PR #252, opened 2022-08-25
self-consistent equilibrium attributes

ddudt:
`Equilibrium.axis` and consequently `eq.Ra_n` & `eq.Za_n` were returning the axis shape from the initial guess, and never got updated after an equilibrium solve. This is now changed so that calling `eq.axis` always evaluates its position at rho=0 to be self-consistent with `eq.R_lmn` & `eq.Z_lmn`. 

Other notes: 
- The axis setter method was removed, since we shouldn't allow the user to assign an axis shape that is inconsistent with the rest of the flux surfaces. 
- Added a test that compares the axis shape to computing R & Z at rho=0. 

**Question:** Should we make a similar change to `Equilibrium.surface`? The same problem exists where the surface attribute could be inconsistent with the flux surfaces shapes (for example, after a boundary perturbation).

### DIFF
--- a/desc/configuration.py
+++ b/desc/configuration.py
@@ -329,9 +329,6 @@ class _Configuration(IOAble, ABC):
         eq_NFP = self.NFP
         surf_NFP = self.surface.NFP if hasattr(self.surface, "NFP") else self.NFP
         axis_NFP = self._axis.NFP
-        print("eq:", eq_NFP)
-        print("surf:", surf_NFP)
-        print("axis:", axis_NFP)
 
         if not (eq_NFP == surf_NFP == axis_NFP):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


desc/configuration.py
```diff
@@ -348,6 +340,16 @@ def __init__(
         if "L_lmn" in kwargs:
             self.L_lmn = kwargs.pop("L_lmn")
 
+        # ensure number of field periods agree
+        eq_NFP = self.NFP
+        surf_NFP = self.surface.NFP if hasattr(self.surface, "NFP") else self.NFP
+        axis_NFP = self.axis.NFP
```
f0uriest, line 346: with the new changes i think this should use self._axis.NFP, since this is checked before setting the initial guess

f0uriest (approved):

dpanici (approved):

PR #254, opened 2022-08-30
Allow hdf5 to store None attributes correctly

f0uriest:
As discussed, this fixes the problem of ``None`` attributes not being stored correctly (since hdf5 doesn't have a null type). It just stores it as the string ``"None"`` and then parses it on the loading side. Also refactored the hdf5 reader to avoid some duplicated code.

### DIFF
--- a/tests/test_input_output.py
+++ b/tests/test_input_output.py
@@ -13,6 +13,7 @@ from desc.utils import equals
 from desc.grid import LinearGrid
 from desc.basis import FourierZernikeBasis
 from desc.transform import Transform
+from desc.equilibrium import Equilibrium
 
 
 def test_vmec_input(tmpdir_factory):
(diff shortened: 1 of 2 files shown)


tests/test_input_output.py
```diff
@@ -273,3 +274,12 @@ def test_copy():
         rtol=1e-10,
         atol=1e-10,
     )
+
+
+def test_save_none(tmpdir_factory):
+    tmpdir = tmpdir_factory.mktemp("none_test")
+    eq = Equilibrium()
+    eq._iota = None
```
ddudt, line 282: `eq._iota = None` will be redundant with the latest changes on `constrain-toroidal-current-density`, which default to vacuum profiles with `iota  = None`

ddudt (dismissed):

tests/test_input_output.py
f0uriest: yeah it's just something to make sure this test actually tests the correct thing regardless of what defaults we pick

ddudt: makes sense

ddudt (approved):

PR #256, opened 2022-08-30
Better callback

f0uriest:
Firstly, renames ``objective.callback`` to ``objective.print_value`` since that's really all it does (in the computing parlance, a callback is usually something that gets called each step in an iterative code and returns a boolean to stop computation early if some condition is met).

Also, it attempts to use a compiled method for evaluating the error where possible. However, there is still the issue that the ``ObjectiveFunction`` method calls the sub ``_Objective`` methods in a loop, and those are generally never compiled. So we may want to reconsider that a bit.

### DIFF
--- a/desc/objectives/objective_funs.py
+++ b/desc/objectives/objective_funs.py
@@ -166,7 +166,10 @@ class ObjectiveFunction(IOAble):
             State vector.
 
         """
-        f = self.compute_scalar(x)
+        if self.compiled and self._compile_mode in {"scalar", "all"}:
+            f = self.compute_scalar(x)
+        else:
+            f = jnp.sum(self.compute(x) ** 2) / 2
         print("Total (sum of squares): {:10.3e}, ".format(f))
         kwargs = self.unpack_state(x)
         for obj in self.objectives:
@@ -268,7 +271,7 @@ class ObjectiveFunction(IOAble):
             mode = "scalar"
         elif mode == "auto":
             mode = "lsq"
-
+        self._compile_mode = mode
         # variable values are irrelevant for compilation
         x = np.zeros((self.dim_x,))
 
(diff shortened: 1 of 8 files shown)


desc/objectives/objective_funs.py
```diff
@@ -166,11 +166,14 @@ def callback(self, x):
...
 
         """
-        f = self.compute_scalar(x)
+        if self.compiled and self._compile_mode in {"scalar", "all"}:
+            f = self.compute_scalar(x)
+        else:
+            f = jnp.sum(self.compute(x) ** 2) / 2
         print("Total (sum of squares): {:10.3e}, ".format(f))
         kwargs = self.unpack_state(x)
         for obj in self.objectives:
```
f0uriest, line 175: @ddudt this is the issue mentioned above. If we want to print the values of the individual objectives, there's always going to be some overhead since we only ever compile the combined compute function

dpanici (approved):
seems good but do have a question on where the overhead for the compute of individual objectives comes from. If it is due to jit compiling can't we just not jit compile those individual functions?

desc/objectives/objective_funs.py
f0uriest: Sort of. Whether you jit it and use it once or don't jit at all, either way that function eval takes a relatively long time (in some cases one time jit is faster than no jit) since we don't use it anywhere else so can't amortize the cost. It's not a ton of overhead, maybe a few seconds. But depends on how detailed we want the printout to be if there are multiple objectives

dpanici: hm ok, I would think it would be useful to have both the overall objective value and the individual ones, so you can see if maybe the overall objective decrease was due only to like one sub-objective decreasing.

Though if there are a lot of sub-objectives maybe this would be less useful

unalmis (approved):

desc/objectives/objective_funs.py
ddudt: I think it's fine to leave it as is for now. If we are really concerned about computation speed we can run with `verbose=0` to suppress the callback output anyways.

PR #261, opened 2022-09-12
Plot results action

dpanici:
- bump matplotlib versions to tested compatible versions `matplotlib>=3.3.0,<=3.6.0,~=3.4.3`
- change github action so that the `pytest-mpl` results are always published to the github, even on test failure (which is maybe the time when we care the most about how the tests failed...)
- adds `norm_F` to `plot_FSA`
- tightens tolerances on all tests, but especially 1D plots
- changes equilibrium used in plot tests to DSHAPE to avoid inconsistencies in SOLOVEV energy minimization
- fix `matplotlib` version to `3.5.0` in CI 
- make DESC input file reader agnostic to the case of the input options (i.e. `spectral_indexing=ANSI` in the input file will work now and register as `ansi` internally)
- update docs to reflect automatic continuation parameters in VMEC to DESC conversion
- update installation docs
- add a github workflow that triggers when you add the `test_mpl` label to a PR that tests versions 3.0.0 - 3.6.0 of matplotlib against our plotting tests to see which versions of matplotlib will pass our plotting tests


dpanici:
good news: it uploads the plot test results even if it fails
bad news: test tolerances of 50 are really high, this failed test had a rms image diff of 60, which is barely above the default threshold of 50:
image


while this one DID pass the 50 tolerance... yet is plotting a completely different quantity than is expected:

image

dpanici:
also unclear if these will only upload the results of build 3.9...

unalmis (dismissed):

unalmis:
We should reduce the tolerances then. 
- Looks like we shouldn't be using high tolerance on 1-d plots since most of the plot is the white canvas background.
- I wonder if the minor numerical difference in the scale's text accounts for most of the error tolerance. maybe testing relative and absolute difference separately would allow lowering tolerance more.
- For example, we could use `@pytest.mark.mpl_image_compare(remove_text=True)` to just compare relative differences (color / curve shape). Then run a numerical test on the axis scale to see if it's in the range we want.

f0uriest:
We could also look at fixing the mpl version for the ci tests to minimize variability

dpanici:
TODO list for this PR:
- fix mpl version for CI
- reduce tolerances for plot_1d tests to make them actually useful
- add norm_F option to `plot_fsa` function (it would be useful and might as well add it to this plotting PR)

dpanici:
> We could also look at fixing the mpl version for the ci tests to minimize variability

From looking at the CI, the same version (3.4.2) of matplotlib and mpl (0.16.1) are used for each of the tests (and these match what I have locally), so it might not be that

dpanici:
Tests seem better with DSHAPE as the base, but sometimes images do not align well or the axes labels change... solution could be to specify the levels for 2D plots?

f0uriest:
``pytest-mpl`` suggests also fixing the version of FreeType which is used to render text (might be redundant with removing text from plots before comparison)
https://github.com/matplotlib/pytest-mpl#package-version-dependencies

dpanici:
unfortunately not immediately obvious to me how to pin [Freetype ](http://freetype.org/)version as it is not a python package but instead a C binary

f0uriest:
> unfortunately not immediately obvious to me how to pin [Freetype ](http://freetype.org/)version as it is not a python package but instead a C binary

ok, from what I've been reading its likely that's not a source of errors anyways. We can always remove text if needed.

f0uriest:
Matt Landreman had asked about using more recent versions of mpl, which I don't think we've tested, this PR might be a good opportunity to try with 3.5 or 3.6 and make sure that works (and pin that version for CI testing)

dpanici:
mpl 3.6 seems to work
only changes we needed to make were to add an `ax` argument to `fig.colorbar()` for the 3D plot functions `plot_3d` and `plot_coils`

do we want to make the version fixed to 3.6 for the main code too? 

I can check if earlier versions of matplotlib work with these changes to see what sort of range we can put but also I cannot figure out a smarter way to change the _CI_requirements.txt to fix mpl==3.6.0 without just pasting all the requirements.txt in there and fixing matplotlib==3.6.0 that way. might just do that

dpanici:
Other notes
- starting in 3.7 `fig.set_tight_layout()` is deprecated and will be replaced with `figure.set_layout_engine('tight')`
- also in 3.7, `matplotlib.cm.get_cmap;` use instead `matplotlib.colormaps[name] instead` (`name` here being the name of the colormap e.g. `'viridis'`
-

f0uriest:
Yeah we should try to figure out a range of MPL versions that work and put those in the requirements file.
For testing, it might be easier to just use the regular requirements file, and then manually install mpl==3.6.0 or whatever after calling pip install in the github workflow

dpanici:
TODO:
Check the following mpl versions to see which work (work as in will pass against the fixed mpl version 3.6.0? or work as in give a reasonable plot? bc the latter could be true but the plot tests will still fail if the former is not true, i.e. could be just a slight misalignment or resizing of one of the elements)
- [ ] 3.3
- [ ] 3.4
- [ ] 3.5
- [ ] 3.6
Ah also seems mpl 3.6.0 has python 3.8 as its [minimal python version required](https://matplotlib.org/stable/devel/min_dep_policy.html).  We should either fix to 3.5.0 instead of 3.6.0, or stop testing python 3.7.
our numpy min is still good enough for the mpl min versions required, at least up to current mpl 3.6.0

f0uriest:
> TODO: Check the following mpl versions to see which work (work as in will pass against the fixed mpl version 3.6.0? or work as in give a reasonable plot? bc the latter could be true but the plot tests will still fail if the former is not true, i.e. could be just a slight misalignment or resizing of one of the elements)
> 
>     * [ ]  3.3
> 
>     * [ ]  3.4
> 
>     * [ ]  3.5
> 
>     * [ ]  3.6
>       Ah also seems mpl 3.6.0 has python 3.8 as its [minimal python version required](https://matplotlib.org/stable/devel/min_dep_policy.html).  We should either fix to 3.5.0 instead of 3.6.0, or stop testing python 3.7.
>       our numpy min is still good enough for the mpl min versions required, at least up to current mpl 3.6.0

yeah im planning on adding python 3.10 and removing 3.7 (which i believe will be deprecated soon)

unalmis:
Could be useful to test the color scale [on these types of plots](https://github.com/PlasmaControl/DESC/pull/261/files#diff-2a4d377329aaa46f339823add962c572e5206b89c82bf371c0b7f73d121b6c82) numerically.

```python
ymin, ymax = axes.get_ylim()
assert ymin not too small
assert ymax not too big
```
- [relevant stack overflow]`https://stackoverflow.com/questions/26131607/matplotlib-get-ylim-values`
- [get_ylim api](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.get_ylim.html)
- [get all axis limits](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.axis.html)

dpanici:
> Could be useful to test the color scale's magnitude [on these types of plots](https://github.com/PlasmaControl/DESC/pull/261/files#diff-2a4d377329aaa46f339823add962c572e5206b89c82bf371c0b7f73d121b6c82) numerically.
> 
> ```python
> ymin, ymax = axes.get_ylim()
> assert ymin not too small
> assert ymax not too big
> ```
> 
> * [relevant stack overflow](https://stackoverflow.com/questions/26131607/matplotlib-get-ylim-values)
> * [get_ylim api](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.get_ylim.html)
> * [get all axis limits api](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.axis.html)

yea I am not sure the better way to go about it, fixing the colorscale ylims so that if the values plotted are the same then they should be the same, or to let it float and check they are reasonable ranges. I am leaning towards former as it will then make the tests more consistent.
Must say I am tired of this PR tho

unalmis:
> > Could be useful to test the color scale's magnitude [on these types of plots](https://github.com/PlasmaControl/DESC/pull/261/files#diff-2a4d377329aaa46f339823add962c572e5206b89c82bf371c0b7f73d121b6c82) numerically.
> > ```python
> > ymin, ymax = axes.get_ylim()
> > assert ymin not too small
> > assert ymax not too big
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * [relevant stack overflow](https://stackoverflow.com/questions/26131607/matplotlib-get-ylim-values)
> > * [get_ylim api](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.get_ylim.html)
> > * [get all axis limits api](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.axis.html)
> 
> yea I am not sure the better way to go about it, fixing the colorscale ylims so that if the values plotted are the same then they should be the same, or to let it float and check they are reasonable ranges. I am leaning towards former as it will then make the tests more consistent. Must say I am tired of this PR tho

yea the former is better, no need to change anything then

dpanici:
Final TODO for this PR:
- [ ] see if can figure out why 3.4.3 and 3.4.0 mpl versions fail, bc then we can say we only need mpl >= 3.3.0
- [ ] maaybe fix colorbar limits but I don't think it is necessary unless that is the reason why above versions are failing.

dpanici:
image
3.4.0 fails just bc the images are slightly shifted from eachother... just going to raise the tolerance on that and say that one is fine. 

3.4.3 we get some sort of an error with the colorbar that I don't think is important to worry about, if we ignore that we can say we are compatible with `matplotlib>=3.3.0` excepting only `3.4.3`

f0uriest:
Sounds good to me

f0uriest:
everything seems ok, though [tests/baseline/test_fsa_F_normalized_vac.png](https://github.com/PlasmaControl/DESC/pull/261/files#diff-d369bcc1b86a833544fd84a91755db76da2ba10e1240b0c00695b02667f6d2d3) seems blank?

dpanici:
> everything seems ok, though [tests/baseline/test_fsa_F_normalized_vac.png](https://github.com/PlasmaControl/DESC/pull/261/files#diff-d369bcc1b86a833544fd84a91755db76da2ba10e1240b0c00695b02667f6d2d3) seems blank?

good catch, forgot to change the ylim from the SOLOVEV vac case

unalmis (approved):

.github/workflows/pytest.yml
```diff
@@ -35,12 +35,14 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r devtools/dev-requirements.txt
+          pip install matplotlib==3.5.0
```
ddudt, line 38: Why `3.5.0` here if mpl tests uses up to `3.6.0`?

devtools/dev-requirements_conda.yml
```diff
@@ -0,0 +1,45 @@
+name: desc-env
```
ddudt, line 1: When we add new requirements, do we have to add them to both `devtools/dev-requirements.txt` and `devtools/dev-requirements_conda.yml`?

docs/installation.rst
```diff
@@ -54,33 +67,27 @@ More info can be found here: https://jax.readthedocs.io/en/latest/developer.html
...
 
-Then install python dependencies:
-
-.. code-block:: bash
-
-   conda create --name desc-env python=3.8 # create a new conda virtual environment
-   conda activate desc-env
-   # install what you can of the requirements with conda, ends up being all but jax, jaxlib and nvgpu
-    conda install colorama "h5py>=3.0.0" "matplotlib>=3.0.0,<=3.4.2" "mpmath>=1.0.0" "numpy>=1.2.0" "netcdf4>=1.5.4" psutil "scipy>=1.5.0" shapely termcolor
+    conda create --name desc-env python=3.8
```
ddudt, line 77: Don't we test on up to Python 3.9? If the IBM power architecture requires 3.8 for some reason, we should make an explicit note of that.

requirements_conda.yml
```diff
@@ -0,0 +1,17 @@
+name: desc-env
```
ddudt, line 1: Same as earlier comment: do new requirements have to be added to both `requirements.txt` and this file?

ddudt (commented):
Additional comments on testing figures: 
- `test_3d_J.png` -- The new figure looks weird at zeta=0
- `test_plot_comparison_no_theta.png` -- Where is legend entry 0 in the new figure? 
- `test_plot_field_line.png` -- The field line looks like it was changed from a rational surface to an irrational surface. Is this because it was changed from using SOLOVEV to DSHAPE? 
- `test_plot_field_lines.png` -- The old plot appears to show 3 separate rational field lines; the new plot appears to show a single field line with different colors for each toroidal transit. Is that the intended new functionality?

.github/workflows/pytest.yml
f0uriest: mpl 3.6.0 has a minimum python version of 3.8, while we still run tests on 3.7 (for now). The point here is just to pin to (any) reasonable version of mpl for testing consistency, with the assumption that different mpl versions will all work with possible minor differences in plot alignment etc.

docs/installation.rst
unalmis: Btw besides removing shapely and fixing a typo in the numpy version, I decided not to change the install instructions for the IBM power architecture as I can't test it.

devtools/dev-requirements_conda.yml
unalmis: Yes. One is for if you do `pip install -r requirements.txt`. The other is when you use conda environments instead. I could write a shell script to sync changes between them in `devtools/`. That wouldn't make it in this pr.

I think the conda version is nice because it sets up the environment for you. It's worse to manually create a conda environment then do `conda install` and `pip install ` because of package dependency issues. I figured if people aren't installing desc with `pip install desc-opt`, they are probably using conda anyway.

requirements_conda.yml
unalmis: [Yea](https://github.com/PlasmaControl/DESC/pull/261/files#r978885865)

PR #262, opened 2022-09-19
VMEC wout conversion

ddudt:
Fixes a few issues when saving a DESC solution in the VMEC "wout" format: 
- There was a bug with using `Transform.fit()` for double Fourier series on multiple surfaces simultaneously. Performing the fit one surface at a time corrects this, but there could be room for speed improvements. 
- Rescale the Jacobian saved as `gmnc` & `gmns` to reflect the VMEC radial coordinate convention of `s = rho^2`. 
- Add the Boozer currents `I` and `G` and the Mercier stability fields to the output data, and associated tests. 
- Add tests for previously untested quantities, such as Lambda and Jacobian. 
- Records the other output data that is still missing from the wout file and needs to be added. 

Also fixes a few issues when reading a VMEC input file:
- Now allows for comma-separated lists of numbers, such as: `AC = 1.0, 0.5, 0.2`
- Now allows for non-stellarator symmetric axis initial guesses using the inputs `RAXIS_CS` and `ZAXIS_CC`


desc/vmec.py
```diff
@@ -328,14 +329,23 @@ def save(cls, eq, path, surfs=128, verbose=1):
         ai = file.createVariable("ai", np.float64, ("preset",))
         ai.long_name = "rotational transform coefficients"
         ai[:] = np.zeros((file.dimensions["preset"].size,))
-        # only using up to 10th order to avoid poor conditioning
-        ai[:11] = PowerSeriesProfile.from_values(
-            s_full, eq.iota(r_full), order=10, sym=False
-        ).params
+        if eq.iota is not None:
+            # only using up to 10th order to avoid poor conditioning
+            ai[:11] = PowerSeriesProfile.from_values(
```
dpanici, line 334: if `iota` here is already a `PowerSeriesProfile` in `rho`, couldn't we just take its coefficients and divide its mode numbers by 2 to get the corresponding series in `s` exactly? instead of fitting a power series

desc/vmec.py
```diff
@@ -328,14 +329,23 @@ def save(cls, eq, path, surfs=128, verbose=1):
...
+            ).params
+        else:
+            ai[:] = np.zeros((file.dimensions["preset"].size,))
 
         ac = file.createVariable("ac", np.float64, ("preset",))
         ac.long_name = "normalized toroidal current density coefficients"
-        ac[:] = np.zeros((file.dimensions["preset"].size,))
+        if eq.current is not None:
+            # only using up to 10th order to avoid poor conditioning
+            ac[:11] = PowerSeriesProfile.from_values(
```
dpanici, line 344: same comment on fitting vs using an existing `PowerSeriesProfile` coefficients

unalmis:
I had added some code to vmec.py last week or a couple weeks before. Might be good to merge that from fix current branch to avoid redoing something.

desc/vmec.py
ddudt: The only downside to your recommendation is it adds more specific logic to maintain. If the underlying profile is already a power series, then the fitting should already work pretty well.

f0uriest:
@ddudt ok, after trying a whole bunch of stuff that didn't work I tried swapping to use solovev for the vmec tests and that seems to just barely work. I'd recommend re-running it with ``ns`` reduced to 64 to be safe (the fact that python 3.7 took 1 hr and 3.8/3.9 took 1.5 hrs makes it seem like its scraping the memory limit).
Also I had to remove one of the tests for the mercier stuff, you should take a look and see if that's something we can fix.

Otherwise it looks like the only outstanding thing would be adding tests for the new stuff in the ``InputReader``

f0uriest (approved):

dpanici (approved):

PR #202, opened 2022-05-24
Jacobian sign fixes

f0uriest:
Ensures DESC always works in a coordinate system with positive coordinate jacobian.

The sign of the jacobian is determined by the orientation of theta. If theta increases clockwise, it is right handed (positive jacobian), otherwise it is left handed (negative jacobian)

When the jacobian is negative, (ie left handed coordinates) some of our equations are incorrect and some of the output vector quantities (B,J,etc) are off by a sign.

To ensure consistency, we now convert any inputs to have a right handed system (ie, theta increasing clockwise). This is summarized below:

- sign( $B_\phi$ ) determined by $\Psi$
- $\Psi$ * $\iota$ > 0 = $B_{poloidal}$ goes clockwise = $J_\phi$ > 0

| $\Psi$ | $\iota$ | $B_{\phi}$ | $B_{poloidal}$ | $J_\phi$ |
|---|---|---|-----|---|
| - | - | - | CW  | + |
| - | + | - | CCW | - |
| + | - | + | CCW | - |
| + | + | + | CW  | + |

TODO:

- [x] fix signs in outputs saved in `desc.examples`
- [x] fix signs in inputs in `examples/DESC`
- [x] double check vmec output
- [x] regenerate precise QH test benchmarks with correct jacobian

PR #267, opened 2022-09-21
Testing

### LINKED ISSUE #265: Streamline Tests
Tests are taking over an hour to run now, this is fine but might be worth checking if we can get same coverage with lighter tests, a good start would be to go in reverse order of the individual test runtimes and see if reducing resolution for tests that don't try to be accurate and compare to VMEC can help

f0uriest:
Upgrades to testing:

- Add full testing of "precise_QH" optimization (broken into stages for parallelization)
- Add tests to compare different methods of computing vacuum heliotron (fixed current vs minimizing J)
- Add testing for python 3.10
- Remove testing for python 3.7
- Mark tests as unit or regression 
  - **NOTE**: all tests must be marked with either `@pytest.mark.unit` or `@pytest.mark.regression` - unmarked tests will not be run as part of the CI testing
  - `regression` tests are ones that test large portions of the code against VMEC/STELLOPT/SIMSOPT base cases - these may only be run occasionally.
  - `unit` tests are for smaller tests that only test certain functionality - these will be run every time.
  - `unit` tests should make limited use of fixtures, currently only `DSHAPE_current`, `SOLOVEV`, and `HELIOTRON_vac` are used in the unit tests.
- `regression` tests are only run on Python 3.9
- `unit` tests are run on Python 3.8 and 3.10
- Parallelize tests on CI using [pytest-split](https://github.com/jerry-git/pytest-split) - currently split into 4 groups for regression tests and 3 groups for unit tests. This can be tweaked in the future to more evenly distribute the load
- Get rid of use of ``unittest`` in favor of always using ``pytest``

Resolves #265


f0uriest:
As part of this PR, I'm reworking the test fixtures so that when running unit tests we only run 1-2 fixtures, ideally 1 vacuum stellarator without current constraint, and 1 finite beta tokamak with current constraint. All the unit tests will just use those as fixtures, and the others will be left for regression tests. Sound reasonable?

dpanici:
> As part of this PR, I'm reworking the test fixtures so that when running unit tests we only run 1-2 fixtures, ideally 1 vacuum stellarator without current constraint, and 1 finite beta tokamak with current constraint. All the unit tests will just use those as fixtures, and the others will be left for regression tests. Sound reasonable?

Seems reasonable to me, no need to have unit tests cause every regression equilibrium to be ran

dpanici:
would be useful once tests pass to do memory profiling with https://pypi.org/project/pytest-monitor/

which does not work with Unittest apparently

f0uriest:
@unalmis Would it be possible to have the stability tests use ``DSHAPE_current`` and ``HELIOTRON_vac`` instead of the regular ``DSHAPE`` and ``HELIOTRON``? I'm trying to reduce the number of solves needed for the core tests, but not sure if those cases work well as tests for the mercier stuff.

unalmis:
> @unalmis Would it be possible to have the stability tests use ``DSHAPE_current`` and ``HELIOTRON_vac`` instead of the regular ``DSHAPE`` and ``HELIOTRON``? I'm trying to reduce the number of solves needed for the core tests, but not sure if those cases work well as tests for the mercier stuff.

DSHAPE_current should work fine. It converges to the same equilibrium as regular DSHAPE. The tolerances may need to be changed though since any tweak can change numbers on the mercier stuff.

A vacuum case would not be a good test, even ignoring the fact that D_well goes to 0.

How about using one live solve and the other from examples.get()?

f0uriest:
Ah yeah that's a good idea. I'll give that a try

unalmis:
Btw, ` desc/examples/DSHAPE_current` needs "current" to be all caps for `examples.get()` to work.

ddudt (approved):
Comments on changed figures:
- `test_fsa_F_normalized_vac.png` the ylim should be changed because part of the plot is cut off.
- `test_plot_comparison.png` & `test_plot_comparison_no_theta.png` why were the number of equilibria being compared reduced from 3 to 2? 

See other comments in code. Otherwise it looks good.

.test_durations
```diff
@@ -0,0 +1,293 @@
+{
+    "tests/test_backend.py::TestBackend::test_put": 0.16319112002383918,
```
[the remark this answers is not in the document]
f0uriest: This is needed by ``pytest-split`` in order to split the tests into groups of approximately equal duration. We should update it occasionally but it doesn't need to be on every change. By default it will assume any tests not listed here take the average amount of time (of the times listed).

desc/io/ascii_io.py
```diff
@@ -12,10 +12,17 @@ def write_ascii(fname, eq):
         dictionary of equilibrium parameters.
 
     """
-    if eq.pressure.__class__.__name__ != "PowerSeriesProfile":
-        raise TypeError("Equilibrium must have power series profiles for ascii io")
-    if eq.iota.__class__.__name__ != "PowerSeriesProfile":
-        raise TypeError("Equilibrium must have power series profiles for ascii io")
+    # can't import other stuff in io due to circular imports, so have to check by name
+    assert (
+        eq.pressure.__class__.__name__ == "PowerSeriesProfile"
```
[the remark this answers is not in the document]
f0uriest: Yeah, the only reason for it was DCON3D, not sure if we should still keep it for that, or just remove it.

tests/conftest.py
```diff
@@ -204,6 +222,64 @@ def HELIOTRON_vacuum(tmpdir_factory):
...
+        "desc_nc_path": desc_nc_path,
+        "vmec_nc_path": vmec_nc_path,
+        "booz_nc_path": booz_nc_path,
+    }
+    return HELIOTRON_vacuum2_out
+
+
+@pytest.fixture(scope="session")
+def precise_QH(tmpdir_factory):
+    """Fun initial condition for precise QH optimization."""
```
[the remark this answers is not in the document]
f0uriest: This was your docstring actually, I'd prefer to keep it as fun fwiw.

tests/inputs/input.QSC_r2_5.5_vmec_desc
```diff
@@ -0,0 +1,3218 @@
+# This DESC input file was auto generated from the VMEC input file
+# /home/fouriest/SCHOOL/Princeton/PPPL/DESC/tests/inputs/input.QSC_r2_5.5_vmec
+# on 09/27/2022 at 16:40:31.
+# For details on the various options see https://desc-docs.readthedocs.io/en/stable/input.html
```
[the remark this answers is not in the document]
f0uriest: No I think this was committed by accident.

tests/test_plotting.py
```diff
@@ -35,404 +35,564 @@
...
 
 
+# @pytest.mark.unit
+# @pytest.mark.solve
+# @pytest.mark.mpl_image_compare(tolerance=50)
+# def test_plot_qh_optimization(precise_QH):
+#     """Test plotting B in boozer coordinates from precise QH optimization."""
+#     eq = EquilibriaFamily.load(load_from=str(precise_QH["optimal_h5_path"]))
+#     fig, ax = plot_boozer_surface(eq)
+#     return fig
```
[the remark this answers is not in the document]
f0uriest: Up to you. Does this exercise code that isn't otherwise tested, or verify properties of the precise qh optimization we can't get with the other tests (I believe they just check that the non-axisymmetric Boozer modes are small). 
If we want to keep it it's probably best to combine it with the `test_precise_QH_optimization3` test.

devtools/dev-requirements_conda.yml
```diff
@@ -39,6 +39,7 @@ dependencies:
   - pytest-benchmark
   - pytest-cov >= 2.6.0
   - pytest-mpl = 0.16.1
```
unalmis, line 41: This one's missing `pytest-monitor` that the pip requirements have.

tests/benchmarks/benchmark.py
```diff
@@ -107,7 +107,7 @@ def test_DSHAPE_run(tmpdir_factory, benchmark):
 @pytest.mark.benchmark(min_rounds=1, max_time=300, disable_gc=True, warmup=False)
 def test_DSHAPE_current_run(tmpdir_factory, benchmark):
     """Benchmark the DSHAPE fixed toroidal current example."""
-    input_path = ".//tests//benchmarks//DSHAPE_current"
+    input_path = ".//tests//inputs//DSHAPE_current"
```
unalmis, line 110: Now that `tests/benchmarks` actually has these files, we should revert this change.

tests/benchmarks/benchmark.py
```diff
@@ -147,7 +147,7 @@ def test_HELIOTRON_run(tmpdir_factory, benchmark):
 @pytest.mark.benchmark(min_rounds=1, max_time=300, disable_gc=True, warmup=False)
 def test_HELIOTRON_vacuum_run(tmpdir_factory, benchmark):
     """Benchmark the HELIOTRON vacuum (fixed current) example."""
-    input_path = ".//tests//benchmarks//HELIOTRON_vacuum"
+    input_path = ".//tests//inputs//HELIOTRON_vacuum"
```
unalmis, line 150: Now that tests/benchmarks actually has these files, we should revert this change.

tests/benchmarks/benchmark.py
f0uriest: I'd rather keep all the inputs in the `inputs` folder, I'll modify the benchmarks PR to use those same files so we don't have dupes.

devtools/dev-requirements_conda.yml
f0uriest: good catch

f0uriest:
> Comments on changed figures:
> 
>     * `test_fsa_F_normalized_vac.png` the ylim should be changed because part of the plot is cut off.
> 
>     * `test_plot_comparison.png` & `test_plot_comparison_no_theta.png` why were the number of equilibria being compared reduced from 3 to 2?
> 
> 
> See other comments in code. Otherwise it looks good.

Just realized `test_fsa_F_normalized_vac` is actually the same as `test_fsa_F_normalized` since they're using the same equilibria. I'd rather keep all the plotting tests using the same equilibria and vacuum tokamaks don't really make sense, so I removed that test for now.

The comparison plots changed because I reduced the `DSHAPE_current` example solve from 3 steps to 2 to speed things up

unalmis (approved):

tests/test_plotting.py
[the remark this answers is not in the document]
f0uriest: this is being added back as part of the `test_precise_QH_optimization3` test, just need to regenerate the baseline image.

.github/workflows/unittest.yml
```diff
@@ -23,7 +23,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ['3.8', '3.10']
```
dpanici, line 26: May be a non-issue but why do we not run unit tests on 3.9?

Is it that we assume our regression tests are exercising all the things the unit tests would anyways, so we don't need to unit test on 3.9?

Is it not possible for some unit test that covers a functionality that is not exercised by regression tests fails on 3.9, which then would not be caught? Or am I overthinking it.

desc/io/ascii_io.py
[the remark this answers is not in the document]
dpanici: I'd vote in favor of keeping, if DCON3D ever gets to the point that it is used in teh community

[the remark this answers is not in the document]
f0uriest: OK yeah, I'm in favor of keeping it for now, at least until we figure out what's happening with DCON. In the future we may need to add current profiles, but that would also require adding that capability on the DCON side so I think we're ok restricting it to fixed iota for now.

.github/workflows/unittest.yml
f0uriest: We've never actually had any issues with different python versions, and it's probably overkill to test with multiple versions. I'm figuring if there are any issues due to python versions they'll be pretty noticeable, so splitting the versions between unit and regression tests seems fine. Going forward it would probably make more sense to use 1-2 versions of python and possibly test with different versions of jax/numpy/matplotlib etc. But in that case the number of jobs gets pretty huge so I think we're better off only doing that at specific times (like we do now with the jax and mpl version test actions), not on every PR.

dpanici: That's fair, yea I think you're right in it is mainly different jax/mpl versions that cause issues, not different python versions

f0uriest: As for this specifically, I just removed it to reduce the number of jobs running (we're only allowed 20 concurrent jobs for the whole PlasmaControl organization), since adding `pytest-split` runs a lot of stuff in parallel. Right now unit tests will take 6 jobs (2 python versions x 3 groups) and regression tests will take 4 jobs (1 python version x 4 groups). We can play around with this more in the future if needed.

ddudt (approved):

tests/test_plotting.py
```diff
@@ -35,404 +35,543 @@
...
     fig, ax = plot_section(eq, "R")
     return fig
 
 
+@pytest.mark.unit
+@pytest.mark.solve
 @pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_2d)
-def test_section_F(DSHAPE):
-    eq = EquilibriaFamily.load(load_from=str(DSHAPE["desc_h5_path"]))[-1]
+def test_section_F(DSHAPE_current):
```
dpanici, line 251: the new figure for this has only 2 orders of magnitude on the colorscale, is the new soln just have a radial force component that spans less orders of magnitude?

f0uriest: Yeah it's slightly worse in terms of force error, I reduced the resolution and number of continuation steps a bit to make it faster.

tests/test_plotting.py
```diff
@@ -35,404 +35,543 @@
...
     fig, ax = plot_2d(eq, "|F|", log=True, grid=grid, figsize=(4, 4))
     return fig
 
 
+@pytest.mark.unit
+@pytest.mark.solve
 @pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_2d)
-def test_2d_g_tz(DSHAPE):
-    eq = EquilibriaFamily.load(load_from=str(DSHAPE["desc_h5_path"]))[-1]
+def test_2d_g_tz(DSHAPE_current):
```
dpanici, line 103: the jacobian of the new current-solved DSHAPE seems the opposite sign of the old one? not really a problem just asking as the color scheme seems inverted

dpanici (changes requested):
Looks good, left just a couple comments but before I approve can you update the CONTRIBUTING.rst to say that for tests you **must put the pytest markers `unit` or `regression` in order for the tests to be ran by CI**?

And maybe the command to save the pytest time log as well for `pytest-split`, it is just nice to have the relevant test commands and info in one spot. But if you think it is unnecessary then that's fine

tests/test_plotting.py
f0uriest: Yeah it's due to some sign issues. The regular `DSHAPE` example should give positive current, but because of the sign issues (see #202) it's negative. So the `DSHAPE_current` example has negative input current, which then gives negative iota, flipping the helicity compared to the regular `DSHAPE`

f0uriest:
Yeah good call. I added that and a few other things to the contributing guide. I left out instructions for getting the test durations, since we won't need to do it that often, and also because we're probably better off just manually editing that file to balance things better (running it locally I got very different results from what we see on CI, so I ended up manually adjusting some of the times in there to better match what we see on CI so it will balance more evenly, but one of the biggest issues is just the variety of hardware the CI tests use, it can lead to a factor of 2x speedup or slowdown).

ddudt (approved):

unalmis (approved):

dpanici (approved):

Issue #268, opened 2022-09-22
Loading in Equilibria saved before constraint-toroidal-current PR results in errors due to nonexistent eq.current property


dpanici:
Per title, if loading in an equiilbrium solved before the toroidal current constraint PR, it won't have a `self.current` property, and so will result in an error if you try to compute anything from it

can easily fix by having a `_set_up()`  function for Configuration to check if the loaded eq has a `current` property, and if it does not it sets it to `None`

PR #269, opened 2022-09-22
add _set_up() to Configuration to add a self.current property if miss…

dpanici:
…ing after an Equiilbrium is loaded (for backwards compatibility with older versions of DESC prior to the toroidal current constraint PR

addresses #268


f0uriest (approved):
Good catch.

f0uriest (dismissed):

desc/configuration.py
```diff
@@ -396,6 +396,12 @@ def __init__(
 
     # TODO: allow user to pass in arrays for surface, axis? or R_lmn etc?
     # TODO: make this kwargs instead?
+    def _set_up(self):
+        """Called after loading, to ensure object has all properties needed for current DESC version.
+        Allows for backwards-compatibility with equilibria saved/ran with older DESC versions"""
+        if not hasattr(self, "current"):
```
f0uriest, line 402: A possibly more general fix would be something like:
```
for atttr in self._io_attrs_:
    if not hasattr(self, attr):
    self.attr = None
```
(i forget the correct syntax)

dpanici: good idea, changed it

f0uriest (approved):

PR #272, opened 2022-09-23
Remove use_jit from Derivative, move to objectives

f0uriest:
Also adds jit to objective.jvp methods which should speed up perturbations a bit


f0uriest:
@dpanici looks like some file names got changed and its breaking the benchmark action?

dpanici (approved):
Seems fine, benchmark run does not seem to show a ton of improvement though

f0uriest:
Yeah it probably isn't too noticeable. Also the benchmarks are pretty noisy, we might want to increase the ``min_rounds`` to 3 or something so we at least get a decent average.

PR #274, opened 2022-09-27
Benchmark upgrades

f0uriest:
- Increase number of rounds for benchmarking to get better statistics
- Add benchmarks for perturbations (1st and 2nd order)
- Fix formatting of PR comment (column alignment and text color)
- Add +/- stdev to PR comment to get better idea of variance in estimated times
- Lines are now highlighted green/red if the dt is within 1 stdev of 0

An example of the new comment format:
```diff
|         benchmark_name          |        dt(%)        |        dt(s)        |      t_new(s)      |      t_old(s)      | 
| ------------------------------- | ------------------- | ------------------- | ------------------ | ------------------ |
+foo1                             | -5.10e+01+/-7.0e+00 | -4.58e+03+/-6.3e+02 | 4.40e+03+/-1.8e+01 | 8.99e+03+/-6.3e+02 |
-foo2                             | +5.16e+01+/-2.0e+01 | +2.03e+03+/-7.7e+02 | 5.96e+03+/-5.1e+02 | 3.93e+03+/-5.8e+02 |
-foo3                             | +3.62e+02+/-3.7e+01 | +5.84e+03+/-6.0e+02 | 7.45e+03+/-3.5e+02 | 1.61e+03+/-4.9e+02 |
 foo4                             | -1.46e+01+/-7.9e+01 | -1.66e+02+/-9.0e+02 | 9.69e+02+/-3.6e+01 | 1.13e+03+/-9.0e+02 |
-foo5                             | +2.96e+02+/-2.3e+02 | +1.04e+03+/-7.9e+02 | 1.39e+03+/-7.6e+02 | 3.50e+02+/-2.2e+02 |
-foo6                             | +3.92e+01+/-1.4e+01 | +2.81e+03+/-1.0e+03 | 9.97e+03+/-7.7e+02 | 7.16e+03+/-6.4e+02 |
```

(Note that somewhat ironically we can't run the benchmarks on this PR since the benchmark actions will be different here vs on master)


dpanici:
that coloring is what I tried to do initially but never could get it to work... I like it

tests/benchmarks/benchmark.py
```diff
@@ -1,10 +1,52 @@
...
+    main(args)
+
+    SOLOVEV_out = {
+        "input_path": input_path,
+        "desc_h5_path": desc_h5_path,
+        "desc_nc_path": desc_nc_path,
+        "vmec_nc_path": vmec_nc_path,
+        "booz_nc_path": booz_nc_path,
+    }
+    return SOLOVEV_out
```
ddudt, line 49: This looks like it is duplicating code from `DESC/tests/conftest.py`. Can we reference that existing code, or does it need to also be included in the benchmark directory for some reason?

f0uriest: Because the benchmarks are in their own folder, pytest does't know about ``conftest.py`` in the parent directory, so we need to include this here as well.

dpanici (approved):

dpanici:
The code cov project failing is due to what? I had the same thing on a previous PR, I thought it was maybe bc I had gotten rid of a test that called the scalar minimization but here we don't mess with any tests at all...

ddudt (approved):

f0uriest:
> The code cov project failing is due to what? I had the same thing on a previous PR, I thought it was maybe bc I had gotten rid of a test that called the scalar minimization but here we don't mess with any tests at all...

Yeah I'm not sure, the line it says changed is in the `Optimizer` class, so no idea why it would have changed from this PR. My hunch is just some bug with coverage or codecov parsing.

PR #276, opened 2022-09-27
Near axis

### LINKED ISSUE #220: Add initial guesses for solutions using Qsc?
For fixed bdry, could take Qsc solution out to rho=0.3 or 0.4 and then scale to the boundary

ddudt:
Adds two new methods:

`Equilibrium.from_near_axis` allows users to load in a solution from `pyQSC` or `pyQIC`. 

`FourierRZToroidalSurface.from_near_axis` allows users to create boundary surfaces that are approximately QP/QI based on an unpublished analytic model shared by Matt Landreman. 

Resolves #220


ddudt:
I would like to extend the `FourierRZToroidalSurface.from_near_axis` method to handle more generic cases, such as with a major radius other than 1 and with non-stellarator symmetric terms. But I think that can wait for a future update.

desc/equilibrium.py
```diff
@@ -265,6 +270,106 @@ def change_resolution(
         if N_grid is not None and N_grid != self.N_grid:
             self._N_grid = N_grid
 
+    @classmethod
+    def from_near_axis(
+        cls, eq, r=0.1, L=None, M=8, N=None, ntheta=None, spectral_indexing="ansi"
```
f0uriest, line 275: could we rename `eq` here to something like near_axis_eq or `naeq` or something? we use `eq` a ton of other places to refer specifically to DESC equilibrium objects

desc/equilibrium.py
```diff
@@ -265,6 +270,106 @@ def change_resolution(
...
+            inputs["pressure"] = np.array([[0, -eq.p2 * r ** 2], [2, eq.p2 * r ** 2]])
+            inputs["iota"] = None
+            inputs["current"] = np.array([[2, 2 * np.pi / mu_0 * eq.I2 * r ** 2]])
+            inputs["axis"] = FourierRZCurve(
+                R_n=np.concatenate((np.flipud(eq.rs[1:]), eq.rc)),
+                Z_n=np.concatenate((np.flipud(eq.zs[1:]), eq.zc)),
+                NFP=eq.nfp,
+            )
+            inputs["surface"] = None
+        except AttributeError:
```
f0uriest, line 331: maybe say what attribute is missing in the error message?

desc/geometry/surface.py
```diff
@@ -469,6 +470,58 @@ def from_input_file(cls, path):
         )
         return surf
 
+    @classmethod
+    def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
```
f0uriest, line 474: are there any sensible defaults for these parameters?

f0uriest (approved):
Assuming the tests pass it looks good to me

dpanici:
requirements is missing `qsc`?

f0uriest:
> requirements is missing `qsc`?

Yeah I'm fixing it and a few other things, will push in a few mins.

devtools/dev-requirements.txt
```diff
@@ -24,6 +24,7 @@ pytest-cov >= 2.6.0
 pytest-mpl == 0.16.1
 shapely >= 1.8.2
 pytest-monitor
+qsc
```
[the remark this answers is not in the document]
f0uriest: I forgot that pickle still needs to import the class on loading, so its kind of moot. But still good to have a consistent baseline saved for testing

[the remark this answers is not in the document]
ddudt: Should we revert the test function back to the old way of type checking instead of the try-except?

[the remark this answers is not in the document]
f0uriest: Eh i still prefer the duck typing way, its more "pythonic"

f0uriest (dismissed):

desc/equilibrium.py
```diff
@@ -265,6 +270,108 @@ def change_resolution(
...
+        na_eq : Qsc or Qic
+            Near-axis solution generated by pyQSC or pyQIC.
+        r : float
+            Radius of the desired boundary surface (in meters).
+        L : int (optional)
+            Radial resolution. Default 2*M for `spectral_indexing`==fringe, else M
+        M : int (optional)
+            Poloidal resolution. Default is 8
+        N : int (optional)
+            Toroidal resolution. Default is the max resolution provided by na_eq.nphi
```
f0uriest, line 290: could this potentially be an issue if nphi is super large?

desc/equilibrium.py
```diff
@@ -265,6 +270,108 @@ def change_resolution(
...
+                R_n=np.concatenate((np.flipud(na_eq.rs[1:]), na_eq.rc)),
+                Z_n=np.concatenate((np.flipud(na_eq.zs[1:]), na_eq.zc)),
+                NFP=na_eq.nfp,
+            )
+            inputs["surface"] = None
+        except AttributeError as e:
+            raise ValueError("Input must be a pyQSC or pyQIC solution.") from e
+
+        rho, _ = special.js_roots(L, 2, 2)
+        grid = LinearGrid(rho=rho, theta=ntheta, zeta=na_eq.nphi, NFP=na_eq.nfp)
```
f0uriest, line 337: do we want to use the "ocs" concentric grid for fitting here? Or at least oversample by default?

desc/equilibrium.py
ddudt: Yes. But I want the option to be able to capture the full toroidal resolution from the near-axis solution if necessary. We could change the default to `N=8` or something?

desc/geometry/surface.py
```diff
@@ -469,6 +470,58 @@ def from_input_file(cls, path):
...
+
+        Returns
+        -------
+        surface : FourierRZToroidalSurface
+            Surface with given geometric properties.
+
+        """
+        a = np.sqrt(elongation) / aspect_ratio  # major axis
+        b = 1 / (aspect_ratio * np.sqrt(elongation))  # minor axis
+        epsilon = (2 - np.sqrt(4 - mirror_ratio ** 2)) / mirror_ratio
```
f0uriest, line 498: should we include a note in the docstring that this needs mirror_ratio < 1?

desc/equilibrium.py
ddudt: This is the radial spacing used in the quadrature grid. We don't necessarily need to oversample.

desc/geometry/surface.py
ddudt: Yes that's a good idea.

desc/equilibrium.py
f0uriest: could maybe make it default to M==N, but have N=np.inf mean use all of them?

desc/equilibrium.py
```diff
@@ -307,7 +308,10 @@ def from_near_axis(
                 elif spectral_indexing == "fringe":
                     L = 2 * M
             if N is None:
-                N = int((na_eq.nphi - 1) / 2)
+                if N == np.inf:
```
f0uriest, line 311: don't think this should be nested like this

f0uriest (approved):

desc/equilibrium.py
```diff
@@ -308,10 +308,10 @@ def from_near_axis(
                 elif spectral_indexing == "fringe":
                     L = 2 * M
             if N is None:
-                if N == np.inf:
-                    N = int((na_eq.nphi - 1) / 2)
-                else:
-                    N = M
+                N = M
+            if N == np.inf:
```
ddudt, line 312: Won't this never be true unless `M=np.inf`? I don't understand this change @f0uriest

desc/equilibrium.py
dpanici: why do we need the quad grid radial spacing here though? isn;'t the ocs supposed to be the optimal sampling grid for fitting zernike

desc/equilibrium.py
f0uriest: or if ``N is not None``

f0uriest: You had it inside the if block, where we already have ``N is None`` so then it would never be `N == np.inf``

desc/geometry/surface.py
```diff
@@ -469,6 +470,61 @@ def from_input_file(cls, path):
         )
         return surf
 
+    @classmethod
+    def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
+        """Create a surface from a near-axis model for quasi-poloidal/quasi-isodynamic.
+
+        Parameters
+        ----------
+        aspect_ratio : float
```
dpanici, line 479: is there a formula in a paper that can be referenced here for this function?

desc/equilibrium.py
ddudt: Oh I see. I'm tired...

desc/geometry/surface.py
ddudt: There is a formula, but it's unpublished. Just a PDF Matt Landreman sent me.

dpanici (approved):

desc/geometry/surface.py
dpanici: Ok, maybe not necessary right now but would be good to remember if it becomes published, I like having formulas for things

Issue #277, opened 2022-09-28
LinearGrids with same nodes have different spacing/weights


f0uriest:
There are 2 methods for creating a grid with a specified number of nodes (as opposed to specified basis resolution). The first (which isn't officially documented) is to just pass integers for `theta`, `zeta` etc, which will linearly space that many nodes, as in `grid1` below. The other is to pass in arrays directly, as is done with `grid2` below.
```
ntheta = 5
nzeta = 4
NFP = 3
grid1 = LinearGrid(rho=1, theta=ntheta, zeta=nzeta, NFP=NFP, endpoint=False)
grid2 = LinearGrid(rho=1, theta=np.linspace(0,2*np.pi,ntheta, endpoint=False), zeta=np.linspace(0,2*np.pi/NFP, nzeta, endpoint=False), NFP=NFP)

print("Nodes agree:", np.allclose(grid1.nodes, grid2.nodes))
print("Spacing agree:", np.allclose(grid1.spacing, grid2.spacing))
```
Gives that the nodes agree but spacing doesn't. 

The spacing for `grid1` seems correct, as it has all nodes having the same spacing and weights. While the spacing for `grid2` seems way off, with `dz` ranging from 0.2 to 4.9

This further causes the weights to be computed incorrectly, possibly giving the wrong averages etc.

dpanici:
![image](https://user-images.githubusercontent.com/37969854/192907536-c9799d59-88d2-4b4f-83fe-911e572a60b0.png)

this is the culprit, the spacing for when it is not given an int is calculated as if it is a finite difference. To be honest I do not understand what this spacing is, if we just want spacing would we not just do 
```dt[0:-1] = t[1:] - t[0:-1]
dt[-1] = 2pi - t[-1]
```
We want spacing to add up to 1 for dr and 2pi for dt and dz. 
- If it is passed in a linearly spaced set of points in any of the coordiantes, the spacing should be uniform
 - this could be easily accomplished by subtracting the second point in the array from the first
 - however this fails for an array that is NOT linearly spaced
  - this begs the question, should we allow such a spacing in the `LinearGrid` class? it seems the logic here is trying to come up with some sort of sensible spacing/weighting for non-uniformly spaced grid points, but maybe those should be put in the `Grid` class and not the `LinearGrid` class anyways

f0uriest:
I think something like `np.diff` could work for a more general case. Just need to be careful with theta and zeta because they're periodic.

[...]

PR #279, opened 2022-09-29
Grid spacing

unalmis:
See #277

### DIFF
--- a/desc/grid.py
+++ b/desc/grid.py
@@ -423,7 +423,6 @@ class LinearGrid(Grid):
             self._L = len(np.atleast_1d(rho))
         if np.isscalar(rho) and (int(rho) == rho) and rho > 0:
             r = np.flipud(np.linspace(1, 0, int(rho), endpoint=axis))
-            # FIXME: endpoint
             dr = 1 / r.size * np.ones_like(r)
         else:
             r = np.atleast_1d(rho)
@@ -432,9 +431,6 @@ class LinearGrid(Grid):
                 dr[0] = r[1] - r[0]
                 dr[1:-1] = (r[2:] - r[:-2]) / 2
                 dr[-1] = 1 - r[-1]
-                if endpoint:
-                    dr[0] = dr[0] / 2
-                    dr[-1] = dr[0]
             else:
                 dr = np.array([1.0])
 
@@ -785,10 +781,8 @@ class ConcentricGrid(Grid):
         drho = np.zeros_like(rho)
         if rho.size > 1:
             # FIXME: I'd assume concentric grid has the same problem?
-            #    but maybe not if there is always a node at rho=1
             #    drho[0] = rho[1] - rho[0]
             #    drho[-1] = 1 - rho[-1]
-            #    endpoint weights
             drho[0] = (rho[0] + rho[1]) / 2
             drho[1:-1] = (rho[2:] - rho[:-2]) / 2
             drho[-1] = 1 - (rho[-2] + rho[-1]) / 2
--- a/desc/grid.py
+++ b/desc/grid.py
@@ -423,14 +423,18 @@ class LinearGrid(Grid):
             self._L = len(np.atleast_1d(rho))
         if np.isscalar(rho) and (int(rho) == rho) and rho > 0:
             r = np.flipud(np.linspace(1, 0, int(rho), endpoint=axis))
+            # FIXME: endpoint
             dr = 1 / r.size * np.ones_like(r)
         else:
             r = np.atleast_1d(rho)
             dr = np.zeros_like(r)
             if r.size > 1:
-                dr[0] = (r[0] + r[1]) / 2
+                dr[0] = r[1] - r[0]
                 dr[1:-1] = (r[2:] - r[:-2]) / 2
-                dr[-1] = 1 - (r[-2] + r[-1]) / 2
+                dr[-1] = 1 - r[-1]
+                if endpoint:
+                    dr[0] = dr[0] / 2
+                    dr[-1] = dr[0]
             else:
                 dr = np.array([1.0])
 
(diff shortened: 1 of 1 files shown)


desc/grid.py
```diff
@@ -447,34 +450,75 @@ def _create_nodes(
             if self.sym:
                 t += t[1] / 2
             dt = 2 * np.pi / t.size * np.ones_like(t)
+            if endpoint and t.size > 2:
+                # When endpoint is True, the theta = 0 node is duplicated at 2 pi.
```
f0uriest, line 454: I think we already do this in `_scale_weights`, should double check that we're not doing it twice

f0uriest: yeah the grids for surfaces and curves use `endpoint=True`, so that's probably what's causing the failing test

unalmis: Yes I have fixed that

desc/grid.py
```diff
@@ -428,9 +428,17 @@ def _create_nodes(
             r = np.atleast_1d(rho)
             dr = np.zeros_like(r)
             if r.size > 1:
-                dr[0] = (r[0] + r[1]) / 2
-                dr[1:-1] = (r[2:] - r[:-2]) / 2
-                dr[-1] = 1 - (r[-2] + r[-1]) / 2
+                axis = np.isclose(r[0], 0)
+                # dr[0] = r[int(axis)]
+                # dr[1:-1] = (r[2:] - r[:-2]) / 2
+                # dr[-1] = 1 * r.size / (r.size - axis) - r[-1 - (not axis)]
```
f0uriest, line 434: I actually think this is more correct. If the nodes aren't evenly spaced for some reason, this will correctly attribute half of the distance from `n-1` to `n` and from `n` to `n+1`

desc/grid.py
```diff
@@ -447,17 +455,33 @@ def _create_nodes(
...
-                dt[0] = (t[0] + t[1]) / 2
+                # in case t[0] != 0, we add it evenly between dt[0] and dt[-1]
+                dt[0] = t[1] - t[0] / 2
                 dt[1:-1] = (t[2:] - t[:-2]) / 2
-                dt[-1] = 2 * np.pi - (t[-2] + t[-1]) / 2
+                dt[-1] = 2 * np.pi - t[-1] + t[0] / 2
+                # maybe below 2 lines would be better than above 3 lines
+                # dt = np.diff(t, append=2 * np.pi + t[0] / 2)
+                # dt[0] += t[0] / 2
+                dt *= (t.size - endpoint) / t.size
```
f0uriest, line 470: wouldn't this give `dt=0` if there's only 1 theta coordinate and `endpoint=True`?

f0uriest: wait nvm i see this is inside the if block

unalmis: Yes, it's in the if block. Line 470 is pretty opaque. Basically, lines 463-466 will assign spacing in a way that already scales the weights down for repeated nodes similar to `_scale_weights`. So if there is an extra node at 2pi, I upscale the weights so that `_scale_weights` can do it's job.

desc/grid.py
unalmis: Ok, I'll use switch it to this. Since the CI tests won't complete on the current run, I had the unit tests run locally and 218/218 passed.

unalmis: Making this change passes 217/218 unit tests, failing `test_vmec_save_1`. Looking into it

unalmis:
The zeta and theta spacing inconsistency is fixed. For one thing this means surface averages should now be correct when an arrays are past into linear grids.

If there's a duplicate node in the array though, I don't think `_scale_weights` changes this correctly (here and on `master`). It scales the total volume weight of the node. In general, if there is a theta node repeated N times, we need to just divide the theta spacing of each of those nodes by N, without modifying the rho or zeta spacing. This will make that repeated node contribute the same theta weight as a non-repeated theta node. The current method with `_scale_weights` modifies the rho and zeta spacing when there is a repeated theta node which can cause drho > 1 and dzeta > 2pi.

All the unit tests passed locally. If the regression tests (excluding codecov) pass, y'all technically could be merge this if you needed it now. It fixes the original issue - wrong zeta and theta spacing when constructing a linear grid with an array.

I reverted the changes for the rho spacing back to what they are on `master`. This means passing an array to make a linear grid may still give different rho spacing when `nrho != 1 `. I think the array spacing is actually more correct.

This notebook will be helpful to see the different spacing computed by the old / new methods.
[grid_spacing.zip](https://github.com/PlasmaControl/DESC/files/9681348/grid_spacing.zip)

desc/grid.py
```diff
@@ -447,31 +450,62 @@ def _create_nodes(
...
+                SUP = 2 * np.pi / self.NFP  # supremum
+                dz[0] = z[1] + (SUP - (z[-1] % SUP)) % SUP
+                dz[1:-1] = z[2:] - z[:-2]
+                dz[-1] = z[0] + (SUP - (z[-2] % SUP)) % SUP
+                dz /= 2
+                dz *= self.NFP
+                if z.size == 2:
+                    dz[-1] = dz[0]
+                # if endpoint and dz[0] == (dz[-1] % (SUP * self.NFP)):
+                #     FIXME: dz already has the correct spacing
```
f0uriest, line 506: Is this something we need to fix now? Do duplicate nodes not get weighted correctly?

unalmis: I put that comment there to remind myself or others to eventually fix it.

Our `_scale_weights` does not reduce duplicate node weight correctly - here or on `master`. The notebook in my above comment can help show what follows.

A simple example of this is `nrho=1, ntheta=5, nzeta=1, endpoint=True`. The correct theta spacing for this `[0.78539816 1.57079633 1.57079633 1.57079633 0.78539816]` with drho = 1 always, dzeta = 2pi always. And at this point in the code that is what the spacing is (also both here and on `master`).

`_scale_weights` then detects the duplicate nodes at 0 and 2pi and scales down the node volume drho * dtheta * dzeta. So the end result of the grid spacing is:
```
[[0.82982653 0.65174424 5.21395388]
 [1.04551592 1.64229256 6.56917025]
 [1.04551592 1.64229256 6.56917025]
 [1.04551592 1.64229256 6.56917025]
 [0.82982653 0.65174424 5.21395388]]
```
The reason I think this is wrong is because a rho surface integral uses the dtheta * dzeta portion of the node volume, ignoring drho. Similar for other surface labels. If we wanted the correct rho surface integral we should leave the drho unchanged and scale up /down dtheta * dzeta.

f0uriest: Ahhh I see. Can you make a new issue to keep track of that? Otherwise this PR looks good to go.

f0uriest (dismissed):

desc/grid.py
```diff
@@ -447,31 +450,62 @@ def _create_nodes(
             if self.sym:
                 t += t[1] / 2
             dt = 2 * np.pi / t.size * np.ones_like(t)
+            # if endpoint:
+            #     FIXME: dt already has the correct weights
+            #         _scale_weights() function is about to mess it up
+            #         can test with endpoint = True, ntheta = 3
+
```
ddudt, line 457: Can this be deleted?

desc/grid.py
```diff
@@ -447,31 +450,62 @@ def _create_nodes(
...
         else:
             t = np.atleast_1d(theta)
             dt = np.zeros_like(t)
             if t.size > 1:
-                dt[0] = (t[0] + t[1]) / 2
-                dt[1:-1] = (t[2:] - t[:-2]) / 2
-                dt[-1] = 2 * np.pi - (t[-2] + t[-1]) / 2
+                # choose dt to be half the cyclic distance of the surrounding two nodes
+                SUP = 2 * np.pi  # supremum
```
ddudt, line 463: Variable naming convention: is this all upper-case just because it's a hard-coded constant?

tests/test_configuration.py
```diff
@@ -410,19 +410,21 @@ class TestGetSurfaces:
...
-        assert surf.rho == 0.5
 
     @pytest.mark.unit
     def test_get_zeta_surface(self):
         """Test getting a constant zeta surface."""
         eq = Equilibrium()
         surf = eq.get_surface_at(zeta=np.pi)
-        np.testing.assert_allclose(surf.compute_surface_area(), np.pi * (1.0) ** 2)
         assert surf.zeta == np.pi
+        np.testing.assert_allclose(surf.compute_surface_area(), np.pi * 1 ** 2)
```
ddudt, line 427: Do we want to make a variable for the radius like the `R0` variable that was added in the previous test?

ddudt (approved):
See minor comments. 

It would be good to fix the duplicate node issue for Linear Grids with `endpoint=True` ASAP. That gets used a lot in plotting, so the plots could be showing inaccurate results for solutions with current constraints if the flux surface averages aren't being computed correctly.

desc/grid.py
unalmis: I'll remove it

desc/grid.py
unalmis: Yes; if you prefer something else just let me know.

ddudt (approved):

PR #280, opened 2022-09-29
Add retry steps to codecov upload

f0uriest:
- Should try uploading coverage up to 5 times for each action, and will fail the CI if all of the uploads fail.
- Renames artifacts to not overwrite them due to `pytest-split` groups-
- Renames workflows and job steps to be cleaner.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -46,16 +46,65 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v2
         with:
-          name: artifact-${{ matrix.python-version }}
+          name: artifact-${{ matrix.python-version }}-${{ matrix.group }}
           path: |
             ./cov.xml
             ./mpl_results.html
             ./prof.db
       - name: Upload coverage
+        continue-on-error: true
+        id : codecov1
         uses: codecov/codecov-action@v2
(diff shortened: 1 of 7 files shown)


f0uriest:
Good news is that retrying all the CI several times, the coverage uploaded correctly every time. 
Medium news is I'm not sure if this fix is actually doing anything, since it always works on the first try so the retries aren't necessary. But probably still a good safeguard.

.github/workflows/unittest.yml
```diff
@@ -41,21 +41,70 @@ jobs:
...
+        with:
+          name: codecov-umbrella # optional
+          files: ./cov.xml
+          fail_ci_if_error: true # optional (default = false)
+          verbose: true # optional (default = false)
+      - name: set the status                          # set the workflow status if command failed
+        if: always()
+        run: |
+          if ${{ steps.codecov1.outcome=='success' || steps.codecov2.outcome=='success' || steps.codecov3.outcome=='success' || steps.codecov4.outcome=='success' || steps.codecov5.outcome=='success' }}; then
+             echo fine
```
dpanici, line 107: is echo fine here necessary?

dpanici (approved):

PR #282, opened 2022-09-30
Fix profiles for vacuum solve

### LINKED ISSUE #275: Properly handle fixing unused degrees of freedom
For example, the vacuum solve by minimizing current doesn't use pressure at all. However, a use may think that they still need to include a `FixPressure` constraint, and this will throw an error with array sizes not matching in the `project`/`recover` functions. It should be simple to ignore constraints that aren't used (or to possibly throw a more informative error if the constraints make the problem infeasible or something)

### LINKED ISSUE #278: Solving vacuum equilibria from input file is somewhat broken
If an input file has `objective = vacuum` and no profiles given, the equilibrium is generated with a current profile. However, this is not constrained during optimization, and the resulting equilibrium can have a current profile with very large coefficients (~1e5). These seem to cancel to give a reasonable (if oscillatory) current profile, but it could cause issues with the optimizer since the step length its taking are O(1e4). 

Explicitly including `i = 0` in the input file seems to work, allowing iota to vary and giving reasonable step sizes. 

We should probably define what sort of behavior we want when `objective = vacuum`. Does this mean:
- Minimize J with I=0
- Minimize J with iota free
- something else?

f0uriest:
- Allows for constraining args that aren't used in the objective
- Fixes current profile for vacuum solves
- Also fixes a bug in `CurrentDensity` that prevented grids other than `jacobi` from being used.

In testing it seems like using fixed `I=0` is better than using unfixed iota when solving for vacuum cases. They both give similar results, but using fixed `I=0` gives slightly lower current density in the end (they both give the same iota profile)

Resolves #275
Resolves #278


dpanici (approved):

ddudt:
If constraining `I = 0` seems to always be better than minimizing `J = 0` with varying iota, do we still want to support the latter case? It was originally a hack to allow for vacuum solves before we finished implementing the current constraint. 

We could still keep `objective = vacuum` as an input option for minimizing current density instead of force balance, but could do it subject to `I = 0` instead of varying iota.

f0uriest:
Yeah with this PR, using `objective=vacuum` and giving no profiles will default to `I=0`, you can still force it to do the free iota version by including iota profile coefficients (which won't be fixed, just used as an initial guess). The differences are small, so its possible in some cases doing free iota might be better (or faster) so I don't think it hurts to leave it as an option.

ddudt (approved):
OK, looks good to me.

PR #283, opened 2022-10-03
plot_boundary function & test

ddudt:
Adds a new plot function to plot boundary surfaces and multiple toroidal angles together in a single plot. This is a popular plot format in stellarator optimization papers when comparing boundary shapes. It is similar to the existing `plot_surfaces` function, but focuses on the LCFS rather than internal flux surfaces.


desc/plotting.py
```diff
@@ -1282,6 +1282,132 @@ def plot_surfaces(eq, rho=8, theta=8, zeta=None, ax=None, **kwargs):
     return fig, ax
 
 
+def plot_boundary(eq, zeta=None, axis=True, ax=None, **kwargs):
+    """Plot stellarator boundary at multiple toroidal coordinates.
+
+    Parameters
+    ----------
+    eq : Equilibrium
```
f0uriest, line 1290: This might be a good place to allow passing in a `Surface` object directly. IE, make the plot routine expect a surface, or use `eq.surface`?

desc/plotting.py
```diff
@@ -1282,6 +1282,132 @@ def plot_surfaces(eq, rho=8, theta=8, zeta=None, ax=None, **kwargs):
...
+
+    assert (
+        len(kwargs) == 0
+    ), f"plot surfaces got unexpected keyword argument: {kwargs.keys()}"
+
+    if zeta is None:
+        if eq.N == 0:
+            zeta = 1
+        else:
+            zeta = 4
```
f0uriest, line 1353: another minor thing: for most of the other stellarator plots we default to 6 cross sections, should that be the same here, or do we want to change it elsewhere?

desc/plotting.py
```diff
@@ -1282,6 +1282,132 @@ def plot_surfaces(eq, rho=8, theta=8, zeta=None, ax=None, **kwargs):
...
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Object from which to plot.
+    zeta : int or array-like or None
+        Values of zeta to plot boundary surface at.
+        If an integer, plot that many contours linearly spaced in [0,2pi).
+        Default is 1 contour for axisymmetric equilibria or 4 for non-axisymmetry.
+    axis : bool
```
f0uriest, line 1296: might want to rename this to avoid confusion with `ax` argument.

f0uriest: (also not sure its totally necessary, I haven't seen the axis position included in these sorts of plots before)

desc/plotting.py
ddudt: We default to 6 cross-sections when they are in separate subplots. I think 4 is a good default for this function that plots them all in the same subplot, because otherwise it starts to get too crowded.

desc/plotting.py
ddudt: Renamed to `plot_axis`. 

I have seen it before and I like it, but I changed the default to not include the axis points.

tests/test_plotting.py
```diff
@@ -297,6 +299,15 @@ def test_plot_surfaces_no_theta(DSHAPE_current):
     return fig
 
 
+@pytest.mark.unit
+@pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_2d)
+def test_plot_boundary():
+    """Test plotting boundary."""
+    eq = get("W7-X")
```
ddudt, line 306: @f0uriest @dpanici It might be worth changing all the plotting tests to use the `desc/examples/` files like this function does, instead of the `DSHAPE_current` solution we are presently using for everything. That would allow us to remove the `@pytest.mark.slow` and `@pytest.mark.solve` markers, and allow us to test on stellarator solutions.

ddudt: OK yeah it probably wouldn't save much time. But I do think it would be good to have more plotting tests use stellarator examples, because there is often different functionality for stellarators vs tokamaks.

f0uriest (changes requested):
Should also add the new function to the API docs, otherwise fine.

f0uriest (approved):

PR #285, opened 2022-10-04
Fixes to wrapped objective and jit compiling

f0uriest:
- Previously, the optimizer would reach a point where no decrease was possible but due to noisy gradients from the wrapped objective it would keep trying until dx < xtol. However, the final step that it tried would be different from the final step accepted, and the wrong one would be returned as the "optimal" result. 
- We now compare the steps accepted by the optimizer to the internal steps in the wrapped objective to only return the accepted ones.
- Fixes a bug where the `use_jit` arg passed to `objective.build` would override any previously set value for `use_jit` (such as in the class constructor)
- Removes `use_jit` from the WrappedEquilibriumObjective` class, in favor of just specifying it separately for the sub-objectives, since the wrapped objective generally isn't exposed to the user.


dpanici (approved):

desc/optimize/utils.py
```diff
@@ -292,3 +292,15 @@ def evaluate_quadratic(J, g, s, diag=None):
...
+def find_matching_inds(arr1, arr2):
+    """Find indices into arr2 that match rows of arr1."""
+    arr1, arr2 = map(np.atleast_2d, (arr1, arr2))
+    inds = []
+    for i, a in enumerate(arr2):
+        x = np.isclose(arr1, a, rtol=1e-14, atol=1e-14)
+        j = np.where(x.all(axis=1))[0]
+        if len(j):
+            inds.append(i)
+    return np.asarray(inds)
```
ddudt, line 306: I get errors when I try to run the following:
```a = np.array([0, 2, 5, 9, 3, 4, 8])
b = np.array([-4, -5, 0, 1, 3])
inds = find_matching_inds(a, b)
```

And in the doc string, does it return the indices of arr2 or arr1?

f0uriest: updated the docstring to make it clearer. The two arrays should have the same number of columns, and the indices are into `arr2`

ddudt (approved):

PR #288, opened 2022-10-06
Optimizer

### LINKED ISSUE #253: Better guess for iniital trust radius
Right now the initial trust region radius defaults to 1, which is often too big allowing the optimizer to go off in weird directions. Usually something smaller like 1e-2 works well, but we should test more to see if there is a good universal default, or implement something based on the scale of the problem as detailed in https://epubs.siam.org/doi/epdf/10.1137/1.9780898719857.ch17

f0uriest:
- Cleans up and removes a lot of old/unused/unecessary code
- Gets rid of custom Hessian update stuff in favor of just using scipy's BFGS class
- Adds new options for setting the initial trust region in `lsqtr` and `fmintr` via the option `initial_trust_radius`:
  - A numeric value uses that value.
  - `"scipy"` uses the default from scipy, which is the norm of the scaled `x`. This was also the default value in prior versions of DESC and will remain so for the time being.
  - `"conngould"` uses a recommended value from Conn & Gould that scales the trust region so that the Cauchy step is on the boundary - generally more conservative than `"scipy"`
  - `"mix"` uses a geometric mean between `"scipy"` and `"conngould"`
- Adds a basic stochastic gradient descent optimizer `sgd` for noisy objectives. Currently only supports either a fixed step size or standard momentum. In the future may be updated to include other step size rules such as RMSProp or Adam.
- Adds utility function for performing modified Cholesky factorizations of indefinite matrices, needed to ensure that the `dogleg` and `subspace` methods work for general Hessian matrices.


desc/optimize/stochastic.py
```diff
@@ -0,0 +1,194 @@
+"""Function for minimizing a scalar function of multiple variables."""
```
ddudt, line 1: This looks copied from the `fmin_scalar.py` file and should probably be updated.

ddudt (approved):
It would be good to include unit tests for all of the `optimize/utils.py` functions. Presently they are only covered through regression tests. Otherwise these changes look good.

desc/optimize/stochastic.py
f0uriest: That description works for both so I think it's fine, the relevant detail is in the docstring.

f0uriest:
Yeah ideally more unit tests would be good but I think they can wait for a future PR. Most of the functions never get used on their own anyway.

dpanici (approved):
looks good, the default tols which are sensitive to if the objective is normalized or not might need to be changed in Normalization though?  gtol is reduced in this one

PR #291, opened 2022-10-09
Linting and general code cleanup

f0uriest:
- Lots of small formatting changes
- Adds a GH action to run ``flake8`` with a few plugins on each PR
- Adds a few things to ``dev-requirements`` for code style and linting
- **NOTE**: should now run ``flake8`` in the root of the repository before commits to flag any issues. Unlike ``black``, it will not fix them for you automagically, so you might have to make edits.
- Can override a particular error code on a per line basis by adding ``# noqa: CODE`` to that line.


desc/backend.py
```diff
@@ -58,15 +66,15 @@
     )
 )
 
-if use_jax:
+if use_jax:  # noqa: C901 - FIXME: simplify this, define globally and then assign?
```
dpanici, line 69: that error code is too much branching logic?

f0uriest: Basically, it refers to [cyclomatic complexity](https://en.m.wikipedia.org/wiki/Cyclomatic_complexity), or how many different paths there are through a function

dpanici (approved):
This must have been a bit tedious to do, thanks Rory

ddudt (approved):
I didn't review every single line, but I approve of the changes overall. Thanks.

PR #294, opened 2022-10-13
Fix plotting iota of a current-constrained with plot_1d function

dpanici:
Currently, plotting functions use the plot grid for both plotting and computing the requested quantity. However, `iota` for a current-profile-constrained equilibrium must be computed using flux surface averages, and these will not be accurate with the usual 1D grid used for plot_1d

this fix creates a temporary equilibrium with an iota profile calculated from the original, and uses that equilibrium to compute the data to plot. Fix applied to plot_1d and plot_2d and plot_section
TODO:
- [x] regen figs for plot_section if needed


desc/plotting.py
```diff
@@ -434,7 +434,15 @@ def plot_1d(eq, name, grid=None, log=False, ax=None, **kwargs):
     if len(plot_axes) != 1:
         return ValueError(colored("Grid must be 1D", "red"))
 
-    data, label = _compute(eq, name, grid, kwargs.pop("component", None))
+    if (
+        eq.iota is None
```
f0uriest, line 438: would it be cleaner to move this stuff to the `_compute` function?

dpanici: Ah, you are probably right

f0uriest (approved):

PR #295, opened 2022-10-14
Add normalizing factors to objectives

### LINKED ISSUE #325: Add normalizing factors to non-dimensionalize objective functions
Right now we compute everything in physical units (Amps, Newtons, Tesla etc). This is good for physics comparison, but makes weighting between different objectives in multi-objective optimization a bit trickier.

In the past we've tried non-dimensionalizing things dynamically by defining a dimensionless metric, but if the normalization is allowed to change during optimization it can cause issues (not to mention making the problem more nonlinear).

Instead we should pre-compute some static scaling factors based on "characteristic scales" in the problem (ie, toroidal flux + area gives a characteristic B field, pressure + radius gives a characteristic force, etc). The goal isn't to be exact, but just to get things all to roughly the same order of magnitude so they can be compared more meaningfully, similar to how we usually plot force error normalized by a pressure gradient.

f0uriest:
- Adds a new function `compute_scaling_factors` to compute characteristic dimensional quantities for an equilibrium
- These are then used to create normalizing scale factors for most objectives
- Should help a bit with poor conditioning in the optimizers, and give dimensionless metrics for comparison.

Resolves #325


f0uriest:
Would appreciate a sanity check on the units/scaling factors. They seem reasonable to me but there might be better options.

also @unalmis the stability objectives say they're dimensionless, but from the equations it looks like some of them have units? Can you confirm?

unalmis:
Must be a typo in the docstrings/print callback statements.

All the 
- All the Mercier stability critera (D_shear, D_well, D_merc...) are 1/ webers^2.
    - `D_shear` is just (diota/dpsi)^2  = 1 / webers^2.
    - `D_Mercier` is the sum of the other 4 stability quantities
- `magnetic well ` is fine (dimensionless)

dpanici:
What sorts of scale do the normalized quantities get? Are the normalized values relatively consistent across equilibria?

What you've chosen for each seems reasonable to me though

dpanici (dismissed):

desc/objectives/_qs.py
```diff
@@ -253,25 +280,37 @@ class QuasisymmetryTwoTerm(_Objective):
...
+            target=target,
+            weight=weight,
+            normalize=normalize,
+            normalize_target=normalize_target,
+            name=name,
+        )
+
         self._print_value_fmt = (
-            "Quasi-symmetry ({},{}) error: ".format(self.helicity[0], self.helicity[1])
+            "Quasi-symmetry two-term ({},{}) error: ".format(
```
ddudt, line 310: Should be `"Quasi-symmetry ({},{}) two-term error: "` if we want to follow the same formatting as the Boozer objective.

ddudt (changes requested):
1. We should also add the normalization options to the linear objectives, and make it part of the objectives API in the `_Objective` class. In typical usage the linear objectives are always satisfied exactly so the normalizations won't matter, but we should allow more general usage. 
2. Do we have any evidence that these normalizations actually improve the optimizations? The default is set to `normalize=True`, but none of the tests use it. **I will veto changing the default behavior until we are confident that it is usually an improvement.** And either way, we should include optimization tests that use the normalizations.

desc/objectives/normalization.py
```diff
@@ -0,0 +1,29 @@
...
+    """Compute dimensional quantities for normalizations."""
+    scales = {}
+    R10 = eq.Rb_lmn[eq.surface.R_basis.get_idx(M=1, N=0)]
+    Z10 = eq.Zb_lmn[eq.surface.Z_basis.get_idx(M=-1, N=0)]
+    R00 = eq.Rb_lmn[eq.surface.R_basis.get_idx(M=0, N=0)]
+
+    scales["R0"] = R00
+    scales["a"] = np.sqrt(np.abs(R10 * Z10))
+    scales["V"] = 2 * np.pi * scales["R0"] * scales["a"]
+    scales["A"] = np.pi * scales["a"] ** 2
```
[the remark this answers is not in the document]
f0uriest: Yeah I just wanted to keep it simple and not have to deal with transforming to real space. I checked several equilibria and these approximations agree within a few percent.

desc/objectives/normalization.py
```diff
@@ -0,0 +1,29 @@
...
+    scales["B_T"] = eq.Psi / scales["A"]
```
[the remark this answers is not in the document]
f0uriest: yeah ill fix that

desc/objectives/normalization.py
```diff
@@ -0,0 +1,29 @@
...
+    iota = eq.get_profile("iota")(np.linspace(0, 1, 20))
+    scales["B_P"] = scales["B_T"] * np.mean(np.abs(iota))
+    scales["B"] = np.sqrt(scales["B_T"] ** 2 + scales["B_P"] ** 2)
+    scales["I"] = scales["B_P"] / (2 * np.pi * mu_0)
```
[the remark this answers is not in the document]
f0uriest: yep

desc/objectives/_qs.py
f0uriest: fixed

f0uriest:
I think it's hard to say that they help the optimization, since effectively they're redundant with ``weight``, but the idea is that they make it easier to choose ``weight`` without having to think about the scales of different objectives. So it's entirely a UI improvement, it doesn't change any of the actual physics. But I see your point. I ran some examples locally that seem fine, I just didn't use normalization in any of the tests because then we'd have to mess with the weights to get the same result.

f0uriest:
Ultimately to get "good" results we'll still have to play around with the weights, the hope is just that this makes that a bit easier and more intuitive.

ddudt:
> I think it's hard to say that they help the optimization, since effectively they're redundant with `weight`, but the idea is that they make it easier to choose `weight` without having to think about the scales of different objectives. So it's entirely a UI improvement, it doesn't change any of the actual physics. But I see your point. I ran some examples locally that seem fine, I just didn't use normalization in any of the tests because then we'd have to mess with the weights to get the same result.

OK, I see your point too. I still think we should have a test somewhere that uses normalization so those new lines of code are hit. 

And what do you think about my first review comment of adding the normalization arguments to the `_Objectives` API?

f0uriest:
Yeah I think that makes sense, I just wasn't sure since some of the objectives are already dimensionless, but I guess we can just default to ``scale=1`` in that case. I'll try to add it to the rest of the objectives today.

desc/objectives/utils.py
```diff
@@ -153,14 +166,16 @@ def factorize_linear_constraints(constraints, objective_args):
         else:
             unfixed_args.append(arg)
             A_ = obj.derivatives["jac"][arg](jnp.zeros(obj.dimensions[arg]))
-            b_ = obj.target
+            # using obj.compute instead of obj.target to allow for correct scale/weight
+            b_ = -obj.compute(jnp.zeros(obj.dimensions[arg]))
```
dpanici, line 170: why is there a negative sign here?

desc/objectives/utils.py
```diff
@@ -153,14 +166,16 @@ def factorize_linear_constraints(constraints, objective_args):
...
             if A_.shape[0]:
                 Ainv_, Z_ = svd_inv_null(A_)
             else:
                 Ainv_ = A_.T
             A[arg] = A_
             b[arg] = b_
-            Ainv[arg] = Ainv_
+            # need to undo scaling here to work with perturbations
+            Ainv[arg] = Ainv_ * obj.weight / obj.normalization
```
dpanici, line 179: why is obj.weight multiplied here? this is related to the bug you had mentioned about it not working correctly when the objective is weighted?

desc/objectives/_wrappers.py
```diff
@@ -229,27 +230,19 @@ def jac(self, x):
...
-        }
-        Gx = {arg: Gx[:, self._objective.x_idx[arg]] for arg in self._objective.args}
-        for arg in self._eq_objective.args:
-            if arg not in Fx.keys():
-                Fx[arg] = jnp.zeros((self._eq_objective.dim_f, self.dimensions[arg]))
-            if arg not in Gx.keys():
-                Gx[arg] = jnp.zeros((self._objective.dim_f, self.dimensions[arg]))
-        Fx = jnp.hstack([Fx[arg] for arg in arg_order if arg in Fx])
-        Gx = jnp.hstack([Gx[arg] for arg in arg_order if arg in Gx])
+        Fx = align_jacobian(Fx, self._eq_objective, self._objective)
```
dpanici, line 233: this is just a refactor of the old code?

desc/objectives/utils.py
f0uriest: The residual is computed as f(x) = Ax-b, so f(0) = -b, so we need the negative sign to get the correct b

desc/objectives/utils.py
f0uriest: Yeah, `Ainv` here is only used as the relation between R and Rb, eg dR/dRb = Ainv, so we want that to be the unweighted version.

A R = Rb
(w A) R = w Rb
dR/dRb = inv(w A) * w

desc/objectives/_wrappers.py
f0uriest: yeah we were doing this in a few places and will probably need to also do it for more multiobjective/constrained stuff so made sense to just roll it into a utility function.

desc/objectives/_stability.py
```diff
@@ -184,7 +182,9 @@ def compute(self, R_lmn, Z_lmn, L_lmn, p_l, i_l, c_l, Psi, **kwargs):
             self._iota,
             self._current,
         )
-        return self._shift_scale(compress(self.grid, data["D_Mercier"]))
+        f = compress(self.grid, data["D_Mercier"], surface_label="rho")
+        w = compress(self.grid, self.grid.weights, surface_label="rho")
+        return self._shift_scale(f * w)
```
ddudt, line 187: In the `ToroidalCurrent` objective the radial weights are computed as follows:
```
I = compress(self.grid, data["current"], surface_label="rho")
w = compress(self.grid, self.grid.spacing[:, 0], surface_label="rho")
return self._shift_scale(I * w)
```
Why would it be any different for the Mercier objectives? I mean the difference between using `self.grid.weights` vs `self.grid.spacing[:, 0]`

desc/objectives/_stability.py
```diff
@@ -329,4 +353,6 @@ def compute(self, R_lmn, Z_lmn, L_lmn, p_l, i_l, c_l, Psi, **kwargs):
             self._iota,
             self._current,
         )
-        return self._shift_scale(compress(self.grid, data["magnetic well"]))
+        f = compress(self.grid, data["magnetic well"], surface_label="rho")
+        w = compress(self.grid, self.grid.weights, surface_label="rho")
```
ddudt, line 357: see previous comment

tests/test_objective_funs.py
```diff
@@ -93,6 +96,7 @@ def test(eq):
             W = obj.compute(
                 eq.R_lmn, eq.Z_lmn, eq.L_lmn, eq.p_l, eq.i_l, eq.c_l, eq.Psi
             )
```
ddudt, line 98: Looks like we can delete these lines since they are now redundant with line 99.

ddudt (changes requested):

tests/test_objective_funs.py
f0uriest: yep, done

dpanici (approved):

PR #297, opened 2022-10-25
Fix bug in ``ToroidalCurrent`` objective

f0uriest:
- Allows grid to have more than 1 rho surface to target a profile rather than a scalar value
- Uses different compute function to avoid computing extra unneeded qs stuff

### DIFF
--- a/desc/objectives/_generic.py
+++ b/desc/objectives/_generic.py
@@ -8,6 +8,7 @@ import desc.compute as compute_funs
 from desc.backend import jnp
 from desc.basis import DoubleFourierSeries
 from desc.compute import arg_order, compute_quasisymmetry_error, data_index
+from desc.compute.utils import compress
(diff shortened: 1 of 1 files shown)


unalmis (approved):

desc/objectives/_generic.py
```diff
@@ -272,5 +270,5 @@ def compute(self, R_lmn, Z_lmn, L_lmn, i_l, c_l, Psi, **kwargs):
             self._iota,
             self._current,
         )
-        I = 2 * jnp.pi / mu_0 * data["I"]
+        I = compress(self.grid, data["current"], surface_label="rho")
         return self._shift_scale(I)
```
unalmis, line 274: Don't need to wait for another CI for this, but I'd suggest eventually changing `I` to `current` for consistency.

PR #298, opened 2022-10-26
Fix bug when recovering full solution after ``optimal_perturb``

f0uriest:
as of v0.6.1, the recover operation from optimal_perturb would give R_lmn etc that are inconsistent with Rb_lmn due to recovering x with the wrong particular solution. This fixes the issue and adds a test to ensure the interior agrees with the boundary.

Note that this doesn't affect the optimize method based on ``optimal_perturb`` in a loop, since re-solving the equilibrium at each step effectively eliminates the issue.


desc/perturbations.py
```diff
@@ -468,12 +468,9 @@ def optimal_perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
         raise ValueError(
             "Invalid dimension: opt_subspace must have {} rows.".format(c.size)
         )
-    if objective_g.dim_f < dim_opt:
```
dpanici, line 471: what made this not needed anymore? can we now do this?

dpanici (approved):

ddudt (approved):

PR #299, opened 2022-10-26
Allow selecting profile when loading vmec equilibrium

f0uriest:
Previously we always grab ``ai`` from the VMEC wout file, even if the VMEC equilibrium was solved with fixed current. Unfortunately, VMEC doesn't say in the ouput file whether it was a fixed iota or fixed current equilibrium so we can't easily detect it. Instead ``VMECIO.load`` now has an argument ``profile`` which can be set to ``"iota"`` or ``"current"`` to select the appropriate profile.

### DIFF
--- a/desc/io/input_reader.py
+++ b/desc/io/input_reader.py
@@ -1318,7 +1318,7 @@ class InputReader:
         )
         # scale current profile
         if curr_tor is not None:
-            inputs["current"][:, 1] *= curr_tor / np.sum(inputs["current"][:, 1])
+            inputs["current"][:, 1] *= curr_tor / (np.sum(inputs["current"][:, 1]) or 1)
         # delete unused profile
         if iota_flag:
             del inputs["current"]
--- a/desc/vmec.py
+++ b/desc/vmec.py
@@ -30,7 +30,7 @@ class VMECIO:
     """Performs input from VMEC netCDF files to DESC Equilibrium and vice-versa."""
 
     @classmethod
-    def load(cls, path, L=-1, M=-1, N=-1, spectral_indexing="fringe"):
+    def load(cls, path, L=-1, M=-1, N=-1, spectral_indexing="fringe", profile="iota"):
         """Load a VMEC netCDF file as an Equilibrium.
 
         Parameters
@@ -45,6 +45,8 @@ class VMECIO:
             Toroidal resolution. Default = NTOR from VMEC solution.
         spectral_indexing : str, optional
             Type of Zernike indexing scheme to use. (Default = ``'fringe'``)
+        profile : {"iota", "current"}
+            Which profile to use.
 
         Returns
         -------
@@ -52,6 +54,7 @@ class VMECIO:
             Equilibrium that resembles the VMEC data.
 
         """
+        assert profile in ["iota", "current"]
         file = Dataset(path, mode="r")
         inputs = {}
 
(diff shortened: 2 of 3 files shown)


dpanici (dismissed):

desc/vmec.py
```diff
@@ -100,9 +103,19 @@ def load(cls, path, L=-1, M=-1, N=-1, spectral_indexing="fringe"):
...
+            inputs["iota"][:, 1] = file.variables["ai"][:]
+            inputs["current"] = None
+        if profile == "current":
+            inputs["current"] = np.zeros((preset, 2))
+            inputs["current"][:, 0] = np.arange(0, 2 * preset, 2)
+            inputs["current"][:, 1] = file.variables["ac"][:]
+            inputs["current"][:, 1] *= (
+                file.variables["ctor"][:] / (np.sum(inputs["current"][:, 1]) or 1),
+            )
+            inputs["iota"] = None
```
ddudt, line 118: I'm not sure this is correct. The VMEC `ac` output is whatever coefficients were used in the input file, not the actual coefficients of current profile. There are two concerns:
1. The default VMEC input is to give the derivative of the current profile wrt the VMEC radial coordinate s, but DESC expects the full profile as a function of rho. I think we are missing an integration factor here. 
2. The current profile also needs to be scaled to the total value of `ctor`, which you are already accounting for. 

See `VMECIO.parse_vmec_inputs` starting at line 1309 of `input_reader.py` for how both of these are handled already.

ddudt (changes requested):

desc/vmec.py
f0uriest: Think this is fixed now

desc/vmec.py
```diff
@@ -100,9 +103,29 @@ def load(cls, path, L=-1, M=-1, N=-1, spectral_indexing="fringe"):
...
         inputs["pressure"][:, 1] = file.variables["am"][:] * p0
-        inputs["iota"] = np.zeros((preset, 2))
-        inputs["iota"][:, 0] = np.arange(0, 2 * preset, 2)
-        inputs["iota"][:, 1] = file.variables["ai"][:]
+        if profile == "iota":
+            inputs["iota"] = np.zeros((preset, 2))
+            inputs["iota"][:, 0] = np.arange(0, 2 * preset, 2)
+            inputs["iota"][:, 1] = file.variables["ai"][:]
+            inputs["current"] = None
+        if profile == "current":
```
dpanici, line 111: Is it worth printing here that DESC is assuming the default current mode in VMEC (I'(s))? as technically in VMEC you can specify the I(s) profile if you pass a certain flag

desc/io/input_reader.py
```diff
@@ -1318,7 +1318,7 @@ def parse_vmec_inputs(vmec_fname, threshold=0):  # noqa: C901 - FIXME: simplify
         )
         # scale current profile
         if curr_tor is not None:
-            inputs["current"][:, 1] *= curr_tor / np.sum(inputs["current"][:, 1])
+            inputs["current"][:, 1] *= curr_tor / (np.sum(inputs["current"][:, 1]) or 1)
```
dpanici, line 1321: what does the " or one" here do?

dpanici (approved):
Looks fine, just left one question

desc/vmec.py
f0uriest: do we know what that flag is? If so we could detect it and throw an error or convert it?

desc/io/input_reader.py
f0uriest: just avoids dividing by zero. IE, if the current profile is all zeros it would end up getting set to NaN, this prevents that case.

desc/vmec.py
ddudt: I think it's `pcurr_type`. If `pcurr_type = power_series`, then it's expecting the derivative profile. If `pcurr_type = power_series_I`, then it's expecting the I(s) profile.

desc/vmec.py
```diff
@@ -100,9 +103,29 @@ def load(cls, path, L=-1, M=-1, N=-1, spectral_indexing="fringe"):
...
+                np.vstack(
+                    (
+                        inputs["current"][:, 0] + 2,
+                        inputs["current"][:, 1] * 2 / (inputs["current"][:, 0] + 2),
+                    )
+                ).T,
+                ((1, 0), (0, 0)),
+            )
+            inputs["current"][:, 1] *= (
+                file.variables["ctor"][:] / (np.sum(inputs["current"][:, 1]) or 1),
```
ddudt, line 126: We should also add the "or 1" to line 1321 of `desc/io/input_reader.py`.

ddudt (approved):
Approving, but see comments before your merge it in. 

The current integration looks correct now, but I would like to check it. Can we add a test for that?

desc/vmec.py
ddudt: It looks correct now, but have you tested it?

desc/vmec.py
f0uriest: I already did

desc/vmec.py
f0uriest: Yes I tested it locally, it's basically copy pasted from the input reader so I figured we don't need another test for it (plus that means another saved vmec thing to compare against etc)

desc/vmec.py
f0uriest: Actually in the input reader right now we just throw a warning, do we want to do the same, or just skip the integration if its ``pcurr_type = power_series_I``?

ddudt: Checking the input type would be "better", but I think just throwing a warning is fine for now.

PR #301, opened 2022-10-26
Ensure equilibrium and surface have same symmetry

f0uriest:
If the surface is symmetric but the underlying equilibrium isn't, then the non-symmetric modes aren't constrained properly.
If you create a surface it defaults to ``sym="auto"`` which in most cases is True, and then when creating an equilibrium from that, the equilibrium inherits the surfaces symmetry. However, when creating an equilibirum without a surface, the equilibrium first defaults to ``sym=False`` before the default surface is created with ``sym=True``.
The fix is to make sure the default surface has the same symmetry as the equilibrium, and also threw in some asserts to make sure it stays that way.


dpanici (approved):

ddudt (approved):
This works, but wouldn't it be better to set the default surface first, then set the default Equilibrium symmetry based on that surface? The way you have it now, `eq = Equilibrium()` will result in `eq.sym = False` even though the geometry of its surface should really be symmetric.

f0uriest:
Sort of. The default surface is symmetric, but if you're just creating an arbitrary equilibrium we're defaulting to non-symmetric because its assumed you might later do stuff like changing boundary shapes, and we don't want to limit that. I think it's consistent with what we assume elsewhere in the code.

ddudt:
OK. Sounds good to me.

PR #304, opened 2022-11-02
Try fixing codecov upload

f0uriest:


dpanici (approved):
I assume this needs to be merged in before we can see if it works, since actions run off of the master branch actions?

PR #305, opened 2022-11-03
Fix bug where if fixing a surface with only 1 mode in basis, modes_id…

dpanici:
…x gets reference before assignment in constraint.build

If trying to constraint an Equilibrium with a surface that has only 1 mode in one of its bases, then the `modes_idx` variable inside of the `FixBoundary` object corresponding to the single mode basis never gets set, but then gets referenced when trying to set the target and weights as the default values for `self.target` and `self.weight` match the logic criterion for a user-supplied `target` and `weight`

This fix adds logic to check that the defaults are not triggering the logic for user-supplied target/weight


dpanici:
I fixed this bug on the #306  PR and cherry-picked it to make this PR, but realize now that that is backwards (I should probably have made this PR, made the fix, then cherrypick this PR's commit to the other branch)
Hopefully this does not mess too much with the tree

f0uriest (approved):

ddudt (approved):

PR #306, opened 2022-11-03
Constrain axis and modes

### LINKED ISSUE #195: Having multiple constraints that target the same argument is not supported currently
On objectives branch, If you have more than one constraint that has the same argument (i.e. two constraints that both take in `R_lmn`), then when the constraint `Ainv` matrices are being built inside of `_build_linear_constraints`, whichever constraint comes later in the constraints list will overwrite the `Ainv` matrix of the other constraint (since both are being accessed by the same dict key)

This came up when I was running with two constraints that both targeted `L_lmn`, the lambda axis constraint and the Poincare constraint that constrained lambda at the zeta=0 plane. Can be seen in this code block:

```
from desc.objectives import ObjectiveFunction
from desc.objectives import (
    LCFSBoundaryR,
    LCFSBoundaryZ,
    LambdaGauge,
    PoincareLambda,
    FixedPressure,
    FixedIota,
    FixedPsi,
    PoincareBoundaryR,
    PoincareBoundaryZ,
)
from desc.objectives import ForceBalance
from desc.basis import ZernikePolynomial, FourierZernikeBasis,FourierZernike_to_PoincareZernikePolynomial
from desc.geometry import ZernikeRZToroidalSection
from desc.equilibrium import EquilibriaFamily, Equilibrium


eq = EquilibriaFamily.load(load_from='../examples/DESC/SOLOVEV_output.h5')[-1]

Rb_lmn, Rb_basis = FourierZernike_to_PoincareZernikePolynomial(eq.R_lmn, eq.R_basis)
Zb_lmn, Zb_basis = FourierZernike_to_PoincareZernikePolynomial(eq.Z_lmn, eq.Z_basis)

surf = ZernikeRZToroidalSection(
    R_lmn=Rb_lmn,
    modes_R=Rb_basis.modes[:, :2].astype(int),
    Z_lmn=Zb_lmn,
    modes_Z=Zb_basis.modes[:, :2].astype(int),
    spectral_indexing=eq._spectral_indexing,
)

eq_poin = Equilibrium(
    surface=surf,
    pressure=eq.pressure,
    iota=eq.iota,
    Psi=eq.Psi,  # flux (in Webers) within the last closed flux surface
    NFP=eq.NFP,  # number of field periods
    L=eq.L,  # radial spectral resolution
    M=eq.M,  # poloidal spectral resolution
    N=eq.N,  # toroidal spectral resolution
    L_grid=eq.L_grid,  # real space radial resolution, slightly oversampled
    M_grid=eq.M_grid,  # real space poloidal resolution, slightly oversampled
    N_grid=eq.N_grid,  # real space toroidal resolution
    sym=True,  # explicitly enforce stellarator symmetry
    bdry_mode="poincare",
)
eq_poin.L_lmn = (
    eq.L_lmn
)  # initialize the poincare eq with the lambda of the original eq
eq_poin.change_resolution(
    eq_poin.L, eq_poin.M, 1
)  # add toroidal modes to the equilibrium
eq_poin.N_grid = 2  # set resolution of toroidal grid

constraints = (
    PoincareBoundaryR(),
    PoincareBoundaryZ(),
    PoincareLambda(),  # this constrains lambda at the zeta=0 surface, using the eq's current value of lambda
    LambdaGauge(),
    FixedPressure(),
    FixedIota(),
    FixedPsi(),
)
objectives = ForceBalance()
obj = ObjectiveFunction(objectives, constraints)
obj.build(eq_poin)
obj.rebuild_constraints(eq_poin) # fails because the LambdaGauge Ainv is being used for the PoincareLambda, so the size is incorrect as they have different targets
```

### LINKED ISSUE #321: Add ability to constrain near axis behavior

dpanici:
- Add new constraints:
  - `FixAxisR` and `FixAxisZ`, which will fix the axis of an `Equilibrium` during a solve
  - `FixModeR` and `FixModeZ` which will fix specific modes of an `Equilibrium` during a solve
  - `FixSumModesR` and `FixSumModesZ` which fix sums of modes of an `Equilibrium` during a solve
  - `FixThetaSFL` to fix lambda=0 during an equilibrium solve 
- Utility functions to use these constraints to constrain a DESC equilibrium solve with NAE O(rho) behavior 
- Add logic inside of `factorize_linear_constraints` to allow multiple linear constraints targeting the same argument
  - Works by stacking the constraint matrices that target the same argument
  - checks that the constraints are compatible by making sure the two constraint matrices, when stacked, have a rank equal to the total number of rows  
- Changes `FixLambdaGauge` to now no longer constrain lambda to be 0 at the axis (which seems to result in better equiilbria) 

Resolves #195 
Resolves #321


f0uriest:
Overall looks good. Main thing I still want to go over is how we handle fixed vs non-fixed constraints. It looks like this will properly combine non-fixed things together, but not sure about how it combines fixed constraints targeting the same arg, but with possibly different modes? Or combining fixed and non-fixed args in a way that may not have a solution?

One possible check to add at the very end of ``factorize_linear_constraints`` would be to verify that ``xp`` actually satisfies all of the constraints and the dimension of ``Z`` is nonzero.

desc/objectives/utils.py
```diff
@@ -135,32 +137,61 @@ def factorize_linear_constraints(constraints, objective_args):
...
         if len(obj.args) > 1:
             raise ValueError("Linear constraints must have only 1 argument.")
         arg = obj.args[0]
         if arg not in objective_args:
             continue
-        constraint_args.append(arg)
+        if arg not in constraint_args:  # hope this does not mess
+            constraint_args.append(arg)
         if obj.fixed and obj.dim_f == obj.dimensions[obj.target_arg]:
             # if all coefficients are fixed the constraint matrices are not needed
```
f0uriest, line 155: What would happen if the user gives 2 constraints that fix the same modes to different values here?

dpanici: Good question, the way it is coded now I think the second would overwrite the first? I only was thinking about fixing not all the modes so this conditional was not being activated, there should be some error thrown I think saying that the constraints are incompatible like I have in the case where not all modes are fixed

desc/objectives/linear_objectives.py
```diff
@@ -179,7 +179,780 @@ class FixBoundaryZ(_Objective):
...
+        fixed_params = Z_lmn[self._idx]
+        return self._shift_scale(fixed_params)
+
+    @property
+    def target_arg(self):
+        """str: Name of argument corresponding to the target."""
+        return "Z_lmn"
+
+
+class FixSumModesR(_Objective):
```
f0uriest, line 937: Would it be possible to combine this with the FixModesR class? We could just detect based on whether ``modes`` is a list or modes or a list of lists of modes

dpanici: Yes, I just wanted to make a simple way to start getting results this week

dpanici:
Need to generalize `SumModes` to allow -1 in the A matrix

dpanici:
Make sure same issue as #380  does not happen using fixed axis and fixed mode objectives

dpanici:
TODO:
- [ ] make sure fixes from #380 work here for FixModeR and FixSumModesR
- [ ] clean up code for NAE constraints and add better documentaiton
- [ ] add example of how to use fixed axis, fixing NAE behavior (O(rho) at least)
- [ ] combine FixMode and FixSumModes?
- [ ] Increase test coverage

f0uriest:
If the O(r^2) stuff isn't quite ready/tested yet we could move that to another PR and just do the O(r) stuff and fixed axis for now, I think that's probably what will get used the most.

desc/objectives/linear_objectives.py
```diff
@@ -166,106 +166,1058 @@ def build(self, eq, use_jit=True, verbose=1):
...
+
+        super().__init__(
+            eq=eq,
+            target=target,
+            weight=weight,
+            name=name,
+            normalize=normalize,
+            normalize_target=normalize_target,
+        )
+        self._print_value_fmt = "Fixed-Z modes error: {:10.3e} (m)"
```
f0uriest, line 1071: move to class attributes

f0uriest (changes requested):
A few other general questions:

- QSC uses Boozer coordinates. When applying the NAE constraints, we convert the Boozer toroidal angle into our standard toroidal angle, but we still use Boozer theta. Is this necessary? Could we write the constraints to use a general theta?
- The API for getting NAE constraints is a bit different from the others, in that we need to pass in the QSC and eq objects to the utility function. Would it be possible to put the logic from `nae_utils` into the build method of a specialized class?

desc/objectives/linear_objectives.py
```diff
@@ -166,106 +166,1058 @@ def build(self, eq, use_jit=True, verbose=1):
...
-        self._args = ["Z_lmn"] if self._fixed_boundary else ["Zb_lmn"]
+        if modes is None or modes is False:
+            raise ValueError(
+                f"modes kwarg must be specified or True with FixSumModesR! got {modes}"
+            )
+        self._sum_weights = sum_weights
+        if target is not None:
+            if target.size > 1:
+                raise ValueError(
+                    "FixSumModesR only accepts 1 target value, please use multiple"
```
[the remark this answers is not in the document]
dpanici: Yes, I just was trying to get it cleaned up and this was the easiest way to address its current shortcomings

desc/objectives/linear_objectives.py
```diff
@@ -166,106 +166,1058 @@ def build(self, eq, use_jit=True, verbose=1):
...
+        Returns
+        -------
+        f : ndarray
+            Lambda gauge symmetry errors.
+
+        """
+        return jnp.dot(self._A, L_lmn)
+
+
+class FixLambdaZero(_Objective):
```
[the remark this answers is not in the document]
dpanici: True, I can change that

desc/objectives/linear_objectives.py
```diff
@@ -166,106 +166,1058 @@ def build(self, eq, use_jit=True, verbose=1):
...
+            Level of output.
+
+        """
+        idx = np.arange(eq.L_basis.num_modes)
+        modes_idx = idx
+        self._idx = idx
+
+        self._dim_f = modes_idx.size
+
+        # use axis parameters as target if needed
```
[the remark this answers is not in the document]
dpanici: this is I think an erroneous comment from an earlier copy/paste, I'll remove

desc/objectives/linear_objectives.py
```diff
@@ -166,106 +166,1058 @@ def build(self, eq, use_jit=True, verbose=1):
...
+            if idx.size < modes.shape[0]:
+                warnings.warn(
+                    colored(
+                        "Some of the given modes are not in the axis, "
+                        + "these modes will not be fixed.",
+                        "yellow",
+                    )
+                )
+        # hm, if we ...
+        # TODO: check for all bdryR things if work when False is passed in
```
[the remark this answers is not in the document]
dpanici: no, I added tests for it

desc/objectives/linear_objectives.py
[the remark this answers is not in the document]
f0uriest: OK. It's fine for now since in most cases the user won't be using these objectives directly. I could also see making `SumModes` into a base class for `FixBoundary` and `FixAxis` at some point.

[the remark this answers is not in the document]
dpanici: Yea that would be eventually useful, but for now since it is behind the NAE utils API I think it is ok, and the FixAxisR/Z are direct replacements for FixBoundaryR/Z so their use is obvious

dpanici:
> A few other general questions:
> 
> * QSC uses Boozer coordinates. When applying the NAE constraints, we convert the Boozer toroidal angle into our standard toroidal angle, but we still use Boozer theta. Is this necessary? Could we write the constraints to use a general theta?

I believe it is necessary for the O(rho) constraints. The way these constraints were derived was converting the X(rho,theta*,phi), Y(rho,theta*,phi) NAE flux surface positions into R(rho,theta*,phi), Z(rho,theta*,phi). These functions in the NAE formalism have the form (at constant phi) of
image

we then match that to the form of a function in the plane in Zernike:
image

getting the constraint eqns:
image

this is only possible since we assumed the poloidal angle is the same for each of these functional forms. If the Zernike one was generalized, we would have to replace $\theta$ everywhere by $\theta_{DESC} + \lambda(\rho,\theta_{DESC},\phi)$ and the simple relation between the NAE coefficients and the Zernike coefficients is lost (we can't simple just match terms on either side anymore, and I am not sure of how we would deal with all of the cos($\lambda$) terms which are each functions of $(\rho,\theta_{DESC}.\phi$) 

> * The API for getting NAE constraints is a bit different from the others, in that we need to pass in the QSC and eq objects to the utility function. Would it be possible to put the logic from `nae_utils` into the build method of a specialized class?

Do you mean like an objective class `NearAxisConstraint` that just internally creates the necessary constraints and takes as argument a qsc, but not necessarily an equilibrum, and the NAE constraints that are in the nae_utils.py would be created when this overarching constraint is built? I suppose I see how that could be useful/more intuitive than having this helper function

> * Also, I assume all this stuff also works for QIC? can we add a test to make sure?

I have not tested but can add a test

> * In the examples/notebook, the NAE constraints are always applied to an equilibrium that was first created by doing `Equilibrium.from_near_axis`. Is this necessary? Or could you just create some arbitrary equilibrium and then apply the NAE behavior from a QSC equilibrium?

This is not actually something I've tried, I can test it. But, currently the util function for NAE constraints assumes that the axis you want fixed is the equilibrium's current axis, which I assume is the qsc axis. It does not make much sense to fix O(r) behavior with a different axis. I should change the function so it always tries to fix the QSC axis, that is an oversight on my part.

- [ ] Add pyQsc test
- [ ] Add pyQic test (for both of these, upload a Qsc equilibrium object pickle so that the CI does not need us to pip install qsc and qicna
- [ ] Change NAE util function to get desired axis from the qsc equilibrium passed in, instead of assuming passed-in DESC equilibrium has the correct axis
- [ ] test on solving with an equilibrium that is not just a fit from the qsc equilibrium (but has same axis shape)

desc/objectives/utils.py
```diff
@@ -83,12 +87,89 @@ def get_fixed_boundary_constraints(
...
 
+def get_fixed_axis_constraints(profiles=True, iota=True):
+    """Get the constraints necessary for a fixed-axis equilibrium problem.
+
+    Parameters
+    ----------
+    profiles : bool
+        Whether to also return constraints to fix input profiles.
+    iota : bool
+        Whether to add FixIota or FixCurrent as a constraint.
```
ddudt, line 98: Would it be cleaner to have the inputs be `iota=False, current=False` instead of `profiles` and `iota`? That feels more intuitive to me, but the advantage of the existing code is that it prevents having both profiles constrained.

desc/objectives/nae_utils.py
```diff
@@ -0,0 +1,239 @@
...
+    bases["Rbasis_cos"] = Rbasis
+    bases["Rbasis_sin"] = Rbasis_sin
+    bases["Zbasis_cos"] = Zbasis
+    bases["Zbasis_sin"] = Zbasis_sin
+
+    coeffs = {}
+    coeffs["R_1_1_n"] = R_1_1_n
+    coeffs["R_1_neg1_n"] = R_1_neg1_n
+    coeffs["Z_1_1_n"] = Z_1_1_n
+    coeffs["Z_1_neg1_n"] = Z_1_neg1_n
```
ddudt, line 112: maybe we could rename these `"R_1_+1_n"` and `"R_1_-1_n"`?

desc/objectives/nae_utils.py
```diff
@@ -0,0 +1,239 @@
...
+    Zbasis_cos = bases["Zbasis_cos"]
+    Rbasis_sin = bases["Rbasis_sin"]
+    Zbasis_sin = bases["Zbasis_sin"]
+
+    # R_11n
+    for n, NAEcoeff in zip(Rbasis_cos.modes[:, 2], coeffs["R_1_1_n"]):
+        sum_weights = []
+        modes = []
+        target = NAEcoeff * r
+        for k in range(1, int((desc_eq.L + 1) / 2) + 1):  # FIXME: add by 1 desc_eq.L+1
```
ddudt, line 165: Does this still need to be fixed?

desc/objectives/nae_utils.py
```diff
@@ -0,0 +1,239 @@
...
+        modes = []
+        target = NAEcoeff * r
+        for k in range(1, int((desc_eq.L + 1) / 2) + 1):
+            modes.append([2 * k - 1, -1, n])
+            sum_weights.append([(-1) ** k * k])
+        modes = np.atleast_2d(modes)
+        sum_weights = -np.atleast_1d(sum_weights)
+        Rcon = FixSumModesR(target=target, sum_weights=sum_weights, modes=modes)
+        Rconstraints += (Rcon,)
+    # Z_11n
```
ddudt, line 198: I think these comments and variable names will be clearer if you use the consistent notation `Z_1_+1_n` etc.

desc/objectives/utils.py
dpanici: I was just sticking in with the convention we already have, as Rory said

desc/objectives/nae_utils.py
dpanici: With `neg1` I can make the dictionary keys line up better with the variable names I have used, is my defense for keeping it

desc/objectives/nae_utils.py
dpanici: I made the comments consistent with the variable names

f0uriest:
Looks mostly good to go, only question left is if we want the NAE example notebook to replace the QSC one? or combine the two?

dpanici:
I'd think combine the two, since the NAE notebook is a literal copy of the Qsc one. I can do that now

f0uriest (approved):

PR #309, opened 2022-11-07
Add pre-commit hooks

f0uriest:
Adds a config file for ``pre-commit`` that runs black, flake8, and other tools on each commit. To install, run ``pip install pre-commit`` followed by ``pre-commit install`` in the root of the repo.

Also upgrades to a newer more stable version of ``black`` and ``flake8`` along with resulting formatting fixes


dpanici (approved):
looks good, I like pre-commit.

PR #310, opened 2022-11-09
Add jit method to objectives

f0uriest:
- Because JAX compiles `self` as a static argument for jitted methods, updating the attributes of the objective (target, weight etc) are not correctly updated in the jitted code.
- This adds a `jit` method that will either jit the compute method, or delete and re-jit it, to ensure that any updates to `self` are correctly handled.
- Also moves some boilerplate code to the base class build method.
- `_set_derivatives` now no longer takes `use_jit` argument, as this is now part of a separate method.

### DIFF
--- a/desc/objectives/_equilibrium.py
+++ b/desc/objectives/_equilibrium.py
@@ -125,10 +125,7 @@ class ForceBalance(_Objective):
         if verbose > 1:
             timer.disp("Precomputing transforms")
 
-        self._check_dimensions()
-        self._set_dimensions(eq)
-        self._set_derivatives(use_jit=use_jit)
-        self._built = True
+        super().build(eq=eq, use_jit=use_jit, verbose=verbose)
 
     def compute(self, R_lmn, Z_lmn, L_lmn, p_l, i_l, c_l, Psi, **kwargs):
         """Compute MHD force balance errors.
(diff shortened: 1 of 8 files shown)


ddudt (approved):

dpanici (dismissed):
looks good, also tested on Della ( jax 0.3.23, jaxlib 0.3.22) and locally (jax version=0.2.25, jaxlib version=0.1.76) and both works fine

edit: only tested rejit, I see other tests don't pass now on CI

dpanici (commented):
Need to fix tests still

dpanici:
Seems like if the logic inside `self.jit()` is passed so that `self.compute` is deleted, it then makes the argspec for the `self.compute` return `[]` as the argument, and so then `self.args` once it gets set again in `_set_derivatives()` inside of `self.jit()`, is equal to `[]`, instead of the correct `R_lmn` etc

this happens only during `perturb` as in compute, the objective's compute is not already jitted.

image

the print statement above for `after del` is after the new self.compute is set by the line `self.compute = jit(self.compute). I think this line is jitting `None`, as the `self.compute` function also does nothing when called after running `self.jit()` with an already built and jitted objective

dpanici (approved):

ddudt (approved):

Issue #230, opened 2022-07-01
nfev = 1


tmqian:
When `nfev = 1`, it doesn't actually set it to 1 iteration, instead it runs to more than 500 steps. Found by Dario Daniel and Tony at Greifswald.

PR #311, opened 2022-11-11
fix bug where if maxiter=0 with lsq-exact optimizer, it would never t…

dpanici:
- fix bug where if maxiter=0 with lsq-exact optimizer, it would never terminate since check_termination would never get called as nfev>max_nfev from the start
- fix bug where if a desc bfgs method is used, the xscale would still default to `hess` and then the optimizer throws an error
- passing a string of the name of an optimizer to `eq.solve` now will actually use the requested optimizer
- fix bug where when using `sgd`, `method` is passed instead of `self.method`
- add tests for maxiter=0 and maxiter=1 for desc least squares and scalar


f0uriest:
I thought we fixed this a while ago? Both ``success`` and ``message`` get initialized, so we don't have to call ``check_termination``

dpanici:
```
import desc.examples
eq = desc.examples.get('SOLOVEV')
eq.solve(maxiter=0,verbose=3)
```
yields this on master

image

dpanici:
it was fixed for #230 nfev=1 but not maxiter=0

f0uriest:
ah ok we should check if `fmin_scalar` is also affected, we might need to fix that as well

ddudt (changes requested):
Looks good, but we should include small tests for these edge cases.

desc/optimize/least_squares.py
```diff
@@ -221,6 +221,27 @@ def lsqtr(  # noqa: C901 - FIXME: simplify this
 
         actual_reduction = -1
 
+        if max_nfev or maxiter == 0:
```
f0uriest, line 224: do we need this if? or can we always just check termination before starting the loop to see if we're already at a local min?

desc/optimize/optimizer.py
```diff
@@ -494,7 +498,7 @@ def jac(x_reduced):
                 x0=x0_reduced,
                 grad=grad_wrapped,
                 args=(),
-                method=method,
+                method=self.method,
```
f0uriest, line 501: i think this needs to be method since we parse the string above to detect some options

f0uriest (changes requested):
Did we also check that `fmintr` is not affected by this same bug?

desc/optimize/optimizer.py
dpanici: the parse only happens for the elif statement above, not inside this elif statement, so method is undefined here. also currently no options affect the sgd method

desc/optimize/least_squares.py
dpanici: a few of the values needed for check_termination are not initially defined until the while loop, and the way the code is structured, it always check for termination since it will always enter the loop once. this just catches the case where with 0 maxiter, it never enters the loop and so never checks for termination

dpanici:
> Did we also check that `fmintr` is not affected by this same bug?

yep, the test with `dogleg-bfgs`  method tests fmintr

ddudt (approved):

PR #314, opened 2022-11-15
Allow Optimizer instance or string in equilibrium methods

### LINKED ISSUE #313: Inconsistent argument type for `optimizer` in `eq.solve` and `eq.optimize`
in `eq.solve`, it can be a string or an optimizer, but in `eq.optimize` it must be an optimizer.

I would like both to be strings or optimizers so that you can pass in "dogleg" or whatever and have it work without an extra line making the optimizer. But I know passing in Optimizer objects lets you potentailly reuse it. 

How advantageous is reusing the Optimizer object? Is it so useful that we must require Optimizers to be passed? or can we allow string names as well to be passed?

f0uriest:
Resolves #313

### DIFF
--- a/desc/equilibrium.py
+++ b/desc/equilibrium.py
@@ -402,7 +402,7 @@ class Equilibrium(_Configuration, IOAble):
             Objective function to solve. Default = force balance on unified grid.
         constraints : Tuple
             set of constraints to enforce. Default = fixed boundary/profiles
-        optimizer : str or Optimzer (optional)
+        optimizer : str or Optimizer (optional)
             optimizer to use
         ftol : float
             Relative stopping tolerance on objective function value.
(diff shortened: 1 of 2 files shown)


dpanici (approved):

desc/equilibrium.py
```diff
@@ -402,8 +402,8 @@ def solve(
             Objective function to solve. Default = force balance on unified grid.
         constraints : Tuple
             set of constraints to enforce. Default = fixed boundary/profiles
-        optimizer : string
-            Optimization algorithm. Default = "lsq-exact".
+        optimizer : str or Optimizer (optional)
+            optimizer to use
```
ddudt, line 406: Can we change this (and the other description on line 527) to:
```Optimizer to use. Default algorithm is "lsq-exact".```

ddudt (approved):
Approved, but see single comment about clarifying the documentation.

PR #316, opened 2022-11-17
Logging overhaul2

MothVine:
Primary addition is in __init__.py; adds functions set_console_logging and set_logfile_logging.  All print statements have been routed to python's logging module, which requires libraries to not log anything natively- will need to use these two functions to view logging information in console or written to file.

To write to logs, instead of print statments, one uses "logging.warning("string"), where 'warning can be replaced by the log level of your choice.  Activating a lower log level will display all statments at that level or higher.  A quick call to one of the "set logging" functions in a fresh new environment will need to be gotten used to.  The lowercase is important!  Do not use "logging.WARNING("string").  This just returns the numeric value of the log level.

The native log levels - listed as follows- allow numeric inputs as well- levels "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL" correspond to 10, 20, 30, 40, and 50 respectively.

Currently, I have written most standard log messages as "WARNING", termination and error messages as "ERROR", with "CRITICAL" being unused, and synonymous with silent running.  "INFO" displays info on individual iterations, and "DEBUG" shows timing info along with all inner and outer loop iteration info.

Added a flag for logging levels, and made the --verbose and --quiet flags refer to set_console_logging() levels.  Also added --logging, which can be used as -l, -ll, and -lll as well to access WARNING, INFO, and DEBUG respectively.

Other changes- currently some of scipy's functions we use automatically write to stdout through print statements- which is annoying, and required helper functions to capture their writing to stdout and redirect it to the base logging system.  Which is exactly the reason libraries are not supposed to have authoritative printing and logging behaviours; our compliance with the logging library is agnostic to user intent.

Old tests for the functionality of our i/o have been rewritten, but might probably require more tests now to make sure these function as intended- especially as the way I've written the scipy stdout capturing function requires python3.

Questions: 
1.  Currently set_console_logging admits a parameter to print to stderr, or stdout- if we don't plan to use this, should remove it.  Same with set_logfile_logging, admiting file names other than desc.log.
2. Logfile default is set to DEBUG to show all possible info, console logging is set to WARNING to display less info.  Should this change?
3.  We have some error and warning messages using print statements, and not the standard library's warning and error modules we sometimes import.  I have converted these to logging.error() statements- should I reroute those to the standard error modules?  Or should I rewrite all error and warning info to be handled by the logger?
4.  Currently logging mostly matches our prior 0, 1, 2 levels of verbosity.  The higher level ERROR could be made more use of- and CRITICAL is currently not used at all.  Do we want to change what the different log levels mean currently? 
5.  Currently I have the logfile desc.log being overwritten any time the log level is set to a different value, with a flag to make this optional.  Do we want overwriting old log files to be the default, or not-overwriting old desc.log log files to be the default?
6.  If we're swapping from print statements everywhere to python's core logging, what's the best way to alert users to new functionality, given DESC no longer emits messages aside from initialization messages by default?


desc/optimize/optimizer.py
```diff
@@ -405,24 +395,27 @@ def callback(x_reduced):
...
-                    method=self.method[len("scipy-") :],
-                    jac=grad_wrapped,
-                    hess=hess_wrapped,
-                    tol=gtol,
-                    callback=callback,
-                    options={"maxiter": maxiter, "disp": disp, **options},
-                )
+                #Sends stdout output of scipy method to logger
+                scipy_stringIO = StringIO()
+                with redirect_stdout(scipy_stringIO):
```
f0uriest, line 406: will this print iterations as it goes, or just capture all of them and dump to log when it finishes?

desc/verbose.txt
```diff
@@ -0,0 +1,41 @@
+__main__.py:    if ir.args.verbose:
```
f0uriest, line 1: what is this file for?

f0uriest (commented):
A few notes below. As a general note I think it would be good to consolidate the logging levels down to just info and debug, we could basically just shift everything down 1 level so warning -> info, info -> debug, debug -> debug
I was also wondering if it's possible to locally override the logging behavior. IE, default logging is info or warning or whatever, but when I call a particular function I want more info?

desc/equilibrium.py
```diff
@@ -223,10 +224,10 @@ def resolution(self):
 
     def resolution_summary(self):
         """Print a summary of the spectral and real space resolution."""
-        print("Spectral indexing: {}".format(self.spectral_indexing))
-        print("Spectral resolution (L,M,N)=({},{},{})".format(self.L, self.M, self.N))
-        print("Node pattern: {}".format(self.node_pattern))
-        print(
+        logging.debug("Spectral indexing: {}".format(self.spectral_indexing))
```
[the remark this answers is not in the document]
MothVine: Can definitely shift the logging levels down a step.

The set_console_logging and set_logfile_logging functions will override their handlers if called again with new settings.  To start with, you might use the defaults with a single "desc.set_console_logging()" call with no args.  But, if you want more info, you can call "desc.set_console_logging(loglevel="DEBUG")" and run your prior command again.  Or, before running anything, you can call "desc.set_logifile_logging()" whose defaults currently capture all logs- so you know you've got a record of all possible logs somewhere, without clogging the console screen.

Currently I don't have a function to turn logging off entirely again, without setting the loggers manually to "CRITICAL", or using a -q flag, which I probably should do for consistency.

desc/objectives/objective_funs.py
```diff
@@ -558,7 +544,7 @@ def compute_scalar(self, *args, **kwargs):
     def print_value(self, *args, **kwargs):
         """Print the value of the objective."""
         x = self._unshift_unscale(self.compute(*args, **kwargs))
-        print(self._print_value_fmt.format(jnp.linalg.norm(x)))
+        logging.warning(self._print_value_fmt.format(jnp.linalg.norm(x)))
```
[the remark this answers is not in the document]
MothVine: Gotcha.  I've altered several of the print-formating function in the utils modules like pretty_print and the timer printing functions, because they seem like things desc primarily calls, rather than the user.  Are there other functions like this meant specifically for the user to output to stdout as well that I should revert?

desc/io/input_reader.py
```diff
@@ -88,7 +91,7 @@ def parse_args(self, cl_args=None):
         if args.version:
             import desc
 
-            print(desc.__version__)
+            logging.warning(desc.__version__)
```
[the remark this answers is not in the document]
MothVine: Yep, good catch

desc/verbose.txt
MothVine: That's an intermediate output file I git grep'ed into; removing.  Wasn't quite finished last night waiting for unittests to come back; had to fix two other bugs like this I omitted as well.

desc/optimize/optimizer.py
MothVine: This is a good question; I'll mock up a test version and find out.  My assumption was that defining redirect_stdout with the try/yeild/finally syntax would handle the output on a line-by-line basis but I've not worked with StringIO type things in python before.

MothVine: Hmm, it does capture them all and dumps them to log when finished- the formatting output is more sensible than other methods I looked at, but it still counts it as one large log.  Unsure how to handle this then.

MothVine: I think I can emit a DEBUG-level log statement that labels the chunk of log scipy will generate in a way that keeps the text's intended columns, and add INFO level logs before and after labelling the LSQ message block's beginning and termination message clearly.

desc/objectives/_geometry.py
```diff
@@ -96,16 +100,18 @@ def build(self, eq, use_jit=True, verbose=1):
...
 
         self._profiles = get_profiles(self._data_keys, eq=eq, grid=self.grid)
         self._transforms = get_transforms(self._data_keys, eq=eq, grid=self.grid)
 
         timer.stop("Precomputing transforms")
-        if verbose > 1:
-            timer.disp("Precomputing transforms")
+        timer.disp("Precomputing transforms")
+
+        if self._normalize:
```
dpanici, line 112: this if block does not belong here, this is the reason why the tests are failing. Aspect ratio is a unitless quantity, so it does not need normalization.

dpanici:
Also here are the instructions for pre-commit, after you do these you can run `pre-commit run --all-files` locally and fix all the flags that it gives you, re-running that command to check until each one passes, then commit, that'll be the easiest way to deal with the flake8 stuff I think

> We recommend installing pre-commit with `pip install pre-commit` and then running `pre-commit install` from the root of the repository. This will automatically run a number of checks every time you commit new code, reducing the likelihood of committing bad code.

MothVine:
Closing PR due to significant changes addressed in the logging_overhaul3 branch.

PR #318, opened 2022-11-21
Improve numerical stability of perturbations and wrapped objective

### LINKED ISSUE #319: Figure out why optimization sometimes stalls
In certain optimization scenarios we find that the optimizer will stall at a certain point, trying smaller and smaller steps but not finding a descent direction, despite the gradient being nonzero. Generally this happens early in the optimization, before much progress has been made, and we know that a better solution exists.
This would seem to imply that the jacobian/gradient is wrong such that the gradient and newton directions are not descent directions, but the cause is unclear.

f0uriest:
- Adds a small amount of regularization to optimal perturbations to reduce bad conditioning
- Cleans up how the wrapped objective stores data, and prevents re-solving the same equilibrium twice on an accepted step.
- Add option to include constant term in perturbations. Defaults to True for regular perturbations, since it tends to help force balance when perturbations are large, but is set to False when doing optimization when perturbations are smaller

Resolves #319


ddudt (dismissed):

desc/perturbations.py
```diff
@@ -251,15 +251,13 @@ def perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
...
-        f = objective.compute(x)
-
         # 1st partial derivatives wrt both state vector (x) and input parameters (c)
         if verbose > 0:
             print("Computing df")
         timer.start("df computation")
         Jx = objective.jac(x)
         Jx_reduced = Jx[:, unfixed_idx] @ Z @ scale
-        RHS1 = f + objective.jvp(tangents, x)
+        RHS1 = objective.jvp(tangents, x)
```
dpanici, line 260: what was the reason for this change again?

f0uriest: Formally the `f` term shouldn't be there (its order epsilon^0 rather than epsilon^1). And also in theory, it may lead to the perturbations trying too hard to decrease force error rather than keeping force the same but moving to a new region of parameter space, leaving the solution stuck in the wrong local minima.

I re-ran some of the examples locally and it seems like removing `f` improves some cases and makes others worse. I'll revert it for now until we can be sure.

tests/test_examples.py
```diff
@@ -262,48 +261,60 @@ def run_qh_step(n, eq):
...
-    np.testing.assert_allclose(rho_err, 0, atol=1e-4)
-    np.testing.assert_allclose(theta_err, 0, atol=1e-4)
+    rho_err, theta_err = area_difference_desc(eq1, eq1a, Nr=1, Nt=1)
+    # only need crude tolerances here to make sure the boundaries are
+    # similar, the main test is ensuring its not pathological and has good qs
+    assert rho_err.mean() < 1
+
+    grid = LinearGrid(M=eq1a.M_grid, N=eq1a.N_grid, NFP=eq1a.NFP, sym=False, rho=1.0)
+    data = eq1a.compute("|B|_mn", grid, M_booz=eq1a.M, N_booz=eq1a.N)
+    idx = np.where(np.abs(data["B modes"][:, 1] / data["B modes"][:, 2]) != 1)[0]
```
dpanici, line 276: should be B_modes here not B modes

tests/test_examples.py
```diff
@@ -262,48 +261,60 @@ def run_qh_step(n, eq):
...
-    np.testing.assert_allclose(rho_err, 0, atol=1e-4)
-    np.testing.assert_allclose(theta_err, 0, atol=1e-4)
+    rho_err, theta_err = area_difference_desc(eq2, eq2a, Nr=1, Nt=1)
+    # only need crude tolerances here to make sure the boundaries are
+    # similar, the main test is ensuring its not pathological and has good qs
+    assert rho_err.mean() < 1
+
+    grid = LinearGrid(M=eq2a.M_grid, N=eq2a.N_grid, NFP=eq2a.NFP, sym=False, rho=1.0)
+    data = eq2a.compute("|B|_mn", grid, M_booz=eq2a.M, N_booz=eq2a.N)
+    idx = np.where(np.abs(data["B modes"][:, 1] / data["B modes"][:, 2]) != 1)[0]
```
dpanici, line 295: same as above

tests/test_examples.py
```diff
@@ -262,48 +261,60 @@ def run_qh_step(n, eq):
...
-    np.testing.assert_allclose(rho_err, 0, atol=1e-4)
-    np.testing.assert_allclose(theta_err, 0, atol=1e-4)
+    rho_err, theta_err = area_difference_desc(eq3, eq3a, Nr=1, Nt=1)
+    # only need crude tolerances here to make sure the boundaries are
+    # similar, the main test is ensuring its not pathological and has good qs
+    assert rho_err.mean() < 1
 
     grid = LinearGrid(M=eq3a.M_grid, N=eq3a.N_grid, NFP=eq3a.NFP, sym=False, rho=1.0)
     data = eq3a.compute(["|B|_mn", "B modes"], grid=grid, M_booz=eq3a.M, N_booz=eq3a.N)
     idx = np.where(np.abs(data["B modes"][:, 1] / data["B modes"][:, 2]) != 1)[0]
```
dpanici, line 315: same as above

tests/test_examples.py
f0uriest: `B modes` is the correct key, I just forgot that we need to explicitly ask for that from `eq.compute` after the recent changes.

desc/objectives/_wrappers.py
```diff
@@ -139,7 +141,13 @@ def _update_equilibrium(self, x, store=False):
         After updating, if store=False, self._eq will revert back to the previous
         solution when store was True
         """
-        if jnp.allclose(x, self._x_old, rtol=1e-14, atol=1e-14):
+        from desc.optimize.utils import f_where_x
+
+        # first check if its something we've seen before, if it is just return
```
dpanici, line 146: is this is where we basically can see if we've already seen the same x, then the optimizer did not accept the step and we dont need to do anything?

desc/objectives/_wrappers.py
```diff
@@ -249,17 +268,33 @@ def jac(self, x):
...
         # possibly better way: Gx @ np.eye(Gx.shape[1])[:,self._unfixed_idx] @ self._Z
         Fx_reduced = Fx[:, self._unfixed_idx] @ self._Z
         Gx_reduced = Gx[:, self._unfixed_idx] @ self._Z
-
         Fc = Fx @ dxdc
-        Fx_reduced_inv = jnp.linalg.pinv(Fx_reduced, rcond=1e-6)
-
         Gc = Gx @ dxdc
+
+        # FIXME (@f0uriest): need to import here to avoid circular dependencies
```
dpanici, line 279: is this easily fixable?

desc/optimize/_scipy_wrappers.py
```diff
@@ -100,8 +100,7 @@ def hess_wrapped(xs):
         hess_allf.append(H)
         return H / (np.atleast_2d(x_scale).T * np.atleast_2d(x_scale))
 
-    def callback(xs):
-        x1 = xs / x_scale
```
dpanici, line 104: why was x_scale gotten rid of here?

dpanici (dismissed):
Looks good beyond my asking for clarification at a couple spots, and we shuold check that failing qh opt test

desc/optimize/_scipy_wrappers.py
f0uriest: The callback is called from `grad_wrapped` after rescaling, so we don't need to rescale again

desc/objectives/_wrappers.py
f0uriest: not really, we'd need to move a bunch of stuff around. I started to do it on the `rc/lagrangian` branch but there's no rush.

desc/objectives/_wrappers.py
f0uriest: the more common case is that we saw this same `x` when we tried this step and evaluated `f` but haven't actually moved to the new x yet. After the step is accepted, we want to compute the jacobian, but we don't want to re-solve the equilibrium because we already solved it when trying that step the first time

dpanici:
sorry for another comment but the ESTELL test failed, seems if the tolerance was raised slightly it would pass tho (to like 4.6e-2 )

tests/test_examples.py
```diff
@@ -218,8 +218,8 @@ def run_qh_step(n, eq):
 
     objective = ObjectiveFunction(
         (
-            QuasisymmetryTwoTerm(helicity=(1, -eq.NFP), grid=grid, normalize=False),
-            AspectRatio(target=8, weight=1e1, normalize=False),
+            QuasisymmetryTwoTerm(helicity=(1, eq.NFP), grid=grid),
```
ddudt, line 221: I think the helicity should be negative for this example: `helicity=(1, -eq.NFP)`.

f0uriest:
Yeah I'm running it locally to make sure it still looks correct. Also regenerating the QH test plot

dpanici:
> Yeah I'm running it locally to make sure it still looks correct. Also regenerating the QH test plot

I'll approve and just approve again once we see tests pass.

dpanici (approved):

tests/test_examples.py
```diff
@@ -262,48 +261,60 @@ def run_qh_step(n, eq):
...
+    rho_err, theta_err = area_difference_desc(eq1, eq1a, Nr=1, Nt=1)
+    # only need crude tolerances here to make sure the boundaries are
+    # similar, the main test is ensuring its not pathological and has good qs
+    assert rho_err.mean() < 1
+
+    grid = LinearGrid(M=eq1a.M_grid, N=eq1a.N_grid, NFP=eq1a.NFP, sym=False, rho=1.0)
+    data = eq1a.compute(["|B|_mn", "B modes"], grid, M_booz=eq1a.M, N_booz=eq1a.N)
+    idx = np.where(np.abs(data["B modes"][:, 1] / data["B modes"][:, 2]) != 1)[0]
+    B_asym = np.sort(np.abs(data["|B|_mn"][idx]))[:-1]
+    np.testing.assert_array_less(B_asym, 1e-1)
```
ddudt, line 278: This isn't actually a good test for QS because of Issue #370. But we can fix this test as part of PR #399.

desc/perturbations.py
```diff
@@ -588,39 +590,56 @@ def optimal_perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
...
+        # some scaling to improve conditioning and rescale trust region
+        wf, _ = compute_jac_scale(Fx_reduced)
+        wg, _ = compute_jac_scale(Gx_reduced)
+        wx = wf + wg
+        Fxh = Fx_reduced * wx
+        Gxh = Gx_reduced * wx
+        if cutoff is None:
+            cutoff = np.finfo(Fx_reduced.dtype).eps * np.max(Fx_reduced.shape)
+        uf, sf, vtf = np.linalg.svd(Fxh, full_matrices=False)
+        sf += sf[-1]  # add a tiny bit of regularization
```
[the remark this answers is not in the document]
f0uriest: Adding a constant to the singular values reduces the condition number. Ideally we'd use a trust region thing but that's tricky to do for both x and c, and emperically I found that the trust region step was usually adding something on the order of the smallest singular value of the rescaled jacobian, so that's what I used here.

tests/test_examples.py
f0uriest: Since we changed the sign of the Jacobian, it changed the sign of theta for this case. So the helicity needs to change as well. Technically it should be (-1, -NFP) but figured this was less confusing.

dpanici (approved):

tests/test_examples.py
ddudt: OK yes I checked it and that's correct that the helicity is positive now.

ddudt (approved):

PR #334, opened 2022-12-05
Refactor ``compute_`` functions to take consistent arguments, add depdencies

### LINKED ISSUE #331: Make existing compute functions all take the same arguments of dictionaries/kwargs

### LINKED ISSUE #332: Add dependencies to all items in data index

f0uriest:
Part 1 of major refactoring of ``compute`` functions.
- All ``compute_`` functions now take the same arguments:
  - ``params`` : dict of ndarray
    Parameters from the equilibrium, such as R_lmn, Z_lmn, i_l, p_l, etc
  - ``transforms`` : dict of Transform
    Transforms for R, Z, lambda, etc
  - ``profiles`` : dict of Profile
    Profile objects for pressure, iota, current, etc
  - ``data`` : dict of ndarray
    Data computed so far, generally output from other compute functions
  - ``kwargs`` : dict
    Other arguments needed by some functions, such as helicity
- All items in ``data_index`` now have dependencies listed. These only include direct depdencies, the full dependecy tree is built recursively when the ``data_index`` is first imported. Dependencies include:
- ``eq.compute`` now accepts multiple names for computing multiple quantities on the same grid.
- ``grid``, ``data``, and other args now must be passed as keyword arguments to ``eq.compute``

Resolves #331 
Resolves #332


f0uriest:
Looks like this reduces compile times by ~50%!
A little surprised the compute/jacobian times are similar, my guess is that maybe JAX was able to optimize away a lot of the redundant calculations before, so all this does is simplify it a bit so that doesn't take as long but the resulting compiled function is similar.

tests/test_compute_funs.py
```diff
@@ -467,3 +513,13 @@ def test_compute_grad_p_volume_avg():
     eq = Equilibrium()  # default pressure profile is 0 pressure
     pres_grad_vol_avg = eq.compute("<|grad(p)|>_vol")["<|grad(p)|>_vol"]
     np.testing.assert_allclose(pres_grad_vol_avg, 0)
+
+
+@pytest.mark.unit
+def test_compute_everything():
+    """Make sure we can compute everything without errors."""
```
ddudt, line 520: Good, I was going to suggest making this test!

ddudt (approved):
This is awesome -- thanks for doing all this work!

desc/equilibrium/configuration.py
```diff
@@ -917,88 +914,69 @@ def L_basis(self):
...
-    def compute(self, name, grid=None, data=None, **kwargs):  # noqa: C901 - FIXME
+    def compute(
+        self,
+        *names,
+        grid=None,
+        params=None,
+        transforms=None,
+        profiles=None,
+        data=None,
+        **kwargs,
```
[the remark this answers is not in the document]
f0uriest: yeah it was the least bad option I could think of. One other approach would just be to have the function take `*args` and then parse them based on type but I thought that was too messy and could lead to some weird logic. This way felt like the cleanest.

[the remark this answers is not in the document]
ddudt: OK sounds good. I like the choice you made.

desc/compute/_field.py
```diff
@@ -486,7 +355,7 @@ def compute_magnetic_field_magnitude(
             * (
                 data["B^zeta_t"]
                 * (
-                    dot(data["e_theta_t"], data["e_zeta"])
+                    dot(data["e_theta_t"], data["e_zeta"])  # TODO: this is just g_tz_t
```
dpanici, line 358: if this is a dependency, could we reuse g_tz_t calculation?

desc/compute/data_index.py
```diff
@@ -18,6 +26,12 @@
     "description": "Radial coordinate, proportional to the square root of the toroidal flux",
     "fun": "compute_flux_coords",
     "dim": 1,
+    "dependencies": {
+        "params": [],
+        "transforms": {"grid": []},
```
dpanici, line 31: what does having a "grid" in transforms dict do? does this specify that ONLY the grid is used from the transform and not any of the evaluated coefficients/derivaitves ?

desc/compute/utils.py
```diff
@@ -11,45 +12,308 @@
...
-    key : str
-        Key indicating a quantity from data_index.
-    R_transform : Transform, optional
-        Transforms R_lmn coefficients to real space.
-    Z_transform : Transform, optional
-        Transforms Z_lmn coefficients to real space.
-    L_transform : Transform, optional
-        Transforms L_lmn coefficients to real space.
+    names : str
+        Names of the quantity to compute.
```
dpanici, line 21: tiny comment but should be quantity(s) here

desc/compute/utils.py
```diff
@@ -11,45 +12,308 @@
...
+
+    Returns
+    -------
+    derivs : dict of list of int
+        Orders of derivatives needed to compute key.
+        Keys for R, Z, L, etc
+    """
+
+    def _get_derivs_1_key(key):
+        if "full_dependencies" in data_index[key]:
```
dpanici, line 111: when would this not be the case? since as soon as the compute module is initialized it build the full tree

desc/compute/data_index.py
f0uriest: Yeah basically. There are some things that need to know about the grid, and we used to usually get the one from R transformer, but now we won't always have any given transform, so figured it's better to just rely on the grid by itself (which should be the same as the grid in all the transforms)

desc/compute/utils.py
f0uriest: We also call these during the initialization to build the full dependency tree, so need this check. Also if for some reason you added something to the data index at runtime (which should be possible and possibly useful in the future)

desc/compute/utils.py
```diff
@@ -11,45 +12,308 @@
...
+    """
+    return (
+        _has_data(qty, data)
+        and _has_params(qty, params)
+        and _has_profiles(qty, profiles)
+        and _has_transforms(qty, transforms)
+    )
+
+
+def _has_data(qty, data):
```
dpanici, line 286: slightly confusing to have this named "has_data" when it really is "_does_not_have_data" since it will return False if the data is present in the data dict already

dpanici (approved):

desc/compute/utils.py
f0uriest: its really "has data dependencies" (ie all the things needed to compute x, and if we already have x there's no reason to compute it again) but that was too long.

ddudt (approved):

PR #336, opened 2022-12-05
Improved Benchmarks

### LINKED ISSUE #281: Benchmarks on CI run out of memory
Ugh. We might need to reduce the number of trials or overall number of benchmarks, or see if we can split the job in a sensible way.

Also it would be nice if we modified the action to always use the `benchmark.py` from the current branch, so that if we make changes to that we can test as part of a single PR. Right now we use the PR branch version of `benchmark.py` for running code on the PR branch, but then compare to running the master branch version of `benchmark.py`. Ideally we wouldn't be changing the file too often, but it makes debugging stuff a pain.

f0uriest:
Breaks benchmarks into "small" and "large" varieties, along with different ones for cpu and gpu.
Small benchmarks just profiles individual objective and jacobian calls, and should be small enough to work on CI without running out of memory but still give an indication of performance changes.
Large benchmarks profile full equilibrium solves and optimizations and should be run manually on a cluster.

Note that running benchmarks on this PR won't work because the benchmark files on master are different.

Resolves #281

### DIFF
--- a/tests/benchmarks/benchmark_gpu_large.py
+++ b/tests/benchmarks/benchmark_gpu_large.py
@@ -1,10 +1,13 @@
-"""Benchmarks for timing comparison."""
+"""Benchmarks for timing comparison on gpu backend."""
 
 import os
 
 import numpy as np
 import pytest
 
+import desc
+
+desc.set_device("gpu")
 from desc.__main__ import main
 from desc.basis import FourierZernikeBasis
 from desc.equilibrium import EquilibriaFamily
@@ -39,7 +42,7 @@ def SOLOVEV(tmpdir_factory):
     print("exec_dir=", exec_dir)
     print("cwd=", cwd)
 
-    args = ["-o", str(desc_h5_path), input_filename, "--numpy", "-vv"]
+    args = ["-o", str(desc_h5_path), input_filename, "--numpy", "-vv", "-g"]
     main(args)
 
     SOLOVEV_out = {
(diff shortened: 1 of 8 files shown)


ddudt (approved):

tests/benchmarks/benchmark_gpu_large.py
```diff
@@ -0,0 +1,308 @@
...
+    """Benchmark 1st order perturbations."""
+
+    def setup():
+        eq = EquilibriaFamily.load(load_from=str(SOLOVEV["desc_h5_path"]))[-1]
+        objective = get_equilibrium_objective()
+        constraints = get_fixed_boundary_constraints()
+        tr_ratio = [0.01, 0.25, 0.25]
+        dp = np.zeros_like(eq.p_l)
+        objective.build(eq)
+        dp[np.array([0, 2])] = 8e3 * np.array([1, -1])
```
dpanici, line 259: should np in this file be replaced with jnp?

tests/benchmarks/benchmark_gpu_small.py
```diff
@@ -0,0 +1,346 @@
+"""Benchmarks for timing comparison on gpu (that are small enough to run on CI)."""
+
+import numpy as np
```
dpanici, line 3: same for this file

tests/benchmarks/benchmark_gpu_large.py
f0uriest: no it should be fine, it's just used for setting up the problem. The actual compute will still run on the gpu

dpanici (approved):
approved subject to seeing if np should be jnp in gpu tests

PR #341, opened 2022-12-06
VMEC load

ddudt:
Improves how the boundary conditions are applied when calling `VMECIO.load`. Also fixes a few minor miscellaneous things.


desc/objectives/linear_objectives.py
```diff
@@ -89,6 +89,8 @@ def build(self, eq, use_jit=True, verbose=1):
             idx = np.arange(eq.surface.R_basis.num_modes)
         else:  # specified modes
             modes = np.atleast_2d(self._modes)
+            if np.isfortran(modes):
```
f0uriest, line 92: what is this needed for? I think ``isfortran`` is deprecated (https://numpy.org/doc/stable/reference/generated/numpy.isfortran.html)

Could we just always make it a contiguous array if its really needed?

f0uriest (approved):

dpanici (approved):

PR #343, opened 2022-12-08
Fix indexing bug in biot-savart

f0uriest:
Arrays were being broadcast incorrectly, resulting in an error if there was more than 1 evaluation point.


dpanici (approved):

dpanici:
Test for this fix though?

dpanici (approved):

PR #344, opened 2022-12-08
Add TargetIota objective

f0uriest:
Bringing this back from the dead. Now that we can compute with fixed current, it's useful to be able to target a particular iota profile.


desc/objectives/_generic.py
```diff
@@ -320,4 +331,152 @@ def compute(self, R_lmn, Z_lmn, L_lmn, i_l, c_l, Psi, **kwargs):
...
+        Note: has no effect for this objective.
+    grid : Grid, optional
+        Collocation grid containing the nodes to evaluate at.
+    name : str
+        Name of the objective function.
+    """
+
+    _scalar = False
+    _linear = False
+    _units = "(n/a)"
```
ddudt, line 366: In other objectives (like aspect ratio) we say `"(dimensionless)"`. We should be consistent -- which do we prefer?

tests/test_objective_funs.py
```diff
@@ -99,14 +100,26 @@ def test(eq):
...
         test(Equilibrium(node_pattern="quad", current=PowerSeriesProfile(0)))
 
+    @pytest.mark.unit
+    def test_target_iota(self):
+        """Test calculation of iota profile."""
+
+        def test(eq):
+            obj = TargetIota(target=1, weight=2, eq=eq)
+            iota = obj.compute(eq.R_lmn, eq.Z_lmn, eq.L_lmn, eq.i_l, eq.c_l, eq.Psi)
+            np.testing.assert_allclose(iota, -2 / 3)
```
ddudt, line 110: Actual iota is 0 but the target is 1, so the error is -1. Then with a weight of 2 I would expect the result to be -2. How do we get -2/3?

ddudt (dismissed):

tests/test_objective_funs.py
f0uriest: This is the grid spacing thing I mentioned. The residual is `(f-target)*weight*spacing` and in this case the default grid has `L_grid` radial nodes (and `L_grid` is 3

desc/objectives/_generic.py
f0uriest: yeah dimensionless is probably better ill change it

tests/test_objective_funs.py
ddudt: Ah OK. Sounds good.

ddudt (approved):

dpanici (approved):

PR #346, opened 2022-12-08
Elongation

ddudt:
Creates a new objective function to compute maximum elongation, defined as semi-major radius / semi-minor radius. 

This is based on a Ramanujan approximation for the perimeter of an ellipse. There are better approximations that could be used, but the math gets a lot more complicated and this one seems good enough. 

I couldn't add `"elongation"` to the `data_index` because it requires two separate grids: a quadrature grid to compute the surface areas, and a linear grid on the boundary surface to compute the perimeters. We could use the same LinearGrid for both computations, but that would be sub-optimal for the area calculations. So all of the new logic is only in the objective function itself. I'm open to suggestions for improvement.


f0uriest:
What about using the quadrature grid for the perimeter as well? For even modest values of L there are nodes almost to the edge, and we could also add a correction like 1/rhomax if needed?
image

f0uriest:
I think that could be handled by the `surface_integral` function, by taking the perimeter of each flux surface and then just indexing out the last one.
That also raises another question: This is elongation on a single cross section right? Would it matter which zeta the grid is on? I think for other stuff like area or minor radius we do an average over all the zetas in the grid.

ddudt:
> We probably could modify `surface_integrals` to add an option for it. But we really want a "line integral" instead of a "surface integral" in this case, so it might be best to add a new util function.

So this works, but for the HELIOTRON example (which has exact elliptical surfaces and L_grid=36) the error in the elongation increases from 1e-5 with the linear grid to 1e-2 with the quadrature grid. Probably still acceptably small, but disappointing.

ddudt:
> We probably could modify `surface_integrals` to add an option for it. 

So this works, but for the HELIOTRON example (which has exact elliptical surfaces and L_grid=36) the error in the elongation increases from 1e-5 with the linear grid to 1e-2 with the quadrature grid. Probably still acceptably small, but disappointing.

f0uriest:
That seems kind of weird. The poloidal spacing should be the same, and the radial max is 0.9989 compared to 1 so it should only be like 0.1% worse? And what about correcting by dividing through by 0.9989?

Also is that relative or absolute error in the perimeter? And how does that translate to error in the elongation?

desc/compute/_core.py
```diff
@@ -838,18 +848,43 @@ def compute_geometry(
             )
         )
     if check_derivs("V_r(r)", R_transform, Z_transform):
+        P = compress(  # perimeter
```
f0uriest, line 851: would this (and major/minor radii) be useful things to add to the data index as well?

desc/compute/_core.py
```diff
@@ -838,18 +848,43 @@ def compute_geometry(
...
-        data["A"] = jnp.mean(
-            surface_integrals(
-                grid, jnp.abs(data["sqrt(g)"] / data["R"]), surface_label="zeta"
-            )
-        )
+        data["A"] = jnp.mean(A)
         data["R0"] = data["V"] / (2 * jnp.pi * data["A"])
         data["a"] = jnp.sqrt(data["A"] / jnp.pi)
         data["R0/a"] = data["V"] / (2 * jnp.sqrt(jnp.pi * data["A"] ** 3))
+        data["a/b"] = jnp.max(a / b)
```
f0uriest, line 887: might be a bit confusing since the "a" here is different than data["a"]? Could call it "kappa" (common term for elongation in tokamaks) or just elongation?

Also do we want to make it mean instead of max? so it agrees more with the way we compute area?

desc/compute/_core.py
ddudt: I don't think so, because it's the perimeter of each toroidal cross-section. I can't think of anything else that would be useful for. And most of the other geometric quantities computed here are 0-dimension scalar values for the whole device.

desc/compute/_core.py
ddudt: Maximum elongation is what we care about penalizing in optimization. If we used mean instead, we could end up with a circular cross-section (elongation=0) that "cancels out" another cross-section with very high elongation, and that would be bad.

ddudt: Regarding the name: I'm not sure what is best. I've seen "epsilon" in the stellarator literature, but that is also used for a lot of other things. I wanted to keep something similar to the other names we use like "R0/a" for aspect ratio, etc. But if you think "a" is too confusing I would change it to "elongation".

ddudt: changed to `"a_major/a_minor"`

f0uriest (dismissed):

f0uriest (approved):

desc/compute/_geometry.py
```diff
@@ -182,6 +182,57 @@ def _R0_over_a(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="a_major/a_minor",
+    label="a_{major} / a_{minor}",
+    units="~",
+    units_long="None",
+    description="Elongation",
```
dpanici, line 190: Can we specify somewhere that for stellarators this is the maximum elongation across all zeta angles? this is not clear from just looking at the description or the docstrings

desc/compute/utils.py
```diff
@@ -538,6 +541,14 @@ def surface_integrals(grid, q=jnp.array([1]), surface_label="rho"):
     q = jnp.atleast_1d(q)
     nodes, unique_idx, ds, max_surface_val = _get_grid_surface(grid, surface_label)
 
+    if max_surface:
+        max_rho = grid.nodes[grid.unique_rho_idx[-1], 0]
+        idx = np.nonzero(grid.nodes[:, 0] == max_rho)[0]
+        q = q[idx]
+        nodes = nodes[idx]
+        unique_idx = (unique_idx * nodes.size / grid.num_nodes).astype(int)
```
dpanici, line 549: can you explain this line?

dpanici (changes requested):
Add to the docstring somewhere (either description of `elongation` in data_index or the docstring of the objective) that it is the max elongation over all zeta points in the grid used

desc/compute/utils.py
ddudt: After extracting only the nodes at the max rho surface in line 548, `unique_idx` no longer corresponds to the correct values in the `nodes` array. `nodes` now only corresponds to nodes on a single flux surface, but `unique_idx` is still pointing to indices in the full list from `grid.nodes`. Based on how the grids are sorted this is corrected by dividing all of the indices by the number of rho surfaces. 

Multiplying by `nodes.size / grid.num_nodes` is equivalent to dividing by `grid.num_rho`, the number of rho surfaces. I made that change to make it more clear.

dpanici (approved):

f0uriest (dismissed):

f0uriest (dismissed):

f0uriest (approved):

PR #349, opened 2022-12-13
Fix composite profiles and add profiles for kinetic quantities

### LINKED ISSUE #327: Allow equilibrium to have pressure/density/temperature profiles

### LINKED ISSUE #347: adding/multiplying/scaling profiles for equilibrium pressure
For the j_bootstrap work I tried to specify an equilibrium pressure profile by multiplying and summing density and temperature profiles, using `ScaledProfile`, `SumProfile` and `ProductProfile`. The equilibrium solve then gave an error. Minimal example:
~~~~py
import numpy as np
from scipy.constants import elementary_charge

from desc.objectives import *

ne = PowerSeriesProfile(3.0e19 * np.array([1, -1]), modes=[0, 10])
Te = PowerSeriesProfile(2.0e3 * np.array([1, -1]), modes=[0, 2])
Ti = Te
pressure = elementary_charge * (ne * Te + ne * Ti)
print("pressure params:", pressure.params)

LM_resolution = 6
eq = Equilibrium(
    pressure=pressure,
    iota=PowerSeriesProfile([1.61]),
    Psi=np.pi,  # so B ~ 1 T                                                                                
    NFP=1,
    L=LM_resolution,
    M=LM_resolution,
    N=0,
    L_grid=2 * LM_resolution,
    M_grid=2 * LM_resolution,
    N_grid=0,
    sym=True,
)

eq.solve(
    constraints=get_fixed_boundary_constraints(),
    objective=ObjectiveFunction(objectives=ForceBalance()),
)
eq.save(__file__ + ".h5")
~~~~
Should this work, in principle? Output, using master from today:
~~~~
pressure params: (1.602176634e-19, ((array([ 3.e+19,  0.e+00,  0.e+00,  0.e+00,  0.e+00, -3.e+19]), array([ 2000., -2000.])), (array([ 3.e+19,  0.e+00,  0.e+00,  0.e+00,  0.e+00, -3.e+19]), array([ 2000., -2000.]))))
Building objective: force
Precomputing transforms
  File "/Users/mattland/Library/CloudStorage/Box-Box/work22/./20221212-09_desc_pressure_minimal_example", line 34, in <module>
    eq.solve(
  File "/Users/mattland/DESC/DESC/desc/equilibrium/equilibrium.py", line 473, in solve
    result = optimizer.optimize(
  File "/Users/mattland/DESC/DESC/desc/optimize/optimizer.py", line 290, in optimize
  File "/Users/mattland/DESC/DESC/desc/objectives/objective_funs.py", line 173, in build
    objective.build(eq, use_jit=self.use_jit, verbose=verbose)
  File "/Users/mattland/DESC/DESC/desc/objectives/_equilibrium.py", line 156, in build
    super().build(eq=eq, use_jit=use_jit, verbose=verbose)
  File "/Users/mattland/DESC/DESC/desc/objectives/objective_funs.py", line 598, in build
    self._set_dimensions(eq)
  File "/Users/mattland/DESC/DESC/desc/objectives/objective_funs.py", line 502, in _set_dimensions
    self._dimensions["p_l"] = eq.pressure.params.size
AttributeError: 'tuple' object has no attribute 'size'
~~~~
We can kind of get around this by adding `pressure = pressure.to_powerseries(order=12)`, but the fitting changes the coefficients a bit.

f0uriest:
- `Profile.params` for composite profiles (`Sum`, `Product`, etc) now returns a concatenated array rather than a tuple, and each class has its own way of parsing an array of params into its constituent parts
- Adds attributes to `Equilibrium` for kinetic profiles of temperature and density
- Refactors how the profiles are parsed in `__init__` to make it a bit cleaner

- [x] Make `compute_pressure` work with kinetic profiles
- [x] Add constraints to fix kinetic profiles
- [x] Add args for kinetic profiles to objectives

Resolves #347 
Resolves #327


desc/profiles.py
```diff
@@ -422,13 +446,10 @@ def compute(self, params=None, grid=None, dr=0, dt=0, dz=0):
...
         """
-        if params is None:
-            params = [None] * len(self._profiles)
-        if isinstance(params, dict):
-            params = [params]
+        params = self._parse_params(params)
         f = 0
         for i, profile in enumerate(self._profiles):
-            f += profile.compute(params[i % len(params)], grid, dr, dt, dz)
+            f += profile.compute(params[i], grid, dr, dt, dz)
```
unalmis, line 452: why did we need to modulo this before, and why do we not need to do it now

f0uriest: the new getter/setter and _parse param methods should ensure that params is always the right length

unalmis (approved):
looks good to me, but I'd recommend adding a test to compare equilibrium output when solving with the new kinetic quantities (maybe with perturbation too)

desc/equilibrium/configuration.py
```diff
@@ -309,51 +332,71 @@ def __init__(  # noqa: C901 - FIXME: break this up into simpler pieces
         self._pressure = None
         self._iota = None
         self._current = None
+        self._electron_temperature = None
+        self._electron_density = None
+        self.ion_temperature = None
```
ddudt, line 337: Missing a leading underscore? `self._ion_temperature`

desc/equilibrium/configuration.py
```diff
@@ -837,12 +886,143 @@ def pressure(self, new):
...
 
+    @property
+    def electron_temperature(self):
+        """Profile: Electron temperature (eV) profile."""
+        return self._electron_temperature
+
+    @electron_temperature.setter
+    def electron_temperature(self, new):
+        if isinstance(new, Profile) or (new is None):
+            self._electron_temperature = new
```
ddudt, line 908: Could we use `self.parse_profile` here to allow other input formats besides Profile types? Same for all the other profile setter methods.

desc/objectives/_equilibrium.py
```diff
@@ -159,6 +159,14 @@ def compute(self, *args, **kwargs):
             Spectral coefficients of I(rho) -- toroidal current profile.
         Psi : float
             Total toroidal magnetic flux within the last closed flux surface (Wb).
+        Te_l : ndarray
+            Spectral coefficients of Te(rho) -- electron temperature profile.
+        ne_l : ndarray
+            Spectral coefficients of ne(rho) -- electron density profile.
+        Ti_l : ndarray
+            Spectral coefficients of Ti(rho) -- ion temperature profile.
```
ddudt, line 167: We should include units for these parameters -- `(eV)` and `(1/m^3)` -- like we do for all the other parameters that are not dimensionless. Same for all other objective function documentations.

desc/objectives/linear_objectives.py
```diff
@@ -917,6 +922,436 @@ def target_arg(self):
...
+            self._normalization = scales["p"]
+        super().build(eq, profile, use_jit, verbose)
+
+    def compute(self, Te_l, **kwargs):
+        """Compute fixed electron temperature errors.
+
+        Parameters
+        ----------
+        Te_l : ndarray
+            parameters of the electron temperature profile.
```
ddudt, line 1016: Should add units here too (same with all other compute methods in this file that have dimensional parameters).

tests/test_profiles.py
```diff
@@ -130,19 +142,23 @@ def test_sum_profiles(self):
...
+        zp = -pp.to_fourierzernike()
 
-        f = pp + sp - (-zp)
+        f = pp + sp - zp
         x = np.linspace(0, 1, 50)
         f.grid = 50
         np.testing.assert_allclose(f(), 3 * (pp(x)), atol=1e-3)
 
         params = f.params
+        assert params.size == len(sp.params) + len(pp.params) + len(zp.params) + 1
```
ddudt, line 153: Where does the extra `+ 1` come from? And why is it not needed in line 180 of the product profile test?

tests/test_profiles.py
```diff
@@ -240,3 +297,127 @@ def test_profile_conversion(self):
...
+            L_grid=2 * LM_resolution,
+            M_grid=2 * LM_resolution,
+            N_grid=0,
+            sym=True,
+        )
+        eq1.solve(
+            constraints=get_fixed_boundary_constraints(kinetic=False),
+            objective=ObjectiveFunction(objectives=ForceBalance()),
+            maxiter=5,
+        )
```
ddudt, line 345: FYI this should be equivalent to the simpler: `eq1.solve(objective-"force", maxiter=5)`. Same for `eq2`.

tests/test_profiles.py
```diff
@@ -240,3 +297,127 @@ def test_profile_conversion(self):
...
+        assert np.all(np.isnan(data1["ne"]))
+        assert np.all(np.isnan(data1["Te"]))
+        assert np.all(np.isnan(data1["Ti"]))
+        assert np.all(np.isnan(data1["Zeff"]))
+        assert np.all(np.isnan(data1["ne_r"]))
+        assert np.all(np.isnan(data1["Te_r"]))
+        assert np.all(np.isnan(data1["Ti_r"]))
+        assert np.all(np.isnan(data1["Zeff_r"]))
+        assert np.all(data2["Te"] == data2["Ti"])
+        assert np.all(data2["Te_r"] == data2["Ti_r"])
```
ddudt, line 421: No similar tests for density?

ddudt (changes requested):
My main comment is that I think it would be good to add a profile for ion density, as an option for users to give instead of the atomic number profile. Also I think we should add to the atomic number profile documentation how it is used to compute the ion density.

desc/compute/utils.py
```diff
@@ -12,7 +12,21 @@
...
+    "Z_lmn",
+    "L_lmn",
+    "p_l",
+    "i_l",
+    "c_l",
+    "Psi",
+    "Te_l",
+    "ne_l",
+    "Ti_l",
+    "Zeff_l",
```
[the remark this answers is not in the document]
f0uriest: My idea was to preserve the existing order where possible, the exception being the boundary terms because in some places we assume those are the last elements.

desc/compute/_profiles.py
```diff
@@ -144,10 +312,15 @@ def _gradpsi_mag(params, transforms, profiles, data, **kwargs):
...
-    data=[],
+    data=["Te", "ne", "Ti", "Zeff"],
 )
 def _p(params, transforms, profiles, data, **kwargs):
-    data["p"] = profiles["pressure"].compute(params["p_l"], dr=0)
+    if profiles["pressure"] is not None:
+        data["p"] = profiles["pressure"].compute(params["p_l"], dr=0)
+    else:
+        data["p"] = elementary_charge * (
+            data["ne"] * data["Te"] + data["Ti"] * data["ne"] / data["Zeff"]
```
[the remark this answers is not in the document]
f0uriest: This is correct. The number of electrons should be equal to the number of ions * effective charge of each ion, so `ne = sum(ni_k) * Zeff` where `Zeff = sum(Zi_k*ni_k)/sum(ni_k)` for `k` different ion species.
The decision to use Zeff as opposed to ion density is based on the following:
- This seems to be a more common convention
- ion density is difficult to measure, so in practice it is usually estimated based on electron density and a rough estimate of Zeff
- Using ion density by itself wouldn't allow us to account for impurities which can be important for neoclassical and turbulent transport.
- Similarly, if we wanted to properly handle impurities we'd need profiles for multiple ions as well as their charge state which gets way more complicated
- Using Zeff like this guarantees global quasi-neutrality. Allowing the user to specify electron and ion density independently could result in them giving a non-neutral plasma in which case all of our equations are super wrong.

tests/test_profiles.py
f0uriest: It's because in `f = pp + sp - zp`, `-zp` first gets parsed as a scaled profile with scale=-1, then a sum profile. So there's an extra parameter for the -1. In the product profile we don't negate anything so no extra scale parameter.

tests/test_profiles.py
f0uriest: Yes I just wanted it to be more explicit and clearer what the difference was (kinetic vs non)

tests/test_profiles.py
f0uriest: If the temperatures and pressures are the same I don't think we need an explicit test for density. And it would be a pretty trivial test, just `ne==ne` since there's no extra math involved, its just a pass-through.

desc/compute/utils.py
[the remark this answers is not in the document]
ddudt: OK. At some point we should probably fix those parts of the code that assume the boundary terms are the final parameters (perturbations, anywhere else?) so that this arg order can be completely arbitrary.

desc/compute/_profiles.py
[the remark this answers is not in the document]
ddudt: Thanks for the explanation. As one counter-point from my experience, SFINCS uses the density profile for each species instead of Zeff. But it could be tricky for us to implement an arbitrary number of ion profiles for each species, and your point about enforcing quasi-neutrality is important. I'm content with this choice now.

ddudt (approved):

unalmis (approved):

desc/compute/_profiles.py
```diff
@@ -133,6 +133,174 @@ def _gradpsi_mag(params, transforms, profiles, data, **kwargs):
...
+
+
+@register_compute_fun(
+    name="Ti",
+    label="T_i",
+    units="eV",
+    units_long="electron-Volts",
+    description="Ion temperature",
+    dim=1,
+    params=["Te_l"],
```
dpanici, line 227: params here does not match params in the compute function, should be Ti_l

desc/objectives/linear_objectives.py
```diff
@@ -673,6 +673,11 @@ def build(self, eq, use_jit=True, verbose=1):
             Level of output.
 
         """
+        if eq.pressure is None:
```
dpanici, line 676: maybe a stupid question, but if a user has supplied kinetic profiles to an equilibrium but then just wants the pressure to be fixed without caring about the kinetic profiles... this probably is a bad use case is that ever a case we should worry about supporting? would mean nonlinear constraint tho since it is a constraint on the products of T and n

desc/objectives/linear_objectives.py
```diff
@@ -917,6 +922,436 @@ def target_arg(self):
...
+            Level of output.
+
+        """
+        if eq.electron_density is None:
+            raise RuntimeError(
+                "Attempting to fix electron density on an equilibrium with no "
+                + "electron density profile assigned"
+            )
+        profile = eq.electron_density
+        if self._normalize:
```
dpanici, line 1113: why have this hardcoded here instead of put in the normalizations dict? at least so that we can change if necessary later from one place and also have all of our normalizations visible in one file

desc/objectives/linear_objectives.py
```diff
@@ -917,6 +922,436 @@ def target_arg(self):
...
+        """
+        if eq.electron_temperature is None:
+            raise RuntimeError(
+                "Attempting to fix electron temperature on an equilibrium with no "
+                + "electron temperature profile assigned"
+            )
+        profile = eq.electron_temperature
+        if self._normalize:
+            scales = compute_scaling_factors(eq)
+            self._normalization = scales["p"]
```
dpanici, line 1007: is B^2 / 2 mu0 a good normalization for this? Te is probably O(1e3)-O(1e4) right? this pressure normalization for W7-X, to give context, is O(1e5). I guess that is in the ballpark but could be a bit high. Not saying to change but just commenting

desc/perturbations.py
```diff
@@ -371,7 +371,7 @@ def perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
                 value, np.where(np.abs(value) < 10 * np.finfo(value.dtype).eps)[0], 0
             )
             # don't set nonexistent profile (values are empty ndarrays)
-            if not (key == "c_l" or key == "i_l") or value.size:
+            if value.size:
```
dpanici, line 374: why did we have that not block for `c_l` and `i_l`?

desc/profiles.py
```diff
@@ -511,13 +545,20 @@ def compute(self, params=None, grid=None, dr=0, dt=0, dz=0):
...
-        for i, profile in enumerate(self._profiles):
-            f *= profile.compute(params[i % len(params)], grid, dr, dt, dz)
+        if dt > 0 or dz > 0:
+            raise NotImplementedError(
+                "Poloidal and toroidal derivatives of ProductProfiles have not "
+                + "been implemented yet"
+            )
+        params = self._parse_params(params)
+        f = 0
+        derivs = combination_permutation(len(self._profiles), dr)
```
dpanici, line 555: what does this block do?

dpanici (approved):
Left a few clarification requests, and one change that if you can justify you don't need to make (regarding hardcoding ne normalization inside the objectives, this seems to be the only place it appears so it is probably fine to leave as is). Otherwise is approved

dpanici (changes requested):
One small error in compute fxn that needs fixing, in the compute fxn for T_i the incorrect param is listed

desc/compute/_profiles.py
f0uriest: good catch

desc/objectives/linear_objectives.py
f0uriest: Yeah in that case it gets a lot trickier. I don't think it would be a common use case that we need to accommodate, at least not for now.

desc/objectives/linear_objectives.py
f0uriest: good point

desc/objectives/linear_objectives.py
f0uriest: its not perfect. I'll update it to be `scales['p'] / (scales['n'] * elementary_charge)`, but since density is usually O(1/elementary charge) it comes out about the same. The main issue is we don't have any inherent scale for n and T other than what we commonly see in experiments, but that's not "universal" in the same way that pressure balance etc is.

desc/perturbations.py
f0uriest: to prevent setting current on fixed iota eq and vice verse, but its not needed since in that only `c_l` or `i_l` would have size>0, so we only need to check the size.

unalmis (approved):

dpanici (approved):

ddudt (approved):

PR #350, opened 2022-12-13
Add implementation of `Surface.compute_curvature`

f0uriest:
Adds calculation of gaussian, mean, and principle curvatures to surface classes.


f0uriest:
Not sure why it says the new lines aren't covered because they definitely are. Might be more codecov shenanigans

desc/geometry/surface.py
```diff
@@ -306,9 +306,54 @@ def _get_transforms(self, grid=None):
...
+            fourier coefficients for R, Z. Defaults to self.R_lmn, self.Z_lmn
+        grid : Grid or array-like
+            toroidal coordinates to compute at. Defaults to self.grid
+            If an integer, assumes that many linearly spaced points in (0,2pi)
+
+        Returns
+        -------
+        K, H, k1, k2 : ndarray, shape(k,)
+            Gaussian, mean and 2 principle curvatures at points specified in grid.
+        """
```
ddudt, line 344: This function should be moved to the parent `Surface` class, rather than copying it for both types of surfaces.

desc/geometry/surface.py
```diff
@@ -306,9 +306,54 @@ def _get_transforms(self, grid=None):
...
+        E, F, G = self._compute_first_fundamental_form(R_lmn, Z_lmn, grid)
+        L, M, N = self._compute_second_fundamental_form(R_lmn, Z_lmn, grid)
+        a = E * G - F**2
+        b = F * M - L * G - E * N
+        c = L * N - M**2
+        r1 = (-b + jnp.sqrt(b**2 - 4 * a * c)) / (2 * a)
+        r2 = (-b - jnp.sqrt(b**2 - 4 * a * c)) / (2 * a)
```
ddudt, line 351: I trust this is correct, but I don't understand where your equations are coming from. Don't you only need the second fundamental form to calculate the principal curvature?

ddudt (changes requested):

desc/geometry/surface.py
dpanici: Agreed, but can make the compute fundamental forms just be not defined in the parent class as they need specific implementation in the subclass

desc/geometry/surface.py
dpanici: Also interested in the math ^, not familiar with fundamental forms but this link seems to suggest to me that K = c/a from Def 3.1?

https://math.uchicago.edu/~may/REU2012/REUPapers/Chase.pdf

ddudt (changes requested):
It looks like you forgot to remove the `compute_curvature` method from the `FourierRZToroidalSurface` class.

ddudt (approved):

dpanici (approved):

Issue #228, opened 2022-06-30
Useful Tutorial/Examples


dpanici:
Listing here some ideas for useful things to have examples for, anyone feel free to add

- [ ] How to access/use automatic differentiation derivative information  (say, if you loaded in a DESC eq to your other code, how can you access useful derivatives to avoid doing finite differences thru equilibria?) - would need to use JAX anyways -> related to #861 
- [x] How to write an Objective Function : explain `build` method more, what passing `derivs=` to a transform means, why we need to copy `iota` and `pressure` when an objective function depends on those two quantities (so if we modify the grid we don't modify the existign grid, which may be linear while we usually use ConcentricGrid for the collocation method)
- [x] small example showing how to check the objective compute value (objective.compute(*objective.xs(eq)))
- [ ] Update dev_guide as current has since been added as well as an argument
- [x] explain how fixed current mode and fixed iota mode differ

dpanici:
created branch [dev_guide](https://github.com/PlasmaControl/DESC/tree/dev_guide) to start writing explanations of various code design choices. 

Would be useful to also include input here from @unalmis and @pkim1818 as the most recent people to learn how to develop the code, feel free to add things to the [notebook](https://github.com/PlasmaControl/DESC/tree/dev_guide/docs/notebooks/dev_guide) in the branch

dpanici:
#328

dpanici:
Something where we at least show like what exactly `jac_scaled` is taking jacobian of could be useful.

dpanici:
#1304

PR #353, opened 2022-12-14
Dev guide

dpanici:
- [x] split dev guide notebooks into smaller parts
- [ ] fill out empty notebooks
- [ ] update compute and objectives notebook to reflect the changes in master that will come from compute function refactor


docs/dev_guide/backend.rst
```diff
@@ -0,0 +1,23 @@
...
+With that goal, in functions which can benefit from JAX, we use the following import statement: ``from desc.backend import jnp``.
+``desc.backend.jnp`` is an alias to ``jax.numpy`` if JAX is installed and ``numpy`` otherwise.
+
+While ``jax.numpy`` attempts to serve as a drop in replacement for ``numpy``, it imposes some constraints on how the code is written.
+For example, ``jax.numpy`` arrays are immutable.
+This means in-place updates to elements in arrays is not possible.
+To update elements in ``jax.numpy`` arrays, memory needs to be allocated to create a new array with the updated element.
+Similarly, JAX's JIT compilation requires code flow structures such as loops and conditionals to be written in a specific way.
+
+The utility functions in ``desc.backend`` provide a simple interface to perform these operations.
```
unalmis, line 23: We should add answers to (good) questions like this here:
https://github.com/PlasmaControl/DESC/pull/854#discussion_r1501910822
https://github.com/PlasmaControl/DESC/pull/854#discussion_r1501911035

unalmis:
- [ ] Add Yigit's [suggestions](https://github.com/PlasmaControl/DESC/pull/1212#issuecomment-2303623158) to grid guide

YigitElma:
> * [x]  Add Yigit's [suggestions](https://github.com/PlasmaControl/DESC/pull/1212#issuecomment-2303623158) to grid guide

This is done on #1304

YigitElma:
@unalmis @dpanici There are some duplicate notebooks on `dev_guide` and `notebooks/dev_guide` folders. Are there any differences with those? Can you delete the old one and put the most updated one to the `notebooks/dev_guide` folder?

Also, what is `compute_2.ipynb`?

docs/dev_guide/compute.rst
```diff
@@ -0,0 +1,96 @@
...
+  ``"rtz"`` indicating it is a function of :math:`\rho, \theta, \zeta`. Profiles and flux surface
+  quantities would have ``coordinates="r"`` indicating it only depends on `:math:\rho`
+* ``data``: What other physics quantities are needed to compute this quantity. In our
+  example, we need the radial derivative of pressure ``p_r``, the Jacobian determinant
+  ``sqrt(g)``, and contravariant components of current and magnetic field. These dependencies
+  will be passed to the compute function as a dictionary in the ``data`` argument. Note
+  that this only includes direct dependencies (things that are used in this function).
+  For example, we need ``sqrt(g)``, which itself depends on ``e_rho``, etc. But we don't
+  need to specify ``e_rho`` here, that dependency is determined automatically at runtime.
+* ``kwargs``: If the compute function requires any additional arguments they should
```
dpanici, line 80: add link to a cmpute fxn that uses this, or change it to be `f_C` so `helicity` kwarg is shown #1332

docs/dev_guide/adding_compute_funs.rst
```diff
@@ -1,5 +1,6 @@
+=============================
 Adding new physics quantities
------------------------------
+=============================
```
dpanici, line 3: don't neeed to change this one

dpanici:
Use master for grid.ipynb as well

YigitElma:
Try to solve #1402, #228 and #775 here.

docs/index.rst
```diff
@@ -52,6 +52,20 @@
    notebooks/tutorials/EffectiveRipple.ipynb
 
 
+.. toctree::
```
f0uriest, line 55: move this after the API docs

docs/dev_guide/notebooks/transform.ipynb
```diff
@@ -0,0 +1,255 @@
+{
```
f0uriest, line 1: <p>they convert the PDE to a set of nonlinear algebraic equations, not ODEs</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/353/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

f0uriest, line 1: <p>Remove reference to <code>unique</code> option since this will be removed soon </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/353/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

f0uriest (changes requested):
Few minor fixes

docs/dev_guide/notebooks/backend.ipynb
```diff
@@ -0,0 +1,227 @@
+{
```
unalmis, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    %timeit jnp.linalg.qr(A_jnp)
</pre><p>The reason this is faster may only be because it never gets computed. Please append <code>.block_until_ready()</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/353/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

unalmis, line 1: <p>If we are going to mention this, then we should say "For example..., or we need the operation to be performed in a static manner rather than traced dynamically at compile time." That is the main reason I would use numpy instead of JAX when developing.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/353/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

unalmis, line 1: <ul><li>I would rather it say "plan is to remove  the<code>numpy</code> backend since code that relies on the numpy backend instead of JAX is not automatically tested for correctness by the GitHub CI."</li><li>I think we can still do finite difference with JAX? just not by default. maybe don't include  "for those who wat to compare the results with finite difference".</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/353/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

docs/dev_guide/notebooks/transform.ipynb
unalmis, line 1: <p>"On the other hand, the <code>fft</code> method uses a Fast (Discrete) Fourier Transform (FFT) which whose linearithmic performance is favorable to quadratic scaling of matrix transforms."</p><p> </p><p>Why do I suggest this: to reduce undefined variables $n$ which here tries to function as both the spectral resolution and number of points which is not usually the case. Also adding Discrete to clarify that we don't compute the direct approximation to the fourier transform rather the fourier series interpolation.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/353/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

unalmis (dismissed):
left some comments, would be nice to make those quick updates as you address Rory's

unalmis:
and thanks

docs/dev_guide/notebooks/backend.ipynb
YigitElma: I think for this operation you don't need. Usually, if a function returns multiple stuff (like Q and R here), you don't need to use `block_until_ready()`.
It throws,
`AttributeError: 'QRResult' object has no attribute 'block_until_ready'`

unalmis: Don't you need to append `.block_until_ready` to each of the jax arrays returned in that tuple then?
a, b = QR()
a.block...
b.block...

YigitElma: <p>But in general, I am also not sure which example should I give, because the speed depends on the hardware and matrix sizes a lot.</p>

YigitElma: <p>Good point, but can you use it inside jit function? I don't think so.</p>

unalmis: We do use numpy inside jitted and differentiable functions

unalmis: if you google the "static jax" the relevant doc comes up
> Use numpy for operations that you want to be static; use jax.numpy for operations that you want to be traced.

unalmis: new developers will read the desc documentation before the jax documentation, so it may mislead them to include a timing example like this that does not include block until ready.

if we still want to have this example to prove jax can be faster, then maybe mention usually you need block until ready, or link to the jax docs that do.

YigitElma: But I think that has some other drawbacks like multiple compilations etc. For every different value of the static input, the function has to compiled again. Anyway, I am making some changes.

docs/dev_guide/notebooks/transform.ipynb
YigitElma: I think the changes I made should clarify this.

unalmis (approved):

f0uriest (approved):

PR #355, opened 2022-12-15
Bumb JAX version

f0uriest:
- Minor fixes to follow JAX's requirement that inputs to `jnp` functions be `ndarrays`
- Increases maximum JAX version to `0.4.1` (latest as of 12/14/22)
- Removes `jaxlib` from requirements in favor of specifying `jax[cpu]` which is now the preferred syntax, and will automatically install a compatible version of `jaxlib`

Tested on all JAX version `0.2.11` thru `0.4.1`


devtools/dev-requirements_conda.yml
```diff
@@ -14,8 +14,7 @@ dependencies:
   - termcolor
   - pip
   - pip:
-    - jax >= 0.2.11, <= 0.2.25
-    - jaxlib >= 0.1.69, <= 0.1.76
+    - jax[cpu] >= 0.2.11, <= 0.4.1
```
ddudt, line 17: We only test on `jax >= 0.3.0`; should we update the minimum requirement?

requirements.txt
```diff
@@ -1,7 +1,6 @@
 colorama
 h5py >= 3.0.0
-jax >= 0.2.11, <= 0.2.25
-jaxlib >= 0.1.69, <= 0.1.76
+jax[cpu] >= 0.2.11, <= 0.4.1
```
ddudt, line 3: Same comment about updating the minimum requirement to `jax >= 0.3.0`.

requirements_conda.yml
```diff
@@ -12,6 +12,5 @@ dependencies:
   - termcolor
   - pip
   - pip:
-    - jax >= 0.2.11, <= 0.2.25
-    - jaxlib >= 0.1.69, <= 0.1.76
+    - jax[cpu] >= 0.2.11, <= 0.4.1
```
ddudt, line 15: Same comment about updating the minimum requirement to `jax >= 0.3.0`.

ddudt (approved):
I feel like we should update the minimum required version of JAX to `0.3.0` since that's what we test with. It's nice to accommodate a wider range, but it could potentially lead to problems if things aren't backwards-compatible. And we should encourage people to use the latest version of JAX.

PR #359, opened 2022-12-16
Remove use of np.std in utils.py to avoid numerical errors fo…

### LINKED ISSUE #358: Transform falls back to direct1 Method when toroidal resolution N>=20
need to still find exactly where/why this is happening, but when using `Equilibrium.from_near_axis`, I find that at high enough toroidal resolution, the transforms for the surface revert to `direct1` method instead of `direct2`:

```python
from qsc import Qsc

name =  'precise QA'
qsc = Qsc.from_paper(name)
r=0.1

# works
desc_eq_1 = Equilibrium.from_near_axis(qsc,
                                         r= r,
                                         L=6,
                                         M=6,
                                         N=19,
                                         ntheta=45)
# increment N by 1, throws warning about reverting to direct1
desc_eq_1 = Equilibrium.from_near_axis(qsc,
                                         r= r,
                                         L=6,
                                         M=6,
                                         N=20,
                                         ntheta=45)
```

warning being: 
```python
/home/dpanici/DESC/desc/transform.py:319: UserWarning: direct2 method requires that node pattern is the same on each zeta plane, falling back to direct1 method```

### PR DESCRIPTION
**Pull request by dpanici**
…r large arrays
Resolves #358

### CODE REVIEW

**Comment by f0uriest** [comment]
Would a better solution be to avoid use of `std` altogether? IE, for `isalmostequal` we could do `np.allclose(x, x[0])` and for `islinspaced` could do `np.allclose(np.diff(x), np.diff(x)[0])`

File: tests/test_utils.py [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+    assert isalmostequal(
+        grid_large.nodes[:, :2].T.reshape((2, zeta_cts[0], -1), order="F")
+    )
+
+
+@pytest.mark.unit
+def test_islinspaced():
+    """Test that islinspaced function works on large arrays."""
+    grid_small = LinearGrid(rho=1, M=1, N=10)
+    zeta_vals = np.unique(grid_small.nodes[:, 2])
```
ddudt, line 30: can use `zeta_vals = grid.nodes[grid.unique_zeta_idx, 2]`

ddudt (changes requested):
I agree with Rory that the better solution would be to avoid calling `np.std` altogether.

dpanici:
addressed the comments. Took me way too long to get a good solution for using `np.allclose` for when the axis may be something other than -1... numpy indexing and broadcasting can be confusing

desc/utils.py
```diff
@@ -271,7 +271,7 @@ def issorted(x, axis=None, tol=1e-12):
     return np.all(np.diff(x, axis=axis) >= -tol)
 
 
-def isalmostequal(x, axis=-1, tol=1e-12):
+def isalmostequal(x, axis=-1, rtol=1e-5, atol=1e-8):
```
f0uriest, line 274: should probably tighten these tolerances, otherwise seems fine

f0uriest (approved):

f0uriest (approved):

f0uriest (dismissed):

tests/test_utils.py
```diff
@@ -9,27 +9,47 @@
...
+    zeta_vals = grid_large.nodes[grid_large.unique_zeta_idx, 2]
     assert islinspaced(zeta_vals)
+
+    # on a 2D array
+    zz = np.tile(zeta_vals, (2, 1))
+    zz[1, :] *= 2
+    assert islinspaced(zz)
+
+    # 0D arrays will return True
+    assert isalmostequal(np.array(0))
```
ddudt, line 55: this should be `assert islinspaced(np.array(0))`

desc/utils.py
```diff
@@ -271,7 +271,7 @@ def issorted(x, axis=None, tol=1e-12):
     return np.all(np.diff(x, axis=axis) >= -tol)
 
 
-def isalmostequal(x, axis=-1, tol=1e-12):
+def isalmostequal(x, axis=-1, rtol=1e-6, atol=1e-12):
```
ddudt, line 274: need to add `rtol` and `atol` to documentation

ddudt (changes requested):
This solution seems overly complicated, but it works.

dpanici:
> This solution seems overly complicated, but it works.

yea without np.std I could not figure out how to get the same functionality as cleanly using np.allclose (namely the ability to have the axis of comparison be arbitrary)

f0uriest (approved):

ddudt (approved):

PR #362, opened 2022-12-19
Refactor compute functions

### LINKED ISSUE #270: More efficient compute functions
Right now we use the ``check_derivs`` function to determine what things to compute within a given compute function. As we add more things, this can get slower and slower because we compute extra things that may not be needed and forward mode autodiff will still keep track of them. I've been thinking about a new method, outlined below.

The main idea is that each compute function would only compute a single quantity from the data index. Most of them could be simple one liner lambda functions.
Each would take 3 args: 
``data`` : a dictionary of real space things computed up to this point
``coeffs``: dictionary of spectral coefficients
``transforms``: dictionary of transform and profile objects
(maybe also something for kwargs etc)

The data index would be updated to include a list of what things that quantity directly depends on, like so:
```
data_index["lambda_z"] = {
    "label": "\\partial_{\\zeta} \\lambda",
    "units": "rad",
    "units_long": "radians",
    "description": "Poloidal stream function, first toroidal derivative",
    "fun": lambda data, coeffs, transforms: transforms['L'].transform(coeffs['L_lmn'], dz=1),
    "depends_on": []
    "dim": 1,
    "L_derivs": [[0, 0, 1]],
}
data_index["B^theta"] = {
    "label": "B^{\\theta}",
    "units": "T \\cdot m^{-1}",
    "units_long": "Tesla / meters",
    "description": "Contravariant poloidal component of magnetic field",
    "fun": lambda data, coeffs, transforms: data["B0"] * (data["iota"] - data["lambda_z"])
    "depends_on" : ["B0", "iota", "lambda_z"],
    "dim": 1,
    "R_derivs": [[0, 0, 0], [1, 0, 0], [0, 1, 0], [0, 0, 1]],
    "L_derivs": [[0, 0, 0], [0, 1, 0], [0, 0, 1]],
}
```

Note that ``lambda_z`` doesn't depend on anything, since its directly computed from the spectral coefficients. The L_derivs etc still work the same to figure out what orders of derivatives are needed (could possibly rework it so the list of derivs is also created recursively)

All this would be wrapped by one main compute funciton like so:

```
def compute(qty, data, coeffs, transforms):
	if qty in data:
		return data
	for thing in data_index[qty]['depends_on']:
		if thing not in data:
                        # compute dependencies recursively
			data = compute(thing, data, coeffs, transforms)
	data[qty] = data_index[qty]["fun"](data, coeffs, transforms)
	return data
```

This would eliminate the possibility of computing more things than needed, and ideally make adding new computeable things a lot easier.

Steps:

- [x] #331
- [x] #332
- [ ] #356
- [ ] #357

### LINKED ISSUE #356: Test/profile recursive compute function

### LINKED ISSUE #357: Break up compute functions into one item each

f0uriest:
This PR breaks all the `compute_` functions up into small functions that each only compute 1 thing, and uses the dependencies in the `data_index` plus the main `compute` function to recursively compute all the required quantities.

- Breaks compute stuff into more submodules to make organization a bit easier
- Adds new helper `register_compute_fun` for decorating a function to add it to the `data_index`
- Metadata formerly in the `data_index` file is now spread across all the files, next to the function that computes the thing. The `data_index` is built at runtime.
- All objective functions now use the standard `compute` function.

There are a lot of changes so would appreciate as many eyes as possible. Most of its just cut/paste/refactor stuff, and all the tests pass, but we should especially check stuff that doesn't get tested like `label`, `description`, `units` etc.

Resolves #270 
Resolves #356 
Resolves #357


desc/compute/_basis_vectors.py
```diff
@@ -0,0 +1,577 @@
...
+)
+def _e_sub_zeta(params, transforms, profiles, data, **kwargs):
+    data["e_zeta"] = jnp.array([data["R_z"], data["R"], data["Z_z"]]).T
+    return data
+
+
+@register_compute_fun(
+    name="e_rho_r",
+    label="\\partial_{\\rho} \\mathbf{e}_{\\rho}",
+    units="m",
```
unalmis, line 61: These are consistent with what they were before. Do we intentionally keep it as meters for simplicity?  If not the rho derivatives should introduce some unit cancellations.

f0uriest:
change compute to take str or list of str

unalmis (approved):

unalmis (approved):

desc/compute/_equil.py
```diff
@@ -4,118 +4,423 @@
...
+        + data["J^theta"] * data["e_theta"].T
+        + data["J^zeta"] * data["e_zeta"].T
+    ).T
+    return data
+
+
+@register_compute_fun(
+    name="J_R",
+    label="J_{R}",
+    units="A \\cdot m^{-3}",
```
dpanici, line 90: I am pretty sure this should be A / m^2, the contravariant components J^rho etc yes are A/m^3, but when you go to lab frame you multiple J^rho by e_rho etc which then gives A/m^3. Besides, physically current density is A/m^2

desc/compute/_equil.py
```diff
@@ -4,118 +4,423 @@
...
+)
+def _J_R(params, transforms, profiles, data, **kwargs):
+    data["J_R"] = data["J"][:, 0]
+    return data
+
+
+@register_compute_fun(
+    name="J_phi",
+    label="J_{\\phi}",
+    units="A \\cdot m^{-3}",
```
dpanici, line 108: same comment as above

desc/compute/_equil.py
```diff
@@ -4,118 +4,423 @@
...
+)
+def _J_phi(params, transforms, profiles, data, **kwargs):
+    data["J_phi"] = data["J"][:, 1]
+    return data
+
+
+@register_compute_fun(
+    name="J_Z",
+    label="J_{Z}",
+    units="A \\cdot m^{-3}",
```
dpanici, line 126: same comment as above

desc/compute/_equil.py
```diff
@@ -4,118 +4,423 @@
...
+)
+def _J_Z(params, transforms, profiles, data, **kwargs):
+    data["J_Z"] = data["J"][:, 2]
+    return data
+
+
+@register_compute_fun(
+    name="|J|",
+    label="|\\mathbf{J}|",
+    units="A \\cdot m^{-3}",
```
dpanici, line 144: same comment as above

desc/compute/_equil.py
```diff
@@ -4,118 +4,423 @@
...
+)
+def _J_dot_B(params, transforms, profiles, data, **kwargs):
+    data["J*B"] = dot(data["J"], data["B"])
+    return data
+
+
+@register_compute_fun(
+    name="J_parallel",
+    label="\\mathbf{J} \\cdot \\mathbf{b}",
+    units="A \\cdot m^{-3}",
```
dpanici, line 180: same comment as above

desc/compute/_equil.py
```diff
@@ -4,118 +4,423 @@
...
+    data["|J|"] = jnp.linalg.norm(data["J"], axis=-1)
+    return data
+
+
+@register_compute_fun(
+    name="J*B",
+    label="\\mathbf{J} \\cdot \\mathbf{B}",
+    units="N / m^{3}",
+    units_long="Newtons / cubic meter",
+    description="Bootstrap current",
```
dpanici, line 164: Maybe worth mentioning this is NOT in units of current density in the description?

desc/compute/_profiles.py
```diff
@@ -0,0 +1,451 @@
...
+    data=["rho"],
+)
+def _psi(params, transforms, profiles, data, **kwargs):
+    data["psi"] = params["Psi"] * data["rho"] ** 2 / (2 * jnp.pi)
+    return data
+
+
+@register_compute_fun(
+    name="psi_r",
+    label="\\partial_{\\rho} \\psi' = \\partial_{\\rho} \\Psi / (2 \\pi)",
```
dpanici, line 29: you have a psi' here, should be just psi

desc/compute/_profiles.py
```diff
@@ -0,0 +1,451 @@
...
+    data=["rho"],
+)
+def _psi_r(params, transforms, profiles, data, **kwargs):
+    data["psi_r"] = 2 * params["Psi"] * data["rho"] / (2 * jnp.pi)
+    return data
+
+
+@register_compute_fun(
+    name="psi_rr",
+    label="\\partial_{\\rho\\rho} \\psi' = \\partial_{\\rho\\rho} \\Psi / (2 \\pi)",
```
dpanici, line 47: you have a psi' here, should be just psi

desc/compute/_profiles.py
```diff
@@ -0,0 +1,451 @@
...
+    )
+    return data
+
+
+@register_compute_fun(
+    name="iota",
+    label="\\iota",
+    units="~",
+    units_long="None",
+    description="Rotational transform",
```
dpanici, line 235: maybe worth putting here that this is technically iotabar (iota normalized by 2pi), this is in our docs pages but would be useful to have in this description too to be clear

desc/compute/_profiles.py
```diff
@@ -0,0 +1,451 @@
...
+    data["G_r"] = surface_averages(transforms["grid"], data["B_zeta_r"])
+    return data
+
+
+@register_compute_fun(
+    name="current",
+    label="\\frac{2\\pi}{\\mu_0} I",
+    units="A",
+    units_long="Amperes",
+    description="Net toroidal current",
```
dpanici, line 423: could add "enclosed by flux surfaces" to be clear this is not just total net toroidal current

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _Y(params, transforms, profiles, data, **kwargs):
+    data["Y"] = data["R"] * jnp.sin(data["phi"])
+    return data
+
+
+@register_compute_fun(
+    name="lambda",
+    label="\\lambda",
+    units="m",
```
dpanici, line 878: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda(params, transforms, profiles, data, **kwargs):
+    data["lambda"] = transforms["L"].transform(params["L_lmn"], 0, 0, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_r",
+    label="\\partial_{\\rho} \\lambda",
+    units="m",
```
dpanici, line 896: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_r(params, transforms, profiles, data, **kwargs):
+    data["lambda_r"] = transforms["L"].transform(params["L_lmn"], 1, 0, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_t",
+    label="\\partial_{\\theta} \\lambda",
+    units="m",
```
dpanici, line 914: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_t(params, transforms, profiles, data, **kwargs):
+    data["lambda_t"] = transforms["L"].transform(params["L_lmn"], 0, 1, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_z",
+    label="\\partial_{\\zeta} \\lambda",
+    units="m",
```
dpanici, line 932: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_z(params, transforms, profiles, data, **kwargs):
+    data["lambda_z"] = transforms["L"].transform(params["L_lmn"], 0, 0, 1)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_rr",
+    label="\\partial_{\\rho \\rho} \\lambda",
+    units="m",
```
dpanici, line 950: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_rr(params, transforms, profiles, data, **kwargs):
+    data["lambda_rr"] = transforms["L"].transform(params["L_lmn"], 2, 0, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_tt",
+    label="\\partial_{\\theta \\theta} \\lambda",
+    units="m",
```
dpanici, line 968: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_tt(params, transforms, profiles, data, **kwargs):
+    data["lambda_tt"] = transforms["L"].transform(params["L_lmn"], 0, 2, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_zz",
+    label="\\partial_{\\zeta \\zeta} \\lambda",
+    units="m",
```
dpanici, line 986: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_zz(params, transforms, profiles, data, **kwargs):
+    data["lambda_zz"] = transforms["L"].transform(params["L_lmn"], 0, 0, 2)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_rt",
+    label="\\partial_{\\rho \\theta} \\lambda",
+    units="m",
```
dpanici, line 1004: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_rt(params, transforms, profiles, data, **kwargs):
+    data["lambda_rt"] = transforms["L"].transform(params["L_lmn"], 1, 1, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_rz",
+    label="\\partial_{\\rho \\zeta} \\lambda",
+    units="m",
```
dpanici, line 1023: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_rz(params, transforms, profiles, data, **kwargs):
+    data["lambda_rz"] = transforms["L"].transform(params["L_lmn"], 1, 0, 1)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_tz",
+    label="\\partial_{\\theta \\zeta} \\lambda",
+    units="m",
```
dpanici, line 1042: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_tz(params, transforms, profiles, data, **kwargs):
+    data["lambda_tz"] = transforms["L"].transform(params["L_lmn"], 0, 1, 1)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_rrr",
+    label="\\partial_{\rho \\rho \\rho} \\lambda",
+    units="m",
```
dpanici, line 1061: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_rrr(params, transforms, profiles, data, **kwargs):
+    data["lambda_rrr"] = transforms["L"].transform(params["L_lmn"], 3, 0, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_ttt",
+    label="\\partial_{\\theta \\theta \\theta} \\lambda",
+    units="m",
```
dpanici, line 1079: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_ttt(params, transforms, profiles, data, **kwargs):
+    data["lambda_ttt"] = transforms["L"].transform(params["L_lmn"], 0, 3, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_zzz",
+    label="\\partial_{\\zeta \\zeta \\zeta} \\lambda",
+    units="m",
```
dpanici, line 1097: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_zzz(params, transforms, profiles, data, **kwargs):
+    data["lambda_zzz"] = transforms["L"].transform(params["L_lmn"], 0, 0, 3)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_rrt",
+    label="\\partial_{\\rho \\rho \\theta} \\lambda",
+    units="m",
```
dpanici, line 1115: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_rrt(params, transforms, profiles, data, **kwargs):
+    data["lambda_rrt"] = transforms["L"].transform(params["L_lmn"], 2, 1, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_rtt",
+    label="\\partial_{\\rho \\theta \\theta} \\lambda",
+    units="m",
```
dpanici, line 1134: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_rtt(params, transforms, profiles, data, **kwargs):
+    data["lambda_rtt"] = transforms["L"].transform(params["L_lmn"], 1, 2, 0)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_rrz",
+    label="\\partial_{\\rho \\rho \\zeta} \\lambda",
+    units="m",
```
dpanici, line 1153: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_rrz(params, transforms, profiles, data, **kwargs):
+    data["lambda_rrz"] = transforms["L"].transform(params["L_lmn"], 2, 0, 1)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_rzz",
+    label="\\partial_{\\rho \\zeta \\zeta} \\lambda",
+    units="m",
```
dpanici, line 1172: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_rzz(params, transforms, profiles, data, **kwargs):
+    data["lambda_rzz"] = transforms["L"].transform(params["L_lmn"], 1, 0, 2)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_ttz",
+    label="\\partial_{\\theta \\theta \\zeta} \\lambda",
+    units="m",
```
dpanici, line 1191: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
+)
+def _lambda_ttz(params, transforms, profiles, data, **kwargs):
+    data["lambda_ttz"] = transforms["L"].transform(params["L_lmn"], 0, 2, 1)
+    return data
+
+
+@register_compute_fun(
+    name="lambda_tzz",
+    label="\\partial_{\\theta \\zeta \\zeta} \\lambda",
+    units="m",
```
dpanici, line 1210: this should be in units of radians, not meters

desc/compute/_core.py
```diff
@@ -1,654 +1,1242 @@
...
-        data["R0/a"] = data["R0"] / data["a"]
-    if has_dependencies("V_rr(r)", params, transforms, profiles, data):
-        data["V_rr(r)"] = surface_integrals(
-            transforms["grid"], data["sqrt(g)_r"] * jnp.sign(data["sqrt(g)"])
-        )
 
+@register_compute_fun(
+    name="lambda_rtz",
+    label="\\partial_{\\rho \\theta \\zeta} \\lambda",
+    units="m",
```
dpanici, line 1229: this should be in units of radians, not meters

dpanici (approved):

desc/compute/_field.py
```diff
@@ -4,800 +4,1858 @@
...
-def compute_magnetic_tension(params, transforms, profiles, data=None, **kwargs):
-    """Compute magnetic tension."""
-    data = compute_contravariant_current_density(
-        params,
-        transforms,
-        profiles,
-        data=data,
-        **kwargs,
+@register_compute_fun(
+    name="<|B|>",
```
unalmis, line 1378: The flux surface average definitions I have read in papers (and D'haeseleer's flux coords book) are usually equivalent to
$$\langle Q \rangle\ = \dfrac{\iint d\theta\ d\zeta\ \sqrt{g}\ Q\}{\iint d\theta\ d\zeta\  \sqrt{g}\}$$

The denominator is the derivative of signed volume wrt rho. The jacobean `sqrt(g)` is not enforced to be positive.

For the magnetic well, we take volume to be considered always positive. With this convention the flux surface average formula becomes
$$\langle Q \rangle\ = \dfrac{\iint d\theta\ d\zeta\ \lvert \sqrt{g} \rvert\ Q\}{\iint d\theta\ d\zeta\  \lvert \sqrt{g} \rvert\}$$

1. Since we are moving  `<B^2>` and `<B^2>_r` from the magnetic well compute function into the `data_index`, maybe we should add something like 'uses magnitude of jacobean' to the description?
2. `<|B|>` was not used in the magnetic well compute function. Is `<|B|>` planned to be used in a context with the magnitude of jacobean in the average, or do we want that for consistency?

unalmis (dismissed):

desc/compute/_qs.py
```diff
@@ -2,131 +2,296 @@
...
+    data["zeta_B"] = data["zeta"] + data["nu"]
+    return data
+
+
+@register_compute_fun(
+    name="sqrt(g)_B",
+    label="\\sqrt{g}_{B}",
+    units="~",
+    units_long="None",
+    description="Jacobian determinant of Boozer coordinates",
```
ddudt, line 177: Similar comment to a previous one about the Jacobians. I suggest the description: 
`"Jacobian of the Boozer coordinate system transformation"`

desc/compute/_qs.py
```diff
@@ -2,131 +2,296 @@
...
+    label="Boozer modes",
+    units="~",
+    units_long="None",
+    description="Boozer harmonics",
+    dim=1,
+    params=[],
+    transforms={"B": [[0, 0, 0]]},
+    profiles=[],
+    function_of="rtz",
+    data=["sqrt(g)_B", "|B|", "rho", "theta_B", "zeta_B"],
```
ddudt, line 229: Does it actually need this data? I don't think it should have any data dependencies.

desc/compute/_qs.py
```diff
@@ -2,131 +2,296 @@
...
+        "|B|_z",
+        "G",
+        "I",
+        "B*grad(|B|)",
+    ],
+    helicity="helicity",
+)
+def _f_C(params, transforms, profiles, data, **kwargs):
     M = kwargs.get("helicity", (1, 0))[0]
     N = kwargs.get("helicity", (1, 0))[1]
```
ddudt, line 264: Suggested change:
```
(M, N) = kwargs.get("helicity", (1, 0))
```

desc/compute/_qs.py
```diff
@@ -2,131 +2,296 @@
...
+    function_of="rtz",
+    data=["sqrt(g)_B", "|B|", "rho", "theta_B", "zeta_B"],
+)
+def _B_modes(params, transforms, profiles, data, **kwargs):
+    data["B modes"] = transforms["B"].basis.modes
+    return data
 
+
+@register_compute_fun(
+    name="f_C",
```
ddudt, line 237: Do we like the names `f_C` and `f_T` for these quasi-symmetry quantities? This is what I called them in my paper, but it's probably not the best practice to have every new quantity named `f_X`. I feel like `QS_C` or `QS_two-term` would be better. But also annoying for users if we change it on them.

ddudt (changes requested):
Only minor comments. I mentioned some things that might not be necessary to change -- sorry if I was being too picky. 

(This took me hours to review!)

f0uriest:
@ddudt thanks for the (many) comments. It sounds like most of them are things we should address but are mostly things that were issues before this PR and were basically just copied from the data index. In order to not have this PR get more bloated I propose merging this in and then starting another one specifically for cleaning up the names/labels/simplify algebra etc. 
I had planned to do something like that, as when refactoring things I noticed several places we can simplify code and reduce dependencies etc, but wanted to keep this as mostly just reorganizing things without getting too into the weeds on content.
Thoughts?

desc/compute/_equil.py
```diff
@@ -4,118 +4,441 @@
...
+    description="Divergence of plasma current density perpendicular to magnetic field",
+    dim=1,
+    params=[],
+    transforms={},
+    profiles=[],
+    function_of="rtz",
+    data=["J^rho", "p_r", "|B|"],
+)
+def _div_J_perp(params, transforms, profiles, data, **kwargs):
+    data["div(J_perp)"] = (mu_0 * data["J^rho"] * data["p_r"]) / data["|B|"] ** 2
```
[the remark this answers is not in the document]
f0uriest: not sure, I think @dpanici implemented this originally?

desc/compute/_field.py
f0uriest: would it matter, since the sign of the jacobian would cancel anyways?

unalmis: If the jacobean sign is constant throughout the whole vector you're right that it won't matter. Otherwise they could give different results

desc/compute/_basis_vectors.py
```diff
@@ -0,0 +1,610 @@
...
+    name="e_rho",
+    label="\\mathbf{e}_{\\rho}",
+    units="m",
+    units_long="meters",
+    description="Covariant radial basis vector",
+    dim=3,
+    params=[],
+    transforms={},
+    profiles=[],
+    function_of="rtz",
```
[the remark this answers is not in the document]
unalmis: I like coordinates more than variables. For string vs list the differences are in the below use cases
```
if coordinates == ["r"]: # list
if coordinates == "r":   # string
```
```
if "r" in coordinates and "t" in coordinates and "z" not in coordinates: # either
if coordinates == ["r", "t"]: # list
if coordinates ==  "rt": # string
```

unalmis (approved):
just adding back previous approval after pushing commit to resolve 2 comments

ddudt (approved):
@f0uriest OK, I'm approving this with the understanding that my minor comments will be addressed in future PRs.

desc/compute/_field.py
```diff
@@ -4,800 +4,1858 @@
...
+    data["grad(|B|^2)"] = (
+        data["grad(|B|^2)_rho"] * data["e^rho"].T
+        + data["grad(|B|^2)_theta"] * data["e^theta"].T
+        + data["grad(|B|^2)_zeta"] * data["e^zeta"].T
+    ).T
+    return data
+
+
+@register_compute_fun(
+    name="|grad(|B|^2)|/2mu0",
```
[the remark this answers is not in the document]
f0uriest: I'd vote for leaving it as is unless there is a strong reason to change

desc/compute/_equil.py
```diff
@@ -4,118 +4,441 @@
...
+    units="J",
+    units_long="Joules",
+    description="Plasma thermodynamic energy",
+    dim=0,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    function_of="",
+    data=["p", "sqrt(g)"],
+    gamma="gamma",
```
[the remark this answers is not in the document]
f0uriest: `gamma` is the adiabatic index, this is how you register that the compute function takes an additional keyword argument.

desc/compute/_geometry.py
```diff
@@ -0,0 +1,354 @@
...
+    L = data["L_sff"]
+    M = data["M_sff"]
+    N = data["N_sff"]
+    a = E * G - F**2
+    b = F * M - L * G - E * N
+    c = L * N - M**2
+    r1 = (-b + jnp.sqrt(b**2 - 4 * a * c)) / (2 * a)
+    r2 = (-b - jnp.sqrt(b**2 - 4 * a * c)) / (2 * a)
+    data["curvature_k1"] = jnp.maximum(r1, r2)
+    data["curvature_k2"] = jnp.minimum(r1, r2)
```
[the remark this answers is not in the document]
f0uriest: because the calculation of k1 and k2 are nearly identical (and I didn't really want to add more intermediate quantities to the data index), I thought it would be best to always return both from either function, so that in theory only one of them ever needs to be called.

desc/compute/_metric.py
```diff
@@ -0,0 +1,605 @@
...
+def _e_theta_x_e_zeta(params, transforms, profiles, data, **kwargs):
+    data["|e_theta x e_zeta|"] = jnp.linalg.norm(
+        cross(data["e_theta"], data["e_zeta"]), axis=1
+    )
+    return data
+
+
+@register_compute_fun(
+    name="|e_zeta x e_rho|",
+    label="|e_{\\zeta} \\times e_{\\rho}|",
```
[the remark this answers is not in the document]
f0uriest: I'd prefer to keep it as is. e_zeta x e_rho is the outward normal (assuming positive jacobian), which is what we generally want and is consistent with the others which are also outward normals. Admittedly it doesn't matter since its just the norm, but i think it makes sense if you think of rho -> theta -> zeta as being cyclical

desc/compute/_qs.py
f0uriest: I'd vote for leaving it as is for consistency. But yeah for future stuff we should come up with something besides `f_X`

desc/compute/_basis_vectors.py
```diff
@@ -0,0 +1,610 @@
...
+def _e_sub_zeta_tz(params, transforms, profiles, data, **kwargs):
+    data["e_zeta_tz"] = jnp.array([data["R_tzz"], data["R_tz"], data["Z_tzz"]]).T
+    return data
+
+
+@register_compute_fun(
+    name="e^rho",
+    label="\\mathbf{e}^{\\rho}",
+    units="m^{-1}",
+    units_long="inverse meters",
```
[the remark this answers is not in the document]
f0uriest: I think we use `inverse meters` or `inverse square meters` in most places

desc/compute/_field.py
```diff
@@ -4,800 +4,1858 @@
...
+            + data["B^zeta"] * dot(data["e_zeta_t"], data["e_zeta"])
+        )
+        + data["B^theta"]
+        * data["B^zeta"]
+        * (
+            dot(data["e_theta_t"], data["e_zeta"])
+            + dot(data["e_zeta_t"], data["e_theta"])
+        )
+    ) / data["|B|"]
+    return data
```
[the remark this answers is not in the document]
f0uriest: I assume this is what #384 is about?

desc/compute/_metric.py
```diff
@@ -0,0 +1,605 @@
...
+from .data_index import register_compute_fun
+from .utils import cross, dot
+
+
+@register_compute_fun(
+    name="sqrt(g)",
+    label="\\sqrt{g}",
+    units="m^{3}",
+    units_long="cubic meters",
+    description="Jacobian determinant",
```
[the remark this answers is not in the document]
f0uriest: I don't think so. It's the square root of the determinant of the metric tensor g_ij, but thats the same as the regular determinant of the coordinate transformation Jacobian `du_i/dx_j`. In order to distinguish it from the Jacobian matrix or the metric tensor itself I'd vote for `Jacobian determinant of flux coordinate system`

desc/compute/_qs.py
```diff
@@ -2,131 +2,296 @@
...
+    dim=1,
+    params=[],
+    transforms={"w": [[0, 0, 0]]},
+    profiles=[],
+    function_of="rtz",
+    data=["w_mn", "G", "I", "iota", "lambda"],
+)
+def _nu(params, transforms, profiles, data, **kwargs):
+    GI = data["G"] + data["iota"] * data["I"]
+    w = transforms["w"].transform(data["w_mn"])
```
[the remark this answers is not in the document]
f0uriest: is `w` ever needed by itself? Also, is there a standard name for it? Or is it really just defined as some intermediate quantity in Hirshman's paper?

desc/compute/_field.py
f0uriest: Well if the surfaces are nested it should have a constant sign throughout the volume, so it shouldn't be an issue. And once #202 is merged it should be fine also.

PR #363, opened 2022-12-19
Add Incompatible Constraint Check and change normalization for B_p to avoid having a 0 normalization factor

dpanici:
- Changes Normalization for B_P to check if iota is zero, and if so makes the scale factor the same as B_T

- Also add incompatible constraint check (Rory's implementation from #306 ) to see if xp satisfies the constraints (cherry-picked from #306  which will be merged soon but this provides a test for why I needed to change the normalization)

This is because if iota=0 (like for example in a newly-made object `Equilibrium(L=3,M=3,N=3)` where by default the current profile is 0 and thus the iota profile is zero, the normalization for B_P will be zero, which then causes the normalizaiton for I to be zero

This causes issues if you try to check feasibility with the particular solution for the initial solve, as if you check the constraint for I, the normalization is 0 so it returns `NaN` for `constraint.compute(xp['c_l'])`


dpanici:
-[ ] loop thru, set to1 if 0

desc/objectives/normalization.py
```diff
@@ -26,4 +26,9 @@ def compute_scaling_factors(eq):
     scales["F"] = scales["p"] / scales["a"]
     scales["f"] = scales["F"] * scales["V"]
     scales["Psi"] = abs(eq.Psi)
+    # replace 0 scales to avoid normalizing by zero
```
f0uriest, line 29: this is good, but also for the particular case of `iota=0` I think the thing you had for `B_P` also makes sense, so that at least `I` will be order 1/mu0 instead of just 1

desc/objectives/utils.py
```diff
@@ -204,6 +204,18 @@ def recover(x_reduced):
         dx = put(jnp.zeros(dim_x), unfixed_idx, Z @ x_reduced)
         return jnp.atleast_1d(jnp.squeeze(xp + dx))
 
+    # check that all constraints are actually satisfiable
```
f0uriest, line 207: We should add a test for this if possible.

desc/objectives/linear_objectives.py
```diff
@@ -160,7 +160,10 @@ def build(self, eq, use_jit=True, verbose=1):
 
     def compute(self, *args, **kwargs):
         """Compute deviation from desired boundary."""
-        x = kwargs.get(self.args[0], args[0])
+        if len(args):
+            x = kwargs.get(self.args[0], args[0])
+        else:
+            x = kwargs.get(self.args[0])
```
ddudt, line 166: Why would there ever be more than one argument? `FixBoundaryR` and `FixBoundaryZ` should take only `R_lmn` or `Z_lmn`.

dpanici: I don't quite remember the issue that prompted this, @f0uriest do you remember? Something to do with R_lmn and Rb_lmn I think

The direct issue I think was that if the passed-in just one arg, args[0] throws an indexing error, and this if statement prevents that.

f0uriest: The issue isn't more than 1 arg, its 0 args. IE, if you just pass things as kwargs by unpacking a dict. This will make sure that it works whether you pass it as a positional or keyword argument.

f0uriest (approved):

dpanici:
only iota=0 part of the `test_force_balance_grids` test is failing, so related to the normalization when iota=0

f0uriest (approved):

ddudt (approved):

Issue #360, opened 2022-12-19
Make Stability Objectives Print Compute un-normed objectives


dpanici:

dpanici:
#364

PR #364, opened 2022-12-19
Have stability objectives print their max, min and average value at end of solve

dpanici:
Stability objective normed values are meaningless when what really matters is the sign, so this makes the print statement for the end-of-solve change so that it prints the max, min and average values for MagneticWell and MercierStability

addresses #360


f0uriest (dismissed):

dpanici:
pytest output

ddudt (approved):

f0uriest (approved):

unalmis (approved):

PR #367, opened 2022-12-20
add return_data flag to most plotting fxns

### LINKED ISSUE #361: Return Data computed when using a plotting function

dpanici:
- passing `return_data=True` to a plotting function will have it return a `plot_data` dict with the values computed and used to plot, in addition to the usual `fig,ax`
Resolves #361


desc/plotting.py
```diff
@@ -1511,6 +1647,10 @@ def plot_boundaries(eqs, labels=None, zeta=None, ax=None, **kwargs):
         Z = coords["Z"].reshape(
             (grid.num_theta, grid.num_rho, grid.num_zeta), order="F"
         )
+
+        plot_data[f"R_{i}"] = R
```
f0uriest, line 1651: would it make more sense to return a dict of lists that matches up with the list of inputs?

desc/plotting.py
```diff
@@ -1649,42 +1801,77 @@ def plot_comparison(
...
+                axis_marker="o",
+                axis_size=0,
+                label=labels[i % len(labels)],
+                title_font_size=title_font_size,
+                xlabel_fontsize=xlabel_fontsize,
+                ylabel_fontsize=ylabel_fontsize,
+                return_data=return_data,
+            )
+            for key in _plot_data.keys():
+                plot_data[key + "_" + labels[i % len(labels)]] = _plot_data[key]
```
f0uriest, line 1835: same as above, might be better to return dict of lists

desc/plotting.py
```diff
@@ -1889,17 +2104,29 @@ def plot_boozer_modes(
                 linestyle=linestyle,
                 linewidth=linewidth,
             )
+        plot_data[f"|B|_{M}{N}"] = B_mn[:, i]
```
f0uriest, line 2107: probably cleaner to return `B_mn` as an array and then a separate array for `B modes`

desc/plotting.py
```diff
@@ -2199,7 +2456,9 @@ def plot_grid(grid, **kwargs):
     Parameters
     ----------
     grid : Grid
-        Grid to plot.
+        Grid to plot.    return_data : bool
```
f0uriest, line 2459: missing newline

desc/plotting.py
```diff
@@ -2295,13 +2555,16 @@ def plot_basis(basis, **kwargs):
...
 
+
     Returns
     -------
     fig : matplotlib.figure.Figure
         Figure being plotted to.
     ax : matplotlib.axes.Axes, ndarray of axes, or dict of axes
         Axes used for plotting. A single axis is used for 1d basis functions,
-        2d or 3d bases return an ndarray or dict of axes.
+        2d or 3d bases return an ndarray or dict of axes.    return_data : bool
```
f0uriest, line 2565: missing newline

desc/plotting.py
f0uriest: also it looks like this doesn't have the return_data option

desc/plotting.py
f0uriest: also looks like return_data option isnt implemented

desc/plotting.py
```diff
@@ -2749,6 +3020,8 @@ def plot_field_lines_sfl(
         spacing in phi to sample field lines along, in radians. Default is 1e-2.
     ax : matplotlib AxesSubplot, optional
         Axis to plot on.
+        if True, return the data plotted as well as fig,ax
```
f0uriest, line 3023: missing argument?

desc/plotting.py
```diff
@@ -3061,7 +3336,12 @@ def plot_field_lines_real_space(
...
     if return_B_interp:
-        return fig, ax, field_line_coords, B_interp
+        return (
+            fig,
+            ax,
+            field_line_coords,
+            B_interp,
+        )
     else:
         return fig, ax, field_line_coords
```
f0uriest, line 3346: might make sense to not return field line coords here, to be consistent with the others

f0uriest (changes requested):
Few comments above. Main thing is using lists instead of more keys in the dict and a few typos

dpanici:
Going to not touch the field_lines_real_space function, we don't test it and I think it needs to be refactored anyways and maybe field line traciing should be moved to something more general than the plotting functions

desc/plotting.py
```diff
@@ -2190,16 +2461,22 @@ def plot_qs_error(
...
 
-def plot_grid(grid, **kwargs):
+def plot_grid(grid, return_data=False, **kwargs):
     """Plot the location of collocation nodes on the zeta=0 plane.
 
     Parameters
     ----------
     grid : Grid
-        Grid to plot.
+        Grid to plot.    return_data : bool
```
ddudt, line 2476: typo?

ddudt (approved):
The returned data keys do not always match the keys from `data_index`. I understand why you couldn't do this, but it could get confusing. Would it be worth listing the keys it returns in each plotting function's documentation? Or maybe the users can figure it out for themselves.

ddudt (approved):

desc/plotting.py
```diff
@@ -1649,42 +1848,84 @@ def plot_comparison(
...
-            axis_color=colors[i % len(colors)],
-            axis_alpha=0,
-            axis_marker="o",
-            axis_size=0,
-            label=labels[i % len(labels)],
-            title_font_size=title_font_size,
-            xlabel_fontsize=xlabel_fontsize,
-            ylabel_fontsize=ylabel_fontsize,
-        )
+        if return_data:
```
f0uriest, line 1861: might be cleaner to always return data from plot_surfaces here and only return it from plot_comparison if asked (just to avoid big duplicated function call), but also ok how it is.

f0uriest (approved):

PR #376, opened 2022-12-22
improve plot_1d

### LINKED ISSUE #249: plot_1d improvement
I'd expect `desc.plotting.plot_1d` to plot any 1d quantity, not just those that can be evaluated on 1d linear grids.

I'm not aware of a way to plot 1d quantities computed on normal grids through DESC. `desc.plotting.plot_fsa` doesn't fill this need because it computes the flux-surface average of the given quantity with `sqrt(g)` when it's not always desired.

To plot `data["I"]`, the cleanest way (after `mercier` is merged to `master`) is:
```
def get_plot_data():
    # use the desc.compute.utils.compress function which is not in public docs
    x = compress(grid, eq.compute("rho", grid)["rho"])
    y = compress(grid, eq.compute("I", grid)["I"])
    return x, y
# make custom plots
```
We should add a function or modify `plot_1d` so that a user can do `plot_1d(eq, "I")` and get what they expect.

### LINKED ISSUE #368: Add warning or error when trying to Compute FSA Quantity with plot_1d
if you do `plot_1d(eq,name="magnetic well")`, the code does not complain and it plots fine, but the result is wrong bc it used a 1D grid to do a  FSA. Should have a warning thrown if this happens to use plot_fsa instead

unalmis:
Before this PR, `plot_1d` would generate misleading plots for single variable functions that needed to be evaluated on 2-d grids.

This PR is based off `rc/compute`. The only changes this PR introduces are in `plotting.py` and `test_plotting.py`.

Resolves #249 
Resolves #368


unalmis:
The comparisons below show this branch gives same results for plot_1d and plot_fsa like we want while master does not.

mwell_master
mwell_plot_1d
current_master
current_plot_1d

unalmis:
One thing this PR does is add an option to plot surface averages without the factor of the sqrt(g).  

Apparently surface averages without the `sqrt(g)` jacobean are called _theta averages_ in the literature. See equation 19 in https://aip.scitation.org/doi/pdf/10.1063/1.5098313. If that naming convention isn't standard, then it might be better to rename `plot_fsa` to `plot_surface_average` with a boolean to using sqrt(g) by default. If that is preferred I can push that commit here to update it.

desc/plotting.py
```diff
@@ -946,9 +971,156 @@ def plot_fsa(
...
+        M=M,
+        N=N,
+        norm_F=norm_F,
+        ax=ax,
+        return_data=return_data,
+        **kwargs,
+    )
+
+
+def plot_tsa(
```
f0uriest, line 990: What is the use case for this?

Also, from a naming perspective, a theta surface to me sounds like theta=const, rho,zeta vary. So in keeping with the idea of a `fsa` being a function of rho with theta/zeta averaged out, a `tsa` sounds like it should be a function of theta, with rho/zeta averaged out?

unalmis: Sometimes the surface average of interest of a quantity does not include the sqrt(g) factor. For example, B_theta is a quantity where the surface average without the sqrt(g) factor is the toroidal current. I don't know what a surface average of B_theta with the sqrt(g) factor represents.

The plotted example quantity for plot_fsa given in the docs is B_theta. It's likely that the intent of plotting that quantity should be done by function like plot_tsa instead, which does not use sqrt(g), so I added the plot_tsa function.

I agree that 'theta surface average' is poorly named. I called it that because the paper I linked made it seem like a known standard convention.

f0uriest: OK. I'd vote for removing `plot_tsa` and just adding an option to not include `sqrt(g)` in `plot_fsa`
As for renaming `plot_fsa` to `plot_surface_average` I'd lean towards not, just to keep a consistent API but I could go either way. If we do change it, we should keep `plot_fsa` as a wrapper to `plot_surface_averages` for now and have it throw a deprecation warning until we remove it fully in a future version (and in general we should start trying to make use of deprecation warnings rather than just suddenly changing API stuff where possible.)

f0uriest (approved):

PR #381, opened 2022-12-23
Fix issue where if passing modes in, the indices used to index the target do not match up with the indices of the A matrix

### LINKED ISSUE #380: Project/recover with `fixed_boundary=True` and specified modes gives wrong results
MWE:

```
n=1
eq = load(".//tests//inputs//precise_QH_step0.h5")[-1]


grid = LinearGrid(
    M=eq.M, N=eq.N, NFP=eq.NFP, rho=np.array([0.6, 0.8, 1.0]), sym=True
)

    (
        QuasisymmetryTwoTerm(weight=1e-2, helicity=(1, -eq.NFP), grid=grid),
        AspectRatio(target=8, weight=1e2),
    ),
    verbose=0,
)
R_modes = np.vstack(
    (
        [0, 0, 0],
        eq.surface.R_basis.modes[
            np.max(np.abs(eq.surface.R_basis.modes), 1) > n + 1, :
        ],
    )
)
Z_modes = eq.surface.Z_basis.modes[
    np.max(np.abs(eq.surface.Z_basis.modes), 1) > n + 1, :
]
constraints = (
    FixBoundaryR(modes=R_modes, fixed_boundary=True, normalize=False),
    FixBoundaryZ(modes=Z_modes, fixed_boundary=True, normalize=False),
)
for con in constraints:
    con.build(eq, verbose=0)
objective.build(eq)
from desc.objectives.utils import factorize_linear_constraints
xp, A, Ainv, b, Z, unfixed_idx, project, recover = factorize_linear_constraints(
    constraints,
    objective.args,
)

from scipy.linalg import block_diag
from desc.compute import arg_order
A_full = block_diag(*[A[arg] for arg in arg_order if arg in A.keys()])
b_full = jnp.concatenate([b[arg] for arg in arg_order if arg in b.keys()])

x1 = objective.x(eq)
x2 = recover(project(x1))
print(np.max(np.abs(x1-x2)))
print(np.max(np.abs(A_full @ xp - b_full)))
print(np.max(np.abs(A_full @ x1 - b_full)))
print(np.max(np.abs(A_full @ x2 - b_full)))
print(np.max(np.abs(A_full @ Z)))
```

Gives the following output:
```
0.3333333333333334
2.220446049250313e-16
1.0
2.220446049250313e-16
2.7755575615628914e-16
```

so doing project + recover gives a different answer, and the initial x from `objective.x` doesn't satisfy the constraints?

My hunch is its something to do with indexing for the targets and stuff but I can't quite figure it out.

dpanici:
- when a modes array is passed in, rearranges that modes array to match the order that the modes appear in eq.surface.R_lmn, so that the A matrix indices match up with the target when target[idx] is called
- also fixes conditional for applying weights only if non-default weights are passed in
- same fix for Z
Resolves #380


dpanici:
The test is maybe not great, if someone can think of a simpler test let me know/go ahead and make the change, I just used the one from the issue #380 but I think it should be testable at a smaller level

desc/objectives/linear_objectives.py
```diff
@@ -291,8 +303,11 @@ def build(self, eq, use_jit=True, verbose=1):
         # use given targets and weights if specified
         if self.target.size == modes.shape[0] and None not in self.target:
             self.target = self._target[modes_idx]
-        if self.weight.size == modes.shape[0] and self.weight != np.array(1):
+        if self.weight.size == modes.shape[0] and self.weight.size > 1:
```
f0uriest, line 306: I feel like we can get rid of these conditionals. We can assume that the user is either passing in targets and weights of the correct size, or target=None. I think another source of confusion is that in the parent class we explicitly broadcast target and weight to dim_f when I'm pretty sure we don't need to (and it would make it less confusing if we didn't)

The only thing we'd need to check is 
```
if target is None:
    target = eq.Rb_lmn[modes]
```

f0uriest:
OK, i think this works, the remaining question is that with this fix, the rows of A will be ordered in the same way as the basis modes array, which may be different from whatever order the user passed in. Do we want to change the order of A, or change the order of b to match what the user expects?

desc/objectives/objective_funs.py
```diff
@@ -565,24 +565,22 @@ def jit(self):
...
-            self._target = np.zeros(1)
-            self._weight = np.ones(1)
-        if np.unique(self.target).size == 1:
-            self._target = np.repeat(self.target[0], self.dim_f)
-        if np.unique(self.weight).size == 1:
-            self._weight = np.repeat(self.weight[0], self.dim_f)
-
-        if self.target.size != self.dim_f:
+
+        def _is_broadcastable(shp1, shp2):
```
ddudt, line 569: This could be worth moving to `desc/utils.py`

tests/test_linear_objectives.py
```diff
@@ -180,3 +185,152 @@ def test_build_init():
...
+
+
+@pytest.mark.unit
+def test_FixBoundary_with_single_weight():
+    """Test Fixing boundary with only a single, passed weight."""
+    eq = Equilibrium()
+    w = 1.1
+    FixZ = FixBoundaryZ(modes=np.array([[0, -1, 0]]), fixed_boundary=True, weight=w)
+    FixZ.build(eq)
+    print(FixZ._weight)
```
ddudt, line 330: rogue print statement

ddudt (commented):
I agree the tests are unnecessarily complex. I can try to find a more minimal example to use instead.

ddudt:
> OK, i think this works, the remaining question is that with this fix, the rows of A will be ordered in the same way as the basis modes array, which may be different from whatever order the user passed in. Do we want to change the order of A, or change the order of b to match what the user expects?

This only affects cases where the user explicitly passes in a custom `target`, right? In those cases, I vote for changing the order of `b` to match `A`, which will be the same order as the basis modes. In other words, re-order the `target` array the user passes based on what they passed for the `modes` array, so everything now matches the order of the sorted basis.

f0uriest:
OK, I'm fine with that. I added a note in the docstring to make users aware of the possible re-ordering.

f0uriest (dismissed):

dpanici:
- [ ] check by using a different example in test where target is passed
- [ ] maybe add a if self.target is not None, rerarrange the target in same way as the modes were (use idx)

desc/objectives/linear_objectives.py
```diff
@@ -51,6 +51,12 @@ class FixBoundaryR(_Objective):
     name : str
         Name of the objective function.
 
+
+    Notes
+    -----
+    If specifying particular modes to fix, the rows of the resulint constraint `A`
```
ddudt, line 57: Typo: "resulting" instead of "resulint"?

Same typo in `FixBoundaryZ` as well.

desc/objectives/linear_objectives.py
```diff
@@ -128,7 +134,17 @@ def build(self, eq, use_jit=True, verbose=1):
                         "yellow",
                     )
                 )
+
         self._dim_f = idx.size
+        if self.target is not None:  # rearrange given target to match modes order
+            if self._modes is True or self._modes is False:
+                raise RuntimeError(
```
ddudt, line 141: Making sure I understand why this error is necessary: we can't use the `target` values that the user passes in because they could be assuming a different order than the basis, right?

ddudt (approved):
Minor typo, otherwise looks good.

desc/objectives/linear_objectives.py
dpanici: basically yea, we wanna make sure that the target and modes arrays are corresponding to eachother, and so if a target is manually passed in you better be also passing in the modes array as a full array

f0uriest (approved):

ddudt (approved):

PR #382, opened 2022-12-28
Galerkin and Spectral Condensation

### LINKED ISSUE #324: Add objectives for Galerkin force error and spectral condensation
These would be good to have in the main code, both for testing and for users to try experimenting with. They just need to be cherry picked from the `constrained` branch. Ideally we can get those in first since they're fairly self contained before moving on to the rest of the constrained stuff.

pkim1818:
Adding objectives for galerkin projection of force error and spectral condensation.

Resolves #324


dpanici:
Can someone remind me what the issue was with the Galerkin projection? did it just not work well?

dpanici:
Been messing around with the branch for fun, it kind of works especially if you allow your projected basis resolution to exceed the equilibrium resolution, but something is broken with the SpectralCondensation, in that the factorize_linear_constraints acts strangely.

https://gist.github.com/dpanici/315967873d9a82c96a451729fd44aeba

YigitElma:
> Been messing around with the branch for fun, it kind of works especially if you allow your projected basis resolution to exceed the equilibrium resolution, but something is broken with the SpectralCondensation, in that the factorize_linear_constraints acts strangely.
> 
> https://gist.github.com/dpanici/315967873d9a82c96a451729fd44aeba
> 
> cool part for me is the cost consistently gets down to like 1e-12 range, this could be useful for deciding tolerances if we ever wanted to use this inside of proximal projection, for instance

Couple comments,
- For testing, it doesn't matter too much but I think if you just call `eq.set_initial_guess()`, this is not the actual initial guess we start with. Because it uses the axis and also final equilibrium's lambda. So, instead of having scaled surfaces concentric surfaces, you get something way closer to the final equilibrium.
- Did you also try equilibrium with pressure? I tried this myself with `Heliotron` and the default basis (without and factor) gives extremely bad results, but thinks it has a sum of squares error of  2.008e-17
image

- When you increase the basis to project the spectral coefficients, it again increases the size of the Jacobian and I don't think it helps to the purpose. Maybe we can come up with an heuristic to choose the final basis of the spectral coefficients. So, when you have given LMN resolution for R, Z and lambda, what is the highest frequency in 
image

dpanici:
As is written now, not an improvement over pseudospectral approach, and spectral condensation is not working at all as intended so this PR would need work to warrant merging in

desc/objectives/_geometry.py
```diff
@@ -300,6 +300,146 @@ def compute(self, params, constants=None):
...
+            constants = self.constants
+        data = compute_fun(
+            "desc.equilibrium.equilibrium.Equilibrium",
+            self._data_keys,
+            params=params,
+            transforms=constants["transforms"],
+            profiles=constants["profiles"],
+        )
+
+        W = -(data["R_t"] * data["R_tt"] + data["Z_t"] * data["Z_tt"])
```
dpanici, line 438: This is, I think, from eqn 7a  and 7b of Hirshman and Meier 

Image

Image

not sure completely but probably related to these

ddudt: I think `W` in the code is $-I(\theta,\phi)$ from equation 5a, if $f(m)=m$? Assuming $X \equiv R$ and $Y \equiv Z$.

dpanici: There are extra derivitves though on W in the code it seems. I thought I had worked this out a long time ago, let me try to find the math

PR #383, opened 2022-12-28
Ensure transforms build correctly even with empty grid

f0uriest:


ddudt (approved):
What behavior do we want/expect from `Transform.transform()` when it is built with an empty grid?

f0uriest:
> What behavior do we want/expect from `Transform.transform()` when it is built with an empty grid?

It should just return an empty array of size 0 which is what it does, but after fixing #359 there was an edge case where building with an empty grid wouldn't trigger it to skip all the fft/direct2 checks

PR #385, opened 2022-12-30
Simplify |B| derivative computations

### LINKED ISSUE #384: Simplify |B| derivative computations
The formulas for the partial derivates of $|B|$ were derived from the equation:
$B = \sqrt{\left(B^\theta\right)^2 e_\theta \cdot e_\theta + \left(B^\zeta\right)^2 e_\zeta \cdot e_\zeta + 2 B^\theta B^\zeta e_\theta \cdot e_\zeta}$
These formulas could likely be simplified by instead using the equivalent definition: 
$B = \sqrt{B^\theta B_\theta + B^\zeta B_\zeta}$

This would greatly reduce the number of lines of code used by these functions, and could also provide some slight performance improvements for the quasi-symmetry computations.

ddudt:
Resolves #384 

Also adds some new compute quantities, which may be useful in future applications.


ddudt:
@unalmis Can you please re-implement the `iota_rr` computation?  Here is your [old implementation](https://github.com/PlasmaControl/DESC/blob/78d0bbec65d37b6d7e25aacbda17d3c55b4feff6/desc/compute/_core.py), it just needs to be updated with our new notation. 

Otherwise all of the new compute quantities are here, all that is left is to add tests for them.

ddudt:
Can someone please review `TestConstrainCurrent.test_compute_rotational_transform` in `test_constrain_current.py`?  I think this test is supposed to check the iota computations from fixed-current equilibria to their known profiles, but I don't think it's actually doing that.

desc/compute/_metric.py
```diff
@@ -253,9 +253,51 @@ def _sqrtg_zz(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="sqrt(g)_rt",
+    label="\\partial_{\\rho\\theta} \\sqrt{g}",
+    units="m^{3}",
+    units_long="cubic meters",
+    description="Jacobian determinant, second derivative wrt radial coordinate and "
```
f0uriest, line 261: `Jacobian determinant of flux coordinate system`?

desc/compute/_metric.py
```diff
@@ -295,6 +337,53 @@ def _sqrtg_tz(params, transforms, profiles, data, **kwargs):
...
+        "e_rho_z",
+        "e_theta_z",
+        "e_zeta_z",
+        "e_rho_rz",
+        "e_theta_rz",
+        "e_zeta_rz",
+    ],
+)
+def _sqrtg_rz(params, transforms, profiles, data, **kwargs):
+    data["sqrt(g)_r"] = (
```
f0uriest, line 368: why are there 2 things here?

tests/test_compute_funs.py
```diff
@@ -100,83 +100,147 @@ def test_magnetic_field_derivatives(DummyStellarator):
...
+        rtol=rtol,
+        atol=atol * np.nanmean(np.abs(data["B_zeta_r"])),
+    )
+    np.testing.assert_allclose(
+        data["B_zeta_rr"][3:-2],
+        B_sub_zeta_rr[3:-2],
+        rtol=rtol,
+        atol=atol * np.nanmean(np.abs(data["B_zeta_rr"])),
+    )
+    np.testing.assert_allclose(
```
dpanici, line 205: these B, B_r and B_rr don't make sense to me, isn't `B` and `B_r`, `B_rr` the vector magnetic field? so they should not be compared against the magnetic field magnitude's radial derivative? I am unsure how this even is passing the test....

tests/test_compute_funs.py
```diff
@@ -100,83 +100,147 @@ def test_magnetic_field_derivatives(DummyStellarator):
...
             "B_rho",
+            "B_rho_t",
+            "B_rho_tt",
             "B_theta",
+            "B_theta_t",
+            "B_theta_tt",
             "B_zeta",
+            "B_zeta_t",
+            "B_zeta_tt",
             "|B|",
```
dpanici, line 241: add vector B field's theta derivatives

tests/test_compute_funs.py
```diff
@@ -186,81 +250,113 @@ def test_magnetic_field_derivatives(DummyStellarator):
...
             "B_rho",
+            "B_rho_z",
+            "B_rho_zz",
             "B_theta",
+            "B_theta_z",
+            "B_theta_zz",
             "B_zeta",
+            "B_zeta_z",
+            "B_zeta_zz",
             "|B|",
```
dpanici, line 357: add vector B field's zeta derivatives

unalmis:
> Can someone please review `TestConstrainCurrent.test_compute_rotational_transform` in `test_constrain_current.py`? I think this test is supposed to check the iota computations from fixed-current equilibria to their known profiles, but I don't think it's actually doing that.

It's supposed to check if `iota` on fixed-iota equilibria (which comes directly from the known profile) is recovered when we force `iota` to be computed via fixed current (the `current` being the surface average of `B_theta` on that same equilibrium).

The test fails if the formula for `iota` is wrong. For example adding a `+ 0.3` to `data["iota_r"]` [here](https://github.com/PlasmaControl/DESC/blob/e7f128fd712c70789685f12e1100046ae88bb0ec/desc/compute/_profiles.py#L341) fails the test.

It should only be a valid test on fixed-iota equilibrium, so I made that change. Good catch

dpanici (changes requested):
Can we add tests for the magnetic field vector derivatives? should be able to use existing test infrastructure, just need to split the B vectors up into B_R, B_zeta, B_Z by taking `data['B_r'][:,0]1 etc.

dpanici (approved):

variables.csv
```diff
@@ -0,0 +1,323 @@
+Name,Label,Units,Description,Module
```
f0uriest, line 1: I think this file is supposed to be `docs/variables.csv`

unalmis (approved):

dpanici (dismissed):
Approved once the variables.csv file is moved to the right spot

f0uriest (approved):

dpanici (approved):

PR #386, opened 2022-12-31
Duplicate node spacing

### LINKED ISSUE #286: Weights are scaled incorrectly on grids with duplicate nodes
_Originally posted in https://github.com/PlasmaControl/DESC/pull/279#discussion_r987394648_
Link to comment with notebook to test https://github.com/PlasmaControl/DESC/pull/279#issuecomment-1263193774

Our `_scale_weights` does not reduce duplicate node weight correctly

A simple example of this is `nrho=1, ntheta=5, nzeta=1, endpoint=True`. The correct theta spacing for this `[0.78539816 1.57079633 1.57079633 1.57079633 0.78539816]` with drho = 1 always, dzeta = 2pi always. And at this point in the code that is what the spacing is.

`_scale_weights` then detects the duplicate nodes at 0 and 2pi and scales down the node volume drho * dtheta * dzeta. So the end result of the grid spacing is:
```
[[0.82982653 0.65174424 5.21395388]
 [1.04551592 1.64229256 6.56917025]
 [1.04551592 1.64229256 6.56917025]
 [1.04551592 1.64229256 6.56917025]
 [0.82982653 0.65174424 5.21395388]]
```
The reason I think this is wrong is because a rho surface integral uses the dtheta * dzeta portion of the node volume, ignoring drho. Similar for other surface labels. If we wanted the correct rho surface integral we should leave the drho unchanged and scale up /down dtheta * dzeta.

unalmis:
Resolves #286 for the common case of a duplicate node at theta = 2pi and zeta = 2pi / NFP (endpoint).

For grids with an arbitrary number/position of duplicate nodes, it is tricky to weight the spacing correctly for rho, theta, and zeta surfaces simultaneously. I was a bit stubborn and had a mostly working implementation of this, but I don't think it's worth it over the more practical solution of deleting duplicate nodes prior to grid construction.


f0uriest:
Should try to get rid of the commented out code if possible, or add `# noqa: E800` to those lines.

desc/grid.py
```diff
@@ -139,13 +139,30 @@ def _scale_weights(self, dtheta_scale):
...
+
+        # Spacing is the differential element used for integration over surfaces.
+        # For this, 2 columns of the matrix are used.
+        # The reduction of weight on duplicate nodes should be accounted for
+        # by the 2 columns of spacing which span the surface.
+        # Note we rescale 3 columns by the factor that 'should' rescale 2 columns,
+        # so the grid spacing is valid for integrals over all surface labels.
+        self._spacing /= duplicates ** (1 / 2)
+        # Multiply by duplicates^(1/6) to account for the extra division by
+        # duplicates^(1/2) in one of the columns above. 1/2 = (1/6)^(3 columns).
```
unalmis, line 162: I'll probably delete this line and mention that we're trying to preserve double product of each pair of spacing columns.

f0uriest (approved):

f0uriest (dismissed):

f0uriest (approved):

ddudt (approved):

PR #393, opened 2023-01-04
inequality / bound constraints

### LINKED ISSUE #366: Add option for inequality / bound constraints

ddudt:
Resolves #366


tests/test_examples.py
```diff
@@ -188,14 +187,33 @@ def test_1d_optimization(SOLOVEV):
...
+        FixBoundaryZ(modes=eq.surface.Z_basis.modes[0:-1, :]),
+        FixPressure(),
+        FixIota(),
+        FixPsi(),
+    )
+    options = {"perturb_options": {"order": 1}}
+    with pytest.warns(UserWarning):
+        eq.optimize(objective, constraints, options=options)
+
+    np.testing.assert_allclose(eq.compute("R0/a")["R0/a"], 2.6, rtol=2e-4)
```
f0uriest, line 210: Is this the right test for this? Since technically anything between 2.4 and 2.6 would be acceptable?

f0uriest:
Been thinking about this, I'm wondering how it will work with general constrained optimizers that need to know the upper and lower bound and the value independently. 
I might make sense to add attributes for lower and upper bounds, and if the user passes in a tuple target we assign them to upper/lower bounds, or if the user passes in a single target we set `lower==upper==target`. And have some sort of option of whether to do the clamping inside the objective vs not so we can interface with other optimizers that expect the constraints in a different format?

desc/objectives/objective_funs.py
```diff
@@ -481,6 +481,7 @@ def __init__(
         assert np.all(np.asarray(weight) > 0)
         assert normalize in {True, False}
         assert normalize_target in {True, False}
+        assert len(target) == 2 if isinstance(target, tuple) else True
```
dpanici, line 484: I could see a list being passed in here too, do we want to check if it is a list and convert to tuple if so before doing this assertion?

desc/objectives/objective_funs.py
```diff
@@ -565,10 +566,19 @@ def jit(self):
 
     def _check_dimensions(self):
         """Check that len(target) = len(weight) = dim_f."""
-        self._target = np.asarray(self._target)
+        if isinstance(self.target, tuple):
+            self._target = tuple([np.asarray(tar) for tar in self._target])
+            for tar in self.target:
+                if not is_broadcastable((self.dim_f,), tar.shape):
```
dpanici, line 572: this is for if we have some vector of target like in FixBoundaryR, and we passed in a target R_b that is actually a tuple of (R_b_lower,R_b_upper)?

desc/objectives/objective_funs.py
```diff
@@ -624,14 +634,44 @@ def _shift_scale(self, x):
         target = (
             self.target / self.normalization if self._normalize_target else self.target
```
dpanici, line 635: won't this line fail if bounds were passed and `self._normalize_target = True`? since it would attempt to divide a tuple by a float (it would fail if dividing by any number really)

tests/test_objective_funs.py
```diff
@@ -264,3 +265,48 @@ def test_generic_compute():
...
+        GenericObjective("R", target=10 * w, weight=3, grid=grid), eq=eq
+    )
+    f_scaled = obj.compute(obj.x(eq))
+    f_unscaled = obj.objectives[0]._unshift_unscale(f_scaled)
+    assert np.allclose(f_unscaled, data)
+
+    # test objective with a bounded target
+    obj = ObjectiveFunction(
+        GenericObjective(
+            "R", target=(9.4 * w, np.ones_like(data) * 10.6 * w), weight=2, grid=grid
```
dpanici, line 301: change this to have `normalize=True` and I'm pretty sure it'll throw an error

desc/objectives/objective_funs.py
```diff
@@ -624,14 +634,44 @@ def _shift_scale(self, x):
...
+                * self.weight
+            )
+        else:
+            return (x_norm - target) * self.weight
 
-    def _unshift_unscale(self, x):
+    def _unshift_unscale(self, f):
         """Undo target and weighting."""
         target = (
             self.target / self.normalization if self._normalize_target else self.target
```
dpanici, line 657: same comment as above with this line

desc/objectives/objective_funs.py
```diff
@@ -624,14 +634,44 @@ def _shift_scale(self, x):
...
         )
-        return (jnp.atleast_1d(x) / self.normalization - target) * self.weight
+        x_norm = jnp.atleast_1d(x) / self.normalization
+        if isinstance(target, tuple) and len(target) == 2:
+            return (
```
dpanici, line 639: what is this doing?

desc/objectives/objective_funs.py
```diff
@@ -624,14 +634,44 @@ def _shift_scale(self, x):
...
-    def _unshift_unscale(self, x):
+    def _unshift_unscale(self, f):
         """Undo target and weighting."""
         target = (
             self.target / self.normalization if self._normalize_target else self.target
         )
-        return (x / self.weight + target) * self.normalization
+        f_unweighted = f / self.weight
+        if isinstance(target, tuple) and len(target) == 2:
+            return (
```
dpanici, line 661: what is this doing?

tests/test_examples.py
dpanici: yea it is unclear to me how this tests it, I guess just that larger aspect ratio should give lower F balance?

tests/test_objective_funs.py
dpanici: I changed the test for you so it fails now 😈

dpanici (changes requested):
fix bug and explain the parts I asked and it should be good to go

desc/objectives/objective_funs.py
ddudt: Lists are already used for targets in the original sense (previous to this PR), where they get cast to jnp arrays. So giving the target as a tuple is the only allowable way to indicate that it is meant to be lower/upper bounds on the target.

desc/objectives/objective_funs.py
ddudt: Correct. Although these target bounds aren't intended for linear objectives that get used as constraints like `FixBoundaryR`. A better example would be for the `ForceBalance` objective, where you are giving (F_lower, F_upper).

desc/objectives/objective_funs.py
ddudt: I added comments in the code to explain

desc/objectives/objective_funs.py
ddudt: See added comments in the code

tests/test_examples.py
ddudt: I changed the test to solve an equilibrium with a target force balance of `-1e3 N <= F <= +1e3 N`, and check that the final residuals are all nearly 0. 

This works, but I should note that it makes things more challenging for the optimizer. In fact, giving `target=0` yields better results than `target=(-1e3, 1e3)` in this example. I think this is because the optimizer is losing information about the cost landscape when we make most of the residuals 0, and it struggles to minimize the remaining nonzero terms.

desc/objectives/objective_funs.py
dpanici: ok that makes sense,

dpanici (approved):

ddudt:
> Been thinking about this, I'm wondering how it will work with general constrained optimizers that need to know the upper and lower bound and the value independently. I might make sense to add attributes for lower and upper bounds, and if the user passes in a tuple target we assign them to upper/lower bounds, or if the user passes in a single target we set `lower==upper==target`. And have some sort of option of whether to do the clamping inside the objective vs not so we can interface with other optimizers that expect the constraints in a different format?

This is addressed now; see new changes.  

API note: `bounds` will always override `target`

tests/test_objective_funs.py
```diff
@@ -296,33 +297,37 @@ def build(self, eq, use_jit=True, verbose=1):
             super().build(eq, use_jit, verbose)
 
         def compute(self, R_lmn):
-            return self.target * self.weight + self.y * R_lmn**3
+            return 200 + self.target * self.weight - self.y * R_lmn**3
```
f0uriest, line 300: Can you explain the changes in this test? not sure I follow all of them

f0uriest (dismissed):

f0uriest:
Approving for now, but still need to add bounds to the new geometry objectives. I'd also vote for holding off merging this since this changes some API stuff so should be put out as a new release, but there's some other things we can get in first.

tests/test_objective_funs.py
ddudt: This test checks to make sure that `target` and `weight` get updated after re-JIT-compiling. In the old API, the `_Objective.compute` function was JIT-compiled, but in the new API it is the `_Objective.compute_scaled` function that gets JIT-compiled. So even if the `DummyObjective.compute()` function was not modified, the tests would still fail because of the new functionality. Since I had to update the "correct" test values anyways, I decided to modify the compute function to make the scale of the numbers more reasonable. 

I also added new checks to the test related to how `DummyObjective.compute()` can give inconsistent behavior from `DummyObjective.compute_scaled()` before re-JIT-compiling.

f0uriest (approved):

desc/objectives/linear_objectives.py
```diff
@@ -177,8 +177,7 @@ def compute(self, *args, **kwargs):
             x = kwargs.get(self.args[0], args[0])
         else:
             x = kwargs.get(self.args[0])
-        Rb = jnp.dot(self._A, x)
-        return self._shift_scale(Rb)
+        return jnp.dot(self._A, x)
```
dpanici, line 180: Does it make any sense to impement bounds on FixBoundaryR and Z? Like if you want to let a mode vary only a bit?

dpanici (dismissed):
Looks good

desc/objectives/linear_objectives.py
ddudt: The bounds do get applied to linear objectives, just like every other Objective function. The bounds are implemented in `_Objective.compute_scaled`, which internally calls `_Objective.compute`. All of the objective Jacobians are computed using the scaled version (with weights, etc.), and the constraint utility function logic also uses the new scaled compute functions.

dpanici: I see that, but where in the linear objectives do you put the bounds argument I mean? In all the other objectives the bounds is an argument passed in where you specify a tuple of upper and lower bounds, instead of giving a target. This is missing in all of the linear objectives. Maybe I am misunderstanding your point here

ddudt: Oh good catch! I just added them.

dpanici (dismissed):

f0uriest (dismissed):

dpanici:
leaving a comment/clarification for my sake, this PR specifically implements inquality constraints for any objective by using a quadratic penalty term that has a dead-zone defined by the bounds passed into the objective (so cost is 0 inside these bounds and is ~ (x-x_bound)^2 where x_bound is the upper or lower bound depending on x > x_upper_bound or x<x_lower_bound)

So any desired inequality constraint MUST be passed in as an OBJECTIVE to the optimization problem, not as a constraint (since constraints are almost always assumed to be linear and then treated by optimizing in the nullspace of that constraint, which does not work for linear inequalities)

f0uriest (approved):

tests/test_examples.py
f0uriest: Was this ever fixed? Right now it looks like the only bound is on the aspect ratio but the assertion is that its a particular value?

dpanici (approved):

tests/test_examples.py
ddudt: This function `test_1d_optimization_bounds` was replaced with `test_solve_bounds`. I'm not sure why it isn't updating in this PR GUI here, but when I pull the code locally it is correct.

f0uriest: Ah ok I see. thanks.

PR #395, opened 2023-01-05
Add calculation of force error with anisotropic pressure tensor

### LINKED ISSUE #312: Allow computing equilibria with anisotropic pressure
From Eduardo's series of papers we know that in theory allowing pressure anisotropy can improve QS. We have most of what we need for that, so we should add that capability.

f0uriest:
Resolves #312


unalmis (changes requested):
There are places we assume that the partial derivative of pressure with respect to rho is the total derivative. I suggest making the following changes.

- [ ] Make new variables $\langle p \rangle$ and $\partial_{\rho} \langle p \rangle$, where the surface averages are the infinitesimal volume averages (i.e. weighted with the $\sqrt g$).
- [ ] Push the ``dp_dpsi`` factor [here](https://github.com/PlasmaControl/DESC/blob/61797b69e6991b0284e76e0e493d9240edc779e0/desc/compute/_stability.py#L111) into the integrals to correctly compute the Mercier magnetic well.
- [ ] Push ``p_r`` [here](https://github.com/PlasmaControl/DESC/blob/61797b69e6991b0284e76e0e493d9240edc779e0/desc/compute/_stability.py#L226) into the surface average of `|B|^2` for the magnetic well. This is most easily done by replacing `p_r` with the new variable `<p>_r`.

Is `F_rho` fine as is?

f0uriest:
It's a bit tricky, since the derivation of mercier stability (and most MHD etc) assumes a scalar pressure that is a flux function, so I'm not 100% that just taking the surface average is the "correct" thing to do if pressure isnt a scalar or isnt a flux function.

In the cases we're interested in, the anisotropy is very small, so it shouldn't make much difference if we just ignore any variation.

dpanici:
Do you have your derivation of the anisotropic force? I just want to double check it as it is not immediately obvious to me that yours is equivalent when comparing it against the one in Eduardo's paper (https://arxiv.org/pdf/2008.04715.pdf eqn 37)

f0uriest:
> Do you have your derivation of the anisotropic force? I just want to double check it as it is not immediately obvious to me that yours is equivalent when comparing it against the one in Eduardo's paper (https://arxiv.org/pdf/2008.04715.pdf eqn 37)

yeah he was working in normalized units so missing some factors of mu0, I re-derived it here:
[pressure_anisotropy.pdf](https://github.com/PlasmaControl/DESC/files/12422886/pressure_anisotropy.pdf)

desc/objectives/linear_objectives.py
```diff
@@ -1897,6 +1897,109 @@ def compute(self, p_l, **kwargs):
...
+            )
+        profile = eq.anisotropy
+        super().build(eq, profile, use_jit, verbose)
+
+    def compute(self, a_lmn, **kwargs):
+        """Compute fixed pressure profile errors.
+
+        Parameters
+        ----------
+        d_lmn : ndarray
```
dpanici, line 1991: says d_lmn, but it should be a_lmn

dpanici (dismissed):

dpanici:
Tests not passing due to the `test_everything` compute test. @unalmis Maybe can you put some instruction somewhere on how to properly add new things to the compute indices in order to avoid this test failing? 

I think in this case what is happening is while on master grad(p) is only in the grad(rho) direction, here grad(p) has grad(theta) as well as grad(zeta). One of those axis limits must eval to NaN, as on this branch grad(p) on axis returns NaN

f0uriest:
Fixed the test. 

The benchmark data is automatically updated when you run the tests if there are missing keys

dpanici (dismissed):

dpanici (approved):

dpanici (approved):

dpanici (dismissed):

dpanici (approved):

desc/objectives/_equilibrium.py
```diff
@@ -216,6 +216,201 @@ def compute(self, *args, **kwargs):
...
+        data = compute_fun(
+            "desc.equilibrium.equilibrium.Equilibrium",
+            self._data_keys,
+            params=params,
+            transforms=constants["transforms"],
+            profiles=constants["profiles"],
+        )
+        f = (
+            data["sqrt(g)"]
+            * constants["transforms"]["grid"].weights
```
ddudt, line 407: I don't think we should multiply by the quadrature weights here after PR #484. 

Also you should add this objective to the new test from that PR to make sure it is independent of the grid resolution.

desc/objectives/_equilibrium.py
```diff
@@ -216,6 +216,201 @@ def compute(self, *args, **kwargs):
...
+            * data["F_anisotropic"].T
+        ).T
+
+        return f.flatten(order="F")  # to line up with quad weights
```
ddudt, line 411: I don't understand why this is necessary -- isn't `f` already a 1D array?

tests/test_examples.py
```diff
@@ -62,6 +63,27 @@ def test_SOLOVEV_results(SOLOVEV):
     np.testing.assert_allclose(theta_err, 0, atol=1e-4)
 
 
+@pytest.mark.regression
+@pytest.mark.solve
+def test_SOLOVEV_anisotropic_results(SOLOVEV):
```
ddudt, line 68: This is good, but it doesn't actually test that solving with non-zero anisotropy works correctly. Do we have any benchmark cases we could use for that?

desc/compute/_profiles.py
```diff
@@ -464,10 +500,14 @@ def _p_r(params, transforms, profiles, data, **kwargs):
...
-    data=["p_r", "e^rho"],
+    data=["p_r", "p_t", "p_z", "e^rho", "e^theta", "e^zeta"],
 )
 def _gradp(params, transforms, profiles, data, **kwargs):
-    data["grad(p)"] = (data["p_r"] * data["e^rho"].T).T
+    data["grad(p)"] = (
+        data["p_r"] * data["e^rho"].T
+        + jnp.where(data["p_t"] == 0, 0, data["p_t"] * data["e^theta"].T)
+        + jnp.where(data["p_z"] == 0, 0, data["p_z"] * data["e^zeta"].T)
+    ).T
```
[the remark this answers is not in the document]
unalmis: e^zeta is finite at axis, so you should be able to leave that alone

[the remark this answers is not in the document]
f0uriest: yeah ill change it. @ddudt `e^theta` however isn't so to get a physical answer we need to make sure we handle that case separately otherwise it returns NaN

desc/compute/utils.py
```diff
@@ -1303,4 +1303,5 @@ def body(i, mins):
     "electron_density": "ne_l",
     "ion_temperature": "Ti_l",
     "atomic_number": "Zeff_l",
+    "anisotropy": "a_lmn",
```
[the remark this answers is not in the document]
f0uriest: that would be anisotropic scalar pressure (ie, scalar pressure but a function of r,t,z). What we actually want is an anisotropic pressure tensor, hence the need for the new profile to represent the difference between the parallel and perpendicular pressures (see the notes linked above for a full description)

desc/objectives/_equilibrium.py
```diff
@@ -216,6 +216,201 @@ def compute(self, *args, **kwargs):
...
+            Equilibrium that will be optimized to satisfy the Objective.
+        use_jit : bool, optional
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        eq = eq or self._eq
+        if self._grid is None:
+            if eq.node_pattern is None or eq.node_pattern in [
```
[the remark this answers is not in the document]
f0uriest: yeah but ill do another PR for that

desc/objectives/_equilibrium.py
f0uriest: will do

desc/objectives/_equilibrium.py
f0uriest: No, this `f` is a Nx3 vector so needs to be flattened. Because of the anisotropy we can't split into rho/helical parts so we just use the vector in the lab frame.

tests/test_examples.py
f0uriest: Not really. Frank and I are working on comparing to near axis stuff but still in progress.

dpanici (approved):

dpanici (dismissed):

PR #399, opened 2023-01-10
QS Boozer updates

### LINKED ISSUE #370: Fix helicity indexing in boozer transform
Need to do Ptolemy identity to convert to `m*t-n*z` form

ddudt:
Resolves #370.


ddudt:
It looks like the only test that is failing is `plot_boozer_modes`, and the image that the CI generates is just very slightly shifted from the baseline image I uploaded. Can someone else try re-uploading it, in case it is a problem with my version of `matplotlib` (3.4.3)?

f0uriest (dismissed):

f0uriest (dismissed):

desc/objectives/_qs.py
```diff
@@ -127,7 +127,7 @@ def build(self, eq, use_jit=True, verbose=1):
             N_booz=self.N_booz,
         )
         self._matrix, self._modes, self._idx = ptolemy_linear_transform(
-            self._transforms["B"].basis, (self.helicity[0], self.helicity[1] / eq.NFP)
+            self._transforms["B"].basis, self.helicity[0]
```
dpanici, line 130: fix typo, should not have indexing into helicity here

dpanici (approved):

PR #400, opened 2023-01-11
fix plot_qs_error bug

dpanici:
Remove data=data from compute in plot_qs_error to fix bug, and make plot_qs_error test use a stellarator to actually test it

this fixes the bug because before the compute fxn refactor, passing in data=data was fine bc the keys of that dict would just be overwritten. But now, passing in data=data has different functionality (it avoids redundant computations) which causes the qs calcs in the plotting fxn to have the same value for each rho surface (I believe the specific problem is, data['|B|_mn'] etc already exists so it does not re-calculate that, and so it uses the SAME data dict for every rho surface even tho the grid  has changed, because it thinks everything has already been calculated)


tests/test_plotting.py
```diff
@@ -717,12 +718,14 @@ def test_plot_boozer_surface(DSHAPE_current):
 @pytest.mark.unit
 @pytest.mark.solve
 @pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_1d)
-def test_plot_qs_error(DSHAPE_current):
+def test_plot_qs_error():
     """Test plotting qs error metrics."""
-    eq = EquilibriaFamily.load(load_from=str(DSHAPE_current["desc_h5_path"]))[-1]
-    fig, ax, data = plot_qs_error(eq, helicity=(0, 0), log=False, return_data=True)
+    eq = get("HELIOTRON")
+    fig, ax, data = plot_qs_error(eq, helicity=(1, 19), log=False, return_data=True)
```
ddudt, line 724: I would recommend using the WISTELL-A example, since that is optimized for QS and HELIOTRON is not:
```
eq = get("WISTELL-A")
fig, ax, data = plot_qs_error(eq, helicity=(1, -eq.NFP), log=False, return_data=True)
```

ddudt (changes requested):

tests/test_plotting.py
```diff
@@ -720,8 +720,8 @@ def test_plot_boozer_surface(DSHAPE_current):
 @pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_1d)
 def test_plot_qs_error():
     """Test plotting qs error metrics."""
-    eq = get("HELIOTRON")
-    fig, ax, data = plot_qs_error(eq, helicity=(1, 19), log=False, return_data=True)
+    eq = get("WISTELL-A")
+    fig, ax, data = plot_qs_error(eq, helicity=(1, eq.NFP), log=False, return_data=True)
```
ddudt, line 724: The helicity needs to be negative for this example: `helicity=(1, -eq.NFP)`

ddudt (changes requested):

tests/test_plotting.py
```diff
@@ -717,12 +718,14 @@ def test_plot_boozer_surface(DSHAPE_current):
...
-def test_plot_qs_error(DSHAPE_current):
+def test_plot_qs_error():
     """Test plotting qs error metrics."""
-    eq = EquilibriaFamily.load(load_from=str(DSHAPE_current["desc_h5_path"]))[-1]
-    fig, ax, data = plot_qs_error(eq, helicity=(0, 0), log=False, return_data=True)
+    eq = get("WISTELL-A")
+    fig, ax, data = plot_qs_error(eq, helicity=(1, eq.NFP), log=False, return_data=True)
     for string in ["rho", "f_T", "f_B", "f_C"]:
         assert string in data.keys()
+        if string != "rho":
```
f0uriest, line 727: might want to include a comment about why we're doing this test, out of context it looks kind of weird

f0uriest (commented):
The legend on the plot has f_B twice?

tests/test_plotting.py
ddudt: Also I would recommend `log=True` so that we can actually see the f_B values in the plot.

f0uriest (approved):

ddudt (approved):

PR #405, opened 2023-01-15
Add missing unit test markers

unalmis:
`tests/test_surfaces.py::TestFourierRZToroidalSurface::test_curvature` and `tests/test_surfaces.py::test_surface_orientationtests/test_surfaces.py::test_surface_orientation` fail when I test locally. I think the CI doesn't catch this because they were missing the unit test markers.


f0uriest:
Ahh I thought I added those markers in another PR but maybe they got lost. Should be fixed now.

f0uriest (approved):

PR #406, opened 2023-01-16
Add objective for penalizing plasma-vessel distance

### LINKED ISSUE #339: Add objective for penalizing distance between plasma boundary and winding surface / vacuum vessel
There are a lot of possible ways but the simplest brute force would just be to compute R,Z on a dense grid on both the plasma surface and the exterior surface and minimize 1/softmin(dr) or something.

f0uriest:
Resolves #339


desc/objectives/_geometry.py
```diff
@@ -357,3 +363,147 @@ def compute(self, *args, **kwargs):
...
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        if self._surface_grid is None:
+            self._surface_grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
+        if self._plasma_grid is None:
+            self._plasma_grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
+        if not np.allclose(self._surface_grid.nodes[:, 0], 1):
```
dpanici, line 448: why does the surface_grid even check rho points? should they not just be theta,zeta points as the surface is a FourierRZToroidalSurface, and so does not really have any rho dependence?

tests/test_objective_funs.py
```diff
@@ -353,3 +355,45 @@ def test_target_profiles():
...
+    # For equally spaced grids, should get true d=1
+    surf_grid = LinearGrid(M=5, N=6)
+    plas_grid = LinearGrid(M=5, N=6)
+    obj = PlasmaVesselDistance(
+        eq=eq, plasma_grid=plas_grid, surface_grid=surf_grid, surface=surface
+    )
+    d = obj.compute(*obj.xs(eq))
+    np.testing.assert_allclose(d, a_s - a_p)
+
+    # for unequal M, should have error of order M_spacing*a_p
```
dpanici, line 381: can you explain this geometry?

dpanici (dismissed):

tests/test_objective_funs.py
```diff
@@ -353,3 +355,45 @@ def test_target_profiles():
...
+    surf_grid = LinearGrid(M=5, N=6)
+    plas_grid = LinearGrid(M=10, N=6)
+    obj = PlasmaVesselDistance(
+        eq=eq, plasma_grid=plas_grid, surface_grid=surf_grid, surface=surface
+    )
+    d = obj.compute(*obj.xs(eq))
+    assert abs(d.min() - (a_s - a_p)) < 1e-14
+    assert abs(d.max() - (a_s - a_p)) < surf_grid.spacing[0, 1] * a_p
```
dpanici, line 389: maybe I am missing something here, for same sapcing shouldn't d.max = a_s + a_p (at a given phi)...

desc/objectives/_geometry.py
```diff
@@ -357,3 +363,147 @@ def compute(self, *args, **kwargs):
...
+        params = self._parse_args(*args, **kwargs)
+        data = compute_fun(
+            self._data_keys,
+            params=params,
+            transforms=self._transforms,
+            profiles=self._profiles,
+        )
+        plasma_coords = rpz2xyz(jnp.array([data["R"], data["phi"], data["Z"]]).T)
+        d = jnp.linalg.norm(
+            plasma_coords[:, None, :] - self._surface_coords[None, :, :], axis=-1
```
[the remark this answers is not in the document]
dpanici: And does this only check distances at the given phi cross-sections from grid zeta values, or does it check distances btwn every point on the surface against every other point (so even phi=0 and phi=pi values are compared), I assume not the latter?

desc/objectives/_geometry.py
f0uriest: In theory yes, any rho coordinate will be ignored. So this check is more to make the user aware that there might be duplicate points which are wasteful (ie, same theta/zeta but different rho are really the same point).

desc/objectives/_geometry.py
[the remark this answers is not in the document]
f0uriest: Yes, this is the distance matrix between every point on the plasma boundary and every point on the bounding surface, and it will look at every cross section. If the grids are different resolution or staggered or something then the closest point to phi=0 might be at phi=2pi etc.

tests/test_objective_funs.py
f0uriest: The geometry is the same for all test cases. Plasma is a circular tokamak of R0=10, a=1, and the boundary surface is the same with a=2, so the true distance is 1 for all points. However, if the grids are different, the pointwise distances that we compute won't be the true min, but will have an error on the order of the arclength between points, which is the grid spacing * minor radius (basically the triangle inequality)

tests/test_objective_funs.py
f0uriest: This d is the minimum distance from each point on the plasma to any point on the boundary, not the full distance matrix. So d.max() - a_s - a_p is really the maximum error in the distance estimate, which should be O(spacing*a)

dpanici (approved):

desc/objectives/_geometry.py
```diff
@@ -360,6 +365,150 @@ def compute(self, *args, **kwargs):
...
+        if self._surface_grid is None:
+            self._surface_grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
+        if self._plasma_grid is None:
+            self._plasma_grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
+        if not np.allclose(self._surface_grid.nodes[:, 0], 1):
+            warnings.warn("Surface grid includes off-surface pts, should be rho=1")
+        if not np.allclose(self._plasma_grid.nodes[:, 0], 1):
+            warnings.warn("Plasma grid includes interior points, should be rho=1")
+
+        self._dim_f = self._plasma_grid.num_nodes
```
ddudt, line 453: Change so that `dim_f`, `target`, and `weight` correspond to the `surface_grid`, since that is fixed during optimization and more intuitive.

ddudt (changes requested):

PR #408, opened 2023-01-16
Add objectives for penalizing strong shaping

### LINKED ISSUE #407: Add objectives for targeting curvature etc

f0uriest:
Resolves #407

### DIFF
--- a/desc/objectives/__init__.py
+++ b/desc/objectives/__init__.py
@@ -8,7 +8,13 @@ from ._equilibrium import (
     RadialForceBalance,
 )
 from ._generic import GenericObjective, RotationalTransform, ToroidalCurrent
-from ._geometry import AspectRatio, Elongation, Volume
+from ._geometry import (
+    AspectRatio,
+    Elongation,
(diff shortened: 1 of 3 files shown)


unalmis (approved):

ddudt (commented):
I think there is room for improvement in the tests here. Can we add one where we know the expected curvatures for a given equilibrium example, such as the simple `eq = Equilibrium()` case or something similar? Also we should try running some sort of optimization with these objectives as a sanity check, even if we don't include that in a test.

f0uriest:
> I think there is room for improvement in the tests here. Can we add one where we know the expected curvatures for a given equilibrium example, such as the simple `eq = Equilibrium()` case or something similar? Also we should try running some sort of optimization with these objectives as a sanity check, even if we don't include that in a test.

We already test the formulas in the surface class so I think they're fine. I would like tot try using it for optimization but ideally we'd need the bounds in #393 to do it properly

ddudt (approved):

dpanici (approved):

PR #409, opened 2023-01-17
Remaining bugs with duplicate nodes, add grid developer guide

unalmis:
- [x] Squash bugs
    A list of bug fixes follows. The use case that would uncover these bugs is rather narrow, which is why it wasn't detected by previous tests, and is unlikely to affect most stuff. All of these bugs involve duplicate nodes, however, none of them were introduced by #386 . This PR just finds bugs that were missed by that one.

    - Fixed issue where theta and zeta surface integrals on grids with duplicate node at 0 and 2 pi (or 2 pi/NFP) (e.g. when endpoint=True) computed a value that was off by a factor of 2 only on the surface at 0 and 2 pi. All other surfaces were computed correctly.
    In other words, 2 elements of the returned array of length `grid.num_surface_label` were off by factor 2.
    - More massaging of inputs so that the grid class works with weird inputs. This is required for correctness of some of the algorithms, but unlikely to be the cause of any bugs as a runtime error would have been raised immediately.
    - Added logic so `LinearGrid`s with duplicate nodes and `symmetry` have correct weights and spacing.
    - If the user supplies an incorrect marker for endpoint, for example by providing `endpoint=False` to the `LinearGrid` constructor while also providing arrays with nodes that include a duplicate endpoint, the weights would be computed incorrectly. This was fixed so that the correct weights are computed  regardless of the user's tricks.

- [x] Add bug catchers
    - many more tests
- [x] Add bug repellent
    - add walk-through of `grid.py` to developer guide explaining everything with nice visuals


unalmis:
html of thr jupyter notebook: [grid_doc.zip](https://github.com/PlasmaControl/DESC/files/10660039/grid_doc.zip)

docs/notebooks/dev_guide/grid.ipynb
```diff
@@ -0,0 +1,1225 @@
+{
```
f0uriest, line 1: <p>Shouldn't the rho integrals should be $\int_0^1 \rho d\rho$?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/409/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

f0uriest, line 1: <p>θ</p><p>&lt;sup&gt;3 coordinate &gt; π</p><p> Typo?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/409/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

unalmis: Yes. That was supposed to be a footnote marker at the end of the sentence.

unalmis: Do you mean that in the sense that $\rho$ is the area Jacobian for basic polar geometries, with $dA = \rho d\rho d\theta$?

If we wanted to compute the actual area of a $\zeta$ surface, we would weight it by the area Jacobian for that surface in our geometry: $\int_0^1 \int_0^{2\pi} \lvert \ e_{\rho} \times e_{\theta} \rvert d\rho d\theta$

When I mention "node area" in this document I am just referring to the product of the differential elements in the columns of `grid.spacing` for the row associated with that node. For a $\zeta$ surface the unweighted area, which is the sum of these products over all the nodes on the surface, would be $\int_0^1 \int_0^{2\pi} d\rho d\theta = 2\pi$.

I think that is an invariant the grid should try to keep. That way when we actually supply a Jacobian factor in the integral, whether that be for an area or volume, we know that the integral covers the entire domain.

f0uriest: Yeah I was thinking of it like the area in the computational domain which in this case is a disk, so the area should be pi. But I can see your point as well about keeping the Jacobian separate. I'm fine with it if everyone else is.

f0uriest (dismissed):

tests/test_grid.py
```diff
@@ -74,34 +75,65 @@ def test_linear_grid(self):
...
+            # and ntheta is even when endpoint is False
+            ntheta = 6 + ntheta_is_odd
             nzeta = 7
             NFP = 3
-            grid1 = LinearGrid(
+            theta = np.linspace(0, 2 * np.pi, ntheta, endpoint=endpoint)
+            if sym:
+                # When users supply nodes and set symmetry to true, they need
+                # to ensure their nodes are placed symmetrically.
+                # Here we change the array to match the effect of line 481
```
ddudt, line 89: We should avoid referencing specific line numbers, since those will change as the file is edited in future PRs.

ddudt (dismissed):

f0uriest (dismissed):

f0uriest (approved):

PR #413, opened 2023-01-18
make default tolerances when converting input file from VMEC to DESC more reasonable and clear

dpanici:
change default tolerances from 0 to more reasonable values, and make more clear that default nfev is 100. Adds test as well

### DIFF
--- a/desc/io/input_reader.py
+++ b/desc/io/input_reader.py
@@ -690,13 +690,14 @@ class InputReader:
             )
 
         f.write("\n# solver tolerances\n")
+        defaults = {"ftol": "1e-2", "xtol": "1e-6", "gtol": "1e-6", "nfev": "100"}
         for key in ["ftol", "xtol", "gtol", "nfev"]:
             f.write(
                 key
                 + " = {}\n".format(
                     ", ".join(
                         [
-                            str(inp[key]) if inp[key] is not None else str(0)
+                            str(inp[key]) if inp[key] is not None else defaults[key]
                             for inp in inputs
                         ]
                     )
--- a/tests/test_input_output.py
+++ b/tests/test_input_output.py
@@ -35,6 +35,15 @@ def test_vmec_input(tmpdir_factory):
         v.pop("output_path")
     assert all([equals(in1, in2) for in1, in2 in zip(vmec_inputs, desc_inputs)])
 
+    # test default solver tolerances are correct
+    defaults = {"ftol": "1e-2", "xtol": "1e-6", "gtol": "1e-6", "nfev": "100"}
+    with open(path) as f:
+        lines = f.readlines()
+        for line in lines:
+            for key in defaults.keys():
+                if line.find(key) != -1:
+                    assert line.split("=").split(",")[-1] == defaults[key]
+
 
 @pytest.mark.unit
 def test_near_axis_input_files():


desc/io/input_reader.py
```diff
@@ -690,13 +690,14 @@ def write_desc_input(filename, inputs, header=""):
...
         f.write("\n# solver tolerances\n")
+        defaults = {"ftol": "1e-2", "xtol": "1e-6", "gtol": "1e-6", "nfev": "100"}
         for key in ["ftol", "xtol", "gtol", "nfev"]:
             f.write(
                 key
                 + " = {}\n".format(
                     ", ".join(
                         [
-                            str(inp[key]) if inp[key] is not None else str(0)
+                            str(inp[key]) if inp[key] is not None else defaults[key]
```
f0uriest, line 700: I'm wondering if it might be better to not write these tolerances at all if they're not assigned. Everywhere else in the code we have defaults=`None` for tolerances, because the default tolerances get assigned here: https://github.com/PlasmaControl/DESC/blob/83ae6d1b0ed17d71dceb3d9528c803a75ef89c18/desc/optimize/optimizer.py#L507 and in theory we might want different defaults for different optimizers etc. When parsing an input file, these default to `None` as well if the field isn't present: https://github.com/PlasmaControl/DESC/blob/83ae6d1b0ed17d71dceb3d9528c803a75ef89c18/desc/io/input_reader.py#L170 so that would correctly propagate into the optimizer where the defaults get set. This way we only have a single place where we assign default values so it's hopefully easier to maintain.

dpanici: Yea that is probably the better way to approach this

dpanici:
Further question then @f0uriest @ddudt , the default for lsqtr is ftol=1e-6 which is really low for an equilibrium solve, in practice we usually use ftol=1e-2, so can I change that in the optimizer?

dpanici:
Seems the default tolerances don't correctly use xtol at least, as here I run desc on the SOLOVEV VMEC case, and even though the input file has no solver tols specified, the xtol triggers at 1e-04, instead of 1e-06 like the default should according to the lsqtr function.

```
Starting optimization
       0              1         1.6078e-10                                    4.29e-04
       1              2         1.2645e-13      1.61e-10       2.87e-04       4.85e-05
Optimization terminated successfully.
`xtol` condition satisfied.
         Current function value: 1.264e-13
         Total delta_x: 2.866e-04
         Iterations: 1
         Function evaluations: 2
         Jacobian evaluations: 2
```

[vmec_desc_SOLOVEV_output.txt](https://github.com/PlasmaControl/DESC/files/10451543/vmec_desc_SOLOVEV_output.txt)

f0uriest:
> Further question then @f0uriest @ddudt , the default for lsqtr is ftol=1e-6 which is really low for an equilibrium solve, in practice we usually use ftol=1e-2, so can I change that in the optimizer?

I had set the default in `lsqtr` to be smaller to keep it more "general purpose". For our specific case 1e-2 is fine, and thats what gets set in `Optimizer` which is where we always call it from. I'd prefer to keep the 1e-6 default in `lsqtr` for the rare cases where we call if by itself (such as in tests, and possibly in other cases where we just need a general purpose optimizer without going through the whole `Equilibrium`/`Objective`/`Optimizer` setup.

desc/io/input_reader.py
```diff
@@ -691,6 +691,13 @@ def write_desc_input(filename, inputs, header=""):
 
         f.write("\n# solver tolerances\n")
         for key in ["ftol", "xtol", "gtol", "nfev"]:
+            at_least_one_is_none = False
+            for inp in inputs:
+                if inp[key] is None:
+                    at_least_one_is_none = True
+            if at_least_one_is_none:
+                break  # don't write if an input tolerance is None
```
ddudt, line 699: This won't write any of the stopping criteria if at least one of them is None. Why not only skip the ones that are None? If a value is passed in for one parameter, it seems odd to overwrite it with the default just because some other parameter was not specified.

dpanici: The logic was just getting a bit annoying to do. I can make that change

dpanici:
- [ ] add actual test for vmec to DESC conversion input file
- [ ] add a test for write_desc_input, exercising when only 1 of the ftol list etc is None

desc/io/input_reader.py
```diff
@@ -640,7 +640,7 @@ def parse_inputs(self, fname=None):  # noqa: C901 - FIXME: simplify this
         return inputs_list
 
     @staticmethod
-    def write_desc_input(filename, inputs, header=""):
+    def write_desc_input(filename, inputs, header=""):  # noqa: 16 fxn too complex
```
f0uriest, line 643: syntax for this should be `# noqa: C901` to ignore that specific error, otherwise it ignores everything.

desc/io/input_reader.py
```diff
@@ -686,18 +686,26 @@ def write_desc_input(filename, inputs, header=""):
...
             )
 
         f.write("\n# solver tolerances\n")
         for key in ["ftol", "xtol", "gtol", "nfev"]:
+            inputs_not_None = []
+            for inp in inputs:
+                if inp[key] is not None:
+                    inputs_not_None.append(inp)
+            if not inputs_not_None:  # an  empty list evals to False
+                break  # don't write line if all input tolerance are None
```
f0uriest, line 700: I think we want `continue` (skip to next iteration) here instead of `break` (exit the loop entirely)

tests/test_input_output.py
```diff
@@ -35,6 +38,59 @@ def test_vmec_input(tmpdir_factory):
...
+def test_write_desc_input_Nones(tmpdir_factory):
+    """Test converting writing DESC input file when an input tol is None."""
+    # tests how None is handled as a passed-in input to the input writer
+    input_path = "./tests/inputs/DSHAPE"
+    tmpdir = tmpdir_factory.mktemp("desc_inputs")
+    tmp_path = tmpdir.join("DSHAPE")
+    shutil.copyfile(input_path, tmp_path)
+    ir = InputReader(cl_args=[str(tmp_path)])
+
+    ftols_input_with_none = [1e-2, None, 1e-3]
```
f0uriest, line 69: can you explain this expected behavior a bit? I thought if it's `None` it doesn't get written?

desc/io/input_reader.py
dpanici: you're right, I added a test for this now (gtol will not be written but rest will be)

tests/test_input_output.py
dpanici: True, but in this case the reader will assume that the 1e-3 is now used for the last 2 equilibria so I need to check that the 3 eq steps are 1e-2, 1e-3, 1e-3

in the string comparison part of the test I then check that the line itself only has 2 elements, 1e-2 and 1e-3

f0uriest (approved):

desc/io/input_reader.py
```diff
@@ -1361,12 +1364,12 @@ def parse_vmec_inputs(vmec_fname, threshold=0):  # noqa: C901 - FIXME: simplify
                 inputs_arr[i]["L_grid"] = 2 * inputs_arr[i]["L"]
                 inputs_arr[i]["N"] = 0
                 inputs_arr[i]["N_grid"] = 0
-                inputs_arr[i]["pres_ratio"] = 0
+                inputs_arr[i]["pres_ratio"] = 0.0
                 inputs_arr[i]["curr_ratio"] = 0
```
ddudt, line 1368: also change this to a float?

tests/test_input_output.py
ddudt: I would prefer if `None` got replaced with the default tolerance, but since those are set by the Optimizer that isn't really possible.

ddudt (approved):

PR #415, opened 2023-01-19
Speedup qs Boozer plot functions

### LINKED ISSUE #414: Slow plot_qs_error possibly due to ptolemy transform with default M_booz, N_booz
Plot_qs_error on a L=M=9, N=10 equilibrium is taking quite a while to run when the `fB=True` flag is given, and it seems a lot of this time is being spent in the ptolemy transform part of the function. Could be worth taking a look at the ptolemy_linear_transform function to see if it can be optimized

dpanici:
- make plot-qs-error faster by only calculating linear transform matrix once
 - makes same change for plot_boozer_modes

speeds both up by like 70% (based off the testing suite)

resolves #414

### DIFF
--- a/desc/plotting.py
+++ b/desc/plotting.py
@@ -2170,7 +2170,8 @@ def plot_boozer_modes(
             "|B|_mn", eq=eq, grid=grid, M_booz=M_booz, N_booz=N_booz
         )
         data = eq.compute("|B|_mn", grid=grid, transforms=transforms)
-        matrix, modes = ptolemy_linear_transform(transforms["B"].basis)
+        if i == 0:
+            matrix, modes = ptolemy_linear_transform(transforms["B"].basis)
         b_mn = np.atleast_2d(matrix @ data["|B|_mn"])
         B_mn = np.vstack((B_mn, b_mn)) if B_mn.size else b_mn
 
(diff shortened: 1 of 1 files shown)


dpanici:
Also removes unused imports in plotting.py, . this should've been caught by the flake8 check in an earlier PR I thought... unless my if statement changed things somehow

f0uriest (approved):

f0uriest:
Would this same thing help `plot_boozer_modes`?

f0uriest:
and @ddudt in general is there a faster/vectorized way to do the ptolemy transforms? I feel like creating the matrix should just be a couple of `np.where`s, and then we could use that matrix in `ptolemy_identity_fwd` and `ptolemy_identity_rev`

ddudt (approved):

ddudt:
> and @ddudt in general is there a faster/vectorized way to do the ptolemy transforms? I feel like creating the matrix should just be a couple of `np.where`s, and then we could use that matrix in `ptolemy_identity_fwd` and `ptolemy_identity_rev`

It is a bunch of `np.where`s, but we have to loop through each mode in the basis and repeat the search at each iteration. I tried to find some sort of pattern to the matrix as a shortcut, but I couldn't come up with something that generalized well. There almost certainly is a better/faster way to do it, I just couldn't find it and it figured it wasn't worth spending too much time on.

PR #417, opened 2023-01-23
Refactor ptolemy matrix to avoid loops, allow custom mode order

### LINKED ISSUE #416: Refactor ptolemy_linear_transform to be faster
related to #414 , can probably refactor the underlying function to be faster by vectorizing the matrix creation

f0uriest:
Resolves #416 

Avoids looping over the whole matrix by using a pattern when m,n !=0 and then some hardcoded stuff for the edge cases. Also allows a custom double angle `modes` array, allowing for different ordering.

Also refactors `ptolemy_identity_*` to use the transform matrix instead of loops.
Keeps existing API for `ptolemy_identity_*` to avoid rewriting a bunch of the VMEC conversion stuff.

### DIFF
--- a/desc/objectives/_qs.py
+++ b/desc/objectives/_qs.py
@@ -127,7 +127,7 @@ class QuasisymmetryBoozer(_Objective):
             N_booz=self.N_booz,
         )
         self._matrix, self._modes, self._idx = ptolemy_linear_transform(
-            self._transforms["B"].basis, self.helicity
+            self._transforms["B"].basis, helicity=self.helicity
         )
 
         timer.stop("Precomputing transforms")
--- a/desc/objectives/_qs.py
+++ b/desc/objectives/_qs.py
@@ -127,7 +127,9 @@ class QuasisymmetryBoozer(_Objective):
             N_booz=self.N_booz,
         )
         self._matrix, self._modes, self._idx = ptolemy_linear_transform(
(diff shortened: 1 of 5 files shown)


ddudt (dismissed):
Approving, assuming the tests pass.

dpanici:
Comment on the math that lets us do this transformation as we forgot one time and it was embarrassing to remember it:

VMEC double angle convention is like :
$A cos(m\theta - n\zeta) + B cos(m\theta + n\zeta)$

Using Ptolemy identity:  
$A cos(m\theta) cos(n\zeta) + A sin(m\theta)sin(n\zeta) + B cos(m\theta)cos(n\zeta) - B sin(m\theta)sin(n\zeta)$

Collect like terms:  
$(A+B) cos(m\theta) cos(n\zeta)  + (A - B) sin(m\theta)sin(n\zeta)$

DESC's convention is double-term like the above lines, so the corresponding DESC coefficients $C,D$ are:
$C cos(m\theta) cos(n\zeta)  + D sin(m\theta)sin(n\zeta) =(A+B) cos(m\theta) cos(n\zeta)  + (A - B) sin(m\theta)sin(n\zeta)$

yielding
$C = A+B$
$D = A - B$

Then rearranging for what the VMEC coefficients a,b are in terms of the DESC coefficients yields
$A = (C+D)/2$
$B = (C-D)/2$

dpanici (approved):

PR #423, opened 2023-01-27
Bootstrap current

landreman:
This PR introduces two main components. The first is a compute function for the bootstrap current, `"<J*B> Redl"`, using the formula by Redl et al, Physics of Plasmas 28, 022502 (2021). This formula for the bootstrap current is valid in axisymmetry, quasi-axisymmetry, and quasi-helical symmetry, but not in other stellarators. The bootstrap calculation uses the new kinetic profiles (electron density, electron temperature, ion temperature, and atomic number) of `Equilibrium` objects. The 2nd new component is an `Objective`, presently named `BootstrapRedlConsistency`, for optimizing the current or iota profile to be consistent with the Redl formula. The objective is essentially the same expression used in Landreman Buller Drevlak, Physics of Plasmas 29, 082501 (2022).

For examples of using the new objective, see the last two tests in `tests/test_bootstrap.py`. The basic idea is to include `BootstrapRedlConsistency(helicity=(1,0))` or `BootstrapRedlConsistency(helicity=(1,NFP))` in your objective, while removing constraints on iota or current. When optimizing prescribed-iota equilibria, you should exclude `FixIota()` from the constraints. When optimizing a prescribed-current equilibrium, you should typically still include `FixCurrent(indices=[0])` in the constraints, so the current is mostly allowed to vary, but it still goes to 0 on axis as it should. When optimizing prescribed-current equilibria, it seems important to set `gtol=0` in the optimizer to prevent it from terminating early, I think because the current parameters in SI units are >> 1.

For some of the tests, in which the bootstrap current is compared to full drift-kinetic calculations with the SFINCS code, DESC input and output files for the relevant configurations were added to `tests/inputs`. Hope this is ok.

As this is my first significant PR, just let me know if there's anything I should pay attention to. For instance I wasn't sure what to pay attention to in the output of `pylint` and `flake8`.


tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+                "R": R,
+                "iota": iota,
+                "epsilon": epsilon,
+                "f_t": f_t,
+                "psi_edge": psi_edge,
+            }
+            J_dot_B_data = j_dot_B_Redl(geom_data, ne, Te, Ti, Zeff, helicity_N)
+
+            # Change False to True in the next line to plot the data for debugging.
+            if False:
```
f0uriest, line 455: could have this use `pytest-mpl` to benchmark the plot against a known correct one?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+                    )
+
+                integral = quad(integrand, 0, 1 / Bmax[jr])
+                f_t[jr] = 1 - 0.75 * fsa_B2[jr] * integral[0]
+            np.testing.assert_allclose(
+                f_t_data["f_t"][1:], f_t[1:], rtol=0.001, atol=0.001
+            )
+
+            # Use True in this "if" statement to plot extra info:
+            if False:
```
f0uriest, line 161: same comment about `pytest-mpl`

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+                    "psi_edge": psi_edge,
+                }
+                J_dot_B_data = j_dot_B_Redl(geom_data, ne, Te, Ti, Zeff, helicity_N)
+
+                L31s[j_nu_star, :] = J_dot_B_data["L31"]
+                L32s[j_nu_star, :] = J_dot_B_data["L32"]
+                alphas[j_nu_star, :] = J_dot_B_data["alpha"]
+                nu_e_stars[j_nu_star, :] = J_dot_B_data["nu_e_star"]
+                nu_i_stars[j_nu_star, :] = J_dot_B_data["nu_i_star"]
+                # np.testing.assert_allclose(J_dot_B_data["nu_e_star"], target_nu_e_star)
```
f0uriest, line 593: are these needed?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+                # np.testing.assert_allclose(J_dot_B_data["nu_i_star"], target_nu_i_star)
+
+            # Change False to True in the next line to plot the data for debugging.
+            if False:
```
f0uriest, line 597: same comment about `pytest-mpl`

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+                    test(grid_type, kwargs, L_factor, M_factor, N_factor, rho_exponent)
+                )
+
+        np.testing.assert_allclose(np.array(results), 0.5, rtol=2e-4)
+
+    @pytest.mark.unit
+    def test_BootstrapRedlConsistency_resolution(self):
+        """Confirm that the objective function is ~independent of grid resolution."""
+        helicity = (1, 0)
+        filename = ".//tests//inputs//DSHAPE_output_saved_without_current.h5"
```
f0uriest, line 1091: might be better to use the `DSHAPE` test fixture here, which is re-run each time to make sure changes to the rest of the code don't break the bootstrap stuff.

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+
+        for grid_type, kwargs in zip(grid_types, kwargss):
+            for L, M, N in zip(Ls, Ms, Ns):
+                results.append(test(grid_type, kwargs, L, M, N))
+
+        results = np.array(results)
+        np.testing.assert_allclose(results, np.mean(results), rtol=0.02)
+
+    # MJL: Should this test be decorated with @pytest.mark.slow ?
+    @pytest.mark.unit
```
f0uriest, line 1160: this should probably be marked as a `regression` test. We try to limit the number of full solve/optimizations we have to do for unit tests.

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+
+        eq.solve(
+            verbose=3,
+            ftol=1e-8,
+            constraints=get_fixed_boundary_constraints(kinetic=True),
+            optimizer=Optimizer("lsq-exact"),
+            objective=ObjectiveFunction(objectives=ForceBalance()),
+        )
+
+        eq.save("test_bootstrap_consistency_iota_initial.h5")
```
f0uriest, line 1220: Should preferably save to a temporary directory if needed (using `TmpDir` fixture) so we don't generate a bunch of local files when running tests

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+        )
+        eq, _ = eq.optimize(
+            verbose=3,
+            objective=objective,
+            constraints=constraints,
+            optimizer=Optimizer("scipy-trf"),
+            ftol=1e-6,
+        )
+
+        eq.save("test_bootstrap_consistency_iota_final.h5")
```
f0uriest, line 1257: same note about saving to temp directory

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+        J_dot_B_Redl = compress(grid, data["<J*B> Redl"])
+
+        assert np.max(J_dot_B_MHD) < 4e5
+        assert np.max(J_dot_B_MHD) > 0
+        assert np.min(J_dot_B_MHD) < -5.1e6
+        assert np.min(J_dot_B_MHD) > -5.4e6
+        np.testing.assert_allclose(J_dot_B_MHD, J_dot_B_Redl, atol=5e5)
+
+    # MJL: Should this test be decorated with @pytest.mark.slow ?
+    @pytest.mark.unit
```
f0uriest, line 1276: same note about making this a regression test

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+
+        eq.solve(
+            verbose=3,
+            ftol=1e-8,
+            constraints=get_fixed_boundary_constraints(kinetic=True, iota=False),
+            optimizer=Optimizer("lsq-exact"),
+            objective=ObjectiveFunction(objectives=ForceBalance()),
+        )
+
+        eq.save("test_bootstrap_consistency_current_initial.h5")
```
f0uriest, line 1335: same note about saving to temp directory

tests/test_bootstrap.py
```diff
@@ -0,0 +1,1390 @@
...
+        eq, _ = eq.optimize(
+            verbose=3,
+            objective=objective,
+            constraints=constraints,
+            optimizer=Optimizer("scipy-trf"),
+            ftol=1e-6,
+            gtol=0,  # It is critical to set gtol=0 when optimizing current profile!
+        )
+
+        eq.save("test_bootstrap_consistency_current_final.h5")
```
f0uriest, line 1374: same note about saving to temp directory

f0uriest (changes requested):
Cleaned up some of the formatting stuff and left comments/questions on a bunch of stuff. Just a high level pass for now, I'll go over it in more detail next week.

tests/test_bootstrap.py
landreman: This has been changed to `regression`.

tests/test_bootstrap.py
landreman: This has been changed to `regression`.

tests/test_bootstrap.py
landreman: Now `TmpDir` is used.

tests/test_bootstrap.py
landreman: Now `TmpDir` is used.

tests/test_bootstrap.py
landreman: Now `TmpDir` is used.

tests/test_bootstrap.py
landreman: Now `TmpDir` is used.

tests/test_bootstrap.py
landreman: These asserts are now uncommented.

tests/test_bootstrap.py
landreman: I attempted to address this, copying the pattern from the other test files.

desc/compute/_bootstrap.py
```diff
@@ -0,0 +1,504 @@
...
+            denominator=denominator,
+        )
+        return lambda_integral + lambda_weights[jlambda] * lambd[jlambda] / (
+            Bmax * Bmax * compress(grid, flux_surf_avg_term)
+        )
+
+    lambda_integral = fori_loop(0, n_gauss, body_fun, jnp.zeros(grid.num_rho))
+
+    f_t = 1 - 0.75 * fsa_B2 * lambda_integral
+    f_t_data = {
```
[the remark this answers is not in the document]
landreman: Yeah, I was debating whether to add the trapped fraction, Bmax, etc to the data index. Here's the reason it's not done at the moment. There are 2 ways I can think of to compute these quantities for the Redl bootstrap current formula when a config is not perfectly quasisymmetric. In method 1, the one implemented in this PR, Bmax, Bmin, and f_t are computed from the full not-quite-QS field. In method 2, which is not in this PR but which I'm hoping to try soon, the B field is transformed to Boozer coordinates and non-QS modes are filtered out before evaluating Bmax, Bmin, and f_t. I expect method 2 will be more accurate but slower / more memory needed, but haven't had a chance to check. Methods 1 and 2 would re-use all the code in `trapped_fraction()`, but use different input values for `modB` and `sqrt_g`, and give different answers for `f_t`, `Bmax`, etc. 

Some possible ways to proceed:

- Leave things as they are for now. I can try out Method 2 in a branch, and see which of the two methods works better. Then we re-assess whether to refactor `trapped_fraction()` into several regular `compute` functions.
- I could refactor `trapped_fraction()` into several regular `compute` functions now. When I try out Method 2 (in a branch), there might be a little bit of repeated code, and there would be 2 variants of these quantities in the data index, e.g. `f_t` and `f_t Boozer filtered`, `Bmax` and `Bmax Boozer filtered`, etc.
- Other ideas welcome too.

desc/objectives/_bootstrap.py
```diff
@@ -0,0 +1,216 @@
...
+            self.helicity[1] == 0 or abs(self.helicity[1]) == eq.NFP
+        ), "Helicity toroidal mode number should be 0 (QA) or +/- NFP (QH)"
+        self._dim_f = self.grid.num_rho
+        self._data_keys = ["<J*B>", "<J*B> Redl", "rho"]
+        self._args = get_params(self._data_keys)
+
+        # Try to catch cases in which density or temperatures are specified in the
+        # wrong units. Densities should be ~ 10^20, temperatures are ~ 10^3.
+        rho = eq.compute("rho", grid=self.grid)["rho"]
+        if jnp.any(eq.electron_temperature(rho) > 50e3):
```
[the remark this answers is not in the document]
landreman: A few lines above, some checks are now done to ensure the equilibrium has kinetic profiles.

desc/objectives/_bootstrap.py
```diff
@@ -0,0 +1,216 @@
...
+    each weighted by the square root of the denominator.
+
+    Parameters
+    ----------
+    helicity : 2-element tuple
+        First entry must be 1. Second entry is the toroidal mode number of
+        quasisymmetry, used for evaluating the Redl bootstrap current
+        formula. Set to 0 for axisymmetry or quasi-axisymmetry; set to +/- NFP for
+        quasi-helical symmetry.
+    rho_exponent: float
```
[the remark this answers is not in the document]
landreman: Sure, `rho_exponent` has now been removed.

desc/compute/_bootstrap.py
```diff
@@ -0,0 +1,504 @@
...
+        "dnds_term",
+        "dTeds_term",
+        "dTids_term",
+    ]
+    J_dot_B_data = geom_data.copy()
+    for v in variables:
+        J_dot_B_data[v] = eval(v)
+    J_dot_B_data["<J*B>"] = J_dot_B
+
+    if plot:
```
f0uriest, line 389: could this be moved to a standalone function in `desc.plotting`?

desc/compute/_bootstrap.py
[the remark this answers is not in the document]
f0uriest: I would vote for the 2nd option. If the boozer filtered method does seem significantly better (and not too slow) we could refactor things to always use that, otherwise I think having 2 versions could be useful. My hunch is that the boozer filtering will be quite a bit slower so in many cases we may still want to use the current implementation.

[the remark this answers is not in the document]
landreman: Sounds good. This function has now been refactored and split up into standard compute functions, so `trapped fraction` and intermediate quantities are all added to the data index.

tests/test_bootstrap.py
landreman: I haven't used `pytest-mpl` before - in your experience, do the comparisons with saved reference figures work reliably across different OSs and versions of matplotlib? I'm happy to add a `pytest-mpl` check here, or alternatively `np.testing.assert_allclose` could be used to compare the computed data to reference data since that would be 100% matplotlib-version-independent. What would you prefer?

desc/compute/_bootstrap.py
```diff
@@ -0,0 +1,504 @@
...
+    f_t = geom_data["f_t"]
+
+    if Zeff is None:
+        Zeff = PowerSeriesProfile(1.0, modes=[0])
+    if not isinstance(Zeff, Profile):
+        # Zeff is presumably a number. Convert it to a constant profile.
+        Zeff = PowerSeriesProfile([Zeff], modes=[0])
+
+    # Evaluate profiles on the grid:
+    ne_rho = ne(rho)
```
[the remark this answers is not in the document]
landreman: This has now been addressed - ne, Te, Ti, and Zeff are specified as data dependencies instead of evaluating the profiles.

desc/compute/_bootstrap.py
```diff
@@ -0,0 +1,504 @@
...
+        "<1/B>"
+    ]
+    geom_data["psi_edge"] = params["Psi"] / (2 * jnp.pi)
+
+    ne = profiles["electron_density"]
+    Te = profiles["electron_temperature"]
+    Ti = profiles["ion_temperature"]
+    Zeff = profiles["atomic_number"]
+    # The "backup" PowerSeriesProfiles that follow here are necessary
+    # for test_compute_funs.py::test_compute_everything
```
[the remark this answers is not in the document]
landreman: I think this may be a non-issue now that ne/Te/Ti/Zeff are obtained using data dependencies - `test_compute_everything` seems to work fine for the bootstrap compute function.

tests/test_bootstrap.py
f0uriest: It depends on what tolerances are used for the plot comparison, but we've found it to perform reasonably well across different versions, though we still pin the matplotlib version on CI for consistency. If it's not something that will be plotted regularly than `np.testing.assert_allclose` might make more sense, we mostly use `pytest-mpl` for the routines in `desc.plotting` to ensure that they look the way we think they should.

desc/compute/_bootstrap.py
landreman: For simplicity I've removed this plotting section. I haven't used it lately, and probably others wouldn't use it enough to be worth maintaining it.

f0uriest:
Looks pretty good, only outstanding thing I flagged is the normalization in the objective function. Would also be good to have @ddudt and @dpanici take a look before merging.

tests/test_bootstrap.py
landreman: `pytest-mpl` is now used here, and for the other similar spots in `test_bootstrap.py` you flagged, so there are no more `if False:` statements.

tests/test_bootstrap.py
landreman: `pytest-mpl` is now used here.

tests/test_bootstrap.py
landreman: `pytest-mpl` is now used here.

desc/objectives/_bootstrap.py
```diff
@@ -0,0 +1,216 @@
...
+    slightly generalized to allow different radial weighting, and with a factor of
+    1/2 for consistency with desc conventions.
+
+    f_{boot} = numerator / denominator
+
+    where
+
+    numerator = (1/2) \int_0^1 d\rho \rho^p [<J \cdot B>_{MHD} - <J \cdot B>_{Redl}]^2
+
+    denominator = \int_0^1 d\rho \rho^p [<J \cdot B>_{MHD} + <J \cdot B>_{Redl}]^2
```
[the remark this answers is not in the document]
f0uriest: Not totally sure what you mean by non-standard? I remember I went back and forth on this for a while and convinced myself that its correct the way it is but I'm open to being wrong. 

My reasoning was that the individual unweighted residual $f_i$ (ie $J_{mhd} - J_{redl}$ or similar) would be independent of the grid resolution, so the total sum of squares residual $F = \sum_i f_i^2$ would scale like $N$ for $N$ nodes. 

The weighted sum of squares is $F = 1/2 \sum_i (w_i f_i / n_i)^2$ where $f_i$ are the residuals in SI units, $n_i$ is a normalization factor, and $w_i$ is the weight. The weights should scale like $V/N$ where $V$ is the total volume, and $N$ is the total number of nodes.

We want $F$ to be approximately invariant under different grid discretizations. For simplicity, assume all the residuals/weights/normalizations are the same, and a total of $N$ nodes

Then $F = 1/2 N (w f / n)^2 = 1/2 N (V/N f / n)^2 = 1/2 V^2 f^2 / (N n^2) $

For $F$ to be invariant under changes in $N$, we need $n$ to depend on $N$ as $1/\sqrt{N}$, hence dividing through by `sqrt(self._dim_f)`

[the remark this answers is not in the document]
landreman: Agreed, the present normalizations give results that are ~independent of grid resolution. I'm thinking though that it is satisfying when the scalar objective is a discrete approximation of the integral ½∫(func - target)² dρ. This isn't necessary, but it's convenient for communicating how exactly the objective is defined. Presently, the quadrature weights are applied before squaring, so for grids with nonuniform radial weights, the scalar objective differs from the rho integral. Here's an example, using `ToroidalCurrent`:
~~~~python
import numpy as np
from desc.objectives import *
from desc.grid import *
from desc.profiles import *
from desc.objectives.normalization import compute_scaling_factors

eq = Equilibrium(iota=PowerSeriesProfile([0.95, 0, -0.7]))
grid = QuadratureGrid(L=100, M=100, N=0)

obj = ObjectiveFunction(ToroidalCurrent(grid=grid), eq)
scalar_obj = obj.compute_scalar(obj.x(eq))

current = eq.compute("current", grid=grid)["current"]
scales = compute_scaling_factors(eq)
# Evaluate ½∫(func - target)² dρ:
integral = 0.5 * sum(grid.weights * (current / scales["I"])**2) / (4 * np.pi**2)

print("     scalar objective:", scalar_obj)
print("½∫(func - target)² dρ:", integral)
~~~~
When run with master, the objective and integral differ:
~~~~
     scalar objective: 0.0005487245134347142
½∫(func - target)² dρ: 0.0004476907631056368
~~~~
With the modified weights [shown here](https://github.com/PlasmaControl/DESC/compare/ml/normalizations), where you multiply by the sqrt of the weights before squaring, the objective can be made to match the rho integral:
~~~~
     scalar objective: 0.00044769076310562615
½∫(func - target)² dρ: 0.0004476907631056368
~~~~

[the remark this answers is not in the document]
f0uriest: Ahhh I see what you mean. That seems like a better option. We should change that in the other profile objectives as well.

ddudt (commented):
Can any of the plots generated in these new tests be turned into more general plotting routines for `desc/plotting.py`?

desc/objectives/_bootstrap.py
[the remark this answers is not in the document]
ddudt: I support this change

desc/compute/_field.py
```diff
@@ -2753,3 +2806,75 @@ def _B_dot_gradB_z(params, transforms, profiles, data, **kwargs):
...
+def _min_tz_modB(params, transforms, profiles, data, **kwargs):
+    data["min_tz |B|"] = expand(
+        transforms["grid"], surface_min(transforms["grid"], data["|B|"])
+    )
+    return data
+
+
+@register_compute_fun(
+    name="effective r/R0",
+    label="(r / R_0)_{effective}",
```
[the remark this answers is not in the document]
f0uriest: that was my interpretation

desc/compute/_field.py
```diff
@@ -2753,3 +2806,75 @@ def _B_dot_gradB_z(params, transforms, profiles, data, **kwargs):
...
+    units="~",
+    units_long="None",
+    description="Effective local inverse aspect ratio, based on max and min |B|",
+    dim=1,
```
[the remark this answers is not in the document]
f0uriest: I think having it as a flux function makes sense, since it is a ~local quantity

desc/compute/utils.py
```diff
@@ -626,3 +636,63 @@ def surface_averages(
...
+        Quantity to find max.
+    surface_label : str
+        The surface label of rho, theta, or zeta to compute integration over.
+
+    Returns
+    -------
+    maxs : ndarray
+        Maximum of x over each surface in grid.
+
+    """
```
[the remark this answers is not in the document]
f0uriest: done

desc/compute/_equil.py
```diff
@@ -438,5 +461,5 @@ def _W(params, transforms, profiles, data, **kwargs):
     data=["W_p", "W_B"],
 )
 def _beta_vol(params, transforms, profiles, data, **kwargs):
-    data["<beta>_vol"] = data["W_p"] / data["W_B"]
+    data["<beta>_vol"] = -data["W_p"] / data["W_B"]
```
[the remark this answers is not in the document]
f0uriest: done

desc/compute/_bootstrap.py
```diff
@@ -0,0 +1,399 @@
...
+        )
+
+    lambda_integral = fori_loop(0, n_gauss, body_fun, jnp.zeros(grid.num_rho))
+
+    trapped_fraction = 1 - 0.75 * compress(grid, data["<B^2>"]) * lambda_integral
+    data["trapped fraction"] = expand(grid, trapped_fraction)
+    return data
+
+
+def j_dot_B_Redl(
```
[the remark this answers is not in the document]
landreman: There are a couple reasons `j_dot_B_Redl()` was separated out from `_compute_J_dot_B_Redl()`. First, there is that 2nd method for computing Bmin/Bmax/trapped_fraction based on filtering the Boozer spectrum, not yet coded up. That 2nd method would re-use all the code in `j_dot_B_Redl()` but not the code in `_compute_J_dot_B_Redl()`. Second, separating the two functions made it easier to do some tests involving prescribed `trapped fraction` data rather than an actual Equilibrium, specifically `test_Redl_second_pass` and `test_Redl_figures_2_3` and `test_Redl_figures_4_5`. (One way around this could be to directly call the compute functions in the tests, as in `tests/test_bootstrap.py:trapped_fraction()`).

desc/compute/_field.py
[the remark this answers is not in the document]
landreman: Right, `r` here is the local minor radius of the (typically interior) flux surface, rather than the minor radius `a` of the boundary.

desc/compute/_field.py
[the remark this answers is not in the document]
landreman: Right, this quantity is different on different flux surfaces, rather than a global dim=0 quantity.

desc/compute/_bootstrap.py
```diff
@@ -0,0 +1,399 @@
...
+    n_gauss = kwargs.get("n_gauss", 20)
+    base_nodes, base_weights = roots_legendre(n_gauss)
+    # Rescale for integration on [0, 1], not [-1, 1]:
+    lambd = (base_nodes + 1) * 0.5
+    lambda_weights = base_weights * 0.5
+
+    grid = transforms["grid"]
+    Bmax = data["max_tz |B|"]
+    modB_over_Bmax = data["|B|"] / Bmax
+    sqrt_g = jnp.abs(data["sqrt(g)"])
```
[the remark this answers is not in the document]
landreman: This abs() has now been removed.

desc/objectives/_bootstrap.py
```diff
@@ -0,0 +1,203 @@
...
+                "Electron temperature is surprisingly high. It should have units of eV"
+            )
+        if jnp.any(eq.ion_temperature(rho) > 50e3):
+            warnings.warn(
+                "Ion temperature is surprisingly high. It should have units of eV"
+            )
+        # Profiles may go to 0 at rho=1, so exclude the last few grid points from lower
+        # bounds:
+        rho = rho[rho < 0.85]
+        if jnp.any(eq.electron_density(rho) < 1e17):
```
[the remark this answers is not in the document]
landreman: A upper bound check on the density has now been added.

landreman:
> Can any of the plots generated in these new tests be turned into more general plotting routines for `desc/plotting.py`?

I think the new `mpl_image_compare` figures are either pretty specific to testing, so not very likely that users would make variations of them, or basically already covered by the existing `plot_1d()` function. The figures from `test_Redl_figures_2_3`, `test_Redl_figures_4_5`, `test_Redl_sfincs_QA`, and `test_Redl_sfincs_QH` are all for verifying that figures from publications can be reproduced. For plotting `trapped fraction` or `<J*B> Redl` as in `test_trapped_fraction_Kim`, `test_Redl_sfincs_QA`, or `test_Redl_sfincs_QH`, a user could use `plot_1d()` (and SFINCS data as in the latter 2 plots would not be available.) One new plot routine we could contemplate would be a figure with radial profiles of `<J*B>` and `<J*B> Redl` on the same axis, to check self-consistency of the current profile, though that too is perhaps sufficiently covered already by `plot_1d()`.

desc/objectives/_bootstrap.py
```diff
@@ -0,0 +1,261 @@
...
+        self._helicity = helicity
+        if hasattr(self, "_print_value_fmt"):
+            units = "(T)"
+            self._print_value_fmt = (
+                "Quasi-symmetry ({},{}) Boozer error: ".format(
+                    self.helicity[0], self.helicity[1]
+                )
+                + "{:10.3e} "
+                + units
+            )
```
landreman, line 261: For this setter that you added, I wasn't sure if you would rather have this section deleted vs edited to fix units and print bootstrap instead of QS/Boozer.

ddudt: Good catch, I forget to delete that when I copied it over from the QS objectives.

desc/compute/_field.py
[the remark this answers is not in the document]
ddudt: OK in that case I like using `r`.

f0uriest (approved):
Looks good to go as far as I can see. Would be good to let @ddudt and @dpanici take one final look. @landreman Once they give the ok feel free to merge whenever.

ddudt (approved):

f0uriest (approved):

desc/objectives/_qs.py
```diff
@@ -205,6 +208,7 @@ def helicity(self, helicity):
                 + "{:10.3e} "
                 + units
             )
+        warnings.warn("Re-build objective after changing the helicity!")
```
dpanici, line 211: Is it worth just forcing a rebuild here after the helicity is changed? Or better yet, adding
```python
if self._helicity != helicity 
self._built = False
```
here before changing `self._helicity` so that if the helicity was changed, the objective is no longer built? I think this would be a good change to add

dpanici (changes requested):
My bad for the delay in reviewing, I think making the QSBoozer's flag`self._built=False` if the helicity is changed from what it previously was is a good change to make, other than that it all looks good

desc/objectives/_qs.py
landreman: This has now been implemented for all the Objectives that use helicity: `BootstrapRedlConsistency`, `QuasisymmetryBoozer`, and `QuasisymmetryTwoTerm`.

f0uriest (approved):

ddudt (approved):

dpanici (approved):

landreman:
Thanks everyone for your help with this.

PR #425, opened 2023-01-31
Add compute functions for field line quantities

### LINKED ISSUE #422: Add field line coordinates to compute functions
- [ ] b (unit vector)
- [ ] n (unit vector)
- [ ] alpha (field line label)
- [ ] grad alpha
- [ ] kappa (field line curvature)
- [ ] sfl theta

f0uriest:
Resolves #422


desc/compute/_field.py
```diff
@@ -2753,3 +2782,60 @@ def _B_dot_gradB_z(params, transforms, profiles, data, **kwargs):
...
+    data["kappa"] = -(
+        cross(data["b"], mu_0 * data["J"] + cross(data["b"], data["grad(|B|)"])).T
+        / data["|B|"]
+    ).T
+    return data
+
+
+@register_compute_fun(
+    name="kappa_n",
+    label="\\kappa_n",
```
ddudt, line 2810: Would it be more clear to use `"\\kappa \\cdot \\hat{n}"`?

desc/compute/_field.py
```diff
@@ -2753,3 +2782,60 @@ def _B_dot_gradB_z(params, transforms, profiles, data, **kwargs):
...
+    data=["kappa", "n"],
+)
+def _kappa_n(params, transforms, profiles, data, **kwargs):
+    data["kappa_n"] = dot(data["kappa"], data["n"])
+    return data
+
+
+@register_compute_fun(
+    name="kappa_g",
+    label="\\kappa_g",
```
ddudt, line 2828: Similar comment as above. Not sure that people will know `\\kappa_g` means "geodesic curvature". But not a big deal either way (it only really affects plotting).

desc/compute/_profiles.py
```diff
@@ -133,6 +133,49 @@ def _gradpsi_mag(params, transforms, profiles, data, **kwargs):
...
+    ],
+)
+def _chi(params, transforms, profiles, data, **kwargs):
+    chi_r = compress(transforms["grid"], data["chi_r"], surface_label="rho")
+    drho = compress(
+        transforms["grid"], transforms["grid"].spacing[:, 0], surface_label="rho"
+    )
+    chi = jnp.cumsum(drho * chi_r)
+    data["chi"] = expand(transforms["grid"], chi, surface_label="rho")
+    return data
```
ddudt, line 176: In the cases where we know the iota profile as a power series, I think we can perform this integration exactly:
$\chi = \int_0^1 \chi' d\rho = \int_0^1 \psi' \iota d\rho = \psi(1) \iota(1) - \psi(0) \iota(0) - \int_0^1 \psi \iota' d\rho = \psi(1) \iota(1) - \psi(0) \iota(0) - \frac{\Psi}{2\pi} \int_0^1 \rho^2 \iota' d\rho$

But I'm not sure how that would generalize to other cases (like with a current profile input).

desc/compute/_core.py
```diff
@@ -78,6 +78,96 @@ def _zeta(params, transforms, profiles, data, **kwargs):
...
+    description="Field line label",
+    dim=1,
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["theta_sfl", "zeta", "iota"],
+)
+def _alpha(params, transforms, profiles, data, **kwargs):
+    data["alpha"] = data["theta_sfl"] - data["iota"] * data["zeta"]
```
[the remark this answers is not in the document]
ddudt: OK I've convinced myself why the equation is correct with the minus sign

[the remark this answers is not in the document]
f0uriest: As an explanation, we want $d\alpha = 0$ along a field line, and we also have $\iota = d\theta/d\zeta$ along a field line, which becomes $d\theta - \iota d\zeta = 0$, so then $\alpha = \theta - \iota \zeta$.

I think it might have been weird before the jacobian sign fix, because in that case there's an extra implicit minus sign on iota

desc/compute/_field.py
f0uriest: This seems to be fairly standard notation in books on differential geometry, and its also what's used in Helander 2014

desc/compute/_profiles.py
f0uriest: Yeah there is probably a way to do this more accurately for specific cases, but not in general (at least that is easy to implement). And ideally we want this to be general enough to handle all cases without having to know the underlying representation for iota or current.

desc/compute/_field.py
f0uriest: see prev comment

desc/compute/_core.py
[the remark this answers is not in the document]
ddudt: Thanks. What about my comment on alpha in different coordinate systems (namely Boozer coordinates)?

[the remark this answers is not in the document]
f0uriest: Shouldn't the field line label be independent of coordinates? As long as we all agree on theta=zeta=0 where alpha=0 I would think that changing coordinates shouldn't change the field line label.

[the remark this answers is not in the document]
f0uriest: $\alpha = \theta - \iota \zeta$

$\theta_B = \theta + \iota \nu$

$\zeta_B = \zeta + \nu$

$\alpha_B = \theta_B - \iota \zeta_B = \theta + \iota \nu - \iota \zeta - \iota \nu = \alpha$

ddudt (approved):

dpanici (approved):

PR #426, opened 2023-01-31
Improve handling of indefinite hessian matrices in fmintr

f0uriest:
The trust region subproblems used in `fmintr` require the Hessian matrix to be positive definite. In the case that it isn't, we perform a modified Cholesky factorization such that the factors give a positive definite approximation to the true Hessian. The previous method for doing this modified factorization, while robust, is quite slow, likely because it's not taking advantage of optimized low level BLAS stuff.

This PR updates the modified Cholesky routine to use a bisection search to find a small diagonal correction to ensure the factors are positive definite. This is much faster in practice, often only 2-3x more expensive that a regular un-modified Cholesky factorization, and the results seem to be about as good on most problems I've tested with.

Some benchmarking for matrix of size 8000x8000:
```
Old cholmod
2min 3s ± 2.15 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
New cholmod
6.25 s ± 119 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
Regular cholesky
1.83 s ± 67.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

### CODE REVIEW

File: desc/optimize/utils.py [hunk 1]
```diff
@@ -2,76 +2,97 @@
...
+        L = jnp.linalg.cholesky(A + alphas[kbest] * eye)
+        # check if it succeeded
+        isnan = jnp.any(jnp.isnan(L))
+        # adjust bounds for correction
+        klow = isnan * kbest + (1 - isnan) * klow
+        khigh = isnan * khigh + (1 - isnan) * kbest
+        kbest = (klow + khigh) // 2
+        # if it succeeded, mark it as the best so far
+        Lbest = cond(isnan, lambda _: Lbest, lambda _: L, 1)
+    return Lbest
```
ddudt, line 95: So this starts with the most conservative correction, tries all possible values for alpha (maxiter different values), and then returns the smallest correction that worked? 
Wouldn't it be better to start with the smallest value of alpha, and then return the first one that works? 
That way we won't always have to do `maxiter` iterations -- we could get lucky on the first or second try.

ddudt: On that same train of thought, this doesn't actually need to be a separate function from `chol(A)` if you start with trying alpha=0.

f0uriest: In that case in the worst case you need 1 extra attempted factorization, and in testing that seemed to be fairly common. If you start with the smallest one and it doesn't work, the next one you'd try would be the middle one, and if that works you'd try a quarter etc, so in almost all cases it will still take the full maxiter. Starting with the smallest one in the best case terminates after 1 iteration, in the worst case requires maxiter+1, and it seemed like the worst case was actually fairly common (because we're only searching over a log scale).

ddudt (approved):

PR #427, opened 2023-02-02
Dd/misc

### LINKED ISSUE #421: Update theory docs for positive Jacobian
Plots in theory docs have theta ccw, but DESC now uses cw theta to ensure a positive Jacobian. Should also confirm current definition of sfl theta

ddudt:
Documentation updates + other trivial changes. 

Resolves #421


f0uriest (dismissed):

docs/theory_general.rst
```diff
@@ -45,9 +45,14 @@ DESC solves for the map between the cylindrical and flux coordinate systems thro
 .. math::
   R(\rho, \theta, \zeta) \\ \\ Z(\rho, \theta, \zeta) \\ \\ \lambda(\rho, \theta, \zeta)
 
-It assumes the flux functions for the pressure :math:`p(\rho)` and rotational transform :math:`\iota(\rho)` profiles are given, in addition to the total toroidal flux through the plasma volume :math:`\psi_a`.
+It assumes the flux functions for the pressure profile :math:`p(\rho)` and either the rotational transform :math:`\iota(\rho)` or toroidal current :math:`I(\rho)` profile are given.
+Additionally, it needs the total toroidal magnetic flux through the plasma volume :math:`\Psi_a`, where :math:`\psi_N = \frac{\Psi}{2\pi\Psi_a}`.
```
dpanici, line 49: is this right?

I thought $\psi_N = \psi / \psi_a$ where $\psi_a = \Psi_a / 2\pi$ and $\Psi_a$ is the total enclosed toroidal flux, in Wb, and $\psi = \Psi / 2\pi$

so you do not need the $2\pi$ in this definition of $\psi_N$

dpanici: my gripe here is capital $\Psi$ is in Wb and not normalized by $2\pi$, but with your current eqn it tells me that $\Psi_a = \Psi / (2 \pi \psi_N)$, which says that at the LCFS where $psi_N = 1$, $Psi_a = \Psi/ 2\pi$ which is wrong, as it should just be $\Psi_a = \Psi$

dpanici (changes requested):
Change the psi equation to remove the errant $2\pi$

docs/theory_general.rst
ddudt: removed the $2\pi$ factor and added clarification

f0uriest (approved):

f0uriest (approved):

f0uriest (approved):

dpanici (approved):

PR #428, opened 2023-02-02
Clean up examples

### LINKED ISSUE #402: Check and update Existing examples and Add precise_qh and precise_qa scripts to docs and to examples

### LINKED ISSUE #412: Examples that need tweaking as default input file is not robust
based on running examples on CPU on Stellar, bolded ones I should change:

 - [ ]  **ATF**: surfaces no longer nested after last step, time: 3.56 min
 - [ ]  **ARIES-CS**: Seems to run ok, long though, time: 3.23 hours
 - [ ]  **CFQS**: surfaces no longer nested after 4th step, time: 18 min
 - DSHAPE: runs fine, time: 40 sec
 - DSHAPE_current: runs fine, time: 50 sec
 - ESTELL: runs fine, time: 13 min
 - HELIOTRON: runs fine, time: 10.4 min
 - [ ]  **HSX**: immediately goes unnested, due to only being 1 tep and probably the bdry is too strongly shaped. Also uses scipy-trf instead of lsq-exact
 - NCSX: runs fine, time: 41 min
 - QAS: runs fine, time:  8.5 min
 - SIMSOPT_QA: runs fine, time: 2.3 min
 - W7-X: on step 7, the radial spectral resolution is higher than the radial grid resolution, but this is because I switched it to ANSI... so maybe not a good representation of this example. ends up going unnested at last step but will re-run with current defaults
 - WISTELL-A: seems to run fine, time: 20.1 min

### LINKED ISSUE #430: Regenerate example equilibrira .h5 files with newer ones

f0uriest:
- Removes VMEC examples, as they were unused and should be supplanted by the in-development stellarator database
- Moved all DESC example inputs/outputs to `desc.examples`
- Makes `desc.examples.get` case insensitive.
- Adds precise QA + QH examples

Resolves #412 
Resolves #430 
Resolves #402


dpanici:
Other thing we can do is making sure that examples can run on a reasonable laptop without running into memory issues (so maybe having a lowres/highres version for some where the lowres we know can run with 16GB of RAM or something)

dpanici:
- [ ] Remove VMEC files
 - [ ] remove DESC files that are difficult to work with

desc/examples/precise_QA.py
```diff
@@ -0,0 +1,88 @@
...
+    RotationalTransform,
+)
+from desc.optimize import Optimizer
+
+# create initial equilibrium
+surf = FourierRZToroidalSurface(
+    R_lmn=[1, 0.166, 0.1],
+    Z_lmn=[-0.166, -0.1],
+    modes_R=[[0, 0], [1, 0], [0, 1]],
+    modes_Z=[[-1, 0], [0, -1]],
```
ddudt, line 32: We should be able to get this precise QA result starting from an axisymmetric initial equilibrium.

ddudt (commented):
So we can't get good solutions for HSX or CFQS? That's disappointing.

f0uriest:
> So we can't get good solutions for HSX or CFQS? That's disappointing.

Not really sure. The ones we had didn't converge very well and I didn't spend much time trying to fix them. At least with HSX I'm pretty sure its just a resolution issue. CFQS seemed to probably need smaller boundary perturbation steps. It didn't feel like having those was super necessary though.

desc/examples/precise_QA.py
f0uriest: I had tried that and it occasionally got stuck as a tokamak depending on how the objectives were weighted, starting from non-axisymmetry seemed to converge to something good more robustly regardless of weighting.

ddudt: Is there a set of weights that does make it converge? I thought it worked in the past. It would be nice to show we can reproduce the result starting from the same initial condition as SIMSOPT.

ddudt:
> > So we can't get good solutions for HSX or CFQS? That's disappointing.
> 
> Not really sure. The ones we had didn't converge very well and I didn't spend much time trying to fix them. At least with HSX I'm pretty sure its just a resolution issue. CFQS seemed to probably need smaller boundary perturbation steps. It didn't feel like having those was super necessary though.

OK. If we don't want to spend time on it now, we should at least make an issue to remember to add working examples eventually. These are the only QS stellarator experiments that exist so it would be nice to show we can represent them.

desc/examples/precise_QA.py
f0uriest: I'm having trouble finding a case that starts from axisymmetry and converges correctly. I remember we had similar issues in trying to perturb a tokamak into a stellarator to reduce current, it seems there's some deep local minimum at axisymmetry. 

Another possibility is that its an issue with vacuum tokamaks not really being well defined, so the initial equilibrium is weird in some way.

ddudt (approved):

dpanici (approved):

PR #431, opened 2023-02-06
Read VMEC Multi-line input files

### LINKED ISSUE #182: Handle Multi-line VMEC inputs when converting to a DESC input

dpanici:
- Adds capability of InputReader to read VMEC files that have inputs spanning multiple lines, i.e. 
```
NITER_ARRAY =  100 1000
                           10000
```
 - Also fixes bug when VMEC input file has duplicated lines, DESC now will just use the last duplicated line (which is what VMEC does)
 - resolves #182


desc/io/input_reader.py
```diff
@@ -834,7 +834,32 @@ def parse_vmec_inputs(vmec_fname, threshold=0):  # noqa: C901 - FIXME: simplify
...
+            elif line.find("&") != -1:  # only care about the INDATA section
+                break
+            if line.find("=") != -1:
+                vmeclines_no_multiline.append(line)
+            else:  # is a multi-line input,append the line to the previous
+                vmeclines_no_multiline[-1] += " " + line
+
+        # Some bool flags to ensure that if a line is duplicate in
+        # the VMEC file, we do not add it twice to the inputs
+        RAXIS_already_read = False
```
f0uriest, line 857: why just these keys? Are there other things that might be written twice? And do we want to throw an error in that case? IE, how do we know which one is the "right" axis?

desc/io/input_reader.py
```diff
@@ -834,7 +834,32 @@ def parse_vmec_inputs(vmec_fname, threshold=0):  # noqa: C901 - FIXME: simplify
         pres_scale = 1.0
         curr_tor = None
 
-        for line in vmec_file:
+        # Loop which makes multi-line inputs a single line
+        vmeclines_no_multiline = []
+
+        for line in vmec_file.readlines():
+            comment = line.find("!")
+            line = (line.strip() + " ")[0:comment].strip()
```
f0uriest, line 842: don't we want to completely ignore a line if it starts with `!`?

f0uriest (commented):
Main question is whether it would be better to just bite the bullet and convert the VMEC input reader stuff to use `f90nml`

dpanici:
- [ ] try f90nml again
- [ ] make += a = for Axis and Bdry coeffs

desc/io/input_reader.py
dpanici: this line does that if `!` is at the begining, and if it is later in the line it includes everything up until that point in the line then ignores the rest

dpanici:
> * [ ]  make += a = for Axis and Bdry coeffs

@f0uriest @ddudt do you remember why I had made this comment? I want to wrap this up but forget... guess that I should not leave stale PRs for 2 months before returning

dpanici:
this fix only helps, it is not a full f90nml change but that can be done another time. I think it is worth putting in as is now

f0uriest (approved):

PR #434, opened 2023-02-11
fix bug in scipy message handling which results in error (attempting …

dpanici:
…str + list) if scipy exits with a warning

`result["message"]` is a list, so must access its first index in order to get the string.


tests/test_optimizer.py
```diff
@@ -370,3 +373,52 @@ def test_maxiter_1_and_0_solve():
...
+            constraints=constraints,
+            objective=obj,
+            optimizer=opt,
+            ftol=1e-12,
+            xtol=1e-12,
+            gtol=1e-12,
+        )
+        # result["message"] is a list for scipy-trust-exact but a string for scipy-trf
+        # why?
+        assert "Maximum number of iterations has been exceeded" in result["message"][0]
```
ddudt, line 418: Figure this out before finishing this PR. We should have a consistent `result["message"]` type across all optimizers.

ddudt (changes requested):

f0uriest:
The issue here is how I had to override scipy's stopping criteria for the `minimize` routines. `message` in that case is a list just so it acts as a wider scoped variable that can be changed from within the callback function. the scipy least squares routines already have all the stopping criteria we want so I didn't have to do as many kludges there. I think I fixed this in #394 so I'll merge this into that.

PR #440, opened 2023-02-22
Ensure averages are performed on the correct grids in `eq.compute`

### LINKED ISSUE #438: Throw warning when computing FSA qtys with insufficient number of points
Just to let people know the results may be wrong if there's only 1 pt in the grid for example.

Not sure the best place is, easiest would probably be in compute.compute but could also move it lower to the surface average function itself so it only throws the error when it really means to?

f0uriest:
Previously, if the user requests data from `eq.compute` on a specified grid, that grid is also used to compute any volume or surface averages, leading to incorrect results in many cases.

This PR adds a pre-computation to `eq.compute` that checks if any of the quantities requested (or their dependencies) are volume or surface averages, and first performs those calculations on a full grid to ensure the correct result. The averaged qtys are then used as a seed for any further calculations needed on the user's originally specified grid.

Resolves #438


desc/compute/_profiles.py
```diff
@@ -52,7 +52,7 @@ def _psi_r(params, transforms, profiles, data, **kwargs):
     params=["Psi"],
     transforms={},
     profiles=[],
-    coordinates="",
+    coordinates="r",
```
unalmis, line 55: This is a constant, so it could? be either an empty string or all coordinates.

desc/equilibrium/configuration.py
```diff
@@ -1082,7 +1090,56 @@ def compute(
...
+            # these should all be 0d quantities so don't need to compress/expand
+            data0d = {key: val for key, val in data0d.items() if key in dep0d}
+            data.update(data0d)
+
+        if len(dep1d):
+            grid1d = LinearGrid(
+                rho=grid.nodes[grid.unique_rho_idx, 0],
+                M=self.M_grid,
+                N=self.N_grid,
+                NFP=self.NFP,
```
unalmis, line 1122: `sym=self.sym`?

desc/compute/_profiles.py
```diff
@@ -71,7 +71,7 @@ def _psi_rr(params, transforms, profiles, data, **kwargs):
...
```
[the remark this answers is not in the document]
f0uriest: My thinking was to make it the same as the other profile stuff since that's it's physical meaning

desc/compute/_profiles.py
f0uriest: Same as above

desc/equilibrium/configuration.py
f0uriest: Yeah good catch

f0uriest: Yeah good catch

ddudt (changes requested):
I think we should distinguish between quantities that are always 0 (`0`, `B^rho`, `psi_rr`, `psi_rrr`), and scalar quantities that need to be computed on the quadrature grid on a whole volume. 

I recommend using `coordinates="0"` for the former, and keeping `coordinates=""` or maybe using `coordinates="vol"` for the latter. In both cases they return a scalar value that can be broadcast to any other grid size. But the former will always be 0 (and therefore does not need a grid to be computed), while the latter will use a quadrature grid. 

Also it's odd to me that you had to regenerate the baseline images -- this shouldn't have changed any functionality, right?

f0uriest:
> I think we should distinguish between quantities that are always 0 (`0`, `B^rho`, `psi_rr`, `psi_rrr`), and scalar quantities that need to be computed on the quadrature grid on a whole volume.
> 
> I recommend using `coordinates="0"` for the former, and keeping `coordinates=""` or maybe using `coordinates="vol"` for the latter. In both cases they return a scalar value that can be broadcast to any other grid size. But the former will always be 0 (and therefore does not need a grid to be computed), while the latter will use a quadrature grid.

I thought it would be better to keep the stuff that's 0 marked as depending on either `r` or `rtz` as in general they do, it just so happens in our coordinate system or with our choice of psi they're constant. For example, if we wanted to do the RFP thing then `psi_rr` may no longer be 0, but it would still be a function of `r`. Note that in the current code, and this version, things that are `0` are still arrays of size `num_nodes` while things marked `coordinates=""` area true scalars in the data dictionary.


> Also it's odd to me that you had to regenerate the baseline images -- this shouldn't have changed any functionality, right?

It does change things slightly, as previously we fit iota to splines and used that profile, which will be slightly different from the true flux surface averaged iota that we now use. It also affects some of the normalizations for force, which I'm realizing now are slightly wrong (we had the wrong NFP for the quadrature grids)

tests/test_constrain_current.py
```diff
@@ -16,17 +16,35 @@ class _ExactValueProfile:
...
         self.eq = eq
         self.grid = grid
+        self.transforms = get_transforms("current_rr", eq=eq, grid=grid)
+        self.profiles = get_profiles("current_rr", eq=eq, grid=grid)
+        self.params = get_params("current_rr", eq=eq)
 
     def compute(self, params, dr, *args, **kwargs):
         if dr == 0:
             # returns the surface average of B_theta in amperes
-            return self.eq.compute("current", grid=self.grid)["current"]
```
unalmis, line 23: Good job catching this.

This test was failing on this PR before this last commit when `self.grid` was `ConcentricGrid`.  The assertion statement caught that iota on the 2 surfaces closest to the axis didn't match with the given power series.

 This was because the `profiles["current"].compute` call [here](https://github.com/PlasmaControl/DESC/blob/cb2313a7d03397a2f71904a64896da766643a2cc/desc/compute/_profiles.py#L556) was monkey patched to call this function. In this PR, `self.eq.compute("current", grid=self.grid)["current"]` now does a surface average on on a `LinearGrid` rather than the `self.grid` in the argument. So when computing `iota` in this test (before the last commit), we were doing a surface average on a `LinearGrid` as well as on `ConcentricGrid` [here](https://github.com/PlasmaControl/DESC/blob/cb2313a7d03397a2f71904a64896da766643a2cc/desc/compute/_profiles.py#L577).

Ideally this shouldn't matter because the surface average should yield the same result regardless of what grid is used.
But I think the fact that `ConcentricGrid` has few nodes on the surface closest to the axis was causing that discrepancy.

ddudt (dismissed):

ddudt:
> > I think we should distinguish between quantities that are always 0 (`0`, `B^rho`, `psi_rr`, `psi_rrr`), and scalar quantities that need to be computed on the quadrature grid on a whole volume.
> > I recommend using `coordinates="0"` for the former, and keeping `coordinates=""` or maybe using `coordinates="vol"` for the latter. In both cases they return a scalar value that can be broadcast to any other grid size. But the former will always be 0 (and therefore does not need a grid to be computed), while the latter will use a quadrature grid.
> 
> I thought it would be better to keep the stuff that's 0 marked as depending on either `r` or `rtz` as in general they do, it just so happens in our coordinate system or with our choice of psi they're constant. For example, if we wanted to do the RFP thing then `psi_rr` may no longer be 0, but it would still be a function of `r`. Note that in the current code, and this version, things that are `0` are still arrays of size `num_nodes` while things marked `coordinates=""` area true scalars in the data dictionary.
> 
> > Also it's odd to me that you had to regenerate the baseline images -- this shouldn't have changed any functionality, right?
> 
> It does change things slightly, as previously we fit iota to splines and used that profile, which will be slightly different from the true flux surface averaged iota that we now use. It also affects some of the normalizations for force, which I'm realizing now are slightly wrong (we had the wrong NFP for the quadrature grids)

OK, I'm fine with that. Thanks.

unalmis (dismissed):

unalmis (approved):

ddudt (approved):

tests/test_constrain_current.py
unalmis: @dpanici
@ddudt
Last week we had thought it was weird why a surface average on a linear grid and concentric grid were not exactly the same (not just different at the axis).
I think this is expected though.
The difference between them decreases with resolution increase, meaning the accuracy of the average being captured is dependent on how well midpoint rule integration approximates the true average, and the linear grid has more nodes.

```python
eq = get("W7-X")
L, M, N, NFP, sym = eq.L_grid, eq.M_grid, eq.N_grid, eq.NFP, eq.sym
cg = ConcentricGrid(L=L, M=M, N=N, NFP=NFP, sym=sym)
rho = np.copy(compress(cg, cg.nodes[:, 0]))
lg = LinearGrid(rho=rho, M=M, N=N, NFP=NFP, sym=sym)

def compute_iota(data, grid):
    current_term = (
        mu_0
        / (2 * np.pi)
        * data["current"]
        / data["psi_r"]
    )
    num = (
        data["lambda_z"] * data["g_tt"] - (1 + data["lambda_t"]) * data["g_tz"]
    ) / data["sqrt(g)"]
    den = data["g_tt"] / data["sqrt(g)"]
    num_avg = surface_averages(grid, num)
    den_avg = surface_averages(grid, den)
    return (current_term + num_avg) / den_avg

    
data_cg = eq.compute(["current", "psi_r", "lambda_z", "lambda_t", "g_tt", "g_tz", "sqrt(g)"], grid=cg)
data_lg = eq.compute(["current", "psi_r", "lambda_z", "lambda_t", "g_tt", "g_tz", "sqrt(g)"], grid=lg)

iota_cg = compress(cg, compute_iota(data_cg, cg))
iota_lg = compress(lg, compute_iota(data_lg, lg))
print("absolute difference")
print(abs(iota_cg - iota_lg))
print("relative difference")
print(abs(iota_cg - iota_lg) / abs(iota_lg))
```
```python
absolute difference
[4.99024289e-01 1.03736010e-03 1.00308820e-05 1.26244516e-06
 3.89644578e-07 8.73197681e-09 3.73933381e-08 2.29957960e-08
 5.23922110e-08 7.08192308e-08 3.95633758e-08 1.11696806e-08
 3.24810173e-08 1.07613550e-08 9.37128031e-09 1.99362553e-08
 0.00000000e+00]
relative difference
[5.82936911e-01 1.21172527e-03 1.17144310e-05 1.47355976e-06
 4.54329755e-07 1.01627698e-08 4.33882724e-08 2.65566947e-08
 6.00864063e-08 8.04441105e-08 4.43891265e-08 1.23499529e-08
 3.53475139e-08 1.15298979e-08 9.90430237e-09 2.08536925e-08
 0.00000000e+00]
```

f0uriest: The concentric grid and linear grid have the same nodes out the outermost surface, so they should agree there, but as you go in the concentric grid looses some nodes so we expect them to differ, which lines up with the values you got. The variation of the function over the surface also decreases as you go in, so its not that big a difference (relative error < 0.1% throughout most of the volume)

dpanici (approved):

PR #444, opened 2023-03-03
Update notebooks and add new tutorials for Toroidal Current Constraint and Solving from pyQSC initial Guess 

### LINKED ISSUE #441: Update the 00,01,02,03 tutorials to reflect new compute function
Currently the 01_Using_DESC_Inputs notebook has these lines

image
 where after it says 

>  Notice that the first derivatives with respect to $\rho$ of the pressure and rotational transform were also returned. Both profiles were provided as inputs when solving this equilibrium, so there is no computational burden to return the higher order derivatives too. Don't rely on this. In general, requesting a quantity will return all its _lower order derivatives_ too. If you want both `B` and `B_r`, you should call `eq.compute("B_r")`.

This is incorrect with the new compute function changes, as we now need to ask for "p","p_r" each if we want both of them in data.

dpanici:
- updates notebooks and fixes some types
 - add Toroidal current tutorial
 - add tutorial showing how to use the `from_near_axis` feature
 - fix broken examples link

Resolves #441


docs/notebooks/Toroidal_current_constraint.ipynb
```diff
@@ -0,0 +1,837 @@
+{
```
f0uriest, line 1: <p>We probably shouldn't do a vacuum tokamak since it's not really well defined. Maybe do a fixed nonzero current case? And then add 3d shaping, then perturb to zero current?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/444/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

f0uriest, line 1: <p>Should regenerate the current plot after PR #440</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/444/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='37'/>

desc/transform.py
```diff
@@ -57,7 +57,11 @@ def __init__(
         self._basis = basis
         self._rcond = rcond if rcond is not None else "auto"
 
-        if not (self.grid.NFP == self.basis.NFP) and grid.node_pattern != "custom":
+        if (
+            self.grid.num_zeta > 1
```
ddudt, line 61: Unequal NFP between the grid and basis can still cause problems for a single zeta coordinate, if the angle is not 0.

f0uriest (dismissed):

docs/notebooks/DESC_Solve_from_pyQSC.ipynb
```diff
@@ -0,0 +1,250 @@
...
+   "source": [
+    "qsc_eq = Qsc.from_paper(\"precise QA\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ed61eac3-54f8-49a2-bb46-23e73e07ea06",
+   "metadata": {},
+   "source": [
+    "Then, to make the DESC equilibrium, the `Equilibrium` class has a method `Equilibrium.from_near_axis`. This method creates a DESC `Equilibrium` based off of the `pyQsc` equilibrium. It requires as input the desired DESC FourierZernike resolution, as well as the radius at which you want to evaluate the qsc equilibrium at to make the DESC equilibrium's boundary. The equilibrium's initial `R_lmn`, `Z_lmn` Fourier-Zernike coefficients are fit to the `R,Z` evaluated from the `pyQsc` equilibrium, and the initial `L_lmn` are 0 (because the `pyQsc` equilibrium uses Boozer angles, so there is no poloidal stream function)"
```
ddudt, line 65: Clarification because it's been a while since I implemented this: the poloidal angle is kept as the Boozer poloidal angle, but the toroidal angle is converted to the standard toroidal angle. $(\theta_B, \phi)$ are generally not straight field-line coordinates, so are we sure it's correct to set $\lambda=0$?

docs/notebooks/DESC_Solve_from_pyQSC.ipynb
```diff
@@ -0,0 +1,250 @@
...
+   "source": [
+    "ntheta = 75\n",
+    "r = 0.15\n",
+    "desc_eq = Equilibrium.from_near_axis(\n",
+    "    qsc_eq,  # the Qsc equilibrium object\n",
+    "    r=r,  # the finite radius (m) at which to evaluate the Qsc surface to use as the DESC boundary\n",
+    "    L=8,  # DESC radial resolution\n",
+    "    M=8,  # DESC poloidal resolution\n",
+    "    N=8,  # DESC toroidal resolution\n",
+    "    ntheta=ntheta,\n",
```
ddudt, line 83: suggest to add the comment:  `# number of poloidal grid points used in the conversion`

docs/notebooks/DESC_Solve_from_pyQSC.ipynb
```diff
@@ -0,0 +1,250 @@
...
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "plot_comparison(\n",
+    "    eqs=[desc_eq, eq_fit],\n",
+    "    labels=[\"DESC-solved equilibrium\", \"NAE surfaces\"],\n",
+    "    zeta=2,\n",
+    "    figsize=(12, 12),\n",
+    "    theta=0,\n",
```
ddudt, line 216: What does the plot look like if you include the theta* contours? That might answer my previous question about if lambda is supposed to be zero or not

desc/transform.py
f0uriest: good point, ill fix this

docs/notebooks/DESC_Solve_from_pyQSC.ipynb
f0uriest: image
Looks pretty similar on the symmetry planes but a bit off on the others. Lambda probably isn't zero but I'm not sure if it's possible/easy to compute what it should be.

f0uriest (approved):

PR #446, opened 2023-03-06
Additional VMEC outputs

### LINKED ISSUE #445: Add VMEC output quantities needed for particle tracing
Request by Rogerio:

- [x] beta_{pol, tor}
- [ ] beta_{axis, vol} - vol seems to be a profile?
- [x] q_factor
- [x] j_{curu, curv} - are these just FSA of J^theta etc?
- [x] r_b{tor,tor0}
- [ ] bdotgradv (source [here](https://github.com/PrincetonUniversity/STELLOPT/blob/3b0f12d31926e4900c15b473fcafb01ed90605c7/VMEC2000/Sources/Input_Output/jxbforce.f#L658))
- [x] b_0 (source [here](https://github.com/hiddenSymmetries/VMEC2000/blob/e5de70486235be997e95ca4a129441d51b796226/Sources/Input_Output/eqfor.f#L362))

ddudt:
Resolves #445. 

`bdotgradv`, `beta_vol`, and `betaxis` are still not computed. But according to Rogerio, he doesn't actually need those values for his particle tracing -- he only need the output entry to exist to avoid errors. So for now they are set to 0 until we can figure out what that quantity really is.


desc/compute/_equil.py
```diff
@@ -462,6 +462,74 @@ def _W_B(params, transforms, profiles, data, **kwargs):
...
+    units_long="Joules",
+    description="Plasma magnetic energy enclosed by flux surfaces",
+    dim=1,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="",
+    data=["|B|", "e_theta", "e_zeta", "Z"],
+)
+def _W_B_of_r(params, transforms, profiles, data, **kwargs):
```
f0uriest, line 478: not sure I totally understand this. It looks like divergence theorem or something to turn the volume integral into a surface integral but what is the divergence of?

ddudt: I took the divergence theorem that we use to compute `V(r)` ([see code here](https://github.com/PlasmaControl/DESC/blob/master/desc/compute/_geometry.py#L39)) and just multiplied it by B^2. 

I wasn't sure about the math, but this matches the VMEC output for `beta_vol` to within a few percent.

f0uriest (dismissed):

dpanici (approved):

f0uriest (approved):

PR #448, opened 2023-03-08
Refactor iota computation and add limit

unalmis:
- [x] Refactor the computation of iota and add ability to compute iota at the rho=0 axis
    - Computation shows the limit appears to be a continuous extension of iota to the axis. As long as it actually is, I think this is sufficient to formally justify that interchange of the integral and limit we discussed.  [my_math.pdf](https://github.com/PlasmaControl/DESC/files/11736797/current_derivation.pdf)
- [x] Correct QAS input file
    - The input file on master for QAS was generated sometime between September (I think) and October 15. The one I uploaded in thie PR is the one generated by DESC on October 15 (and could be regenerated when this PR was made) that had better agreement with VMEC's QAS solve. Although, the VMEC input [input.QAS.txt](https://github.com/PlasmaControl/DESC/files/11680002/input.QAS.txt) to DESC input converter recently (as in sometime close to June 6 2023) gives something different now.
    - The reason this change is included in this PR is because one test on QAS for the iota limit computation includes a comparison to a hardcoded value of iota that comes from solving the QAS input file. So if I want that test to be more accurate I should make sure QAS is being solved well.
- [x] Implement [efficient API for limits.pdf](https://github.com/PlasmaControl/DESC/files/11697684/limit.api.pdf)
    - Skipped part 3 for now due to Rory's suggestion; although the broadcasting is not an issue (updated the pdf to show why)


tests/test_compute_utils.py
```diff
@@ -95,30 +96,35 @@ def test_surface_integrals(self):
         should also be done on grids with duplicate nodes (e.g. endpoint=True).
         """
 
-        def test(surface_label, grid):
-            q = np.random.random_sample(size=grid.num_nodes)
+        def test_B_theta(surface_label, grid, eq):
+            q = eq.compute("B_theta", grid=grid)["B_theta"]
```
unalmis, line 100: I had changed this earlier to confirm B_theta was computed correctly before we fixed that plotting issue. It's better than a random quantity test, so I kept the change.

tests/test_compute_utils.py
unalmis, line 1: I don't think we need live solves for these tests.

unalmis:
Not an urgent PR (can review whenever). I just added you all since we discussed this PR today.

desc/compute/_profiles.py
```diff
@@ -644,40 +615,216 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
+    profiles=[],
+    coordinates="r",
+    data=["g_tt", "g_tt_r", "sqrt(g)", "sqrt(g)_r"],
+)
+def _iota_0_den_r(params, transforms, profiles, data, **kwargs):
+    den = data["g_tt"] / data["sqrt(g)"]
+    den_r = (data["g_tt_r"] - den * data["sqrt(g)_r"]) / data["sqrt(g)"]
+    data["iota_0_den_r"] = surface_averages(transforms["grid"], den_r)
+    # TODO: limit at axis
+    #    I will need to check if limit and derivative interchangeable.
```
unalmis, line 802: This is an old comment; there is a simpler way to do this.

desc/compute/_profiles.py
```diff
@@ -644,40 +615,212 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
+    data=["lambda_t", "lambda_z", "g_tt", "g_tz", "sqrt(g)"]
+    + ["lambda_rt", "g_tt_rr", "g_tz_r", "g_tz_rr", "sqrt(g)_r", "sqrt(g)_rr"],
+)
+def _iota_0_num(params, transforms, profiles, data, **kwargs):
+    num = (
+        data["lambda_z"] * data["g_tt"] - (1 + data["lambda_t"]) * data["g_tz"]
+    ) / data["sqrt(g)"]
+    data["iota_0_num"] = surface_averages(transforms["grid"], num)
+
+    if transforms["grid"].axis.size:
```
f0uriest, line 647: Not sure this is the best way to incorporate these, since in most cases we don't need the axis value but the way the dependencies work it will still calculate all the higher order derivatives needed for the axis value.

unalmis: Good point. It seems to avoid unnecessarily computing limits, we should only do so when explicitly specified. I think the most convenient option for the user would be something like:
```python
eq.compute(["iota", "|B|"], axis=True)
```

I plan to add a `limit_params` list to the compute functions.
Then tweak how the dependency list is built to only include `limit_params` 
if `axis` was specified.

Then inside the actual compute function, the condition will be
```python
if axis and transforms.grid.axis.size:
    # replace axis indices with value at limit
```

dpanici:
@unalmis what is the status on this? I am starting to write my paper where I use on-axis quantitites from DESC as a verification against near-axis theory, and it could be nice to be able to just say "on-axis values from DESC" and not have to sweep under the rug the fact that we are calculating things at $\rho=1e-6$ instead of at 0 (though maybe not completely necessary)

unalmis:
> @unalmis what is the status on this? I am starting to write my paper where I use on-axis quantitites from DESC as a verification against near-axis theory, and it could be nice to be able to just say "on-axis values from DESC" and not have to sweep under the rug the fact that we are calculating things at ρ=1e−6 instead of at 0 (though maybe not completely necessary)

Would you like to just compute iota and its derivatives at 0, or are you asking for the status of https://github.com/PlasmaControl/DESC/issues/486? If its the former, I can probably do the math for the `iota_r` and `iota_rr` and add them with tests within the next few days.

dpanici:
> > @unalmis what is the status on this? I am starting to write my paper where I use on-axis quantitites from DESC as a verification against near-axis theory, and it could be nice to be able to just say "on-axis values from DESC" and not have to sweep under the rug the fact that we are calculating things at ρ=1e−6 instead of at 0 (though maybe not completely necessary)
> 
> Would you like to just compute iota and its derivatives at 0, or are you asking for the status of #486? If its the former, I can probably do the math for the `iota_r` and `iota_rr` and add them with tests within the next few days.

Just iota (not its derivatives) and |B| (which I assume is the harder of the 2 here)

unalmis:
@dpanici You can use this branch to compute `|B|` at the limit as well now.  Just give a grid with a node on the axis and it will be handled automatically.
Many of the limit quantities are easy to do once we know iota. If there are quantities you'd prefer to have before others let me know. I'll need to improve the api before this PR can be merged.

dpanici:
> @dpanici You can use this branch to compute `|B|` at the limit as well now. Just give a grid with a node on the axis and it will be handled automatically. Many of the limit quantities are easy to do since we know iota's limit. If there are quantities you'd prefer to have before others let me know. I'll need to improve the api before this PR can be merged.

Awesome, thanks!

desc/compute/_profiles.py
unalmis: @f0uriest Here's my plan for the API
[limit api.pdf](https://github.com/PlasmaControl/DESC/files/11606378/limit.api.pdf)

f0uriest: This looks great! I think its more important to make sure criteria 1&2 are satisfied, ensuring criteria 3 is satisfied seems trickier and could potentially get messy. I'd start with just computing even the limit dependencies over the full grid, and we can see how slow it is. My hunch is it won't be that bad in practice, since most of the optimization grids don't include the axis, so the axis terms would be mainly for post-hoc plotting and analysis where speed isn't as important. It should also simplify the broadcasting etc.

f0uriest: we can get a sense of how much extra computation is done by just looking at the ratio of the number of limit dependencies vs regular dependencies for each quantity.

tests/test_compute_utils.py
unalmis, line 1: Just ignore the changes in tests/test_compute_utils.py. After the `function integral` PR is merged i'll update this.

desc/compute/__init__.py
```diff
@@ -52,16 +52,37 @@
...
+                "params": get_params(key, True),
+                "profiles": get_profiles(key, True),
+            }
+            for _key, val in full_with_axis.items():
+                if _key == "transforms":
+                    continue
+                if len(full[_key]) >= len(val):
+                    # one is a deep copy of the other; dereference to save memory
+                    full_with_axis[_key] = full[_key]
+        data_index[key]["full_with_axis_dependencies"] = full_with_axis
```
unalmis, line 85: I feel like this is clunky, but I don't have a better idea at the moment.

desc/equilibrium/configuration.py
```diff
@@ -1134,9 +1138,11 @@ def compute(
...
                 dep1d,
                 params=params,
-                transforms=get_transforms(dep1d, eq=self, grid=grid1d, **kwargs),
-                profiles=get_profiles(dep1d, eq=self, grid=grid1d),
-                data=None,
+                transforms=get_transforms(
+                    dep1d, grid1d.axis.size, eq=self, grid=grid1d, **kwargs
+                ),
+                profiles=get_profiles(dep1d, grid1d.axis.size, eq=self, grid=grid1d),
+                data=None,  # Todo: pass in data0d?
```
unalmis, line 1145: wasn't an issue now apparently just commented for future. in any case i see why we don't need this logic inside the compute utils dependency building recursion.

desc/compute/utils.py
```diff
@@ -134,9 +146,14 @@ def get_data_deps(keys):
...
+            if "full_with_axis_dependencies" in data_index[key]:
+                return data_index[key]["full_with_axis_dependencies"]["data"]
+        elif "full_dependencies" in data_index[key]:
             return data_index[key]["full_dependencies"]["data"]
         deps = data_index[key]["dependencies"]["data"]
+        if has_axis:
+            deps += data_index[key]["dependencies"]["limit_data"]
         if len(deps) == 0:
             return deps
         out = deps.copy()
```
unalmis, line 159: curious why we did this

unalmis:
Since adding the limit API stuff, unit tests 3.10.2 stalls for 6 hours then fails on this PR on Github's ci. Not sure if I've broke something. On my machine with python 3.10.10 ([local_envrionment .txt](https://github.com/PlasmaControl/DESC/files/11679916/kaya_version.txt)), `pytest -m unit tests` succeeds in 49 minutes.

desc/compute/data_index.py
```diff
@@ -45,18 +46,21 @@ def register_compute_fun(
         Coordinate dependency. IE, "rtz" for a function of rho, theta, zeta, or "r" for
         a flux function, etc.
     data : list of str
-        Names of other items in the data index needed to compute qty
+        Names of other items in the data index needed to compute qty.
+    limit_data : list of str
```
f0uriest, line 50: `limit_data` could be ambiguous, maybe `axis_data` or `axis_limit_data`?

desc/compute/utils.py
```diff
@@ -114,17 +114,29 @@ def _compute(names, params, transforms, profiles, data=None, **kwargs):
                 data=data,
                 **kwargs,
             )
-            data = data_index[name]["fun"](params, transforms, profiles, data, **kwargs)
+            if transforms["grid"].axis.size:
```
f0uriest, line 117: does this work correctly?
I'm thinking of the following scenario:
- user wants some data at the axis
- above block triggers, computing the regular data dependencies, but maybe doesn't have the required data to compute the axis part yet so thats nan
- this block triggers, asking to compute the axis dependencies, but if they were already computed above they get skipped, leaving us with nans at the axis?

desc/compute/utils.py
f0uriest: It's to avoid modifying the actual data index, since `deps = data_index[key]["dependencies"]["data"]` is just a pointer to the list. I think you might need the same thing on line 156.

```python
a = {"b": [1,2,3]}
c = a['b']
c += [4,5,6]

print(a)
>>> {'b': [1, 2, 3, 4, 5, 6]}

File: desc/compute/__init__.py [hunk 9]
```diff
@@ -52,16 +52,37 @@
...
+        full = {
+            "data": get_data_deps(key, False),
+            "transforms": get_derivs(key, False),
+            "params": get_params(key, False),
+            "profiles": get_profiles(key, False),
+        }
         data_index[key]["full_dependencies"] = full
 
+        full_with_axis_data = get_data_deps(key, True)
+        if len(full["data"]) >= len(full_with_axis_data):
```
f0uriest, line 68: is this possible?

desc/compute/__init__.py
f0uriest: can you explain a bit what's going on here? Not sure I understand.

desc/compute/data_index.py
unalmis: i'll pick the latter

f0uriest:
> Since adding the limit API stuff, unit tests 3.10.2 stalls for 6 hours then fails on this PR on Github's ci. Not sure if I've broke something. On my machine with python 3.10.10 ([local_envrionment .txt](https://github.com/PlasmaControl/DESC/files/11679916/kaya_version.txt)), `pytest -m unit tests` succeeds in 49 minutes.

is it stalling consistently? I've noticed that sometimes it would stall on other branches too, but usually restarting it fixed the issue.

desc/compute/__init__.py
unalmis: - Short answer: only the equality is possible, I suppose the inequality was just a defensive programming habit on my part.
- Long answer: `data_index[key]["full_dependencies"]["data"]` stores all the names of the quantities that `key` depends on. `data_index[key]["full_with_axis_dependencies"]["data"]` stores all the names of the quantities that `key` and the limit at the axis of `key` depend on. The purpose of creating the extra `full_with_axis` dictionary is the same as the normal one (described in the comment at line 55).

It is possible for some quantity and all its dependencies to manifestly not depend on anything extra to evaluate its limit (e.g. `data["rho"]`). In that case, `data_index[key]["full_dependencies"]["data"]` is a list which is identical to `data_index[key]["full_with_axis_dependencies"]["data"]`. Hence, when the length of these lists are equal, we know that `data_index[key]["full_with_axis_dependencies"]["data"]` will duplicate `data_index[key]["full_dependencies"] `. So I simply assign to `data_index[key]["full_with_axis_dependencies"]["data"]` the same reference that was assigned to `data_index[key]["full_dependencies"]["data"]`

desc/compute/__init__.py
unalmis: Similar to [this comment](https://github.com/PlasmaControl/DESC/pull/448/files/ea78cdc354c0fc6fd3997a7ea789e77c91a7875b#r1222004632). Inside this `else` branch, we know that `key` requires an additional dependency to evaluate its limit at the axis. Typically this dependency will need the next higher order derivative of some quantity, so it will require an additional (or higher order) `transform`.  The evaluation of the limit will usually not need any additional profile like `current` or `iota`. `data["B0"]` is a good example of this.

In this loop of length 4, my code claims that when the length of the list returned by `get_params(key, has_axis=True)` is equal to the length of the list returned by `get_params(key, has_axis=False)`, then they must be duplicates of one another because no additional `params` are needed to evaluate the limit. Instead of keeping references to two copies of the same object, I just use the full_dependencies params reference. (This doesn't work on `transforms` because a list of transforms with equal length may have contents that differ.)

desc/compute/utils.py
unalmis: I think my modification on line 372 of the function `_has_dependencies` accounts for this. The inside of this if statement is executed whenever a dependency is missing (whether that be a normal dependency or one that is only needed to evaluate the limit at the axis). The `_has_dependencies` function contains logic that states quantities that are only required to evaluate the limit are considered dependencies if and only if `transforms["grid"].axis.size > 0`.

If we always use `get_data_deps(keys, has_axis=grid.axis.size > 0)` as our input to the compute function, then the `if transforms["grid"].axis.size"` is redundant. I've kept it because I think it improves readability and prevents computing the limit dependencies when they are not needed (for example if the user used `get_data_deps(keys, has_axis=not(grid.axis.size > 0))`.

desc/compute/utils.py
unalmis: Ahh this I need to correct then. The logic here had assumed `deps` was already a copy.  I guess I got used to working with jax arrays and was primed to think a new copy is always created on modification.

desc/compute/__init__.py
unalmis: I felt like it was clunky because we create two dictionaries whose internals are mostly identical, but we compute and store each them separately. The intention is to save memory by making the internals (depth level of 1 internals) point to the same instance when I believe the two things are copies.

unalmis:
> > Since adding the limit API stuff, unit tests 3.10.2 stalls for 6 hours then fails on this PR on Github's ci. Not sure if I've broke something. On my machine with python 3.10.10 ([local_envrionment .txt](https://github.com/PlasmaControl/DESC/files/11679916/kaya_version.txt)), `pytest -m unit tests` succeeds in 49 minutes.
> 
> 
> 
> is it stalling consistently? I've noticed that sometimes it would stall on other branches too, but usually restarting it fixed the issue.

Failed twice but passed now.

desc/plotting.py
```diff
@@ -2306,11 +2306,22 @@ def plot_boozer_surface(
     N_booz = kwargs.pop("N_booz", 2 * eq.N)
     title_font_size = kwargs.pop("title_font_size", None)
 
+    # Todo: review this please
```
f0uriest, line 2309: What is the issue here?

unalmis: There shouldn't be any. I marked the objectives that used two grids with this comment in case these should be handled differently. I don't see any problem if `grid_compute` and `grid_plot` differ in whether they have an axis.

f0uriest: OK yeah it seems fine to me. And I think @ddudt had some changes on another branch that got rid of the separate grids for plotting/computing?

desc/compute/_profiles.py
```diff
@@ -478,7 +478,8 @@ def _gradp_mag_vol(params, transforms, profiles, data, **kwargs):
     transforms={"grid": []},
     profiles=["iota", "current"],
     coordinates="r",
-    data=["psi_r", "lambda_t", "lambda_z", "g_tt", "g_tz", "sqrt(g)"],
+    data=["psi_r", "iota_0_num", "iota_0_den"],
```
dpanici, line 481: maybe I misunderstand the dependency logic, but isn't the iota_0_den and iota_0_num axis limit data? not data?

dpanici: oh nvm I see, `iota_0_num` is what you call the zero current iota. That name could be a bit confusing when we are talking about axis limits... but I can't think of a better name for it right now. Maybe `iota_zero_curr` or `iota_vac`?

desc/equilibrium/configuration.py
dpanici: this is in case there are 1d quantities which rely on 0D quantities?

desc/objectives/utils.py
```diff
@@ -313,7 +313,7 @@ def factorize_linear_constraints(constraints, objective_args):  # noqa: C901
 
     def project(x):
         """Project a full state vector into the reduced optimization vector."""
-        x_reduced = jnp.dot(Z.T, (x - xp)[unfixed_idx])
```
dpanici, line 316: any reason for changing dot to matmul? I think they are basically equivalent for 1-D and 2-D array operations, so I am fine with the change

tests/test_axis_limits.py
```diff
@@ -0,0 +1,77 @@
...
+            assert np.isfinite(b_mag_fsa_no_sqrt_g).all()
+            np.testing.assert_allclose(
+                b_mag_fsa_no_sqrt_g[:-1], b_mag_fsa_no_sqrt_g[1:], atol=epsilon
+            )
+            # check value
+            np.testing.assert_allclose(
+                b_mag_fsa_no_sqrt_g[0], expected_at_axis, atol=epsilon
+            )
+
+        test(get("W7-X"), 2.708108)
```
dpanici, line 56: where did the expected on-axis value come from?

tests/test_axis_limits.py
```diff
@@ -0,0 +1,77 @@
...
+            lg = LinearGrid(rho=rho, M=5, N=5, NFP=eq.NFP, sym=eq.sym)
+            iota = compress(lg, eq.compute("iota", grid=lg)["iota"])
+            # check continuity
+            assert np.isfinite(iota).all()
+            np.testing.assert_allclose(iota[:-1], iota[1:], atol=epsilon)
+            # check value
+            np.testing.assert_allclose(iota[0], expected_at_axis, atol=epsilon)
+
+        eq = desc.io.load(load_from=str(DSHAPE_current["desc_h5_path"]))[-1]
+        test(eq, -0.994167)
```
dpanici, line 76: these axis values come from where?

desc/objectives/utils.py
unalmis: In the dev_guide PR, my math showed `x_reduced` = $Z^T (x - xp)$, so I was expecting to see either `Z.T @ ((x-xp)[unfixed_idx])` or `jnp.matmul(Z.t, (x-xp)[unfixed_idx])` in the code. `jnp.dot` was confusing with the math notation I am familiar with for vectors since $Z^T \cdot (x-xp) = \langle Z^T, x-xp \rangle = Z (x-xp) \neq$ `x_reduced`.
I know now that `dot` is general in numpy to make the operation well-defined, but I liked matmul more

desc/equilibrium/configuration.py
unalmis: yea i can update the comment to say pass it in when there is one of those

tests/test_axis_limits.py
unalmis: same as https://github.com/PlasmaControl/DESC/pull/448#discussion_r1228688339

dpanici (approved):

desc/compute/utils.py
```diff
@@ -188,17 +213,19 @@ def _get_derivs_1_key(key):
     return {key: np.unique(val, axis=0).tolist() for key, val in derivs.items()}
 
 
-def get_profiles(keys, eq=None, grid=None, **kwargs):
+def get_profiles(keys, has_axis, eq=None, grid=None, **kwargs):
```
f0uriest, line 216: For this and the similar `get_` stuff, could we move the `has_axis` arg to the end and default to `False`? I think the most common use case is when a grid is passed in, so the `has_axis` in that case is redundant, so this would save some code and keep the API the same in most cases.

desc/compute/utils.py
```diff
@@ -257,13 +287,15 @@ def get_params(keys, eq=None, **kwargs):
     return params
 
 
-def get_transforms(keys, eq, grid, **kwargs):
+def get_transforms(keys, has_axis, eq, grid, **kwargs):
```
f0uriest, line 290: similarly here, `has_axis` seems redundant with `grid`

desc/objectives/_geometry.py
```diff
@@ -528,7 +534,9 @@ def build(self, eq, use_jit=True, verbose=1):
 
         self._dim_f = surface_grid.num_nodes
         self._data_keys = ["R", "phi", "Z"]
-        self._args = get_params(self._data_keys)
+        self._args = get_params(
+            self._data_keys, has_axis=plasma_grid.axis.size or surface_grid.axis.size
```
f0uriest, line 538: FYI the surface grid should never have a point at the axis (since its a 2d grid over a surface) and realistically the plasma grid shouldn't either.

desc/plotting.py
ddudt, line 1: Please don't reformat this file to add more lines of code, as long as the existing format is within the max line length limit.

ddudt (commented):
Can we make a list of which quantities can now be computed at the axis and which axis limits still need to be implemented?

desc/compute/data_index.py
```diff
@@ -45,18 +46,21 @@ def register_compute_fun(
...
     Should only list *direct* dependencies. The full dependencies will be built
-    recursively at runtime using each quantities direct dependencies.
+    recursively at runtime using each quantity's direct dependencies.
     """
     deps = {
         "params": params,
         "transforms": transforms,
         "profiles": profiles,
         "data": data,
+        "axis_limit_data": [] if axis_limit_data is None else axis_limit_data,
```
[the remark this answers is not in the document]
unalmis: Using a mutable object as a default parameter can lead to bugs. Because the default parameter of the function stores a reference to some list, if that list is modified, then the default value of the function changes accordingly to the modified list.
We don't plan to ever modify a quantity's `axis_limit_data` list, so you're right that we can use the same empty list for `axis_limit_data` for all quantities that don't specify it in the `compute_funs`. But we use this same `None` to empty list design pattern in other places of the code too.

f0uriest (approved):

unalmis:
I'm going to merge this since the changes since Dario's approval were cosmetic

PR #451, opened 2023-03-13
issue with symmetric grids

unalmis:
Fixes issues with symmetric grids.

- [x] On `LinearGrid`, specifying nodes at coordinates $\theta \in$ [0, $\pi$] is now sufficient when `sym=True`.
- [x] When `sym=True`, nodes at coordinates $\theta$ mod $\pi = 0$ should not have their $d\theta$ scaled up.
- [x] `LinearGrid` can now handle non-uniformly spaced nodes if the user passes them when `sym=True`.
- [x] Corrects `grid.weights` on `ConcentricGrid` with `sym=True`.
    - Resolving example equilibria and tightening those test tolerances to be done in different PR.
- [x] More tests.


unalmis:
If some of the grid spacing tests are a chore to review, it might be easier to just check results directly. For convenience:
```python
import sys
import os
# sys.path.append(os.path.abspath("../../"))
from desc.grid import *
from desc.plotting import plot_grid
import numpy as np

theta = np.linspace() or integer
lg = LinearGrid(rho=, theta=, zeta=, sym=)
print(lg.spacing[:, 1])
plot_grid(lg)
```

tests/test_grid.py
```diff
@@ -531,27 +725,127 @@ def test(grid, basis, err_msg, true_avg=1):
...
+                dr = np.empty_like(r_unique)
+                dr[0] = (r_unique[0] + r_unique[1]) / 2
+                dr[1:-1] = (r_unique[2:] - r_unique[:-2]) / 2
+                dr[-1] = 1 - (r_unique[-2] + r_unique[-1]) / 2
+            else:
+                dr = np.ones(grid.num_rho) / grid.num_rho
+            expected_integral = np.sum(dr * compress(grid, function_of_rho))
+            true_integral = np.log(1.35 / 0.35)
+            midpoint_rule_error_bound = max(dr) ** 2 / 24 * (2 / 0.35**3)
+            right_riemann_error_bound = dr[0] * (1 / 0.35 - 1 / 1.35)
```
unalmis, line 782: This is what I was mentioning earlier.
A line integral over rho is accurate to what we expect (within error bound of midpoint rule). Of course it would be better if the error was guaranteed to be smaller (like how Simpson's rule does better than midpoint rule).
The would be relevant to make integrals over three dimensions (rho, theta, and zeta), more accurate since those include an integral over rho.

But I'm pretty sure it's not an issue since the integrals over all coordinates that we are interested in, are volume integrals that include the Jacobian factor - which our grids are designed to integrate well.

dpanici: Yea, so if someone wants to do an integral over rho (like say integrate the surface area of each surface over rho to get volume), they should also know to multiply by the correct metric quantity so that the end result is volume

tests/test_grid.py
```diff
@@ -135,6 +141,28 @@ def test(sym, endpoint, axis, ntheta_is_odd):
...
+        """Test that specifying theta nodes from [0, pi] is sufficient."""
+        # uniform spacing
+        theta = np.linspace(0, 2 * np.pi, 10)
+        lg_1 = LinearGrid(L=5, theta=theta, N=5, sym=True)
+        lg_2 = LinearGrid(L=5, theta=theta[theta <= np.pi], N=5, sym=True)
+        np.testing.assert_allclose(lg_1.nodes, lg_2.nodes)
+        np.testing.assert_allclose(lg_1.spacing, lg_2.spacing)
+        np.testing.assert_allclose(lg_1.weights, lg_2.weights)
+
+        # non-uniform spacing
```
dpanici, line 155: what exactly is the expected behavior for a linear grid with non-uniformly spaced points?

desc/grid.py
```diff
@@ -53,50 +53,69 @@ def __init__(self, nodes, sort=True):
...
         2. Rescale theta spacing to preserve dtheta weight.
             Need to rescale on each theta coordinate curve by a different factor.
-            dtheta should = 2pi / number of nodes remaining on that theta curve
-
-        Returns
-        -------
-        dtheta_scale : ndarray
-            The multiplicative factor to scale the theta spacing for each theta curve.
-                number of nodes / (number of nodes - number of nodes to delete)
+            dtheta should = 2pi / number of nodes remaining on that theta curve.
```
dpanici, line 69: what is meant by "theta curve"?

unalmis: the intersection of a rho and zeta surface

dpanici: Isn't that just a toroidal cross-section?

tests/test_grid.py
unalmis: The rule used to compute the spacing when an array is given is: $d\theta$ is chosen to be half the cyclic distance of the surrounding two nodes. In other words, if we parameterize a circle’s perimeter from 0 to $2\pi$ , and place points on it according to the given array (e.g. theta = np.linspace(0, 2pi, 4)), then the $d\theta$ assigned to each node will be half the parameterized distance along the arc between its left and right neighbors. The process is the same for $d\zeta$.

Visuals:
Here we are visualizing the $d\theta$ spacing of a $\theta$ curve (intersection of $\rho$ and $\zeta$ surface).
Let the node's coordinates be at the values given by the filled circles.
The $d\theta$ spacing assigned to each node is the length of arc of the same color.

Example: uniform spacing
linearly spaced

Example: non-uniform spacing
non linear spacing

Example: Two nodes with symmetry set to False
another example

Example: The same two nodes with symmetry set to True
Notice now the red node is given more weight, because symmetry says there is a duplicate of that node (in black) across the axis of symmetry.
symmetry

desc/grid.py
unalmis: I thought a toroidal cross-section is just a zeta surface (surface of constant zeta, which is two dimensional) and the intersection of a rho surface and zeta surface is a curve (one dimensional line).

tests/test_grid.py
unalmis: In this sense LinearGrid can actually work better than the base Grid for custom grids. User can give nodes at whatever location when they pass in arrays, and the spacing will be a midpoint rule. In Grid it's just assumed constant.

desc/grid.py
dpanici: I see, thanks

tests/test_grid.py
dpanici: Ok. I don't see the use case of calling `LinearGrid` with a non-uniform spacing in theta/zeta but this seems a good way to treat that case. Thanks for the explanation

dpanici (approved):

PR #458, opened 2023-03-16
Add axis as optimizable DoF

### LINKED ISSUE #322: Add ability to optimize using axis as independent DOF

f0uriest:
- Add axis degrees of freedom to arg_order
- Add new constraints for enforcing self-consistency between plasma/axis
- FixAxis constraints now only fix Xa_n

Resolves #322


desc/vmec.py
```diff
@@ -147,19 +139,8 @@ def load(
...
-        )
-        objective = ObjectiveFunction(constraints, eq=eq, verbose=0)
-        _, _, _, _, _, project, recover = factorize_linear_constraints(
-            constraints, objective.args
-        )
-        args = objective.unpack_state(recover(project(objective.x(eq))))
-        for key, value in args.items():
-            setattr(eq, key, value)
+        eq.surface = eq.get_surface_at(rho=1)
+        eq.axis = eq.get_axis()
```
ddudt, line 143: We don't want to change this! When we load in a VMEC solution, we do a least-squares fit of the R & Z data. `eq.get_surface_at(rho=1)` is not guaranteed to be the same surface as the VMEC boundary after this fit, and therefore we need to apply the linear constraints to force the internal surfaces to be self-consistent with the desired boundary surface. We probably want this same behavior for the axis, so that the resulting DESC equilibrium will have the exact same surface and axis shapes as the loaded VMEC solution.

ddudt: We should also probably add a test for this

ddudt (changes requested):

desc/vmec.py
f0uriest: Not sure that's right. We do the least squares fit, but then we compute the target boundary based on that fit, so projecting the boundary should basically be an identity operation?

ddudt: `inputs["surface"]` is created from the VMEC boundary, and used to initialize the Equilibrium. Then `eq.R_lmn` and `eq.Z_lmn` get set by the least-squares fit, so they are generally not self-consistent with `eq.surface`. But we want to keep the surface the same and update `eq.R_lmn` and `eq.Z_lmn` accordingly, which is what the feasible solution from `factorize_linear_constraints` does. This works because the default target for `FixBoundaryR` is `eq.surface.R_lmn` (and similar for Z)

f0uriest: Ahh I see. I'll revert this. Enforcing the same surface makes sense, I'm not sure about enforcing the same axis, since the behavior near the VMEC axis may not be analytic.

ddudt: I agree that VMEC's axis may not be good, but considering the purpose of this function is to load in the VMEC solution I feel like we should preserve their axis shape. 

Keep in mind that the least-squares fit is weighted heavily towards the boundary, since that is where most of VMEC's surfaces are. So we could actually end up with a very different axis location than the VMEC solution if we don't correct it.

ddudt (approved):

ddudt (approved):

daniel-dudt (approved):

ddudt (dismissed):

daniel-dudt (approved):

PR #459, opened 2023-03-16
Refactor linear constraints

### LINKED ISSUE #449: Separate boundary constraint into AR=Rb and Rb=Rb_target
Right now the `FixBoundary*` constraints do two things, depending on if `fixed_boundary` is true or false, and its kind of klunky, especially if we add other constrained optimizers.

Rather than treating Rb, Zb etc specially, it's probably better to treat them the same as other args, but we just always need to include a constraint that AR=Rb etc. Then the optional constraint that the user would see is something that fixes Rb=Rb_target.

The AR=Rb constraint would be automatically applied when needed, transparent to the user. We should likely also do the same for the lambda gauge constraint.

f0uriest:
- Allows linear constraints to take more than 1 argument
- Refactors `factorize_linear_constraints` to return full `A`,`b` rather than dicts.
- Add new constraints for enforcing consistency between `X_lmn` and `Xb_lmn`
- `FixBoundaryX` now only fixes `Xb_lmn`, with the self consistency constrained being applied automatically.

Resolves #449


desc/objectives/objective_funs.py
```diff
@@ -170,16 +177,17 @@ def build(self, eq, use_jit=None, verbose=1):
         # build objectives
         self._dim_f = 0
         for objective in self.objectives:
-            if verbose > 0:
-                print("Building objective: " + objective.name)
-            objective.build(eq, use_jit=self.use_jit, verbose=verbose)
+            if not objective.built:
```
dpanici, line 180: If the equilibrium changes resolution, but the same objective is used, with this conditional here the objective won't be rebuilt, even if one explicitly did 
```
eq = Equilibrium(L=3,M=3)
f_obj = ForceBalance()
obj = ObjectiveFunction(f_obj)
eq.solve(maxiter=5,objective=obj)
eq.change_resolution(L=5,M=5)
obj.build(eq) # it is already built, so this won't do anything
eq.solve(maxiter=5,objective=obj) # fails with a mismatch in shapes
```
I guess one could re-create the `ObjectiveFunction` entirely, but even then, if it is built with previously-created `Objective` objects, then this will not correctly re-build those objects.

I guess my gripe here is that having this here means one MUST re-create new `Objective` objects in order to change the resolution, which might not be a problem but I think if one changes resolution and then rebuilds an `ObjectiveFunction` with that eq, that should be enough to have that `ObjectiveFunction` be built with the correct resolutions

desc/objectives/utils.py
```diff
@@ -86,6 +83,21 @@ def get_fixed_boundary_constraints(
     return constraints
 
 
+def maybe_add_self_consistency(constraints):
+    """Add self consistency constraints if needed."""
```
dpanici, line 87: "If needed" just means if the constraints do not currently contain R consistency, Z consistency, or FixLambdaGauge, it will add them to the constraints? So It will almost have to add these?

f0uriest: Yeah basically. The assumption is that the user doesn't need to specify those constraints explicitly. In the same vein, I currently have it so the `get_fixed_boundary_constraints` etc don't return the self consistency constraints, but we can change that if we want. My thinking was that the stuff returned by `get_fixed...` is stuff the user may take as a guide for custom constraints and modifying stuff, but we don't want them to be messing with the self consistency stuff.

desc/objectives/objective_funs.py
f0uriest: The motivation for this was that the user will often call `objective.build` themselves before passing it to the optimizer, and the optimizer would then build again. Similarly, in the course of a regular solve/optimization/perturb run we end up rebuilding stuff way more often than we need to so was trying to cut down on that.

I think this has been a general problem for a while, and I'm not sure what the best solution is. Do we want to always rebuild the objective, even if it's not needed? Or do we try to detect when things change, which might not always be easy?

I'm open to suggestions.

desc/objectives/utils.py
```diff
@@ -237,95 +246,63 @@ def factorize_linear_constraints(constraints, objective_args):  # noqa: C901
...
+                *[jnp.zeros(obj.dimensions[arg]) for arg in obj.args]
+            )
+            for arg in args
+        }
+        # using obj.compute instead of obj.target to allow for correct scale/weight
+        b_ = -obj.compute_scaled(*[jnp.zeros(obj.dimensions[arg]) for arg in obj.args])
+        A.append(A_)
+        b.append(b_)
+
+    A_full = jnp.vstack([jnp.hstack([Ai[arg] for arg in args]) for Ai in A])
```
dpanici, line 272: So here, A contains the dict which has A matrices for each arg, and Ai[arg] is empty(?) if the objective does not rely on that arg? so then when the inner jnp.hstack happens, it creates an A matrix for each objective, then  the vstack puts the A matrices for all the objectives together?

f0uriest: Not empty, it will just be a matrix of zeros of the appropriate size if that constraint doesn't take that arg.

f0uriest: but otherwise yes

f0uriest: `A` itself is a list, one for each constraint, and each element of that is a dictionary of the jacobian wrt each arg

desc/objectives/utils.py
```diff
@@ -237,95 +246,63 @@ def factorize_linear_constraints(constraints, objective_args):  # noqa: C901
...
+        # some values might be fixed, but they still show up in other constraints
+        # this is where the fixed cols have >1 nonzero val
+        # for fixed variables, we delete that row and col of A, but that means
+        # we need to subtract the fixed value from b so that the equation is balanced.
+        # eg 2 x1 + 3 x2 + 1 x3= 4 ;    x1 = 1.5
+        # combining gives 3 x2 + 1 x3 = 1, with x1 now removed
+        b_full = put(
+            b_full,
+            unfixed_rows,
+            b_full[unfixed_rows]
```
dpanici, line 295: did we not do this subtraction before?

f0uriest: no, we didn't need to because all the constraints were uncoupled

dpanici: maybe I am missing something but what makes them coupled in this PR now?

desc/objectives/objective_funs.py
dpanici: Could we make it more robust by adding a try/except block to the line that would fail in the common cases? like for this case:
```python
eq = Equilibrium(L=3,M=3)
f_obj = ForceBalance()
obj = ObjectiveFunction(f_obj)
eq.solve(maxiter=5,objective=obj)


eq.change_resolution(L=5,M=5)
obj.build(eq)
eq.solve(maxiter=5,objective=obj)
```
the error is in line `165         x0 = objective.x(eq)` in desc/optimize/optimizer.py and is a `ValueError`

and for here
```python
eq = Equilibrium(L=3,M=3)
f_obj = ForceBalance()
obj = ObjectiveFunction(f_obj)
eq.solve(maxiter=5,objective=obj)
eq.change_resolution(L=5,M=5)
obj = ObjectiveFunction(f_obj)
obj.build(eq)
eq.solve(maxiter=5,objective=obj)
```

it is line `153             objective.compile(mode, verbose)` of desc/optimize/optimizer.py and is also a `ValueError`

something like
```python
            try:
                objective.compile(mode, verbose)
            except ValueError:
                objective.compile(mode, verbose)
```

I know this may not seem clean though but it would alleviate the double build issue you mention while also addressing the possible confusion the above cases could cause.

Though for this to actually work, we would need to add a `force_rebuild=False` flag to ObjectiveFunction so that if you call `ObjectiveFunction.build(eq,force_rebuild=True)`, it will rebuild every `Objective` it contains even if they are built already.

My vote would be in favor of this approach, or if we don't want to have these try/except blocks, then still including the `force_rebuild` flag with a default `False` and adding something in the description of `ObjectiveFunction` that if the eq resolution is changed one must rebuild with force_rebuild, or recreate the underlying objectives.

desc/objectives/utils.py
f0uriest: The new self consistency constraints have the form of `A*X - Xb` = 0`, so they couple `X` and `Xb`

desc/objectives/objective_funs.py
f0uriest: I think a better option would be to somehow automatically check if something needs to be built within the build method, so you could call it multiple times and have it do nothing if not needed. We could probably do something like cache the resolution properties of the eq when its first built and then compare that to a new eq if you call build again. 

Tricky thing might be profiles and grids. We should go over it at the meeting later.

f0uriest: OK for now I went with the try/catch since that covers most of the use cases and is fairly simple. Also added tests for the above use cases.

Eventually we should come up with a better option though.

dpanici (approved):
Edit: I like the try/except and we can deal with future edge cases later, but holding off on approval until the tests pass

desc/objectives/linear_objectives.py
```diff
@@ -18,7 +18,161 @@
...
+            Level of output.
+
+        """
+        modes = eq.surface.R_basis.modes
+        idx = np.arange(eq.surface.R_basis.num_modes)
+
+        self._dim_f = idx.size
+        self._A = np.zeros((self._dim_f, eq.R_basis.num_modes))
+        for i, (l, m, n) in enumerate(eq.R_basis.modes):
+            if eq.bdry_mode == "lcfs":
```
ddudt, line 83: What about the Poincare cross-section BC case? We should at least throw a `NotImplementedError`.

ddudt (dismissed):

desc/objectives/linear_objectives.py
dpanici: clever error message

dpanici (approved):

PR #461, opened 2023-03-16
Add missing attributes to new classes

f0uriest:


dpanici (approved):
thanks for catching this

PR #463, opened 2023-03-17
Verification against pyQIC

dpanici:
pyQIC has slight differences from pyQSC that requires some changes to the current code to have the NAE constraints work correctly

- pyQIC phi grid is shifted slightly from being `np.linspace(0,2*np.pi,nphi), so makes sure that the correct grid nodes in phi are used when fitting the near-axis surfaces for the initial DESC equilibrium
- fixes Psi sign issues that can arise btwn pyQSC/pyQIC and DESC


dpanici:
even with current fixes, there are disagreements with pyQIC we need to sort out. The |B| mismatch and iota make me thing that the way I fit the NAE toroidal Fourier coefficients is assuming something that is true in pyQSC but not in pyQIC

iota on axis not matching the NAE iota:
image


B strength on axis not matching the NAE input,:
image


lambda not converging unless N is high:
image

dpanici:
TODO:
- [ ] decide on how to treat when NAE constraints are constraining a mode to be ~0 (O(1e-13)), like can happen if a stellarator asymmetric NAE eq (with asym ONLY appearing at O(r^2) like in example used in QIC test), but O(r) constraints being used (where the coefficients will be stell sym)

f0uriest (approved):

f0uriest (approved):

PR #468, opened 2023-03-29
Use jax for perturbation stuff

f0uriest:
Previously we were using regular numpy for a lot of the operations in `perturb`, especially for factorizing the jacobian. This seems to be a major bottleneck on GPUs and might explain the low GPU utilization we were seeing. For reference, when running on an A100, the optimizer (using `jnp`) takes ~7 seconds per step, which includes computing and factorizing the jacobian. The perturb part using regular `np` took ~3 mins to do basically the same thing (the 2nd order stuff added negligible time after compilation)

### DIFF
--- a/desc/perturbations.py
+++ b/desc/perturbations.py
@@ -2,10 +2,9 @@
 
 import warnings
 
-import numpy as np
 from termcolor import colored
 
-from desc.backend import put, use_jax
+from desc.backend import jnp, put, use_jax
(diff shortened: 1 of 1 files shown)


dpanici:
wow this seems like it should give a big speedup

dpanici (approved):
running the gpu benchmark on della but this seems good to merge in withouot waiting for the benchmark result

ddudt (approved):

PR #469, opened 2023-03-30
Dp/nae stell asym

dpanici:
When the NAE equilibrium is stellarator asymmetric bc of O(r^2) effects, the DESC equilibrium fit is also asymmetric. BUT, the NAE O(r) constraints are stellarator symmetric. 

This leads to the first problem, which is that the constraints will be setting the stellarator asymmetric parts of the equilibrium's O(r) coefficients to roughly 0, which can cause some weirdness when solving

To allevaite the issue I thought to drop any NAE constraints whose target was below a threshold value (of like 1e-12 or so). However, this means when solving, the equilibrium's stellarator asymmetric coefficients are free to move as they wish, and this results in a completely different looking equilibrium than the NAE equilibrium (as expected I guess, since there is so much more freedom that we are not constraining).

I also need to test that the NAE constraints work as intended with O(r) stellarator asymmetric equilibria. 

This PR then is attempting to find the best way to address each of the above issues, and add tests for stellarator asymmetric equilibria. Suggestions are welcome for how to fix these.


dpanici:
Test this again @dpanici

dpanici:
image
iota is off near-axis, so the non-stell-sym O(r) constraints seem to be off somehow

PR #470, opened 2023-03-30
Add Contract Equilibrium Utilities

### LINKED ISSUE #467: Add utilities for expanding or contracting equilibrium by redefining flux surface labels

dpanici:
- Adds a new utility function ``desc.equilibrium.contract_equilibrium`` which takes in an ``Equilibrium`` object and an argument ``inner_rho``, and returns a new ``Equilibrium`` with original ``Equilibrium``'s ``inner_rho`` flux surface as its boundary.

Resolves #467 

image


dpanici:
@f0uriest what did you have in mind for the expand utility? it is not as straightforward as contraction because it is not obvious how one should continue the surface, and extrapolate the profiles (pressure may become negative, etc)

dpanici:
for vacuum could just do this, ignoring profiles, and use this to evaluate NAE at a low aspect ratio where the pyQSC rootfind fails, and just expand the higher aspect ratio one out to get it to the correct aspect ratio

dpanici:
@ddudt add your contract util here

dpanici:
Maybe use constant offset surface for expansion

dpanici:
just do contract @dpanici

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+    """Return a new profile whose rho=1 val is value of the old profile at inner_rho.
+
+    Args
+    ----
+        profile (Profile): Profile to scale
+        inner_rho (float): rho to take as the new LCFS rho for the new profile
+    Returns
+    -------
+        new_profile (Profile): new profile whose last value is the inner_rho value of
+        the original profile
```
ddudt, line 198: This doesn't match our usual docs formatting

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+        new_profile = PowerSeriesProfile(params=params_list, modes=modes_list)
+    elif isinstance(profile, SplineProfile):
+        n_knots = profile.params.size
+        new_knots = np.linspace(0, 1, n_knots, endpoint=True)
+        new_vals = profile(np.linspace(0, inner_rho, n_knots, endpoint=True))
+        new_profile = SplineProfile(
+            values=new_vals, knots=new_knots, method=profile._method
+        )
+    else:
+        raise NotImplementedError(
```
ddudt, line 216: Instead of implementing logic specific to each profile type, we can just do `Profile.from_values`

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+    ----
+        eq (Equilibrium): Equilibrium to contract.
+        inner_rho (float): rho value (<1) to contract the Equilibrium to
+
+    Returns
+    -------
+        eq_inner: New Equilibrium object, contracted from the old one such that
+            eq.pressure(rho=inner_rho) = eq_inner.pressure(rho=1), and
+            eq_inner LCFS = eq's rho=inner_rho surface.
+            Note that this will not be in force balance, and so must be re-solved.
```
ddudt, line 235: This doesn't match our usual docs formatting

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+    surf_inner.rho = 1.0
+    from .equilibrium import Equilibrium
+
+    eq_inner = Equilibrium(
+        surface=surf_inner,
+        pressure=pressure,
+        iota=iota,
+        current=current,
+        Psi=eq.Psi
+        * inner_rho**2,  # flux (in Webers) within the last closed flux surface
```
ddudt, line 258: It would be better to do `eq.compute("Psi")` at `inner_rho` in case we ever generalize the definition of `rho`

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+            values=new_vals, knots=new_knots, method=profile._method
+        )
+    else:
+        raise NotImplementedError(
+            f"Expected PowerSeriesProfile or SplineProfile, got {type(profile)}"
+        )
+    return new_profile
+
+
+def contract_equilibrium(eq, inner_rho):
```
ddudt, line 222: Might be worth adding a `copy` argument to control if this modifies the original Equilibrium or returns a new copy

tests/test_equilibrium.py
```diff
@@ -369,3 +374,50 @@ def test_backward_compatible_load_and_resolve():
     f_obj = ForceBalance(eq=eq)
     obj = ObjectiveFunction(f_obj, use_jit=False)
     eq.solve(maxiter=1, objective=obj)
+
+
+@pytest.mark.unit
+def test_contract_equilibrium():
```
ddudt, line 380: This only tests that the values at the new boundary match the values at the old internal surface. Suggestion for a new test: do a plotting image test with `plot_comparison((eq, eq_half_rho))` to check that the flux surfaces actually match on all of the surfaces

ddudt (changes requested):
I can help implement these changes I suggested

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+    ----
+        profile (Profile): Profile to scale
+        inner_rho (float): rho to take as the new LCFS rho for the new profile
+    Returns
+    -------
+        new_profile (Profile): new profile whose last value is the inner_rho value of
+        the original profile
+    """
+    params_list = []
+    modes_list = []
```
VascoSch92, line 201: these just list are just used in the first if. I will replace them inside the if for clarity.

Moreover, you can create them using a list compression.

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+        inner_rho (float): rho value (<1) to contract the Equilibrium to
+
+    Returns
+    -------
+        eq_inner: New Equilibrium object, contracted from the old one such that
+            eq.pressure(rho=inner_rho) = eq_inner.pressure(rho=1), and
+            eq_inner LCFS = eq's rho=inner_rho surface.
+            Note that this will not be in force balance, and so must be re-solved.
+    """
+    errorif(not (inner_rho < 1 and inner_rho > 0), ValueError, "inner_rho should be <1")
```
VascoSch92, line 237:
```suggestion
    errorif(not (inner_rho < 1 and inner_rho > 0), ValueError, f"inner_rho should be between  0 and 1 (not included), but got {inner_rho}")
```
the error message doesn't match what are you checking

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+    iota = None
+    if eq.iota is not None:
+        iota = scale_profile(eq.iota, inner_rho)
+    elif eq.current is not None:
+        current = scale_profile(eq.current, inner_rho)
+
+    surf_inner = eq.get_surface_at(rho=inner_rho)
+    surf_inner.rho = 1.0
+    from .equilibrium import Equilibrium
+
```
VascoSch92, line 251:
```suggestion
    surf_inner.rho = 1.0
    
    from .equilibrium import Equilibrium
```

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,100 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+    errorif(not (inner_rho < 1 and inner_rho > 0), ValueError, "inner_rho should be <1")
+    # create new profiles for contracted equilibrium
+    # pressure
+    pressure = scale_profile(eq.pressure, inner_rho)
+    current = None
+    iota = None
+    if eq.iota is not None:
+        iota = scale_profile(eq.iota, inner_rho)
+    elif eq.current is not None:
+        current = scale_profile(eq.current, inner_rho)
```
VascoSch92, line 246: are those two mutually exclusive?

tests/test_equilibrium.py
dpanici: It also tests if |B| and |F| are the same, which are functions of the surface geometry. I can add R Z and lambda as well to the tests

desc/equilibrium/utils.py
ddudt: But instead of throwing this `NotImplementedError` we should do `Profile.from_values` in this `else` block. That way the code will still work for other profile types. And maybe throw a warning instead. 

Also what you are doing for the `SplineProfile` is equivalent to fitting `from_values`, so it's really only the `PowerSeriesProfile` that we can do analytically.

tests/test_equilibrium.py
ddudt: My point is that we should also test at some intermediate surface, not just the boundary. If you add another couple of surfaces to the comparison in this test I'll be happy.

ddudt (changes requested):

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,135 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+        )
+    return new_profile
+
+
+def contract_equilibrium(eq, inner_rho, copy=True):
+    """Create a new equilibrium by using an inner surface of the passed-in equilibrium.
+
+    Parameters
+    ----------
+        eq : Equilibrium
```
f0uriest, line 230: too many indents here

desc/equilibrium/utils.py
```diff
@@ -181,3 +183,135 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+    errorif(
+        not (inner_rho < 1 and inner_rho > 0),
+        ValueError,
+        f"inner_rho should positive and <1, instead got {inner_rho}",
+    )
+
+    # create new profiles for contracted equilibrium
+    errorif(
+        eq.anisotropy,
+        NotImplementedError,
```
f0uriest, line 253: i think it should work fine if you do `Profile.from_values` as dan suggested

f0uriest (changes requested):

tests/test_equilibrium.py
dpanici: testing `data_keys = ["|B|", "|F|", "R", "Z", "lambda", "p", "iota"]` in the volume of the smaller equilibrium is literally testing at multiple surfaces, so this test covers what you want.

desc/equilibrium/utils.py
dpanici: does `SplineProfile` have a `from_values` method? I do not see one

dpanici:
if profile type is MTanHProfile, this will return a new MTanHProfile fit of the original profile values up to the specified inner_rho, which is not exactly the intention right, since the new profiile would have a pedestal still at the new edge just due to the functional form of the fit?
 we would actually rather have the returned profile be the original profile but just cut off at `inner_rho`...

desc/equilibrium/utils.py
```diff
@@ -181,3 +188,165 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+    ----------
+        profile : Profile
+            Profile to scale
+        inner_rho: float
+            rho to take as the new LCFS rho for the new profile
+    Returns
+    -------
+        new_profile : Profile
+            new profile whose last value is the inner_rho value of
+        the original profile
```
ddudt, line 206: indent

desc/equilibrium/utils.py
```diff
@@ -181,3 +188,165 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
     else:
         raise TypeError("Got unknown axis type {}".format(axis))
     return axis
+
+
+def scale_profile(profile, inner_rho, eq):
```
ddudt, line 193: I still think this whole function is unnecessary and could be replaced with `Profile.from_values` for all profile types. We should enforce `from_values` as a required method of every profile class. The only ones that are missing are `SplineProfile` (which is trivial to implement) and `TwoPowerProfile`

tests/test_equilibrium.py
```diff
@@ -375,3 +380,67 @@ def test_backward_compatible_load_and_resolve():
...
+
+
+@pytest.mark.unit
+def test_scale_profile():
+    """Test scale_profile utility function."""
+    eq = get("SOLOVEV")
+    p = eq.pressure
+    p_spline = p.to_spline(method="cubic2")
+    p_zern = p.to_fourierzernike()
+    p_mtanh = p.to_mtanh(order=4, ftol=1e-12, xtol=1e-12)
```
ddudt, line 429: Missing `PowerSeriesProfile` and `TwoPowerProfile`. We should make this test do that parameterization thing where it tests on a list of classes

ddudt (changes requested):

tests/test_equilibrium.py
dpanici: `p` is already a powerseries profile

dpanici:
Add option to NOT "contract" the profile, as a fallback (or to always make a spline as a fallback) or the user can also specify this.

desc/equilibrium/utils.py
ddudt, line 1: Should we put this new function in `desc/compat.py` instead? I think it's thematically more similar to the functions there, which edit the geometry without actually changing the equilibrium.

desc/equilibrium/utils.py
```diff
@@ -182,3 +184,124 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+
+    def scale_profile(profile, rho):
+        if profile is None:
+            return profile
+        is_power_series = isinstance(profile, PowerSeriesProfile)
+        if contract_profiles and is_power_series:
+            # only PowerSeriesProfile both
+            # a) has a from_values
+            # b) can safely use that from_values to represent a
+            #    subset of itself.
```
ddudt, line 234: Technically the `SplineProfile` also meets these two criteria, and we shouldn't throw a warning if the profile is already a spline

desc/equilibrium/utils.py
```diff
@@ -182,3 +184,124 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+            return profile.from_values(x=x, y=y)
+        elif contract_profiles:
+            warnif(
+                not isinstance(profile, SplineProfile),
+                UserWarning,
+                f"{profile} is not a PowerSeriesProfile or SplineProfile,"
+                " so cannot safely contract using the same profile type."
+                "falling back to fitting the values with a SplineProfile",
+            )
+            x = np.linspace(0, 1, eq.L_grid)
```
ddudt, line 247: We should probably add a kwarg for the user to specify how many spline points to use in this fit. The default can still be `eq.L_grid`

desc/equilibrium/utils.py
```diff
@@ -182,3 +184,124 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+        iota=iota,
+        current=current,
+        electron_density=electron_density,
+        electron_temperature=electron_temperature,
+        ion_temperature=ion_temperature,
+        atomic_number=atomic_number,
+        anisotropy=anisotropy,
+        Psi=float(
+            eq.compute("Psi", grid=LinearGrid(rho=inner_rho, NFP=eq.NFP))["Psi"][0]
+        ),  # flux (in Webers) within the new last closed flux surface
```
ddudt, line 280: Technically we don't need to include NFP here, since the compute function for `"Psi"` is just $\Psi = \Psi_0 \rho^2$. So for simplicity we could use `grid=Grid([inner_rho, 0, 0])`

ddudt (changes requested):

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ for compatibility with other codes which expect such files from the Booz_Xform c
 - Allows specification of Nyquist spectrum maximum modenumbers when using ``VMECIO.save`` to save a DESC .h5 file as a VMEC-format wout file
 - Adds a new objective ``desc.objectives.ExternalObjective`` for wrapping external codes with finite differences.
 - DESC/JAX version and device info is no longer printed by default, but can be accessed with the function `desc.backend.print_backend_info()`.
+- Adds a new utility function ``desc.equilibrium.contract_equilibrium`` which takes in an ``Equilibrium`` object and an argument ``inner_rho``, and returns a new ``Equilibrium`` with original ``Equilibrium``'s ``inner_rho`` flux surface as its boundary.
```
ddudt, line 33: This need to be updated now that we moved the function to `desc/compat.py`

tests/test_compat.py
```diff
@@ -323,3 +330,118 @@ def test_rotate_zeta():
...
+@pytest.mark.unit
+def test_contract_equilibrium_warns_errors():
+    """Test contract_equilibrium utility function warning for profiles."""
+    eq = get("ESTELL")
+    rho = 0.5
+    eq.pressure += eq.pressure
+    with pytest.warns(UserWarning, match="SplineProfile"):
+        eq_half_rho = contract_equilibrium(eq, rho, profile_num_points=50)
+    eq.anisotropy = FourierZernikeProfile()
+    with pytest.raises(ValueError, match="FourierZernike"):
```
ddudt, line 421: This test is failing because it does not trigger the correct warning. You added an `anisotropy` profile, but the `pressure` profile gets checked first, so it gives the same warning as in the line above. The easiest solution would be to change this to `eq.pressure = FourierZernikeProfile()` instead of setting the anisotropy.

ddudt (changes requested):

YigitElma (approved):
To be honest, I am not happy with the name but if people are fine with it, no problem, it is my nitpicking. Approving anyway

YigitElma (approved):

YigitElma (dismissed):

ddudt (dismissed):

desc/compat.py
```diff
@@ -330,3 +331,135 @@ def _get_new_coeffs(fun):
     eq_rotated.axis = eq_rotated.get_axis()
 
     return eq_rotated
+
+
+def contract_equilibrium(
```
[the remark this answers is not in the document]
ddudt: I would be fine with "crop", but I think "contract" is fine. Originally we were also planning to have an "expand" option, so contract/expand went well together.

desc/compat.py
```diff
@@ -330,3 +331,135 @@ def _get_new_coeffs(fun):
...
+        pressure=pressure,
+        iota=iota,
+        current=current,
+        electron_density=electron_density,
+        electron_temperature=electron_temperature,
+        ion_temperature=ion_temperature,
+        atomic_number=atomic_number,
+        anisotropy=anisotropy,
+        Psi=float(
+            eq.compute("Psi", grid=LinearGrid(rho=inner_rho))["Psi"][0]
```
f0uriest, line 437: can't we just do `Psi=eq.Psi*inner_rho`

YigitElma: There was a discussion about generalizing the flux label, should we add a note?

YigitElma (dismissed):

f0uriest (approved):

YigitElma (approved):

f0uriest (approved):

PR #472, opened 2023-03-30
change benchmark functions' perturb to use new syntax

dpanici:


f0uriest (approved):
Benchmarks won't work till this is merged to master, since they checkout master and use that benchmark, but the ones on the current branch seem to work.

PR #473, opened 2023-03-30
Add pure isodynamic metric/objective

### LINKED ISSUE #436: Isodynamic objective function
It could be useful to add a compute & objective function for isodynamicity: 
$v_d \cdot \nabla \psi \propto (b \times \nabla \ln B) \cdot \nabla \psi$
(Eq. 64.5 in Helander 2014)

It won't be satisfiable but could still be interesting to try optimizing for.

f0uriest:
Adds compute function and objective for targeting isodynamicity (NOT quasi-isodynamicity), or enforcing radial drift vanishes at each point rather than just on average. Might not be a good optimization target but could be a useful metric or possibly combined with other stuff like QS.

Resolves #436


desc/compute/_field.py
ddudt, line 1: Can we move this function to `compute/_qs.py`? That way it will be with all of the other functions that are proxies for particle drifts, like QS and QI. Maybe we could rename the file to something more general than QS.

desc/objectives/_qs.py
```diff
@@ -508,3 +508,136 @@ def compute(self, *args, **kwargs):
...
+    grid : Grid, ndarray, optional
+        Collocation grid containing the nodes to evaluate at.
+    name : str
+        Name of the objective function.
+
+    """
+
+    _scalar = False
+    _linear = False
+    _units = "(~)"
```
ddudt, line 550: In other objectives we have been using `_units = "(dimensionless)"`. I don't care either way, but we should be consistent.

desc/objectives/_qs.py
```diff
@@ -508,3 +508,136 @@ def compute(self, *args, **kwargs):
...
+    _print_value_fmt = "Isodynamicity error: {:10.3e} "
+
+    def __init__(
+        self,
+        eq=None,
+        target=0,
+        bounds=None,
+        weight=1,
+        normalize=False,
+        normalize_target=False,
```
ddudt, line 560: The default for all of our dimensionless objectives is `normalize=False`, like we have here. That is fine for scalar objectives (like aspect ratio, etc.) where the normalization doesn't do anything. But for non-scalar objectives like this one, it could still be useful to normalize by the number of collocation points: `self._normalization = 1 / jnp.sqrt(self._dim_f)`

ddudt (changes requested):
We should add tests for this new objective.

desc/objectives/_qs.py
```diff
@@ -508,3 +508,134 @@ def compute(self, *args, **kwargs):
...
+        ----------
+        eq : Equilibrium, optional
+            Equilibrium that will be optimized to satisfy the Objective.
+        use_jit : bool, optional
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        if self.grid is None:
```
ddudt, line 589: Remember to update this objective with the new `self._grid` instead of `self.grid` syntax introduced in PR #459. And similar for `self._transforms["grid"]` instead of `self.grid`.

ddudt (dismissed):

ddudt (approved):

dpanici (approved):

PR #476, opened 2023-03-31
Make cos(mt-nz), sin(mt-nz) -> cos(mt)sin(mz) etc. conversion more general

dpanici:
- add fix for vmec utility bug that would cause xn,xm to be empty if the Fourier modes given had only sin symmetry
- Also generalized conversion to work for inputs where the first mode is not the `m=0, n=0` mode


dpanici:
Further fixes:

- [ ] If the VMEC mode number basis does NOT have the 0,0 mode as the first mn, (or is lacking it, for example if it is a sin symmetric series), then our util functions will fail as they are off by an index (https://github.com/PlasmaControl/DESC/blob/f67a7d9475e0d97870c9aa5bf3022b3c6907850c/desc/vmec_utils.py#L97). Either update docstring to reflect this and add check when using these functions, or update function to work for arbitrary inputs

f0uriest (approved):

PR #480, opened 2023-04-05
fix vartheta bug

### LINKED ISSUE #477: Plot vartheta lines look off on latest master
on current master
image

before current master (`git checkout b73756b`)

image

probably related to the symmetry stuff?

unalmis:
Resolves #477


dpanici (changes requested):
Can you add a `test_plot_surfaces_HELIOTRON` so we can catch changes like this in the future?

ddudt (approved):

desc/grid.py
```diff
@@ -199,8 +199,11 @@ def _create_nodes(self, nodes):
 
         """
         nodes = np.atleast_2d(nodes).reshape((-1, 3)).astype(float)
-        nodes[nodes[:, 1] != 2 * np.pi, 1] %= 2 * np.pi
-        nodes[nodes[:, 2] != 2 * np.pi / self.NFP, 2] %= 2 * np.pi / self.NFP
+        # Do not alter nodes given by the user for custom grids.
+        # In particular, do not modulo nodes by 2pi or 2pi/NFP.
+        # This may cause the surface_integrals() function to fail recognizing
+        # surfaces outside the interval [0, 2pi] as duplicates. However, most
+        # surface integral computations are done with LinearGrid anyway.
```
unalmis, line 206: Basically, if you call surface_integrals on a custom Grid that has say three duplicate nodes at theta=pi/4, pi/4 + 2pi, pi/4 + 4 pi, then the returned array will have the correct value of the computation at index corresponding to theta=pi/4 and will have a value of 0 at the indices corresponding to theta=pi/4 + 2 pi, + 4pi.
Only the integrals of the surface that is duplicated would be affected. Rho surface and zeta surface integrals would not be affected (even though these surfaces would have duplicate nodes in the theta coordinate) in this example.

Surface integrals usually use one of the predefined types, so I don't think this is a big deal.

f0uriest (approved):

dpanici (changes requested):
I'd like to see a warning for the surface integral case that could give wrong answers, but besides that looks great

desc/grid.py
f0uriest: I would vote against a warning for this edge case. The only time the base `Grid` class gets used is in some of the custom root finding stuff we do, where we don't want to have it throwing warnings at every iteration if at the end it gets the right answer. 

If a user uses a custom grid for computing stuff using `eq.compute` the logic in there will ensure that averages are done correctly.

dpanici: Ok that is fair

dpanici (approved):

PR #481, opened 2023-04-05
Use consistent normalizations when plotting.

f0uriest:
- We currently normalize force by either thermal or magnetic pressure
- However, this means that low beta cases can look worse than they really are, since we're normalizing by a small quantity.
- This change makes it so we always use the magnetic pressure, which is well defined and of reasonable magnitude in all cases of interest, and is also consistent with the way we normalize the objective functions


dpanici (changes requested):
This is ok, a small part of me is unsure of changing it like this bc in my paper all the normalizations/plots were wrt pressure though. Could we add a flag to ask to norm_F by pressure instead of defaulting to always using |B|? Just so I can make comparable plots without needing to rework the plot functions again.

tests/baseline/test_fsa_F_normalized.png
ddudt, line 1: need to change the y limits on this plot

ddudt (commented):
I don't love this idea. The point of normalizing by pressure is to give a percent error for JxB-grad(p). Even in a low beta case, normalizing by pressure still gives a useful relative error metric. Having the option for normalizing by magnetic pressure could be useful to add, but I don't think it should be the default.

f0uriest:
The main goal was just to have the normalizing quantity be "smooth" in some sense, instead of jumping from one value to another when we go from vacuum to finite pressure. Another option would be to use the min/max/mean etc of the different terms but that seems more complicated than it's worth.

dpanici:
@f0uriest yea something like that, so I can do `plot_section(eq,"|F|",norm_F=True, normalize_with_pressure=True)` and just have that flag be False by default

docs/variables.csv
```diff
@@ -237,6 +237,7 @@ Name,Label,Units,Description,Module
 ``grad(|B|^2)_zeta``,:math:`(\nabla B^{2})_{\zeta}`,Tesla squared,Covariant toroidal component of magnetic pressure gradient,``desc.compute._field``
 ``grad(|B|^2)``,:math:`\nabla B^{2}`,Tesla squared / meters,Magnetic pressure gradient,``desc.compute._field``
 ``|grad(|B|^2)|/2mu0``,:math:`|\nabla B^{2}/(2\mu_0)|`,Newton / cubic meter,Magnitude of magnetic pressure gradient,``desc.compute._field``
+``<|grad(|B|^2)|/2mu0>_vol``,:math:`\langle |\nabla B^{2}/(2\mu_0)| \rangle_{vol}`,Newtons per cubic meter,Volume average of magnitude of magnetic pressure gradient,``desc.compute._field``
 ``(curl(B)xB)_rho``,:math:`((\nabla \times \mathbf{B}) \times \mathbf{B})_{\rho}`,Tesla squared,Covariant radial component of Lorentz force,``desc.compute._field``
```
dpanici, line 241: missing magnetic tension vol avg entry

dpanici (dismissed):

f0uriest:
Leaving this here for reference:

plotting the different volume averaged pressures for all the equilibria in our examples

image


The two magnetic terms are usually of similar magnitude, and pressure as well when its not zero or really small

ddudt (changes requested):
Can we have all 3 normalization options? Magnetic pressure gradient can stay as the default, and we already have pressure gradient, so just need to add an option for magnetic tension. Instead of a bunch of flags like `normalize_with_pressure=True` we could use a string like `normalization="pressure"` etc.

dpanici (approved):

ddudt (approved):

PR #483, opened 2023-04-08
Dommaschk potentials implemented in JAX

dpanici:
Following [Dommaschk 1986 paper]( https://doi.org/10.1016/0010-4655(86)90109-8), implementing general Dommaschk potentials in JAX so can automatically differentiate to find the field

TODO:
- [x] Add tests for each function added
- [x] Add new MagneticField subclass DommaschkMagneticField which uses autodiff to grad(dommaschk_potential) to find the B field
- [x] replace CD_mk and CN_mk with the expressions in eqns 31 and 32 of the paper, so they are applicable for any m,k
- [ ] Possibly verify against simsopt implementation - I don't think this is needed, I also verified thru field tracing that we reproduce Fig. 2 in the 1986 dommaschk paper.
- [x] jaxify for loops in dommaschk potential functions
- [x] other ways to speed up evaluation or fitting?
- [x] implement stellarator symmetry option for fitting ( makes $a_{ml}=d_{ml}=0$ for even $l$ and $b_{m,l}=c_{m,l}=0$ for odd $l$, so effectively reduced number of modes by 2, useful for fitting)
- [x] put magnetic_fields in its own directory and break out dommaschk utility functions?

Pushing off improving the fit logic to #928 , this PR has been good to go for a while now, might as well get it in.


desc/magnetic_fields.py
```diff
@@ -1034,6 +1004,238 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
...
+    c_arr : 1D array-like of float
+        c_m_l coefficients of V_m_l terms, which multiply the cos(m*phi)*N_m_l-1 term
+    d_arr : 1D array-like of float
+        d_m_l coefficients of V_m_l terms, which multiply the sin(m*phi)*N_m_l-1 terms
+    B0: float
+        scale strength of the magnetic field's 1/R portion
+
+    """
+
+    def __init__(self, ms, ls, a_arr, b_arr, c_arr, d_arr, B0=1.0):
```
f0uriest, line 1035: Would it be possible to have default values for m,l,a,b,c,d? Like something that would just give a simple 1/R toroidal field?

desc/magnetic_fields.py
```diff
@@ -1034,6 +1004,238 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
...
+
+    @classmethod
+    def fit_magnetic_field(  # noqa: C901 - FIXME - simplify
+        cls, field, coords, max_m, max_l, sym=False, verbose=1
+    ):
+        """Fit a vacuum magnetic field with a Dommaschk Potential field.
+
+        Parameters
+        ----------
+            field (MagneticField or callable): magnetic field to fit
```
f0uriest, line 1072: could field here also be just the B field values at the coordinate locations?

desc/magnetic_fields.py
```diff
@@ -1034,6 +1004,238 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
...
+            verbose (int): verbosity level of fitting routine, > 0 prints residuals
+        """
+        # We seek c in  Ac = b
+        # A will be the BR, Bphi and BZ from each individual
+        # dommaschk potential basis function evaluated at each node
+        # c is the dommaschk potential coefficients
+        # c will be [B0, a_00, a_10, a_01, a_11... etc]
+        # b is the magnetic field at each node which we are fitting
+
+        if not isinstance(field, _MagneticField):
```
f0uriest, line 1090: `if callable(field)` might be clearer here

desc/magnetic_fields.py
```diff
@@ -1034,6 +1004,238 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
...
+                        abcd_zero_due_to_sym_inds[which_coef].append(0)
+                    else:
+                        abcd_zero_due_to_sym_inds[which_coef].append(1)
+
+                ms.append(m)
+                ls.append(l)
+        for i in range(4):
+            abcd_zero_due_to_sym_inds[i] = jnp.asarray(abcd_zero_due_to_sym_inds[i])
+        assert (
+            round(jnp.sum(len(a_s) + len(b_s) + len(c_s) + len(d_s))) == num_modes - 1
```
f0uriest, line 1165: why round and sum?

desc/magnetic_fields.py
```diff
@@ -1034,6 +1004,238 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
...
+        c_s = []
+        d_s = []
+        zero_due_to_sym_inds = []
+        abcd_zero_due_to_sym_inds = [
+            [],
+            [],
+            [],
+            [],
+        ]  # indices that should be 0 due to symmetry
+        for l in range(max_l + 1):
```
f0uriest, line 1138: can this be done with np.where type operations? might be easier to understand

desc/magnetic_fields.py
```diff
@@ -1034,6 +1004,238 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
...
+        X = jnp.asarray(X)
+
+        jac = jit(jacfwd(get_B_dom, argnums=1))(coords, X, params["ms"], params["ls"])
+
+        A = jac.reshape((rhs.size, len(X)), order="F")
+
+        # now solve Ac=b for the coefficients c
+
+        Ainv = scipy.linalg.pinv(A)
+        c = jnp.matmul(Ainv, rhs)
```
f0uriest, line 1219: `jnp.linalg.lstsq` would be more stable here, and automatically gives residuals, condition numbers etc.

desc/magnetic_fields.py
```diff
@@ -1087,6 +1289,319 @@ def odefun(rpz, s):
...
+def alpha(m, n):
+    """Alpha of eq 27, 1st ind comes from C_m_k, 2nd is the subscript of alpha."""
+    # modified for eqns 31 and 32
+
+    def false_fun(m_n):
+        m, n = m_n
+        return (-1) ** n / (gamma(m + n + 1) * gamma(n + 1) * 2.0 ** (2 * n + m))
+
+    def bool_fun(n):
+        return jnp.any(n < 0)
```
f0uriest, line 1318: do we need `jnp.any` in all of these? aren't m, n always scalars?

desc/magnetic_fields.py
dpanici: I think I was tired when writing this code as yea that is stupid of me to have wrote it like that...

desc/magnetic_fields.py
dpanici: I think you're right, I was just going trigger happy on the jnp statements when I converted it to JAX

dpanici:
@YigitElma these are a set of vacuum potentials i.e. each of these individually satisfies Laplace's equation, could be useful for getting analytic solutions of vacuum and seeing how their poincare XS compare to the full soln/ if you can just use the poincare XS to get the full field solution with DESC

f0uriest (approved):

f0uriest (approved):

desc/magnetic_fields/_dommaschk.py
```diff
@@ -0,0 +1,520 @@
...
+                    and output (num_nodes,3) as the B field in cylindrical rpz basis.
+                if ndarray, must be an ndarray of the magnetic field in rpz,
+                    of shape (num_nodes,3) with the columns being (B_R, B_phi, B_Z)
+            coords (ndarray): shape (num_nodes,3) of R,phi,Z points to fit field at
+            max_m (int): maximum m to use for Dommaschk Potentials
+            max_l (int): maximum l to use for Dommaschk Potentials
+            sym (bool): if field is stellarator symmetric or not.
+                if True, only stellarator-symmetric modes will
+                be included in the fitting
+            verbose (int): verbosity level of fitting routine, > 0 prints residuals
```
ddudt, line 99: looks like this is over-indented, might cause issues for the docs?

ddudt (approved):
Approving, but check docstring formatting

ddudt (approved):

ddudt (approved):

PR #484, opened 2023-04-10
Objective.compute_scalar independent of grid resolution

### LINKED ISSUE #482: Objective weights/normalizations
We should review how we are weighting and normalizing our objective functions. Questions to consider and check the implementation of: 
- For "local" errors, do we want the sum of squares (scalar) error to be proportional to the error integrated over the volume? 
- Should weights/normalizations be applied in the `compute` method, or the `compute_scalar` method? 
- Should all objective values be independent of the grid resolution, even if the user specifies `normalize=False`? (This is not currently the case, but I think it should be)
- What terms should be considered "weights" and which terms should be considered "normalizations"? (I propose normalizations should only be making the output units dimensionless, not making them independent of resolution)

### LINKED ISSUE #589: printing objective sum of squares error
`ObjectiveFunction.print_value` does not always print the true sum of squares error. 

Minimal working example: 
```
from desc.objectives import ObjectiveFunction, AspectRatio

eq = Equilibrium(L=2, M=2, N=0)
obj = ObjectiveFunction(AspectRatio(eq, target=7))
eq.solve(objective=obj, verbose=3)
```

Gives the output:
```
Starting optimization
Using method: lsq-exact
       0              1          4.500e+00                                    5.294e-15   
`gtol` condition satisfied.
         Current function value: 4.500e+00
         Total delta_x: 0.000e+00
         Iterations: 0
         Function evaluations: 1
         Jacobian evaluations: 1
Timer: Solution time = 210 ms
Timer: Avg time per step = 210 ms
Start of solver
Total (sum of squares):  5.000e+01, 
Aspect ratio:  3.000e+00 (dimensionless)
Aspect ratio:  3.000e+00 (normalized)
```

The initial aspect ratio is 10 and the target value is 7, so the error is 3. The corresponds to the cost of `1/2 * 3**2 = 4.5`. 
But the printed sum of squares is `1/2 * 10**2 = 50`. 

We should edit the `print_value` function so it gives the same sum of squares value as the optimizer displays.

ddudt:
This PR attempts to resolve most of the problems raised in Issue #482. Most importantly, it makes sure that all objectives with "local" errors now have their `compute_scalar` values roughly independent of the grid resolution, both when normalized and not. 

These changes also have `Objective.normalization` now only used to make the values dimensionless; it removes the need to reference `Objective.dim_f` in the compute logic. 

Also updates `Objective.print_value` to print the min/max/mean of the error instead of just the total, and uses abs error when target is 0, and only prints total error for linear objectives (assumed they're used as exactly satisfied constraints so we don't need lots of details). A

Printing should also be more intuitive, printing the actual value of the aspect ratio, volume etc rather than the error.

Resolves #482 
Resolves #589


f0uriest:
Can we make a separate PR for [f429b20](https://github.com/PlasmaControl/DESC/pull/484/commits/f429b20fe65d287b49d3ab6666f55e4accfe0e0f)? Doing it here messes up the diff and makes it hard to see what actually changed.

ddudt:
> Can we make a separate PR for all the alphabetization stuff? Doing it here messes up the diff and makes it hard to see what actually changed.

Yes, but you specifically said at the last dev meeting to just start alphabetizing as we edit files for other reasons...

f0uriest:
> > Can we make a separate PR for all the alphabetization stuff? Doing it here messes up the diff and makes it hard to see what actually changed.
> 
> Yes, but you specifically said at the last dev meeting to just start alphabetizing as we edit files for other reasons...

yeah that was my bad, I forgot that git sometimes doesn't recognize what changed in the most intuitive way.

desc/compute/_bootstrap.py
```diff
@@ -312,7 +312,7 @@ def j_dot_B_Redl(
 @register_compute_fun(
     name="<J*B> Redl",
     label="\\langle\\mathbf{J}\\cdot\\mathbf{B}\\rangle_{Redl}",
-    units="T A m^{-2}",
+    units="N / m^{3}",
```
f0uriest, line 315: I prefer the units as T*A/m^2. Although this has units of Newtons, its not really a force

f0uriest:
Reweighting the force balance objective seems to give generally way worse results:
image

image

We may want to reconsider some of the stuff in this PR. If the goal is just to make things roughly independent of grid resolution it might be better just to throw in a scalar fudge factor without rescaling the relative weighting between nodes.

f0uriest (approved):

dpanici (changes requested):
do we want a changelog entry for this?

tests/test_examples.py
```diff
@@ -500,7 +500,7 @@ def test_simsopt_QH_comparison():
     # Cancel factor of 1/2 in desc objective which is not present in simsopt:
     aspect_weight = np.sqrt(2)
     # Also scale QS weight so objective is approximately independent of grid resolution:
```
dpanici, line 502: why this factor, if it is not actually something that is related to the grid anymore?

f0uriest: there's still a difference between us and simsopt, due to our theta/zeta being [0,2pi] and VMEC u,v being [0,1], and an extra factor of 1/2, but I realize I forgot a square root so ill add that.

desc/objectives/objective_funs.py
```diff
@@ -938,6 +942,24 @@ def build(self, eq=None, use_jit=True, verbose=1):
...
         self._set_dimensions(eq)
         self._set_derivatives()
+
+        # set quadrature weights if they haven't been
+        if hasattr(self, "_constants") and ("quad_weights" not in self._constants):
+            if self._coordinates == "":
+                w = jnp.ones((self.dim_f,))
+            elif self._coordinates == "rtz":
+                w = self._constants["transforms"]["grid"].weights
+                w *= jnp.sqrt(self._constants["transforms"]["grid"].num_nodes)
```
dpanici, line 952: For my own understanding, is the logic behind this method of making independent of resolution as follows

We have some error vector $f_i$ with $i=0->N$ where $N$ is num_nodes.

when we scale, we multiply by `grid.weights`=$d\rho d\theta d\zeta$ and then by $\sqrt{N}$

`grid.weights` is $\approx (1/N^{1/3})(2\pi/{N^{1/3})(2\pi/{N^{1/3}) \approx 4\pi^2 / N$ if we assume that the grid points are linearly spaced and that the number of unique rho, theta, and zeta nodes are the same, so $N_{rho} = N^{1/3}$ etc. 

So when we square each scaled error and sum, we get something like $\sum_i f_i^2 * 16 \pi^4/N^2 N = 16\pi^4 \sum_i f_i^2 / N \approx 16 \pi^2 f_i^2/N * N \approx 16 \pi^2 f_i^2$ where I assumed $f_i$ is constant across grid points, and this is thus independent of grid resolution

dpanici (approved):

f0uriest (approved):

PR #485, opened 2023-04-12
Add interfaces for constrained optimizers

### LINKED ISSUE #418: Add constrained optimizers from scipy
- [ ] SLSQP
- [ ] trust_constr

f0uriest:
- Adds interfaces to `trust-constr` and `SLSQP` methods of `scipy.optimize.minimize`. These methods can handle general nonlinear constraints, though their performance on badly scaled problems like those encountered in stellarator optimi isn't great.
- `Objective` and `ObjectiveFunction` `compute` methods have now been separated into `compute_unscaled` which returns the raw physics value of the objective, `compute_scaled`, which returns the normalized value, and `compute_scaled_error` which returns the normalized difference between the physics value and the target/bounds. Similarly, `jac_scaled` and `jac_unscaled` are the relevant derivatives (note that `jac_scaled` is equivalent to `jac_scaled_error` as the constant target drops out). `grad` and `hess` methods still correspond to `compute_scalar` which returns the sum of squares of `compute_scaled_error`
- Adds new `ObjectiveFunction` attributes `target_scaled` and `bounds_scaled` which return vectors of the scaled values from each sub-objective.
- Adds automatic scaling of variables for `scipy.optimize.minimize` methods, using the hessian at the initial point.
- Ensures objectives don't have both bounds and target set at the same time. This occasionally caused issues if one of them had a default value without the user realizing.
- Adds a bunch of missing docstrings and args to linear objectives- Adds documentation on how to interface with new optimizers.
- Adds a documentation notebook with a simple example of using constrained optimizers
- Adds a table of optimizer info to docs

Resolves #418


desc/optimize/_constraint_wrappers.py
```diff
@@ -308,21 +373,15 @@ def __init__(
...
-            if not isinstance(
-                con,
-                (
-                    ForceBalance,
-                    RadialForceBalance,
-                    HelicalForceBalance,
-                    CurrentDensity,
-                ),
-            ):
+            if not con._equilibrium:
```
ddudt, line 376: OK so this flag is true if it can't be used as a nonlinear constraint for the proximal projection? Does that only ever apply to the equilibrium objectives? If not, the flag should be renamed something more general/descriptive.

ddudt (dismissed):
see minor comments

docs/adding_optimizers.rst
```diff
@@ -0,0 +1,222 @@
...
+The ``register_optimizer`` decorator takes 6 arguments. In all cases the values can either
+be single entries or lists, to register multiple versions of the same basic algorithm.
+For example, here we register a standard version of ``ipopt`` that uses all derivative
+information, and ``ipopt-bfgs`` which only uses approximate Hessians. The necessary fields
+are:
+
+
+- ``name`` (``str``) : name you want to give the optimization method. This is what you will
+  pass to ``desc.optimize.Optimizer`` to select the given method.
+- ``description`` (``str``) : A short description of the method, with relevant links
```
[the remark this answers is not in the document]
f0uriest: what Do you mEan

desc/optimize/_constraint_wrappers.py
f0uriest: The way things are currently written, the projection method only works for equilibrium constraints. In theory we could figure out "projections" for other constraints but that's not the goal here. The `Objective._equilibrium` flag is just meant to be a replacement for the few places where we would list out all the different objectives that define an equilibrium condition.

desc/optimize/optimizer.py
```diff
@@ -146,15 +147,45 @@ def optimize(  # noqa: C901 - FIXME: simplify this
             linear_constraints = maybe_add_self_consistency(linear_constraints)
         if len(linear_constraints):
             objective = LinearConstraintProjection(objective, linear_constraints)
+            if nonlinear_constraint is not None:
+                nonlinear_constraint = LinearConstraintProjection(
```
dpanici, line 151: this is meant to trigger the error in the init, bc this means that a nonlinear constraint was not caught and wrapped by the earlier maybe_wrap?

desc/optimize/optimizer.py
```diff
@@ -146,15 +147,45 @@ def optimize(  # noqa: C901 - FIXME: simplify this
...
+                    nonlinear_constraint, linear_constraints
+                )
         if not objective.built:
             objective.build(eq, verbose=verbose)
+        if nonlinear_constraint is not None and not nonlinear_constraint.built:
```
dpanici, line 156: if the nonlinear constraint is not None at this point, doesn't that mean it wasn't wrapped properly? I am a bit confused at the logic

desc/optimize/optimizer.py
f0uriest: no this is just doing the linear constraint projection on top of the nonlinear constraint, so that the objective and constraint both take the same reduced state vector.

desc/optimize/optimizer.py
dpanici: My confusion is that in line 141
```python
 objective, nonlinear_constraint = _maybe_wrap_nonlinear_constraints(
            objective, nonlinear_constraint, self.method, options
        )
```
the `_maybe_wrap_nonlinear_constraints`  if there is  non-None `nonlinear_constraint` passed in, seems to return objective, None

```python
if wrapper is not None and wrapper.lower() in ["prox", "proximal"]:
        objective = ProximalProjection(
            objective,
            constraint=nonlinear_constraint,
            perturb_options=perturb_options,
            solve_options=solve_options,
        )
        nonlinear_constraint = None
    return objective, nonlinear_constraint
```
But ok maybe I see my confusion. This is just checking if the wrapper is proximal projection, at which point we fold the nonlinear constraint into the ProximalProjection objective which takes care of it after each solver iteration, so we no longer need to consider the nonlinear constraint. That is only something optimizer needs to consider if we are using like a constrained optimizer with a nonlinear constraint.

docs/adding_optimizers.rst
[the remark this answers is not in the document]
ddudt: All of the items in this list start without capitalization, except for this line. Then in the next list all of the items are properly capitalized. So I'm asking you to capitalize the items in this list.

desc/optimize/optimizer.py
f0uriest: Yes exactly

f0uriest:
Removing the constrained optimization example notebook for now, finding a case that works well with the scipy optimizers is difficult. Will add back in later when we get the augmented lagrangian going.

ddudt (dismissed):

docs/adding_optimizers.rst
```diff
@@ -0,0 +1,222 @@
...
+- ``objective`` (``ObjectiveFunction``) : Function to minimize.
+- ``constraint`` (``ObjectiveFunction``) : Constraint to satisfy.
+- ``x0`` (``ndarray``) : Starting point.
+- ``method`` (``str``) : Name of the method to use (this will be the same as the name
+  the method was registered with)
+- ``x_scale`` (``array_like``) : Characteristic scale of each variable.
+- ``verbose`` (``int``) : level of output to console - 0  : work silently,
+  1 : display a termination report, 2 : display progress during iterations
+- ``stoptol`` (``dict``) : Dictionary of stopping tolerances, with keys ``{"xtol", "ftol",
+  "gtol", "maxiter", "max_nfev", "max_njev", "max_ngev", "max_nhev"}``
```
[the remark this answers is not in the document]
ddudt: @f0uriest

dpanici (approved):

PR #487, opened 2023-04-13
Update install docs and Tutorial notebook

dpanici:
- Update installation docs with more specific instructions on how to install DESC with CPU or with CPU+GPU support, along with examples for clusters we've tested. 
- Fix warning in optimization notebook which occurs because `lsq-exact` is used with `eq.optimize`, when it needs to be `proximal-lsq-exact`
- Bump python version is `setup.py` to be `python>=3.8` to reflect the range of versions we test on


docs/installation.rst
```diff
@@ -22,42 +24,133 @@ Or from GitHub (for development builds)
...
+    # only need to do ONE of these conda env create commands, not both
+    ## standard build
     conda env create --file requirements_conda.yml
-    # developer build (if you want to run tests)
+    ## developer build (if you want to run tests)
     conda env create --file devtools/dev-requirements_conda.yml
+
+
     conda activate desc-env
+    python setup.py install
```
unalmis, line 43: In this comment, I use "DESC" to refer to the directory of that name and `desc` to refer to the actual `desc` python module or package.

From the slack it looks like the motivation to add the `python setup.py install` command is to avoid `desc module not found error`.
To avoid this error, my suggestion is to either
- Add the path to the DESC directory to your computer's Python path. The instructions for doing so may depend on operating system.
- Or, when you want to use `desc`, open your terminals from the DESC directory. This is equivalent to changing your terminal session to DESC with some `cd path/to/DESC`. I've always used this option. Note that Jupyter Notebooks will need the following preamble before using the `desc` module. For example,
```python
# do this
import sys
import os
relative_path_to_DESC_string = "../../"
sys.path.append(os.path.abspath(relative_path_to_DESC_string))
# before this
import desc.io
```

I do not recommend using `python setup.py install` as a means to add `desc` to the Python path.  It may confuse developers.

The command adds a frozen copy of the DESC directory you cloned from GitHub to your python path. Any changes you make as you develop code will be ignored.
For example, suppose you run `python setup.py install`, then later add `raise Exception("Prevent users from making Grids")` in the constructors of every grid in `grid.py`. Then suppose you start a terminal somewhere other than DESC path. You would expect the following code to raise that expection.
```python
conda activate desc-env
python
from desc.grid import *
lg = LinearGrid(L=2, M=2, N=2)
```
But it will not (I tested), because the `desc` module that is being used was frozen to modifications after the `python setup.py install` command.

Moreover, if you start a terminal from the DESC directory, then the code
```python
conda activate desc-env
python
from desc.grid import *
lg = LinearGrid(L=2, M=2, N=2)
``` 
will raise that expection.
This is because Python searches the local directory with highest priority for modules. In this case, the `desc` module that is being used are the live files in the DESC directory.

unalmis, line 43: Unrelated, but the `setup.py` file contains the line `python_requires="~=3.7"`. Should this be changed to `>= 3.8`? The `~=` specifier means 3.7 is compatible. https://peps.python.org/pep-0440/#version-specifiers

docs/installation.rst
```diff
@@ -22,42 +24,133 @@ Or from GitHub (for development builds)
...
-Or from GitHub (for development builds)
+Or from GitHub (for development builds):
 
 .. code-block:: sh
 
     git clone https://github.com/PlasmaControl/DESC.git
     cd DESC
 
-    module load anaconda  # load your python module
+    module load anaconda  # load your python module, this command may depending on cluster
```
unalmis, line 67: may vary

unalmis (changes requested):
Some comments about adding `python setup.py install`. Other than that looks great. Also some of the QS error plots look different. Is that just because the optimization algorithm changed?

dpanici:
> A comment about adding `python setup.py install`. Other than that looks great. Also some of the QS error plots look different. Is that just because the optimization algorithm changed?

The QS plots in the notebook I think were from back before we fixed a bug in the way f_boozer was calculated, as they now line up more with what our plot tests show.

unalmis (approved):

docs/installation.rst
unalmis: For the reasons discussed above, if you have used the command `python setup.py install`, I recommend uninstalling that. The only way to do this is by removing the environment it was installed in entirely. Then to install DESC, the instructions in this PR are fine. To make explicit:
```sh
conda activate base
# remove the environment where `python setup.py install` was used
conda remove --name desc-env --all
# install desc as usual
cd DESC
conda env create --file devtools/dev-requirements_conda.yml
conda activate desc-env
# if you want, add DESC directory to python path so that you can use `desc` module from any directory
export PYTHONPATH="${PYTHONPATH}:${HOME}/Documents/project/DESC/" # replace with your path to DESC directory
source ~/.bashrc
```

rahulgaur104: Ok, thanks Kaya! I did see the problem that you mentioned. When I added a debug statement to a DESC core script, ran python3 setup.py install, and ran the code. It didn't stop at the debug breakpoint. 

So if I install DESC using the procedure that you have just explained, make changes to the source, say putting a debugger in perturbations.py, and then run my python3 script with DESC will the code stop at the debug breakpoint.

unalmis: Yes. It will certainly stop if you replace your debugger with an `Exception`. If it somehow does not stop at the debugger, my guess is it has something to do with jax and just in time (jit) compilation techniques in the code. Although in my experience, if jax & jit are the cause of some unexpected behavior, they will complain with a warning.

rahulgaur104: Ok, I'll think about it. Thanks!

unalmis:
I'll push a commit that handles the python path issue better

dpanici:
> I'll push a commit that handles the python path issue better

I had sort of put that part onto the user, they should know how to add to their pythonpath by adding to their .bashrc

if they add the export statement to their .bashrc then it should indeed be persistent right?

unalmis:
> > I'll push a commit that handles the python path issue better
> 
> I had sort of put that part onto the user, they should know how to add to their pythonpath by adding to their .bashrc
> 
> if they add the export statement to their .bashrc then it should indeed be persistent right?

Yes it will be persistent if you modify the bashrc file. If you use the export statement as a terminal command it will disappear when your machine restarts.

In any case, I think there is a simpler solution than telling people to modify their bashrc. The command `pip install --no-deps -e .` is equivalent to that `python setup.py install` command you had before; except that it installs a live version of DESC directory. I think this is the typical way to install a development environment (see [here](https://github.com/conda/conda-build/issues/4251#issuecomment-1053460542)).

I'll push this change. If you don't like it just revert it (or tell me to).

unalmis (dismissed):

unalmis (dismissed):

unalmis:
I tested all the installs except for IBM Power Architecture (Traverse). Everything works except the Della GPU one (on both 12715b52bdbd7a0dc2ed19cae98c46b1d588b372 and 324d06e5609e1615d21500df5d8713db69b2156f). The error it gives is attached [della_gpu_error.txt](https://github.com/PlasmaControl/DESC/files/11349453/della_gpu_error.txt). (The stellar GPU one works fine).

unalmis (dismissed):

f0uriest (approved):

unalmis (dismissed):

dpanici:
- [ ] see why della fails

dpanici:
> I tested all the installs except for IBM Power Architecture (Traverse). Everything works except the Della GPU one (fails on both instructions from [12715b5](https://github.com/PlasmaControl/DESC/commit/12715b52bdbd7a0dc2ed19cae98c46b1d588b372) and [324d06e](https://github.com/PlasmaControl/DESC/commit/324d06e5609e1615d21500df5d8713db69b2156f)). The error it gives is attached [della_gpu_error.txt](https://github.com/PlasmaControl/DESC/files/11349453/della_gpu_error.txt). (The stellar GPU one works fine).

Thanks for testing all these btw, that is really helpful.

Do you get this error when trying to run on a GPU, but not on CPU?

I also got that error when running on CPU after fresh installing DESC on della (but if I run on the head node , but if I include the command `module load cudatoolkit/11.7 cudnn/cuda-11.x/8.2.0` in my slurm script before running DESC, I can run without any issues on the GPU

unalmis (approved):
Yes; I only had the issue with gpu on della. The last commit resolves them

docs/installation.rst
```diff
@@ -121,4 +232,18 @@ You can also try running an example input file:
...
+ - **Solution**: This may be caused by DESC not being on your PYTHONPATH, or your environment containing DESC not being activated.
+     - Try adding the DESC directory to your PYTHONPATH manually by adding the line `export PYTHONPATH="$PYTHONPATH:path/to/DESC"` (where `/path/to/DESC` is the path to the DESC folder on your machine) to the end of your ~/.bashrc (or other shell configuration) file.
+       You will also need to run `source ~/.bashrc` after making the change to ensure that your path updates properly for your current terminal session.
+     - Try ensuring you've activated the conda environment that DESC is in ( `conda activate desc-env` ), then retry using DESC.
+ - **Problem**: I've installed DESC, but when I check my installation I get an error `ModuleNotFoundError: No module named 'termcolor'` (or another module which is not `desc```).
+ - **Solution**: you likely are not running python from the environment in which you've installed DESC.
+     - Try ensuring you've activated the conda environment that DESC is in( `conda activate desc-env` ), then retry using DESC.
```
ddudt, line 255: The code snippets `with formatting like this` don't seem to be rendering correctly here. It might be a formatting issue with being a bullet point list.

ddudt (dismissed):

ddudt:
> I tested all the installs except for IBM Power Architecture (Traverse). Everything works except the Della GPU one (fails on both instructions from [12715b5](https://github.com/PlasmaControl/DESC/commit/12715b52bdbd7a0dc2ed19cae98c46b1d588b372) and [324d06e](https://github.com/PlasmaControl/DESC/commit/324d06e5609e1615d21500df5d8713db69b2156f)). The error it gives is attached [della_gpu_error.txt](https://github.com/PlasmaControl/DESC/files/11349453/della_gpu_error.txt). (The stellar GPU one works fine).

We should update the time stamps in the instructions to reflect this recent testing. I'm talking about the lines like this for each of the clusters:
> These instructions were tested and confirmed to work on the Della cluster at Princeton as of 10-13-2022.

unalmis (approved):

f0uriest (approved):

ddudt (approved):

PR #493, opened 2023-04-24
Dp/softmin

dpanici:
Implement softmin as an option for PlasmaVesselDistance objective
- Uses jax.scipy.special.logsumexp for the softmax implementation, and softmin is just `-(softmax(-array))`
- adds a parameter `alpha` that determines how approximate the `softmin` is, with `softmin -> min` as `alpha -> inf`
- softmin will also automatically multiply by an effective alpha to make the values of an array at least > 1 (to try to help avoid inaccuracies that occur when values <1 are used with the softmin which results in inaccurate estimates or even incorrect signs)

softmax is defined as $log(e^x + e^y)$ 
The parameter alpha can be added to this to make it more closely approximate the true max $$max(x,y) \approx lim_{\alpha\rightarrow \infty} log(e^{\alpha x} + e^{\alpha y}) / \alpha$$

Then softmin is defined as `-softmax(-array)`


dpanici:
- [ ] log sum exp

dpanici:
bounded by log(N) tightly, subtract out log(N) where N is number of variables

dpanici:
with the parameter $\alpha=1$, the softmax is always greater than the actual max bc $log(\sum_i^N e^x_i) > log(e^x_{max})$. Then because we find the softmin by taking `-softmax(-arr)`, the softmin is also always going to be less than the actual min. So this should provide a conservative estimate of these values, which is the way we would like to lean for this type of calculation (i.e. if our bound is we want a minimum distance from the vessel, using softmin should get us at least that distance away,  if not more, since it is a conservative estimate of the minimum that is lower than the real minimum)

dpanici:
Bigger Issue: for small numbers, the log will end up being negative, meaning that the outcome of the softmin function isn't even the same sign as the inputs

desc/objectives/utils.py
```diff
@@ -371,6 +372,43 @@ def align_jacobian(Fx, objective_f, objective_g):
...
+    arr: ndarray, the array which we would like to apply the softmin function to.
+    alpha: float, the parameter smoothly transitioning the function to a hardmin.
+        as alpha increases, the value returned will come closer and closer to
+        min(arr).
+
+    Returns
+    -------
+    softmin: float, the soft-minimum of the array.
+    """
+    return -jax_softmax(-arr, alpha)
```
dpanici, line 409: @f0uriest if I were to add a line here that checked that alpha*arr was > 1, if not it multiplied by a factor to make it so, then divided by it, would that end up messing with the optimization? it is multiplication and division by the same quantity so it should be fine, though when it gets jitted I am not sure if that will freeze the if statement logic or not

like

```python
if jnp.any(jnp.abs(alpha*arr)<1):
    factor = jnp.min(jnp.abs(alpha*arr))
else:
    factor = 1
return -jax_softmax(-arr/factor,alpha)*factor
```

to avoid cases where it is inaccurate bc array elements are <1

dpanici: or maybe
```python
from jax import lax

arr_times_alpha = alpha*arr
min_val = jnp.min(jnp.abs(arr_times_alpha))
return lax.cond(jnp.any(jnp.abs(arr_times_alpha)<1),
                lambda arr_times_alpha: logsumexp(arr_times_alpha / min_val ) / alpha * min_val,
                lambda arr_times_alpha: logsumexp(arr_times_alpha) / alpha,
                arr_times_alpha)
 ```
for the softmax function. This does add some overhead (it is ~twice as slow as not having this conditional), and I am not even sure how much this helps with this issue according to the tests I am running right now...

Comment (dpanici) [reply desc/objectives/utils.py:409]: This should work fine with jit, but I am not sure how this affects the derivative since it technically is using the min function...

File: desc/objectives/_geometry.py [hunk 2]
```diff
@@ -415,6 +424,9 @@ class PlasmaVesselDistance(_Objective):
         Collocation grid containing the nodes to evaluate surface geometry at.
     plasma_grid : Grid, ndarray, optional
         Collocation grid containing the nodes to evaluate plasma geometry at.
+    use_softmin: Bool, use softmin or hard min.
+    alpha: float, parameter used for softmin. The larger alpha, the closer the softmin
```
f0uriest, line 428: is this needed if we're manually ensuring that alpha*arr > 1? Should we include a note about that here?

f0uriest (changes requested):
I'd update the docstring to make clear that alpha is only a "suggestion"

desc/objectives/_geometry.py
```diff
@@ -389,6 +390,14 @@ class PlasmaVesselDistance(_Objective):
     plasma grid. For dense grids, this will approximate the global min, but in general
     will only be an upper bound on the minimum separation between the plasma and the
     surrounding surface.
+    NOTE: for best results, use this objective in combination with either MeanCurvature
```
f0uriest, line 393: could use an extra space here to avoid formatting issues

desc/objectives/_geometry.py
dpanici: ensureing alpha*arr>1 isn't sufficient to ensure that the softmin is accurate, I just have it there as a check bc we know for sure if alpha*arr<1, the softmin is inaccurate. Allowing alpha to be larger than that minimum threshold value lets the user ensure that the softmin is accurate/approaching the hardmin. I will include a note about that

f0uriest (approved):

f0uriest (approved):

PR #494, opened 2023-04-27
Dp/XYZ_coils_and_BNORM

### LINKED ISSUE #460: Add support for reading in coils from make_grid type files
codes like REGCOIL output `coil.XXX` text files in the format

```
X_1  Y_1 Z_1 Current
X_2  Y_2 Z_2 Current
...
X_nseg  Y_nseg Z_nseg Current
(repeated for each coil in the coilset)
```
It would be useful to be able to interface with these outputs files
- [x] Add methods to `FourierXYZCoil` to read in and fit the curves from these files
- [x] Add new coil class `XYZCoil` to be able to just store the coil segment locations in space (without needing to Fourier Transform)

dpanici:
The main thing this PR adds is a new curve/coil class (SplineXYZ), I/O support for MAKEGRID-formatted coils, and the ability to read in and save BNORM-style files containing information on B normal on plasma surface.

also has pytest ignore RuntimeWarnings from Cython about numpy array sizes changing, as if they are just a warning then they are benign, and if it would ever cause issues it would probably instead be a RuntimeError (https://stackoverflow.com/questions/40845304/runtimewarning-numpy-dtype-size-changed-may-indicate-binary-incompatibility)

resolves #460

## SplineXYZCurve and SplineXYZCoil
- New classes which accept coordinates in real space cartesian, and spline fits them with an arbitrary angle (which by default is normalized arclength). This is the coil format typically used in MAKEGRID-formatted coils files outputted by other codes like REGCOIL and FOCUS, and is the format accepted by MAKEGRID. 
- type of spline is any of the types that exist in `desc/interpolate.py`

## Coil and MAKEGRID Conventions
Where possible the conventions have been chosen to be similar to what is already used and expected by the [MAKEGRID code](https://princetonuniversity.github.io/STELLOPT/MAKEGRID.html). This was chosen to make DESC able to read in MAKEGRID coilis files from other sources, and to be able to have DESC-saved files be readable by other codes as well.

 - CoilSets have an NFP attribute, but this is only for informational purposes of what the nominal magnetic field periodicity is for this CoilSet. It does not affect any computations
 - Therefore, CoilSet objects should always have the full CoilSet in them, even if there is a discrete rotational symmetry.
 - CoilGroups (that is, the groups of coils which can be assigned the same current in MAKEGRID) must be handled by the user by specifying the name of each coil in the CoilSet when saving to the MAKEGRID formatted file. This could be changed in the future to also have a CoilSet name be used to provide the coil group, but as of right now this is the current support for that compatibility

# TODO List
- [x] Change XYZcurve and coil to take in points, but then use them as spline knots and values, then have it calculate the points on a grid in theta 
- [x] decide what theta should be (arclength by default)
- [x] Add more methods for XYZcurve for computing curvature etc
- [x] Add ability to save XYZcoils as txt file in MAKEGRID format (for possible compatibility/use in other codes like FOCUS)
- [x] fix loading in MAKEGRID coil file to allow for coil group names other than Modular
- [ ] allow saving in MAKEGRID coil file to specify the group name (probably using the name of the coilset, and if a coilset is made of multiple coiilsets then loop thru them and use each name of the coilset as the coil group) - punting on this until I understand coil groups better in MAKEGRID, if someone else does and wants to contribute feel free to
- [x] add test that MAKEGRID loading works with periods > 1
- [x] change MAKEGRID NFP to be more sensible - really just put it in line with the community standard as I know it, which is the coils. file always has EVERY coil, and the period argument is just to signal that the magnetic field should be unique only across 2pi/NFP, but it is not used to save space by saving only the coils over a single field period.
- [x] Add ability to save in BNORM output for compatibility with other codes
- [x] check sign of current for saving to MAKEGRID to ensure that when read it keeps the sign of B the same
- [x] Add tests for XYZCurve with different interpolation methods other than cubic2
- [x] add XYZCurve and XYZCoil to docs API
- [x] make a pretty description of this PR to compete with @unalmis 
- [x] refactor compute methods to work with changes made in #572 , model after #583
  - [x] replace compute methods of the XYZCurve and XYZCoil to use the generic compute method
  - [x] add compute method for XYZCurve unique quantities etc into the compute/_curve.py (if needed)
  - [x] remove grid attribute from XYZCurve
- [x] change name to SplineXYZCurve


dpanici:
Although the methods for computing the curvature torsion etc for XYZCurve should be the same as the FourierXYZCurve, I am finding that if I say fit a FourierXYZCurve with an XYZCurve then compare the curvature, torsion etc, it is not correct compared to the original FourierXYZCurve. I suspect this is just due to spline fitting only preserving the values accurately, not the values of the derivatives. So I did not add a test yet for these compute methods.

desc/coils.py
```diff
@@ -311,7 +366,7 @@ def compute_magnetic_field(self, coords, params={}, basis="rpz"):
         field : ndarray, shape(n,3)
             magnetic field at specified points, in either rpz or xyz coordinates
         """
-        if isinstance(params, dict):
+        if isinstance(params, dict) or params is None:
```
dpanici, line 369: this is in the commit message, but to explain this, I have been using SumMagneticField  to sum both a Coil and a MagneticField, and this was needed because the params is an empty dict by default for a coil, but it is None for a MagneticField (iirc)

dpanici:
Waiting for #572 et al then will refactor this to work with the change

desc/coils.py
f0uriest: This was a mistake on my part, I should have made it default to None for all of them.

desc/coils.py
```diff
@@ -220,7 +273,9 @@ class CoilSet(Coil, MutableSequence):
 
     _io_attrs_ = Coil._io_attrs_ + ["_coils"]
 
-    def __init__(self, *coils, name=""):
+    def __init__(
+        self, *coils, name=""
+    ):  # FIXME: if a list of of Coils is passed, this fails...
```
f0uriest, line 278: Could use `flatten_list` here, on one of my branches I added a flag to also have it flatten tuples.

desc/coils.py
```diff
@@ -430,6 +485,185 @@ def from_symmetry(cls, coils, NFP, sym=False):
...
+
+        By default, each coil is assigned to the same Coilgroup in MAKEGRID
+        with the name "Modular". For more details see the MAKEGRID documentation
+        https://princetonuniversity.github.io/STELLOPT/MAKEGRID.html
+
+        Parameters
+        ----------
+        filename : str or path-like
+            path save CoilSet as a file in MAKEGRID txt format
+        NFP : int, default 1
```
f0uriest, line 582: this seems a bit confusing to the way we normally use NFP. Are we always assuming that the coilset goes around the whole plasma? or does the coilset only cover 1 field period?

desc/geometry/curve.py
```diff
@@ -807,6 +808,76 @@ def compute_length(self, X_n=None, Y_n=None, Z_n=None, grid=None):
         theta = transform.grid.nodes[:, 2]
         return jnp.trapz(T, theta)
 
+    @classmethod
+    def from_XYZCurve(cls, xyz_curve, N=10, grid=None):
```
f0uriest, line 812: probably don't need both from and to methods. For profiles we have `to_*` methods in the base class, and then each subclass just has a `from_values` method to fit from raw points, we could probably due something similar here

desc/geometry/curve.py
```diff
@@ -1218,3 +1289,440 @@ def compute_length(self, center=None, normal=None, r_n=None, grid=None):
...
+
+    Parameters
+    ----------
+    X, Y, Z: array-like
+        points for X, Y, Z describing a closed curve
+    knots : ndarray
+        arbitrary theta values to use for spline knots,
+        should be an 1D ndarray of same length as the input.
+        If None, defaults to using an equal-arclength angle as the knot
+    period: float
```
f0uriest, line 1305: I think it would be better to always have the period be 2pi, for consistency with the other curve types. If the user supplies their own knots we could just rescale them to be in [0,2pi]

desc/geometry/curve.py
```diff
@@ -1218,3 +1289,440 @@ def compute_length(self, center=None, normal=None, r_n=None, grid=None):
         T = jnp.linalg.norm(T, axis=1)
         theta = transform.grid.nodes[:, 2]
         return jnp.trapz(T, theta)
+
+
+class XYZCurve(Curve):
```
f0uriest, line 1294: I'd prefer if this was `SplineXYZCurve` for consistency with the other names

desc/geometry/curve.py
```diff
@@ -1218,3 +1289,440 @@ def compute_length(self, center=None, normal=None, r_n=None, grid=None):
...
+    @X.setter
+    def X(self, new):
+        if len(new) == len(self._knots):
+            self._X = jnp.asarray(new)
+        else:
+            raise ValueError(
+                "X should have the same size as the knots, "
+                + f"got {len(new)} X values for {len(self._knots)} knots"
+            )
+        assert np.allclose(
```
f0uriest, line 1391: probably want to move this assert to before we assign the new value

desc/magnetic_fields.py
```diff
@@ -59,6 +63,64 @@ def biot_savart(eval_pts, coil_pts, current):
...
+    m, n, Bnorm_mn = ptolemy_identity_fwd(xm, xn, Bnorm_mn, np.zeros_like(Bnorm_mn))
+    basis = DoubleFourierSeries(int(np.max(m)), int(np.max(n)), sym=False, NFP=eq.NFP)
+    Bnorm_mn_desc_basis = np.zeros((basis.num_modes,))
+    for i, (mm, nn) in enumerate(zip(m, n)):
+        idx = basis.get_idx(L=0, M=mm, N=nn)
+        Bnorm_mn_desc_basis[idx] = Bnorm_mn[0, i]
+
+    if eval_grid is None:
+        eval_grid = LinearGrid(
+            rho=jnp.array(1.0), M=2 * basis.M, N=2 * basis.N, NFP=eq.NFP
```
f0uriest, line 114: would it be better to use `eq.M_grid` and `eq.N_grid` here?

desc/magnetic_fields.py
```diff
@@ -59,6 +63,64 @@ def biot_savart(eval_pts, coil_pts, current):
     return B
 
 
+def read_BNORM_file(fname, eq, eval_grid=None, scale_by_curpol=True):
+    """Create BNORM-style .txt file containing Bnormal Fourier coefficients.
```
f0uriest, line 67: Think this description is wrong

desc/magnetic_fields.py
```diff
@@ -113,6 +175,141 @@ def __call__(self, coords, params={}, basis="rpz"):
         """Compute magnetic field at a set of points."""
         return self.compute_magnetic_field(coords, params, basis)
 
+    def compute_Bnormal(self, surface, grid_M=40, grid_N=40, NFP=None, grid=None):
+        """Create BNORM-style .txt file containing Bnormal Fourier coefficients.
```
f0uriest, line 179: docstring doesn't match signature and implementation

desc/magnetic_fields.py
```diff
@@ -113,6 +175,141 @@ def __call__(self, coords, params={}, basis="rpz"):
...
+            which is expected by most other codes that accept BNORM files,
+            by default True
+
+        Returns
+        -------
+        None
+        """
+        if sym != "sin":
+            raise UserWarning(
+                "BNORM code assumes that |B| has sin symmetry,"
```
f0uriest, line 264: does this mean it wouldn't work for stellarator asymmetric cases? Should we even include the sym option?

desc/magnetic_fields.py
```diff
@@ -113,6 +175,141 @@ def __call__(self, coords, params={}, basis="rpz"):
...
+        basis_N=24,
+        grid=None,
+        sym="sin",
+        scale_by_curpol=True,
+    ):
+        """Create BNORM-style .txt file containing Bnormal Fourier coefficients.
+
+        Parameters
+        ----------
+        eq : Equilibrium
```
f0uriest, line 233: we could probably generalize this to be either an `Equilibrium` or a `Surface`

desc/magnetic_fields.py
```diff
@@ -225,8 +422,8 @@ class ToroidalMagneticField(MagneticField):
     _io_attrs_ = MagneticField._io_attrs_ + ["_B0", "_R0"]
 
     def __init__(self, B0, R0):
-        assert np.isscalar(B0), "B0 must be a scalar"
-        assert np.isscalar(R0), "R0 must be a scalar"
+        assert float(B0) == B0, "B0 must be a scalar"
```
f0uriest, line 425: if we're making this change we should also explicitly cast the values to floats before assigning so that they're actually scalars not size 1 arrays

desc/magnetic_fields.py
dpanici: I am unsure, the BNORM code assumes this so any output that is not stell sym I am not even sure how to represent in their format.

desc/geometry/curve.py
dpanici: because we have the base classes in a separate file, I dont think we can do a similar thing without running into circular imports. I at least could not figure it out

desc/geometry/curve.py
dpanici: I will change this before the PR is finalized

desc/coils.py
dpanici: really useful, thanks

desc/coils.py
dpanici: by assumption yea this is assuming it goes around the whole plasma, if NFP=1 (think of it as NFP of the coilset).

 If NFP is larger than 1, then only the first len(coils)/NFP are unique, and so it saves only those ones (though I just realized I need to change the from_makegrid method to work correctly with that

f0uriest: OK for now I guess. Still feels a bit confusing. I'm wondering if maybe a coilset by definition should only have coils over 1 period, but have an NFP attribute saying how many times it repeats, and when you call `compute_magnetic_field` it automatically sums over the different periods. That also eliminates the need to have explicit constraints that coil 1 is the same thing as coil 1+NFP, just rotated.

f0uriest: I guess that would only work for modular coils though...

dpanici: It is confusing I agree and I dont like how I've defined it. I like that better, I think I'll probably change to that

desc/geometry/curve.py
f0uriest: we could just do a local import of the necessary class in the `to_*` method in the base class?

desc/coils.py
dpanici: your way could work for non modular coils if we just make sure the NFP is one for them...

desc/magnetic_fields.py
```diff
@@ -59,6 +64,81 @@ def biot_savart(eval_pts, coil_pts, current):
...
+        Bnormal distribution from the BNORM Fourier coefficients,
+        evaluated on the given eval_grid
+    """
+    if isinstance(surface, EquilibriaFamily):
+        surface = surface[-1]
+    if isinstance(surface, Equilibrium):
+        eq = surface
+        surface = eq.surface
+    else:
+        eq = None
```
f0uriest, line 101: Do we want to add checking for Stellarator symmetry?

desc/magnetic_fields.py
```diff
@@ -59,6 +64,81 @@ def biot_savart(eval_pts, coil_pts, current):
...
+    data = np.genfromtxt(fname)
+
+    xm = data[:, 0]
+    xn = -data[:, 1]  # negate since BNORM uses sin(mu+nv) convention
+    Bnorm_mn = data[:, 2] / curpol  # these will only be sin terms
+
+    # convert to DESC Fourier representation i.e. like cos(mt)*cos(nz)
+    m, n, Bnorm_mn = ptolemy_identity_fwd(xm, xn, Bnorm_mn, np.zeros_like(Bnorm_mn))
+    basis = DoubleFourierSeries(
+        int(np.max(m)), int(np.max(n)), sym=False, NFP=surface.NFP
```
f0uriest, line 123: Shouldn't this be symmetric?

desc/magnetic_fields.py
```diff
@@ -59,6 +64,81 @@ def biot_savart(eval_pts, coil_pts, current):
...
+    )
+    Bnorm_mn_desc_basis = np.zeros((basis.num_modes,))
+    for i, (mm, nn) in enumerate(zip(m, n)):
```
f0uriest, line 126: Could use copy_coeffs here?

desc/magnetic_fields.py
```diff
@@ -90,7 +170,7 @@ def __sub__(self, x):
         return self.__add__(-x)
 
     @abstractmethod
-    def compute_magnetic_field(self, coords, params={}, basis="rpz"):
+    def compute_magnetic_field(self, coords, params={}, basis="rpz", grid=None):
```
f0uriest, line 173: Should probably make default params none

desc/magnetic_fields.py
```diff
@@ -113,6 +197,154 @@ def __call__(self, coords, params={}, basis="rpz"):
...
 
+    def compute_Bnormal(self, surface, grid=None, NFP=None):
+        """Compute Bnormal from self on the given surface.
+
+        Parameters
+        ----------
+        surface : Surface or Equilibrium
+            Surface to calculate the magnetic field's Bnormal on.
+            If an Equilibrium is supplied, will use its boundary surface.
+        grid : Grid, optional
```
f0uriest, line 208: Should also have grid for discretizing the source

desc/magnetic_fields.py
```diff
@@ -113,6 +197,154 @@ def __call__(self, coords, params={}, basis="rpz"):
...
+        if NFP is None:
+            NFP = surface.NFP
+        if grid is None:
+            grid = LinearGrid(
+                rho=jnp.array(1.0), M=2 * surface.M, N=2 * surface.N, NFP=NFP
+            )
+        coords = surface.compute("x", grid=grid, basis="xyz")["x"]
+        rs_t = surface.compute("e_theta", grid=grid, basis="xyz")["e_theta"]
+        rs_z = surface.compute("e_zeta", grid=grid, basis="xyz")["e_zeta"]
+        surf_normal = cross(rs_t, rs_z)
```
f0uriest, line 235: This isn't generally a unit vector, is that ok? Could use n_rho

desc/magnetic_fields.py
```diff
@@ -113,6 +197,154 @@ def __call__(self, coords, params={}, basis="rpz"):
...
+        if isinstance(surface, Equilibrium):
+            surface = surface.surface
+        if NFP is None:
+            NFP = surface.NFP
+        if grid is None:
+            grid = LinearGrid(
+                rho=jnp.array(1.0), M=2 * surface.M, N=2 * surface.N, NFP=NFP
+            )
+        coords = surface.compute("x", grid=grid, basis="xyz")["x"]
+        rs_t = surface.compute("e_theta", grid=grid, basis="xyz")["e_theta"]
```
f0uriest, line 233: It'll be slightly faster if you compute all of these in one call

desc/magnetic_fields.py
```diff
@@ -113,6 +197,154 @@ def __call__(self, coords, params={}, basis="rpz"):
...
+        """Compute Bnormal from self on the given surface.
+
+        Parameters
+        ----------
+        surface : Surface or Equilibrium
+            Surface to calculate the magnetic field's Bnormal on.
+            If an Equilibrium is supplied, will use its boundary surface.
+        grid : Grid, optional
+            Grid of points on the  surface to calculate the Bnormal at and then
+            fit with a DoubleFourierSeries, if None defaults to a LinearGrid with twice
```
f0uriest, line 210: The code doesn't look like it does the fitting, just evaluating

desc/magnetic_fields.py
```diff
@@ -113,6 +197,154 @@ def __call__(self, coords, params={}, basis="rpz"):
...
+                rho=jnp.array(1.0), M=2 * surface.M, N=2 * surface.N, NFP=NFP
+            )
+        coords = surface.compute("x", grid=grid, basis="xyz")["x"]
+        rs_t = surface.compute("e_theta", grid=grid, basis="xyz")["e_theta"]
+        rs_z = surface.compute("e_zeta", grid=grid, basis="xyz")["e_zeta"]
+        surf_normal = cross(rs_t, rs_z)
+        B = self.compute_magnetic_field(coords, basis="xyz")
+
+        Bnorm = jnp.sum(B * surf_normal, axis=-1)
+        return Bnorm
```
f0uriest, line 239: Should we also return x on the surface here?

desc/magnetic_fields.py
```diff
@@ -113,6 +197,154 @@ def __call__(self, coords, params={}, basis="rpz"):
...
+            If an Equilibrium is supplied, will use its boundary surface.
+        fname : str
+            name of file to save the BNORM Bnormal Fourier coefficients to.
+        basis_M : int, optional
+            Poloidal resolution of the DoubleFourierSeries used to fit the Bnormal
+            on the plasma surface, by default 24
+        basis_N : int, optional
+            Toroidal resolution of the DoubleFourierSeries used to fit the Bnormal
+            on the plasma surface, by default 24
+        grid : Grid, optional
```
f0uriest, line 266: Should also have a source grid

f0uriest: And ability to pass in params for field?

desc/magnetic_fields.py
```diff
@@ -198,13 +430,23 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz"):
...
         for i, field in enumerate(self._fields):
-            B += field.compute_magnetic_field(coords, params[i % len(params)], basis)
+            B += (
+                field.compute_magnetic_field(
+                    coords, params[i % len(params)], basis, grid=grid
+                )
+                if isinstance(field, Coil)
+                else field.compute_magnetic_field(
+                    coords, params[i % len(params)], basis
+                )
```
f0uriest, line 448: Since the base class has grid as part of the API we should update the others as well, which will avoid the need for this logic

desc/compute/_curve.py
```diff
@@ -501,6 +502,241 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+    data=["s"],
+    parameterization="desc.geometry.curve.SplineXYZCurve",
+    basis="basis",
+)
+def _x_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
+    xq = data["s"]
+
+    Xq = interp1d(
+        xq,
+        params["_knots"],
```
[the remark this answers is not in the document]
dpanici: these are the attributes of the XYZCurve, they are underscored I assumed because it is how it is accessed. Can I make them not underscored and it will just use the getter?

desc/geometry/core.py
```diff
@@ -69,14 +69,23 @@ def compute(
...
             NFP = self.NFP if hasattr(self, "NFP") else 1
-            grid = LinearGrid(N=2 * self.N + 5, NFP=NFP, endpoint=True)
-        if isinstance(grid, numbers.Integral):
+            grid = LinearGrid(N=2 * N + 5, NFP=NFP, endpoint=True)
+        elif isinstance(grid, numbers.Integral):
             NFP = self.NFP if hasattr(self, "NFP") else 1
             grid = LinearGrid(N=grid, NFP=NFP, endpoint=True)
+        elif isinstance(grid, Grid):
+            NFP = grid.NFP
+        elif not isinstance(grid, Grid):
```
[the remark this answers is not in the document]
dpanici: ah you are right

desc/compute/_curve.py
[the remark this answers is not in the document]
f0uriest: yeah, if there is a property with the non-underscored name we should use that.

desc/compute/_curve.py
```diff
@@ -501,6 +502,310 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+        params["X"],
+        method=params["_method"],
+        derivative=0,
+        period=params["_period"],
```
f0uriest, line 532: for curves isn't the period always 2pi?

desc/compute/_curve.py
```diff
@@ -501,6 +502,310 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+    )
+    Zq = interp1d(
+        xq,
+        params["_knots"],
+        params["Z"],
+        method=params["_method"],
+        derivative=0,
+        period=params["_period"],
+    )
```
f0uriest, line 628: could move this stuff into the if block so we only do the extra work when needed, or if we want to always do it its probably easier to just list x and y as dependencies?

desc/geometry/curve.py
```diff
@@ -533,3 +630,243 @@ def set_coeffs(self, n, r=None):
...
+
+    _io_attrs_ = Curve._io_attrs_ + ["_X", "_Y", "_Z", "_basis", "_transform"]
+
+    def __init__(
+        self,
+        X,
+        Y,
+        Z,
+        knots=None,
+        method="cubic",
```
f0uriest, line 726: for the SplineXYZCoil and all the `to_SplineXYZCurve` methods the default method is `cubic2`, any reason its different here?

desc/coils.py
```diff
@@ -560,6 +632,192 @@ def from_symmetry(cls, coils, NFP, sym=False):
 
         return cls(*coilset)
 
+    @classmethod
+    def from_makegrid_coilfile(cls, coil_file, method="cubic"):
```
f0uriest, line 636: similar comment, we should have a consistent default interpolation method

desc/coils.py
```diff
@@ -560,6 +632,192 @@ def from_symmetry(cls, coils, NFP, sym=False):
...
+            raise OSError(
+                "first 3 lines in file must be the header lines,"
+                + " each of length 2 (after .split())!"
+                + f"Line(s) {lines[np.where(header_lines_not_as_expected)[0]]}"
+                + " is not length 2"
+            )
+
+        for i, (start, end) in enumerate(zip(coilinds[0:-1], coilinds[1:])):
+            coords = np.genfromtxt(lines[start + 1 : end])
+            if i % 20 == 0:
```
f0uriest, line 697: is this assuming the coil only has 20 points?

desc/compute/_curve.py
dpanici: yea now it is, i should change this

desc/geometry/curve.py
dpanici: I had ran into some issues when loading from a MAKEGRID file where cubic2 interpolation failed at the endpoints and gave NaN, but cubic did not have that same issue

desc/coils.py
dpanici: no, it is a holdout from debugging (making sure I knew the load function was working by printing progress), I will remove

desc/compute/_curve.py
dpanici: listing x,y as dependencies (by listing "x" as a dependency) would stilll make it possible to incorrectly use an rpz "x" if that was already present in the data dictionary when this is called. I will just move into if block

desc/coils.py
```diff
@@ -560,6 +636,192 @@ def from_symmetry(cls, coils, NFP, sym=False):
...
+        np.savetxt(
+            coilsFilename,
+            save_arr,
+            delimiter=" ",
+            header=header,
+            footer=footer,
+            fmt="%14.8e",
+            comments="",  # to avoid the # appended to the start of the header/footer
+        )
+        # now need to re-load the file and place coilgroup markers at end of each coil
```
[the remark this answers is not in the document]
dpanici: I don't know a smart way to do so while also using np.savetxt to save the data

desc/coils.py
```diff
@@ -560,6 +636,192 @@ def from_symmetry(cls, coils, NFP, sym=False):
...
+        for i in range(len(coil_end_inds)):
+            name = coils[i].name if coils[i].name != "" else "1 Modular"
+            real_end_ind = int(
+                np.sum(coil_end_inds[0 : i + 1]) + 2
+            )  # to account for the 3 header lines
+            lines[real_end_ind] = lines[real_end_ind].strip("\n") + f" {name}\n"
+        with open(coilsFilename, "w") as f:
+            f.writelines(lines)
+
+        print(f"Saved coils file at : {coilsFilename}")
```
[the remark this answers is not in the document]
dpanici: we don't, I can remove

desc/magnetic_fields.py
dpanici: yea I think I had it as false before I fixed another bug we had in the VMEC mode stuff, changed it to sym = 'sin' now

desc/coils.py
```diff
@@ -560,6 +636,192 @@ def from_symmetry(cls, coils, NFP, sym=False):
...
+        for i in range(int(len(coils))):
+            coil = coils[i]
+            coords = coil.compute("x", basis="xyz", grid=grid)["x"]
+
+            contour_X = np.asarray(coords[0:-1, 0])
+            contour_Y = np.asarray(coords[0:-1, 1])
+            contour_Z = np.asarray(coords[0:-1, 2])
+
+            currents = np.ones_like(contour_X) * float(coil.current)
+            # close the curves
```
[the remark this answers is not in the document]
dpanici: thanks for catching that. Technically I think it was ok because when I index coords I did `[:-1]` so the last point was not included, but it only works assuming the grid had endpoint=True, I fixed it now

desc/geometry/core.py
```diff
@@ -69,14 +69,23 @@ def compute(
             Computed quantity and intermediate variables.
 
         """
+        # set a default numpts for the SplineXYZCurve
+        N = self.N if hasattr(self, "N") else self.X.size
```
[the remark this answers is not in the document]
dpanici: I tried messing around with the logic to make that change but some of the tests were not passing, I will leave it as is for now

desc/geometry/core.py
```diff
@@ -158,6 +158,77 @@ def __repr__(self):
...
+
+        """
+        from .curve import SplineXYZCurve
+
+        coords = self.compute("x", grid=grid, basis="xyz")["x"]
+        return SplineXYZCurve.from_values(
+            coords, knots=knots, method=method, name=name, basis="xyz"
+        )
+
+    # TODO: to_rz method for converting to FourierRZCurve representation
```
f0uriest, line 229: We could probably add this, along with a check that phi/zeta is monotonic as a function of the original curve parameter, and throw a warning if its not?

desc/geometry/curve.py
```diff
@@ -428,75 +369,58 @@ def Z_n(self, new):
...
             coords_xyz = coords
         else:
-            coords_rpz = coords
             coords_xyz = rpz2xyz(coords, phi=coords[:, 1])
-        grid = LinearGrid(zeta=coords_rpz[:, 1], NFP=1, sym=False)
+        X = coords_xyz[:, 0]
+        Y = coords_xyz[:, 1]
+        Z = coords_xyz[:, 2]
+
+        assert np.allclose(X[-1], X[0], atol=1e-14), "Must pass in a closed curve!"
```
f0uriest, line 401: does it need to be closed explicitly? Wouldn't the fit take care of that?

desc/geometry/curve.py
```diff
@@ -753,8 +617,8 @@ def __init__(
             knots = knots - knots[0]
             knots = (knots / knots[-1]) * 2 * np.pi
 
-        self._knots = knots
-        self._method = method
+        self.knots = knots
```
f0uriest, line 620: do we want getters/setters for these?

desc/magnetic_fields.py
f0uriest: This same description is also used for `compute_Bnormal` and `save_BNORM_file`

desc/geometry/core.py
f0uriest: can be a new PR though

desc/geometry/curve.py
dpanici: I'd rather keep this requirement, reason being if an angle is passed in, it is easier to rescale to [0,2pi] if I know that the last point is the endpoint, otherwise I have no way of knowing of knowing where to put the last point before doing the rescale

(use case: you pass in an arbitrary angle from 0->4 but it is not equally spaced, then I have no way of extrapolating to the endpoint)

desc/geometry/core.py
dpanici: Might save as another PR if I don't add in the next hour or so. 

I just realized, should I also be enforcing monotonicity of the curve s parameter? to ensure that the curve is always being traversed in one direction?

f0uriest: it should be monotonic when the points are ordered "correctly", so maybe we just sort `s` and argsort `x` based on that?

f0uriest (approved):

desc/geometry/core.py
dpanici: I think I am fine leaving it as is for now, and put the onus on the user to make sure it is monotonic, which most sensible use cases where user is supplying s, it should be

tests/test_coils.py
```diff
@@ -254,6 +325,142 @@ def test_dunder_methods(self):
...
+    """Test loading in and saving CoilSets from MAKEGRID format files."""
+    Ncoils = 22
+    input_path = f"./tests/inputs/coils.MAKEGRID_format_{Ncoils}_coils"
+    tmpdir = tmpdir_factory.mktemp("coil_files")
+    tmp_path = tmpdir.join(f"coils.MAKEGRID_format_{Ncoils}_coils")
+    shutil.copyfile(input_path, tmp_path)
+
+    coilset = CoilSet.from_makegrid_coilfile(str(tmp_path))
+    assert len(coilset) == Ncoils  # correct number of coils
+    # TODO: add better tests for this?
```
ddudt, line 339: Did you want to add better tests in this PR?

desc/magnetic_fields.py
```diff
@@ -59,6 +64,85 @@ def biot_savart(eval_pts, coil_pts, current):
...
+        Grid of points on the plasma surface to evaluate the Bnormal at,
+        if None defaults to a LinearGrid with twice
+        the surface grid's poloidal and toroidal resolutions
+    scale_by_curpol : bool, optional
+        Whether or not to un-scale the Bnormal coefficients by curpol
+        before calculating Bnormal, by default True
+        (set to False if it is known that the BNORM file was saved without scaling
+        by curpol)
+        requires an Equilibrium to be passed in
+
```
ddudt, line 87: extra line here

desc/magnetic_fields.py
```diff
@@ -109,10 +197,187 @@ def compute_magnetic_field(self, coords, params={}, basis="rpz"):
...
+        # "BNORM scales B_n by curpol=(2*pi/nfp)*bsubv(m=0,n=0)
+        # where bsubv is the extrapolation to the last full mesh point of
+        # bsubvmnc."
+        # this corresponds to 2pi/NFP*G(rho=1) in DESC
+        curpol = (
+            (
+                2
+                * jnp.pi
+                / surface.NFP
+                * eq.compute("G", grid=LinearGrid(rho=jnp.array(1)))["G"]
```
ddudt, line 369: Don't we need to set `M` and `N` in order to get a good flux surface average? Otherwise it will default to returning $B_\theta$ at $\theta=\zeta=0$.

desc/coils.py
```diff
@@ -312,11 +361,15 @@ class CoilSet(Coil, MutableSequence):
         collection of coils
     currents : float or array-like of float
         currents in each coil, or a single current shared by all coils in the set
+    name : str
+        name of this CoilSet
+
     """
 
-    _io_attrs_ = Coil._io_attrs_ + ["_coils"]
+    _io_attrs_ = Coil._io_attrs_ + ["_coils"] + ["_NFP"]
```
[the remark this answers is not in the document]
f0uriest: Is nfp still an attribute? I thought that was removed

desc/magnetic_fields.py
f0uriest: the logic in eq.compute should use the correct M and N for surface averaged quantities

desc/coils.py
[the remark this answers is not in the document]
dpanici: Ah I do think it was removed

setup.cfg
```diff
@@ -49,7 +49,10 @@ filterwarnings=
     ignore::pytest.PytestUnraisableExceptionWarning
     ignore::RuntimeWarning:desc.compute
     # Ignore division by zero warnings.
+    ignore:numpy.ndarray size changed:RuntimeWarning
```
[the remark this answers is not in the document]
unalmis: the comments need to be on their own line fyi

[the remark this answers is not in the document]
dpanici: ^ yea it thinks that you are naming a different module otherwise. And I think that this is fine, only we look at it and we can figure out which go with which

desc/coils.py
```diff
@@ -560,6 +613,207 @@ def from_symmetry(cls, coils, NFP, sym=False):
...
+            [
+                len(lines[0].split()) != 2,
+                len(lines[1].split()) != 2,
+                len(lines[2].split()) != 2,
+            ]
+        )
+        if np.any(header_lines_not_as_expected):
+            raise OSError(
+                "first 3 lines in file must be the header lines,"
+                + " each of length 2 (after .split())!"
```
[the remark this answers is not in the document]
dpanici: purposely not putting the period at the end here to be annoying

tests/test_coils.py
dpanici: yep, done

desc/grid.py
```diff
@@ -733,10 +733,12 @@ def _create_nodes(  # noqa: C901
         self._endpoint = (
             t.size > 0
             and z.size > 0
-            and (t[0] == 0 and t[-1] == THETA_ENDPOINT)
-            and (z[0] == 0 and z[-1] == ZETA_ENDPOINT)
-        )
-
+            and (
```
dpanici, line 736: @unalmis I think this change is fine, what I am fixing here is if we do `LinearGrid(zeta=np.linspace(0,2*np.pi,100,endpoint=True), endpoint=True)` , this line would revert the endpoint bool to `False` because the theta grid is only `np.array([0.]`. I think that we should be able to have endpoint true for LinearGrids with a single point in one angle and then the full array in the other, and I use that in other part of the code to check if to complete a zeta curve I need to add another point, or if the grid takes care of it

f0uriest (dismissed):

desc/compute/data_index.py
```diff
@@ -133,6 +136,10 @@ def _decorator(func):
         "desc.geometry.curve.FourierPlanarCurve",
         "desc.geometry.core.Curve",
     ],
+    "desc.coils.SplineXYZCoil": [
+        "desc.geometry.curve.SplineXYZCurve",
+        "desc.geometry.core.Curve",
+    ],
```
unalmis, line 142: Could you add these new parameterizations to `tests/test_compute_funs/test_compute_everything` and an assert statement to that test like
```python
assert things.keys() == data_index.keys(), "Missing a parameterization to test against master."
```

desc/geometry/curve.py
```diff
@@ -533,3 +559,217 @@ def set_coeffs(self, n, r=None):
...
+        self._Y = Y
+        self._Z = Z
+
+        if knots is None:
+            # find equal arclength angle-like variable, and use that as theta
+            # L_along_curve / L = theta / 2pi
+            lengths = jnp.sqrt(
+                (X[0:-1] - X[1:]) ** 2 + (Y[0:-1] - Y[1:]) ** 2 + (Z[0:-1] - Z[1:]) ** 2
+            )
+            thetas = 2 * jnp.pi * np.cumsum(lengths) / jnp.sum(lengths)
```
unalmis, line 623: You're mixing calls to `jnp` and `np` here. just pointing out if your intention was to use only numpy. also don't need to change but i like `np.diff(X)` more than `X[0:-1] - X[1:]`.

desc/grid.py
unalmis: Yes this change is fine.

Since it is somewhat related, I want to remind just to make sure you're aware: we prioritize getting the surface area spacing correct over the line lengths when there are duplicate nodes. So while the correct "length" of  the zeta curve on this grid (`spacing[:, 2].sum()`) would be: $2 \pi$, it will be computed as $6.3$ instead. This number will converge to $2\pi$ as the number of points on the grid grows. 

I'm reading through the PR to see if this might mess anything up.

dpanici: Hm, I think that is fine? when we do curve computations we dont call grid.spacing

unalmis: I agree I didn't see any issues with this.

tests/test_compute_funs.py
```diff
@@ -1189,6 +1189,20 @@ def test_compare_quantities_to_vmec():
     np.testing.assert_allclose(J_dot_B_desc, J_dot_B_vmec, rtol=0.005)
 
 
+do_not_test_these_values = {
+    "desc.geometry.curve.SplineXYZCurve": [
+        "x_s",
```
f0uriest, line 1194: why not test these? It's a regression test so as long as we're comparing apples to apples it should be ok? We can have a separate test that compares things to spectral basis curves

dpanici: Thought something different was happening with the test, just regened the master data for spline XYZ curve and it works now so they are tested now

f0uriest (approved):

PR #498, opened 2023-05-01
Fix theta orientation in vmec output

f0uriest:
What the title says. Previously we correctly flipped R,Z,lambda, but forgot to also flip the derived quantities like B, J, etc.

### DIFF
--- a/desc/vmec.py
+++ b/desc/vmec.py
@@ -718,10 +718,23 @@ class VMECIO:
             cos_transform = Transform(
                 grid=grid, basis=cos_basis, build=False, build_pinv=True
             )
+
+            def cosfit(x):
+                y = cos_transform.fit(x)
+                return np.where(cos_transform.basis.modes[:, 1] < 0, -y, y)
+
+            def sinfit(x):
+                y = sin_transform.fit(x)
+                return np.where(sin_transform.basis.modes[:, 1] < 0, -y, y)
+
         else:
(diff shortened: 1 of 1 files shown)


desc/vmec.py
```diff
@@ -775,9 +788,9 @@ def save(cls, eq, path, surfs=128, verbose=1):  # noqa: C901 - FIXME - simplify
...
         for i in range(surfs - 1):
             if eq.sym:
-                x_mn[i, :] = cos_transform.fit(data[i, :])
+                x_mn[i, :] = cosfit(data[i, :])
             else:
-                x_mn[i, :] = full_transform.fit(data[i, :])
+                x_mn[i, :] = fullfit(data[i, :])
         xm, xn, s, c = ptolemy_identity_rev(m, n, x_mn)
         gmnc[0, :] = 0
         gmnc[1:, :] = -c  # negative sign for negative Jacobian
```
ddudt, line 796: Do we still think these hard-coded negative signs are needed after this change? I think so, but want to double check.

ddudt (approved):

desc/vmec.py
f0uriest: they do seem to be. Flipping the m modes flips the effective definition of theta, but all the quantities are still off by a minus sign, likely because of the jacobian being a factor going into B, J etc.

PR #499, opened 2023-05-04
Vector integral

unalmis:
Generalize surface integral routines to work with vector valued functions.


desc/grid.py
```diff
@@ -668,7 +668,7 @@ def change_resolution(self, L, M, N, NFP=None):
                 axis=len(self.axis) > 0,
                 endpoint=self.endpoint,
             )
-            self._enforce_symmetry()
```
unalmis, line 671: forgot to comment in earlier pr. just avoids unneeded computation.

tests/test_compute_utils.py
unalmis, line 1: Adds a test for the new functionality. The other changes are from https://github.com/PlasmaControl/DESC/pull/448/files#r1163580738 since it helped me test faster.

desc/compute/utils.py
```diff
@@ -691,15 +690,23 @@ def surface_integrals(grid, q=jnp.array([1]), surface_label="rho", max_surface=F
...
+        integrals = jnp.histogram(nodes_modulo, bins=bins, weights=integrands)[0]
         # By modulating nodes, we 'moved' all the area from
         # surface_label=max_surface_val to the surface_label=0 surface.
         # With the correct value of the integral on this surface stored
         # in integrals[0], we copy it to the duplicate surface at integrals[-1].
-        assert integrals[-1] == 0
-        integrals = put(integrals, -1, integrals[0])
-    return expand(grid, integrals, surface_label)
+        if has_endpoint_dupe:
+            # could assert integrals[-1] == 0 but it offends jax
```
unalmis, line 700: Prior to coding with desc, I would use assertion statements for checks which I didn't want running in production code. For example, on a function designed to be logarithmic time worst case, an assert statement could call a function which checks program correctness in linear time. On checks I did want running in production code, I would throw errors (`raise AssertionError`). An example of this would be guards in functions to prevent bad user input. Then running python with the `python -O` (means optimized mode) would disable assertions but keep the errors.

I think the default for desc module is without the `-O`. If we want to uncomment this, I'll need to do that `jax.cond` thing or some trick with jnp.where

tests/test_compute_utils.py
unalmis, line 1: Adds a test for the new functionality. The other changes are from https://github.com/PlasmaControl/DESC/pull/448/files#r1163580738 since it helped me test faster.

desc/backend.py
```diff
@@ -270,3 +271,29 @@ def while_loop(cond_fun, body_fun, init_val):
         while cond_fun(val):
             val = body_fun(val)
         return val
+
+    def vmap(fun, out_axes=0):
+        """A numpy implementation of jax.lax.map whose API is a subset of jax.vmap.
```
unalmis, line 276: I tested this by copying the same function to the if use_jax section and running the tests.

f0uriest: Would `jnp.vectorize` work? We already use that for some other things I think. If not this is fine.

f0uriest (dismissed):

desc/compute/utils.py
f0uriest: there is a jax-ish way to do that but still experimental and probably not ideal:
https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.checkify.checkify.html

desc/backend.py
unalmis: I couldn't get `jnp.vectorize`  to work when the function changes dimension. The performance of vmap should be at least as fast as vectorize though.
```python
import jax.numpy as jnp
from jax import vmap

a = jnp.arange(6)
inputs = jnp.stack([a, a[::-1], -a])


def f(x):
    return x**3


def g(x):
    # downsample
    return f(x)[: x.size // 2]


def numpy_vmap(f, f_inputs):
    f_inputs = jnp.atleast_2d(f_inputs)
    f_outputs = jnp.stack([f(f_input) for f_input in f_inputs], axis=1)
    return jnp.atleast_1d(jnp.squeeze(f_outputs))


def jax_vmap(f, f_inputs):
    f_inputs = jnp.atleast_2d(f_inputs)
    f_outputs = vmap(f, out_axes=1)(f_inputs)
    return jnp.atleast_1d(jnp.squeeze(f_outputs))


print("1d input, vectorize works like vmap")
print(numpy_vmap(f, a))
print(jax_vmap(f, a))
print(jnp.vectorize(f)(a).T)
print("2d input, vectorize works like vmap")
print(numpy_vmap(f, inputs))
print(jax_vmap(f, inputs))
print(jnp.vectorize(f)(inputs).T)
print("vmap still works, but vectorize fails")
print(numpy_vmap(g, inputs))
print(jax_vmap(g, inputs))
print(jnp.vectorize(g)(inputs).T)
```

desc/compute/utils.py
```diff
@@ -691,15 +690,22 @@ def surface_integrals(grid, q=jnp.array([1]), surface_label="rho", max_surface=F
     # 'nodes', so that the duplicate surface has nodes with the same
     # surface label and is treated as one, like in the previous paragraph.
 
-    integrals = jnp.histogram(nodes_modulo, bins=bins, weights=ds * q)[0]
+    def integrate_component(integrands):
```
unalmis, line 693: I think `vmap` could get compiled each time `surface_integrals` is called because `integrate_component` is a nested function. Would this alternative avoid that
```python
def surface_integrals(...):
    ....
    vector_of_integrals = jnp.atleast_1d(
        jnp.squeeze(vmap_integrate_component(nodes_modulo, bins, vector_of_integrands))
    )

def vmap_integrate_component(nodes_modulo, bins, vector_of_integrands):
    return vmap(functools.partial(integrate_component, nodes_modulo, bins), out_axes=1)(vector_of_integrands)

def integrate_component(nodes_modulo, bins, integrands):
    return jnp.histogram(nodes_modulo, bins=bins, weights=integrands)[0]
```

f0uriest: I don't think we have to worry too much. For most of the regular stuff the compilation doesn't add much overhead (mainly on derivatives). JAX is also usually pretty good about dealing with inlined functions

unalmis:
See #501

PR #500, opened 2023-05-05
Add PEST jacobian, use to test for nestedness

### LINKED ISSUE #490: Update `is_nested` to also check vartheta contours
Because our computational domain is rho/theta/zeta and doesn't know anything about lambda/vartheta, the jacobian can be everywhere positive but we can still get self-intersections in vartheta. 

- [ ] add compute function for jacobian in flux/pest coordinates (we already have one for boozer)
- [ ] use that to check nestedness instead of regular jacobian

f0uriest:
Resolves #490


desc/compute/_basis_vectors.py
```diff
@@ -556,6 +556,32 @@ def _e_sub_zeta_tz(params, transforms, profiles, data, **kwargs):
...
+    coordinates="rtz",
+    data=[
+        "0",
+        "R_t",
+        "Z_t",
+        "lambda_t",
+    ],
+)
+def _e_sub_theta_pest(params, transforms, profiles, data, **kwargs):
+    dtdv = 1 / (1 + data["lambda_t"])
```
unalmis, line 578: don't need to change but i think `dt_dv` is more clear here

tests/test_configuration.py
```diff
@@ -490,6 +490,20 @@ def test_is_nested():
...
+    # just mess with lambda, so rho contours are the same
+    eq.L_lmn += 1e-1 * np.random.default_rng(seed=3).random(eq.L_lmn.shape)
+    grid = QuadratureGrid(10, 10, 0, NFP=eq.NFP)
+    g1 = eq.compute("sqrt(g)", grid=grid)["sqrt(g)"]
+    g2 = eq.compute("sqrt(g)_PEST", grid=grid)["sqrt(g)_PEST"]
+    assert np.all(g1 > 0)  # regular jacobian will still be fine
+    assert np.any(g2 < 0)  # PEST jacobian should be negative
+    assert not eq.is_nested()
+
+
```
unalmis, line 506: Did you want to add that plasma volume test?

```python
data = eq.compute(["V", "V(r)"], grid)
V_PEST = np.sum(data["sqrt(g)_PEST"] * (1 + data["lambda_t"]) * transforms["grid"].weights)
np.testing.assert_allclose(data["V"], V_PEST)
# or maybe
# divergence theorem: integral(dV div [0, 0, Z]) = integral(dS dot [0, 0, Z])
V_of_r_PEST = np.abs(surface_integrals(grid,
    (1 + data["lambda_t"]) * cross(data["e_theta_PEST"], data["e_zeta"])[:, 2] * data["Z"],
))
np.testing.assert_allclose(data["V(r)"], V_of_r_PEST)
```

unalmis (dismissed):

tests/test_configuration.py
rahulgaur104: The strange Jacobian is a consequence of non-monotonicity between theta and theta_PEST. So maybe we could add a function that checks that instead of varying lambda to make the Jacobian fail. Not urgent, though.

desc/compute/_basis_vectors.py
rahulgaur104: I agree with Kaya. If R_t = partial R/partial t, then partial theta/partial v should be dt_dv

ddudt (dismissed):
Looks good, but I like Kaya's suggestions.

tests/test_configuration.py
f0uriest: I didn't add the volume test because when you add in the `(1+ lambda_t)` term it cancels the 1/(1+ lambda_t) term in the PEST jacobian and the overall thing becomes exactly the same as our regular jacobian, so it's not really testing anything.

unalmis (approved):

PR #501, opened 2023-05-05
Function integral

### LINKED ISSUE #271: Profile different options for surface averages, compress/expand operations
There are a few different ways to do surface averaging operations and compress/expanding values to the full grid. We should test and profile different options (both for the forward computation and the jacobian, when used in the full end to end calculation)

- [x] Indexing with grid.unique_rho_idx etc
- [x] Precomputed matrices - ie, surface_average(x) = A*x, expand(x) = P*x
- [x] Histogram method
- [x] For loop
- [x] possibly others.

Another idea: make compress/expand etc methods of grid class?

unalmis:
Generalize integration routine to work with
- [x] vector-valued functions. An example is averaging `data["B"]` on each surface.
- [x] function-valued functions.
- [x] vector-valued function-valued functions. An example would be: $$\symbf{A}(u^\prime, v^\prime) = \frac{1}{4\symrm{\pi}} \iint du dv \frac{\symbf{x}\_{u^{\prime}}(\symbf{B\_{0}}^{\prime} \cdot \symbf{x}\_{v{^\prime}}) - \symbf{x}\_{v^{\prime}} (\symbf{B_{0}}^{\prime} \cdot \symbf{x}\_{u^{\prime}})}{\left \lVert \symbf{x}(u, v) - \symbf{x}(u^\prime, v^\prime) \right \rVert}$$
- [x] integral transform of a vector kernel of 5 variables. This was the motivation to do this pr (This one is the physically meaningful formula) (see [this comment](https://github.com/PlasmaControl/DESC/pull/501#issuecomment-1573000305) if interested in difference with above) 
$$\symbf{A}(u, v) = \frac{1}{4\symrm{\pi}} \iint du^\prime dv^\prime \frac{\symbf{x}\_{u^{\prime}}(\symbf{B\_{0}}^{\prime} \cdot \symbf{x}\_{v{^\prime}}) - \symbf{x}\_{v^{\prime}} (\symbf{B_{0}}^{\prime} \cdot \symbf{x}\_{u^{\prime}})}{\left \lVert \symbf{x}(u, v) - \symbf{x}(u^\prime, v^\prime) \right \rVert}$$
- [x] New method `line_integrals`
- [x] Resolves #271


unalmis:
The ability to integrate function-valued functions should essentially come for free with this new implementation of integration routine.

There are two options (both are tested and work) that we can pick from that would be nice to get feedback on. Pinging y'all so you see; it's not urgent. @ddudt @f0uriest @dpanici 

1. Fully generate the matrix operator that does the surface integral, let's call this `masks` then apply it.
This has the advantage that `masks` can be precomputed in `grid.py`.
If this option is chosen, it may be worth looking into storing `masks` as a sparse matrix for reduced memory and maybe faster computations. See https://jax.readthedocs.io/en/latest/jax.experimental.sparse.html for `jnp` and https://docs.scipy.org/doc/scipy/reference/sparse.html for `np`.
```python
def surface_integrals(...)
    ...
    # compute `masks` either each time surface_integrals is called, or once in grid.py
    def mask_surface(i):
        return inverse_idx == i
    masks = vmap(mask_surface)(jnp.arange(unique_idx.size))
    
    integrals = masks @ jnp.nan_to_num(integrands)
    return integrals
```
2. Don't compute the matrix explicitly. Clearly would use less memory than option 1. Precomputing the matrix is of course not possible with this option.
```python
def surface_integrals(...)
    ...
    # avoid explicitly generating `masks`
    def integrate_surface(i):
        return jnp.nansum(integrands, where=inverse_idx == i])
    integrals = vmap(integrate_surface)(jnp.arange(unique_idx.size))
    return integrals
```

This branch has a finished implementation option 1 with some details not shown here. (although GitHub isn't updating the PR for some reason). I'll do some timing and maybe memory tests to see which is better. Questions:
1. Is the sparse matrix thing worth looking into? Seems like its designed for this use case, but they say its experimental.
2. Speed is unlikely to be issue with either method. Before I profile things, does anyone have a preference to save memory and therefore not store the precomputed matrix?

desc/compute/_geometry.py
```diff
@@ -196,15 +196,19 @@ def _R0_over_a(params, transforms, profiles, data, **kwargs):
...
+                jnp.sqrt(data["g_tt"]),
+                line_label="theta",
+                fix_surface=("rho", max_rho),
+            ),
             surface_label="zeta",
-            max_surface=True,
-        ),
-        surface_label="zeta",
+        )
+        / max_rho
```
unalmis, line 211: Should this be multiplied by `max_rho` instead?

unalmis:
Here are my timing results. The changes to `surface_integrals` are improvements. The changes made to `surface_min` are improvements if you're using numpy, but I think not if you're using jax.
[timing.zip](https://github.com/PlasmaControl/DESC/files/11467762/timing.zip)

desc/compute/_geometry.py
f0uriest: i think this is correct. P ~ r, if we compute P0 at r=r0 but want the value at r=1, thats P1 = P0*1/r0

desc/compute/utils.py
```diff
@@ -620,20 +612,93 @@ def tupleset(t, i, value):
...
+    -----
+        It is assumed that the integration surface has area 4π^2 when the
+        surface label is rho and area 2π when the surface label is theta or
+        zeta. You may want to multiply q by the surface area Jacobian.
+
     Parameters
     ----------
     grid : Grid
         Collocation grid containing the nodes to evaluate at.
     q : ndarray
```
f0uriest, line 698: Should update this with some explanation of the expected shapes of the inputs for function or vector valued integrands

f0uriest (changes requested):
This looks correct, though for my use case it might not actually be that helpful, since creating the function valued integrand over all the surfaces would be like N^6, so I've been doing it manually in a loop of size N^2 to avoid memory issues.

One other thing to consider:
As far as I can see, if the integrand is function values, then we'd expect it to be a square array (ie, pairwise interactions) so we could always assume your f is always either size 1 or size(g)
Similarly, it might make more sense to NOT expand function valued integrands back to the full grid, both for memory savings, and consistency. With regular scalar we have something like g(r,t,z) where we integrate out t and z, but want it to broadcast correctly against other things like h(r,t,z) so we expand back. For function valued integrands, its more like f(r,t,z,t1,z1) where we integrate out over t1 and z1, leaving the result already in the correct shape to broadcast against other functions like h(r,t,z)

Thoughts?

desc/compute/utils.py
unalmis: Updated. In any case the only requirement is that the first dimension has size grid.num_nodes as usual.

unalmis:
> This looks correct, though for my use case it might not actually be that helpful, since creating the function valued integrand over all the surfaces would be like N^6, so I've been doing it manually in a loop of size N^2 to avoid memory issues.
> 
> One other thing to consider: As far as I can see, if the integrand is function values, then we'd expect it to be a square array (ie, pairwise interactions) so we could always assume your f is always either size 1 or size(g) Similarly, it might make more sense to NOT expand function valued integrands back to the full grid, both for memory savings, and consistency. With regular scalar we have something like g(r,t,z) where we integrate out t and z, but want it to broadcast correctly against other things like h(r,t,z) so we expand back. For function valued integrands, its more like f(r,t,z,t1,z1) where we integrate out over t1 and z1, leaving the result already in the correct shape to broadcast against other functions like h(r,t,z)
> 
> Thoughts?

I think I understand now. Let's say h(r, t, z) is the volume Jacobian. Currently if the broadcast `(data["sqrt(g)"] * q.T).T` succeeds, then so will `(data["sqrt(g)"] * surface_integrals(q).T).T`, regardless of the dimension of `q` where `q.ndim <=3`.

As you mention, when the input `q` has 2 dimensions, we typically expect a square array for `q` because `f` and `g` as defined in my comments for the surface integral function will typically both have size `grid.num_nodes`. Because it is a square matrix, we need to be careful about broadcasting; wherever a broadcast succeeds with `q` it will also succeed with `q.T`. Yet only one of those operations can be our intention. The same applies to the algorithm in `surface_integrals`; we should make sure we are doing the integration along the correct axis. You and I have our axes swapped.

The way I visualize what this PR does is as follows.

- As always when the quantity to integrate `q` has shape `(grid.num_nodes, )` for example `q=data["sqrt(g)"]`, I think of it as a function over the nodes in some space like R^3. Each node is mapped to a number and `surface_integrals(q)` will sum, over a particular surface, all of those numbers. It does this for each surface. The output will again be a function in R^3, but now each surface is mapped to a number, which is of course constant over the surface.
- This PR generalizes the integration routine to work for a more complicated integrand, that is when the dimensionality of the integrand `q` increases. When `q` has shape `(grid.num_nodes, f_size)`, now each node is mapped to `f_size` amount of numbers which represents a particular function discretized at `f_size` points. `surface_integrals(q)` will sum, over a particular surface, all of those functions. It does this for each surface. The output maps each surface to this summed function. If there are 10 surfaces on the grid, this approach semantically performing 10 surface integrals, so the output will be ten such unique summed functions. (Before expanding, `surface_integrals(q).shape == (10, f_size)`.).

The point is, each node is mapped to an array `f` and this PR is summing all the `f`s of the node on a surface in the computational domain (rho, theta, zeta).

Now I think what you want to do is still map each node to `f` but instead perform `surface_integrals(f)` for every node. Then your output will map each node to a number which is potentially different from all the other nodes on a particular surface in the computational domain (rho, theta, zeta). This approach can be semantically thought of as performing `grid.num_nodes` surface integrals. With this intention, I agree that you should NOT upsample/expand the output because you already have your `grid.num_nodes` amount of unique quantities which is going to be broadcastable with quantities like `data["sqrt(g)"]`.

Although the intentions of these two algorithms are different, their implementations only differ by a transpose.
When `q` has shape (grid.num_nodes, f_size), the main operation in the aforementioned approach is `expand(masks @ q)`. This is broadcastable with shape `(grid.num_nodes, f_size)`. If the latter approach is your intention, the main operation is `(masks @ q.T).T` which is still broadcastable but now with shape `(f_size, surface_label_unique_size)`. The last dimension can of course be expanded with an additional broadcast, but I believe your intention was to say that expansion is unnecessary because the broadcast is already valid.

My vote is to keep `surface_integrals` as is currently in this PR because the semantics of the above approach are more consistent with the one dimensional case. That is, there are only 10 unique values of the surface label, there is also 10 surface integrals being performed. For your use case, we should define a new function because it's intention is semantically different than before (semantically there are `grid.num_nodes `surface_integrals being performed).

unalmis:
For your comment about memory, I don't follow your memory analysis. In either case of the above algorithms, no extra computation is being performed. So iff you can get the job done like
```python
for i in range(N):
    q_i = make quantity
    do stuff with surface_integrals(q_i)
```
Then I think generating the input for `surface_integrals(q_all_i)` will only use `N * q_i` memory. If that's too expensive, then I think your only option is loops.

unalmis:
I've updated to not `expand` `integrals` when `q` is a function-valued integrand (square matrix). With the implementation, the following still applies for any `q`
> Let's say h(r, t, z) is the volume Jacobian. Currently if the broadcast (data["sqrt(g)"] * q.T).T succeeds, then so will (data["sqrt(g)"] * surface_integrals(q).T).T, for any dimension of `q` from 0 to 3.

unalmis:
I'll add a test for basic case (where I compare the output to a value I compute by hand) for integrals of the type in the description of this PR. That should resolve ambiguity about how you want your input to be transposed.

f0uriest:
> Or did you have a different algorithm for integrating a 2d quantity such as:
Still map each node to some f but instead perform surface_integrals(f) for every node. Then your output will map each node to a function which is potentially different from other nodes' on a particular surface in the computational domain (rho, theta, zeta). With this intention, you should not expand the output because you already have your grid.num_nodes amount of unique quantities which is going to be broadcastable with quantities like data["sqrt(g)"].

Ahh yeah this is what I was thinking of. The function `F(r,t,z) = integral(f(r,t,z,t1,z1), t1, z1)` would be a function of all 3 coordinates, so it would require `grid.num_nodes` different surface integrals to evaluate everywhere. I agree it probably makes sense to make that a separate function.

For the memory analysis I was treating `N` is the number of nodes in 1 dimension, so the grid would have `N^3`, and if you're doing an all-to-all surface op, it would be `N^6` (or maybe could reshape things to get down to `N^5` since its really `N` integrals each of which is integrating `N^2 x N^2` points)

f0uriest (approved):

desc/compute/utils.py
```diff
@@ -649,61 +749,141 @@ def surface_integrals(grid, q=jnp.array([1]), surface_label="rho", max_surface=F
...
+    # products. However, timing results showed no difference.
+    axis_to_move = (integrands.ndim == 3) * 2
+    integrals = jnp.moveaxis(
+        masks @ jnp.moveaxis(integrands, axis_to_move, 0), 0, axis_to_move
+    )
+    return integrals
+
+
+def surface_integral_transforms(grid, q=jnp.array([1.0]), surface_label="rho"):
+    """Compute the integral transform of a kernel over each surface in the grid.
```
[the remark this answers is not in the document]
unalmis: I don't see the issue

unalmis:
> For the memory analysis I was treating `N` is the number of nodes in 1 dimension, so the grid would have `N^3`, and if you're doing an all-to-all surface op, it would be `N^6` (or maybe could reshape things to get down to `N^5` since its really `N^3` integrals each of which is integrating `N^2` points)

Then maybe the `surface_integrals_tranform_map` method will be useful wrapper for you.

desc/compute/utils.py
```diff
@@ -649,63 +781,120 @@ def surface_integrals(grid, q=jnp.array([1]), surface_label="rho", max_surface=F
...
-    # 'nodes', so that the duplicate surface has nodes with the same
-    # surface label and is treated as one, like in the previous paragraph.
-
-    integrals = jnp.histogram(nodes_modulo, bins=bins, weights=ds * q)[0]
+    # Todo: Define masks as a sparse matrix once sparse matrices are
+    #       are no longer experimental in jax.
+    # The ith row of masks is True only at the indices which correspond to the
+    # ith surface. The integral over the ith surface is the dot product of the
+    # ith row vector and the vector of integrands of all surfaces.
+    masks = inverse_idx == jnp.arange(unique_size)[:, jnp.newaxis]
```
unalmis, line 793: This is a boolean array, so even though it's shape is N x N^3, its memory is N^4/8 <= 3N^3 = grid.nodes almost always, and therefore much smaller than the grid itself.

dpanici: is this $N^{4/8}$ or $N^4 / 8$?

unalmis: the latter. seems scary but it actually uses less memory (https://github.com/PlasmaControl/DESC/pull/501#issuecomment-1546307972) than the implementation on master. I think the idea is that matrix operations are easy to differentiate through.

unalmis: The sparse matrix is actually implementable now with `masks = sparse.BCOO.fromdense(inverse_idx == jnp.arange(unique_size)[:, jnp.newaxis])`. but it didn't seem worth it

f0uriest (dismissed):
Looks great, the `surface_integral_map` stuff is basically exactly what I need.

One question: Given that the `surface_integral_map` thing only needs to know about the grid, is there any advantage to creating it once in the base `compute` and then passing the function in as a dependency to all the lower level stuff? To avoid creating the big matrix multiple times? From the timing it seemed like that's not a bottleneck, but something to consider (for a future PR, I think this one is fine as is)

unalmis:
> One question: Given that the `surface_integral_map` thing only needs to know about the grid, is there any advantage to creating it once in the base `compute` and then passing the function in as a dependency to all the lower level stuff? To avoid creating the big matrix multiple times?

Basically just that. It could be worthwhile to make it a dependency of certain quantities in `data_index`. Though I would suggest making it an intermediate dependency, that is we should delete it after all things are finished computing. If it's persistent, there's probably not much advantage over just precomputing it in `grid.py`. I didn't do that because it seemed redundant to store three objects each of size $N^4 / 64$ in the grid when they can be reconstructed from the already stored `inverse_idx` each of size $N^3$.

f0uriest (dismissed):

desc/compute/utils.py
```diff
@@ -683,63 +813,124 @@ def surface_integrals(grid, q=jnp.array([1]), surface_label="rho", max_surface=F
...
+        # by the last two dimensions of `integrands`, for every element along the
+        # previous dimension of `integrands`. Therefore, when `integrands` has three
+        # dimensions, the second must hold g. We may choose which of the first and
+        # third dimensions hold f and v. The choice below transposes `integrands` to
+        # shape (v.size, g.size, f.size). As we expect f.size >> v.size, the
+        # integration is in theory faster since numpy optimizes large matrix
+        # products. However, timing results showed no difference.
+        axis_to_move = (integrands.ndim == 3) * 2
+        integrals = jnp.moveaxis(
+            masks @ jnp.moveaxis(integrands, axis_to_move, 0), 0, axis_to_move
```
unalmis, line 918: The GitHub CI failed (twice) a new augmented Lagrangian test because a value is positive with order $10^{-12}$ when it should be negative. I can't reproduce the failure locally with `pytest -k test_constrained_AL_scalar tests/test_optimizer.py`. I'll try testing on Princeton's clusters.

Seems unlikely but I'll try changing `@` to `jnp.matmul(precision=Precision.HIGH)` and see if it fixes.

f0uriest (approved):

desc/compute/utils.py
```diff
@@ -820,22 +1194,29 @@ def surface_min(grid, x, surface_label="rho"):
...
     def body(i, mins):
         mins = put(mins, inverse_idx[i], jnp.minimum(x[i], mins[inverse_idx[i]]))
         return mins
 
-    return fori_loop(0, len(inverse_idx), body, mins)
+    mins = fori_loop(0, inverse_idx.size, body, mins)
+    # The above implementation was benchmarked to be more efficient, after jit
+    # compilation, than the alternative given in the two lines below.
+    # masks = inverse_idx == jnp.arange(unique_size)[:, jnp.newaxis]  # noqa: E501,E800
+    # mins = jnp.amin(x[jnp.newaxis, :], axis=1, initial=jnp.inf, where=masks)  # noqa: E501,E800
```
ddudt, line 1221: Do we need to keep this record in the code?

ddudt (approved):

desc/compute/utils.py
f0uriest: Could probably eventually move to some sort of design doc or dev guide but I think it's fine to leave it in.

unalmis: We don't need to. Some of my other comments may be redundant too; I mainly included them to make reviewing easier. I'll remove this comment from my add_all_limits pr. If there are other cosmetic changes just mark them and I'll do the same.

PR #503, opened 2023-05-08
Adding augmented lagrangian for constrained optimization

### LINKED ISSUE #419: Add wrapper to turn inequality constraints into bound constraints via slack variables
Or equality constraints via quadratic slack variables?

pkim1818:
augmented lagrangian (scalar and least squares) along with utils for inequality constraints.

Resolves #419


f0uriest (dismissed):

desc/optimize/_desc_wrappers.py
```diff
@@ -1,11 +1,200 @@
...
+        x_scale[j]. Improved convergence may be achieved by setting x_scale such that
+        a step of a given size along any of the scaled variables has a similar effect
+        on the cost function. If set to ‘jac’, the scale is iteratively updated using
+        the inverse norms of the columns of the Jacobian matrix.
+    verbose : int
+        * 0  : work silently.
+        * 1 : display a termination report.
+        * 2 : display progress during iterations
+    stoptol : dict
+        Dictionary of stopping tolerances, with keys {"xtol", "ftol", "gtol",
```
dpanici, line 51: missing `ctol` in this list

desc/optimize/_desc_wrappers.py
```diff
@@ -1,11 +1,200 @@
...
```
dpanici, line 145: missing `ctol` in this list

desc/equilibrium/equilibrium.py
```diff
@@ -564,7 +565,7 @@ def optimize(
             Objective function to satisfy. Default = fixed-boundary force balance.
         optimizer : str or Optimizer (optional)
             optimizer to use
-        ftol, xtol, gtol : float
+        ftol, xtol, gtol, ctol : float
```
dpanici, line 568: probably a dumb question but ctol is the tolerance on satisfying the constraints given? So if we give `F balance < 1e-2` as the nonlinear constraint, the `ctol` would hit if we have `F balance` change by `1e-2` or less?

desc/optimize/aug_lagrangian.py
```diff
@@ -0,0 +1,336 @@
...
+        condition is disabled.
+    xtol : float or None, optional
+        Tolerance for termination by the change of the independent variables.
+        Default is 1e-8. Optimization is stopped when
+        ``norm(dx) < xtol * (xtol + norm(x))``. If None, the termination by
+        this condition is disabled.
+    gtol : float or None, optional
+        Absolute tolerance for termination by the norm of the gradient. Default is 1e-8.
+        Optimizer teriminates when ``norm(g) < gtol``, where
+        If None, the termination by this condition is disabled.
```
dpanici, line 78: missing `ctol` in the docstring here

desc/optimize/aug_lagrangian_ls.py
```diff
@@ -0,0 +1,327 @@
...
+        condition is disabled.
+    xtol : float or None, optional
+        Tolerance for termination by the change of the independent variables.
+        Default is 1e-8. Optimization is stopped when
+        ``norm(dx) < xtol * (xtol + norm(x))``. If None, the termination by
+        this condition is disabled.
+    gtol : float or None, optional
+        Absolute tolerance for termination by the norm of the gradient. Default is 1e-8.
+        Optimizer teriminates when ``norm(g) < gtol``, where
+        If None, the termination by this condition is disabled.
```
dpanici, line 78: missing `ctol` in the docstring

desc/optimize/least_squares.py
```diff
@@ -389,6 +389,7 @@ def lsqtr(  # noqa: C901 - FIXME: simplify this
         cost=cost,
         fun=f,
         grad=g,
+        v=v,
```
dpanici, line 392: Is this used elsewhere in the code? or just good to have in the result?

desc/equilibrium/equilibrium.py
f0uriest: no, `ctol` is a criteria on the infinity norm (max abs) of the constraint violation, not a relative tolerance like `ftol`

So if your constraint is `F < 1e-2` and `ctol = 1e-4` then it would really stop when `F < (1e-2 + 1e-4)`

tests/test_optimizer.py
```diff
@@ -725,3 +736,185 @@ def hess(x):
...
+
+    constraint = NonlinearConstraint(con, -np.inf, 0, conjac, conhess)
+    x0 = rng.random(n)
+
+    out1 = fmin_auglag(
+        fun,
+        x0,
+        grad,
+        hess=hess,
+        bounds=(-jnp.inf, jnp.inf),
```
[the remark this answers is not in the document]
f0uriest: these are bounds on the independent variable, not the function value.

tests/test_optimizer.py
```diff
@@ -725,3 +736,185 @@ def hess(x):
...
+    constraints += (
+        Volume(bounds=Vbounds),
+        AspectRatio(bounds=ARbounds),
+        MagneticWell(bounds=(0, jnp.inf)),
+        ForceBalance(bounds=(-1e-3, 1e-3), normalize_target=False),
+    )
+    H = eq.compute(
+        "curvature_H", grid=LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
+    )["curvature_H"]
+    obj = ObjectiveFunction(MeanCurvature(target=H))
```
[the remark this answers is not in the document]
f0uriest: The initial equilibrium is Mercier unstable, so the goal is to stabilize it without changing the shape too much.

tests/test_optimizer.py
```diff
@@ -725,3 +736,185 @@ def hess(x):
...
+    V = eq.compute("V")["V"]
+    AR = eq.compute("R0/a")["R0/a"]
+    constraints += (
+        Volume(target=V),
+        AspectRatio(target=AR),
+        MagneticWell(bounds=(0, jnp.inf)),
+        ForceBalance(bounds=(-1e-3, 1e-3), normalize_target=False),
+    )
+    # Dummy objective to return 0, we just want a feasible solution.
+    obj = ObjectiveFunction(GenericObjective("0"))
```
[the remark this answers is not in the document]
f0uriest: Yes, this is just meant to find a feasible solution (as before, initial guess is Mercier unstable) that is stable with the same aspect ratio and volume

desc/optimize/utils.py
```diff
@@ -220,27 +357,29 @@ def print_iteration_nonlinear(
     if cost is None or abs(cost) == np.inf:
         cost = " " * 15
     else:
-        cost = "{:^15.4e}".format(cost)
+        cost = "{: ^15.3e}".format(cost)
```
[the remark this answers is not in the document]
f0uriest: its part of the format specifier: https://docs.python.org/3/library/string.html#format-specification-mini-language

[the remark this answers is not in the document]
ddudt: This is just the fill character? It sounds like it defaults to a space anyways. My point is that we should be consistent: lines 360 and 365 explicitly ask for a space, while lines 370 and 375 (and everywhere else in the code, I think) omit it (and I think they give the same behavior.

ddudt (approved):

desc/optimize/utils.py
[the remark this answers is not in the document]
f0uriest: this tells it to print a space or a minus sign. It's really only an issue for things that can be negative (most things we print aren't, but with the augmented lagrangian the cost and delta cost can be) and it's also only an issue when we want to align things vertically.

ie, as we had it previously, things would look like:
```
Cost reduction
1.234
-1.234
2.234
-4.567
``` 

while this change makes it looks like
```
Cost reduction
 1.234
-1.234
 2.234
-4.567
``` 

so that things are lined up all pretty

PR #504, opened 2023-05-09
New function for mapping between coordinate systems

f0uriest:
- Introduces a new function/method `Equilibrium.map_coordinates` which generalizes the existing methods `compute_theta_coordinates` and `compute_flux_coordinates`, but allows mapping between arbitrary coordinates.
- Adds analytic derivatives of many things to `desc.compute` so that analytic jacobians can be used for the mapping.
- Replaces usage of old `compute_*_coordinates` in `desc.plotting` with the new `map_coordinates` function.
- Also renames `zeta` -> `phi` in places in `desc.plotting` to be consistent with when we mean the computational coordinate `zeta` vs the physical coordinate `phi`

This will be needed for plotting things correctly with a generalized toroidal angle, but may also be useful in standalone.


desc/equilibrium/coords.py
```diff
@@ -5,14 +5,143 @@
...
+    evaluates outbasis at those locations.
+
+    NOTE: this function cannot be JIT compiled or differentiated with AD.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium to use
+    coords : ndarray, shape(k,3)
+        2D array of input coordinates. Each row is a different
```
dpanici, line 29: shouldn't this be "each column"? shape(k,3) means there are k rows and 3 columns?

or do you mean coordinate like (R1,phi1,Z1) versus (R2,phi2,Z2) i.e. different points in space?

desc/equilibrium/coords.py
```diff
@@ -5,14 +5,143 @@
...
+        idx = put(idx, i, k)
+        return idx
+
+    idx = fori_loop(0, len(coords), _distance_body, idx)
+    yk = yg[idx]
+    alpha = 0.5
+    yk = fixup(yk)
+    resk = residual(yk)
+
+    for i in range(maxiter):
```
dpanici, line 106: Just to make sure I get the algorithm.

You find the initial guesses with a coarse grid nearest neighbor.

Then you are doing a Newton method on $$\mathbf{f}(\mathbf{x}) = \mathbf{yk} - coords$$
where $$\mathbf{x} = (\rho,\theta,\zeta)$$ and  $$\mathbf{x}^{i+1} =  \mathbf{x}^{i} - \mathbf{Df}^{-1}(\mathbf{x}^{i}) @ \mathbf{f}(\mathbf{x^i})$$
then moving $x^i$ over and defining $$\mathbf{d} = \mathbf{x}^{i+1} - \mathbf{x}^i$$ and multiplying by the jacobian yields the least squares system you solve for $\mathbf{d}$:
$$\mathbf{Df}(\mathbf{x}^{i}) \mathbf{d} = -\mathbf{f}(\mathbf{x}^{i})$$

Then you update the guess at the correct coordinates with $\mathbf{d}$ using a backtracking line search.

Did I miss something tho or should your `resk` be negative, based off of my derivation here? 

Or wait, you define $\mathbf{d}$ the opposite of me, so that $\mathbf{x}^{i+1} = \mathbf{x}^i - \mathbf{d}$, ok that clears up my confusion

desc/plotting.py
```diff
@@ -1095,29 +1095,47 @@ def plot_section(
         fig, ax = plot_section(eq, "J^rho")
 
     """
+    if "nzeta" in kwargs:
```
dpanici, line 1098: did not rename this in the docstring

dpanici (changes requested):
I had a tiny change in the docstring of plot_section

desc/plotting.py
f0uriest: fixed

dpanici (approved):

dpanici (approved):

unalmis (approved):

desc/equilibrium/coords.py
```diff
@@ -5,14 +5,143 @@
...
+
+    basis_derivs = [f"{X}_{d}" for X in inbasis for d in ("r", "t", "z")]
+    for key in basis_derivs:
+        assert (
+            key in data_index.keys()
+        ), f"don't have recipe to compute partial derivative {key}"
+
+    # can't use AD or jit compile this because of grid stuff
+    def compute(y, basis):
+        data = eq.compute(basis, grid=Grid(y, sort=False))
```
unalmis, line 57: Can you jit this by passing in the already constructed grid as parameter instead of y?

f0uriest: in theory yes but the grid would need to be a static arg, so it would recompile on every iteration, so probably not worth it.

PR #505, opened 2023-05-09
Fix docs building

f0uriest:
- RTD has been failing due to timeouts recently, so this PR reduces the toctree depth which seems to make it build a lot faster.
- Also fixes lots of small formatting issues that were causing warnings and errors during the docs build
- RTD is now set to fail on warnings, to hopefully avoid future formatting problems from creeping into master.


desc/objectives/_equilibrium.py
```diff
@@ -15,14 +15,23 @@
...
+    r"""Radial and helical MHD force balance.
 
-    F_rho = sqrt(g) (B^zeta J^theta - B^theta J^zeta) - grad(p)
-    f_rho = F_rho |grad(rho)| dV  (N)
+    Given force densities:
 
-    F_helical = sqrt(g) J^rho
-    e^helical = -B^zeta grad(theta) + B^theta grad(zeta)
-    f_helical = F_helical |e^helical| dV  (N)
+    Fᵨ = √g (B^ζ J^θ - B^θ J^ζ) - ∇ p
```
dpanici, line 22: woah how did you make this format like that with the special characters?

dpanici (approved):

desc/objectives/_equilibrium.py
unalmis: i think you can just copy paste the output of latex since it's unicode

f0uriest: Yeah just Unicode. Figured thats the best option for having both the raw text and html docs be readable

PR #518, opened 2023-05-30
Add wrapper for user defined objective function

### LINKED ISSUE #185: Easily wrap user defined objective functions
In addition to the predefined objectives, we should be able to wrap a function from the user in an ``Objective`` to make it easy to work with the rest of the code. One possibility would be to allow the user to define any function of the form ``func(data) -> ndarray`` where ``data`` is a dictionary containing any of the quantities from ``data_index``. We can then parse the user supplied function to find out what data is actually needed and create an objective that first calls the necessary compute function to get the raw variables, then calls the user function to calculate the final desired quantity.

The following snippet shows an example of parsing user code to determine which variables are needed:
```
def foo(data):
    x = data["x"]
    y = data["y"]
    z = x*y
    return z

import inspect, re
pattern = r"data\[(.*?)\]"
src = inspect.getsource(foo)
variables = re.findall(pattern, src)
variables = [s.replace("'", "").replace('"', "") for s in variables]
print(variables)
>>> ['x', 'y']
```

### LINKED ISSUE #234: Easy wrapper for generic linear constraints
Similar to what @ddudt was doing for the vmec subspace thing, it would be handy to have a class that can easily wrap any user defined linear constraint of the form ``Ax=b``

possible idea:
```
class LinearConstraintFromUser:
    """Wrap a user supplied linear constraint

     Parameters
     ----------------
    fun : callable
        function that takes an equilibrium and returns A matrix and b vector and list of args, such that the constraint is A * x[args] - b = 0

    """
    def __init__(self, fun):
```

f0uriest:
Resolves #185 
Resolves #234


dpanici (dismissed):

desc/objectives/_generic.py
```diff
@@ -12,6 +14,149 @@
...
+            grid = QuadratureGrid(eq.L_grid, eq.M_grid, eq.N_grid, eq.NFP)
+        else:
+            grid = self._grid
+
+        def getvars(fun):
+            pattern = r"data\[(.*?)\]"
+            src = inspect.getsource(fun)
+            variables = re.findall(pattern, src)
+            variables = [s.replace("'", "").replace('"', "") for s in variables]
+            return variables
```
unalmis, line 111: well that's neat

unalmis (dismissed):

desc/objectives/_generic.py
```diff
@@ -12,6 +14,149 @@
...
+        self,
+        fun,
+        eq=None,
+        target=None,
+        bounds=None,
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        grid=None,
+        name="generic",
```
ddudt, line 72: change to `"custom"`?

ddudt (changes requested):
Very cool! Just one minor comment.

ddudt (approved):

unalmis (approved):

PR #523, opened 2023-05-30
Make plotting kwargs a bit more uniform

### LINKED ISSUE #420: Make plot kwargs uniform across different functions
On me as I did this change originally, some plot functions take `zeta` which can be either a float or an int, others take `nzeta` which can only be an int. Should change this to be consistent API

f0uriest:
- Renames some keyword arguments to plotting functions to have more consistent names, and adds some missing kwargs.
- Also renames all instances of `zeta`, `nzeta`, `nphi` to just `phi`, which can be a float, int, or array of toroidal angles to plot.

Resolves #420


ddudt (approved):

rahulgaur104 (approved):
Seems straightforward.

PR #525, opened 2023-05-31
Add functions for finding most/least rational surfaces

### LINKED ISSUE #6: Add node option to avoid axis / rational surfaces
Flag for avoiding axis could help with debugging stuff there

Also having option to place at "most rational" surfaces.

f0uriest:
Given an iota profile, we can find the "most rational" surfaces in the plasma. We can them attempt to avoid these, giving some notion of "least rational" surfaces. This PR provides 2 main functions and a number of utilities for finding rational surfaces, and finding points most distant from them.

Resolves #6

### DIFF
--- a/desc/grid.py
+++ b/desc/grid.py
@@ -1,11 +1,17 @@
 """Classes for representing flux coordinates."""
 
 import numpy as np
-from scipy import special
+from scipy import optimize, special
 
 from desc.io import IOAble
 
-__all__ = ["Grid", "LinearGrid", "QuadratureGrid", "ConcentricGrid"]
+__all__ = [
+    "Grid",
+    "LinearGrid",
+    "QuadratureGrid",
+    "ConcentricGrid",
+    "find_most_rational_surfaces",
+]
 
 
 class Grid(IOAble):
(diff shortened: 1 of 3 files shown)


dpanici (dismissed):

desc/grid.py
```diff
@@ -1101,3 +1118,182 @@ def most_rational(a, b):  # pragma: no cover
...
+        values of iota at least rational surfaces
+    rho_rat : ndarray
+        rho values of lowest order rational surfaces
+    io_rat : ndarray
+        iota values at lowest order rational surfaces
+    """
+    rho_rat, io_rat = find_most_rational_surfaces(
+        iota, nrational, atol, itol, eps, **kwargs
+    )
+    io = find_most_distant(io_rat, n, tol=atol, **kwargs)
```
ddudt, line 1297: Shouldn't we also pass in the following arguments here: `a=iota(0), b=iota(1)`?
Otherwise there could be a nice "area of irrationality" near the axis/edge that gets missed.

ddudt (commented):
The way this computes the "most irrational" surfaces as furthest away from the most rational surfaces isn't perfect, but I can't think of a better method at the moment.

desc/grid.py
f0uriest: yeah good point, will do.

f0uriest:
I agree the "most irrational" part isn't perfect, but from doing a bit of literature review it seems like there probably aren't many better options that are simple to implement and would give reasonable results. IE, in theory the golden ration is "very irrational" and anything similar to that with small variations in the coefficients of the continued fraction would also be very irrational, but we don't want a bunch of nodes clustered around one very irrational value.

In testing it seems like just finding points farthest from the most rationals seems to give "sufficiently irrational" results. The method is also connected to Leja points and there's a bunch of theory on that and using them as quadrature and interpolation nodes, so I think it's reasonable.

ddudt (approved):

PR #537, opened 2023-06-15
Make desc classes compatible with jax pytrees

### LINKED ISSUE #532: Turn DESC classes into JAX pytrees
In JAX parlance, a "pytree" is a (possibly nested) container of arrays and data, like `list`, `dict`, `tuple` etc. Basically, it means its a type that JAX knows about, and can be passed into jit compiled functions as normal, without having to close over or declare as a static argument.

see: https://jax.readthedocs.io/en/latest/pytrees.html

It would be useful, and possibly necessary, to allow DESC classes to act as pytrees. I *think* it should be straightforward:

- need to define a function/method for each class that converts the class data into a tuple of arrays or other jax types
- another function to re-create the object from a tuple

We can probably reuse a lot of the io infrastructure, with the caveat that we don't save all attributes of objects but we probably do want to use all of them in the pytree, so maybe a bit complicated.

### LINKED ISSUE #540: Reduce memory usage
Currently DESC seems to use quite a bit of memory, so that we can't solve equilibria with spectral resolution > L=M=N~14, even on an A100 GPU with 80gb of memory. There are probably things we can do to reduce the memory usage and let us go to higher resolution.

f0uriest:
- When we "close over" stuff like transforms and other big arrays in the `objective.compute` methods, it can lead to larger than expected memory usage and longer compile times as those matrices get baked into the computational graph that JAX compiles.
- Previously, we couldn't pass these things in as extra arguments, because arguments to jit compiled funcitons must be jax types (ndarray, or pytree)
- This PR registers all desc `IOAble` classes as jax pytrees allowing them to cross the jit boundary unaltered. 
- The base `IOAble` class has a default method for converting things to pytrees which treats `int`, `bool`, `str` and a few other types as constants, since those are usually used as indices/flags etc that shouldn't be changed during optimization. If a particular class needs other behavior, you can override the default method (but the default seems to work fine for basically all of our existing classes).
- Also adds a new attribute to all objectives, `objective.constants` which is a dictionary of constant properties, such as transforms and profiles. This is then passed into `objective.compute` as an extra arg/kwarg depending on the situation.
- Combined with #539 this significantly reduces memory usage, such that we can go up to `L=M=N=20` with the grid oversampled by 2 in each direction on a single A100 gpu. Free boundary can go up to at least `L=M=N=12` (haven't tried higher but probably can).

Resolves #532 
Resolves #540


desc/io/equilibrium_io.py
```diff
@@ -71,7 +74,79 @@ def load(load_from, file_format=None):
...
+        return ("ndarray", x.shape, tuple(x.flatten()))
+    return x
+
+
+def _unmake_hashable(x):
+    # turn tuple of ints and shape to ndarray
+    if isinstance(x, tuple) and x[0] == "ndarray":
+        return np.array(x[2]).reshape(x[1])
+    return x
+
```
unalmis, line 101: it could be better to make a hashable array class as outlined here https://github.com/google/jax/issues/4572#issuecomment-709809897

f0uriest: I looked into something like that and it's probably overkill. The real issue is hashing the actual values of the array, and converting it to a tuple and hashing that seemed the fastest option of the things I tried.

unalmis: I was thinking just copying the [tuple hash](https://github.com/python/cpython/blob/6e9f83d9aee34192de5d0ef7285be23514911ccd/Objects/tupleobject.c#L319) in python, but it's fine if this works

docs/adding_objectives.rst
```diff
@@ -129,6 +137,10 @@ A full example objective with comments describing key points is given below:
             # equilibrium, though in most cases that isn't necessary.
             self._profiles = get_profiles(self._data_keys, eq=eq, grid=self.grid)
             self._transforms = get_transforms(self._data_keys, eq=eq, grid=self.grid)
```
dpanici, line 139: this is not specific to this PR and I should probably just make an issue, but these lines using the  helper functions are a bit opaque as it is not clear what self._transforms etc are, I guess is not that important but just thought for clarity could be good to have this be more transparent of at least what type of object are and what they contain

dpanici (approved):

desc/objectives/_geometry.py
```diff
@@ -227,6 +233,10 @@ def build(self, eq=None, use_jit=True, verbose=1):
 
         self._profiles = get_profiles(self._data_keys, eq=eq, grid=grid)
         self._transforms = get_transforms(self._data_keys, eq=eq, grid=grid)
```
unalmis, line 235: So these have both attributes and copies of the reference in the constants attribute?

desc/objectives/_qs.py
```diff
@@ -178,15 +184,17 @@ def compute(self, *args, **kwargs):
...
             params=params,
-            transforms=self._transforms,
-            profiles=self._profiles,
+            transforms=constants["transforms"],
+            profiles=constants["profiles"],
         )
-        B_mn = self._matrix @ data["|B|_mn"]
-        return B_mn[self._idx]
+        B_mn = constants["matrix"] @ data["|B|_mn"]
+        return B_mn[constants["idx"]]
```
unalmis, line 197: are we basically making everything with not a small constant size, a constant

unalmis (approved):

desc/objectives/_qs.py
f0uriest: That was the idea. In theory it also allows you to call an objective function and override the default constants, so that in the future we could break some parts out into standalone functions that would be compiled ahead of time, if jax ever gets that working

PR #538, opened 2023-06-15
Add retry step to codecov upload

f0uriest:


dpanici (approved):
I forget, are the test durations something that is automatically updated, or do we need to manually update them periodically? I assume the latter?

f0uriest:
> I forget, are the test durations something that is automatically updated, or do we need to manually update them periodically? I assume the latter?

We have to update manually. In this case i used the median time for each test from the last few months of running on CI

PR #539, opened 2023-06-16
Add "looped" option for computing jacobian matrix

f0uriest:
The standard way jax computes the jacobian matrix is with ``vmap(jvp(fun))``, which attempts to vectorize the jacobian vector product to compute the entire jacobian at once. This is usually the fastest way, but because it has to materialize all the intermediate matrices in memory at the same time, it limits the resolution we can use.

Another option is to instead manually loop over the columns of the jacobian and build it one at a time. This is a bit slower, but much more memory efficient.

Some profiling results:

[looped](https://gist.github.com/f0uriest/9f5f915bede37d3f68868b29979e8bd2)
[master](https://gist.github.com/f0uriest/c3d7c15642bae9e563048c4beb0b9787)

On master, on an A100, we max out around L=M=N=14, with the nodes oversampled by 2 in each direction (probably overkill)

With the looped jacobian for the same specs, we max out around L=M=N=20.

Overall speed is a bit slower with the looped jacobian: 
image

But still a very useful option when we need higher resolution.

### DIFF
--- a/desc/derivatives.py
+++ b/desc/derivatives.py
@@ -150,7 +150,7 @@ class AutoDiffDerivative(_Derivative):
         return self._compute(*args)
 
     @classmethod
-    def compute_jvp(cls, fun, argnum, v, *args):
+    def compute_jvp(cls, fun, argnum, v, *args, **kwargs):
         """Compute df/dx*v.
 
         Parameters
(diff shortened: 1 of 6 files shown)


ddudt (approved):
This is great!

PR #546, opened 2023-06-21
Add fft interpolation for future use in free boundary

f0uriest:
This will be needed for the high order singular integration method used in the free boundary stuff, but figured it can stand on it's own.


desc/interpolate.py
```diff
@@ -8,6 +8,100 @@
...
+        Shift in x to evaluate at. If original data is f(x), interpolates to f(x + sx)
+    dx : float
+        Spacing of source points
+
+    Returns
+    -------
+    fi : ndarray, shape(n, ..., len(sx))
+        Interpolated (and possibly shifted) data points
+    """
+    c = jnp.fft.ifft(f, axis=0)
```
dpanici, line 30: kind of confused, I thought that the interp FFT sohuld work by FFT the input f, then eventually calling ifft to evaluate the coefficients of f at more grid points? I must be misssing something because this starts with ifft?

f0uriest: They're equivalent (`fft` and `ifft` are transposes of each other), it just depends on the sign convention (see https://numpy.org/doc/stable/reference/routines.fft.html#implementation-details)

I'm more used to the positive sign on the forward transform, which in numpy/jax is `ifft`

ddudt (approved):

dpanici (approved):

PR #547, opened 2023-06-23
Add BFGS option for augmented lagrangian

### LINKED ISSUE #553: Implement vector-jacobian products for objectives
In constrained optimization, we often need to evaluate stuff like `y.T * J` where `y` are Lagrange multipliers and `J` is the Jacobian of the constraints. Currently we evaluate this directly using the full Jacobian matrix, but this is a bit wasteful since it can be evaluated much faster with reverse mode AD. JAX already has functionality for this exact sort of thing, so we'd just need to wrap that and add an interface like `objective.jvp`

f0uriest:
- Modifies `fmintr` to allow an externally passed `BFGS` object to maintain state between calls
- Adds BFGS option to augmented lagrangian
- Adds calculation of vector jacobian products to objectives
- uses vjp in augmented lagrangian rather than explicitly computing full jacobian

Resolves #553


ddudt (approved):
I'm not sure why `test_scipy_fail_message` is now failing; otherwise this looks good

f0uriest:
Looks like the latest version of scipy (``1.11.0``, released yesterday) slightly changes how optimizer callbacks are handled. For now I'm just pinning the version to ``< 1.11.0``, I'll work on figuring out a proper fix later.

desc/optimize/_constraint_wrappers.py
```diff
@@ -768,6 +802,36 @@ def hess(self, x):
...
+
+        Parameters
+        ----------
+        v : ndarray
+            Vector to left-multiply the Jacobian by.
+        x : ndarray
+            Optimization variables.
+
+        """
+        raise NotImplementedError
```
dpanici, line 818: couldn't this just call the jac_scaled and multiply the two vectors?

f0uriest: Technically yes but that's the expensive way that this is trying to avoid, so it seemed better to make the user do that explicitly

ddudt (approved):

PR #549, opened 2023-06-23
plot_boozer_surface improvements

daniel-dudt:
- Add `rho` argument for specifying the surface to plot
- Add `fieldlines` arguments for plotting fieldlines over |B| contours
- Reduce default number of contours


desc/plotting.py
```diff
@@ -2283,23 +2288,28 @@ def plot_boozer_surface(
     Examples
     --------
     .. image:: ../../_static/images/plotting/plot_boozer_surface.png
-
     .. code-block:: python
-
```
f0uriest, line 2288: I think removing these lines is causing the docs build to fail.

dpanici (approved):

desc/plotting.py
```diff
@@ -2234,8 +2226,10 @@ def plot_boozer_surface(
     eq,
     grid_compute=None,
     grid_plot=None,
+    rho=1,
```
f0uriest, line 2229: Since we're adding this, does it make sense to move the `grid_compute` and `grid_plot` to be keyword only args? And possibly add similar positional args for ntheta/nzeta? Several of the other plotting functions are similar.

PR #551, opened 2023-06-23
Enforce periodicity convention for alpha and theta_sfl

### LINKED ISSUE #548: Fix periodicity of ``theta_sfl`` and ``alpha``
Both should be periodic on [0,2pi] but this is not currently enforced in the code, leading to issues when trying to map between coordinates.

f0uriest:
Resolves #548

### DIFF
--- a/desc/compute/_core.py
+++ b/desc/compute/_core.py
@@ -326,7 +326,7 @@ def _theta_sfl_z(params, transforms, profiles, data, **kwargs):
     label="\\alpha",
     units="~",
     units_long="None",
-    description="Field line label, defined on [0, 2pi]",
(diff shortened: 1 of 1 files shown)


desc/compute/_core.py
```diff
@@ -326,7 +326,7 @@ def _theta_sfl_z(params, transforms, profiles, data, **kwargs):
     label="\\alpha",
     units="~",
     units_long="None",
-    description="Field line label",
+    description="Field line label, defined on [0, 2pi]",
```
daniel-dudt, line 329: technically it is defined on `[0, 2pi)`

daniel-dudt (changes requested):

ddudt (approved):

PR #552, opened 2023-06-23
Chebyshev polynomials

daniel-dudt:
Adds Basis functions and classes for shifted Chebyshev polynomials. 
- `ChebyshevPolynomial` can be used for radial Profiles. 
- `ChebyshevDoubleFourierSeries` is used for the omnigenity mapping parameters. 

The Chebyshev polynomials are computed from the trigonometric definition $T_n(x) = \cos(n \arccos(x))$ where $x = 2\rho-1$. This makes the evaluations fast for JIT compilation, but is messy to differentiate. Derivatives of the polynomials are not implemented yet, since they are not needed for the omnigenity calculations. To add this functionality in the future, it may be better to compute the Chebyshev polynomials through the recurrence relations.


desc/basis.py
```diff
@@ -1304,6 +1555,30 @@ def powers(rho, l, dr=0):
     return polyval_vec(coeffs, rho).T
 
 
+@jit
+def chebyshev(r, l, dr=0):
```
f0uriest, line 1559: If derivatives aren't implemented we should add an assert to make sure `dr==0`

tests/test_basis.py
```diff
@@ -126,6 +129,16 @@ def test_powers(self):
...
         np.testing.assert_allclose(derivs, correct_ders, atol=1e-8)
 
+    @pytest.mark.unit
+    def test_chebyshev(self):
+        """Test chebyshev function for Chebyshev polynomial evaluation."""
+        l = np.array([0, 1, 2])
+        r = np.linspace(0, 1, 11)  # rho coordinates
+
+        correct_vals = np.array([np.ones_like(r), 2 * r - 1, 8 * r**2 - 8 * r + 1]).T
+        values = chebyshev(r[:, np.newaxis], l, dr=0)
```
f0uriest, line 139: Could use the evaluate method of the class here to hit that with testing as well

desc/basis.py
```diff
@@ -311,6 +316,98 @@ def change_resolution(self, L):
             self._set_up()
 
 
+class ChebyshevPolynomial(Basis):
```
f0uriest, line 319: Not sure what the usefulness of this is if we can't take radial derivatives of the profile?

desc/basis.py
```diff
@@ -1304,6 +1555,30 @@ def powers(rho, l, dr=0):
...
+
+    Returns
+    -------
+    y : ndarray, shape(N,K)
+        basis function(s) evaluated at specified points
+
+    """
+    x = 2 * r - 1  # shift
+    x, l, dr = map(jnp.asarray, (x, l, dr))
+    return jnp.cos(l * jnp.arccos(x))
```
f0uriest, line 1579: Are there any possible cases where AD would try to differentiate this wrt `r`? IE, doesn't the omnigenity stuff require evaluating stuff at new nodes?

desc/basis.py
ddudt: It's not useful for anything at the moment, but could be in the future. Chebyshev profiles would be a nice basis if we can add derivatives.

desc/basis.py
ddudt: I don't think so. In the omnigenity computation, the Chebyshev basis is only used to interpolate the "well shape" on the flux surface of interest. So the transform matrix that calls this function is pre-computed. The Boozer coordinate basis for $|B|$ is the transform matrix with changing nodes that has to be computed on the fly with AD, but that is a double Fourier series.

desc/basis.py
f0uriest: OK, gotcha. There is a PR to get chebyshev stuff into jax natively: https://github.com/google/jax/pull/11093

I'd vote to remove this class for now, since it isn't used currently and likely won't be until there's some refactoring to get derivatives to work.

desc/basis.py
```diff
@@ -769,17 +871,157 @@ def change_resolution(self, L, M):
...
             self._L = L
             self._M = M
+            self._sym = sym if sym is not None else self.sym
             self._modes = self._get_modes(
                 self.L, self.M, spectral_indexing=self.spectral_indexing
             )
             self._set_up()
 
 
+class ChebyshevDoubleFourierBasis(Basis):
```
f0uriest, line 892: Should add this class to ``docs/api.rst``

desc/basis.py
```diff
@@ -1304,6 +1555,31 @@ def powers(rho, l, dr=0):
...
+def chebyshev(r, l, dr=0):
+    """Shifted Chebyshev polynomial.
+
+    Parameters
+    ----------
+    rho : ndarray, shape(N,)
+        radial coordinates to evaluate basis
+    l : ndarray of int, shape(K,)
+        radial mode number(s)
+    dr : int
```
f0uriest, line 1568: possibly hard-code dr=1,2,3 derivatives?

desc/basis.py
```diff
@@ -311,6 +316,98 @@ def change_resolution(self, L):
...
+        Maximum radial resolution.
+
+    """
+
+    def __init__(self, L):
+        self._L = L
+        self._M = 0
+        self._N = 0
+        self._NFP = 1
+        self._sym = False
```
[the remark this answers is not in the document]
ddudt: class removed, so irrelevant now

dpanici (changes requested):
Could you add a test that checks the evaluation of the 3D basis against FourierZernike or something? that seems to be the last test missing for it

desc/basis.py
```diff
@@ -63,6 +65,9 @@ def _enforce_symmetry(self):
         elif self.sym == "even":  # even powers of rho
             non_sym_idx = np.where(self.modes[:, 0] % 2 != 0)
             self._modes = np.delete(self.modes, non_sym_idx, axis=0)
+        elif self.sym == "cos(t)":  # cos(m*t) terms only
+            non_sym_idx = np.where(sign(self.modes[:, 1]) < 0)
+            self._modes = np.delete(self.modes, non_sym_idx, axis=0)
```
unalmis, line 70: Don't need to change but for future you may like this more:
```python
self._modes = self.modes[self.modes[:, 0] % 2 == 0]
```

dpanici (approved):

f0uriest (approved):

PR #555, opened 2023-06-23
allow stellarator symmetry to change

daniel-dudt:
Adds a `sym` argument in `Equilibrium.change_resolution()` to specify stellarator symmetry and change all of the underlying bases accordingly.


tests/test_equilibrium.py
```diff
@@ -181,6 +181,32 @@ def test_eq_change_grid_resolution():
     assert eq.N_grid == 10
 
 
+@pytest.mark.unit
+def test_eq_change_symmetry():
```
f0uriest, line 185: for this it might be better to test if non-symmetric modes are or aren't in the basis, to make sure we didn't just change the label without modifying stuff.

desc/geometry/surface.py
f0uriest, line 1: We should also add the ability to change symmetry to `ZernikeRZToroidalSection` and `FourierRZToroidalCurve` (and possibly other stuff that have a symmetry attribute that I'm forgetting. `MagneticFields`?)

ddudt: Added for `FourierRZCurve` and `ZernikeRZToroidalSection`. That covers all the geometry classes with the symmetry attribute. `MagneticField` does not have symmetry, but that might be worth adding.

ddudt:
FYI the geometry classes always have non-stellarator symmetric grids. I assume we want to keep this and did not allow it to be changed according to the bases.

f0uriest (approved):

PR #443, opened 2023-03-01
Logging overhaul3

MothVine:
Adds DESC-specific logger that is automatically set up to write to a logfile and output to console as a normal print statement would, adds some helper functions for adding other logging handlers to DESC logger.  It also generates a file called "desc.log".

Logging can now happen at several levels, and loggers will display all logs of a given level and higher.  These are DEBUG, INFO, WARNING, ERROR, and CRITICAL.

Currently we use DEBUG for inner-loop timing and iteration information, and INFO for most standard logs.  Warnings from the 'warnings' module will automatically be logged as WARNING level logs, though raising errors will not automatically generate ERROR logs.

To use logger, you will need the following lines before you invoke it:
`import logging`
`logger = logging.getLogger("DESC_logger")`

Writing a log statement can be done by replacing a print statement with an invocation to the logger, and / or one of its log levels:
`logger.debug("Lowest level message")`
`logger.info("Standard DESC info level message")`
`logger.warning("Message that will log with the same severity as a 'warnings.warn()' statement")`
`logger.error("High level message")`
`logger.critical("Highest level log message)`

To allow a function to change the verbosity level of the DESC logger, you will want initiate the DESC logger as usual with
`import logging`
`logger = logging.getLogger("DESC_logger")`

Then in your function, include the following logic:
`if verbose == 0:`
`    logger = set_console_logging(console_log_level="CRITICAL")`
`if verbose == 1:`
`    logger = set_console_logging(console_log_level="INFO")`
`if verbose == 2:`
`    logger = set_console_logging(console_log_level="DEBUG")`

This removes the need for constant verbosity checks; the logger knows what message level the user has set verbosity to, so you can use 'logger.debug("Msg") and 'logger.info("Msg") calls whenever and the output will sort itself out.

Currently a logfile log has the following format:
`2023-02-28 16:56:31.603 :: DESC_logger :: INFO ::  File: objective_funs  Func: build  Line: 189  ::  Building objective: force `

A console log has the following format- but is probably too long, and should be changed before merging- it could be we want no formatting, so it will display like a normal print statement but still be logged.
`INFO     DESC_logger:equilibrium.py:523 Start of solver`

Log levels will display in a color-coded fashion in the console.  Once we know exactly how much information we want to display where, I can finish the unit tests for the newly added functions.

This has changed how tests that compared stdout to expected output strings work, so now a logger is made to capture output to compare to the desired string.  If we decide not to format console logs at all, the old method can be reinstated.

PR #562, opened 2023-06-29
Avoid / catch warnings during testing, fail CI on new warnings

### LINKED ISSUE #534: Resolve warnings during unit and regression tests
During our automated unit and regression tests a lot of random warnings get thrown, some from our code, some from libraries we depend on. Currently we basically just ignore these, but probably shouldn't. 

Pytest allows us to catch warnings stuff like 
```python
with pytest.warns(UserWarning):
    function_that_throws_warning()
```

Ideally we'd refactor code to avoid warnings wherever possible, catch the ones we can't, and then set pytest to raise warnings to errors so that we're alerted if new warnings creep in.

f0uriest:
- This refactors some code to avoid warnings during common use cases and unit/regression tests
- Tests that raise expected warnings are now caught using `pytest.warns`
- Sets default `pytest` behavior to promote uncaught warnings to errors

Resolves #534


dpanici (dismissed):

desc/grid.py
```diff
@@ -203,7 +203,10 @@ def _scale_weights(self):
         temp_spacing = self.spacing.copy()
         temp_spacing = (temp_spacing.T / duplicates ** (1 / 3)).T
         # scale weights sum to full volume
-        temp_spacing *= (4 * np.pi**2 / temp_spacing.prod(axis=1).sum()) ** (1 / 3)
+        if temp_spacing.prod(axis=1).sum():
```
unalmis, line 206: When would this ever be zero?

f0uriest: It happens if the grid is empty for some reason, which occurs in some of our tests:
https://github.com/PlasmaControl/DESC/actions/runs/5583151692/jobs/10203162768#step:6:191

tests/test_optimizer.py
```diff
@@ -561,25 +562,29 @@ def test_wrappers():
 @pytest.mark.slow
 def test_all_optimizers():
     """Just tests that the optimizers run without error, eg tests for the wrappers."""
-    eq = desc.examples.get("SOLOVEV")
-    fobj = ObjectiveFunction(ForceBalance(eq=eq))
-    eobj = ObjectiveFunction(Energy(eq=eq))
+    eqf = desc.examples.get("SOLOVEV")
+    eqe = eqf.copy()
+    eqe._node_pattern = "quad"
```
unalmis, line 567: Seems like this is because node pattern was not an ioable attribute at the time SOLOVEV was solved and saved. Might be good to add a todo marker to remove this when solved in case solovev node pattern is changed in future.

tests/test_optimizer.py
```diff
@@ -463,6 +463,7 @@ def test_scipy_fail_message():
             gtol=1e-12,
         )
         assert "Maximum number of iterations has been exceeded" in result["message"]
+    eq._node_pattern = "quad"
```
unalmis, line 466: Same as below.

unalmis (approved):

tests/test_optimizer.py
```diff
@@ -657,7 +662,8 @@ def test_scipy_constrained_solve():
 def test_solve_with_x_scale():
     """Make sure we can manually specify x_scale when solving/optimizing."""
     # basically just tests that it runs without error
-    eq = Equilibrium(L=2, M=2, N=2, pressure=np.array([1000, -2000, 1000]))
+    with pytest.warns(UserWarning, match="pressure profile is not an even"):
+        eq = Equilibrium(L=2, M=2, N=2, pressure=np.array([1000, -2000, 1000]))
```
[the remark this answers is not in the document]
dpanici: there are some use cases for linear profiles, like in some of Matt Landreman's bootstrap current consistency tests where the profiles are odd

desc/vmec.py
```diff
@@ -162,7 +162,11 @@ def load(
         file.close()
 
         # initialize Equilibrium
-        eq = Equilibrium(**inputs)
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore", message="Left handed coordinates detected"
+            )
+            eq = Equilibrium(**inputs)
```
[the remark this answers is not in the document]
f0uriest: VMEC always uses left handed coordinates, and if the user is loading stuff from a saved output there's not really an easy way to manually change the coordinate system, so not much point in warning them.

Based on the idea [here](https://docs.python.org/3/howto/logging.html) that `warnings.warn` should be about something the user can fix, while `logging.warning` is for stuff they can't fix but should still be noted (but we don't use logging ATM so punting that for now.)

tests/test_stability_funs.py
```diff
@@ -264,7 +264,7 @@ def test_mercier_print(capsys):
 def test_magwell_print(capsys):
     """Test that the magnetic well stability criteria prints correctly."""
     eq = desc.examples.get("HELIOTRON")
-    grid = LinearGrid(L=12, M=12, N=6, axis=False)
+    grid = LinearGrid(L=12, M=12, N=6, NFP=eq.NFP, axis=False)
```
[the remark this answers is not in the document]
f0uriest: We don't need to, a nonsymmetric grid is basically a superset of a symmetric grid. But if NFP is wrong you're effectively using lower toroidal resolution.

tests/test_optimizer.py
f0uriest: No, it is saved, but the example uses a `jacobi` pattern with force balance as the objective. This creates another version that uses `quad` to avoid the warning when creating the energy objective.

dpanici (approved):

desc/vmec.py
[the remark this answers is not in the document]
dpanici: #443

PR #563, opened 2023-06-29
Dp/loss

dpanici:
- Add ability to pass in a custom loss function to any objective, such as `jnp.mean` to take the average as the target or `jnp.max` to take the max as the target

### DIFF
--- a/desc/objectives/_bootstrap.py
+++ b/desc/objectives/_bootstrap.py
@@ -73,6 +73,7 @@ class BootstrapRedlConsistency(_Objective):
         weight=1,
         normalize=True,
         normalize_target=True,
+        loss_function=lambda x: x,
         grid=None,
         helicity=(1, 0),
         name="Bootstrap current self-consistency (Redl)",
--- a/desc/objectives/_generic.py
+++ b/desc/objectives/_generic.py
@@ -90,6 +90,7 @@ class ObjectiveFromUser(_Objective):
         weight=1,
         normalize=True,
         normalize_target=True,
+        loss_function=lambda x: x,
         grid=None,
         name="custom",
     ):
@@ -104,6 +105,7 @@ class ObjectiveFromUser(_Objective):
             weight=weight,
             normalize=normalize,
             normalize_target=normalize_target,
+            loss_function=loss_function,
             name=name,
         )
 
@@ -222,6 +224,7 @@ class GenericObjective(_Objective):
         weight=1,
         normalize=True,
         normalize_target=True,
+        loss_function=lambda x: x,
         grid=None,
         name="generic",
     ):
@@ -236,6 +239,7 @@ class GenericObjective(_Objective):
             weight=weight,
             normalize=normalize,
             normalize_target=normalize_target,
+            loss_function=loss_function,
             name=name,
         )
         self._units = "(" + data_index[self.f]["units"] + ")"
@@ -338,6 +342,7 @@ class ToroidalCurrent(_Objective):
         weight=1,
         normalize=True,
         normalize_target=True,
+        loss_function=lambda x: x,
         grid=None,
         name="toroidal current",
     ):
@@ -351,6 +356,7 @@ class ToroidalCurrent(_Objective):
             weight=weight,
             normalize=normalize,
             normalize_target=normalize_target,
+            loss_function=loss_function,
             name=name,
         )
 
@@ -515,6 +521,7 @@ class RotationalTransform(_Objective):
         weight=1,
         normalize=True,
         normalize_target=True,
+        loss_function=lambda x: x,
         grid=None,
         name="rotational transform",
     ):
@@ -528,6 +535,7 @@ class RotationalTransform(_Objective):
             weight=weight,
             normalize=normalize,
             normalize_target=normalize_target,
+            loss_function=loss_function,
             name=name,
         )
 
--- a/desc/objectives/_qs.py
+++ b/desc/objectives/_qs.py
@@ -64,6 +64,7 @@ class QuasisymmetryBoozer(_Objective):
         weight=1,
         normalize=True,
         normalize_target=True,
+        loss_function=lambda x: x,
         grid=None,
         helicity=(1, 0),
         M_booz=None,
@@ -85,6 +86,7 @@ class QuasisymmetryBoozer(_Objective):
             weight=weight,
             normalize=normalize,
             normalize_target=normalize_target,
+            loss_function=lambda x: x,
             name=name,
         )
 
@@ -258,6 +260,7 @@ class QuasisymmetryTwoTerm(_Objective):
         weight=1,
         normalize=True,
         normalize_target=True,
+        loss_function=lambda x: x,
         grid=None,
         helicity=(1, 0),
         name="QS two-term",
@@ -273,6 +276,7 @@ class QuasisymmetryTwoTerm(_Objective):
             weight=weight,
             normalize=normalize,
             normalize_target=normalize_target,
+            loss_function=lambda x: x,
             name=name,
         )
 
@@ -425,6 +429,7 @@ class QuasisymmetryTripleProduct(_Objective):
         weight=1,
         normalize=True,
         normalize_target=True,
+        loss_function=lambda x: x,
         grid=None,
         name="QS triple product",
     ):
@@ -438,6 +443,7 @@ class QuasisymmetryTripleProduct(_Objective):
             weight=weight,
             normalize=normalize,
             normalize_target=normalize_target,
+            loss_function=lambda x: x,
             name=name,
         )
 
@@ -565,6 +571,7 @@ class Isodynamicity(_Objective):
         weight=1,
         normalize=False,
         normalize_target=False,
+        loss_function=lambda x: x,
         grid=None,
         name="Isodynamicity",
     ):
@@ -578,6 +585,7 @@ class Isodynamicity(_Objective):
             weight=weight,
             normalize=normalize,
             normalize_target=normalize_target,
+            loss_function=lambda x: x,
             name=name,
         )
 
--- a/desc/objectives/linear_objectives.py
+++ b/desc/objectives/linear_objectives.py
@@ -58,6 +58,7 @@ class BoundaryRSelfConsistency(_Objective):
             weight=1,
             normalize=False,
             normalize_target=False,
+            loss_function=lambda x: x,
             name=name,
         )
 
@@ -140,6 +141,7 @@ class BoundaryZSelfConsistency(_Objective):
             weight=1,
             normalize=False,
             normalize_target=False,
+            loss_function=lambda x: x,
             name=name,
         )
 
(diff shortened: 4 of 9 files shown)


desc/objectives/_bootstrap.py
```diff
@@ -49,6 +49,10 @@ class BootstrapRedlConsistency(_Objective):
         Whether target and bounds should be normalized before comparing to computed
         values. If `normalize` is `True` and the target is in physical units,
         this should also be set to True.
+    loss_function : function, optional
```
f0uriest, line 52: This should be `callable` not `function`

desc/objectives/_generic.py
```diff
@@ -201,6 +207,14 @@ class GenericObjective(_Objective):
         Whether target and bounds should be normalized before comparing to computed
         values. If `normalize` is `True` and the target is in physical units,
         this should also be set to True.
+    loss_function : function, optional
+        User-defined loss function to apply to the objective values once computed.
+        Must be a JAX transformable function, e.g. `jnp.mean` for taking the average
+        or `lambda x: 3*x`, etc
+    loss_function : function, optional
```
f0uriest, line 214: duplicated

desc/objectives/_generic.py
```diff
@@ -318,6 +334,14 @@ class ToroidalCurrent(_Objective):
...
```
f0uriest, line 341: duplicated

desc/objectives/_generic.py
```diff
@@ -495,7 +521,11 @@ class RotationalTransform(_Objective):
         values. If `normalize` is `True` and the target is in physical units,
         this should also be set to True.
         Note: has no effect for this objective.
-    grid : Grid, optional
+    loss_function : function, optional
+        User-defined loss function to apply to the objective values once computed.
+        Must be a JAX transformable function, e.g. `jnp.mean` for taking the average
+        or `lambda x: 3*x`, etc
+    grid : Grid, ndarray, optional
```
f0uriest, line 528: does this accept ndarrays? I thought all the objectives only took `Grid` objects?

desc/objectives/_qs.py
```diff
@@ -37,6 +37,13 @@ class QuasisymmetryBoozer(_Objective):
         Whether target and bounds should be normalized before comparing to computed
         values. If `normalize` is `True` and the target is in physical units,
         this should also be set to True.
+    loss_function : function, optional
+        User-defined loss function to apply to the objective values once computed.
+        Must be a JAX transformable function, i.e. jnp.mean for taking the average
+    loss_function : function, optional
```
f0uriest, line 43: duplicated

desc/objectives/objective_funs.py
```diff
@@ -742,6 +747,15 @@ def __init__(
         assert normalize_target in {True, False}
         assert (bounds is None) or (isinstance(bounds, tuple) and len(bounds) == 2)
         assert (bounds is None) or (target is None), "Cannot use both bounds and target"
+        assert callable(loss_function), "Loss function must be callable!"
+        assert (
```
f0uriest, line 751: this might be too restrictive, eg
``len(getfullargspec(jnp.max).args)`` is 6, but all the other args have default values. Besides, the next line will check if its valid when called with a single argument.

desc/objectives/objective_funs.py
```diff
@@ -742,6 +747,15 @@ def __init__(
...
+            len(getfullargspec(loss_function).args) == 1
+        ), "Loss function must accept a single array as an argument!"
+        _test_output_loss = loss_function(jnp.array([1, 1]))
```
f0uriest, line 754: might be better to pass an array of size `dim_f` here and then update the `dim_f` attribute afterwards with the correct output size.

f0uriest (changes requested):
Left a few comments to be addressed

One other thing is we should make it clear in the docs somewhere (possibly in the docstring) when the loss function is applied, ie before/after weighting and bounds etc.

desc/objectives/objective_funs.py
```diff
@@ -902,16 +919,18 @@ def compute(self, *args, **kwargs):
 
     def compute_unscaled(self, *args, **kwargs):
         """Compute the raw value of the objective."""
-        return jnp.atleast_1d(self.compute(*args, **kwargs))
+        return jnp.atleast_1d(
+            self._loss_function(jnp.atleast_1d(self.compute(*args, **kwargs)))
```
ddudt, line 923: Does this need the extra `jnp.atleast_1d` call here? `computed_scaled` and `compute_scaled_error` don't have this.

ddudt (commented):
This is cool. It would be nice to add something similar for the plotting functions too (not in this PR).

desc/objectives/_generic.py
dpanici: it does not, some of the docstrings still had ndarray but I think they all only take Grid so I've updated them accordingly

f0uriest:
Should also update ``docs/adding_objectives.rst`` with the new stuff.

Also, will everything work correctly if the loss function results in the output being different from ``dim_f``?

desc/objectives/linear_objectives.py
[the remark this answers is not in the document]
dpanici: no I had made some changes and I guess my black did not remove those blank lines, so I had to do it myself

dpanici:
> Should also update `docs/adding_objectives.rst` with the new stuff.
> 
> Also, will everything work correctly if the loss function results in the output being different from `dim_f`?

Not yet, I have not tested it enough.

desc/objectives/objective_funs.py
dpanici: I put it in case loss function returned something that was not a 1D ndarray (like jnp.mean() which returns a 0D array)

dpanici:
Need to think of how to change weights etc, if jnp.mean makes the dimension 1, but weights is `dim_f` based off grid. If fxn is linear/affine, should be ok if we apply after the weights, but other functions (softmin etc) it would matter

think of how broadcasting will work if loss_function changes the dimension

dpanici:
- [ ] add option to say where to apply loss
- [ ] change dim_f according to the dim of the loss fnx

desc/objectives/_bootstrap.py
```diff
@@ -49,7 +49,20 @@ class BootstrapRedlConsistency(_Objective):
...
         values. If `normalize` is `True` and the target is in physical units,
         this should also be set to True.
-    grid : Grid, ndarray, optional
+    loss_function : callable, optional
+        User-defined loss function to apply to the objective values once computed.
+        Must be a JAX transformable function, e.g. `jnp.mean` for taking the average
+        or `lambda x: 3*x`, etc.
+        This loss function is called on the raw compute value, before any shifting,
+        scaling, or normalization.
+    where_apply_loss " str,
```
f0uriest, line 58: `"` should be `:`

also in a bunch of other places

f0uriest: also don't need comma after `str`

desc/objectives/_bootstrap.py
```diff
@@ -49,7 +49,20 @@ class BootstrapRedlConsistency(_Objective):
...
+        where to apply the user defined loss function. One of "compute_unscaled"
```
f0uriest, line 59: i'd vote to rename these options to just `scaled` and `unscaled`, feels a bit clearer

desc/objectives/_generic.py
```diff
@@ -52,7 +52,27 @@ class ObjectiveFromUser(_Objective):
...
+        This loss function is called on the raw compute value, before any shifting,
+        scaling, or normalization.
+    where_apply_loss " str,
+        where to apply the user defined loss function. One of "compute_unscaled"
+        or "compute_scaled"
+        If "compute_unscaled", the loss function will apply to
+        the raw objective function values, before the target is applied.
+        If "compute_scaled", the loss function will apply to
+        the scaled objective function values, after the target is applied.
+    where_apply_loss " str,
```
f0uriest, line 68: duplicated text

desc/objectives/_generic.py
```diff
@@ -215,7 +239,27 @@ class GenericObjective(_Objective):
...
```
f0uriest, line 255: duplicated text

desc/objectives/_generic.py
```diff
@@ -350,7 +398,27 @@ class ToroidalCurrent(_Objective):
...
```
f0uriest, line 414: duplicated text

desc/objectives/objective_funs.py
```diff
@@ -760,6 +760,20 @@ class _Objective(IOAble, ABC):
...
+        This loss function is called on the raw compute value, before any shifting,
+        scaling, or normalization.
+    where_apply_loss " str,
+        where to apply the user defined loss function. One of "compute_unscaled"
+        or "compute_scaled"
+        If "compute_unscaled", the loss function will apply to
+        the raw objective function values, before the target is applied.
+        If "compute_scaled", the loss function will apply to
+        the scaled objective function values, after the target is applied.
+
```
f0uriest, line 776: extra blank line?

desc/objectives/objective_funs.py
```diff
@@ -786,13 +800,44 @@ def __init__(
...
         assert np.all(np.asarray(weight) > 0)
         assert normalize in {True, False}
         assert normalize_target in {True, False}
         assert (bounds is None) or (isinstance(bounds, tuple) and len(bounds) == 2)
         assert (bounds is None) or (target is None), "Cannot use both bounds and target"
+        assert callable(loss_function), "Loss function must be callable!"
+        # function should be able to handle input of this type and shape
+        try:
+            _test_output_loss = loss_function(jnp.ones((2,)))
```
f0uriest, line 829: any reason the user couldn't create a loss function that specialized on the shape of the expected input? Would this break that?

desc/objectives/objective_funs.py
```diff
@@ -885,6 +933,7 @@ def jit(self):
         except AttributeError:
             pass
         self.compute_scaled = jit(self.compute_scaled)
+        self.compute_scaled2 = self.compute_scaled
```
f0uriest, line 936: is this used?

desc/objectives/objective_funs.py
dpanici: No, this was just to test that the function accepts arrays. There is probably a better way by having it accept dim_f size array and ensuring no error is raised

desc/objectives/objective_funs.py
dpanici: no it was a debug thing, thanks for catching

dpanici:
@f0uriest @ddudt I am unsure how to deal with if someone want to say do mean iota on the unscaled values. That would change the dim_f of the compute_unscaled to be just 1 instead of `grid.num_rho`. However, `compute_scaled` will be still size `num_rho` because we multiply by a  extra weighting in the `_scale` method of `RotationalTransform`. This would cause issues because the dim_f would be different for these two methods, but that means the scaled and unscaled jacobians are different. 

Does it even make sense to apply weighting after a reduction operation like mean is used on the unscaled values?

dpanici:
- Clear up quadrature vs objective weights #484 
 - possibly apply loss to quadrature weights
 - specify that is only correct for mean/min/max
   - str arg
   - if scalar, dont use quadrature weights in compute scaled

dpanici:
Just have 3 predefined, then one called 'callable' and it will be basically a custom compute scaled, and have it accept f, weights, etc, then have user decide what they want to do (where to apply loss etc)

f0uriest (approved):

PR #565, opened 2023-06-30
Ensure that basis.L , basis.M, basis.N all return integers

dpanici:
- Fixes bug where if a `Basis` object were instantiated with a float passed in as one of the resolutions, such as `PowerSeries(L=1.0)`, then `basis.L` would be a float. This caused issues during the init of most bases, except DoubleFourierSeries, which could cause further headaches when down the line another object using that basis runs into issues because it expected an int and got a float instead for `basis.N`

### DIFF
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -147,17 +147,17 @@ class Basis(IOAble, ABC):
     @property
     def L(self):
         """int: Maximum radial resolution."""
-        return self.__dict__.setdefault("_L", 0)
+        return round(self.__dict__.setdefault("_L", 0))
 
     @property
     def M(self):
         """int:  Maximum poloidal resolution."""
-        return self.__dict__.setdefault("_M", 0)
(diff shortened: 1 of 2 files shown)


desc/basis.py
```diff
@@ -149,16 +149,31 @@ def L(self):
         """int: Maximum radial resolution."""
         return self.__dict__.setdefault("_L", 0)
 
+    @L.setter
+    def L(self, L):
+        assert isinstance(L, int), "Basis Resolution must be an integer!"
```
f0uriest, line 154: This might be overly restrictive, eg `np.int` wouldn't work here.
A better way might be
```python
assert int(L) == L
self._L = int(L)

f0uriest: You may still want to cast it to an int explicitly if it just being a float can cause issues

Ie, ``int(3.0) == 3`` is true

f0uriest (approved):

PR #566, opened 2023-07-04
Fix norm bug in curvature computation

unalmis:
Looks like the matrix norm was being computed instead of the vector norm for normal vector of a rho surface. Affects all curvature quantities. Might be a good idea to review other places we use norm. `grep -r -F 'np.linalg.norm'`

### DIFF
--- a/desc/compute/_geometry.py
+++ b/desc/compute/_geometry.py
@@ -244,7 +244,7 @@ def _a_major_over_a_minor(params, transforms, profiles, data, **kwargs):
     transforms={},
     profiles=[],
     coordinates="rtz",
-    data=["n", "e_theta_t"],
+    data=["n_rho", "e_theta_t"],
 )
 def _L_sff(params, transforms, profiles, data, **kwargs):
     data["L_sff"] = dot(data["e_theta_t"], data["n_rho"])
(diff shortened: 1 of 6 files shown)


desc/compute/_geometry.py
```diff
@@ -233,24 +233,6 @@ def _a_major_over_a_minor(params, transforms, profiles, data, **kwargs):
...
-    description="Unit normal vector to constant rho surface",
-    dim=3,
-    params=[],
-    transforms={},
-    profiles=[],
-    coordinates="rtz",
-    data=["e^rho"],
-)
-def _n_rho(params, transforms, profiles, data, **kwargs):
-    data["n_rho"] = (data["e^rho"].T / jnp.linalg.norm(data["e^rho"])).T
```
unalmis, line 250: I think this should instead be `jnp.linalg.norm(data["e^rho"]), axis=1).T`.

dpanici: I am a little confused, so is `n` wrong or just the `n_rho` quantity?

unalmis: it looks like they are supposed to be the same thing? but n was being computed correctly while n_rho wasn't, so i replaced n_rho with n

desc/plotting.py
```diff
@@ -2798,10 +2798,7 @@ def plot_basis(basis, return_data=False, **kwargs):
         fig, ax = plt.subplots(figsize=kwargs.get("figsize", (6, 4)))
 
         f = basis.evaluate(grid.nodes)
-        plot_data = {}
-        plot_data["n"] = basis.modes[:, 2]
-        plot_data["amplitude"] = []
-        plot_data["zeta"] = z
+        plot_data = {"n": basis.modes[:, 2], "amplitude": [], "zeta": z}
```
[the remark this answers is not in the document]
f0uriest: no, this `n` is the mode number

f0uriest (approved):

PR #567, opened 2023-07-05
convert RZCurve to XYZCurve

daniel-dudt:
Adds the method `FourierRZCurve.to_xyz(N=None)` to convert to a `FourierXYZCurve`. 

Example use case: converting the magnetic axis to same curve type as coils for plotting, etc. 

Converting the other way (from XYZ to RZ) may not always be possible and is still not implemented.


f0uriest (changes requested):
This seems fine, only request is I think a better convention for the `to_*` methods is to use the full name of the class (eg `to_FourierXYZCurve`) to avoid ambiguity (eg with Dario's xyz spline curve). I've been meaning to update the profile classes with that but might as well start here.

f0uriest:
Also converting to an RZ curve should be possible for arbitrary geometries when the generalized toroidal angle is finished.

f0uriest (approved):

dpanici (approved):

PR #568, opened 2023-07-05
Rc/toroidal angle

### LINKED ISSUE #465: Generalize toroidal angle beyond phi cylindrical
This has come up several times now, and would be very useful for dealing with strongly shaped axes, racetrack/figure8 type shapes etc.

The main things to do:

- [ ] Add a toroidal stream function like lambda (what symmetry does it have? what resolution is needed?)
- [ ] Change computation of contravariant B to use generalized toroidal angle
- [ ] Change places where we assume dphi/dzeta = R
- [ ] Add option to *not* use toroidal stream function for backwards compatibility
- [ ] Test to see if it helps spectral convergence
- [ ] Add utility for finding condensed spectrum when fitting from QSC/QIC

f0uriest:
- [ ] near axis constraints with omega
- [ ] constant offset surface
- [ ] Proximal projection
- [ ] Add stuff for spectral condensation / optimal fitting
- [x] Default to fixing zeta==phi? Default to an empty W_lmn, modes_W.
- [ ] to_FourierRZCurve methods should now be possible?
- [ ] Fix calculation of A, a_major/a_minor, other stuff that needs to be computed at const phi, not const zeta. (this should already be done in other prs)
- [ ] changing continuation/perturbation to work with the new coordinates (changing regularization)
- [ ] make the new API compatible with this PR
- [ ] different resolutions for omega?
- [ ] Revert change to plot_surfaces map_coordinates call once toroidal angle is generalized #1952 

Resolves #465


unalmis:
Are you sure that DSHAPE should have non positive curvature? tests/test_objective_funs.py::test_mean_curvature is failing

dpanici:
`test_compute_everything` try also with omega nonzero or something

dpanici:
just load and resave all the eqs after merge, dont resolve as there is not consistnecy to set omega to zero yet on this branch

unalmis:
Note to self: in bounce integrals you used d zeta / B dot grad zeta for field line differential length where the zeta in dzeta denotes the field line zeta, and the zeta in B dot grad zeta denotes DESC's coordinate zeta. this is fine currently, but work through the math rigorously when desc zeta is generalized with stream function. also I need to recheck my legendre transform for computing B_z|r,a if it takes into account generalized zeta.

unalmis:
alpha_t should be corrected to
data["alpha_t"] = data["theta_PEST_t"] - data["iota"] * data["phi_t"]

unalmis:
Also the formula for alpha uses zeta while it's derivatives use phi.

dpanici:
@unalmis I merged master, gonna see which tests fail but I think there needs to be some extra logic added to `"A"` and `"A(z)"` to prevent `ZernikeRZToroidalSection` objects from getting an error when they try to compute `"A"`, or at least there is a test failing due to that. They have no `W_lmn` attribute but somewhere we try to access it when computing with those objects

unalmis:
> @unalmis I merged master, gonna see which tests fail but I think there needs to be some extra logic added to `"A"` and `"A(z)"` to prevent `ZernikeRZToroidalSection` objects from getting an error when they try to compute `"A"`, or at least there is a test failing due to that. They have no `W_lmn` attribute but somewhere we try to access it when computing with those objects

Note they aren't computed correctly here yet because you need to use source grids to do coordinate mapping. I think the error you see is due to #1127 . Also `A(z) and `A` compute functions will change in #1094 .

dpanici:
couple omega to theta pest somehow?

rahulgaur104:
> Note to self:
> 
> * [x]  in bounce integrals you used d zeta / B dot grad zeta for field line differential length where the zeta in dzeta denotes the field line zeta, and the zeta in B dot grad zeta denotes DESC's coordinate zeta. this is fine currently, but work through the math when desc zeta is generalized with stream function. nothing changes with bounce/field line stuff.
> * [x]  and check for any 2d divergence theorems in zeta surface. Edit: The integration over boundaries we do (2d divergence theorems etc.) no longer work. I have done the math to fix them. Once this gets up to date with master, feel free to ping me and I'll push the up

@unalmis  This is up-to-date with master. If you can share your notes here, I can code them up.

unalmis:
I handle all the stoke's theorem stuff with nonzero toroidal stream map in #1094

unalmis:
and I updated the boozer transforms  for this already in master

f0uriest:
We should probably add some sort of new version of the `test_compute_everything` where we compare 2 equilibria with the same physical geometry, one with zeta=phi and one without, mapping the coordinates so we compare things at the same points in real space.

rahulgaur104:
> and I updated the boozer transforms for this already in master, but remember to remove this line:
> 
> https://github.com/PlasmaControl/DESC/blob/dd6ce311a128e09af43725cd85e532dd17f4157b/desc/compute/_omnigenity.py#L68

Ok, done! 

Another error, I don't know if the "A" calculation is still causing errors, but when I run 

pytest -k test_get_zeta_surface

I get 

```python
tests/test_configuration.py F                                                                                                                                                                                                                                                                                                                                                        [100%]

========================================================================================================================================================================================= FAILURES =========================================================================================================================================================================================
__________________________________________________________________________________________________________________________________________________________________________ TestGetSurfaces.test_get_zeta_surface ___________________________________________________________________________________________________________________________________________________________________________

self = <tests.test_configuration.TestGetSurfaces object at 0x7f4ea5a804d0>

    @pytest.mark.unit
    def test_get_zeta_surface(self):
        """Test getting a constant zeta surface."""
        eq = Equilibrium()
        surf = eq.get_surface_at(zeta=np.pi)
        assert surf.zeta == np.pi
        rho = 1
>       np.testing.assert_allclose(surf.compute("A")["A"], np.pi * rho**2)

tests/test_configuration.py:474: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
desc/geometry/core.py:494: in compute
    params = get_params(names, obj=self, basis=kwargs.get("basis", "rpz"))
desc/backend.py:133: in wrapper
    return func(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

keys = ['A'], obj = ZernikeRZToroidalSection at 0x7f4ea2094c50 (name=), has_axis = False, basis = 'rpz'

    @execute_on_cpu
    def get_params(keys, obj, has_axis=False, basis="rpz"):
        """Get parameters needed to compute a given quantity.
    
        Parameters
        ----------
        keys : str or array-like of str
            Name of the desired quantity from the data index
        obj : Equilibrium, Curve, Surface, Coil, etc.
            Object to compute quantity for.
        has_axis : bool
            Whether the grid to compute on has a node on the magnetic axis.
        basis : {"rpz", "xyz"}
            Basis of computed quantities.
    
        Returns
        -------
        params : list[str] or dict[str, jnp.ndarray]
            Parameters needed to compute key.
            If eq is None, returns a list of the names of params needed
            otherwise, returns a dict of ndarray with keys for R_lmn, Z_lmn, etc.
    
        """
        p = _parse_parameterization(obj)
        keys = [keys] if isinstance(keys, str) else keys
        deps = list(keys) + get_data_deps(keys, p, has_axis=has_axis, basis=basis)
        params = []
        for key in deps:
            params += data_index[p][key]["dependencies"]["params"]
        if isinstance(obj, str) or inspect.isclass(obj):
            return params
        temp_params = {}
        for name in params:
>           p = getattr(obj, name)
E           AttributeError: 'ZernikeRZToroidalSection' object has no attribute 'W_lmn'

desc/compute/utils.py:507: AttributeError

```

rahulgaur104:
> I handle all the stoke's theorem stuff with nonzero toroidal stream map in #1094

@unalmis  Is it a good idea to merge your branch into this one or make your branch the base branch?

dpanici:
> > I handle all the stoke's theorem stuff with nonzero toroidal stream map in #1094
> 
> @unalmis Is it a good idea to merge your branch into this one or make your branch the base branch?

I think that #1094 should be fixed/merged first, not made a base branch for here. #1094 does not even have tests passing, and has issues related to #1127 that have to be resolved

desc/compute/_core.py
```diff
@@ -1648,10 +1648,10 @@ def _alpha_tz(params, transforms, profiles, data, **kwargs):
     transforms={},
     profiles=[],
     coordinates="rtz",
-    data=["theta_PEST_z", "phi_z", "iota"],
+    data=["theta_PEST_z", "zeta_z", "iota"],
```
dpanici, line 1651: I think this is wrong, we need this to be `"phi_z"` because now `zeta=phi` is not assumed, but `alpha` as defined within the PEST coordinate system uses the toroidal angle `phi`

rahulgaur104: Yeah, all the phis and their derivatives in this PR are actually what we call zetas so you are right.

unalmis: Yes we had already changed these from zeta to phi in other PRs.

desc/compute/data_index.py
```diff
@@ -58,7 +58,7 @@ def register_compute_fun(  # noqa: C901
     transforms,
     profiles,
     coordinates,
-    data,
+    data=[],
```
dpanici, line 61: Idk why I made this change but we prob dont need this

desc/equilibrium/equilibrium.py
```diff
@@ -177,6 +182,9 @@ def __init__(
         L_grid=None,
         M_grid=None,
         N_grid=None,
+        Lz=0,
```
dpanici, line 185: what is Lz Mz Nz? if for the omega resolution using something like L_W M_W N_W may be clearer

dpanici: And add any new stuff to the docstring too

desc/equilibrium/equilibrium.py
```diff
@@ -266,8 +283,28 @@ def __init__(
...
+        self._Nz = int(setdefault(Nz, self.surface._Nz))
+        self._Mz = int(setdefault(Mz, self.surface._Mz))
+        self._Lz = int(
+            setdefault(
+                Lz,
+                self.Mz if (self.spectral_indexing == "ansi") else 2 * self.Mz,
+            )
+        )
+
+        # RG: NFP, sym, Mz, Nz passed as kwargs, order is sacrosanct
```
dpanici, line 295: just use kwargs for everything if we are adding new arguments here, if that makes it clearer. I am in favor of kwargs over args always

desc/equilibrium/equilibrium.py
```diff
@@ -266,8 +283,28 @@ def __init__(
...
+            self.L,
+            self.M,
+            self.N,
+            NFP=self._surface.NFP if hasattr(self._surface, "NFP") else self.NFP,
+            sym=self.sym,
+            Mz=self.Mz,
+            Nz=self.Nz,
+        )
+
+        # RG: If you must specify NFP here, specify the NFP of the axis
```
dpanici, line 306: what does this mean?

desc/equilibrium/equilibrium.py
```diff
@@ -461,10 +519,10 @@ def __repr__(self):
...
             + " at "
             + str(hex(id(self)))
-            + " (L={}, M={}, N={}, NFP={}, sym={}, spectral_indexing={})".format(
-                self.L, self.M, self.N, self.NFP, self.sym, self.spectral_indexing
-            )
-        )
+            + f" (L={self.L}, M={self.M}, N={self.N}, Lz={self.Lz}, Mz={self.Mz},"
+            + "Nz={self.Nz}, NFP={self.NFP}, sym={self.sym},"
+            + "spectral_indexing={self.spectral_indexing})"
+        )  # RG: not sure if this change is needed
```
dpanici, line 525: we should track the omega resolution so I think yes it is not needed strictly but we want this to be changed

desc/equilibrium/equilibrium.py
rahulgaur104: Well, there is test which checks if axis.NFP is different from surface.NFP and eq.NFP so we must not specify eq.NFP here otherwise that test will fail.

desc/equilibrium/equilibrium.py
rahulgaur104: Using LW, NW, MW for now.

desc/coils.py
```diff
@@ -688,13 +688,15 @@ class FourierRZCoil(_Coil, FourierRZCurve):
...
-        fourier coefficients for R, Z
+        current through coil, in Amperes
+    R_n, Z_n, W_n: array-like
+        fourier coefficients for R, Z, omega
     modes_R : array-like
         mode numbers associated with R_n. If not given defaults to [-n:n]
     modes_Z : array-like
         mode numbers associated with Z_n, defaults to modes_R
+    modes_W : array-like
+        mode numbers associated with W_n, defaults to modes_W
```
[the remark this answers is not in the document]
rahulgaur104: I meant modes_W will only have [0, 0, 0] mode. I am starting to realize that modes_W = [0, 0,0] and W_lmn = np.array([0]) is a better option than empty modes_W and W_lmn arrays.

desc/equilibrium/equilibrium.py
rahulgaur104: The change of variables will be the very last step because I'd have to do this to all the files in examples. tests/inputs/ etc.
I will add the docstrings though

desc/compute/data_index.py
rahulgaur104: Actually if I remove =[], I get the following error
  File "/home/rgaur/toroidal_angle_scripts/rerun_w_Omega_cold.py", line 3, in <module>
  File "/home/rgaur/DESC/desc/equilibrium/__init__.py", line 3, in <module>
  File "/home/rgaur/DESC/desc/equilibrium/equilibrium.py", line 16, in <module>
    from desc.compute import compute as compute_fun
  File "/home/rgaur/DESC/desc/compute/__init__.py", line 29, in <module>
    from . import (
  File "/home/rgaur/DESC/desc/compute/_core.py", line 2102, in <module>
    @register_compute_fun(
     ^^^^^^^^^^^^^^^^^^^^^
TypeError: register_compute_fun() missing 1 required positional argument: 'data'

rahulgaur104: So reverting back for now.

YigitElma: We can remove bunch of lines by defaulting `transforms={}`, `profiles=[]` `coordinates="rtz"` etc

YigitElma: Probably doing this (if we ever do) in a new PR would be better, since this PR has already a lot of changes

f0uriest: using mutable data structures like lists or dicts as default arguments leads to all sorts of weird bugs that are hard to find, I strongly recommend against it.

rahulgaur104: Okay, so leaving as it is. Marking as resolved.

rahulgaur104:
Add detailed description of the progress and the current state.

dpanici:
- [ ] try optimization where you fix all Wb_lmn
- [ ] try optimization where you fix all but a single Wb_lmn mode

rahulgaur104:
[advanced_optimization_generalized_toroidal.ipynb](https://github.com/user-attachments/files/21903949/advanced_optimization_generalized_toroidal.ipynb)

PR #570, opened 2023-07-06
Add toroial resolution argument to get_NAE_constraints

dpanici:
- Allows the `get_NAE_constraints` function to accept a toroidal resolution argument `N`. If this is less than `desc_eq.N`, than the equilibrium's NAE behavior only up to that passed-in toroidal resolution will be constrained. 
- Use-case is mainly for debugging order rho^2, but is a small change that can exist outside of that PR


desc/objectives/nae_utils.py
```diff
@@ -43,6 +46,12 @@ def _calc_1st_order_NAE_coeffs(qsc, desc_eq):
     R0 = qsc.R0_func(phi)
     dR0_dphi = qsc.R0p
     dZ0_dphi = qsc.Z0p
+    if N is None:
+        N = desc_eq.N
+    else:
+        N = np.max([desc_eq.N, N])
```
ddudt, line 52: shouldn't this be `min`?

dpanici: yes, thanks for catching that

f0uriest (approved):

PR #571, opened 2023-07-06
Build docs tables dynamically

f0uriest:
- Rather than having to manually recreate the list of variables every time we change something, this now automatically runs the scripts to generate ``variables.csv`` and ``optimizers.csv`` whenever the docs are built, either locally with ``make html`` or on RTD
- Also cleans up some misformatted or overly long TeX labels


desc/compute/_field.py
```diff
@@ -17,7 +17,7 @@
 
 @register_compute_fun(
     name="B0",
-    label="\\partial_{\\rho} \\psi / \\sqrt{g}",
+    label="\\psi' / \\sqrt{g}",
```
ddudt, line 20: I don't like this change. It feels messy to mix notation (sometimes using `'` for radial derivatives, other times using $\partial_\rho$), and the existing notation is the most explicit. I realize it seems excessive for the radial profiles, but it is clear and avoids any ambiguity.

f0uriest: I agree its not perfect, but I think its still clear what we mean, and `B0` isn't really something the user ever sees directly. The issue with the old way was that some of the derivatives ended up being super long which would mess with the formatting in the list of variables.

One other option might be to remove stuff like `B0` and `iota_zero_current_num` from the public list of variables?

unalmis (approved):

desc/compute/_field.py
dpanici: I'm ok with this change, I don't think psi will ever be a function of theta and zeta, we still have the profiles being explicit partial derivatives

PR #572, opened 2023-07-06
Generalize compute utils and data index for multiple parameterizations

f0uriest:
- One of many steps in generalizing DESC to work with multiple optimizable object types.
- Adds the ability to have multiple "parameterizations" in the `data_index` (eg, different classes, types)
- Right now the only parameterization is `Equilibrium` but in the future will also include curves, coils, surfaces etc.
- Main difference is that when calling `desc.compute.compute` or any of the `get_params/profiles/transforms` etc, you need to specify the parameterization, either as a string or an object of the given type.

Still not convinced this is the best/most elegant way to do this, and open to other suggestions.


desc/compute/data_index.py
```diff
@@ -75,7 +80,63 @@ def _decorator(func):
...
+# from parent classes.
+# This is the least bad solution I've found, since everything else requires
+# crazy circular imports
+# could maybe make this fancier with a registry of compute-able objects?
+_aliases = {
+    "desc.equilibrium.equilibrium.Equilibrium": [
+        "desc.equilibrium.Equilibrium",
+        "Equilibrium",
+    ],
+    "desc.geometry.curve.FourierRZCurve": [
```
dpanici, line 110: Unsure if I understand the aliases, so if I have a compute function that is specific to just FourierRZCurve, but not to XYZcurve (from the other PR I have #494 ), for example it is compute_coordinates, which for FourierRZCurve uses transforms, while XYZCurve uses splines. 
Is the idea that if I put as a paremetrization for my compute function decorator `XYZCurve`, then it will be able to be used by `XYZCurve` objects but not `FourierRZCurve` objects?
But if I say the parametrization for a compute function is `Curve`, then it will be allowed to be used with any subclass of `Curve` (according to our aliases here)?

desc/compute/utils.py
```diff
@@ -41,11 +42,21 @@
...
+
 def _sort_args(args):
     return [arg for arg in arg_order if arg in args]
 
 
-def compute(names, params, transforms, profiles, data=None, **kwargs):
+def compute(parameterization, names, params, transforms, profiles, data=None, **kwargs):
     """Compute the quantity given by name on grid.
 
     Parameters
```
dpanici, line 62: docstring here is missing `parameterization`

desc/compute/utils.py
```diff
@@ -69,27 +80,31 @@ def compute(names, params, transforms, profiles, data=None, **kwargs):
         Computed quantity and intermediate variables.
 
     """
+    p = _parse_parameterization(parameterization)
     if isinstance(names, str):
         names = [names]
     for name in names:
-        if name not in data_index:
+        if name not in data_index[p]:
             raise ValueError("Unrecognized value '{}'.".format(name))
```
dpanici, line 88: should we add to this error `"For parameterization {p}"` ?
Since in the future, if I understand correctly, we are thinking of allowing things to have like `FourierRZSurface.compute("n_rho")`, so if someone instead did `FourierRZSurface.compute("|F|")`, the error should specify that "|F|" is an unrecognized value specifically for the given object ( `FourierRZSurface`)that is being asked to compute it

desc/compute/data_index.py
f0uriest: Yeah exactly. This is basically just to keep track of class inheritance without actually having to import all the classes to do regular `issubclass` and `isinstance` checks. In #583 you can see that there are multiple implementations for curve `x` (for each parameterization) but stuff like the frenet frame, curvature etc there is one shared function that works for any `Curve`

f0uriest: I'll clean it up a bit to be just dealing with inheritance rather than also parsing different paths to the same class, we should just stick to the canonical path.

desc/compute/data_index.py
```diff
@@ -75,7 +80,55 @@ def _decorator(func):
...
+    "desc.geometry.curve.FourierPlanarCurve": [
+        "desc.geometry.core.Curve",
+    ],
+    "desc.geometry.surface.FourierRZToroidalSurface": [
+        "desc.geometry.core.Surface",
+    ],
+    "desc.geometry.surface.ZernikeRZToroidalSection": [
+        "desc.geometry.core.Surface",
+    ],
+    "desc.coils.FourierRZCoil": [
```
dpanici, line 120: do we also want `desc.coils.Coil` for these? For methods that would belong to Coil but not Curve like computing biot-savart

dpanici: Or we can leave for now and I can add in the coils PR #494

f0uriest: Yeah currently there aren't any compute methods for the generic coil class, but if we later add some we'd need to add that here.

docs/index.rst
```diff
@@ -3,6 +3,8 @@
 .. include:: ../README.rst
   :start-after: inclusion-marker-do-not-remove
 
+.. Note: variables.rst is autogenerated by write_variables.py
```
dpanici, line 6: just a note for the future, when we add other parametrizations it might be best to have multiple tabs under List of Variables, one for each parametrization. Either that or have another column where it lists which objects are able to calculate each variable, otherwise the list page will get really long

dpanici (approved):
I really cannot think of something more elegant, this seems to work well enough now and I can see how it will be expanded when we add other parametrizations, so this is good I think

desc/compute/data_index.py
```diff
@@ -75,7 +80,55 @@ def _decorator(func):
...
     return _decorator
+
+
+# This allows us to handle subclasses whos data_index stuff should inherit
+# from parent classes.
+# This is the least bad solution I've found, since everything else requires
+# crazy circular imports
+# could maybe make this fancier with a registry of compute-able objects?
+_class_inheritance = {
+    "desc.equilibrium.equilibrium.Equilibrium": [],
```
unalmis, line 104: Wouldn't it be more natural to have the keys of this dictionary be the superclass and the values be the subclasses? Then if the parameteriztaion is a superclass (key in this dict) then assign `d.copy()` to all values (subclasses). If the parameteriztaion is a subclass then just assign `d.copy()` to that subclass.

desc/compute/utils.py
```diff
@@ -41,15 +42,27 @@
...
+        return klass.__qualname__  # avoid outputs like 'builtins.str'
+    return module + "." + klass.__qualname__
+
+
 def _sort_args(args):
     return [arg for arg in arg_order if arg in args]
 
 
-def compute(names, params, transforms, profiles, data=None, **kwargs):
+def compute(parameterization, names, params, transforms, profiles, data=None, **kwargs):
```
unalmis, line 59: might be more consistent to have this go after names

unalmis (approved):

PR #573, opened 2023-07-07
Change resolution grid attribute

unalmis:
When the ``change_resolution`` method is called on any grid

- The ``NFP`` attribute could be changed in isolation. I think this is a bug because the node placement depends on ``NFP``, so I would expect that the grid nodes should be updated whenever NFP does.
- The ``num_rho``, ``num_theta``,  ``num_zeta``,  ``unique_xxx_idx``, ``inverse_xxx_idx`` attributes were not updated when the nodes were changed.

This pr
- fixes the above issues
- refactors where some attributes are assigned to make it easier to keep track of all the attributes
- copies the ``compress`` and ``expand`` functions from ``compute_utils`` to the grid class. I made this change in the ``add_all_limits`` PR for consistency and have just copied ``grid.py`` from there. The ``add_all_limits`` PR removes these functions from ``compute.utils``.


desc/grid.py
```diff
@@ -77,52 +113,52 @@ def _enforce_symmetry(self):
...
-                self.nodes[to_delete_idx, 0], return_counts=True
-            )
-            assert (
-                2 * np.pi not in self.nodes[:, 1]
-                and off_sym_line_per_rho_surf_count.size
-                >= to_delete_per_rho_surf_count.size
+        if not self.sym:
+            return
+        # indices where theta coordinate is off the symmetry line of theta=0 or pi
+        off_sym_line_idx = self.nodes[:, 1] % np.pi != 0
```
unalmis, line 119: nothing changed in this block i just switched it to a boolean mask instead of integer because its less code

desc/grid.py
```diff
@@ -662,23 +773,24 @@ def change_resolution(self, L, M, N, NFP=None):
             Number of field periods.
 
         """
-        self._NFP = NFP if NFP is not None else self.NFP
-        if L != self.L or M != self.M or N != self.N:
```
unalmis, line 666: Previously NFP could be changed without updating the Grid. But I don't think this makes sense because the node placement depends on NFP

desc/grid.py
```diff
@@ -662,23 +773,24 @@ def change_resolution(self, L, M, N, NFP=None):
...
-                N=N,
-                NFP=self.NFP,
-                axis=len(self.axis) > 0,
-                endpoint=self.endpoint,
+                L=L, M=M, N=N, NFP=NFP, axis=self.axis.size > 0, endpoint=self.endpoint
             )
             # symmetry handled in create_nodes()
             self._sort_nodes()
-            self._find_axis()
-            self._scale_weights()
```
unalmis, line 681: Previously, when the grid changed resolution, none of the unique and inverse idx would be updated.

desc/grid.py
```diff
@@ -48,24 +49,59 @@ class Grid(IOAble):
...
+            self._inverse_theta_idx,
+            self._unique_zeta_idx,
+            self._inverse_zeta_idx,
+        ) = self._find_unique_inverse_nodes()
+        self._L = self.num_rho
+        self._M = self.num_theta
+        self._N = self.num_zeta
+        self._weights = self._scale_weights()
+
+    def _create_nodes(self, nodes):
```
unalmis, line 79: I just changed the order of the functions here to match the order they are called.

desc/grid.py
```diff
@@ -48,24 +49,59 @@ class Grid(IOAble):
...
+        nodes = np.atleast_2d(nodes).reshape((-1, 3)).astype(float)
+        # Do not alter nodes given by the user for custom grids.
+        # In particular, do not modulo nodes by 2pi or 2pi/NFP.
+        # This may cause the surface_integrals() function to fail recognizing
+        # surfaces outside the interval [0, 2pi] as duplicates. However, most
+        # surface integral computations are done with LinearGrid anyway.
+        spacing = (  # make weights sum to 4pi^2
+            np.ones_like(nodes) * np.array([1, 2 * np.pi, 2 * np.pi]) / nodes.shape[0]
+        )
+        return nodes, spacing
```
unalmis, line 104: Since git struggled with this diff, I just removed the code
```python
self._L = len(np.unique(nodes[:, 0]))
self._M = len(np.unique(nodes[:, 1]))
self._N = len(np.unique(nodes[:, 2]))
```
and assigned them in ``__init__``.

desc/grid.py
```diff
@@ -746,9 +861,13 @@ def _create_nodes(self, L=1, M=1, N=1, NFP=1):
...
-        L = self.L + 1
-        M = 2 * self.M + 1
-        N = 2 * self.N + 1
+        self._L = L
+        self._M = M
+        self._N = N
+        self._NFP = NFP
+        L = L + 1
+        M = 2 * M + 1
+        N = 2 * N + 1
```
unalmis, line 870: Also assigning them here to make the oversampling clear.

desc/grid.py
```diff
@@ -48,24 +49,59 @@ class Grid(IOAble):
...
+            self._inverse_rho_idx,
+            self._unique_theta_idx,
+            self._inverse_theta_idx,
+            self._unique_zeta_idx,
+            self._inverse_zeta_idx,
+        ) = self._find_unique_inverse_nodes()
+        self._L = self.num_rho
+        self._M = self.num_theta
+        self._N = self.num_zeta
+        self._weights = self._scale_weights()
```
f0uriest, line 77: Can you explain the motivation for making these methods return values rather than just assigning them?

unalmis: The excerpt from this book explains the memory optimization more clearly than the pep412 docs. https://stackoverflow.com/a/73751450. Although my motivation for the change was just to make it consistent with how the `create_nodes` method returns values, so that it's clearer what the attributes are without having to read the whole file. If it looks ugly I can change back

f0uriest (approved):

tests/test_grid.py
ddudt, line 1: Should we add test to check that changing `NFP` properly updates the nodes?

unalmis: added

desc/grid.py
```diff
@@ -331,6 +316,33 @@ def node_pattern(self):
         """str: Pattern for placement of nodes in (rho,theta,zeta)."""
         return self.__dict__.setdefault("_node_pattern", "custom")
 
+    @property
+    def nodes(self):
+        """ndarray: Node coordinates, in (rho,theta,zeta)."""
+        return self.__dict__.setdefault("_nodes", np.array([]).reshape((0, 3)))
+
+    @nodes.setter
```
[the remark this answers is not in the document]
unalmis: 1. remove the setter methods so that they are not publicly writable
2. set the read only attribute on the numpy arrays to make them immutable
3. recast to jax array

Is your suggestion to do just number 1? Or if you want to to 1&2, then why not do option 3 to avoid the need for this: https://github.com/PlasmaControl/DESC/blob/6fad1a8158b4ed3aed3f75e6dc2d495041f62a06/desc/compute/utils.py#L1209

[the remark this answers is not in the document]
f0uriest: I was just thinking about #1

dpanici (approved):

unalmis:
@dpanici can you merge this at your convenience? others have approved previously. (The codecov fails but no coverage has actually decreased).

PR #581, opened 2023-07-15
Dp/interpolate

dpanici:
- No longer precomputes Df in SplineProfile, as it doesn't really help with speed.
- Fixes bug where approx_df returns None when method is not a cubic spline family
- refactor monotonic spline algorithm to work with 2-D inputs
- add tests for these fixes
- Test all available interpolation methods for interp2d and interp3d

### DIFF
--- a/desc/profiles.py
+++ b/desc/profiles.py
@@ -10,7 +10,7 @@ from desc.backend import jit, jnp, put, sign
 from desc.basis import FourierZernikeBasis, PowerSeries
 from desc.derivatives import Derivative
 from desc.grid import Grid, LinearGrid
-from desc.interpolate import _approx_df, interp1d
+from desc.interpolate import interp1d
 from desc.io import IOAble
 from desc.transform import Transform
 from desc.utils import combination_permutation, copy_coeffs, multinomial_coefficients
@@ -812,9 +812,6 @@ class SplineProfile(Profile):
         self._knots = knots
         self._params = values
         self._method = method
-        self._Dx = _approx_df(
-            self._knots, np.eye(self._knots.size), self._method, axis=0
-        )
 
     def __repr__(self):
         """Get the string form of the object."""
@@ -885,12 +882,8 @@ class SplineProfile(Profile):
             return jnp.zeros_like(xq)
         x = self._knots
         f = params
-        if self._method == "monotonic":
-            df = None  # must create the df at compute time since the BCs
-            #  contain logic depending on f
-        else:
-            df = self._Dx @ f
-        fq = interp1d(xq, x, f, method=self._method, derivative=dr, extrap=True, fx=df)
+
+        fq = interp1d(xq, x, f, method=self._method, derivative=dr, extrap=True)
         return fq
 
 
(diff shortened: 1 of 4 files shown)


dpanici:
@ddudt @f0uriest I can't exactly figure out why the SplineProfile isn't working well for the `monotonic` method. Ive narrowed it down to basically the edge case BC function I think, something about that function makes it be very inaccurate for the first few points of the interpolation

dpanici:
For now I just added a conditional that makes the df not be precomputed, and it works well as is now.

f0uriest:
I think you can't precompute df for the monotonic splines since the monotonicity preserving thing makes it nonlinear

desc/profiles.py
```diff
@@ -885,8 +885,12 @@ def compute(self, params=None, grid=None, dr=0, dt=0, dz=0):
             return jnp.zeros_like(xq)
         x = self._knots
         f = params
-        df = self._Dx @ f
-        fq = interp1d(xq, x, f, method=self._method, derivative=dr, extrap=True, df=df)
+        if self._method == "monotonic":
```
f0uriest, line 888: I think it might actually be faster in many cases to just not precompute `df`:

```python
import numpy as np
from jax import jit
from desc.interpolate import interp1d, _approx_df
from desc.grid import ConcentricGrid
p = PowerSeriesProfile([1,0,-2,0,1])
s = p.to_spline()
grid = ConcentricGrid(24,24,24)
s.grid = grid
p.grid = grid

Dx = _approx_df(
    s._knots, np.eye(s._knots.size), s._method, axis=0
    )

def interp(f):
    return interp1d(
        grid.nodes[:,0], 
        s._knots, 
        f, 
        method=s._method, 
        derivative=0, 
        extrap=True,
       )

def dinterp(f):
    fx = Dx @ f
    return interp1d(
        grid.nodes[:,0], 
        s._knots, 
        f, 
        method=s._method, 
        derivative=0,
        extrap=True, 
        fx=fx
    )

%timeit _ = jit(interp)(s.params).block_until_ready()
%timeit _ = jit(dinterp)(s.params).block_until_ready()
```

without precompute:
```1.03 ms ± 270 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)```
with precompute:
```1.24 ms ± 296 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)```

Computing df with the matrix multiply takes `O(n^2)` operations (since its basically doing it for any possible values of `f`), but in theory calculating it on the fly only requires `O(n)` operations

dpanici: hm I should have profiled that... would have saved me a lot of time this morning lol. Thanks, I will change that, I will keep the changes to the monotonic spline stuff though if it could be useful for the higher dimensional interpolation (I should add it as a method for those too)

dpanici:
I am not adding `monotonic` `monotonic-`0 to interp2d or interp3d as generalizing it to multi dimensions is not directly obvious to me at this point and this is good enough to merge in as is I think. If someone else wants to tackle that that is fine but I think it is fine to not have it in this PR

f0uriest (approved):
Looks good. In general monotonic functions in multiple dimensions are tricky, and we don't really have any use for them so I don't think we need monotonic interpolation in 2d or 3d.

ddudt (approved):

PR #556, opened 2023-06-26
Magnetic axis limits and new variables

unalmis:
[Math to compute the rotational transform in DESC.pdf](https://github.com/PlasmaControl/DESC/files/12137234/rotational_transform_and_limit.pdf)

- [x]  Resolves #486
- [x] Update `desc.plotting.plot_fsa` to be able to plot at the axis 
- [x] Remove `desc.compute.compress` and `desc.compute.expand` functions in favor `grid.compress` and `grid.expand` object methods
- [x] Miscellaneous formatting changes caught when doing above tasks
- [x] Update test durations to prevent GitHub CI timeout failure

If the limits were unable to be computed with the version of l’Hôpital’s below, then I probably added some comments to outline whatever alternative technique I used. Please note that this version of  l’Hôpital’s rule I have used can only resolve the  $\mathbf{0}/0$ indeterminate form not the (something / $\infty$) form.

# An $ℝ^n → ℝ^m$ l’Hôpital’s rule for $\mathbf{0}/0$
Part of my work for this PR included proving non-existence of the limit of some of the more complicated functions. (Recall that for a function $a/b$, if $da/db$ is proven to not exist, we still know nothing about the limit of $a/b$). The functions whose limits do not exist are listed at the top of `tests/test_axis_limits.py`. If you don't want to confirm this non-existence with math then a heuristic alternative is just plotting (script below) and seeing if they blow up. For the limits that do exist, I used the rule below. I didn't think it's widely known that you can use a l’Hôpital’s rule for our functions, so I wrote this explanation.

l’Hôpital’s rule relies on Cauchy’s generalized mean value theorem, which is a property proven unique to functions that map $ℝ → ℝ$ (see [paper 1](https://www.researchgate.net/publication/254212006_On_the_Vector_Form_of_the_Lagrange_Formula_the_Darboux_Property_and_LHopitals_Rule)). The quantities we compute map $ℝ^3 → ℝ^m$, so it would be very wrong to evaluate limits by comparing the end behavior of their derivatives. Our polar coordinates do allow us to write the limit we desire as $\lim_{\rho \to 0} f(\rho, \theta, \zeta)$, but this does not let us reduce our limits to single variable limits. See [here](https://math.stackexchange.com/a/1769377) for an already written explanation reminding this.

[...]

PR #586, opened 2023-07-17
Add 4th order derivative transforms

unalmis:
- [x] Adds Zernike radial polynomial fourth order derivative. This is needed for #556 .
- [x] Ignore division by zero in compute funs.
- [x] Update compute/_core with dummy omega function


desc/transform.py
```diff
@@ -85,13 +85,14 @@ def __init__(
 
     def _set_up(self):
 
-        self.method = self._method
+        self.method = self._method  # assign according to logic in setter function
+        n = np.max(self.derivatives) + 1
```
unalmis, line 89: lower bound to 1

unalmis: Making this change passed the failing test. Should I make both changes. `n = np.maximum(np.amax(self.derivatives), 1) + 1`

f0uriest: probably doesn't hurt, but I think `derivatives` should have a minimum value of 0

unalmis: Ok I'll revert that then

desc/transform.py
```diff
@@ -71,29 +71,19 @@ def __init__(
...
-
         self._built = False
         self._built_pinv = False
-        self._set_up()
+        self._derivatives = self._get_derivatives(derivs)
+        self._sort_derivatives()
+        # assign according to logic in setter function
+        self.method = method
+        # assign according to logic in property function
+        self._matrices = self.matrices
```
unalmis, line 81: Mixing the design patterns of using setters and property functions to assign attributes is a little confusing. If this is done elsewhere would prefer if comments like this are added to clarify.

unalmis: Not sure what the problem is. This is the error.
> tests/test_perturbations.py::test_perturbation_orders - AttributeError: 'Transform' object has no attribute '_matrices'. Did you mean: 'matrices'?
I've added the following guard to the end of `__init__`. It does not raise an error, so this attribute must be getting deleted somewhere?
```python
        self._matrices = self.matrices
        if not hasattr(self, "_matrices"):
            raise ValueError("Kaya does not know how to python. Uno")
        if build:
            self.build()
        if build_pinv:
            self.build_pinv()

        if not hasattr(self, "_matrices"):
            raise ValueError("Kaya does not know how to python. Dos")
```

f0uriest: not totally sure what the issue is, but if you remove the line 
```
        self._matrices = self.matrices
``` 
from init and also replace all the references to ``self._matrices`` elsewhere in the class with ``self.matrices`` it seems to work for me

unalmis: Is the def method setter supposed to be recursive? If you notice it calls itself because it assigns with self.method = rather than self._method. "Fixing" this to make it not recursive generates similar missing attribute errors. It's not clear to me why either of these issues occur from making these changes, but maybe they are related.

I'm hesitant to make your change. Currently it seems the matrices attribute is getting deleted after __init__, and I think your suggestion will avoid that by constantly reconstructing it whenever its called.

f0uriest:
The method setter is recursive, not sure what's causing the problem

desc/transform.py
unalmis: It seems like it's because I removed the `def _set_up` method, which apparently is used in `desc/io/equilibrium_io.py` for something. If that does end up being the issue, we should document the purpose more clearly. I think it is unusual to rely on a private method like `_set_up`.

f0uriest (approved):

desc/basis.py
```diff
@@ -1252,37 +1254,49 @@ def zernike_radial(r, l, m, dr=0):
...
+        f = _jacobi(n, alpha, beta, jacobi_arg, 0)
+        df = _jacobi(n, alpha, beta, jacobi_arg, 1)
+        d2f = _jacobi(n, alpha, beta, jacobi_arg, 2)
+        d3f = _jacobi(n, alpha, beta, jacobi_arg, 3)
         out = (
             (m - 2) * (m - 1) * m * r ** jnp.maximum(m - 3, 0) * f
-            - 12 * (m - 1) * m * r ** jnp.maximum(m - 1, 0) * df
-            + 48 * r ** (m + 1) * d2f
+            - 12 * m**2 * r ** jnp.maximum(m - 1, 0) * df
+            + 48 * (m + 1) * r ** (m + 1) * d2f
```
dpanici, line 1281: can you explain this refactor? It does not seem completely obvious to me how it is the same expression, it seems to me like you removed a `12*m*r**(jnp.maximum(m-1,0) * df` and added a `48*m*r**(m+1)*df`

dpanici (changes requested):
could you add a correctness test for the jacobi derivatives? like some crude check against finite differences or something, just to make sure things are correct.

desc/basis.py
f0uriest: It's correct, I checked it last night. It's just re-grouping terms. We do test the correctness of the jacobi derivatives vs evaluating the zernike polynomials from their polynomial coeffs, and the derivatives.

unalmis: we subtract an extra 12m times the trailing term from the last line. Add an extra 48m times the trailing term from the last line

dpanici: ah my bad the diff was a bit hard to see, I did not see the bottom two lines removed from the original code

f0uriest (dismissed):

f0uriest:
I can't seem to see why the unit tests are failing, nothing that changed should affect them

f0uriest (approved):

dpanici (approved):

tests/test_basis.py
```diff
@@ -142,76 +142,72 @@ def test_chebyshev(self):
             chebyshev(r[:, np.newaxis], l, dr=1)
 
     @pytest.mark.unit
-    def test_zernike_radial(self):
+    def test_zernike_radial(self):  # noqa: C901
         """Test zernike_radial function, comparing to analytic formulas."""
-
-        def test_zernike_radial(self):
```
unalmis: so i found the coverage problem..

unalmis:
Made requested change to merge some changes from #556

desc/compute/_core.py
```diff
@@ -67,7 +67,7 @@ def _rho_r(params, transforms, profiles, data, **kwargs):
     units="~",
     units_long="None",
     description="Radial coordinate, proportional to the square root "
-    + "of the toroidal flux, derivative wrt poloidal coordinate",
+    "of the toroidal flux, derivative wrt poloidal coordinate",
```
unalmis, line 70: If I remember correctly, the cause of this lost `+` sign was a script I used to update a mistake I made while copying pasting descriptions. Shouldn't affect anything since the string is concatenated.

desc/compute/_core.py
f0uriest, line 1: Is there any particular order to the entries in this file? I think at one point they were sorted alphabetically by derivative order and I'd like to keep that if possible to make code generation and maintenance easier.

unalmis: They have now been sorted using the script I made here: https://github.com/PlasmaControl/DESC/blob/basis_vectors/docs/sort_compute_funs.py

unalmis: It looks like sorting them alphabetically doubled the diff size. If you want a custom sort order with lower case first that can be done via the one-line modification to the script as discussed here https://stackoverflow.com/a/28136499.

f0uriest: Thanks! The actual order isn't so important as long as there is some logic to it so its reproducible and we can keep it in order.

f0uriest (approved):

setup.cfg
```diff
@@ -47,6 +47,8 @@ markers=
 filterwarnings=
     error
     ignore::pytest.PytestUnraisableExceptionWarning
+    ignore::RuntimeWarning:desc.compute
```
dpanici, line 50: do we want this change? I thought that Rory wanted to catch warnings like this in the tests, then change the underlying code to avoid it, like in this line: https://github.com/PlasmaControl/DESC/blob/a92c2f9f9822dc843faa017ca042555d86149e48/desc/grid.py#L206

f0uriest: I think it's fine. With the axis stuff there are too many individual places where we divide by zero and then overwrite with the correct value, so ignoring the warning for just the compute module seems reasonable.

tests/test_basis.py
unalmis: Yes I must have copied the function definition in a later commit after I tested. Then black formatting finished the job of tricking me

f0uriest (approved):

f0uriest (approved):

dpanici (approved):

PR #588, opened 2023-07-18
Ability to get an input file from a DESC output file

rahulgaur104:
I think having a routine that can generate a basic DESC input file from a DESC output file would be helpful. 
I'll make further commits to this PR later this week

Update 07/25/23: I have added a function named write_descout_to_input to DESC/desc/io/input_reader.py that can create a simple input file. However, because of limited information in a DESC output file, I do not know how the continuation progresses, i.e., how we go from low-res, zero beta tokamaks to a general stellarator. This causes the optimization to fail.

Update 07/30/23: I have removed the feature that sets the continuation parameters. Now, DESC chooses the continuation parameters automatically, and the input file works. I have also added a simple unit test that checks some of the changes made to input_reader.py. However, the test does not cover a lot of the new code. Maybe we should add more tests.
.


desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+            "L_rad": "L",
+            "M_pol": "M",
+            "N_tor": "N",
+            "L_grid": "L_grid",
+            "M_grid": "M_grid",
+            "N_grid": "N_grid",
+        }.items():
+            f.write(key + " = {}\n".format(" " + str(eval("eq0.{0}".format(val)))))
+
+        print("DESC will automatically choose continuation parameters..\n")
```
dpanici, line 826: this is probably an unnecessary print statement, we can just put this in the docstring

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+
+        f.write("\n\n# solver tolerances\n")
+        try:
+            f.write("ftol = {}\n".format(eq0._bdry_ratio))
```
dpanici, line 830: in this line, _bdry_ration is not the same a ftol, so why is this being accessed?

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+        except AttributeError:
+            print("ftol not given in the output. Setting to 1E-2...")
+            f.write("ftol = {}\n".format(float(1e-2)))
+
+        try:
+            f.write("xtol = {}\n".format(eq0._pres_ratio))
```
dpanici, line 836: same here, I think we never will be including these things in the output file, so maybe it is best to just put them as the default values, but have kwargs that let you change it when you call the function if you want a different ftol, xtol etc from the default

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+            f.write("maxiter = {}\n".format(eq0._pert_order))
+        except AttributeError:
+            print("maxiter not given in the output. Setting to 100...")
+            f.write("maxiter = {}\n".format(int(100)))
+
+        f.write("\n\n# solver methods\n")
+        try:
+            f.write("optimizer = {}\n".format(eq0._optimizer))
+        except AttributeError:
+            print("Optimizer absent in the output. Setting to lsq-exact...")
```
dpanici, line 857: same for this, you can have a kwarg for optimizer if something other than default is desired

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+            char = "i"
+            if len(pres_profile1) >= len(iota_profile1):
+                pres_profile = pres_profile1
+
+                iota_profile = np.zeros((len(pres_profile),))
+                iota_profile[: len(iota_profile1)] = iota_profile1
+            else:
+                pres_profile1 = np.zeros((len(iota_profile1),))
+                pres_profile[: len(pres_profile)] = pres_profile
+        except AttributeError:
```
dpanici, line 892: probably a cleaner way to do this is to use an if statement and check `if eq.iota` for the first block, as `eq.iota= None` when it is a current constrained equilibrium

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+        f.write("\n# spectral resolution\n")
+        for key, val in {
+            "L_rad": "L",
+            "M_pol": "M",
+            "N_tor": "N",
+            "L_grid": "L_grid",
+            "M_grid": "M_grid",
+            "N_grid": "N_grid",
+        }.items():
+            f.write(key + " = {}\n".format(" " + str(eval("eq0.{0}".format(val)))))
```
f0uriest, line 824: I think you probably want `getattr` here instead of `eval`. You might also consider using (fstrings)[https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings] since they're a bit easier to read

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+        except AttributeError:
+            print("Objective not given in the output. Setting to force...")
+            f.write("objective = {}\n".format("force"))
+
+        f.write("spectral_indexing = {}\n".format(eq0._spectral_indexing))
+        f.write("node_pattern = {}\n".format(eq0._node_pattern))
+
+        f.write("\n# pressure and rotational transform/current profiles\n")
+
+        # can't import other stuff in io due to circular imports, so have to check by name
```
f0uriest, line 871: You could also do a local import (inside the function, not at the top of the file)

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+            iota_profile1 = eq0._iota.params.tolist()
+            assert (
+                eq0.iota.__class__.__name__ == "PowerSeriesProfile"
+            ), "Equilibrium must have power series profiles for ascii io"
+            char = "i"
+            if len(pres_profile1) >= len(iota_profile1):
+                pres_profile = pres_profile1
+
+                iota_profile = np.zeros((len(pres_profile),))
+                iota_profile[: len(iota_profile1)] = iota_profile1
```
f0uriest, line 888: you can also use the `change_resolution` method of the `Profile` class here to make sure all the arrays are the same size, which should work correctly regardless of the profiles even/odd parity. This might be wrong if one profile is enforced to be even but the other isn't

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+            for k, (l, m, n) in enumerate(eq0.surface.Z_basis.modes):
+                if abs(eq0.Zb_lmn[k]) > 5e-5:
+                    f.write(
+                        "l: {:3d}\tm: {:3d}\tn: {:3d}\tR1 = {:16.8E}\tZ1 = {:16.8E}\n".format(
+                            int(0), m, n, 0, eq0.Zb_lmn[k]
+                        )
+                    )
+        else:
+            for k, (l, m, n) in enumerate(eq0.surface.R_basis.modes):
+                if abs(eq0.Rb_lmn[k]) > 5e-5 or abs(eq0.Zb_lmn[k]) > 5e-5:
```
f0uriest, line 942: 5e-5 seems like a pretty high cutoff for ignoring Fourier modes, we should at least make this an argument the user can pass in. Take a look at https://github.com/PlasmaControl/DESC/blob/32398c3ad34e3f05296216126f9091730eb78126/desc/io/input_reader.py#L804

desc/io/input_reader.py
```diff
@@ -776,6 +778,185 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+                    f.write(
+                        "l: {:3d}\tm: {:3d}\tn: {:3d}\tR1 = {:16.8E}\tZ1 = {:16.8E}\n".format(
+                            int(0), m, n, eq0.Rb_lmn[k], eq0.Zb_lmn[k]
+                        )
+                    )
+
+        f.write("\n\n# magnetic axis initial guess\n")
```
f0uriest, line 949: We might actually not want to give an axis guess. If it's doing the automatic continuation, the axis position of the final equilibrium is usually a bad guess for the axis position of the initial one (and in some cases may lie outside the plasma boundary)

tests/test_input_output.py
```diff
@@ -110,6 +110,132 @@ def test_write_desc_input_Nones(tmpdir_factory):
...
+    tmp_path = tmpdir.join("input_LandremanPaul_QA")
+    tmpout_path = tmpdir.join("LandremanPaul2022_QA_reactorScale_lowRes.h5")
+    shutil.copyfile(outfile_path, tmpout_path)
+
+    ir = InputReader()
+    ir.descout_to_input(str(tmp_path), str(tmpout_path))
+    desc_input_truth = ".//tests//inputs//LandremanPaul2022_QA_reactorScale_lowRes"
+    desc_input_converted = tmp_path
+
+    with open(desc_input_truth) as f:
```
f0uriest, line 134: instead of manually parsing both these files, its probably easier (and a better test of the code) to use `InputReader.parse_inputs` to get both as a dict of ndarray and then do regular array comparisons

tests/test_input_output.py
```diff
@@ -110,6 +110,132 @@ def test_write_desc_input_Nones(tmpdir_factory):
...
+        lines_correct[counter0].split("\t")[1].split(":")[0].replace(" ", "") == "m"
+    )
+
+    found0 = 1
+
+    i = int(counter1)
+    j = int(counter0)
+
+    # Compare the Fourier modes and ensure that they match, i.e,, found0 = 1
+    if mfirst0 is True:
```
f0uriest, line 175: the more "pythonic" way would just be `if mfirst0:`, which also guards against the possibiltiy that `mfirst0` may not be the python literal `True` (remember, `is` checks if they're the same instance), but possibly some other boolean type like a np.array of bool etc.

desc/io/input_reader.py
rahulgaur104: Removed axis guess

tests/test_input_output.py
rahulgaur104: Done

desc/io/input_reader.py
rahulgaur104: I would be surprised if that is the case; Maybe if you want to lower your objective to very small values. I have set the threshold now to 1E-8.

desc/io/input_reader.py
rahulgaur104: Moved to docstring.

desc/io/input_reader.py
rahulgaur104: Corrected.

desc/io/input_reader.py
rahulgaur104: Using kwargs now.

desc/io/input_reader.py
rahulgaur104: Read from kwargs, else default values.

desc/io/input_reader.py
rahulgaur104: Done.

desc/io/input_reader.py
rahulgaur104: I understand the point but don't know how to implement it. I'll figure out my own way, and your can review it again.

desc/io/input_reader.py
rahulgaur104: Done.

rahulgaur104:
I have made all the changes suggested by you. However, the test fails due to a silly warning. The variables pres_flag and iota_flag are True even though all the coefficients are zero which causes parse_inputs to throw an error when I use it in my test.

tests/inputs/HELIOTRON_vacuum
```diff
@@ -27,7 +27,7 @@ spectral_indexing = ansi
 node_pattern      = jacobi
 
 # pressure and rotational transform/current profiles
-l:   0	p =   0    i = 0
+l:   0	p =   0    i = 1E-6
```
unalmis, line 30: It might not be relevant to what you're testing, but I had to debug something on this file on #556 and ended up adding the below note to this file so we don't forget.
```python
# Note that the vacuum objective will optimize for current density = 0.
# The rotational transform profile coefficients below will only be used
# as an initial guess, as the rotational transform will not be constrained
# throughout the optimization. See GitHub issue #278.
```

rahulgaur104: I am trying to resolve the issue that iota_flag, pres_flag, and curr_flag are True even when the coefficients are all zeros. This gives the following warning 
```
           warnings.warn(
                "Vacuum objective does not use any profiles, "
                + "ignoring presssure, iota, and current"
```
, which fails my tests. The warning shouldn't arise if all the coefficients are zero. So I change that part of parse_inputs().

For the Heliotron, one of the tests checks if there is a warning when we specify iota coefficients. To make sure that test doesn't fail, I made the iota coefficient finite.

tests/inputs/HELIOTRON_iota
f0uriest, line 1: for these new input/output files, can we rename them to avoid possible confusion with the existing heliotron test cases?

tests/inputs/HELIOTRON_iota
```diff
@@ -0,0 +1,46 @@
+# This is the DESC input file for a vacuum heliotron
```
f0uriest, line 1: is this really vacuum? or just fixed iota?

tests/inputs/HELIOTRON_vacuum_iota
f0uriest, line 1: how is this different from `HELIOTRON_iota`?

desc/io/input_reader.py
```diff
@@ -776,6 +778,166 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+        ----------
+        filename : str or path-like
+            name of the file to create
+        inputs : str or path-like
+            path of the DESC output equilibrium file
+        header : str
+            text to print at the top of the file
+
+        DESC will automatically choose continuation parameters unless you
+        have specified those prameters as kwargs
```
f0uriest, line 805: the signature doesn't take kwargs, and it looks like the code doesn't have stuff for dealing with non-default continuation params? That's fine as long as we fix the docstring to match. Also usually extended description like this goes before the `Parameters` block, or as a seperate `Notes` block (with `------` after)

desc/io/input_reader.py
```diff
@@ -776,6 +778,166 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
 
         f.close()
 
+    @staticmethod
+    def descout_to_input(  # noqa: C901 - fxn too complex
+        filename,
+        inputs,
+        header="#DESC-generated input file",
+        ftol=1e-2,
```
f0uriest, line 786: these arguments aren't included in the docstring, we should explain how they're used.

tests/inputs/HELIOTRON_iota
rahulgaur104: Not vacuum. Just fixed iota.

tests/inputs/HELIOTRON_vacuum_iota
rahulgaur104: Not different. I'll remove it.

desc/io/input_reader.py
```diff
@@ -776,6 +778,166 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
 
         f.close()
 
+    @staticmethod
+    def descout_to_input(  # noqa: C901 - fxn too complex
+        filename,
```
f0uriest, line 783: `filename` and `inputs` are a bit confusing, since `inputs` itself is a filename. Might be better to make it `input_filename` and `output_filename`. 
Or better yet, change `inputs` to `eq` and allow it to also be an instance of the `Equilibrium` class? (or load it from a file if its path-like)

desc/io/input_reader.py
rahulgaur104: I tried kwargs, but that means we should check whether to use kwargs provided by the user or the default values. I can't think how to do that without putting if statements for each key in kwargs. More if statements => more tests. I don't want to decrease coverage. 

I'll edit the docstrings.

desc/io/input_reader.py
rahulgaur104: There are other routines in input_reader with confusing terminology like this. Let me try something else.

desc/io/input_reader.py
rahulgaur104: I think the most important kwarg there is objective. We should definitely set the right objective in the input file.

tests/test_input_output.py
```diff
@@ -111,6 +111,94 @@ def test_write_desc_input_Nones(tmpdir_factory):
...
+    desc_input_truth = "./tests/inputs/LandremanPaul2022_QA_reactorScale_lowRes"
+    with pytest.warns(UserWarning):
+        ir2 = InputReader(cl_args=[str(desc_input_truth)])
+        arr2 = ir2.parse_inputs()[-1]["surface"]
+    arr2 = arr2[arr2[:, 1].argsort()]
+    arr2mneg = arr2[arr2[:, 1] < 0]
+    arr2mpos = arr2[arr2[:, 1] >= 0]
+
+    found0 = 0
+    if (
```
f0uriest, line 147: i think something like `np.testing.assert_allclose` would be cleaner here

tests/inputs/HELIOTRON_iota
rahulgaur104: Maybe HELIOTRON_force_iota?

tests/test_input_output.py
```diff
@@ -111,6 +111,94 @@ def test_write_desc_input_Nones(tmpdir_factory):
     assert no_gtol  # fail test if gtol line is written
 
 
+@pytest.mark.unit
+def test_descout_to_input(tmpdir_factory):
```
f0uriest, line 115: does this also test that the profiles are correct? Would be good to test both even/nonsym profiles (and profiles that have several nonzero values) to make sure that parts works correctly too

rahulgaur104: Not right now. Ok, will do.

tests/inputs/HELIOTRON_vacuum
f0uriest: This is expected behavior. `pres_flag`, `iota_flag` etc are true if the given profiles are present in the input file, even if they're all zero. If the code throws a warning that we expect it to throw, you can wrap the part of the test like so:
```python
with pytest.warns(UserWarning):
    some code that throws warning
```

That will make sure the warning is thrown when we expect it to be, but not fail the test.

tests/inputs/HELIOTRON_iota
f0uriest: i was thinking like `iotest_HELIOTRON_...` if its just for this specific test

tests/inputs/HELIOTRON_vacuum
rahulgaur104: I did not change that part of the code in the latest commit. I use with pytest.warns now.

tests/test_input_output.py
rahulgaur104: I have added a simple check for pressure but no test files.

desc/io/input_reader.py
rahulgaur104: Added short description.

desc/io/input_reader.py
```diff
@@ -776,6 +778,176 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+        gtol=1e-6,
+        maxiter=100,
+    ):
+        """Generate a DESC input file from a DESC output file.
+
+        DESC will automatically choose continuation parameters
+
+        Parameters
+        ----------
+        filename : str or path-like
```
f0uriest, line 799: should update these to match signature

desc/io/input_reader.py
```diff
@@ -776,6 +778,176 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+        optimizer : str
+            type of optimizer
+        header : str
+            text to print at the top of the file
+        ftol : float
+            tolerance on the objective function f
+        xtol : float
+            tolerance on the state vector x
+        gtol : float
+            relative tolerance on the prjected gradient g
```
f0uriest, line 814: gtol is an absolute tolerance, xtol and ftol are relative. Also misspelled `projected`

desc/io/input_reader.py
rahulgaur104: Ah, shoot!

desc/io/input_reader.py
rahulgaur104: Corrected

tests/test_input_output.py
```diff
@@ -111,6 +111,103 @@ def test_write_desc_input_Nones(tmpdir_factory):
...
+        np.minimum(
+            np.linalg.norm(arr1mpos[:, 3:] - arr2mpos[:, 3:]),
+            np.linalg.norm(arr1mpos[:, 3:] + arr2mpos[:, 3:]),
+        ),
+        0,
+        atol=1e-8,
+    )
+
+    if np.linalg.norm(pres1[:, 1]) > 0:
+        np.testing.assert_allclose(pres1(pres1[:, 1] > 0), pres2(pres2[:, 1] > 0))
```
f0uriest, line 208: should probably use a different equilibrium for this part, since all the others have pressure=0, and you initialize the pressure array to 0, so its hard to tell if its actually being written correctly.

tests/inputs/HELIOTRON_iota.h5
f0uriest, line 1: there are a bunch of new files here, are they all needed? 
it looks like the only ones used in the test are `iotest_HELIOTRON` and `iotest_HELIOTRON_output.h5`

tests/test_input_output.py
rahulgaur104: Yes, it currently does not test the profiles. I can choose finite pressure in the iotest_HELIOTRON equilibrium.

tests/inputs/HELIOTRON_iota.h5
rahulgaur104: I thought the rest were removed automatically. Let me remove them now.

desc/io/input_reader.py
```diff
@@ -776,6 +778,176 @@ def write_desc_input(filename, inputs, header=""):  # noqa: C901 - fxn too compl
...
+
+        f.write("\n\n# solver methods\n")
+        f.write(f"optimizer = {optimizer}\n")
+        f.write(f"objective = {objective}\n")
+        f.write("spectral_indexing = {}\n".format(eq0._spectral_indexing))
+        f.write("node_pattern = {}\n".format(eq0._node_pattern))
+
+        f.write("\n# pressure and rotational transform/current profiles\n")
+
+        if eq0.iota:
```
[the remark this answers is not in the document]
f0uriest: similarly, should be `iota = -1 + 0.67 r**2`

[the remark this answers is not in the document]
f0uriest: something like this should work:

```python
from desc.utils import copy_coeffs

pres = eq0.pressure
iota = eq0.iota
L = max(pres.basis.L, iota.basis.L)

# pad with zeros for M,N
modes_new = np.array([np.arange(L), np.zeros(L), np.zeros(L)]).T

p_l = copy_coeffs(pres.params, pres.basis.modes, modes_new)
i_l = copy_coeffs(iota.params, iota.basis.modes, modes_new)

for l in np.arange(L):
    f.write(
        "l: {:3d}\tp = {:16.8E}\t{} = {:16.8E}\n".format(
            int(l), p_l[l], char, i_l[l]
        )
    )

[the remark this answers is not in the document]
Comment (rahulgaur104) [reply desc/io/input_reader.py:860]: iotest_HELIOTRON has finite pressure + odd profiles now so it should be sufficient.

File: tests/test_input_output.py [hunk 24]
```diff
@@ -111,6 +111,103 @@ def test_write_desc_input_Nones(tmpdir_factory):
...
+    np.testing.assert_allclose(
+        np.minimum(
+            np.linalg.norm(arr1mpos[:, 3:] - arr2mpos[:, 3:]),
+            np.linalg.norm(arr1mpos[:, 3:] + arr2mpos[:, 3:]),
+        ),
+        0,
+        atol=1e-8,
+    )
+
+    if np.linalg.norm(pres1[:, 1]) > 0:
```
f0uriest, line 165: is there any reason for the conditional here?

f0uriest (approved):

tests/test_input_output.py
rahulgaur104: Only match pressure profiles with nonzero coefficients.
No need to match vacuum pressures.

rahulgaur104:
This is a relatively short PR regardless of the number of commits so I'll just merge it.

PR #590, opened 2023-07-18
Combine ``Configuration`` and ``Equilibrium``, refactor equilibrium init

f0uriest:
Got sick of forgetting what methods/attributes were where, so now they're finally combined.

Also refactors the init method and breaks some stuff out into standalone utilities to reduce the complexity and make it more readable.


f0uriest:
For reviewing, it's probably easier to look through the commits individually to see the changes, git diff kind of mashes them all together:
feb8d9d : just copies all the stuff from Configuration into Equilibrium
ce453f7 : Adds helper functions to utilities
e632960 : Refactors init method

dpanici (dismissed):
Looks good

desc/utils.py
```diff
@@ -487,3 +488,39 @@ def parse_argname_change(arg, kwargs, oldname, newname):
...
     return arg
+
+
+def setdefault(val, default, cond=None):
+    """Return default if val is None or cond is True, otherwise val."""
+    if cond is None:
+        cond = val is not None
+    if cond:
+        return val
+    return default
```
unalmis, line 499: This doesn't do what the docstring says. Recommend change to
```python
return default if val is None or cond else val
```
though you may instead want (below is different, not a refactor)
```python
return val if val is not None or cond else default
```

unalmis (changes requested):

desc/utils.py
unalmis: none defaults to False, so you may also want to just do cond=False

f0uriest: the docstring was wrong, the code was working as intended. Docstring has been updated.

unalmis: It's still confusing. The updated docstring is still wrong in the case cond is False and val is not None. This does the docstring:
```python
def setdefault(val, default, cond=False):
    """Return val if it isn't None (or if cond is True), otherwise default."""
    return val if val is not None or cond else default
```

f0uriest: The code does what it should be doing, I've updated the docstring again to make it more clear. Its a basic switch, with the default condition being `(val is not None)`, the most common use case being to assign default values when the input is `None`

unalmis: Ok cool. Now the code and docstring agree and implement what you want:
```python
def setdefault(val, default, cond=None):
    return val if cond or (cond is None and val is not None) else default
```

unalmis (approved):
I didn't understand this comment https://github.com/PlasmaControl/DESC/commit/e6329605110f1bd430dfd0e52f7c54204757460a#r122231475, but otherwise looks fine

PR #591, opened 2023-07-21
Move update_target method to new subclass just for fixed constraints

f0uriest:
Most `_Objective`s don't really use the `target_arg` or `update_target` methods, and in some case they don't even make sense, so this moves those methods to a new baseclass `_FixedObjective` which comprises most of the linear constraints.


desc/objectives/linear_objectives.py
```diff
@@ -831,9 +832,7 @@ class FixAxisR(_Objective):
 
     """
 
-    _scalar = False
-    _linear = True
-    _fixed = False
```
dpanici, line 836: `_fixed=False` but the parent objective has `_fixed=True`, does this affect the factorize_linear_constraints at all?

f0uriest: It should have been fixed=True after the changes in #458 but wasn't updated. The `fixed` flag isn't actually used anywhere in the code currently, but I figure keep it for now until we decide on a more permanent fix.

dpanici (approved):

PR #592, opened 2023-07-21
Dp/current potential field

dpanici:
Implements `CurrentPotentialField` and `FourierCurrentPotentialField` classes, which allow for computation of the magnetic field from a surface current density given by 
$$\mathbf{K} =  \mathbf{n} \times \nabla \Phi(\theta,\zeta)  $$

where for the `CurrentPotentialField`, the current potential $\Phi$ is a user-specified, arbitrary function of $\theta$ and $\zeta$, with its derivatives also specified by the user.
For `FourierCurrentPotentialField`, the current potential is assumed to have the form
$$\Phi(\theta,\zeta) = \Phi_{sv}(\theta,\zeta) + \frac{G\zeta}{2\pi} + \frac{I\theta}{2\pi} $$

where $(\theta,\zeta)$ are poloidal and toroidal angles on the winding surface, $G$ is current linking the coil surface poloidally (which for modular or helical coils is determined entirely by the target plasma equilibrium, where one finds that $G = \frac{1}{\mu_0}\int_0^{2\pi} B_{\zeta}d\zeta$), and $I$ is the current linking the coil surface toroidally.  $\Phi_{sv}$ is the single-valued part of $\Phi$ and is given by a `DoubleFourierSeries`:
$$\Phi_{sv}(\theta,\zeta) = \sum_m \sum_n \Phi_{mn} \mathcal{F}_m(\theta) \mathcal{F}_n(\zeta)$$

So for the `FourierCurrentPotentialField`, the free parameters of the field are the Fourier coefficients $\Phi_{mn}$, and the net currents $I$, and $G$.

---------------------------------------
TODO
- [x] subclass with FourierRZToroidalSurface so that can have its own Rb_lmn etc, for winding surface optimization (look at Coils classes)
- [x] Modify the compute magnetic fieild to take advantage of fact that it has discrete symmetry (based off Rory's FB https://github.com/f0uriest/DESC_freeb/blob/1d071712af6844bc00b4c4071f31777907b0cb50/desc/singularities.py#L278)
- [x] don't precompute K
- [x] make field assume fourier series + linear secular terms instead of custom functions (maybe keep the general one?)
- [x] Maybe add change_potential_resolution methods to class (to allow changing of the basis)
- [x] add Phi to data index
- [x] pass in modenumbers instead of basis for fourier currentpotenitalfield
- [x] maybe remove AD from general currentpotential field
  - this is because a lot of stuff is sort of annoying when dealing with the AD
- [x] add basis as public attribute of fourier currentpotentialfield
- [x] change Fouriercurrentpoetnal field to change the potential methods when basis resolution is changed, as right now it will fail if the basis resolution is changed bc the potential functions were defined with the initial basis, and are not changed when change_resolution is called


desc/geometry/surface.py
```diff
@@ -60,7 +60,6 @@ def __init__(
         NFP=1,
         sym="auto",
         rho=1,
-        grid=None,
```
dpanici, line 63: these were unused arguments, so I got rid of them.

dpanici:
This class is mainly to be used with current potentials of form $I\theta/2/\pi + G\zeta/2/\pi  + \sum \Phi_{mn} sin(m\theta - n\zeta$
Unsure if it is worth subclassing for this specific implementation,or leaving it general and then having the user define their functions for the potential. Probably leaving it general is fine though I think

desc/magnetic_fields.py
```diff
@@ -59,6 +61,40 @@ def biot_savart(eval_pts, coil_pts, current):
...
+    assert J.shape == rs.shape
+    JdV = J * dV[:, None]
+    B = jnp.zeros_like(re)
+
+    def body(i, B):
+        r = re - rs[i, :]
+        num = jnp.cross(JdV[i, :], r, axis=-1)
+        den = jnp.linalg.norm(r, axis=-1) ** 3
+        B = B + jnp.where(den[:, None] == 0, 0, num / den[:, None])
+        return B
```
unalmis, line 93:
```python
re, rs, J, dV = map(jnp.asarray, (re, rs, J, dV))
assert J.shape == rs.shape

JdV = (dV * J.T).T
dr = re - rs[:, jnp.newaxis]
num = jnp.cross(JdV, dr, axis=-1)
den = jnp.linalg.norm(dr, axis=-1) ** 3
B = 1e-7 * jnp.nansum((num.T / den).T, axis=0)
return B
```
?

I personally prefer to use transposing to to make broadcasting work and reserve creating a new axis for when a larger matrix needs to be materialized (jnp.where duplicates `den` 3 times so you need `den[:, np.newaxis]` there, but for `JdV` you can just transpose). That way it is clear when something is being vectorized (as in my creation of `dr`) and when we're just getting broadcasting to work (like creating `JdV`).

desc/magnetic_fields.py
```diff
@@ -795,3 +831,265 @@ def odefun(rpz, s):
...
+            / surface_grid.NFP
+        )  # divide by NFP here so that the NFP loop
+        # in compute_magnetic_field is correct
+
+        # compute the potential derivatives on the surface
+        phi_t = self._potential_t(theta, zeta)
+        phi_z = self._potential_z(theta, zeta)
+
+        # compute surface normal magnitude
+        ns_mag = jnp.linalg.norm(jnp.cross(data["e_theta"], data["e_zeta"]), axis=1)
```
unalmis, line 1039: if you end up pushing another commit you could also move ns above self._dv and use it for the computation there

desc/geometry/surface.py
f0uriest: thanks, forgot to remove them in #583

desc/magnetic_fields.py
f0uriest: The issue with vectorizing this is that the `dr` array you create can be pretty huge (N^2 for N src/eval pts). Doing it in a loop limits the total memory to O(N), at the expense of being slightly less efficient on GPU at lower resolution.

f0uriest: One possibly in between option would be to do the loop over whichever is smaller of the source/eval pts. right now we loop over the source points, but if we have many sources and few eval pts its a lot more efficient to loop over the eval pts.

f0uriest (commented):
- Are there examples of realistic use cases where the potential would NOT be given as a fourier series + secular terms? I'm wondering if it would be better to have this assume that parameterization, so the user only needs to supply the fourier coefficients rather than functions.

- I'm not crazy about precomputing K on a fixed grid, since I'm not sure that will work if we want to optimize K in the loop. Is it that much more expensive to evaluate on the fly? IE, the user can pass in the surface grid to `compute_magnetic_field`. We could also then add compute functions for `Phi`, `K` etc to `desc.compute._surface.py` (I have something like this on the free boundary branch)

desc/magnetic_fields.py
unalmis: my vote is to loop over the vectorized algorithm in chunks. if nothing else it's cooler
```python
re, rs, J, dV = map(jnp.asarray, (re, rs, J, dV))
assert J.shape == rs.shape
JdV = (dV * J.T).T
JdV = JdV[:, jnp.newaxis]
chunk_size = 30

def body(i, B):
    start = i * chunk_size
    end = min(start + chunk_size, rs.shape[0])
    # loop over vectorized algorithm in chunks to save memory
    dr = re - rs[start:end, jnp.newaxis]
    num = jnp.cross(JdV, dr, axis=-1)
    den = jnp.linalg.norm(dr, axis=-1) ** 3
    B += jnp.nansum((num.T / den).T, axis=0)
    return B

B = 1e-7 * fori_loop(0, rs.shape[0] // chunk_size, body, jnp.zeros_like(re))
return B
```

dpanici:
> * Are there examples of realistic use cases where the potential would NOT be given as a fourier series + secular terms? I'm wondering if it would be better to have this assume that parameterization, so the user only needs to supply the fourier coefficients rather than functions.

My only use case for this right now is the REGCOIL and LM stellarator stuff. Both of those we use fourier series and secular terms (linear in theta and zeta). It might be that for the LM stell stuff we want secular terms that are not just linear in theta and zeta, but at least right now I thin that this can be made to only be this parametrization

> * I'm not crazy about precomputing K on a fixed grid, since I'm not sure that will work if we want to optimize K in the loop. Is it that much more expensive to evaluate on the fly? IE, the user can pass in the surface grid to `compute_magnetic_field`. We could also then add compute functions for `Phi`, `K` etc to `desc.compute._surface.py` (I have something like this on the free boundary branch)

Yea this I had in mind mainly for a fixed surface, since if the surface changes at all K would need to be precomputed. K can be kind of expensive but I dont think as expensive as the biot savart loop is. I can change so it is not precomputed, and add the `compute._surface` functions

desc/magnetic_fields.py
dpanici: smart, I can do that

desc/magnetic_fields.py
f0uriest: does this work with jit/AD if the chunk size isnt an even multiple of the total number of points?

f0uriest:
> My only use case for this right now is the REGCOIL and LM stellarator stuff. Both of those we use fourier series and secular terms (linear in theta and zeta). It might be that for the LM stell stuff we want secular terms that are not just linear in theta and zeta, but at least right now I thin that this can be made to only be this parametrization

ok yeah that sounds cleaner. And maybe rename the class to `FourierCurrentPotentialField` or something, to leave room for other parameterizations in the future

f0uriest:
> Yea this I had in mind mainly for a fixed surface, since if the surface changes at all K would need to be precomputed. K can be kind of expensive but I dont think as expensive as the biot savart loop is. I can change so it is not precomputed, and add the compute._surface functions

This might also be a place where doing FFTs in the poloidal direction could help speed things up. Can probably be a separate PR though.

desc/magnetic_fields.py
unalmis: i couldn't get it to work. if it does its not worth added code complexity

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+        modes are symmetric. If True, non-symmetric modes will be truncated.
+    rho : float [0,1]
+        flux surface label for the toroidal surface
+    name : str
+        name for this surface
+    check_orientation : bool
+        ensure that this surface has a right handed orientation. Do not set to False
+        unless you are sure the parameterization you have given is right handed
+        (ie, e_theta x e_zeta points outward from the surface).
+    surface: FourierRZToroidalSurface, optional, default None
```
f0uriest, line 885: rather than passing this into init it might be better to make a `from_FourierRZSurface` method or `to_CurrentPotentialSurface` method in the surface class. Those methods would then only take the surface and potential functions, not all the rest of the stuff

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+            "_surface_grid",
+            "_params",
+        ]
+    )
+
+    def __init__(
+        self,
+        potential,
+        surface_grid,
+        params={},
```
f0uriest, line 905: should use None for default here

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+        + [
+            "_surface_grid",
+            "_params",
+        ]
+    )
+
+    def __init__(
+        self,
+        potential,
+        surface_grid,
```
f0uriest, line 904: could default to None and have preset default based on resolution of surface? Or remove entirely and just make it defined within the compute method where its needed?

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+    @property
+    def params(self):
+        """Dict of parameters to pass to potential function and its derivatives."""
+        return self._params
+
+    @params.setter
+    def params(self, new):
+        if new != self._params:
+            if len(new) != len(self._params):
+                raise UserWarning(
```
f0uriest, line 963: you probably want `warnings.warn` here, `raise` will make it an error. 
Could also make a `warnif` util similar to `errorif` to handle the conditional.

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+        self._dV = (
+            surface_grid.weights * ns_mag / surface_grid.NFP
+        )  # divide by NFP here so that the NFP loop
+        # in compute_magnetic_field is correct
+
+        # compute the potential derivatives on the surface
+        phi_t = self._potential_t(theta, zeta)
+        phi_z = self._potential_z(theta, zeta)
+
+        # compute K = n x nabla(Phi)
```
f0uriest, line 1031: is it `n x grad(Phi)` or `grad(Phi) x n`? the docstring says one, this says another.

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+                Derivative(
+                    partial(self.potential, **self.params), argnum=1, mode="grad"
+                ).compute,
+                in_axes=[0, 0],
+            )
+
+    def _compute_surface_current(self, surface_grid=None, params=None):
+        if surface_grid is None:
+            surface_grid = self.surface_grid
+        if (params is None) or (len(params) == 0):
```
f0uriest, line 1006: might be better to leave out the `or len(params) == 0` bit? I think all the magnetic fields default to `params=None` at this point.

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+
+        """
+        assert basis.lower() in ["rpz", "xyz"]
+        if isinstance(coords, Grid):
+            coords = coords.nodes
+        coords = jnp.atleast_2d(coords)
+        if basis == "rpz":
+            coords = rpz2xyz(coords)
+
+        if (params is None) or (len(params) == 0):
```
f0uriest, line 1067: i don't think we need this here, the one in `_compute_surface_current` will do the same thing

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+            surface_grid.weights * ns_mag / surface_grid.NFP
+        )  # divide by NFP here so that the NFP loop
+        # in compute_magnetic_field is correct
+
+        # compute the potential derivatives on the surface
+        phi_t = self._potential_t(theta, zeta)
+        phi_z = self._potential_z(theta, zeta)
+
+        # compute K = n x nabla(Phi)
+        self._K = xyz2rpz_vec(
```
f0uriest, line 1032: I think this should be returned rather than stored as an attribute if we want it to be differentiable

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+    @potential.setter
+    def potential(self, new):
+        if new != self._potential:
+            self._potential = new
+            # reset derivatives if potential has changed, using AD
+            self._set_derivatives()
+
+    def _set_derivatives(self, potential_dtheta=None, potential_dzeta=None):
+        if potential_dtheta:
+            self._potential_t = partial(potential_dtheta, **self.params)
```
f0uriest, line 983: if you close over `params` here we won't be able to differentiate wrt them

desc/magnetic_fields.py
f0uriest: also these params aren't passed to the potential functions

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+            B = xyz2rpz_vec(B, x=coords[:, 0], y=coords[:, 1])
+        return B
+
+
+class FourierCurrentPotentialField(CurrentPotentialField):
+    """Magnetic field due to a surface current potential on a toroidal surface.
+
+        surface current K is assumed given by
+
+        K = ∇ Φ x n
```
f0uriest, line 1100: should check whether its this or the flipped, with minus sign

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+        Φ(θ,ζ) = Φₛᵥ(θ,ζ) + Gζ/2π + Iθ/2π
+
+        where n is the winding surface unit normal
+              Phi is the current potential function,
```
f0uriest, line 1104: why indented?

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+        modes are symmetric. If True, non-symmetric modes will be truncated.
+    rho : float [0,1]
+        flux surface label for the toroidal surface
+    name : str
+        name for this surface
+    check_orientation : bool
+        ensure that this surface has a right handed orientation. Do not set to False
+        unless you are sure the parameterization you have given is right handed
+        (ie, e_theta x e_zeta points outward from the surface).
+    surface: FourierRZToroidalSurface, optional, default None
```
f0uriest, line 1145: same as above, might be better to have a separate `from_surface` method

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+        Existing FourierRZToroidalSurface object to create a
+        CurrentPotentialField with, if provided will use this
+        object's R_lmn, Z_lmn etc instead of any passed-in values
+
+    """
+
+    _io_attrs_ = (
+        MagneticField._io_attrs_
+        + FourierRZToroidalSurface._io_attrs_
+        + ["_surface_grid", "Phi_mn", ""]
```
f0uriest, line 1155: need to save I, G as well?

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+    )
+
+    def __init__(
+        self,
+        Phi_mn,
+        basis,
```
f0uriest, line 1161: I think in other places we have the user pass in an array of mode numbers and then make a basis based on that.

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+
+    _io_attrs_ = (
+        MagneticField._io_attrs_
+        + FourierRZToroidalSurface._io_attrs_
+        + ["_surface_grid", "Phi_mn", ""]
+    )
+
+    def __init__(
+        self,
+        Phi_mn,
```
f0uriest, line 1160: could have this and I,G default to 0

desc/magnetic_fields.py
f0uriest: Also, weren't we going to make this a compute function? could treat the potential functions as transforms

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+        modes_Z=None,
+        NFP=1,
+        sym="auto",
+        rho=1,
+        name="",
+        check_orientation=True,
+        surface=None,
+    ):
+        assert isinstance(
+            basis, DoubleFourierSeries
```
f0uriest, line 1177: should make a public attribute for phi basis.

f0uriest: Also do we want the phi basis to share M,N resolution with the underlying surface? if not we might want to make that clear and provide some way to change the resolution of the phi basis separately from the `surface.change_resolution`

Could also do what daniel did for omnigenity and have a new `M_phi` and `N_phi` to control resolution of the potential

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+            f"Expected Phi_mn size {basis.num_modes.size}, instead got {Phi_mn.size}"
+        )
+        self._Phi_mn = Phi_mn
+
+        assert np.isscalar(I), "I must be a scalar"
+        assert np.isscalar(G), "G must be a scalar"
+        self._I = I
+        self._G = G
+
+        def _fourier_potential(theta, zeta, I, G, Phi_mn):
```
f0uriest, line 1190: could these be made into compute functions?

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
+    @property
+    def I(self):  # noqa: E743
+        """Net current linking the plasma and the coils toroidally."""
+        return self._I
+
+    @I.setter
+    def I(self, new):  # noqa: E743
+        assert np.isscalar(new), "I must be a scalar"
+        self._I = new
+        self._params["I"] = new
```
f0uriest, line 1268: a bit confused, is this its own attribute or is it part of the params dict?

setup.cfg
```diff
@@ -47,7 +47,7 @@ markers=
 filterwarnings=
     error
     ignore::pytest.PytestUnraisableExceptionWarning
-    ignore::RuntimeWarning:desc.compute
+    ignore::RuntimeWarning
```
f0uriest, line 50: This covers a lot of possible things (eg, missing io attrs), not sure we want to ignore them everywhere

dpanici: this was an accident, did not mean to include in the commit

desc/magnetic_fields.py
dpanici: Should this be the case for all the params for magnetic field classes?

desc/magnetic_fields.py
dpanici: latter, fixed

desc/magnetic_fields.py
f0uriest: Yeah I think so. Using mutable objects as default arguments can lead to all sorts of weird bugs

desc/magnetic_fields.py
dpanici: True but I cannot vmap it in the next line if we dont close over the params. Maybe not even worth keeping the AD one. I forgot why we need to use vmap for the auto derivative though

desc/magnetic_fields.py
dpanici: It is this, this is opposite the REGCOIL convention, I chose this though so that positive G means positive toroidal field, which is more intuitive to me

image

desc/magnetic_fields.py
```diff
@@ -795,3 +831,464 @@ def odefun(rpz, s):
...
     return r, z
+
+
+class CurrentPotentialField(MagneticField, FourierRZToroidalSurface):
+    """Magnetic field due to a surface current potential on a toroidal surface.
+
+        surface current K is assumed given by
+         K = ∇ Φ x n
+        where n is the winding surface unit normal
+              Phi is the current potential function,
```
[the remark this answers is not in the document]
dpanici: to separate out where I am just describing terms in the eqn

desc/magnetic_fields.py
dpanici: params but added it as an attribute as well to allow checking of it being a scalar. Though this is bypassed when it is passed in as a param I guess

desc/magnetic_fields.py
dpanici: they are decoupled now, can see the Phi basis resolutions through `self.Phi_basis.M` (or N)

desc/magnetic_fields.py
dpanici: I moved FOurierCurrentPotentialField to be its own class instead of subclass with CurrentPotentialFIeld, so this is now just an attribute

desc/magnetic_fields.py
dpanici: removed the AD part as it was complicated to deal with when using the data index

desc/compute/_surface.py
```diff
@@ -427,6 +427,173 @@ def _e_zeta_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwa
...
+    coordinates="tz",
+    data=[],
+    parameterization="desc.magnetic_fields.CurrentPotentialField",
+    basis="basis",
+)
+def _Phi_CurrentPotentialField(params, transforms, profiles, data, **kwargs):
+    data["Phi"] = transforms["potential"](
+        transforms["grid"].nodes[:, 1],
+        transforms["grid"].nodes[:, 2],
+        **params["params"][0]
```
dpanici, line 517: this `[0]` is here becauuse the params["params"] is an `ndarray` type, not the `dict` type I expected, given that the `CurrentPotentialField.params` attribute is a dictionary. I was unsure of a better way to account for the fact that this class allows arbitrary functions for the potential and thus we don't know what the params will be beforehand, so they are not each individually attributes of the class.

desc/magnetic_fields.py
dpanici: It is hard to think of a good default resolution, as even with simple potentials and simple surfaces (like a rotating ellipse), you still need pretty high surface grid resolution to accurately resolve the magnetic fields inside (I have found at least)

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+            modes_Z,
+            NFP,
+            sym,
+            rho,
+            name,
+            check_orientation,
+        )
+
+
+class FourierCurrentPotentialField(_MagneticField, FourierRZToroidalSurface):
```
dpanici, line 1419: The reason why this is not subclassed from `CurrentPotentialField` is that I wanted to be able to have the compute function for `Phi` from this use the transforms and `Phi_basis` cleanly, in that it will get a transform with `get_transforms` and see the `Phi_basis` and use that in the transform. 

That requires separate compute functions for the `Phi` from this `FourierCurrentPotentialField` class and the `CurrentPotentialField` class. However, if the fourier was a subclass of the  `CurrentPotentialField`, our API for compute gets confused as it thinks I've defined the same function twice for the same parametrization. 

I was unable to think of how to avoid this without just separating the two classes. This works OK as the `K` compute variable only uses the calculated derivatives and geometry, so that can have one entry for both classes.

The only unclean part is that I now have duplicated code in the `compute_magnetic_field` function for these two classes, as they are the exact same.

Maybe a possible solution is to make these 2 separate subclasses of an overarching `SurfaceCurrentPotentialField`, but that would only be to avoid duplicating the compute_magnetic_field code

desc/compute/_surface.py
```diff
@@ -427,6 +427,173 @@ def _e_zeta_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwa
     return data
 
 
+@register_compute_fun(
+    name="Phi",
+    label="\\Phi",
+    units="A",
+    units_long="amperes",
```
f0uriest, line 434: think this should be capitalized

desc/compute/_surface.py
```diff
@@ -427,6 +427,173 @@ def _e_zeta_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwa
...
+    description="Surface current potential",
+    dim=2,
```
f0uriest, line 436: isnt this a scalar?

desc/compute/_surface.py
```diff
@@ -427,6 +427,173 @@ def _e_zeta_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwa
...
+    params=["I", "G", "Phi_mn"],
+    transforms={"Phi": [[0, 0, 0]]},
+    profiles=[],
+    coordinates="tz",
+    data=[],
+    parameterization="desc.magnetic_fields.FourierCurrentPotentialField",
+    basis="basis",
```
f0uriest, line 443: dont think this is needed?

desc/compute/_surface.py
```diff
@@ -427,6 +427,173 @@ def _e_zeta_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwa
...
+    coordinates="tz",
+    data=["Phi_t", "Phi_z", "x", "e_theta", "e_zeta", "|e_theta x e_zeta|"],
+    parameterization=[
+        "desc.magnetic_fields.CurrentPotentialField",
+        "desc.magnetic_fields.FourierCurrentPotentialField",
+    ],
+    basis="basis",
+)
+def _K_CurrentPotentialField(params, transforms, profiles, data, **kwargs):
+    data["K"] = (
```
f0uriest, line 590: would it be clearer to have an intermediate qty for `grad(Phi)` and then do `cross(n, grad(Phi))`?

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+    Parameters
+    ----------
+    potential : callable
+        function to compute the current potential. Should have a signature of
+        the form potential(theta,zeta,**params) -> ndarray.
+        theta,zeta are poloidal and toroidal angles on the surface
+    potential_dtheta: callable
+        function to compute the theta derivative of the current potential
+    potential_dzeta: callable
+        function to compute the theta derivative of the current potential
```
f0uriest, line 1156: zeta derivative

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+            self._surface_grid = new
+
+    @property
+    def params(self):
+        """Dict of parameters to pass to potential function and its derivatives."""
+        return self._params
+
+    @params.setter
+    def params(self, new):
+        if new != self._params:
```
f0uriest, line 1244: is this check needed?

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+            if len(new) != len(self._params):
```
f0uriest, line 1245: could use `warnif` here

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+
+        name : str
+            name for this field
+        check_orientation : bool
+            ensure that this surface has a right handed orientation. Do not set to False
+            unless you are sure the parameterization you have given is right handed
+            (ie, e_theta x e_zeta points outward from the surface).
+
+        """
+        if not isinstance(surface, FourierRZToroidalSurface):
```
f0uriest, line 1388: could use `errorif` here

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        cls,
+        surface,
+        potential,
+        potential_dtheta,
+        potential_dzeta,
+        surface_grid,
+        params=None,
+        rho=1,
+        name="",
+        check_orientation=True,
```
f0uriest, line 1358: are these needed here? Aren't they already supplied by the surface object?

desc/magnetic_fields.py
f0uriest: I'd prefer `n x grad(Phi)` since that seems to be much more common in the literature and other codes. It's also what I've been using on the free boundary branch.

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        Poloidal and Toroidal modenumbers corresponding to passed-in Phi_mn coefficients
+    I : float
+        Net current linking the plasma and the coils toroidally
+        Denoted I in the algorithm
+    G : float
+        Net current linking the plasma and the coils poloidally
+        Denoted G in the algorithm
+    surface_grid : Grid
+        grid upon which to evaluate the surface current density K
+    sym_Phi : str or False, one of {"auto","cos","sin",False}
```
f0uriest, line 1450: is it possible for the potential to be asymmetric while the surface is symmetric? It feels cleaner if there's only one `sym` param for both if possible.

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        Fourier coefficients of the double FourierSeries part of the current potential.
+    modes_Phi : array-like, shape(k,2)
+        Poloidal and Toroidal modenumbers corresponding to passed-in Phi_mn coefficients
+    I : float
+        Net current linking the plasma and the coils toroidally
+        Denoted I in the algorithm
+    G : float
+        Net current linking the plasma and the coils poloidally
+        Denoted G in the algorithm
+    surface_grid : Grid
```
f0uriest, line 1448: is this needed here, or just when we compute things?

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        ensure that this surface has a right handed orientation. Do not set to False
+        unless you are sure the parameterization you have given is right handed
+        (ie, e_theta x e_zeta points outward from the surface).
+
+    """
+
+    _io_attrs_ = (
+        _MagneticField._io_attrs_
+        + FourierRZToroidalSurface._io_attrs_
+        + ["_surface_grid", "Phi_mn", "I", "G"]
```
f0uriest, line 1480: i think these should have `_` before them

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+    name : str
+        name for this field
+    check_orientation : bool
+        ensure that this surface has a right handed orientation. Do not set to False
+        unless you are sure the parameterization you have given is right handed
+        (ie, e_theta x e_zeta points outward from the surface).
+
+    """
+
+    _io_attrs_ = (
```
f0uriest, line 1183: We should probably make it clear somewhere that saving this doesn't really work since the potential function can't be serialized. Maybe overwrite the save method to throw an error?

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        B field from this current density K on the surface
+
+    Parameters
+    ----------
+    Phi_mn : ndarray
+        Fourier coefficients of the double FourierSeries part of the current potential.
+    modes_Phi : array-like, shape(k,2)
+        Poloidal and Toroidal modenumbers corresponding to passed-in Phi_mn coefficients
+    I : float
+        Net current linking the plasma and the coils toroidally
```
f0uriest, line 1443: might want to say "plasma and surface" rather than "plasma and coils"

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        return self._I
+
+    @I.setter
+    def I(self, new):  # noqa: E743
+        assert np.isscalar(new), "I must be a scalar"
+        self._I = new
+
+    @property
+    def G(self):
+        """Net current linking the plasma and the coils poloidally."""
```
f0uriest, line 1569: similar thing about "plasma and coils"

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+            M = M if M is not None else self.M
+            N = N if N is not None else self.N
+            Phi_modes_old = self.Phi_basis.modes
+            self.Phi_basis.change_resolution(M=M, N=N, NFP=self.NFP, sym=self.sym_Phi)
+
+            self._Phi_mn = copy_coeffs(self.Phi_mn, Phi_modes_old, self.Phi_basis.modes)
+            self._M_Phi = M
+            self._N_Phi = N
+
+    def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
```
f0uriest, line 1641: if we want to reduce duplicated code we could make this a private standalone function and then have both methods call that, similar to some of the `Equilibrium` methods

f0uriest, line 1641: if we want to reduce duplicated code we could make this a private standalone function and then have both methods call that, similar to some of the `Equilibrium` methods

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        this function then uses biot-savart to find the
+        B field from this current density K on the surface
+
+    Parameters
+    ----------
+    potential : callable
+        function to compute the current potential. Should have a signature of
+        the form potential(theta,zeta,**params) -> ndarray.
+        theta,zeta are poloidal and toroidal angles on the surface
+    potential_dtheta: callable
```
f0uriest, line 1153: can you explain a bit why we can't use AD to get the derivatives? I thought something like we do for `ScalarPotentialField` would work?

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        """Phi toroidal deriv. function, signature (theta,zeta,**params) -> value."""
+        return self._potential_dzeta
+
+    @potential_dzeta.setter
+    def potential_dzeta(self, new):
+        if new != self._potential_dzeta:
+            assert callable(new), "Potential derivative must be callable!"
+            self._potential_dzeta = new
+
+    def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
```
f0uriest, line 1286: if we want to reduce duplicated code we could make this a private standalone function and then have both methods call that, similar to some of the `Equilibrium` methods

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        sym="auto",
+        rho=1,
+        name="",
+        check_orientation=True,
+    ):
+        self._Phi_mn = Phi_mn
+
+        Phi_mn, modes_Phi = map(np.asarray, (Phi_mn, modes_Phi))
+
+        assert issubclass(modes_Phi.dtype.type, np.integer)
```
f0uriest, line 1505: i think `np.issubdtype` is the preferred way to do this

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        sym_Phi = kwargs.pop("sym_Phi", None)
+        assert len(kwargs) == 0, "change_resolution got unexpected kwarg: {kwargs}"
+        self._NFP = NFP if NFP is not None else self.NFP
+        self._sym_Phi = sym_Phi if sym_Phi is not None else self.sym_Phi
+        if L is not None:
+            warnings.warn("Phi basis does not have radial resolution, ignoring L")
+        if len(args) == 2:
+            M, N = args
+        elif len(args) == 3:
+            L, M, N = args
```
f0uriest, line 1625: Most of this was to account for surfaces being either L/M or M/N, but in this case we know its always M/N so we can probably get rid of most of this.

desc/magnetic_fields.py
```diff
@@ -1094,3 +1130,644 @@ def odefun(rpz, s):
...
+        cls,
+        surface,
+        Phi_mn=np.array([0.0]),
+        modes_Phi=np.array([[0, 0]]),
+        I=0,
+        G=0,
+        surface_grid=None,
+        sym_Phi="auto",
+        name="",
+        check_orientation=True,
```
f0uriest, line 1713: are these needed?

desc/compute/_surface.py
dpanici: Maybe, but the contravariant basis vectors are a bit hard to define on the surface, and afaik it involves inverting the covariant metric tensor. It is easiest to go straight to the `n x grad(phi)` expression as then all the contravariant vectors become covariant basis vectors and those are immediately defined for a surface.

desc/magnetic_fields.py
dpanici: mainly when we compute things, BUT I want this class to be like the other MagneticField classes in that it defines a single magnetic field, in this case one given by a surface current, and so I want it to have as attributes the `I` and `G`

desc/magnetic_fields.py
dpanici: It does, but it gets trickier when we want to use the compute index as well, we need params to be something that can be passed in, but to AD with keyword arguments (at least from what I was seeing), it was easier to use `functools.partial` to remove them first before doing the AD, and that means that any times `params` changes, we need to re-create the derivatives

it is probably possible but also I was just having a harder time dealing with it, so I'd rather get this in without it then come back to adding AD later

desc/magnetic_fields.py
dpanici: we can

desc/magnetic_fields.py
dpanici: Sure, I can change it, but begrudgingly as I don't like that a positive poloidal current `G` results in a negative `B_zeta` now

desc/magnetic_fields.py
dpanici: I am not completely sure to be honest, in the fortran regcoil version, the phi symmetry is also decoupled from the symmetry of the equilbrium surface. I need to think more about if the `eq.sym` implies the symmetry of the `Phi`

desc/compute/_surface.py
```diff
@@ -494,6 +494,165 @@ def _e_zeta_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwa
...
+        transforms["Phi"].transform(params["Phi_mn"])
+        + params["G"] * transforms["Phi"].nodes[:, 2].flatten(order="F") / 2 / jnp.pi
+        + params["I"] * transforms["Phi"].nodes[:, 1].flatten(order="F") / 2 / jnp.pi
+    )
+    return data
+
+
+@register_compute_fun(
+    name="Phi_t",
+    label="\\partial_{\\theta}Phi",
```
f0uriest, line 522: think you want `\\Phi`

desc/compute/_surface.py
```diff
@@ -494,6 +494,165 @@ def _e_zeta_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwa
...
+def _Phi_t_FourierCurrentPotentialField(params, transforms, profiles, data, **kwargs):
+    data["Phi_t"] = (
+        transforms["Phi"].transform(params["Phi_mn"], dt=1) + params["I"] / 2 / jnp.pi
+    )
+    return data
+
+
+@register_compute_fun(
+    name="Phi_z",
+    label="\\partial_{\\zeta}Phi",
```
f0uriest, line 543: `\\Phi`

desc/compute/_surface.py
f0uriest: were you able to figure out why its an array of dict instead of just a dict? AFAIK we don't do any weird wrapping or casting

desc/compute/_surface.py
```diff
@@ -494,6 +494,165 @@ def _e_zeta_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwa
...
+        transforms["grid"].nodes[:, 1],
+        transforms["grid"].nodes[:, 2],
+        **params["params"][0]
+    )
+    return data
+
+
+@register_compute_fun(
+    name="Phi_t",
+    label="\\partial_{\\theta}Phi",
```
f0uriest, line 587: `\\Phi`

desc/compute/utils.py
```diff
@@ -362,6 +362,12 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
             transforms["rotmat"] = obj.rotmat
         elif c == "shift":
             transforms["shift"] = obj.shift
+        elif c == "potential":
```
f0uriest, line 365: on one of my branches I changed this to be 
```python
elif hasattr(obj, c):
    transforms[c] = getattr(obj, c)
```
to avoid having all these different conditionals

desc/magnetic_fields.py
```diff
@@ -1094,3 +1129,614 @@ def odefun(rpz, s):
...
+    R_lmn, Z_lmn : array-like, shape(k,)
+        Fourier coefficients for winding surface R and Z in cylindrical coordinates
+    modes_R : array-like, shape(k,2)
+        poloidal and toroidal mode numbers [m,n] for R_lmn.
+    modes_Z : array-like, shape(k,2)
+        mode numbers associated with Z_lmn, defaults to modes_R
+    NFP : int
+        number of field periods
+    sym : bool
+        whether to enforce stellarator symmetry. Default is "auto" which enforces if
```
f0uriest, line 1167: should make it clear we mean symmetry for the surface, not the potential

desc/magnetic_fields.py
```diff
@@ -1094,3 +1129,614 @@ def odefun(rpz, s):
...
+        modes are symmetric. If True, non-symmetric modes will be truncated.
+    rho : float [0,1]
```
f0uriest, line 1169: since these surfaces are generally outside the plasma we might want to remove this?

desc/magnetic_fields.py
```diff
@@ -1094,3 +1129,614 @@ def odefun(rpz, s):
...
+        Denoted I in the algorithm
+    G : float
+        Net current linking the plasma and the surface poloidally
+        Denoted G in the algorithm
+        NOTE: a negative G will tend to produce a positive toroidal magnetic field
+        B in DESC, as in DESC the poloidal angle is taken to be positive
+        and increasing when going in the clockwise direction, which with the
+        convention n x grad(phi) will result in a toroidal field in the negative
+        toroidal direction.
+    sym_Phi : str or False, one of {"auto","cos","sin",False}
```
f0uriest, line 1406: when the options are a discrete set like this you don't have to give the type, ie can just write `sym_Phi : {"auto", "cos", "sin", False}` with the default option first

desc/magnetic_fields.py
```diff
@@ -1094,3 +1129,614 @@ def odefun(rpz, s):
...
+
+    @property
+    def I(self):  # noqa: E743
+        """Net current linking the plasma and the surface toroidally."""
+        return self._I
+
+    @I.setter
+    def I(self, new):  # noqa: E743
+        assert np.isscalar(new), "I must be a scalar"
+        self._I = new
```
f0uriest, line 1502: we may also want to cast these to floats, doing AD with integers can give unexpected results

desc/magnetic_fields.py
```diff
@@ -1094,3 +1129,614 @@ def odefun(rpz, s):
...
+                sym_Phi = False
+            # catch case where only (0,0) mode is given as 0
+            if np.all(Phi_mn == 0.0) and np.all(modes_Phi == 0):
+                sym_Phi = "cos"
+        self._sym_Phi = sym_Phi
+        self._Phi_basis = DoubleFourierSeries(M=M_Phi, N=N_Phi, NFP=NFP, sym=sym_Phi)
+
+        assert np.isscalar(I), "I must be a scalar"
+        assert np.isscalar(G), "G must be a scalar"
+        self._I = I
```
f0uriest, line 1487: should make sure these are floats

desc/magnetic_fields.py
```diff
@@ -1094,3 +1129,614 @@ def odefun(rpz, s):
...
+        source grid upon which to evaluate the surface current density K
+
+    Returns
+    -------
+    field : ndarray, shape(N,3)
+        magnetic field at specified points
+
+    """
+    assert basis.lower() in ["rpz", "xyz"]
+    if isinstance(coords, Grid):
```
f0uriest, line 1705: in other places we've switched to using `if hasattr(coords, "nodes"):` because the actual base class is now `_Grid`

desc/magnetic_fields.py
```diff
@@ -1094,3 +1129,614 @@ def odefun(rpz, s):
...
+    if basis == "rpz":
+        coords = rpz2xyz(coords)
+    surface_grid = grid or LinearGrid(M=30, N=30, NFP=field.NFP)
+
+    # compute surface current, and store grid quantities
+    # needed for integration in class
+    # TODO: does this have to be xyz, or can it be computed in rpz as well?
+    data = field.compute(["K", "x"], grid=surface_grid, basis="xyz", params=params)
+    _K = xyz2rpz_vec(data["K"], phi=surface_grid.nodes[:, 2])
+    _rs = xyz2rpz_vec(data["x"], phi=surface_grid.nodes[:, 2])
```
f0uriest, line 1717: i don't think you want the vector version here, since `rs` is a coordinate not a vector (ie use regular `xyz2rpz`)

desc/magnetic_fields.py
```diff
@@ -1094,3 +1129,614 @@ def odefun(rpz, s):
...
+    # surface grid weights, as we account for that when doing the for loop
+    # over NFP
+    _dV = surface_grid.weights * data["|e_theta x e_zeta|"] / surface_grid.NFP
+
+    def nfp_loop(j, f):
+        # calculate (by rotating) rs, rs_t, rz_t
+        phi = (surface_grid.nodes[:, 2] + j * 2 * jnp.pi / surface_grid.NFP) % (
+            2 * jnp.pi
+        )
+        rs = rpz2xyz_vec(_rs, phi=phi)
```
f0uriest, line 1728: same as above. To rotate this you just need to add `j*2*np.pi/NFP` to the phi coordinate

f0uriest (commented):
A few minor fixes noted. Should also try to get the coverage up a bit higher, it looks like there are a lot of conditionals etc that aren't tested. 

Should also add the new classes to `test_compute_everything`

desc/magnetic_fields.py
dpanici: I am finding that something is messing up when I don't use the vectorized version and instead just use `xyz2rpz`, need to see why that is but leaving it for now

dpanici: fixed it

desc/compute/_surface.py
dpanici: https://github.com/PlasmaControl/DESC/blob/7fcfdd3bbb4964440ce8460d0083374c331c72e7/desc/compute/utils.py#L292

this line turns anything in the params dictionary into a 1D ndarray

dpanici: I could avoid this by grabbing the params dict through the `get_transforms` function instead, but that is kinda confusing. This seems to work for now

f0uriest: hmm ok that might break things if you try to jit or optimize the params of a current potential field though. np arrays get converted to jax arrays but jax arrays can't hold arbitrary objects like dictionaries.

we should probably change that to not cast dicts to arrays, and maybe use `jnp.atleast_1d` to make sure we're not doing anything funky that jax doesn't like

f0uriest: you probably dont want to go through `get_transforms` because that would make it a constant during any optimization

desc/magnetic_fields.py
```diff
@@ -1050,3 +1085,617 @@ def odefun(rpz, s):
...
+        potential_dtheta,
+        potential_dzeta,
+        params=None,
+        R_lmn=None,
+        Z_lmn=None,
+        modes_R=None,
+        modes_Z=None,
+        NFP=1,
+        sym="auto",
+        rho=1,
```
f0uriest, line 1156: should also get rid of this

desc/compute/_surface.py
f0uriest: I think we want something like this:
```python
temp_params = {}
for name in params:
    p = getattr(obj, name)
    if isinstance(p, dict):
        temp_params[name] = p
    else:
        temp_params[name] = jnp.atleast_1d(p)
params = temp_params

f0uriest (approved):

unalmis (commented):
approving

PR #598, opened 2023-07-25
Basis vectors

unalmis:
Generalizes basis vectors to include toroidal stream function and adds their third derivatives.

This PR includes the second derivatives of the covariant basis vectors which were generated by Rory's code, and the following quantities that I coded:
- 3rd derivatives of covariant basis vectors (**review this more thoroughly**)
- 1st derivatives of contravariant basis vectors (Already checked for correctness with finite difference of the contravariant elements of the metric tensor)
- Magnetic axis limits (Already checked for correctness in #556 ).

@f0uriest Here is a diff of `_basis_vectors.py` from this branch with the sorted version `_basis_vectors.py` from #568.
https://diffy.org/diff/a1af98d05be0a.
- This diff was generated with the code: `git diff --no-index rc_toroidal_sorted_basis_vectors.txt _basis_vectors.py > output.diff` applied to [rc_toroidal_sorted_basis_vectors.txt](https://github.com/PlasmaControl/DESC/files/12231980/rc_toroidal_sorted_basis_vectors.txt).


tests/test_axis_limits.py
```diff
@@ -1,95 +1,323 @@
...
+    matches = pattern.findall(src)
+    matches = {s.replace("'", "").replace('"', "") for s in matches}
+    return matches
+
 
 class TestAxisLimits:
     """Tests for compute functions evaluated at limits."""
 
+    @pytest.mark.unit
+    def test_data_index_deps(self):
```
unalmis, line 206: This file is included in this PR, so that reviewers can skip checking what this function tests during their review.

desc/compute/_basis_vectors.py
```diff
@@ -9,686 +20,3414 @@
...
+        "R_rrr",
+        "R_rrrr",
+        "Z_rrrr",
+        "omega_r",
+        "omega_rr",
+        "omega_rrr",
+        "omega_rrrr",
+    ],
+)
+def _e_sub_rho_rrr(params, transforms, profiles, data, **kwargs):
```
unalmis, line 419: I added the third derivatives below the second derivatives because it's helpful to see the previous derivative to check for correctness. If git is struggling with the diff this like this maybe I should move it to the bottom then we can tell GitHub to compare with `rc/toroidal_angle`.

desc/compute/_basis_vectors.py
f0uriest, line 1: Similar question to the one on #586, can we keep a consistent sorting order for all the derivatives of the covariant basis vectors?

unalmis: I have sorted them using the script I added here https://github.com/PlasmaControl/DESC/blob/basis_vectors/docs/sort_compute_funs.py

unalmis:
Think GitHub is showing a diff against an old version of master. If it doesn't resolve itself after the fourth order derivatives PR is merged, then I will check what the issue is.

desc/compute/_field.py
```diff
@@ -203,9 +203,9 @@ def _B0_r(params, transforms, profiles, data, **kwargs):
     data=["B0", "B0_r", "iota", "iota_r", "lambda_z", "lambda_rz"],
 )
 def _B_sup_theta_r(params, transforms, profiles, data, **kwargs):
-    data["B^theta_r"] = data["B0_r"] * (data["iota"] - data["lambda_z"]) + data[
-        "B0"
-    ] * (data["iota_r"] - data["lambda_rz"])
+    data["B^theta_r"] = data["B0_r"] * (data["iota"] - data["lambda_z"]) + (
+        data["B0"] * (data["iota_r"] - data["lambda_rz"])
+    )
     return data
```
unalmis, line 209: needed to throw a parenthesis around this because my regular expression in test_axis_limits.py can't detect quantities with brackets spread on multiple lines.

tests/test_optimizer.py
```diff
@@ -917,6 +917,7 @@ def test_constrained_AL_lsq():
 
 @pytest.mark.slow
 @pytest.mark.regression
+@pytest.mark.xfail
```
unalmis, line 920: See https://github.com/PlasmaControl/DESC/pull/556/files#r1271238780

tests/test_examples.py
```diff
@@ -665,7 +665,7 @@ def test_NAE_QIC_solve():
     grid = LinearGrid(L=10, M=20, N=20, NFP=eq.NFP, sym=True, axis=False)
     iota = compress(grid, eq.compute("iota", grid=grid)["iota"], "rho")
 
-    np.testing.assert_allclose(iota[1], qsc.iota, atol=1e-5)
+    np.testing.assert_allclose(iota[1], qsc.iota, atol=5e-4)
```
unalmis, line 668: see https://github.com/PlasmaControl/DESC/pull/556#discussion_r1271470207

f0uriest:
the finite differencing test seems to fail consistently for ``e_zeta_rtt``, you might want to check that one.

desc/compute/_geometry.py
```diff
@@ -194,7 +194,7 @@ def _R0_over_a(params, transforms, profiles, data, **kwargs):
     transforms={"grid": []},
     profiles=[],
     coordinates="",
-    data=["sqrt(g)", "g_tt"],
+    data=["sqrt(g)", "g_tt", "R"],
```
ddudt, line 197: FYI this change is unnecessary because `"sqrt(g)"` already depends on `"R"`. But if we always want to list all explicit dependencies that is fine.

docs/sort_compute_funs.py
ddudt, line 1: This sorts all compute functions by their *function name*? So the compute quantity `"B^theta_r"` would get sorted as `_B_sup_theta_r`? That's fine, just checking my understanding. 

I like the idea of sorting all the compute functions this way, but I think we need to re-sort some of the files.

tests/test_axis_limits.py
```diff
@@ -1,101 +1,324 @@
...
-        # check continuity
-        np.testing.assert_allclose(quantity[:-1], quantity[1:], atol=epsilon)
-
-        # check expected value at axis
-        if expected_at_axis is None:
-            # fit the data (except axis pt) to a polynomial to extrapolate to axis
-            poly = np.polyfit(rho[1:], quantity[1:], 6)
-            expected_at_axis = poly[-1]  # constant term is same as eval poly at rho=0
-        np.testing.assert_allclose(quantity[0], expected_at_axis, atol=epsilon)
+    def test_limit_existence(self):
```
ddudt, line 263: You added this new test but it looks like it is not actually used?

tests/test_compute_funs.py
```diff
@@ -299,94 +308,94 @@ def test_magnetic_field_derivatives(DummyStellarator):
...
         atol=atol * np.nanmean(np.abs(data["B_tt"])),
     )
 
     # partial derivatives wrt zeta
-    rtol = 1e-4
-    atol = 1e-4
-    num_zeta = 130
+    rtol = 1e-3
+    atol = 1e-3
+    num_zeta = 180
```
ddudt, line 398: Is there a reason we need to increase the tolerances, despite also increasing the finite difference resolution?

tests/test_axis_limits.py
unalmis: It is not used since it won't pass. All the limits are implemented in #556.
I pulled this file from #556 so that reviewers know that the `test_data_index_deps` test passes. (Manually checking that there are no extra dependencies in the compute functions for the basis vectors would be a chore).

desc/compute/_bootstrap.py
```diff
@@ -280,8 +278,8 @@ def j_dot_B_Redl(
     J_dot_B = dnds_term + dTeds_term + dTids_term
 
     # Store all results in the J_dot_B_data dictionary:
-    nu_e_star = nu_e
-    nu_i_star = nu_i
+    nu_e_star = nu_e  # noqa: F841
```
dpanici, line 281: if nu_e_star is never used, we can probably get rid of this line and the next line? F841 is local variable assigned but never used

f0uriest: it is used, though the `eval` loop below. I refactored this in #601 to remove that part and simplify this, but for now this is fine.

f0uriest (dismissed):

tests/test_compute_funs.py
ddudt: that's fine for `rtol`, but in general could be dangerous for `atol` since some derivates might be close to zero

docs/sort_compute_funs.py
ddudt: Do we want to sort the other files too as part of this PR?

unalmis: There's a couple PR drafts out that touch compute funs. Would prefer to do later to minimize merge conflicts

f0uriest (approved):

f0uriest (approved):

desc/compute/_basis_vectors.py
```diff
@@ -27,630 +1356,2063 @@ def _e_sub_rho(params, transforms, profiles, data, **kwargs):
...
     profiles=[],
     coordinates="rtz",
-    data=["R", "R_z", "Z_z"],
+    data=["e_theta", "sqrt(g)"],
+    axis_limit_data=["e_theta_r", "sqrt(g)_r"],
 )
-def _e_sub_zeta(params, transforms, profiles, data, **kwargs):
-    data["e_zeta"] = jnp.array([data["R_z"], data["R"], data["Z_z"]]).T
+def _e_sub_theta_over_sqrt_g(params, transforms, profiles, data, **kwargs):
+    data["e_theta/sqrt(g)"] = transforms["grid"].replace_at_axis(
```
dpanici, line 1389: why is this function needed? ah is it due to axis limits?

unalmis: yes

tests/test_compute_funs.py
```diff
@@ -173,94 +177,94 @@ def test_magnetic_field_derivatives(DummyStellarator):
     B_rr = np.apply_along_axis(my_convolve, 0, data["B"], FD_COEF_2_4) / drho**2
 
     np.testing.assert_allclose(
-        data["B^theta_r"][3:-2],
-        B_sup_theta_r[3:-2],
+        data["B^theta_r"][4:-4],
```
dpanici, line 180: why make these further from the bdry? because you increased nrho so want to keep it at the same rho values roughly?

dpanici (approved):

tests/test_compute_funs.py
unalmis: My guess is same reason as https://github.com/PlasmaControl/DESC/pull/598#discussion_r1282271985

PR #600, opened 2023-07-26
Ensure two term objective helicity is passed correctly

f0uriest:


ddudt (changes requested):
Can you also make sure that the helicity is handled properly for both `QuasisymmetryBoozer` and `QuasisymmetryTwoTerm`? 

Should this change you made of adding `"helicity"` to the `constants` dict also be made to the Two-Term class? 

Also the Boozer class has the checks:
```
assert len(helicity) == 2
assert (int(helicity[0]) == helicity[0]) and (int(helicity[1]) == helicity[1])
```
but they are missing from the Two-Term class.

f0uriest:
The checks are in the setter method for both classes, I think the checks in init for the Boozer one are unneeded.

f0uriest:
We do have tests for the Boozer helicity already, `test_qa_boozer` and `test_qh_boozer`

dpanici (approved):

ddudt (approved):

PR #601, opened 2023-07-26
Misc code cleanup and organization

f0uriest:
- Breaks `get_fixed_boundary_constraints` etc into their own file to avoid circular imports with other utils that are used in objective classes
- Refactors bootstrap computation to avoid flake8 warning and remove use of `eval`
- Relabels some tests: `test_perturbation_orders` is now a regression test as it tends to take a long time, while a few of the example tests are now marked as unit tests since the fixture is already used by other unit tests.
- Simplify constraint projection when loading vmec solution.
- Refactors softmin/max to avoid direct JAX imports


desc/compute/_bootstrap.py
```diff
@@ -189,17 +189,15 @@ def j_dot_B_Redl(
...
-            + 0.23 * (1 - 0.96 * f_t) * jnp.sqrt(nu_e) / jnp.sqrt(Zeff)
-            + 0.13
-            * (1 - 0.38 * f_t)
-            * nu_e
-            / (Zeff * Zeff)
-            * (
-                jnp.sqrt(1 + 2 * jnp.sqrt(Zeff - 1))
-                + f_t * f_t * jnp.sqrt((0.075 + 0.25 * (Zeff - 1) ** 2) * nu_e)
-            )
+        1
```
f0uriest, line 192: This was all just autoformatting, no actual change

desc/objectives/__init__.py
```diff
@@ -55,9 +62,3 @@
     FixThetaSFL,
 )
 from .objective_funs import ObjectiveFunction
-from .utils import (
```
ddudt, line 58: Why not include the utils in the init file?

unalmis:
can you move `def _rotation_matrix_from_normal(normal):` in `desc.compute._curve.py` to `geom_utils`. Otherwise it will get lost if you sort with https://github.com/PlasmaControl/DESC/blob/basis_vectors/docs/sort_compute_funs.py

desc/objectives/__init__.py
f0uriest: The stuff left in utils is not meant to be public, so its not part of the module init.

dpanici (approved):

ddudt (approved):

PR #604, opened 2023-07-31
Allow for optimizing multiple objects

### LINKED ISSUE #560: adding param parsing methods to coils and equilibrium

### LINKED ISSUE #561: Update objective parameter parsing to new style

f0uriest:
For this PR:
- [ ] naming conventions for `things` - can probably come up with a better name
- [x] Add tests for constraints for fixing general objects/degrees of freedom
- [x] increase test coverage

For a future PR:
- [ ] Figure out how to handle collections of optimizeable objects like `CoilSets`
- [ ] Right now `ProximalProjection` is still limited to just the equilibrium, we could possibly generalize that?
- [ ] Constraint for sharing parameters between different objects
- [ ] Allow for objectives to be built with an optimizable object that is not an equilibrium (for example, `MeanCurvature` with `FourierRZToroidalSurface`) 
- [ ] @f0uriest new class for OptimizableCollection
- [ ] @dpanici write tests for possible use cases to test current functionality
- [ ] @ddudt Coil optimization -> coil objectives @kianorr 

Resolves #560 
Resolves #561


desc/objectives/objective_funs.py
```diff
@@ -802,81 +716,23 @@ def __init__(
...
-            self,
-            "_args",
-            [arg for arg in getfullargspec(self.compute)[0] if arg != "self"],
-        )
-        self._eq = eq
-        if eq is None:
+        if things is None:
             warnings.warn(
                 FutureWarning(
                     "Creating an Objective without specifying the Equilibrium to"
```
dpanici, line 722: this warning should now read "Optimizable Object" instead of Equilibrium

desc/optimizable.py
```diff
@@ -0,0 +1,155 @@
...
+class Optimizable(ABC):
+    """Base class for all objects in DESC that can be optimized.
+
+    Sub-classes should decorate optimizable attributes with the
+    ``optimizable_parameter`` decorator.
+    """
+
+    @property
+    def optimizable_params(self):
+        """list: names of parameters that have been declared optimizable."""
```
dpanici, line 19: maybe specify in the docstring it is names as strings? I know it kind of obvious but I think better to be verbose

desc/continuation.py
```diff
@@ -729,17 +717,21 @@ def solve_continuation(  # noqa: C901
...
-                    setattr(eqi, arg, val)
+            eqi.params_dict = eqp.params_dict
             deltas = {}
             del eqp
 
         if not eqi.is_nested(msg="manual"):
             stop = True
 
         if not stop:
+            # TODO: add ability to rebind objectives
```
dpanici, line 728: what prevents this currently? does binding an objective not just say which eq to build with?

desc/equilibrium/equilibrium.py
```diff
@@ -167,6 +168,7 @@ def __init__(
         spectral_indexing=None,
         **kwargs,
     ):
+        super().__init__()
```
dpanici, line 171: the init method of `Optimizable` does not seem to do much, what is this call here for?

desc/utils.py
```diff
@@ -483,7 +483,10 @@ def is_broadcastable(shp1, shp2):
 
 def get_instance(things, cls):
     """Get thing from a collection of things that is the instance of a given class."""
```
dpanici, line 485: get "first" thing

desc/optimize/optimizer.py
```diff
@@ -141,6 +145,13 @@ def optimize(  # noqa: C901 - FIXME: simplify this
             `OptimizeResult` for a description of other attributes.
 
         """
+        things = flatten_list(things, flatten_tuple=True)
+
+        # need local import to avoid circular dependencies
+        from desc.equilibrium import Equilibrium
+
+        eq = get_instance(things, Equilibrium)
```
dpanici, line 153: this line makes it seem we ALWAYS require an equilibrium when we want to optimize things. What if we just want to minimize quadratic flux from coils on a surface (not necessarily an equilibirum surface)?

desc/objectives/utils.py
```diff
@@ -41,40 +54,50 @@ def factorize_linear_constraints(constraints, objective_args):  # noqa: C901
         and recovering x from y.
 
     """
+    for con in constraints:
+        for thing in con.things:
+            if thing not in objective.things:
+                warnings.warn(
+                    f"Optimizable object {thing} is constrained"
```
dpanici, line 61: Could be useful to specify which constraint is the culprit, like `f"Optimizable object {thing} is constrained by {con}"`

desc/objectives/utils.py
```diff
@@ -41,40 +54,50 @@ def factorize_linear_constraints(constraints, objective_args):  # noqa: C901
...
+                f"Linear constraint {con} must use target instead of bounds."
             )
-            for arg in args
-        }
+        A_per_thing = []
+        xz = _tree_zeros_like([t.params_dict for t in con.things])
+        # computing A matrix for each constraint for each thing in the optimization
+        for thing in objective.things:
+            if thing in con.things:
+                # for now we implicitly assume that each linear constraint is bound to
```
dpanici, line 84: what would be an example of a generalization? something like constraining the R0 of every coil in a coilset or something?

desc/equilibrium/equilibrium.py
```diff
@@ -1770,24 +1795,10 @@ def optimize(
...
-        if verbose > 0:
-            print("End of solver")
-            objective.print_value(objective.x(eq))
-            for con in constraints:
-                con.print_value(*con.xs(eq))
-
-        return eq, result
+        return things[0], result
 
     def _optimize(  # noqa: C901
```
dpanici, line 1803: why do we still carry this old optimize method around?

desc/objectives/_generic.py
```diff
@@ -144,35 +146,32 @@ def get_vars(fun):
...
 
-    def compute(self, *args, **kwargs):
+    def compute(self, params, constants=None):
         """Compute the quantity.
 
         Parameters
         ----------
-        args : ndarray
-            Parameters given by self.args.
+        params : dict
```
dpanici, line 163: does it have to be Equilibrium DOFs?

desc/optimizable.py
```diff
@@ -0,0 +1,155 @@
...
+            params[arg] = jnp.atleast_1d(x[x_idx[arg]])
+        return params
+
+    def _sort_args(self, args):
+        """Put arguments in a canonical order. Returns unique sorted elements.
+
+        Actual order doesn't really matter as long as its consistent, though subclasses
+        may override this method to enforce a specific ordering
+        """
+        return sorted(set(list(args)))
```
ddudt, line 122: Does this order really not matter? Because `self.x_idx` depends on this order, so subclasses have to make sure they override that too.

desc/objectives/utils.py
ddudt: It would also be really useful if we could print out what the free variables are when running an optimization (so the opposite of this). That way you can check to make sure you included all the constraints you meant to.

desc/objectives/utils.py
```diff
@@ -121,62 +144,28 @@ def project(x):
...
+            atol=1e-14,
+            rtol=1e-14,
             err_msg="Incompatible constraints detected, cannot satisfy "
             + f"constraint {con}",
         )
 
     return xp, A_full, b_full, Z, unfixed_idx, project, recover
 
 
-def align_jacobian(Fx, objective_f, objective_g):
```
ddudt, line 148: Why is this no longer needed?

desc/optimize/optimizer.py
ddudt: It looks like this logic is to handle the case of proximal projections, and I think your example of not needing an equilibrium would still work fine. But I agree it seems misleading.

desc/objectives/_geometry.py
```diff
@@ -538,37 +526,37 @@ def build(self, eq=None, use_jit=True, verbose=1):
...
         )
         self._constants = {
-            "transforms": transforms,
-            "profiles": profiles,
-            "surface_coords": self._surface_coords,
+            "equil_transforms": equil_transforms,
+            "equil_profiles": equil_profiles,
+            "surface_transforms": surface_transforms,
+            "quad_weights": surface_grid.weights * jnp.sqrt(surface_grid.num_nodes),
         }
```
ddudt, line 560: Do we have a list somewhere in the code or docs of what all the allowed constants are?

desc/perturbations.py
```diff
@@ -760,7 +729,7 @@ def optimal_perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
         if verbose > 0:
             print("Computing d^2g")
         timer.start("d^2g computation")
-        tangents_g = (dxdx_reduced @ dx1_reduced + dxdc @ dc1) @ dxf_dxg
+        tangents_g = dxdx_reduced @ dx1_reduced + dxdc @ dc1
```
ddudt, line 732: What happened to `dxf_dxg`?

desc/objectives/linear_objectives.py
```diff
@@ -39,6 +40,223 @@ def update_target(self, eq):
...
+        -------
+        f : ndarray
+            Fixed degree of freedom errors.
+
+        """
+        return jnp.concatenate([params[par] for par in self._params])
+
+
+class ShareParameter(_Objective):
+    """Require that params from two or more Optimizable objects share the same values.
```
ddudt, line 152: Clarifying question:
So by default, every optimizable "thing" has its own copy of optimizable parameters, right? Like if I am optimizing two coils, they are completely independent. But then, for example, I could use this objective to say they have to have the same position, while letting their other degrees of freedom (tilt, rotation, etc.) be independent?

desc/optimize/_constraint_wrappers.py
```diff
@@ -142,34 +126,38 @@ def recover(self, x_reduced):
...
+    def unpack_state(self, x, per_objective=True):
         """Unpack the state vector into its components.
 
         Parameters
         ----------
         x : ndarray
             State vector, either full or reduced.
+        per_objective : bool
+            Whether to return param dicts for each objective (default) or for each
+            unique optimizable thing.
```
ddudt, line 143: What is the use case for each option?

desc/optimize/_constraint_wrappers.py
```diff
@@ -569,32 +493,76 @@ def build(self, eq=None, use_jit=None, verbose=1):  # noqa: C901
...
-            self._x_old[self.x_idx[arg]] = getattr(eq, arg)
-
+        self._x_old = self.x(eq)
         self._allx = [self._x_old]
         self._allxopt = [self._objective.x(eq)]
         self._allxeq = [self._constraint.x(eq)]
-        self.history = {}
-        for arg in arg_order:
-            self.history[arg] = [np.asarray(getattr(self._eq, arg)).copy()]
+        self.history = [[self._eq.params_dict]]
```
ddudt, line 511: OK so before `history` was a dict of lists, and now it is a list of dicts?

desc/optimize/_constraint_wrappers.py
```diff
@@ -569,32 +493,76 @@ def build(self, eq=None, use_jit=None, verbose=1):  # noqa: C901
...
+
+        x = jnp.atleast_1d(x)
+        if x.size != self.dim_x:
+            raise ValueError(
+                "Input vector dimension is invalid, expected "
+                + f"{self.dim_x} got {x.size}."
+            )
+
+        xs_splits = np.cumsum([self._eq.dimensions[arg] for arg in self._args])
+        # kludge for now if things is only equilibrium
```
ddudt, line 547: Is this going to be a problem in the future?

desc/optimize/_constraint_wrappers.py
```diff
@@ -569,32 +493,76 @@ def build(self, eq=None, use_jit=None, verbose=1):  # noqa: C901
...
+        params = [{arg: xs for arg, xs in zip(self._args, jnp.split(x, xs_splits))}]
+        if per_objective:
+            params = self._unflatten(params)
+        return params
+
+    def x(self, *things):
+        """Return the full state vector from the Optimizable objects things."""
+        # TODO: also check resolution etc?
+        assert [type(t1) == type(t2) for t1, t2 in zip(things, self.things)]
```
ddudt, line 556: Why do we have to check against `self.things`, just as a safety measure? And why not have the input `*things` default to `self.things`?

desc/optimize/_constraint_wrappers.py
```diff
@@ -800,8 +757,6 @@ def jac_scaled(self, x, constants=None):
         # Jacobian matrices wrt combined state vectors
         Fx = self._constraint.jac_scaled(xf, constants[1])
         Gx = self._objective.jac_scaled(xg, constants[0])
-        Fx = align_jacobian(Fx, self._constraint, self._objective)
-        Gx = align_jacobian(Gx, self._objective, self._constraint)
```
ddudt, line 804: I don't understand how we are ensuring these Jacobians are still aligned in the new code

desc/objectives/_geometry.py
dpanici: this is probably something specific to each objective

desc/objectives/_geometry.py
```diff
@@ -523,7 +511,7 @@ def build(self, eq=None, use_jit=True, verbose=1):
             Level of output.
 
         """
-        eq = eq or self._eq
+        eq = self.things[0]
```
dpanici, line 514: Is this enforced somewhere or assumed?

desc/objectives/_profiles.py
```diff
@@ -129,18 +125,25 @@ def build(self, eq=None, use_jit=True, verbose=1):
...
+        constants : dict
+            Dictionary of constant data, eg transforms, profiles etc. Defaults to
+            self.constants
 
         Returns
         -------
-        pressure : ndarray
-            Plasma pressure at specified points.
+        f : ndarray
+            Computed quantity.
```
dpanici, line 144: is this not the pressure? why change the docstring?

desc/objectives/_profiles.py
```diff
@@ -268,33 +267,25 @@ def build(self, eq=None, use_jit=True, verbose=1):
...
+        constants : dict
+            Dictionary of constant data, eg transforms, profiles etc. Defaults to
+            self.constants
 
         Returns
         -------
-        iota : ndarray
-            Rotational transform on specified flux surfaces.
+        f : ndarray
+            Computed quantity.
```
dpanici, line 286: is this not the rotational transform?

desc/objectives/linear_objectives.py
```diff
@@ -39,6 +40,223 @@ def update_target(self, eq):
...
+
+    Parameters
+    ----------
+    things : Optimizable
+        Object whose degrees of freedom are being fixed.
+    params : str or list of str
+        Names of parameters to fix. Defaults to all parameters.
+    target : dict of {float, ndarray}, optional
+        Target value(s) of the objective. Only used if bounds is None.
+        Should have the same tree structure as thing.params. Defaults to things.params.
```
dpanici, line 55: `thing.params` or `things.params`? you mean `things.params` right?

desc/objectives/linear_objectives.py
```diff
@@ -39,6 +40,223 @@ def update_target(self, eq):
...
+    bounds : tuple of dict {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
+        Should have the same tree structure as thing.params.
```
dpanici, line 58: or if it is meant to be `thing.params`, which of the `things` is it referring to?

desc/objectives/linear_objectives.py
```diff
@@ -39,6 +40,223 @@ def update_target(self, eq):
...
+        Parameters
+        ----------
+        use_jit : bool, optional
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        self.things = setdefault(thing, self.things)
+        thing = self.things[0]
```
dpanici, line 116: why is it always the first element of things?

desc/objectives/linear_objectives.py
```diff
@@ -39,6 +40,223 @@ def update_target(self, eq):
...
+
+    _scalar = False
+    _linear = True
+    _fixes = True
+    _units = "(~)"
+    _print_value_fmt = "Fixed parameter error: {:10.3e} "
+
+    def __init__(
+        self,
+        thing=None,
```
dpanici, line 189: this is thing, but it is meant to be things right?

desc/objectives/linear_objectives.py
```diff
@@ -39,6 +40,223 @@ def update_target(self, eq):
...
+        Returns
+        -------
+        f : ndarray
+            Fixed degree of freedom errors.
+
+        """
+        return jnp.concatenate([params[par] for par in self._params])
+
+
+class ShareParameter(_Objective):
```
dpanici, line 151: Make sure this works as intended and for multiple objects

desc/objectives/objective_funs.py
```diff
@@ -1099,28 +1039,10 @@ def print_value(self, *args, **kwargs):
                     + "(normalized)"
                 )
 
-    def xs(self, eq):
+    def xs(self, *things):
         """Return a tuple of args required by this objective from the Equilibrium eq."""
```
dpanici, line 1043: not necessarily an Equilibirum

dpanici:
https://github.com/PlasmaControl/DESC/blob/be26659f1fb80bf5f9ada8a62e3da1096bab29d3/desc/optimize/optimizer.py#L364
To not re-wrap an already-existing nonlinear constraint that is already wrapped by an ObjectiveFunction, change lines here

desc/equilibrium/equilibrium.py
```diff
@@ -1671,19 +1711,7 @@ def solve(
...
-            # don't set nonexistent profile (values are empty ndarrays)
-            if value[-1].size:
-                setattr(eq, key, value[-1])
-
-        if verbose > 0:
-            print("End of solver")
-            objective.print_value(objective.x(eq))
-
-        return eq, result
+        return things[0], result
```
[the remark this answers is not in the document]
f0uriest: exactly

desc/equilibrium/equilibrium.py
f0uriest: we have an automated test for it for a basic case (solovev with target aspect ratio), we could look at getting rid of it in a future version, it would still be available in older versions that are referenced in the paper

desc/objectives/_generic.py
f0uriest: for now, this objective still assumes you're optimizing an equilibrium. We could probably generalize it somehow?

desc/objectives/_geometry.py
f0uriest: yeah, each objective can define their own constants, there's no master list

desc/objectives/_profiles.py
f0uriest: I just copy and pasted the docstring from the other objectives

desc/objectives/linear_objectives.py
f0uriest: That's the idea yes

desc/objectives/linear_objectives.py
f0uriest: This objective should only ever be used with a single optimizable object

desc/objectives/linear_objectives.py
f0uriest: yeah but this is only used with 1 thing at a time so I should change that.

desc/objectives/utils.py
f0uriest: yeah, or the parameter sharing thing we discussed. Still need to think about how that will work but can probably punt to the next PR

desc/objectives/utils.py
f0uriest: Every objective now takes the full set of arguments, so its jacobian will automatically have zeros in the right places.

desc/optimize/_constraint_wrappers.py
f0uriest: basically depends what you're iterating over. IE, for most optimization stuff you want to get the params that will be passed to each objective. But at the end when you're assigning the optimized parameters, you want a single dict for each optimizable object.

desc/optimize/_constraint_wrappers.py
f0uriest: yeah. The outer list is iteration, the inner list is futureproofing for having multiple objects being optimized

desc/optimize/_constraint_wrappers.py
f0uriest: possibly. I haven't really thought of how best to generalize this stuff.

desc/optimize/_constraint_wrappers.py
f0uriest: yeah its a safety check. Ideally we'd also check specific resolution etc but that gets messy. I'll make it default to self.things (same as for the regular ObjectiveFunction)

desc/optimize/_constraint_wrappers.py
f0uriest: Both the objective and constraint are bound to the same things in `Optimizer.optimize`, and they always take the full set of optimizable params as args. The ordering of the parameters is deterministic, so they should always line up.

desc/perturbations.py
f0uriest: That was basically taking the place of `align_jacobian` which is no longer needed since both objectives take the same args.

desc/optimizable.py
```diff
@@ -0,0 +1,155 @@
...
+            key: jnp.asarray(getattr(self, key)).size for key in self.optimizable_params
+        }
+
+    @property
+    def x_idx(self):
+        """dict: dictionary of arrays of indices into array for each parameter."""
+        dimensions = self.dimensions
+        idx = {}
+        dim_x = 0
+        for arg in self.optimizable_params:
```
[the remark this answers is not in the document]
f0uriest: I don't think it has to be sorted accordingly to any particular rule, as long as the order is consistent, though in practice we do sort alphabetically.

desc/optimizable.py
f0uriest: this is really only used once when optimizable_params is first created, after that everything uses the ordering of the optimizable_params list

desc/equilibrium/equilibrium.py
f0uriest: I think this was left over from another change, will remove.

desc/optimize/optimizer.py
f0uriest: I'll add a comment to clarify

desc/objectives/utils.py
f0uriest: @dpanici good idea. 
@ddudt that might get messy, since often there are many free variables. did you have anything particular in mind?

tests/test_objective_funs.py
```diff
@@ -410,37 +402,25 @@ def test(eq):
 
 @pytest.mark.unit
 def test_derivative_modes():
-    """Test equality of derivatives using batched, blocked, looped methods."""
+    """Test equality of derivatives using batched, looped methods."""
```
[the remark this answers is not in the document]
f0uriest: Because all the sub-objectives now take dicts of params instead of the params themselves it got really messy/wasteful trying to get the individual jacobians for each arg. In practice the blocked method was never really used since it had greater memory usage than looped but was slower than batched.

tests/test_objective_funs.py
```diff
@@ -319,19 +311,19 @@ def test(eq):
         # precise_QA should have lower QA than QH
         obj = QuasisymmetryTwoTerm(eq=eq1, helicity=helicity_QA)
         obj.build()
-        f1 = obj.compute_scalar(*obj.xs(eq1))
+        f1 = obj.compute_scalar(*obj.xs(eq1), constants=obj.constants)
```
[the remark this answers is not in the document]
f0uriest: In theory either is fine. Passing in constants should in theory lead to faster compiles since it doesn't try to bake the constants into the compiled code, though in practice who knows.

desc/objectives/_geometry.py
f0uriest: Enforced when we call `super().__init__` and specify `things`

desc/objectives/linear_objectives.py
```diff
@@ -39,6 +40,137 @@ def update_target(self, eq):
...
+        # replace indices=True with actual indices
+        if self._indices and isinstance(self._indices, bool):
+            self._indices = [np.arange(thing.dimensions[par]) for par in self._params]
+        # make sure its iterable if only a scalar was passed in
+        if not isinstance(self._indices, (list, tuple)):
+            self._indices = [self._indices]
+        # replace idx=True with array of all indices, throwing an error if the length
+        # of indices is different from number of params
+        indices = {}
+        for idx, par in zip(self._indices, self._params, strict=True):
```
dpanici, line 137: `strict` is only a keyword of `zip` in python 3.10 or later: https://peps.python.org/pep-0618/

so this would increase our minimum supported python version from 3.8 to 3.10

desc/objectives/linear_objectives.py
```diff
@@ -39,6 +40,137 @@ def update_target(self, eq):
...
+            if idx and isinstance(idx, bool):
```
dpanici, line 138: if idx is not a bool (for example, if it is an array), this statement throws an error because it can't figure out how to apply `and` between the bool on the right and the array on the left. I got this error when trying to do `FixParameter(eq,params="L_lmn")` to try to fix lambda during an optimization

f0uriest: OK, i think just flipping the order of the operands should fix that

dpanici:
I would like to be able to make objectives without an `eq` necessarily as an argument. I  have in mind trying to optimize a surface to be a certain distance from an eq without changing the eq, only changing the surface, while also having a `MeanCurvature` target for the surface. Right now, it seems this is not possible because most objectives still expect eq as their first thing, and the normalizations expect an `Equilibrium` object to calculate normalizations with. From what I see right now we would need to
 - have each objective only calculate the normalizations it needs, and allow the normalizations to be calculated from things other than `Equilibrium`
 - generalize the required `things` argument for the objectives which it makes sense for (lots of the geometry objectives could feasibly be used with just a surface, instead of an equilibrium, for example)

 Is this something we want to change in this PR or is that too large of a change?

f0uriest:
> I would like to be able to make objectives without an `eq` necessarily as an argument. I have in mind trying to optimize a surface to be a certain distance from an eq without changing the eq, only changing the surface, while also having a `MeanCurvature` target for the surface. Right now, it seems this is not possible because most objectives still expect eq as their first thing, and the normalizations expect an `Equilibrium` object to calculate normalizations with. From what I see right now we would need to
> 
>     * have each objective only calculate the normalizations it needs, and allow the normalizations to be calculated from things other than `Equilibrium`
> 
>     * generalize the required `things` argument for the objectives which it makes sense for (lots of the geometry objectives could feasibly be used with just a surface, instead of an equilibrium, for example)
> 
> 
> Is this something we want to change in this PR or is that too large of a change?
> 
> this actually is not a big change at all, with `normalize=False` and just changing the build method of `MeanCurvature` to use the `.M` attribute of the passed-in thing if `M_grid` is not present for the default grid, this seems to work fine. I won't have time to make this change today but I think we should make this small generalization

I agree this would be a good idea but I think probably better to do a separate PR, since ideally we'd want to do it for all of the geometry objectives (and possible generic objectives as well) and figure out normalizations etc.

desc/perturbations.py
```diff
@@ -205,81 +202,75 @@ def perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
...
+    xz = objective.unpack_state(jnp.zeros_like(x), False)[0]
     # tangent vectors
-    tangents = jnp.zeros((objective.dim_x,))
+    tangents = jnp.zeros((eq.dim_x,))
     if "Rb_lmn" in deltas.keys():
         con = get_instance(constraints, BoundaryRSelfConsistency)
-        A = con.derivatives["jac_unscaled"]["R_lmn"](
-            *[jnp.zeros(con.dimensions[arg]) for arg in con.args]
-        )
+        A = con.jac_unscaled(xz)["R_lmn"]
```
rahulgaur104, line 210: In terms of your perturbation-continuation paper, is A = df/dR_lmn? Similarly, what is "tangents"?

desc/perturbations.py
```diff
@@ -610,38 +574,29 @@ def optimal_perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
...
+    dxdc = []
+    xz = objective_f.unpack_state(jnp.zeros_like(xf), False)[0]
+
+    for arg in deltas:
+        if arg not in ["Rb_lmn", "Zb_lmn"]:
+            x_idx = eq.x_idx[arg]
+            dxdc.append(jnp.eye(eq.dim_x)[:, x_idx])
+        if arg == "Rb_lmn":
+            con = get_instance(constraints, BoundaryRSelfConsistency)
+            A = con.jac_unscaled(xz)["R_lmn"]
```
rahulgaur104, line 589: So basically, you get the derivative df/dx and the constraints function applies the constraints somehow.
How does this work?

desc/perturbations.py
```diff
@@ -664,7 +618,6 @@ def optimal_perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
             print("Computing dg")
```
rahulgaur104, line 618: Is dg also needed when I only have one objective like "force" or "current"? It thought dg was a set of objectives other than the force error df

rahulgaur104:
I am trying to understand the purpose of this branch. You seem to be trying to collect all the optimizable quantities, such as R_lmn, p_l, T_i, etc, into "things." This would make it easier for the devs to access optimizable quantities. Is it possible to add more optimizable quantities in the future, say from objectives other than force balance or current? 
What is "things" exactly? An object of a class, list, dict ?

desc/equilibrium/equilibrium.py
```diff
@@ -1866,21 +1878,22 @@ def _optimize(  # noqa: C901
...
         tr_ratio = perturb_options.get(
             "tr_ratio",
             inspect.signature(optimal_perturb).parameters["tr_ratio"].default,
         )
 
         if verbose > 0:
-            objective.print_value(objective.x(eq))
+            objective.print_value(objective.x(self))
+
+        params = orig_params = self.params_dict.copy()
```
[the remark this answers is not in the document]
f0uriest: this is just multiple assignment, it's equivalent to 
```python
params = self.params_dict.copy()
orig_params = self.params_dict.copy()
```

`params` will get modified during the optimization but we also keep track of the original params in case the optimization fails, or if the user doesn't want the equilibrium to be modified in place.

desc/perturbations.py
f0uriest: Not quite. This `A` is the jacobian of the constraint that enforces consistency between the Fourier-Zernike basis in the volume and the double Fourier basis on the boundary. That constraint can be written as `A*R_lmn = Rb_mn`. When we perturb the boundary, we need to translate that into a perturbation of the volume. In the paper notation we have $\partial f/\partial c \Delta c$ where c are the things we're perturbing, and that's basically what we're doing here.

$$\frac{\partial f}{\partial c}\Delta c = \frac{\partial f}{\partial R_{lmn}}\frac{\partial R_{lmn}}{\partial Rb_{mn}}\Delta c = \frac{\partial f}{\partial R_{lmn}}A \Delta c $$

`tangents` is basically just $A \Delta c$, which is a tangent vector in the parlance of differential geometry and JAX

desc/perturbations.py
f0uriest: Yes, there are two different forms of a perturbation. In the first, you fix ahead of time what the changes in the parameters $\Delta c$ are, in which case you only have a single objective (usually force or current). This is what's done in `perturb`

The other one we choose $\Delta c$ to minimize a second objective `g`. This is what's done in `optimal_perturb`. It's basically one step of a constrained optimization loop.

f0uriest:
> I am trying to understand the purpose of this branch. You are collecting all the optimizable quantities, such as R_lmn, p_l, T_i, etc, into "things." This would make it easier for the devs to access optimizable quantities. Is it possible to add more optimizable quantities in the future, say from objectives other than force balance or current? What is "things" exactly? An object of a class, list, dict ?

This PR is basically about changing how we keep track of and organize all the variables that eventually get passed to the optimizer. Previously, everything we wanted to optimize was an attribute of an `Equilibrium`, and we had a hard-coded list of those attributes: https://github.com/PlasmaControl/DESC/blob/b6ede28b7717ed6bb85789f2e3d558a778e6126c/desc/compute/utils.py#L1279

This list also defined how we order the parameters when we put them all in a single array to pass to the optimizer.

The big limitation of this is it doesn't generalize well to doing stuff like coil optimization, winding surface stuff, omnigenity with target fields etc.

So now we introduce a new base class `Optimizable` that tells the code "Hey, this object is something we might want to optimize, here are its degrees of freedom (decorated with `@optimizable_parameter`) and some methods for getting/setting/stacking/unstacking those parameters"

Each objective is bound to one or more `Optimizable` objects. For most objectives, this is just a single `Equilibrium`, but for example, `PlasmaVesselDistance` connects an `Equilibrium` to a `Surface`. Similarly, future work will add stuff like `QuadraticFlux` for connecting an `Equilibrium` or `Surface` to external coils, or the new omnigenity stuff where we have both an `Equilibrium` and a target `OmnigeneousField`. These different objects are what's stored in `things`. Each sub-objective (ie, force, qs, aspect ratio etc) knows about what things it is connected to, and the overall objective (`ObjectiveFunction`) collects all those things into a unique set. The `things` attribute is generally a collection (list, tuple, or nested list, tuple) of `Optimizable` objects, each of which in turn has specific optimizable parameters like `R_lmn`, `p_l`, etc.

The key point is that there's no longer some single master list of all the possible parameters we might want to optimize. Instead, each `Optimizable` object can declare it's own parameters and their sizes, types, meanings etc, without the rest of the code needing to know about it. When we're optimizing stuff, the `ObjectiveFunction` doesn't need to know all the individual parameters that are used, just which `Optimizable` things are involved. Each thing then is in charge of knowing what it's parameters are and how to convert them into a single array that the optimizer sees.

desc/continuation.py
```diff
@@ -105,6 +98,13 @@ def _solve_axisym(
             deltas = get_deltas({"surface": surf_i}, {"surface": surf_i2})
             surf_i = surf_i2
 
+        constraints_i = get_fixed_boundary_constraints(
```
dpanici, line 101: why was this moved into the loop? because now it matters which object is attacehd to the objectives, so we have to rebuild each time to ensure that the correct equilibrium's parameters are being changed when eqi.solve or perturb is called?

f0uriest: yes exactly. Ideally we'll add back in the ability to "re-use" objectives that have already been compiled but this PR was getting long so figured that would wait for another one.

desc/coils.py
```diff
@@ -42,6 +43,7 @@ def __init__(self, current, *args, **kwargs):
         self._current = current
```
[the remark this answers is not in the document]
f0uriest: yeah good catch

desc/objectives/objective_funs.py
```diff
@@ -740,14 +663,55 @@ def weights(self):
...
+            "Cannot unambiguously parse Optimizable objects to individual Objectives, "
+            + f"expected types {expected_types} but got types {got_types}. "
+            + "Try setting Objective.things on each sub Objective individually then "
+            + "call ObjectiveFunction.build().",
+        )
+        # now we know that new and self.things contains instances of unique classes, so
+        # we should be able to just replace like with like
+        things = self._unflatten(new)
+        for obj, t in zip(self.objectives, things[: -len(self._extra_things)]):
+            obj.things = t
```
dpanici, line 705: if things is changed, if any of them have differing resolutions from the original things, the objectivefunction must  be rebuilt. I think the setter should set the `self._built` flag to `False` if things results in a change (or even just whenever things is called) to enforce this and try to pre-empt dimension mismatches due to things changing resolution but the objective not being rebuilt

f0uriest: sounds good

desc/objectives/linear_objectives.py
```diff
@@ -38,6 +39,147 @@ def update_target(self, eq):
...
+        values. If `normalize` is `True` and the target is in physical units,
+        this should also be set to True. Has no effect for this objective.
+    name : str, optional
+        Name of the objective function.
+
+    """
+
+    _scalar = False
+    _linear = True
+    _fixes = True
```
dpanici, line 77: typo, this should be _fixed

desc/objectives/objective_funs.py
```diff
@@ -272,23 +172,65 @@ def build(self, eq=None, use_jit=None, verbose=1):
...
+        from unique flattened version.
+        """
+        from jax.tree_util import tree_flatten, tree_unflatten
+
+        things = setdefault(things, [obj.things for obj in self.objectives])
+
+        flat_, treedef_ = tree_flatten(
+            things, is_leaf=lambda x: isinstance(x, Optimizable)
+        )
+        unique_, inds_ = unique_list(flat_)
```
dpanici, line 215: question to make sure I understand: our classes are also JAX pytrees right?

so when `flat_` comes out, does it contain the attributes which make up each class, and then this line gets only the unique ones, so we know what the the unique list of parameters we need to recreate the `things` are?

or is `flat_` the underlying array values themselves?

In the[ JAX pytrees page](https://jax.readthedocs.io/en/latest/pytrees.html) it seems like flat would just return the array values, but then I am unsure of how this unique_list helps here

dpanici: Ah ok wait I see, it is actually the objects themselves

```python
Building objective: plasma-vessel distance
Precomputing transforms
treedef_=PyTreeDef([[*, *]])
before unique: flat_=[Equilibrium at 0x7fa93414a710 (L=24, M=12, N=4, NFP=12.0, sym=1, spectral_indexing=fringe), FourierRZToroidalSurface at 0x7fa934297190 (name=)]
after  unique: flat_=[Equilibrium at 0x7fa93414a710 (L=24, M=12, N=4, NFP=12.0, sym=1, spectral_indexing=fringe), FourierRZToroidalSurface at 0x7fa934297190 (name=)]
```

desc/objectives/objective_funs.py
```diff
@@ -413,24 +349,29 @@ def print_value(self, x, constants=None):
...
+            Whether to return param dicts for each objective (default) or for each
+            unique optimizable thing.
 
         Returns
         -------
-        kwargs : dict
-            Dictionary of the state components with argument names as keys.
-
+        params : pytree of dict
+            if per_objective is True, this is a nested list of of parameters for each
```
dpanici, line 371: "of of" typo here

desc/objectives/objective_funs.py
```diff
@@ -740,14 +663,57 @@ def weights(self):
             [jnp.ones(obj.dim_f) * obj.weight for obj in self.objectives]
         )
 
+    @property
+    def _all_things(self):
+        """list: all things known to this objective, used and unused."""
+        if not hasattr(self, "_extra_things"):
+            self._extra_things = []
+        return [obj.things for obj in self.objectives] + self._extra_things
```
dpanici, line 671: extra_things just refers to unused things by this objective right?

desc/objectives/objective_funs.py
```diff
@@ -809,81 +774,16 @@ def __init__(
...
-                    arg
-                ] = lambda *args, arg=arg, **kwargs: jnp.zeros(
-                    (1, self.dimensions[arg])
-                )
-                self._derivatives["hess"][
-                    arg
-                ] = lambda *args, arg=arg, **kwargs: jnp.zeros(
-                    (self.dimensions[arg], self.dimensions[arg])
-                )
+        self._grad = Derivative(self.compute_scalar, mode="grad")
```
dpanici, line 781: we can do this without specifying arg indices now because all of the compute functiosn take in `params`?

desc/objectives/objective_funs.py
f0uriest: yeah, the `is_leaf` function stops the flattening when it encounters an `Optimzable` object

f0uriest: The `unique_list` part is to handle the possibility of the same object being in the list multiple times (ie, because the object is connected to multiple objectives)

desc/objectives/objective_funs.py
f0uriest: yes, for example in a constrained problem the objective and constraints may depend on different objects, but we want both to accept the same set of arguments. This is effectively the newer version of the old `align_jacobian` function that @ddudt was asking about

desc/objectives/objective_funs.py
f0uriest: Exactly. Though for objectives that connect multiple objects we might need to modify this. These are currently unused so not an issue for now, but I'll fix it when working on #666

dpanici: objectives like `PlasmaVesselDistance`? or others

desc/objectives/objective_funs.py
dpanici: Got it, like a `PlasmaVesselDistance` and a QS touching the same `Equilibrium` or something

desc/objectives/utils.py
```diff
@@ -41,40 +54,50 @@ def factorize_linear_constraints(constraints, objective_args):  # noqa: C901
...
-        }
+        A_per_thing = []
+        xz = _tree_zeros_like([t.params_dict for t in con.things])
+        # computing A matrix for each constraint for each thing in the optimization
+        for thing in objective.things:
+            if thing in con.things:
+                # for now we implicitly assume that each linear constraint is bound to
+                # only 1  thing, to generalize we need to make jac_scaled work for all
+                # positional args not just the first one.
+                A_ = con.jac_scaled(*xz)
```
dpanici, line 87: another question, where is the implicit assumption here?

dpanici (dismissed):
I think it looks good, I cannot think of any other tests I want to do with it that don't involve some further functionality, so I am happy to merge into master once @ddudt also reviews

desc/objectives/objective_funs.py
f0uriest: For now yes thats the only one that connects two objectives

desc/objectives/utils.py
f0uriest: basically that `jac_scaled` only takes the derivative wrt the first arg, which is the params of the first "thing"

dpanici (approved):

PR #605, opened 2023-08-01
Fix Benchmarks

dpanici:


dpanici:
force merged this because it only affects the benchmarks and not any of the code, and I need to have the changes in Master to test that the `run_benchmarks` label works

PR #610, opened 2023-08-03
change field line integration to use diffrax

### LINKED ISSUE #609: Use Diffrax for field line integration
When `field_line_integrate` is used with a `MagneticField` with `extrap=False`, the integration blows up if the field line leaves the pre-defined domain. In these cases, `MagneticField.compute_magnetic_field` returns `NaN` and the integration should stop. 

A possible resolution is to use the package [Diffrax](https://github.com/patrick-kidger/diffrax) for the integration. This would allow us to create an event to stop the integration when the magnetic field is no longer defined.

dpanici:
Updated comment by @f0uriest 

- `field_line_integrate` now uses `diffrax` instead of `jax.experimental.ode.odeint` which has been soft deprecated.
- Allows better control of bounding box (hard stop of field lines rather than ad-hoc expoential decay)
- In practice seems to be a bit faster. `test_plot_poincare` which calls `field_line_integrate` under the hood went from ~65s to ~55s with these changes.

Resolves #609


requirements.txt
```diff
@@ -9,3 +9,4 @@ nvgpu
 psutil
 scipy >= 1.5.0, < 1.11.0
 termcolor
+diffrax<=0.4.0
```
ddudt, line 12: Keep the requirements list in alphabetical order. 

Maybe we could avoid the runtime warnings by changing the version of `diffrax`?

desc/magnetic_fields.py
```diff
@@ -767,13 +785,36 @@ def field_line_integrate(
...
         maximum number of steps between different phis
+    solver: diffrax.Solver
+        diffrax Solver object to use in integration,
+        defaults to Tsit5(), a RK45 explicit solver
+    terminating_event_fxn: fxn
+        Function which takes as input the state of the ODE solution at each timestep and
+        outputs a bool which, if True, terminates the solve at that timestep.
+        NOTE: If the solve is terminated early, the output returned is still
+        length(phis), however all values from the step point the fxn evaluated
+        to True and on will be inf
```
ddudt, line 796: Would it be better to return NaN instead of Inf for those values? Maybe it doesn't make a difference for plotting.

ddudt (changes requested):

desc/magnetic_fields.py
dpanici: diffrax default is NaN, I could do that yea

desc/magnetic_fields.py
```diff
@@ -782,16 +823,36 @@ def field_line_integrate(
...
         br, bp, bz = field.compute_magnetic_field(rpz, params, basis="rpz").T
         return jnp.array(
             [r * br / bp * jnp.sign(bp), jnp.sign(bp), r * bz / bp * jnp.sign(bp)]
         ).squeeze()
 
-    intfun = lambda x: odeint(odefun, x, phis, rtol=rtol, atol=atol, mxstep=maxstep)
+    # diffrax parameters
+    stepsize_controller = PIDController(rtol=rtol, atol=atol)
+    terminating_event = (
+        DiscreteTerminatingEvent(terminating_event) if terminating_event else None
```
ddudt, line 837: It would be nice if we could make the default terminating event to stop if the magnetic field is undefined. Like if `field.compute_magnetic_field(rpz, params, basis="rpz")` returns NaNs because the integration has left the field's domain. 

That was my desired use case for this PR, but I'm not sure how to implement it.

dpanici: wasted the whole day trying to figure this out... hopefully the diffrax people can tell me https://github.com/patrick-kidger/diffrax/issues/290

dpanici:
The RuntimeWarning for numpy header size being different is still an issue, I think because the warning is coming from C level code (@f0uriest any ideas on how to ignore those?), right now I have in a blanket ignore RuntimeWarning but that is not really what we want, I opened an issue to see if pytest knows how to resolve it https://github.com/pytest-dev/pytest/issues/11304

desc/magnetic_fields.py
ddudt: Sorry... but thanks for looking into it!

f0uriest:
it looks like there are older versions of diffrax that work with older versions of jax. They might not have the discrete terminating event stuff but might still be useful for being able to select different integration schemes etc.

dpanici:
exp. decrease RHS  outside the bounding box (exp(-r)*B)

dpanici:
Alternatively try `solve_ivp` in JAX with Dormand Prince integration method

desc/magnetic_fields.py
```diff
@@ -1271,18 +1342,89 @@ def field_line_integrate(
...
         rpz = rpz.reshape((3, -1)).T
         r = rpz[:, 0]
         br, bp, bz = field.compute_magnetic_field(rpz, params, basis="rpz", grid=grid).T
         return jnp.array(
             [r * br / bp * jnp.sign(bp), jnp.sign(bp), r * bz / bp * jnp.sign(bp)]
         ).squeeze()
 
-    intfun = lambda x: odeint(odefun, x, phis, rtol=rtol, atol=atol, mxstep=maxstep)
+    # diffrax parameters
+    stepsize_controller = PIDController(rtol=rtol, atol=atol, dtmin=min_step_size)
```
f0uriest, line 1354: might be better to make this a user supplied input with default? to allow for other controllers, constant step size etc.

desc/magnetic_fields.py
```diff
@@ -1248,21 +1272,68 @@ def field_line_integrate(
...
+        If not given and one of bounds_R,Z,phi are given, will
+        default to a terminating event which ends integration
+        once R,Z,or phi exit the domain defined by the given bounds
+
+        NOTE: If the solve is terminated early, the output returned is still
+        length(phis), however all values from the step point the fxn evaluated
+        to True and on will be inf
+        state has attributes such as state.y (array of length 3 with current (R,phi,Z))
+        see diffrax documentation for more in-depth information
+    bounds_R: tuple
```
f0uriest, line 1304: is this use case common enough to have these options for it vs just asking the user to supply a `DiscreteTerminatingEvent` object?

dpanici: I think yes

desc/magnetic_fields/_core.py
```diff
@@ -1485,11 +1494,13 @@ def field_line_integrate(
     rtol=1e-8,
     atol=1e-8,
     maxstep=1000,
+    min_step_size=1e-8,
```
f0uriest, line 1497: I tried a few different values for this and it didn't seem to have a major effect on speed, and accuracy is fine as long as its reasonably small. We could do something fancy using the data from Albanese 2015 but I think probably not worth it, and if the user wants that they can still give a particular value for this or even give a different step size controller.

desc/magnetic_fields/_core.py
```diff
@@ -1538,60 +1544,64 @@ def field_line_integrate(
...
+
+    # diffrax parameters
+
+    def default_terminating_event_fxn(state, **kwargs):
+        R_out = jnp.any(jnp.array([state.y[0] < bounds_R[0], state.y[0] > bounds_R[1]]))
+        Z_out = jnp.any(jnp.array([state.y[2] < bounds_Z[0], state.y[2] > bounds_Z[1]]))
+        return jnp.any(jnp.array([R_out, Z_out]))
+
+    kwargs.setdefault(
+        "stepsize_controller", PIDController(rtol=rtol, atol=atol, dtmin=min_step_size)
```
ddudt, line 1571: What does a PID controller have to do with integration?

ddudt (approved):

desc/magnetic_fields/_core.py
f0uriest: It's using a PID controller with y=error estimate and u=step size to choose steps (I think it defaults to just PI control but the theory is all the same)

f0uriest:
@dpanici I know you can't formally approve this but lmk if you have any comments

ddudt (approved):

desc/magnetic_fields/_core.py
ddudt: That's neat

desc/magnetic_fields/_core.py
```diff
@@ -2228,60 +2234,64 @@ def field_line_integrate(
...
         ).T
-        return (
-            decay_factor
-            * jnp.array(
-                [r * br / bp * jnp.sign(bp), jnp.sign(bp), r * bz / bp * jnp.sign(bp)]
-            ).squeeze()
-        )
+        return jnp.array(
+            [r * br / bp * jnp.sign(bp), jnp.sign(bp), r * bz / bp * jnp.sign(bp)]
+        ).squeeze()
```
YigitElma, line 2251:
```suggestion
        # step along the field line
        # Torodial component of the field is used to normalize the step size
        return jnp.array(
            jnp.sign(bp) * [r * br / bp, 1, r * bz / bp]
        ).squeeze()
```

YigitElma: Not that important if you don't change

desc/magnetic_fields/_core.py
```diff
@@ -2228,60 +2234,64 @@ def field_line_integrate(
...
+
+    # diffrax parameters
+
+    def default_terminating_event_fxn(state, **kwargs):
+        R_out = jnp.any(jnp.array([state.y[0] < bounds_R[0], state.y[0] > bounds_R[1]]))
+        Z_out = jnp.any(jnp.array([state.y[2] < bounds_Z[0], state.y[2] > bounds_Z[1]]))
+        return jnp.any(jnp.array([R_out, Z_out]))
+
+    kwargs.setdefault(
+        "stepsize_controller", PIDController(rtol=rtol, atol=atol, dtmin=min_step_size)
```
YigitElma, line 2261: Maybe a bit more explanation of why we use PID would be good

YigitElma: And is the speed up coming this PID or `diffrax` or both?

desc/magnetic_fields/_core.py
f0uriest: it's not that we're normalizing the step or anything, its just because we're using the toroidal angle as our "time" coordinate so that modifies the ODE slightly.

desc/magnetic_fields/_core.py
f0uriest: This is just the default step size controller in diffrax. Basically ever other adaptive ODE method uses something similar, they're just less explicit about it. But under the hood this is basically doing the same thing as matlab rk45 or similar.

I think the speedup is due to lots of little tricks that diffrax does that the old jax method didn't, I don't think its specific to any one thing (and with the default parameters this PID controller is the same as what jax odeint uses internally)

tests/test_magnetic_fields.py
```diff
@@ -1038,24 +1039,51 @@ def test_field_line_integrate(self):
...
-        # this will hit the R bound
-        # (there is no Z bound, and R would go to 10.0 if not bounded)
-        r, z = field_line_integrate(r0, z0, phis, field, bounds_R=(10.05, np.inf))
-        np.testing.assert_allclose(r[-1], 10.05, rtol=3e-4)
-        # this will hit the Z bound
-        # (there is no R bound, and Z would go to 0.1 if not bounded)
-        r, z = field_line_integrate(r0, z0, phis, field, bounds_Z=(-np.inf, 0.05))
-        np.testing.assert_allclose(z[-1], 0.05, atol=3e-3)
+        phis = [0, 2 * np.pi * 25]
+        r, z = field_line_integrate(r0, z0, phis, field, solver=Dopri5())
```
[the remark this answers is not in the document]
f0uriest: Dopri5 and Tsit5 are both rk45 methods, just with slightly different coefficients. There's not any real reason to prefer one over the other, other than to just test that they both work.

YigitElma (approved):

PR #611, opened 2023-08-03
add new versions of JAX to JAX tests

dpanici:


dpanici:
@f0uriest I thought that with `fail-fast: false` in the `jax_tests.yml` it is supposed to let all of them run even if other jobs fail? For some reason it keeps cancelling the jobs because earlier ones failed...

f0uriest:
> @f0uriest I thought that with `fail-fast: false` in the `jax_tests.yml` it is supposed to let all of them run even if other jobs fail? For some reason it keeps cancelling the jobs because earlier ones failed...

yeah it should, and it looks like it is? I think the jobs being cancelled might be a OOM issue, I've noticed that depending how the tests are split it will sometimes cause the runner to kill the job because it uses too much memory.

dpanici:
hm well I also added `continue_on_error`, which I think is safe, it seems to be running better now. Would it cancel without even throwing an OOM if it was OOM?

f0uriest:
> hm well I also added `continue_on_error`, which I think is safe, it seems to be running better now. Would it cancel without even throwing an OOM if it was OOM?

Yeah if it's killed by the os it probably doesn't exit properly and give an error message. You can try cherry picking https://github.com/PlasmaControl/DESC/commit/d2739753ac4683b6c6e2042009a5b300d7e9d4e6 and see if it helps

dpanici:
@f0uriest Looks like we could bump the supported versions of JAX up to the most recent one (the only failing ones are 0.4.9 and 0.4.10 but it is due to a "benign" DeprecationWarning that could be solved by downgrading `mltypes` package, or if we just allow deprecation warnings to not fail the pytest)

What do you think? To change the field line integration to use diffrax they have jax 0.4.13 as their pinned version unfortunately

f0uriest:
Sounds fine to me

ddudt (dismissed):

f0uriest:
Rather than pinning ml_dtypes, which we don't explicitly use, it's probably better just to ignore that deprecation warning

ddudt (changes requested):
Also need to update the JAX version in `requirements_conda.yml` and `devtools/dev-requirements_conda.yml`.

f0uriest (approved):

ddudt (approved):

PR #612, opened 2023-08-07
Surface variance

unalmis:
To compute the variance of quantities over surfaces
[variance_math.pdf](https://github.com/PlasmaControl/DESC/files/12283780/variance_math.pdf)


desc/compute/utils.py
```diff
@@ -1204,6 +1204,112 @@ def surface_integrals_transform(grid, surface_label="rho"):
...
+        The second case is when the weights are intended to assign more weight
+        to samples with less uncertainty. See
+        https://en.wikipedia.org/wiki/Inverse-variance_weighting.
+        The unbiased sample variance for this case is obtained by replacing the
+        effective number of samples in the formula this function implements,
+        :math:`n_e`, with the actual number of samples :math:`n`.
+
+        The third case is when the weights denote the integer frequency of each
+        sample. See
+        https://en.wikipedia.org/wiki/Weighted_arithmetic_mean#Frequency_weights.
```
unalmis, line 1262: FYI, If you read through this link I think you'll find that wikipedia has some errors. That formula should only be used on integer frequeny weights. They suggest that you can still use the formula when the frequency weights are normalized, but in addition to loosing ability to unbias, normalizing to 1 blows up their expression as the denominator is 0.

desc/compute/utils.py
```diff
@@ -1204,6 +1204,112 @@ def surface_integrals_transform(grid, surface_label="rho"):
...
+    n_e = v1**2 / v2
+    # analogous to Bessel's bias correction
+    correction = n_e / (n_e - bool(unbiased))
+
+    q = jnp.atleast_1d(q)
+    # compute variance in two passes to avoid catastrophic round off error
+    mean = (integrate((weights * q.T).T).T / v1).T
+    mean = grid.expand(mean, surface_label)
+    variance = (correction * integrate((weights * ((q - mean) ** 2).T).T).T / v1).T
+    return grid.expand(variance, surface_label) if expand_out else variance
```
unalmis, line 1310: All the `.T` you see are artifacts of how we vectorize the computation in the case `q` is a 2d or 3d array. The first axis of a multidimensional array iterates through the nodes of a grid. Numpy multiplies a 1d and 2d array by multiplying element wise against the last dimension. So we transpose the possible multidimensional array before a multiplication so that the last axis iterates through the nodes of a grid.

tests/test_compute_utils.py
```diff
@@ -373,6 +374,46 @@ def test_surface_averages_shortcut(self):
...
+        n = grid.num_rho * grid.num_theta
+        np.testing.assert_allclose(np.bincount(grid.inverse_zeta_idx), desired=n)
+
+        # something arbitrary
+        q = np.arange(grid.num_nodes)
+        # Check correctness of mean calculation.
+        mean = grid.expand(
+            q.reshape((grid.num_zeta, -1)).mean(axis=-1),
+            surface_label="zeta",
+        )
```
unalmis, line 392: Our predefined grids sort the nodes in chunks of common zeta coordinates. So to compute a quantity that is local to a surface, we can reshape them into zeta surface chunks and take the mean across the chunks.

desc/compute/utils.py
```diff
@@ -1204,6 +1204,112 @@ def surface_integrals_transform(grid, surface_label="rho"):
...
+    Parameters
+    ----------
+    grid : Grid
+        Collocation grid containing the nodes to evaluate at.
+    q : ndarray
+        Quantity to compute the sample variance.
+    weights : ndarray
+        Weight assigned to each sample of ``q``.
+        A good candidate for this parameter is the surface area Jacobian.
+    unbiased : bool
```
ddudt, line 1275: Suggestion: switch this logic to `biased`. Saying `unbiased = False` feels like a confusing double-negative to me, but `biased = True` or `biased = False` is easy to understand.

tests/test_compute_utils.py
```diff
@@ -373,6 +374,46 @@ def test_surface_averages_shortcut(self):
             err_msg="average without sqrt(g) fail",
         )
 
+    @pytest.mark.unit
+    def test_surface_variance(self):
+        """Test that variance is computed correctly when weights are equal."""
```
ddudt, line 379: Do you want to add another test for a case where the weights are not equal?

unalmis: Added. Also moved the tests concerning mean to the surface_average unit test

desc/compute/utils.py
dpanici: When you compute the variance in two passes, what do you mean about catastrophic roundoff error if we only did one pass? I am just curious

unalmis: Catastrophic is the technical term https://en.wikipedia.org/wiki/Catastrophic_cancellation. The difference of squares example they give is what we want to avoid, as the squares will statistically be close. See http://www.cs.yale.edu/publications/techreports/tr222.pdf.

dpanici (approved):

desc/compute/utils.py
```diff
@@ -1212,6 +1212,112 @@ def surface_integrals_transform(grid, surface_label="rho"):
...
+    weights=jnp.array([1.0]),
+    bias=False,
+    surface_label="rho",
+    expand_out=True,
+):
+    r"""Compute the weighted sample variance of ``q`` on each surface of the grid.
+
+    Computes the following quantity on each surface of the grid.
+
+    .. math::
```
f0uriest, line 1227: minor point, but I've been trying to use unicode instead of TeX for math notation in docstrings, so that its readable both in source and formatted form. There's a handy website for converting stuff: https://www.unicodeit.net/

f0uriest (approved):
Mostly seems fine, if you make any other changes the math notation could be cleaned up. My main question is what is the expected use case for this new function? Do you have anything in mind?

unalmis:
> Mostly seems fine, if you make any other changes the math notation could be cleaned up. My main question is what is the expected use case for this new function? Do you have anything in mind?

@dpanici or @Tal-Shpigel can explain this better than me
Also codecov will prevent me from merging this PR (can't force merge).

PR #613, opened 2023-08-07
Add utilities for random surfaces and pressure profiles

### LINKED ISSUE #541: Add utilites to create "random" equilibria
For database studies, testing, examples etc, it would be nice to be able to generate some "random" equilibria that satisfy certain basic criteria, such as major/minor radius, field strength, beta, etc.

- [ ] Random surface with approximate shape that ideally doesn't self intersect
- [ ] Random monotonic profiles
- [ ] Random "smooth" non-monotonic profiles

f0uriest:
Resolves #541


desc/random.py
```diff
@@ -0,0 +1,190 @@
...
+        Random number generator. If None, uses numpys default_rng
+
+    Returns
+    -------
+    pressure : PowerSeriesProfile
+        Random pressure profile.
+    """
+    rng = setdefault(rng, default_rng())
+    if isinstance(n, tuple):
+        n = rng.integers(n[0] // 2, (n[1] + 1) // 2) * 2  # ensure its even
```
ddudt, line 147: It feels weird to have the order of the polynomial be randomly assigned. I would prefer a similar setup as `random_surface`, where the resolution is fixed but there is also a parameter for the user to specify how quickly the coefficients decay. Then you can have a high order polynomial with decaying values that acts low order.

ddudt (changes requested):

desc/random.py
```diff
@@ -0,0 +1,190 @@
...
+        abs(R_mn)[(abs(R_basis.modes[:, 1]) == 1) & (abs(R_basis.modes[:, 2]) == 0)]
+    )
+    Z_scale1 = np.mean(
+        abs(Z_mn)[(abs(Z_basis.modes[:, 1]) == 1) & (abs(Z_basis.modes[:, 2]) == 0)]
+    )
+    R_mn *= R_scale / R_scale1
+    Z_mn *= Z_scale / Z_scale1
+    R_mn[R_basis.get_idx(0, 0, 0)] = R0
+    if not sym:
+        Z_mn[Z_basis.get_idx(0, 0, 0)] = 0  # center at Z=0
```
dpanici, line 103: test for non-sym?

dpanici (changes requested):
should add a non-sym test as well for it

dpanici (dismissed):

desc/random.py
```diff
@@ -0,0 +1,198 @@
...
+    Profile will be a PowerSeriesProfile with even symmetry,
+    enforced to be monotonically decreasing from p0 at r=0 to 0 at r=1
+
+    Could also be used for other monotonically decreasing profiles
+    such as temperature or density.
+
+    Parameters
+    ----------
+    n : int
+        Order of polynomial.
```
ddudt, line 143: Can we change this to `L` to be consistent with our other notation?

desc/random.py
```diff
@@ -0,0 +1,198 @@
...
+        Pressure on axis. If a tuple, treats as min/max for random value.
+    rng : numpy.random.Generator
+        Random number generator. If None, uses numpys default_rng
+
+    Returns
+    -------
+    pressure : PowerSeriesProfile
+        Random pressure profile.
+    """
+    assert (n // 2) == (n / 2), "n should be even"
```
ddudt, line 154: I don't think we should enforce the order to be even. You only generate random coefficients for the even powers of rho anyways, so in the case of an odd order the coefficient for the highest power can also be zero.

ddudt (changes requested):

desc/random.py
f0uriest: The logic in the rest of the function needs the order to be even, and we need even profiles for them to be physical anyways

ddudt (approved):

PR #615, opened 2023-08-07
add Coils and magnetic fields to docs api

dpanici:


dpanici:
black format the doc code

ddudt (changes requested):
update example code to have black formatting

docs/api.rst
```diff
@@ -17,6 +17,18 @@ Basis
     desc.basis.ChebyshevDoubleFourierBasis
     desc.basis.FourierZernikeBasis
 
+Coils
+*****
+
+.. autosummary::
+    :toctree: _api/basis/
```
f0uriest, line 24: this should be `_api/coils/`

docs/api.rst
```diff
@@ -104,6 +116,22 @@ IO
     desc.io.InputReader
     desc.io.load
 
+Magnetic Fields
+***************
+
+.. autosummary::
+    :toctree: _api/basis/
```
f0uriest, line 123: should be `_api/magnetic_fields`

desc/coils.py
```diff
@@ -188,6 +253,40 @@ class FourierPlanarCoil(Coil, FourierPlanarCurve):
...
+        R_coil = 10
+        # circular coil given by center at (0,0,0)
+        # and normal vector in Z direction (0,0,1) and radius 10
+        coil = FourierPlanarCoil(
+            current=I,
+            center=[0, 0, 0],
+            normal=[0, 0, 1],
+            r_n=R_coil,
+            modes=[0],
+            grid=LinearGrid(N=100),
```
f0uriest, line 275: FYI, I'm not sure about having `grid` objects assigned to each coil. I was considering getting rid of it in #583 but felt we should discuss it first. I think it would be better to add something like `source_grid` as an optional argument to `compute_magnetic_field`, especially since after #583 whatever grid you assign to the coil won't be used by the underlying `Curve.compute` method.

dpanici: I agree, that is something we can change in #494 when I update the coil stuff to reflect the #583 changes. This PR though is just for the docs

ddudt (approved):

Issue #603, opened 2023-07-31
Elongation returns NaN for circular cross section


f0uriest:
For circular cross sections (like our default surface and equilibrium starting points), the calculation of elongation returns NaN instead of 1

eg
```python
eq = Equilibrium()
eq.compute("a_major/a_minor")["a_major/a_minor"]
```

I suspect its a numerical issue from taking sqrt of a subtraction that ends up being close to zero. Definitely needs to be fixed, since a circular cross section is a common starting point for optimization.

ddudt:
Resolved by #618.

PR #618, opened 2023-08-11
compute geometry updates

### LINKED ISSUE #597: Compute A(r)
Presently we have a compute function for cross-sectional area `A`, which is a scalar quantity for the area through the $\rho=1$ surface. I would like to compute the same quantity but as a flux function, $A(\rho)$.

ddudt:
Resolves #597 and #603 and adds `Psi` profile (true toroidal magnetic flux, not normalized by $2\pi$.

### DIFF
--- a/desc/compute/_profiles.py
+++ b/desc/compute/_profiles.py
@@ -7,6 +7,24 @@ from .utils import compress, cumtrapz, dot, expand, surface_averages
 
 
 @register_compute_fun(
+    name="Psi",
+    label="\\Psi",
+    units="Wb",
+    units_long="Webers",
+    description="Toroidal flux",
+    dim=1,
+    params=["Psi"],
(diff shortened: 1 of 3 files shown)


desc/compute/_profiles.py
```diff
@@ -6,6 +6,24 @@
...
+    units_long="Webers",
+    description="Toroidal flux",
+    dim=1,
+    params=["Psi"],
+    transforms={},
+    profiles=[],
+    coordinates="r",
+    data=["rho"],
+)
+def _Psi(params, transforms, profiles, data, **kwargs):
```
f0uriest, line 22: do we also want to add derivatives of Psi?

f0uriest (approved):

dpanici (approved):

PR #620, opened 2023-08-15
Fix simsopt QH comparison test

### LINKED ISSUE #599: Simsopt comparison test sort of broken
It seems there's a bug causing the simsopt QH comparison test to give a QA result (with the QS cost going down to machine precision):

image

f0uriest:
There were a few issues:

- After switching the sign of the jacobian, we forgot to switch the sign of the target helicity
- Per the email chain with Matt that inspired the test, 
> Unfortunately the optimizer then fell into an unphysical minimum with very small objective resembling a tall cylinder. I get this sometimes in simsopt too with QH optimizations if the initial condition doesn't have enough axis torsion. Increasing the (m=0,n=4) mode of the initial condition resolved this, and now desc gets the real QH solution with f=0.075, nearly identical to the simsopt result.

So we should have increased Delta from the start, the one in the original test was expected to converge to a really high elongation.

With the fixes, we get a nice reasonable QH:
image

Resolves #599

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -420,7 +420,7 @@ def test_simsopt_QH_comparison():
     nfp = 4
     aspect_target = 8.0
     # Initial (m=0, n=nfp) mode of the axis:
-    Delta = 0.2
+    Delta = 0.4
     LMN_resolution = 6
     # Set shape of the initial condition.
     # R_lmn and Z_lmn are the amplitudes. modes_R and modes_Z are the (m,n) pairs.
@@ -508,7 +508,7 @@ def test_simsopt_QH_comparison():
                 eq=eq, target=aspect_target, weight=aspect_weight, normalize=False
             ),
             QuasisymmetryTwoTerm(
-                eq=eq, helicity=(1, nfp), grid=grid, weight=qs_weight, normalize=False
+                eq=eq, helicity=(-1, nfp), grid=grid, weight=qs_weight, normalize=False
             ),
         )
     )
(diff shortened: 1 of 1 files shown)


tests/test_examples.py
```diff
@@ -420,7 +420,7 @@ def test_simsopt_QH_comparison():
     nfp = 4
     aspect_target = 8.0
     # Initial (m=0, n=nfp) mode of the axis:
-    Delta = 0.2
+    Delta = 0.4
```
ddudt, line 423: I would recommend changing this variable name to `torsion`, since that is what this is controlling. `Delta` makes me think of the VMEC input parameter with the same name.

ddudt (approved):

tests/test_examples.py
rahulgaur104: Whatever name you decide, it would be useful to make it consistent with the Frenet-Serret notation. For example, torsion is tau in Frenet-Serret.

dpanici (approved):

rahulgaur104:
The helicity is now -1. Are we sure this is not going to happen again? What if we set the helicity to helicity_value*sign(jacobian)?

f0uriest:
> The helicity is now -1. Are we sure this is not going to happen again? What if we set the helicity to helicity_value*sign(jacobian)?

Since v0.7.0, the jacobian in DESC is enforced to always be positive so it should be ok

rahulgaur104 (approved):

PR #622, opened 2023-08-15
rho grid bug fixes

### LINKED ISSUE #619: BootstrapRedlConsistency default grid
The default grid for the `BootstrapRedlConsistency` objective includes the $\rho=1$ surface. Typical density and pressure profiles vanish at the boundary surface; this causes division by 0 in the Redl formula, and results in the objective returning a `NaN` value. 

The easy fix is to change the default grid to exclude the $\rho=1$ surface. We may also want to do some checking when the objective is built to ensure the profiles and grid are consistent with each other.

ddudt:
Resolves #619 and also fixes a bug where `PowerSeriesProfile.from_values(sym=True)` did not actually return an even profile.


dpanici (changes requested):
Can you add a test for the PowerSeriesProfile fix? like just going from SplineProfile to PowerSeries and checking that it is even I mean

desc/profiles.py
```diff
@@ -768,6 +766,9 @@ def from_values(
             profile in power series basis fit to given data.
 
         """
+        if sym:
+            x = x**2
+            order = order // 2
```
f0uriest, line 771: I don't think we want to reduce the order here without padding the params later, or passing in the mode numbers explicitly

f0uriest (changes requested):
I think the even profile fix isn't quite right. Also we should leave #619 open for now, or make a new issue to evaluate the limit properly

tests/test_profiles.py
```diff
@@ -346,16 +360,6 @@ def test_profile_errors(self):
             a = sp * pp
             a.params = sp.params
 
-    @pytest.mark.unit
```
ddudt, line 349: I deleted this test because it was redundant with `test_close_values`.

f0uriest (approved):

dpanici (approved):

PR #623, opened 2023-08-15
Ts/bd min

dpanici:
Add new objective to minimize variance of |B| times minimum distance from plasma to winding surface. This is to try to ensure that the equilibrium magnetic field is more feasible to create with the given winding surface by penalizing having larger magnetic fields farther away from the winding surface.


Tal-Shpigel:
We made this objective to create equilibrium surfaces that could be more readily reproduced from a surface current and, ultimately, coils.
This is the problem we had:
Screenshot 2023-08-16 at 19 20 20

Using the objective, minimizing the surface variance of |B|*d_min yielded good initial results:
Screenshot 2023-08-16 at 19 22 21

And was recreated fairly well by a surface current:
trace50_transits_trace_M_50_N_160_alpha_0_modopt_ct32NFP4_opt7

desc/objectives/_geometry.py
```diff
@@ -1077,3 +1078,224 @@ def compute(self, *args, **kwargs):
...
+    weight : float, ndarray, optional
+        Weighting to apply to the Objective, relative to other Objectives.
+        len(weight) must be equal to Objective.dim_f
+    normalize : bool
+        Whether to compute the error in physical units or non-dimensionalize.
+    normalize_target : bool
+        Whether target should be normalized before comparing to computed values.
+        if `normalize` is `True` and the target is in physical units, this should also
+        be set to True.
+    surface_grid : Grid, ndarray, optional
```
f0uriest, line 1122: shouldn't be ndarray

desc/objectives/_geometry.py
```diff
@@ -1077,3 +1078,224 @@ def compute(self, *args, **kwargs):
...
+        Collocation grid containing the nodes to evaluate surface geometry at.
+    plasma_grid : Grid, ndarray, optional
+        Collocation grid containing the nodes to evaluate plasma geometry at.
+    use_softmin: Bool, use softmin or hard min.
```
f0uriest, line 1126: should have a new indented line after datatype, also bool should be lowercase

desc/objectives/_geometry.py
```diff
@@ -1077,3 +1078,224 @@ def compute(self, *args, **kwargs):
...
+            has_axis=plasma_grid.axis.size or surface_grid.axis.size,
+            obj="desc.equilibrium.equilibrium.Equilibrium",
+        )
+
+        timer = Timer()
+        if verbose > 0:
+            print("Precomputing transforms")
+        timer.start("Precomputing transforms")
+
+        self._surface_coords = self._surface.compute("x", basis="xyz")["x"]
```
f0uriest, line 1213: should the surface grid be passed in here?

desc/objectives/_geometry.py
```diff
@@ -1077,3 +1078,224 @@ def compute(self, *args, **kwargs):
...
+            - constants["surface_coords"][None, :, :],
+            axis=-1,
+        )
+
+        if self._use_softmin:  # do softmin
+            dmin_data = jnp.apply_along_axis(softmin, -1, d, self._alpha)
+        else:  # do hardmin
+            dmin_data = d.min(axis=-1)
+
+        B_dmin_data = jnp.array([d_min * B for d_min, B in zip(dmin_data, data["|B|"])])
```
f0uriest, line 1292: can't these just be broadcasted normally?

desc/objectives/_geometry.py
```diff
@@ -1077,3 +1078,224 @@ def compute(self, *args, **kwargs):
             profiles=constants["profiles"],
         )
         return data["L_grad(B)"]
+
+
+class PlasmaCoilFeasibilty(_Objective):
```
f0uriest, line 1083: feel like this name is too general, since there are a lot of different coil feasibility metrics (like the LgradB above). Might want something more specific to this.

dpanici: Yea I could not think of a good one. maybe one like BStrengthDistanceFeasibility or something...

desc/objectives/_geometry.py
dpanici: yes, though when I make that change our test fails... it must not be a good test then

f0uriest: It might also be something weird with array shaping. I'd double check the sizes and values of things to make sure it's doing what you want.

desc/objectives/_geometry.py
f0uriest: What about `BDistanceVariance`?

desc/objectives/_geometry.py
```diff
@@ -1077,3 +1078,225 @@ def compute(self, *args, **kwargs):
...
+        Whether target should be normalized before comparing to computed values.
+        if `normalize` is `True` and the target is in physical units, this should also
+        be set to True.
+    surface_grid : Grid
+        Collocation grid containing the nodes to evaluate surface geometry at.
+    plasma_grid : Grid
+        Collocation grid containing the nodes to evaluate plasma geometry at.
+    use_softmin: bool,
+        use softmin or hard min.
+    alpha: float, parameter used for softmin. The larger alpha, the closer the softmin
```
f0uriest, line 1128: missing a newline here as well after the type

PR #624, opened 2023-08-15
Shear objective

ddudt:
- Adds `Shear` as a new objective to target the shear profile. 
- Moves all of the profile objectives to their own file. 

This PR is based on #556 so that the axis limits are included in the profile objective tests.


desc/objectives/_profiles.py
```diff
@@ -0,0 +1,544 @@
...
+    def print_value(self, *args, **kwargs):
+        """Print the value of the objective."""
+        f = self.compute(*args, **kwargs)
+        print("Maximum " + self._print_value_fmt.format(jnp.max(f)) + self._units)
+        print("Minimum " + self._print_value_fmt.format(jnp.min(f)) + self._units)
+        print("Average " + self._print_value_fmt.format(jnp.mean(f)) + self._units)
+
+
+class Shear(_Objective):
+    """Targets a shear profile (derivative of rotational transform).
```
f0uriest, line 186: should probably put the equation here to make the normalization/definition clear

desc/objectives/_profiles.py
```diff
@@ -0,0 +1,544 @@
...
+        data = compute_fun(
+            "desc.equilibrium.equilibrium.Equilibrium",
+            self._data_keys,
+            params=params,
+            transforms=constants["transforms"],
+            profiles=constants["profiles"],
+        )
+        shear = jnp.where(
+            data["iota"] == 0,
+            data["iota_r"],
```
f0uriest, line 336: is this the correct limit?

desc/objectives/_profiles.py
```diff
@@ -0,0 +1,544 @@
...
+        target=None,
+        bounds=None,
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        grid=None,
+        name="shear",
+    ):
+        if target is None and bounds is None:
+            bounds = (0, np.inf)
```
f0uriest, line 233: why not default to target of 0?

ddudt: because typically you want finite shear for stability and island suppression

desc/objectives/_profiles.py
ddudt: No. Doing l'Hospital's rule gives: $\frac{\rho\iota'}{\iota} \rightarrow 1 + \rho\frac{\iota''}{\iota'}$. 

But the case I'm really trying to handle is the default vacuum tokamak `Equilibrium()` that has zero rotational transform everywhere, and therefore all of the iota derivates will also be 0. So I'm not sure how to handle that case and I'm open to suggestions.

f0uriest: hmm i see. Another possible tricky bit is that iota may not be exactly zero, but down to machine precision, so maybe we should have a threshold?

As for the limit, I think it would be good to implement the "correct" limit for cases that are more like a FRC where iota passes through zero.

For the case where iota is identically zero we could add something like 
`jnp.where(jnp.all(jnp.abs(iota)) < eps, 0, shear)`, to distinguish between iota being zero at a single point vs the whole profile

desc/objectives/_profiles.py
f0uriest: OK, sounds reasonable. Is there any reason it should be positive vs negative? Just trying to think what the most common use cases would be. Something like a target of 1 might also be reasonable

desc/objectives/_profiles.py
ddudt: Done, see the latest changes.

desc/objectives/_profiles.py
ddudt: The sign will depend on the sign of iota, but that would be annoying to compute before the objective is built. I think we can trust the user to override the default when needed. 

I don't want to set a specific target because the actual amount of shear desired will vary greatly. For example, LHD has very large shear, while W7-X wanted low shear to avoid rational surfaces.

desc/objectives/_profiles.py
unalmis: The first call to jnp.where is only correct near $\rho = 0$. To do what you want, you would need something like
```python
indeterminate_1 = (jnp.abs(data["iota"]) < self._eps) & (jnp.abs(data["rho"]) < self._eps`
indeterminate_2 = (jnp.abs(data["iota"]) < self._eps) & (jnp.abs(data["iota_r"]) < self._eps`
shear = jnp.where(indeterminate_1 | indeterminate_2, limit, shear)
# The limit is undefined in this case so just set to zero.
shear = jnp.where(jnp.abs(iota).all() < eps, 0, shear)
```

unalmis: Does `jnp.nan` mess up the objective?

desc/objectives/_profiles.py
```diff
@@ -0,0 +1,549 @@
...
+            params=params,
+            transforms=constants["transforms"],
+            profiles=constants["profiles"],
+        )
+        shear = jnp.where(
+            jnp.abs(data["iota"]) < self._eps,
+            1 + data["rho"] * data["iota_rr"] / data["iota_r"],
+            data["rho"] * data["iota_r"] / data["iota"],
+        )
+        shear = jnp.where(jnp.all(jnp.abs(data["iota"])) < self._eps, 0, shear)
```
f0uriest, line 343: This might cause NaN in the gradient if the previous `jnp.where` divided by zero and then it gets overwritten. Might be better to use `cond` to first check if iota is all zero, then use jnp.where inside that to handle the case where its zero at a point.

desc/objectives/_profiles.py
unalmis: Basically the requirement to apply lhopitals rule is:
```python
# iota needs to be close to zero, and then either rho is close to 0 or iota_r is close to 0
(jnp.abs(data["iota"]) < self._eps) & ((jnp.abs(data["rho"]) < self._eps) | (jnp.abs(data["iota_r"]) < self._eps))
```

unalmis:
> add `shear` as a compute quantity

If you're going to use `np.where` in compute functions it'll help to `git cherry-pick` [bca5fc6](https://github.com/PlasmaControl/DESC/pull/395/commits/bca5fc6c5518040499b7d09f5bb42bea7a70db17)

desc/compute/_profiles.py
```diff
@@ -1474,3 +1474,33 @@ def _current_r(params, transforms, profiles, data, **kwargs):
...
+    dim=1,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="r",
+    data=["0", "iota", "iota_r", "iota_rr", "rho"],
+    axis_limit_data=["iota_den_r", "iota_num_r"],
+)
+def _shear(params, transforms, profiles, data, **kwargs):
+    eps = 1e2 * jnp.finfo(jnp.array([1.0]).dtype).eps
```
ddudt, line 1494: Is there a better way to do this?

f0uriest: might be more understandable if you use `data['iota'].dtype`

desc/compute/_profiles.py
```diff
@@ -1474,3 +1474,33 @@ def _current_r(params, transforms, profiles, data, **kwargs):
...
+    data["shear"] = cond(
+        jnp.all(jnp.abs(data["iota"])) < eps,
+        lambda _: data["0"],  # if iota profile is all 0, set shear to 0
+        lambda _: jnp.where(  # apply l'Hopital rule where necessary
+            (jnp.abs(data["iota"]) < eps)
+            & ((jnp.abs(data["rho"]) < eps) | (jnp.abs(data["iota_r"]) < eps)),
+            1 + data["rho"] * data["iota_rr"] / data["iota_r"],  # limit
+            data["rho"] * data["iota_r"] / data["iota"],  # shear
+        ),
+        1,
```
f0uriest, line 1504: We've been using `None` as a dummy argument in other places, I'd prefer that to `1` which might be confused for an actual value.

desc/compute/_profiles.py
```diff
@@ -1474,3 +1474,33 @@ def _current_r(params, transforms, profiles, data, **kwargs):
...
+
+@register_compute_fun(
+    name="shear",
+    label="\\rho \\partial_{\\rho} \\iota / \\iota",
+    units="~",
+    units_long="None",
+    description="Normalized derivative of the rotational transform",
+    dim=1,
+    params=[],
+    transforms={"grid": []},
```
f0uriest, line 1487: doesn't look like the function actually needs the grid

desc/compute/_profiles.py
```diff
@@ -1474,3 +1474,33 @@ def _current_r(params, transforms, profiles, data, **kwargs):
...
+    profiles=[],
+    coordinates="r",
+    data=["0", "iota", "iota_r", "iota_rr", "rho"],
+    axis_limit_data=["iota_den_r", "iota_num_r"],
```
f0uriest, line 1491: are these needed? they don't seem to be used, and i think the axis limit of the shear is well behaved the way you calculate it

tests/test_objective_funs.py
```diff
@@ -169,6 +170,21 @@ def test(eq):
...
+
+        def test(eq):
+            obj = Shear(target=1, weight=2, eq=eq)
+            obj.build()
+            shear = obj.compute_unscaled(*obj.xs(eq))
+            shear_scaled = obj.compute_scaled_error(*obj.xs(eq))
+            np.testing.assert_allclose(shear, 0)
+            np.testing.assert_allclose(shear_scaled, -2 / np.sqrt(3))
+
+        test(Equilibrium(iota=PowerSeriesProfile(0)))
```
f0uriest, line 185: might want to test with a nonzero iota to make sure the non-degenerate case is correct as well

tests/test_objective_funs.py
```diff
@@ -700,6 +722,10 @@ def test(obj, values, normalize=False):
     obj = RotationalTransform(eq=eq, grid=grid)
     obj.build()
     test(obj, iota)
+    shear = eq.compute("iota_r", grid=grid)["iota_r"]
```
f0uriest, line 725: are these the "correct" values?

desc/compute/_profiles.py
ddudt: Done. But I meant about having this `eps` defined inside the compute function like this.

f0uriest: yeah thats pretty standard. The threshold depends on the application and the data, so a global one usually doesn't work.

f0uriest (dismissed):

desc/compute/_profiles.py
```diff
@@ -1474,3 +1474,32 @@ def _current_r(params, transforms, profiles, data, **kwargs):
...
+        lambda _: data["0"],  # if iota profile is all 0, set shear to 0
+        lambda _: jnp.where(  # apply l'Hopital rule where necessary
+            (jnp.abs(data["iota"]) < eps)
+            & ((jnp.abs(data["rho"]) < eps) | (jnp.abs(data["iota_r"]) < eps)),
+            1 + data["rho"] * data["iota_rr"] / data["iota_r"],  # limit
+            data["rho"] * data["iota_r"] / data["iota"],  # shear
+        ),
+        None,
+    )
+    return data
```
unalmis, line 1505: Typically $\partial_{\rho} \iota$ is 0 at axis because it should be able to be expanded as an even power series near axis so that it is symmetric under the transformation $\rho \to -\rho$. So in that case in order to compute this limit we need an additional application of lhipitals rule which yields $\lim_{\rho \to 0} \rho \frac{\iota}{\partial_{\rho} \iota} = 2 + \rho \frac{\partial_{\rho \rho \rho} \iota}{\partial_{\rho \rho} \iota}$.

We don't have the third derivative of $\iota$ ready to compute, so I think we should only handle the case where $\iota = 0$ everywhere and add `shear` to the not implemented limits list: https://github.com/PlasmaControl/DESC/blob/c0b44414883d202fb30c94e78fbbd5e7d29b59d7/tests/test_axis_limits.py#L53

desc/compute/_profiles.py
```diff
@@ -1474,3 +1474,32 @@ def _current_r(params, transforms, profiles, data, **kwargs):
...
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="r",
+    data=["0", "iota", "iota_r", "iota_rr", "rho"],
+)
+def _shear(params, transforms, profiles, data, **kwargs):
+    eps = 1e2 * jnp.finfo(data["iota"].dtype).eps
+    data["shear"] = cond(
+        jnp.all(jnp.abs(data["iota"])) < eps,
```
unalmis, line 1495: This should be `jnp.all(jnp.abs(data["iota"]) < eps)`

rahulgaur104: To make this consistent with the tokamak definition of shear, you should also put a minus sign in front.

For tokamaks, global shear = rho/q dq/drho and, q = 1/iota. Hence, global_shear = -rho/iota d iota/drho

desc/compute/_profiles.py
unalmis: Also since the limit is undefined in the case iota = 0 everywhere, and I think we are just setting shear to 0 in this case to avoid `nan` appearing in the objective, I feel like it would be better to do this inside the objective function rather than the compute function.

If you choose to continue handling this inside the compute function, can you add a comment that says we set it to zero for numerical reasons and it's not the actual limit

desc/compute/_profiles.py
ddudt: I don't think we want to follow the tokamak sign convention, since this is a stellarator code. When stellarator people talk about "positive shear" I take that to mean an increasing rotational transform.

rahulgaur104: The definitions are the same for both tokamaks and stellarators. Increasing iota radially outwards would mean negative shear. Feel free to double-check this.

ddudt: Here is an example [from this paper](https://www.jstage.jst.go.jp/article/pfr/3/0/3_0_S1004/_pdf/-char/ja) where the stellarator shear is defined to be positive, and the equivalent tokamak shear is negative. The sign is a convention, and it seems like each community chooses to make their own machines have positive values. (Stellarators almost always have radially increasing iota.)

image

rahulgaur104: This is a paper by a relatively small group, and they call tokamaks negative shear. This is the first time I have seen such a convention. Here are some [notes by Helander] (https://pure.mpg.de/rest/items/item_2146015/component/file_2146014/content). Please see the first line on page 24.
Here's a paper by [Hudson and Hegna](https://doi.org/10.1063/1.1622669) where the global shear is proportional to -iota_r. 

Note that negative shear is not the same as positive shear as far as stability is concerned. Negative shear almost always leads to a more stable equilibrium than positive shear. Therefore, I strongly suggest putting a negative sign in the definition.

desc/compute/_profiles.py
unalmis: Thanks for typing this out. I still think my comments above are valid.

> One exception is for Reverse Field Configurations (RFCs), where the shear reverses sign and crosses through $\iota=0$. But even then, this typically happens towards the last closed flux surface and far from the magnetic axis. In this case I think the shear is undefined because $\iota=0$ but $\iota'\neq0$, and our compute function will return `NaN`. 

Yes I agree.

> The only other exception is a vacuum tokamak, which has $\iota(\rho)=0~\forall\rho$ and we already cover that case. Otherwise any amount of torsion, 3D shaping, or toroidal current should give a finite rotational transform on the magnetic axis. 

In this case, the shear: $\rho \frac{\iota'}{\iota}$ is undefined everywhere (including axis) because there is no information about the speed the function approaches any point. At any point, shear is equivalent to the undefined arithmetic operation $0 / 0$ times $\rho$.  Currently we arbitrarily say this limit is 0. I'm guessing this is for numerical purposes of  avoiding  `nan` in the objective?

> I think the actual case we implemented the limit for where $\iota=\iota'=0$ is very rare. 

Yes it seems rare to me too. We failed to implement the limit in this case. In this case the code returns $1 + \rho \frac{\iota''}{\iota'}$. This will evaluate as `nan` because $\iota'$ is 0. The correct limit is

- 1 if $\rho = 0$ and $\iota'' \neq 0$.
- If $\rho = 0$ and $\iota'' = 0$, we need a second application of lhopitals rule which gives $2 + \rho \frac{\partial_{\rho \rho \rho} \iota}{\partial_{\rho \rho} \iota}$. This is also indeterminate so we need a third application $3 + \rho \frac{\partial_{\rho \rho \rho \rho} \iota}{\partial_{\rho \rho \rho} \iota}$. If this last term is not indeterminate we are done. Otherwise we continue applying lhopitals.
- If $\rho \neq 0$ and $\iota'' = 0$, then the same thing as above happens if $\partial_{\rho \rho \rho} \iota = 0$ too. Otherwise the limit is probably not finite.
- If $\rho \neq 0$ and $\iota'' \neq 0$, then the limit is probably not finite. 


This seems pedantic enough that I suggest only implementing what you need for the objective in the vacuum case and adding a comment that it's not the true limit. If I've made a mistake, sorry for using your time.

desc/compute/_profiles.py
ddudt: I concede

desc/compute/_profiles.py
ddudt: This is an important conversation, not a waste of time! I misunderstood your earlier comment, and I now see your point that the limit as implemented was useless. 

Yes, we "arbitrarily" set `shear = 0` when $\iota(\rho)=0~\forall\rho$ to avoid `NaN`, because starting from a vacuum tokamak is such a common use case. It is a physics-informed choice, even if it is not mathematically rigorous. 

I've decided to remove the handling of the limits for other cases where $\iota=0$, but added comments in the code to explain. As we have discussed, the shear *should* be undefined for most of these instances such as in an RFC. And as you pointed out, if $\iota=\iota'=0$ we would need higher order derivates, which I do not think is worth implementing for such a rare edge case.

f0uriest (approved):

unalmis (approved):

f0uriest:
Can you mention the new compute data and objective in the changelog?

unalmis (dismissed):

f0uriest (approved):

PR #627, opened 2023-08-17
Constant Offset Surface Algorithm

dpanici:
Implement constant offset algorithm described in Appendix B of ["An improved current potential method for fast computation of stellarator coil shapes", Landreman (2017) ](https://iopscience.iop.org/article/10.1088/1741-4326/aa57d4)

# Algorithm
-------------
Given a toroidal surface $\mathbf{r}_p(\theta,\zeta)$, we want to find a toroidal surface $\mathbf{r}_c(\theta',\zeta')$ offset from $\mathbf{r}_p$ by a uniform distance $\Delta$. For each pair of values of the poloidal/toroidal angles $\theta',\zeta'$, we can find $\mathbf{r}_c(\theta',\zeta')$ by letting $\theta = \theta'$ and solving the 1D nonlinear rootfinding problem

$$
\begin{equation}
arctan(y(\hat{\zeta}) / x(\hat{\zeta})) - \zeta' = 0
\end{equation}
$$

for the unknown $\hat{\zeta}$ where $x(\hat{\zeta})$ and $y(\hat{\zeta})$ are the Cartesian components of the vector $\mathbf{r}_p(\theta,\zeta) + \Delta \mathbf{n}(\theta, \zeta')$ where $\mathbf{n}$ is the unit normal vector to the surface $\mathbf{r}_p$. With the $\hat{\zeta}$ from this root finding, we then have the $\mathbf{r}_p(\theta', \hat{\zeta})$ which is the point on the original surface such that if we were to move the distance $\Delta$ in the normal direction, we would arrive at the toroidal angle $\zeta'$. Therefore we can find points on the offset surface on any desired grid in the (cylindrical) toroidal angle $\zeta'$, which we can then fit with a Fourier series in poloidal and toroidal angle to find the representation for the offset surface.

# TODO
---------
- [x] add math of how it works
- [x] test analytic derivative more and probably remove it as it takes longer for seemingly the same result
- [x] add type checks etc. inside function
- [x] write own fsolve (or steal the one in map_coords) in JAX so we can vectorize across the points and speed it up
- [x] assert omega is 0 - no way to do so without passing in eq as arg, which I don't want, so for now is not done. Did put in description it only works for tor angle = cyl angle. If in future we generalize this to general tor angle, will  have it accept eq as an arg (possibly optional, unless we also have the tor stream function be a parameter of the surface, in which case dont need to pass in an eq, can just do `surface.compute("omega")`)

Future PR
----------
- [x] Add a util to make FourierRZCurve out of FourierRZToroidalSurface at a constant theta/zeta @rahulgaur104 (evaluate basis at constant theta, and then find the coefficients that describe it directly from the surface's basis)


rahulgaur104:
I will add a [turning number](https://en.wikipedia.org/wiki/Total_curvature) calculation to this PR. This will tell us if the curve self-intersects. We want the turning number to be exactly equal to 1.

dpanici:
> I will add a [turning number](https://en.wikipedia.org/wiki/Total_curvature) calculation to this PR. This will tell us if the curve self-intersects. We want the turning number to be exactly equal to 1.
> 
> To do this, I only need the vectors dt/dl and n, where t and n are the tangent and normal vectors to the curve, respectively, and l is the distance along the curve.

will this help for figure out if a surface intersects? we would need to check a series of curves at constant poloidal angle theta right?

rahulgaur104:
Yes, we would have to choose different curves on the offset surface and check if they self-intersect.

ddudt:
This algorithm seems to be similar to one used in the python CAD package [build123d](https://build123d.readthedocs.io/en/latest/direct_api_reference.html#topology.Face.offset)

desc/geometry/surface.py
```diff
@@ -357,6 +361,200 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
         surf = cls(R_lmn=R_lmn, Z_lmn=Z_lmn, modes_R=modes_R, modes_Z=modes_Z, NFP=NFP)
         return surf
 
+    @classmethod
+    def from_values(cls, coords, nodes, M, N, NFP, sym=True):
```
f0uriest, line 365: Should the other PR be merged first? Or can we combine these two?

desc/geometry/surface.py
```diff
@@ -357,6 +361,200 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
...
+            Rb_lmn,
+            Zb_lmn,
+            R_basis.modes[:, 1:],
+            Z_basis.modes[:, 1:],
+            NFP,
+            sym,
+        )
+        return surf
+
+    @classmethod
```
f0uriest, line 418: Does this need to be a class method? I figured it would be an instance method that returns the new offset surface

setup.cfg
```diff
@@ -48,7 +48,7 @@ markers=
 filterwarnings=
     error
     ignore::pytest.PytestUnraisableExceptionWarning
-    ignore::RuntimeWarning:desc.compute
+    ignore::RuntimeWarning
```
f0uriest, line 51: Are there new warnings to ignore?

tests/test_surfaces.py
```diff
@@ -136,6 +137,140 @@ def test_curvature(self):
...
+        np.testing.assert_allclose(Zneg10_offset, a + offset, atol=1e-2)
+        np.testing.assert_allclose(Zneg11_offset, -b, atol=3e-2)
+        np.testing.assert_allclose(Z1neg1_offset, -b, atol=3e-2)
+
+        np.testing.assert_allclose(
+            R00_offset,
+            eq.surface.R_lmn[eq.surface.R_basis.get_idx(M=0, N=0)],
+            atol=1e-3,
+        )
+        # cannot do the same sort of distance test as the axisymmetric test
```
f0uriest, line 267: Could maybe compute aspect ratio and volume before and after?

desc/geometry/surface.py
dpanici: I think other should be merged first

desc/compute/_metric.py
```diff
@@ -152,6 +152,35 @@ def _e_theta_x_e_zeta_rr(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="|e_theta x e_zeta|_z",
+    label="|e_{\\theta} \\times e_{\\zeta}|",
```
ddudt, line 157: missing the zeta derivative in this label

desc/geometry/surface.py
ddudt: So this method is from the other PR? Or is this a separate implementation?

desc/geometry/surface.py
```diff
@@ -357,6 +361,200 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
...
+                ``x_offset_surface`` : (``grid.num_nodes`` x 3) array of the
+                    position vectors on the offset surface, corresponding to the
+                    ``x`` points on the base_surface (i.e. the points to which the
+                    offset surface was fit)
+        info : tuple
+            2 element tuple containing residuals and number of iterations
+            for each point. Only returned if ``full_output`` is True
+
+        """
+        if grid is None:
```
ddudt, line 478: would be good to include a test that covers the default grid

desc/geometry/surface.py
```diff
@@ -357,6 +361,200 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
...
+            re = rpz2xyz(coords)
+            # NOTE: when generalized toroidal angle is implemented
+            # this must include omega
+            e_theta = jnp.array([R_t, jnp.zeros_like(R_t), Z_t]).T
+            e_zeta = jnp.array([R_z, R, Z_z]).T
+            e_theta_cross_e_zeta = cross(e_theta, e_zeta)
+            n = safediv(
+                e_theta_cross_e_zeta.T, jnp.linalg.norm(e_theta_cross_e_zeta, axis=1)
+            ).T
+            n = rpz2xyz_vec(n, phi=phi)
```
ddudt, line 513: Why not use the usual compute interface for these quantities, like `base_surface.compute("|e_theta x e_zeta|", grid=Grid(nodes))`?

desc/geometry/surface.py
```diff
@@ -357,6 +361,200 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
...
+            n, r, r_offset = n_and_r_jax(nodes)
+            return jnp.arctan(r_offset[0, 1] / r_offset[0, 0]) - zeta
+
+        vecroot = jit(
+            vmap(
+                lambda x0, *p: root_scalar(
+                    fun_jax,
+                    x0,
+                    jac=None,
+                    args=p,
```
ddudt, line 528: Formatting note: if you remove the comma here, I think this whole lambda function will fit on a single line

desc/geometry/surface.py
```diff
@@ -357,6 +361,200 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
...
+                )
+            )
+        )
+        zetas, (res, niter) = vecroot(
+            grid.nodes[:, 2], grid.nodes[:, 1], grid.nodes[:, 2]
```
ddudt, line 533: Are the same values being passed in for both `zeta_hat` and `zeta` to the function `fun_jax`? If so, why bother naming them as separate variables in the function?

setup.cfg
ddudt: we shouldn't ignore all runtime warnings unless there is a good reason we can't avoid them

tests/test_surfaces.py
```diff
@@ -136,6 +137,140 @@ def test_curvature(self):
         np.testing.assert_allclose(data["curvature_k2_rho"], -1)
 
     @pytest.mark.unit
+    def test_constant_offset_surface_circle(self):
+        """Test constant offset algorithm for a circular torus."""
+        s = FourierRZToroidalSurface()
+        grid = LinearGrid(M=3, N=2)
```
ddudt, line 143: Maybe increase the resolution of the surface and then let it use the default grid? That should increase test coverage (see previous comment).

tests/test_surfaces.py
```diff
@@ -136,6 +137,140 @@ def test_curvature(self):
...
+                err_msg=f"Failed test at comparison of {key}",
+            )
+
+    @pytest.mark.slow
+    @pytest.mark.unit
+    def test_constant_offset_surface_rot_ellipse(self):
+        """Test constant offset algorithm for a rotating ellipse."""
+        eq = desc.examples.get("HELIOTRON")
+        s = eq.surface
+        grid = LinearGrid(M=4, N=4, NFP=eq.NFP)
```
ddudt, line 215: See previous comment. Maybe best to use the default grid for one test and a custom grid for the other test so both cases are covered.

ddudt (changes requested):

setup.cfg
dpanici: Meant to revert this, thanks for catching. It was warnings I would get from the scipy root solver but we dont get those anyomre now

desc/geometry/surface.py
dpanici: That other PR is a cherrypick from this PR, now that that one is merged to master, this PR is using the updated method implementation from the master branch

desc/geometry/surface.py
dpanici: we can't `jit` it if we go through the compute interface as the creation of Transform objects is not jittable

desc/geometry/surface.py
dpanici: They are separate variables int he function, one is the zeta on the offset surface, the other is the zeta on the base surface.

Initially we do not know the offset surface zeta, so we use the base surface zeta as an initial guess, hence why here we pass it in as the `x0` for the rootfind

tests/test_surfaces.py
dpanici: added a XS surface area test

desc/geometry/surface.py
dpanici: hm that is a good point, I guess it makes sense that you'd already have the surface and you just want to make a new one based off of it.

desc/geometry/surface.py
f0uriest: It is if you make method="jitable"

dpanici: That works and I made the change, needed a bug fix

I will say that the previous method is about 10-20% faster than going through the compute function, but that is not a huge change.

f0uriest (approved):

desc/geometry/surface.py
```diff
@@ -490,6 +491,125 @@ def from_values(
...
+            dictionary containing  the following data, in the cylindrical basis:
+                ``n`` : (``grid.num_nodes`` x 3) array of the unit surface normal on
+                    the base_surface evaluated at the input ``grid``
+                ``x`` : (``grid.num_nodes`` x 3) array of the position vectors on
+                    the base_surface evaluated at the input ``grid``
+                ``x_offset_surface`` : (``grid.num_nodes`` x 3) array of the
+                    position vectors on the offset surface, corresponding to the
+                    ``x`` points on the base_surface (i.e. the points to which the
+                    offset surface was fit)
+        info : tuple
```
ddudt, line 548: Is there a use case where this info is necessary?

ddudt (approved):

desc/geometry/surface.py
dpanici: if a given surface is kinda crappy it could be useful, I just kept it here to mirror the information given from the rootfinding function

PR #630, opened 2023-08-21
alphabetize

ddudt:
No code is changed, only sorting the files in alphabetical order. The following rules are applied:

1. Classes first, then functions
2. Underscores come before numbers
3. Numbers come before letters

The only exception is that getter & setter methods are at the end of a class, instead of being mixed in with the other methods. 

To-do:

- [ ] compute
- [x] equilibrium
- [x] geometry
- [x] io
- [x] objectives
- [ ] optimize
- [ ] other desc/
- [ ] tests


f0uriest:
OK, spent the past few days playing with this and the script to alphabetize things works fine, but we still get tons of merge conflicts, even if we sort both branches first. Git sees that both branches modified the same lines, even if they're modified to be the same. Resolving all the merge conflicts is a huge pain, and even when I was super careful I still missed stuff and either lost changes or included changes that shouldn't have been.

I'm going to vote that we shelve this, as I don't think the benefit is worth all the aggravation, especially when we can just `Ctrl-F` to find whatever function/method etc we want in a given file. For now I'm going to unsort #484 so we can get that in cleanly.

PR #632, opened 2023-08-22
JITable grid and transform creation

### LINKED ISSUE #608: Plot_Surfaces is very slow when plotting vartheta contours
I've noticed that plot_surfaces is noticeably slower than it was in the mid-past (a few months ago), when the vartheta contours are plotted. Wonder if a change to how those are calculated is the cause (like in map coords or something)

f0uriest:
Adds capability to create `Grid` and `Transform` objects inside a `jit`'ed function

Improves the speed of `map_coordinates` by ~2-10x in local testing.

This PR also switches the evaluation of Zernike polynomials to always use the Jacobi recurrence relation. Previously this was used for L>24 since it is more accurate, with Horner's method used for L<24 since it was faster. Recent improvements in JAX seem to have made the recurrence relation faster, and further testing showed that Horner's method could produce appreciable errors (O 1e-5) on the higher order derivatives (like those needed for $F$ and $J$) even at L<24.

Resolves #608


ddudt (approved):

desc/grid.py
```diff
@@ -51,30 +56,47 @@ class Grid(IOAble):
...
-            self._unique_rho_idx,
-            self._inverse_rho_idx,
-            self._unique_theta_idx,
-            self._inverse_theta_idx,
-            self._unique_zeta_idx,
-            self._inverse_zeta_idx,
-        ) = self._find_unique_inverse_nodes()
+        if jitable:
+            # dont do anything with symmetry since that changes # of nodes
+            # avoid point at the axis, for now. FIXME: make axis boolean mask?
```
unalmis, line 71: The limits in the compute function are jitable because they rely on the `axis.size` parameter which I think is a compile time constant. If you switch this to boolean mask you may need to add a `has_axis` boolean because `jnp.any(axis_mask)` may make jit complain.

f0uriest: Yeah I still don't have a good idea of how to do dynamic grids that may or may not have a point on the axis, so for now just fudging it off the axis which should be fine for 99.999% of cases.

ddudt (approved):

desc/equilibrium/coords.py
```diff
@@ -82,17 +86,27 @@ def map_coordinates(  # noqa: C901
...
 
-    # can't use AD or jit compile this because of grid stuff
+    params = get_params(inbasis + basis_derivs, eq)
+
+    @functools.partial(jit, static_argnums=1)
     def compute(y, basis):
-        data = eq.compute(basis, grid=Grid(y, sort=False))
+        grid = Grid(y, sort=False, jitable=True)
+        profiles = get_profiles(inbasis + basis_derivs, eq, grid, jitable=True)
+        # do surface average to get iota once
```
dpanici, line 95: maybe I dont understand this logic, it seems to me that if a current profile constrained eq is  used, it will get iota every time?

desc/transform.py
```diff
@@ -377,6 +380,12 @@ def build(self):
                     self.grid.nodes, d, unique=True
                 )
 
+        if self.method == "jitable":
```
dpanici, line 383: what is this doing that makes it jitable? precomputing everything?

dpanici (approved):

desc/equilibrium/coords.py
f0uriest: by "once" mean per iteration, as opposed to the standard method which would do the averaging once for `iota` and once for `iota_r` etc.

desc/transform.py
f0uriest: its the `unique=False` part that makes it jitable. Possibly more expensive since theres some duplicate work if the are duplicate nodes, but this is usually for unstructured grids so it should be fine.

unalmis (approved):

PR #638, opened 2023-08-25
Fix bounds_scaled for profile objectives

### LINKED ISSUE #637: Scaling with quadrature weights messes up profile constraints
I'm running an optimization, using `RotationalTransform` as a constraint:
`RotationalTransform(eq=eqfam[-1], bounds=(0.41, 0.43)),`

However, the resulting rotational transform I get is more like 1.7, the issue is seemingly because the profile objectives get multiplied by `sqrt(w)`, which in this case is `0.242`, so effectively its targeting iota of `0.4/0/242 ~ 1.7`

When used as a constraint, we use `Objective.compute_scaled` because we often want to deal with constraints in normalized units to keep everything a similar order of magnitude. However, this also applies quadrature weights which may mess up bounds/targets for non-scalar objectives with non-zero targets/bounds.

It's not an issue if we use it as an objective, because in that case we call `Objective.compute_scaled_error` which applies the target/bounds to the unscaled values, then applies the scaling

Possibly related to #484 ?

f0uriest:
Resolves #637

### DIFF
--- a/desc/objectives/objective_funs.py
+++ b/desc/objectives/objective_funs.py
@@ -707,8 +707,9 @@ class ObjectiveFunction(IOAble):
             else:
                 # need to return something, so use midpoint of bounds as approx target
                 target_i = jnp.ones(obj.dim_f) * (obj.bounds[0] + obj.bounds[1]) / 2
-            if obj._normalize_target:
-                target_i /= obj.normalization
+            target_i = obj._scale(target_i)
+            if not obj._normalize_target:
+                target_i *= obj.normalization
             target += [target_i]
         return jnp.concatenate(target)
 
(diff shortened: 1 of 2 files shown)


dpanici (approved):

desc/objectives/objective_funs.py
```diff
@@ -723,9 +724,11 @@ def bounds_scaled(self):
...
                 lb_i = jnp.ones(obj.dim_f) * obj.target
                 ub_i = jnp.ones(obj.dim_f) * obj.target
-            if obj._normalize_target:
-                lb_i /= obj.normalization
-                ub_i /= obj.normalization
+            lb_i = obj._scale(lb_i)
+            ub_i = obj._scale(ub_i)
+            if not obj._normalize_target:
+                lb_i *= obj.normalization
+                ub_i *= obj.normalization
```
unalmis, line 731: From the linked issue it's clear that it is undesirable to multiply the objective by the quadrature weighs because the target and bounds for the objective were unaware of this scaling. So the solution used here is to also scale the target and bounds by the quadrature weights through the call to `obj._scale`.

`obj._scale` first calls [this method](https://github.com/PlasmaControl/DESC/blob/9428bdaab2ddf50351131b72d258fab362e92828/desc/objectives/_generic.py#L670) for the rotational transform objective in particular (applying the quadrature weight `w`), then in general calls this [method](https://github.com/PlasmaControl/DESC/blob/a2fe711ffa3f3de459ba65e2d5520a91f98841a4/desc/objectives/objective_funs.py#L986) which applies objective weighting.

The changes in `test_objective_funs.py` of multiplying the target and bounds by the objective weights are there now to account for this change. It looks like the target/bounds were previously also unaware of the objective weights which could cause a similar issue to the one discussed in the linked issue.

unalmis (approved):

PR #639, opened 2023-08-26
Fix docs bulding locally

### LINKED ISSUE #636: Update Traverse/Power9 JAX Installation to Work with newer JAX
Our current Traverse/Power9 Install instructions install an old (0.2.19 which is way below where we are at master now)

This is because it is hard to install on power9 architectures

There may be a way to install newer versions, using this conda channel, so should attempt this and update install instructions if successful https://github.com/open-ce/open-ce/issues/549

https://github.com/google/jax/issues/4493

unalmis:


dpanici:
> Maybe also a good branch to resolve #636

After spending a couple hours, I don't think I can get #636 resolved in a reasonable amount of time, so I would not wait on it for this PR

dpanici:
While you are changing the docs/labels, could you fix a few of the labels where we have \\partial{coord} instead of \\partial_{coord}? `e^rho_r` and `e^rho_theta` are two examples of this 
image

something as simple as doing a ctrl-f on "\\partial{\\rho}" and then replace all found instances with "\\partial_{\\rho}" should work I think

dpanici:
Same thing for some of the higher derivatives
image

desc/compute/_basis_vectors.py
```diff
@@ -115,11 +115,7 @@ def _e_sup_rho_t(params, transforms, profiles, data, **kwargs):
             / data["sqrt(g)"] ** 2
         ).T,
         lambda: (
-            (
-                cross(data["e_theta_r"], data["e_zeta_t"])
```
unalmis, line 119: This is always 0 anyway so I just remove it.

dpanici: why is it always zero? And would that be true even with a zeta that is not the cylindrical toroidal angle?

unalmis: Yes because e_zeta_t == e_theta_z (i.e. $\partial_{\theta \zeta} \mathbf{e} = \partial_{\zeta \theta} \mathbf{e}\$) and e_theta = 0 everywhere on the magnetic axis.

dpanici: In the cylindrical case: $e_{\theta} = [R_{\theta}, 0, Z_{\theta}]$ while $e_zeta = [R_{\zeta}, R, Z_{\zeta}]$
$\partial_{\zeta}e_{\theta} = [R_{\theta,\zeta}, 0, Z_{\theta,\zeta}]$ while $\partial_{\theta}e_{\zeta} = [R_{\theta,\zeta}, R_{\theta}, Z_{\theta,\zeta}]$

dpanici: Hm ok I see. Thanks for clarifying for me

dpanici (approved):

f0uriest (approved):

PR #640, opened 2023-08-29
Refactor optimizer subproblems and utilities to use jax control flow

f0uriest:
No real functionality change, just refactors a bunch of code to be jitable to hopefully use more of the GPU.


dpanici (approved):

tests/test_optimizer.py
```diff
@@ -868,6 +868,7 @@ def con(x):
 
 @pytest.mark.slow
 @pytest.mark.regression
+@pytest.mark.xfail
```
ddudt, line 871: Why do these tests fail now after the changes?

ddudt (approved):

tests/test_optimizer.py
f0uriest: These tests are apparently super sensitive to minor numerical differences but should be fixed in #644 soon

Issue #587, opened 2023-07-17
Fourth order derivatives of basis vectors


unalmis:
#556 Adds third order derivatives of the basis vectors needed to compute the magnetic axis limits of quantities.
The axis limit of the second derivative of the rotational transform `iota_rr` requires the fourth radial derivative of the 3D volume Jacobian `sqrt(g)_rrrr` which itself requires fourth order derivatives of basis functions to compute. Should use the basis vectors that have been generalized to include toroidal stream function `omega` to avoid repeating work in the future.

PR #642, opened 2023-08-29
iota vacuum and current contributions

ddudt:


desc/compute/_profiles.py
```diff
@@ -607,6 +607,97 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
+def _iota_vacuum_num(params, transforms, profiles, data, **kwargs):
+    """Vacuum contribution to the numerator of rotational transform formula."""
+    data["iota_vacuum_num"] = surface_integrals(
+        transforms["grid"],
+        (data["lambda_z"] * data["g_tt"] - (1 + data["lambda_t"]) * data["g_tz"])
+        / data["sqrt(g)"],
+    )
+    data["iota_vacuum_num"] = transforms["grid"].replace_at_axis(
+        data["iota_vacuum_num"], 0
+    )
```
ddudt, line 697: I'm not certain what the proper limit should be here. It probably should not be 0, but it doesn't really matter because the proper limit get used for `iota_vacuum` which is what we ultimately care about.

ddudt:
I left the existing `iota` computation alone. Technically `alpha` is now redundant with `iota_current_num` and `beta` is redundant with `iota_vacuum_num`, so there is some repeated code. But I want to be able to compute `iota_vacuum` and `iota_current` even when the iota profile is supplied. So I think there would be circular dependencies otherwise -- but let me know if you see a way to resolve that.

unalmis:
1. We should remove these guards from `iota_den`, `iota_den_r`, `iota_den_rr` so that we can also compute them when an iota profile is supplied instead of a current profile. https://github.com/PlasmaControl/DESC/blob/c0b44414883d202fb30c94e78fbbd5e7d29b59d7/desc/compute/_profiles.py#L1000
2. Heads up to avoid fighting the tests: Remove the `current` dependency on the `iota_den` stuff. We only used them for the above checks. https://github.com/PlasmaControl/DESC/blob/c0b44414883d202fb30c94e78fbbd5e7d29b59d7/desc/compute/_profiles.py#L990
3. We should be able to remove `iota_den` from this line https://github.com/PlasmaControl/DESC/blob/c0b44414883d202fb30c94e78fbbd5e7d29b59d7/tests/test_axis_limits.py#L105

f0uriest:
> 1. We should remove these guards from `iota_den`, `iota_den_r`, `iota_den_rr` so that we can also compute them when an iota profile is supplied instead of a current profile. https://github.com/PlasmaControl/DESC/blob/c0b44414883d202fb30c94e78fbbd5e7d29b59d7/desc/compute/_profiles.py#L1000
> 
>     2. Heads up to avoid fighting the tests: Remove the `current` dependency on the `iota_den` stuff. We only used them for the above checks. https://github.com/PlasmaControl/DESC/blob/c0b44414883d202fb30c94e78fbbd5e7d29b59d7/desc/compute/_profiles.py#L990
> 
>     3. We should be able to remove `iota_den` from this line https://github.com/PlasmaControl/DESC/blob/c0b44414883d202fb30c94e78fbbd5e7d29b59d7/tests/test_axis_limits.py#L105

I think we might not want to do your first point. Because `iota_den` etc are dependencies for `iota`, they get computed even if the iota profile is supplied, but because they are hardcoded to nan the compiler should skip that. If we remove this it may lead to redundant calculations for standard fixed iota cases

desc/compute/_profiles.py
```diff
@@ -607,6 +607,97 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
+    coordinates="r",
+    data=["current", "psi_r"],
+)
+def _iota_current_num(params, transforms, profiles, data, **kwargs):
+    """Current contribution to the numerator of rotational transform formula."""
+    # 4π^2 I = 4π^2 (mu_0 current / 2π) = 2π mu_0 current
+    data["iota_current_num"] = 2 * jnp.pi * mu_0 * data["current"] / data["psi_r"]
+    data["iota_current_num"] = transforms["grid"].replace_at_axis(
+        data["iota_current_num"], 0
+    )
```
unalmis, line 671: Can you add a comment like this here:
```python
# Assume current may be expanded as an even power series of ρ so that
# ∂ᵨ current = 0 at axis.
```

desc/compute/_profiles.py
unalmis: It's fine to put as 0. I think we're basically just assuming that at the time this function is called the configuration is in a physical state (current_r = 0 at axis).
Do you want to add a comment like this:
```python
# Assume current may be expanded as an even power series of ρ so that
# ∂ᵨ current = 0 at axis. Then 0 = lim iota_num = lim iota_current_num + lim iota_vacuum_num
# and the latter limit is also finite because the integrand is finite over the magnetic axis.
```

desc/compute/_profiles.py
```diff
@@ -607,6 +607,97 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="r",
+    data=["iota_den", "iota_current_num"],
+)
+def _iota_current(params, transforms, profiles, data, **kwargs):
+    data["iota_current"] = transforms["grid"].replace_at_axis(
+        data["iota_current_num"] / data["iota_den"], 0
+    )
```
unalmis, line 626: I misled you earlier. For this limit to be 0, we would need to make the invalid assumption that $\partial_{\rho \rho} I = 0$ at axis. We can only assume the first derivative is 0 at axis.

The limit of this one should be `alpha_r / iota_den_r` where `alpha_r` is the same thing as `iota_current_num_r`. See equation 48 in the document here: https://github.com/PlasmaControl/DESC/pull/556#issue-1773649567. It's the first term in the numerator divided by the denominator.

desc/compute/_profiles.py
```diff
@@ -607,6 +607,97 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="r",
+    data=["iota_den", "iota_vacuum_num"],
+    axis_limit_data=["iota_den_r", "iota_num_r"],
+)
+def _iota_vacuum(params, transforms, profiles, data, **kwargs):
+    data["iota_vacuum"] = transforms["grid"].replace_at_axis(
+        data["iota_vacuum_num"] / data["iota_den"],
+        lambda: data["iota_num_r"] / data["iota_den_r"],
```
unalmis, line 647: Similar to https://github.com/PlasmaControl/DESC/pull/642/files#r1309419402, this one's limit should be `beta_r / iota_den_r` where `beta_r` is the same thing as `iota_vacuum_num_r`.

Even though the enclosed current  `I` is 0 at the magnetic axis, for it to not contribute to the rotational transform at axis it needs to approach 0 fast - in particular fast enough that both $\partial_{\rho} I$ and $\partial_{\rho \rho} I$ are 0 axis. $\partial_{\rho \rho} I$ need not be zero at axis, so both iota_current and iota_vacuum may be nonzero at axis.

desc/compute/_profiles.py
ddudt: where is this document?

desc/compute/_profiles.py
```diff
@@ -701,7 +884,7 @@ def _iota_num_r(params, transforms, profiles, data, **kwargs):
             / data["psi_r"] ** 2,
             lambda: (
                 profiles["current"].compute(params["c_l"], dr=2) * data["psi_rr"]
-                - current_r * data["psi_rrr"]
+                - current_r * data["psi_rrr"]  # FIXME: psi_rrr=0
```
ddudt, line 887: `psi_rrr` (and higher derivatives) are always 0. Can we remove all references to them from the compute functions to simplify the code?

desc/compute/_profiles.py
ddudt: FYI the axis limits of `iota_current_num` and `iota_vacuum_num` never get used when computing `iota_current` and `iota_vacuum`, but they are still nice to have in case you want to plot these quantities.

desc/compute/_profiles.py
ddudt: The (non-zero) correct axis limit should be implemented now.

desc/compute/_profiles.py
unalmis: It's at the top of the description of Github Pull request 558. titled math to compute rotational transform in desc

unalmis:
I prefer names like `iota_from_current` and `iota_in_vacuum` by the way.

To compute `iota_from_current` and its derivatives, you should be able to just copy the compute functions for `iota` and replace all references to `iota_num` with the term `alpha` defined in the compute functions for `iota_num`. (similar with `iota_num_r` and `alpha_r`). This will also give you the correct limits.

Same thing for `iota_in_vacuum` except use the term `beta` instead.


Currently on master, the only physics constraint I used to compute these limits is that the magnetic field physical. (This was necessary because there is an if and only if relationship between this and the limit of iota being finite).

FYI, there are two other perhaps weaker physics conditions which can make computing the limits easier. These conditions are that $\partial_{\rho} iota$ and $\partial_{\rho} I$ are both zero at axis. I didn't explicitly enforce these and computed the limits in the more general manner because 
1. whatever shortcuts they provide don't help when computing the limit of the second derivative of $\iota$.
2. didn't want to assume behavior of user-inputted quantities

ddudt:
The new quantities are working except for the axis limit of iota from the current. I think I understand why: 

I changed the references from things like `2 * jnp.pi * mu_0 * profiles["current"].compute(params["c_l"], dr=0)` to instead be like `4 * jnp.pi**2 * data["I"]`. This moves away from depending on the current profile as an input, so that we could compute these new quantities even when the user gives an iota profile instead. This also better matches Kaya's equations. 

The axis limits then depend on the value of `data["I_rr"]` instead of `profiles["current"].compute(params["c_l"], dr=2)`. The later is always well defined at the axis by the input, but the former is computed as a flux surface average of $B_\theta$. It appears that in our code this always evaluates to `I_rr = 0` at the axis, even though it should generally have a finite value. So we need to implement the proper axis limit for that quantity to finish this work.

unalmis:
> The new quantities are working except for the axis limit of iota from the current...

The test was failing because `iota_den` was hardcoded to return `jnp.nan` on fixed iota equilibrium. Commiting the the suggestions here https://github.com/PlasmaControl/DESC/pull/642#issuecomment-1697881035 resolve the issue.

> The axis limits then depend on the value of `data["I_rr"]` ... is computed as a flux surface average of Bθ. It appears that in our code this always evaluates to `I_rr = 0` at the axis, even though it should generally have a finite value. So we need to implement the proper axis limit for that quantity to finish this work.

The computation of $\partial_{\rho \rho} I = \partial_{\rho \rho} \langle B_{\theta} \rangle$ is dependent on $\partial_{\rho \rho} \iota$. All the math is done to compute $\lim_{\rho \to 0} \partial_{\rho \rho} \iota$, but we can't implement it until #587 is resolved. This means the integrand: $\partial_{\rho \rho}  B_{\theta}$ evaluates as `jnp.nan` when the iota profile is not given as a known power series. The surface integral replaces `nan` integrand with zero (see [here](https://github.com/PlasmaControl/DESC/blob/4ced2db2788476af0b21581d8a06aedb62f259f4/desc/compute/utils.py#L852)).

desc/compute/_profiles.py
```diff
@@ -1065,13 +1236,13 @@ def _iota_den_r(params, transforms, profiles, data, **kwargs):
             data["omega_t"]
             * data["g_tz_r"]
             * data["sqrt(g)_rr"]
-            / (2 * data["sqrt(g)_r"] ** 2)
+            / (2 * data["sqrt(g)_r"] ** 2)  # XXX: this factor of 2 not in Eq. 48
```
unalmis, line 1239: This computes what the docstring says. The extra factor of 2 is in $d \alpha / d \rho$, $d \beta / d \rho$, and $d \gamma / d \rho$, and hence cancels out in [equation 48](https://github.com/PlasmaControl/DESC/files/12137234/rotational_transform_and_limit.pdf). Since the code computes equation 49, it is simplest to keep the factors of 2 and compute the limit as in equation 50.

unalmis:
> > I think we might not want to do your first point. Because `iota_den` etc are dependencies for `iota`, they get computed even if the iota profile is supplied, but because they are hardcoded to nan the compiler should skip that. If we remove this it may lead to redundant calculations for standard fixed iota cases
> 
> But Daniel wants to be able to compute `iota_vacuum` and `iota_current` even when the iota profile is supplied. These quantities require `iota_den`

If Rory is okay with [6e9b9e2](https://github.com/PlasmaControl/DESC/pull/642/commits/6e9b9e2a68cbbf8053664985ce96eca7392a11b7) then the [master data ](https://github.com/PlasmaControl/DESC/blob/4ced2db2788476af0b21581d8a06aedb62f259f4/tests/test_compute_funs.py#L1271)will need to be updated in this PR since `iota_den` will no longer be `nan`

f0uriest:
I'm generally fine with the above, it would be good if you did some basic benchmarking to see if it slows things down appreciably, but I doubt there will be a noticeable difference for most use cases

ddudt:
This should be working now, I just need to add tests. 

Everything can be computed fine when the equilibrium has a current profile assigned. The only thing that can't be computed from an iota profile is `iota current`, but you can still get that data manually by doing `data["iota"] - data["iota vacuum"]`.

ddudt:
> This should be working now, I just need to add tests.
> 
> Everything can be computed fine when the equilibrium has a current profile assigned. The only thing that can't be computed from an iota profile is `iota current`, but you can still get that data manually by doing `data["iota"] - data["iota vacuum"]`.

I think I figured out a workaround so you can now compute everything for all types of equilibria. This PR should be ready to go

desc/compute/_profiles.py
```diff
@@ -638,59 +638,161 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
+    current = profiles["current"].compute(transforms["grid"], params["c_l"], dr=0)
+    current_r = profiles["current"].compute(transforms["grid"], params["c_l"], dr=1)
+    current_rr = profiles["current"].compute(transforms["grid"], params["c_l"], dr=2)
+    data["iota_num_r current"] = (
+        2
+        * jnp.pi
+        * mu_0
+        * transforms["grid"].replace_at_axis(
+            (current_r * data["psi_r"] - current * data["psi_rr"]) / data["psi_r"] ** 2,
+            lambda: current_rr / (2 * data["psi_rr"]),
```
unalmis, line 777: just linking to document change https://github.com/PlasmaControl/DESC/pull/642#discussion_r1310570548

unalmis (approved):

desc/compute/_profiles.py
```diff
@@ -638,59 +638,161 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
-
-    Computes 𝛼 + 𝛽 as defined in the document attached to the description
-    of GitHub pull request #556. 𝛼 supplements the rotational transform with an
-    additional term to account for the enclosed net toroidal current.
-    """
+def _iota_num_current(params, transforms, profiles, data, **kwargs):
+    """Current contribution to the numerator of rotational transform formula."""
     if profiles["current"] is None:
-        data["iota_num"] = jnp.nan * data["0"]
+        data["iota_num current"] = jnp.nan * data["0"]
```
f0uriest, line 705: couldn't this be computed as `iota*iota_den-iota_vac`?

desc/compute/_profiles.py
```diff
@@ -638,59 +638,161 @@ def _iota_rr(params, transforms, profiles, data, **kwargs):
...
     dim=1,
     params=["c_l"],
     transforms={"grid": []},
     profiles=["current"],
     coordinates="r",
+    data=["0", "psi_r", "psi_rr"],
+)
+def _iota_num_r_current(params, transforms, profiles, data, **kwargs):
+    if profiles["current"] is None:
+        data["iota_num_r current"] = jnp.nan * data["0"]
```
f0uriest, line 764: couldn't this be computed from `iota`, `iota_num_vac_r` etc?

desc/compute/_profiles.py
```diff
@@ -763,8 +832,63 @@ def _iota_num_r(params, transforms, profiles, data, **kwargs):
...
+)
+def _iota_num(params, transforms, profiles, data, **kwargs):
+    """Numerator of rotational transform formula.
+
+    Computes 𝛼 + 𝛽 as defined in the document attached to the description
+    of GitHub pull request #556. 𝛼 supplements the rotational transform with an
+    additional term to account for the enclosed net toroidal current.
+    """
+    if profiles["current"] is None:
+        data["iota_num"] = jnp.nan * data["0"]
```
f0uriest, line 860: I think if the above fixes are made the following formula works in all cases

f0uriest (approved):
Fine as is (assuming the tests pass and the new code is covered) but may want to consider implementing stuff where we now return NaN, I don't think it would be too complicated

f0uriest:
also make sure to update the changelog

ddudt:
> Fine as is (assuming the tests pass and the new code is covered) but may want to consider implementing stuff where we now return NaN, I don't think it would be too complicated

I implemented these, but you should check my formulas. The logic is a little tricky to avoid circular dependencies. Also these quantities will probably never actually be called by the user, since `"iota_num"` etc. are only intermediate outputs to compute `"iota"`, etc.

desc/compute/_profiles.py
```diff
@@ -753,30 +753,35 @@ def _iota_num_vacuum(params, transforms, profiles, data, **kwargs):
...
-        * jnp.pi
-        * mu_0
-        * transforms["grid"].replace_at_axis(
-            (current_r * data["psi_r"] - current * data["psi_rr"]) / data["psi_r"] ** 2,
-            lambda: current_rr / (2 * data["psi_rr"]),
+    if profiles["iota"] is not None:
+        iota_r = profiles["iota"].compute(transforms["grid"], params["i_l"], dr=1)
+        data["iota_num_r current"] = (
+            iota_r * data["iota_den_r"] - data["iota_num_r vacuum"]
+        )
```
unalmis, line 767: Equations 52 and 56 in the document attached to the top of GitHub PR #556 are helpful here. Recall "iota_num current" is "alpha" and "iota_num vacuum" is "beta" and "iota_den" is "gamma".
Here "iota_num_r current" is $d \alpha / d \rho$. Given $d \iota / d\rho$, we can compute that as
```python
data["iota_num_r current"] = (
    iota_r * data["iota_den"]**2
    + data["iota_num"] * data["iota_den_r"] 
) / data["iota_den"] - data["iota_num_r vacuum"]
```

desc/compute/_profiles.py
```diff
@@ -933,79 +930,82 @@ def _iota_num_rr(params, transforms, profiles, data, **kwargs):
     of GitHub pull request #556. 𝛼 supplements the rotational transform with an
     additional term to account for the enclosed net toroidal current.
     """
-    if profiles["current"] is None:
+    if profiles["iota"] is not None:
```
unalmis, line 933: This should trip a test unless you add `iota` as a profile dependency

desc/compute/_profiles.py
```diff
@@ -1054,84 +1054,92 @@ def _iota_num_rrr(params, transforms, profiles, data, **kwargs):
...
```
unalmis, line 1057: Same thing with adding another profile dependency.

unalmis (changes requested):

unalmis (dismissed):

f0uriest (approved):

f0uriest (approved):

dpanici:
wasany benchmarking done? like how long it takes to to cmpute iota here versus master?

dpanici:
> was any benchmarking done? like how long it takes to to cmpute iota here versus master?

I did htis quickly for the DSHAPE and DSHAPE_current examples, seems it is as expected, in that an equilibrium with an iota profile and an equilibrium with a current profile now take the same time to calculate iota (because the iota profile eq still computes the iota contributions from the geometry/current of the equilibrium, even though it is not necessary)

for the DSHAPE this is a slowdown of about 2X but since in optimization current profiles are almost always used this is fine. the current profile eq's iota compute speed has not changed significantly, it is now slightly slower but only by like 5%

PR #644, opened 2023-08-29
Refactor augmented lagrangian methods to better handle stopping criteria

f0uriest:
- Previously we would use `lsqtr` or `fmintr` as the inner loop of the augmented lagrangian optimizers. This would occasionally cause issues since the inner loop and outer loop could have different stopping criteria, and there's not an easy way to get all the info the outer loop needs from the inner loop, in addition to having to do some hacky things to maintain state between calls to the inner loop optimizers. This would occasionally lead to the optimizer stalling and basically taking the same step over and over.
- This refactors a bunch of things but the main part is copying the core part of `lsqtr` and `fmintr` to the augmented lagrangian methods but modifying it so that tolerances are consistent and both the inner and outer loop are seeing the same state.
- Allows the quadratic penalty parameter to be a vector so that constraints of different magnitude can be scaled differently
- Adds `callback` to augmented lagrangians
- Correctly handles `constants` passed into augmented lagrangian optimizers
- Cleans up links and other tidbits in optimizer documentation


desc/optimize/_desc_wrappers.py
```diff
@@ -13,9 +13,12 @@
 @register_optimizer(
     name=["fmin-auglag", "fmin-auglag-bfgs"],
     description=[
-        "Augmented Lagrangian method with trust region subproblem.",
-        "Augmented Lagrangian method with trust region subproblem. Uses BFGS to"
-        + " approximate hessian",
+        "Augmented Lagrangian trust region method for minimizing scalar valued "
+        + "multivariate function. "
+        + "See https://desc-docs.readthedocs.io/en/stable/_api/optimize/desc.optimize.fmin_auglag.html",  # noqa: E501
+        "Augmented Lagrangian trust region method for minimizing scalar valued "
```
dpanici, line 19: duplicated line?

dpanici (approved):

rahulgaur104:
Should we worry about the fact that the lsq-auglag optimizer doesn't run with Daniel's omnigenity code right now? That may become a problem in the near future.

f0uriest:
> Should we worry about the fact that the lsq-auglag optimizer doesn't run with Daniel's omnigenity code right now? That may become a problem in the near future.

Pretty sure that's an issue with the omnigenity stuff which still needs a lot of work to get into master. Right now omnigenity tests don't even pass on their own (and there's a bunch of merge conflicts)

dpanici:
Update ftol default to 1e-4 etc

dpanici (approved):

rahulgaur104 (dismissed):

dpanici (approved):

rahulgaur104 (approved):

PR #646, opened 2023-08-31
Fix spelling errors, rename abstract base classes

f0uriest:
- Fixes a bunch of typos and spelling errors to shut up my spell checker
- Adds an underscore to abstract base class names so they get sorted first so inheritance works correctly
- Splits `Grid` class into a true ABC `_Grid` that `Grid` now inherits from.
- Fixes a bunch of misc spacing issues that were causing the alphabetizing script trouble.

### DIFF
--- a/desc/coils.py
+++ b/desc/coils.py
@@ -14,12 +14,11 @@ from desc.geometry import (
     FourierXYZCurve,
     SplineXYZCurve,
 )
-from desc.grid import Grid
-from desc.magnetic_fields import MagneticField, biot_savart
+from desc.magnetic_fields import _MagneticField, biot_savart
 from desc.utils import flatten_list
 
 
-class Coil(MagneticField, ABC):
+class _Coil(_MagneticField, ABC):
     """Base class representing a magnetic field coil.
 
     Represents coils as a combination of a Curve and current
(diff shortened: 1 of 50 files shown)


tests/conftest.py
```diff
@@ -15,16 +12,6 @@
 from desc.vmec import VMECIO
 
 
-# print full tracebacks to help find sources of warnings
-def _warn_with_traceback(message, category, filename, lineno, file=None, line=None):
```
f0uriest, line 19: because we raise warnings to errors, this isnt needed anymore.

ddudt (approved):

dpanici:
Can we add the spell checker to the pre-commit? like https://github.com/codespell-project/codespell

f0uriest:
> Can we add the spell checker to the pre-commit? like https://github.com/codespell-project/codespell

Probably but would prefer to do another PR for that since we'll need to come up with a list of words and patterns to ignore (think like all the vmec variable names ...)

desc/coils.py
```diff
@@ -77,7 +76,7 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
             magnetic field at specified points, in either rpz or xyz coordinates
         """
         assert basis.lower() in ["rpz", "xyz"]
-        if isinstance(coords, Grid):
+        if hasattr(coords, "nodes"):
```
dpanici, line 79: is `hasattr` instead of `isinstance` better in some way, or is this just a cleaner way? Why this change is what I am asking

f0uriest: In this case it is "better" since it's what we actually need from the object. In general I was just trying to avoid importing things just for instance checking where possible

PR #648, opened 2023-09-05
Add aliases for data_index

kianorr:
Provides API for aliases of computed quantities. Additionally finds permutations of some specified quantities and assigns them as aliases. The aliases are then added to `data_index`


unalmis:
Could be nice to also do this for the covariant basis vectors. Currently things like `e_zeta_rtt` and `e_rho_ttz` have their own compute functions, but they should always be equal because the partial derivatives commute. Something like 
```python
data["e_zeta_rtt"] = data["e_rho_ttz"]
return data
```
would make the code more maintainable.

desc/compute/data_index.py
```diff
@@ -13,6 +59,7 @@ def register_compute_fun(
     profiles,
     coordinates,
     data,
+    aliases=[],
```
dpanici, line 62: should add this to the docstring of this decorator

desc/compute/_metric.py
```diff
@@ -1306,7 +1306,7 @@ def _g_sup_zz(params, transforms, profiles, data, **kwargs):
     coordinates="rtz",
     data=["e^rho", "e^theta"],
 )
-def _g_sup_rt(params, transforms, profiles, data, **kwargs):
```
f0uriest, line 1309: all the compute functions should also take `**kwargs` to maintain a consistent API

kianorr: Not sure what happened here to be honest - I don't remember deleting **kwargs but I'll fix (probably something dumb I was doing in the beginning)

desc/compute/data_index.py
```diff
@@ -69,6 +116,10 @@ def register_compute_fun(
         "kwargs": list(kwargs.values()),
     }
 
+    permutable_names = ["R_", "Z_", "phi_", "lambda_", "omega_", "sqrt(g)_"]
```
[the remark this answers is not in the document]
kianorr: They're might some complications that it doesn't account for because of the partials of metrics that have both ^ and _ but I will check.

dpanici:
don't include aliases as extra entries in docs, modify write_variables.py to do so, check for functools as module

f0uriest (approved):

kianorr:
@f0uriest in the docs, the aliases column gets cut off a bit depending on screen size. Should we care about that? I talked with dario about adding a line break to the f_c label so there's more space available but the code seems a bit annoying for that.

f0uriest:
If there's any easy fix that would be good but it's not essential/urgent

kianorr:
Okay then I'll just merge it for now since it will probably take more time than it's worth.

And for @unalmis 's comment from before: that's a good idea but I think we'll just keep this PR limited to the API since it's taken me so long already and add in more features later.

PR #649, opened 2023-09-05
VMECIO.save work with all profile types

ddudt:
Allows saving VMEC wout quantities for all types of Equilibrium profiles (pressure vs kinetic, iota vs current). 

Makes a similar change to allow writing DESC input files for Equilibria with any profile type.


desc/io/input_reader.py
```diff
@@ -12,6 +12,8 @@
 from termcolor import colored
 
 from desc import set_device
+from desc.grid import LinearGrid
```
dpanici, line 15: move to local imports not at top

dpanici (changes requested):
- add test for saving to VMEC with kinetic profile DESC eq

 - move imports to avoid circular import

f0uriest (dismissed):

desc/input_reader.py
```diff
@@ -845,91 +848,77 @@ def descout_to_input(  # noqa: C901 - fxn too complex
...
 
         f.write("\n# fixed-boundary surface shape\n")
+
         # boundary parameters
-        if eq0.sym:
-            for k, (l, m, n) in enumerate(eq0.surface.R_basis.modes):
-                if abs(eq0.Rb_lmn[k]) > 1e-8:
+        if eq.sym:
+            for k, (l, m, n) in enumerate(eq.surface.R_basis.modes):
+                if abs(eq.Rb_lmn[k]) > 1e-8:
```
dpanici, line 904: Wait I did not realize, the desc output to input only saves boundary coefficients larger than 1e-8?

I think at the least this threshold should be an input parameter, and I would put it lower to like 1e-10 as the default

dpanici: #696 not for this PR but for the future making this issue

dpanici (approved):

desc/input_reader.py
f0uriest: Iirc you were the one who added this feature and I recommended making it an option at the time

f0uriest (approved):

PR #656, opened 2023-09-08
update Princeton cluster install instructions

ddudt:


docs/installation.rst
```diff
@@ -121,43 +121,16 @@ specific JAX GPU installation instructions, as that is the main installation dif
...
-Using pip install and including GPU capabilities.
-These instructions were tested and confirmed to work on the Stellar cluster at Princeton as of 5-2-2023.
-
-First, install JAX with GPU support (commands taken from `this tutorial <https://github.com/PrincetonUniversity/intro_ml_libs/tree/master/jax>`__ ):
-
-.. code-block:: sh
-
-    module load anaconda3/2022.5
-    CONDA_OVERRIDE_CUDA="11.2" conda create --name desc-env jax "jaxlib==0.4.1=cuda112*" -c conda-forge
+    CONDA_OVERRIDE_CUDA="11.2" conda create --name desc-env jax "jaxlib==0.4.10=cuda112*" -c conda-forge
```
ddudt, line 133: Can we generalize this so that it always fetches the latest version?

f0uriest: Not that I'm aware of, since not all versions of Jax work with all versions of cuda, so "most recent" may not have a unique answer

dpanici: add comment pointing to Picscie tutorial/JAX docs to check that it is latest version (make it more explicit to check if there is something newer if you want)

f0uriest (approved):

dpanici (approved):

f0uriest:
FYI I think the tests won't run for this because the workflow ignores changes to the docs folder.

PR #657, opened 2023-09-08
fix some plotting problems

ddudt:


desc/plotting.py
```diff
@@ -1593,16 +1593,15 @@ def plot_boundary(eq, phi=None, plot_axis=True, ax=None, return_data=False, **kw
     phi = parse_argname_change(phi, kwargs, "zeta", "phi")
 
     figsize = kwargs.pop("figsize", None)
-    cmap = kwargs.pop("cmap", "rainbow")
+    cmap = kwargs.pop("cmap", "hsv")
```
ddudt, line 1596: This is a cyclic colormap, which I prefer because it makes it more clear which cross-sections correspond to which toroidal angles.

desc/plotting.py
```diff
@@ -1627,11 +1626,12 @@ def plot_boundary(eq, phi=None, plot_axis=True, ax=None, return_data=False, **kw
             period=(np.inf, 2 * np.pi, 2 * np.pi),
             guess=grid.nodes,
         ),
+        NFP=eq.NFP,
         sort=False,
     )
 
     if colors is None:
-        colors = _get_cmap(cmap, nz - 1)(np.linspace(0, 1, nz - 1))
+        colors = _get_cmap(cmap, nz)(np.linspace(0, 1, nz))
```
ddudt, line 1634: This change is because of the switch to the cyclic colormap. It will still work fine for other colormaps, it's just that the "end" color won't get used.

desc/grid.py
```diff
@@ -396,20 +396,22 @@ class Grid(_Grid):
     Parameters
     ----------
     nodes : ndarray of float, size(num_nodes,3)
-        node coordinates, in (rho,theta,zeta)
+        Node coordinates, in (rho,theta,zeta)
+    NFP : int
+        Number of field periods (Default = 1).
```
f0uriest, line 401: What does having nfp!=1 do for custom grids?

desc/grid.py
```diff
@@ -396,20 +396,22 @@ class Grid(_Grid):
...
-        whether to sort the nodes for use with FFT method.
+        Whether to sort the nodes for use with FFT method.
     jitable : bool
         Whether to skip certain checks and conditionals that don't work under jit.
         Allows grid to be created on the fly with custom nodes, but weights, symmetry
         etc may be wrong if grid contains duplicate nodes.
     """
 
-    def __init__(self, nodes, sort=True, jitable=False):
+    def __init__(self, nodes, NFP=1, sort=True, jitable=False):
```
f0uriest, line 410: Can you put nfp after sort? Will hopefully minimize any needed changes in existing calls to Grid

desc/plotting.py
f0uriest: I think twilight or twilight_shifted would be preferred over hsv, it looks better in grayscale and for colorblindness 
https://matplotlib.org/stable/tutorials/colors/colormaps.html

f0uriest: I take that back, `twilight` looks pretty ugly. But I'm also really not a fan of `hsv`. Might be a good chance to try some of the cool ones in https://github.com/holoviz/colorcet

f0uriest: or if we're already going to start using plotly for 3d stuff, https://plotly.com/python/builtin-colorscales/#builtin-cyclical-color-scales

desc/grid.py
dpanici: use eq.NFP as default instead?

dpanici: @f0uriest double check plotting and map coords that everything works with NFP=1

f0uriest: looks like the only place we reference `Grid.NFP` is here: https://github.com/PlasmaControl/DESC/blob/da715883e689956c08bf8b03ffd0396d37fb4150/desc/plotting.py#L1660

Probably easiest to just change that to `eq.NFP` and leave `Grid` with NFP of 1

desc/plotting.py
ddudt: @f0uriest You can update the default colormap if you want to, otherwise this PR is done. 

Honestly I don't think `hsv` looks that different from `rainbow` that we had before. Also it's just a default; you can always supply your own colormap.

f0uriest (dismissed):

desc/plotting.py
```diff
@@ -1656,8 +1655,8 @@ def plot_boundary(eq, phi=None, plot_axis=True, ax=None, return_data=False, **kw
             color=colors[i],
             linestyle=ls[i],
             lw=lw[i],
-            label="$\\phi \\cdot NFP/2\\pi = {:.3f}$".format(
-                grid.NFP * phi[i] / (2 * np.pi)
+            label="$\\phi \\cdot N_{{FP}}/2\\pi = {:.2f}$".format(
```
dpanici, line 1658: If you are making this change here, might as well make it to all the other plot functions which use NFP in the latex string

dpanici (changes requested):
Can you change the NFP to N_{FP} for the rest of the functions which use that in their titles/labels?

f0uriest (approved):

dpanici (approved):

PR #658, opened 2023-09-09
Update CI testing breakdown

### LINKED ISSUE #654: Speed up tests running on CI
running all the tests on CI has gotten pretty slow, even after splitting into chunks (1 of the regression tests chunks has been taking 3+ hours).

Right now we run the full unit test suite with 2 different versions of python, which seems wasteful since we've never actually had any issues due to python versions (jax/numpy/matplotlib versions are much more likely to cause problems).

I think we could reduce the duplicated unit tests, and test each remaining chunk on a different python version, which would free up more workers for the regression tests so we can split those more.

f0uriest:
Resolves #654


dpanici (approved):
Looks fine, am surprised that the figures did not need to be regenerated

PR #660, opened 2023-09-10
Reduce resolution of QH test, combine into single test

f0uriest:
- Reduces resolution of precise QH test from 8 to 4 so it runs a lot faster.
- Also seems to make it more robust, tested on several different branches and now passes on all of them.
- Combines into single test rather than breaking into stages.
- No longer comparing flux surfaces at all, just testing magnitude of boozer modes and checking elongation to avoid degenerate solutions, similar to the recent fixes for the simsopt comparison test in #620


dpanici:
How do the flux surfaces compare, out of curiosity?

f0uriest:
> How do the flux surfaces compare, out of curiosity?

Fairly "standard" qh, similar to what we had before and the simsopt one

dpanici (approved):

PR #661, opened 2023-09-11
Add new options to plot_boozer_modes

### LINKED ISSUE #647: Add option to `plot_boozer_modes` to only plot symmetry breaking modes
And possibly also just the max of all the symmetry breaking modes?

f0uriest:
- Adds option to plot only symmetry breaking modes
- Or max symmetry breaking mode
- Also refactors things a bit to simplify the logic

Resolves #647


desc/plotting.py
```diff
@@ -2193,57 +2215,77 @@ def plot_boozer_modes(
...
+        plot_op = ax.semilogy
+        B_mn = np.abs(B_mn)
+    else:
+        plot_op = ax.plot
+
+    if max_only:
+        plot_op(
+            rho,
+            B_mn,
+            label=kwargs.pop("label", ""),
```
dpanici, line 2256: I think that the (M,N) should be listed for Max, otherwise all we know is it is the max mode but we don't know which mode is the max

dpanici (changes requested):
Can you add a default label to the max breaking mode with its mode numbers?

desc/plotting.py
f0uriest: We're taking the point wise max, so there isn't a single M,N. In the test plots you can see that which mode is largest changes with radius. The use case for this is more likely comparing multiple equilibria, in which case the label would be the name

dpanici (dismissed):

desc/plotting.py
```diff
@@ -2193,57 +2215,77 @@ def plot_boozer_modes(
...
+        ylabel = r"Max symmetry breaking Boozer $B_{M,N}$"
+    elif helicity:
+        ylabel = r"Symmetry breaking Boozer $B_{M,N}$"
+    else:
+        ylabel = r"$B_{M,N}$ in Boozer coordinates"
+    ylabel += r" (normalized)" if norm else r" $(T)$"
+    ax.set_ylabel(ylabel, fontsize=ylabel_fontsize)
 
     if kwargs.pop("legend", True):
         fig.legend(**kwargs.pop("legend_kw", {"loc": "center right"}))
```
ddudt, line 2291: `"lower right"` might be a better default option now

tests/test_plotting.py
```diff
@@ -778,14 +778,52 @@ def test_plot_boozer_modes():
     """Test plotting boozer spectrum."""
     eq = get("WISTELL-A")
     fig, ax, data = plot_boozer_modes(
-        eq, M_booz=eq.M, N_booz=eq.N, num_modes=7, return_data=True
+        eq, M_booz=eq.M, N_booz=eq.N, num_modes=7, return_data=True, norm=True
```
ddudt, line 781: Why this change? The other two tests you added also use `norm=True`, so it could be nice to keep this as `norm=False` for more coverage.

f0uriest: Mainly for visual consistency. If they're all normed all the plots line up so its easier to tell that everything is doing what it should be

f0uriest: It shouldn't effect the coverage at all, the way its coded `norm=True` exercises more code

dpanici (approved):

ddudt (dismissed):

PR #665, opened 2023-09-12
util function to ensure positive iota

ddudt:
Adds a function `enforce_positive_iota` to enforce the rotational transform to be positive (without otherwise changing the equilibrium). This is not necessary to enforce, but I find it helpful when comparing equilibria. The implementation is very similar to the existing function `enforce_positive_jacobian`. 

This should have the same effect on ensuring a positive current profile. So maybe we should rename it to `enforce_positive_current` or something more general? 

TODO:

- [x] Properly handle magnetic axis coefficients
- [x] Test when iota profile is assigned
- [x] Test when current profile is assigned


dpanici:
test stability quantities as well before/after

ddudt:
The code coverage checks might fail, but the only lines that aren't covered are the logic for `EquilibriaFamily` instead of an `Equilibrium`

tests/test_compat.py
```diff
@@ -0,0 +1,162 @@
+"""Tests for compatability functions."""
```
f0uriest, line 1: can you also add tests for the directions of the basis vectors? IE, is this flipping the direction of zeta?

f0uriest (dismissed):
Can you show what the surfaces look like before and after for the following cases:
- axisymmetry
- rotating elipse
- more complicated shape like one of the optimized stellarators

ddudt:
> Can you show what the surfaces look like before and after for the following cases:
> 
> * axisymmetry
> * rotating elipse
> * more complicated shape like one of the optimized stellarators

In axisymmetry the boundary does not change at all, we just flip the sign of the iota/current profile. Here is the SOLOVEV example: 
SOLOVEV

For stellarators, the boundary is still the same surface but with the opposite "helicity". 

In the HELIOTRON example (a rotating ellipse), the boundary changes from rotating clockwise to rotating counter-clockwise: 
HELIOTRON_neg
HELIOTRON_pos

In the W7-X example, the positive rotational transform profile is when the boundary is rotating clockwise: 
W7X_neg
W7X_pos

f0uriest:
So isn't this basically just flipping the direction of zeta?

ddudt:
Renamed the function to `flip_helicity` and it will always be applied when called, so could be used to make profiles negative if they are already positive. 

Also added more tests. The basis vectors $e_\rho$, $e_\theta$, and $e^\zeta$ are unchanged. $R_\zeta$ and $Z_\zeta$ do flip sign, which changes $e_\zeta$, but these components of the vector are not used to compute $\sqrt{g}$. The QH helicity flips its sign. 

I want to update the QH test to use `"HSX"` instead of the `"precise_QH"` example, so I will wait on #702 for this PR.

dpanici (approved):

f0uriest (approved):

PR #667, opened 2023-09-13
Move InputReader out of desc.io to allow GPU to initialize correctly

### LINKED ISSUE #663: using GPU from command line is broken
it looks like using a GPU when running desc as a program ( eg, `desc <path_to_input> --gpu`)has been broken since v0.9.2 . The culprit seems to be when we registered all of our classes to work with jax, that had to use metaclasses, which means we have to import jax when the classes are declared, not just when they're created. Long story short, I think the fix is to remove the InputReader class from `desc.io` and either put it in its own module, or maybe just put it in `__main__` or `__init__` ?

f0uriest:
Resolves #663 

Tested it manually on della, there's not really any way to test it automatically.

### DIFF
--- a/desc/__main__.py
+++ b/desc/__main__.py
@@ -2,7 +2,7 @@
 
 import sys
 
-from desc.io import InputReader
+from desc.input_reader import InputReader
 
 
 def main(cl_args=sys.argv[1:]):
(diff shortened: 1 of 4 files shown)


dpanici (approved):

dpanici (changes requested):
fix docs? unsnure why failing, seems pandoc version is not correct, it claims

ddudt (approved):

PR #672, opened 2023-09-14
Fix edge case for passing callable profile target

f0uriest:


tests/test_objective_funs.py
```diff
@@ -620,6 +620,15 @@ def test_target_profiles():
             ]
         ),
     )
+    objp = Pressure(target=lambda x: 2 * x, eq=eqc)
```
dpanici, line 623: does this test exercise the code because somewhere down the line, `grid.nodes` is passed into the callable? just want to make sure I understand

f0uriest: yeah, when the target or bounds are callable we call them with grid.nodes[unique_rho_idx]. `Profile` objects can correctly ignore the theta/zeta parts but if its an arbitrary callable like this that fails.

PR #674, opened 2023-09-15
Add documentation for optimizer options

### LINKED ISSUE #659: Document optimizer options

f0uriest:
- Also moves some previously positional arguments to be keyword options instead to give a more uniform API and make them accessible through the `Optimizer.optimize` interface.
- Combines `fmin-exact`, `fmin-dogleg`, `fmin-subspace` into one method `fmintr`, with the sub methods being selectable by passing `tr_method` in as part of `options`
- Removes stopping criteria on number of gradient/hessian evaluations as this is basically redundant with number of iterations

Resolves #659


desc/optimize/least_squares.py
```diff
@@ -297,10 +326,9 @@ def lsqtr(  # noqa: C901 - FIXME: simplify this
                 tr_decrease_threshold,
                 tr_decrease_ratio,
             )
-            if return_tr:
-                alltr.append(trust_radius)
+            alltr.append(trust_radius)
             alpha *= tr_old / trust_radius
-
+            # TODO: does this need to move to the outer loop?
```
ddudt, line 331: Is this something you still need to check before this PR is merged?

dpanici (approved):

PR #680, opened 2023-09-19
Add Ability to Write Boozmn.nc style output files

dpanici:
Add `make_boozmn_output` utility function to `vmec_utils`, which allows DESC to write boozmn.nc style output files containing Boozer harmonics. I/O based heavily off of the[ hidden symmetries version of booz_xform](https://github.com/hiddenSymmetries/booz_xform/tree/main)

- [x] add documentation (in form of notebook or a page) on our Boozer transform setup and how to use and the meaning of our `"|B|_mn_B"` and how to get into the familiar double angle form - make issue
- [x] deprecate `|B|_mn`
- [x] make issue for asym boozmn and kick can down the road - only issue is the `numnc` from the DESC-saved `boozmn` is the opposite sign to the C++ saved `boozmn`. Not sure what the issue is... the DESC-saved `nu` agrees better with the DESC `nu`, and the other quantities are all in agreement from both C++ and DESC, so it is something subtle


dpanici:
make a norm name in data index

dpanici:
notes to self on Booz_xform

modenumbers are defined [here](https://github.com/PrincetonUniversity/STELLOPT/blob/7c70770b629d5f57f0557455cab5871b91e855c9/BOOZ_XFORM/Sources/setup_booz.f#L33): 
 - if m=0, they only include n from 0 -> nboz
 - else if m is not 0, they include all n modes from -nboz -> nboz
 - this means that they include the (0,0) mode even for their sin terms, but they skip the (0,-nboz -_ nboz) modes. This may explain the discrepancy in modes I see

desc/compute/_qs.py
```diff
@@ -240,25 +273,78 @@ def _zeta_B(params, transforms, profiles, data, **kwargs):
 
 
 @register_compute_fun(
-    name="sqrt(g)_B",
-    label="\\sqrt{g}_{B}",
+    name="jac_Boozer_DESC",
```
f0uriest, line 276: should also document this name change in the changelog

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+        N_booz = 2 * eq.N
+
+    # calculations are done on a "half-grid" of size surfs-1
+    # which starts at s =  psi = 0.5/(surfs-1)
+    # and increments by 1 / (surfs-1)
+    # until it reaches 1-0.5/(surfs-1)
+    # VMEC radial coordinate: s = rho^2 = Psi / Psi(LCFS)
+    s_full = np.linspace(0, 1, surfs)
+    hs = 1 / (surfs - 1)
+    s_half = np.arange(hs / 2, 1, hs)
```
f0uriest, line 544: i think using `linspace + 0.5/nsurfs` is safer, `np.arange` with floating point steps can sometimes give the wrong sized output

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    r_full = np.sqrt(s_full)
+    r_half = np.sqrt(s_half)
+
+    basis = get_transforms(
```
f0uriest, line 548: would it be easier to just create the double fourier basis here, since thats all you need? Otherwise it creates a bunch of other stuff that gets thrown out

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    # just evaluate the fit at the rho to get the B_theta_mn and B_zeta_mn
+    # so this we only only do a single fit
+    transform = Transform(
+        grid=vol_grid,
+        basis=FourierZernikeBasis(
+            L=2 * eq.L, M=M_booz - 1, N=N_booz, sym="cos", NFP=eq.NFP
+        ),
+        build_pinv=True,
+    )
+    B_zeta_lmn = transform.fit(data_vol["B_zeta"])
```
f0uriest, line 607: are we sure B_zeta and B_theta are analytic quantities? B obviously is, but i thought the basis vectors aren't at the axis which would make the components possibly not analytic

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+        # cos symmetric terms
+        transforms = get_transforms(
+            ["|B|_mn", "sqrt(g)_B_mn"],
+            obj=eq,
+            grid=grid,
+            M_booz=M_booz - 1,
+            N_booz=N_booz,
+        )
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
f0uriest, line 642: what warnings get caught here?

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+            data = eq.compute(
+                ["|B|_mn", "R_mn", "Z_mn", "nu_mn", "sqrt(g)_B_mn", "psi_r"],
+                grid=grid,
+                transforms=transforms,
+                data=data,
```
f0uriest, line 647: may also want to include `override_grid=True` here

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+            )
+        if eq.sym:
+            # Z and nu are sin-symmetric, but the transforms used
+            # before are cos symmetric.
+            # We want to use the data that needed cos sym transforms,
+            # but remove the incorrectly calculated Z_mn,nu_mn from the
+            # dictionary first
+            data.pop("nu_mn")
```
f0uriest, line 655: wouldn't it be better to just not compute those qtys in the first place on the wrong grid?

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+                    data=data_sin,
+                )
+
+        else:
+            data_sin = data
+        b_mn = np.where(
+            transforms["B"].basis.modes[:, 1] < 0, -data["|B|_mn"], data["|B|_mn"]
+        )
+        b_mn = np.atleast_2d(matrix @ b_mn)
+        B_mn = np.vstack((B_mn, b_mn)) if B_mn.size else b_mn
```
f0uriest, line 684: isn't B_mn always empty? it doesn't look like it gets populated anywhere

desc/compute/_qs.py
```diff
@@ -165,6 +165,39 @@ def _nu(params, transforms, profiles, data, **kwargs):
...
+        "jac_Boozer_DESC",
+    ],
+)
+def _nu_mn(params, transforms, profiles, data, **kwargs):
+    nodes = jnp.array([data["rho"], data["theta_B"], data["zeta_B"]]).T
+    norm = 2 ** (3 - jnp.sum((transforms["B"].basis.modes == 0), axis=1))
+    data["nu_mn"] = (
+        norm  # 1 if m=n=0, 2 if m=0 or n=0, 4 if m!=0 and n!=0
+        * (
+            transforms["B"].basis.evaluate(nodes).T
```
f0uriest, line 193: can make a new compute qty for this matrix to avoid recomputing

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+
+        r_mn = np.where(
+            transforms["B"].basis.modes[:, 1] < 0, -data["R_mn"], data["R_mn"]
+        )
+        r_mn = np.atleast_2d(matrix @ r_mn)
+        R_mn = np.vstack((R_mn, r_mn)) if R_mn.size else r_mn
```
f0uriest, line 690: same as above

f0uriest: nvm didn't realize this was still part of the loop

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    ixn_b.units = "None"
+    ixn_b[:] = modes[:, 2] * eq.NFP
+
+    iotas = file.createVariable("iota_b", np.float64, ("radius",))
+    iotas.long_name = (
+        "Rotational transform. The radial grid corresponds to the"
+        "requested surfaces, and a 0 is prepended"
+    )
+    iotas.units = "None"
+    if eq.iota is not None:
```
f0uriest, line 855: couldn't we just always use eq.compute to get iota?

desc/vmec_utils.py
```diff
@@ -469,3 +487,509 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    bvco_b.units = "None"
+    bvco_b[0] = 0
+
+    bvco_b[1:] = grid.compress(data_1d["G"])
+
+    # TODO: assuming this is on full mesh
+    presf = file.createVariable("pres_b", np.float64, ("radius",))
+    presf.long_name = "pressure on full mesh"
+    presf.units = "Pa"
+    presf[:] = eq.pressure(r_full)
```
f0uriest, line 886: should use eq.compute here in case of kinetic profiles

tests/test_vmec.py
```diff
@@ -907,3 +910,180 @@ def test_vmec_boundary_subspace(DummyStellarator):
...
+    # on 3 surfaces using boozer resolutions of 15
+    make_boozmn_output(
+        eq, output_path, surfs=surfs, verbose=2, M_booz=boozer_res, N_booz=boozer_res
+    )
+    # load in the .nc file
+    file = Dataset(output_path, mode="r")
+
+    # make half grid in s
+    hs = 1 / (surfs - 1)
+    s_half = np.arange(hs / 2, 1, hs)
```
f0uriest, line 934: same as before

tests/test_vmec.py
```diff
@@ -907,3 +910,180 @@ def test_vmec_boundary_subspace(DummyStellarator):
...
+def test_make_boozmn_output(TmpDir):
+    """Test that booz_xform-style outputs accurately reconstruct quantities."""
+    # load in HELIOTRON equilibrium
+    eq = get("HELIOTRON")
+    output_path = str(TmpDir.join("boozmn_out.nc"))
+
+    boozer_res = 50
+    surfs = 4
+    # Use DESC to calculate the boozer harmonics and create a booz_xform style .nc file
+    # on 3 surfaces using boozer resolutions of 15
```
f0uriest, line 925: 15 or 50?

tests/test_vmec.py
```diff
@@ -907,3 +910,180 @@ def test_vmec_boundary_subspace(DummyStellarator):
...
+    # compare coefficients
+    for i, (quant_DESC, quant_cpp) in enumerate(zip(quantities, quantities_cpp)):
+        np.testing.assert_allclose(
+            quant_DESC, quant_cpp, atol=quant_atols[i], err_msg=quant_names[i]
+        )
+    for key in list(file_cpp.variables.keys()):
+        if key not in [
+            "rmnc_b",
+            "zmns_b",
+            "bmnc_b",
```
f0uriest, line 1075: why are we not comparing these?

tests/test_vmec.py
```diff
@@ -907,3 +910,180 @@ def test_vmec_boundary_subspace(DummyStellarator):
...
+                data[name],
+                atol=5e-4,
+                rtol=1e-4,
+                err_msg=f"{name} at surf index {surf_index}",
+            )
+
+
+@pytest.mark.unit
+def test_make_boozmn_output_against_hidden_symmetries_booz_xform(TmpDir):
+    """Test that booz_xform-style outputs compare well against C++ implementation."""
```
f0uriest, line 1012: can we also add a test for a non-symmetric case?

desc/vmec_utils.py
dpanici: they are needed for the other cosine quantities, however they are not needed for the sin quantities. and if you are not stellarator symmetric, then the grid and basis is the same for everything and this block is not entered.

I could change it tho to be more efficient

tests/test_vmec.py
dpanici: they are already compared above, with the relevant tolerances (they will not be equal to machine precision bc booz xform interpolates all the quantities R,Z, lambda, etc radially to the half grid from the full gid, while we don't have any interpolation error when we go to the radial half grid)

desc/vmec_utils.py
dpanici: I copied these from the save vmec method, which at the time at least did not do this. I will update

tests/test_vmec.py
dpanici: you mean use eq.compute("psi")?

desc/vmec_utils.py
dpanici: I am not certain, but this seems to work out fine and agree basically exactly with just calculating B_zeta_mn on a surface

desc/vmec_utils.py
dpanici: ended up changing so that I use the transforms as well

desc/vmec_utils.py
dpanici: good point

desc/compute/_qs.py
```diff
@@ -301,6 +422,48 @@ def _B_modes(params, transforms, profiles, data, **kwargs):
...
+)
+def _boozer_prefactor_modes_norm(params, transforms, profiles, data, **kwargs):
+    # norm is 1 if m=n=0, 2 if m=0 or n=0, 4 if m!=0 and n!=0
+    norm = 2 ** (3 - jnp.sum((transforms["B"].basis.modes == 0), axis=1))
+    data["Boozer transform prefactor modes norm"] = norm
+    return data
+
+
+@register_compute_fun(
+    name="Boozer transform prefactor",
```
f0uriest, line 446: minor point but I wouldn't call the whole matrix a prefactor, I would think that the matrix really is the transform (in the same way that the dft can be written as a matrix operation). Maybe `Boozer transform matrix`?

dpanici: thanks, this was just a placeholder name anyways

desc/compute/_qs.py
```diff
@@ -301,6 +422,47 @@ def _B_modes(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="Boozer transform modes norm",
+    label="\\nu_{mn} = (\\zeta_{B} - \\zeta)_{mn}",
+    units="rad",
+    units_long="radians",
```
f0uriest, line 429: double check these, i think it should be unitless, and can probably leave the label blank

desc/compute/_qs.py
```diff
@@ -301,6 +422,47 @@ def _B_modes(params, transforms, profiles, data, **kwargs):
...
+    norm = 2 ** (3 - jnp.sum((transforms["B"].basis.modes == 0), axis=1))
+    data["Boozer transform modes norm"] = norm
+    return data
+
+
+@register_compute_fun(
+    name="Boozer transform matrix",
+    label="\\nu_{mn} = (\\zeta_{B} - \\zeta)_{mn}",
+    units="rad",
+    units_long="radians",
```
f0uriest, line 449: double check these, i think it should be unitless, and can probably leave the label blank

desc/compute/_qs.py
dpanici: forgot to change these, thakns for the catch

desc/vmec_utils.py
```diff
@@ -469,3 +486,575 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
     boundary_subspace = block_diag(Rb_subspace, Zb_subspace)
     opt_subspace = null_space(boundary_subspace)
     return opt_subspace
+
+
+def make_boozmn_output(  # noqa: 16 fxn too complex
```
ddudt, line 491: Should we move this into the `VMECIO` class? In my mind it is similar to the `VMECIO.save()` method since they are both outputting NetCDF files in a format that is specific to VMEC.

ddudt: And ideally we would rename them to something like `save_wout` and `save_boozmn`

desc/vmec_utils.py
```diff
@@ -469,3 +486,575 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+        # needed transforms other than the "B" and "w" are already
+        # computed from the compute_data call above
+        # and we have those transforms in transforms_sin
+        profiles = get_profiles(data_keys_sin, obj=eq, grid=grid, jitable=True)
+        # compute data
+        d = compute_fun(
+            "desc.equilibrium.equilibrium.Equilibrium",
+            data_keys_sin,
+            params=eq.params_dict,
+            transforms=transforms_sin,
```
ddudt, line 681: Are you sure you don't have to compute fresh transforms here? Otherwise the grid could be different. 

If it is safe to use `transforms_sin` as pre-computed outside of this function, then why can't we do the same for `trans` in the above `compute_data` function? Or the `profiles` in either function?

tests/test_vmec.py
```diff
@@ -907,3 +910,449 @@ def test_vmec_boundary_subspace(DummyStellarator):
...
+
+    R_mnc = file.variables["rmnc_b"][:].filled()
+    Z_mns = file.variables["zmns_b"][:].filled()
+    B_mnc = file.variables["bmnc_b"][:].filled()
+    nu_mns = file.variables["pmns_b"][:].filled()
+    g_mnc = file.variables["gmn_b"][:].filled()
+
+    quantities = [R_mnc, Z_mns, B_mnc, g_mnc, nu_mns]
+    quant_names = ["R", "Z", "|B|", "sqrt(g)_B", "nu"]
+    quant_atols = [1e-4, 7e-5, 9e-6, 5e-3, 2e-7]
```
ddudt, line 1165: Are you sure these atols are good choices? rtols would be preferable if you can use them. It's not obvious to me what order of magnitude the output coefficients should be, and if the Fourier coefficients are decaying exponentially these atols are probably just ignoring all the higher-order modes.

ddudt (changes requested):

desc/compute/_qs.py
```diff
@@ -301,6 +422,47 @@ def _B_modes(params, transforms, profiles, data, **kwargs):
...
+    units_long="None",
+    description="Inner product norm for boozer modes basis.",
+    dim=1,
+    params=[],
+    transforms={"B": [[0, 0, 0]]},
+    profiles=[],
+    coordinates="rtz",
+    data=[],
+)
+def _boozer_modes_norm(params, transforms, profiles, data, **kwargs):
```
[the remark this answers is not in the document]
dpanici: it is not specific to boozmn, this is the normalization used when you compute the boozer transform in DESC, I just moved it from the compute functions of `B_mn` to its own thing to avoid recomputation

desc/compute/_qs.py
```diff
@@ -270,16 +340,67 @@ def _sqrtg_B(params, transforms, profiles, data, **kwargs):
...
-    )
+    norm = data["Boozer transform matrix"]
+    data["|B|_mn"] = (norm @ (data["jac_Boozer_DESC"] * data["|B|"])) / transforms[
+        "B"
+    ].grid.num_nodes
+    return data
+
+
+@register_compute_fun(
+    name="R_mn",
```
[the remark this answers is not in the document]
dpanici: sure, I will leave `nu_mn` for now though as that specifically refers to the boozer transform quantity, in the future if it is used for other coordinates we can change it

desc/compute/_qs.py
```diff
@@ -301,6 +422,47 @@ def _B_modes(params, transforms, profiles, data, **kwargs):
...
+    params=[],
+    transforms={"B": [[0, 0, 0]]},
+    profiles=[],
+    coordinates="rtz",
+    data=["rho", "theta_B", "zeta_B", "Boozer transform modes norm"],
+)
+def _boozer_transform_matrix(params, transforms, profiles, data, **kwargs):
+    nodes = jnp.array([data["rho"], data["theta_B"], data["zeta_B"]]).T
+    data["Boozer transform matrix"] = (
+        data["Boozer transform modes norm"] * transforms["B"].basis.evaluate(nodes)
```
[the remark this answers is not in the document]
dpanici: Sure but this is already used with the boozer nodes, and is only used in the boozer transform part. I include the norm multiplication here to avoid unnecessary computations as much as necessary

desc/vmec_utils.py
dpanici: the `trans` in the `compute_data` function only is called so that we can get the correct theta_B, zeta_B for that rho surface (as those change from surface to surface, so we do have to re-create some transforms in the for loop to account for the fact that we need to eval things that have a rho dependence, that goes for why we need the get profiles too)

however the B_transform and w_transform are PURELY transforms that are a function of theta and zeta, so those only need to be built once and can be reused for every surface, hence why that is how we populate `trans["B"]` and `trans["w"]` in the comput_data function

now, when we get ot compute_data_sin, we have at this point already calculated `w` and `theta_B`, `zeta_B` etc, literally the only thing left that we must compute is `nu_mn` and `Z_mn`. However, these are sin-symmetric when stellarator sym is True, but the `trans["B"]` is cos symmetric by default when sym is True. Hence we need to use the `transforms_sin` where the transforms are sin symmetric to correctly caluclate `nu_mn` and `Z_mn`

The reason we don't need to re-create the transforms_sin at each step in the loop is because any rho-dependent data has already been computed with the `compute_data` call, so we can just reuse that, so the only transforms that matter in `compute_data_sin` are the transforms which only depend on the theta and zeta, which are precomputed before the loop

it is kind of confusing but this works and it agrees well when comparing computed quantities against the computation using the boozer modes, so I think it is correct according to the tests I have for it

tests/test_vmec.py
dpanici: I am not 100% sure these are great choices to be honest, but htese are the tolerances that I get when comparing. I could try to compare with more surfs (as the main issue with the comparison I think is that the booz xform uses finite diffs to interpolate to the half grid, so they have some inherent error introduced there which DESC won't have)

desc/compute/_qs.py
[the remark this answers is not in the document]
ddudt: Ah OK I think this is needed because we aren't doing the usual `transform.fit()` and instead is more like an integration for each mode number.

desc/compute/_qs.py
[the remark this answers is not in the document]
ddudt: I still think the `"Boozer transform modes norm"` does not need to be its own compute quantity and could be moved into this function, because this is the only placed they are ever used. And the normalizations are somewhat meaningless on their own.

tests/test_vmec.py
ddudt: It feels like there is some circular logic going on here: You think your new boozmn.nc output is correct because it passes your comparison tests, but here you admit that you set the tolerances to whatever they needed to be to make them pass. 

These tolerances are low enough that the code is _probably_ correct, but it depends on the scale of the field in the example you are testing with. And by using an absolute tolerance, I still suspect you are just ignoring most of the high-order Fourier modes. Can we try switching to relative tolerances? That would avoid those two issues.

desc/vmec_utils.py
dpanici: You were right, I must have made a mistake when I tested that initially, it is inaccurate to do it the 3D way. I reverted it back to calculating it surface by surface

dpanici:
Compare not fourier to fourier but in real space the quantities from C++ and DESC

dpanici:
Maybe just throw error for asym for now

desc/vmec_utils.py
```diff
@@ -467,3 +482,547 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+
+    nfp = file.createVariable("nfp_b", np.int32)
+    nfp.long_name = "number of field periods"
+    nfp[:] = NFP
+
+    lasym = file.createVariable("lasym__logical__", np.int32)
+    lasym.long_name = "0 if the configuration is stellarator-symmetric, 1 if not"
+    lasym[:] = not eq.sym
+
+    # FIXME: ns?
```
f0uriest, line 793: still tbd?

desc/compute/_omnigenity.py
```diff
@@ -254,27 +279,73 @@ def _zeta_B(params, transforms, profiles, data, **kwargs):
...
     dim=1,
     params=[],
     transforms={},
     profiles=[],
     coordinates="rtz",
     data=["lambda_t", "lambda_z", "nu_t", "nu_z", "iota"],
 )
-def _sqrtg_B(params, transforms, profiles, data, **kwargs):
-    data["sqrt(g)_B"] = (1 + data["lambda_t"]) * (1 + data["nu_z"]) + (
+def _sqrt_g__Boozer_DESC(params, transforms, profiles, data, **kwargs):
```
dpanici, line 295: too many underscores here

desc/compute/_omnigenity.py
```diff
@@ -284,18 +355,69 @@ def _sqrtg_B(params, transforms, profiles, data, **kwargs):
...
+
+@register_compute_fun(
+    name="R_mn_B",
+    label="R_{mn}^{\\mathrm{Boozer}}",
+    units="m",
+    units_long="meters",
+    description="Boozer harmonics of radial toroidal coordinate of a flux surface",
+    dim=1,
+    params=[],
+    transforms={"B": [[0, 0, 0]]},
```
dpanici, line 382: @rahulgaur104 if you want the zeta_B deriv of R you would do

transforms = {"B": [[0,0,1]]}

then have "R_mn" in your data

and in the compute, do `dR_dzeta_B = transforms["B"].transforms(data["R_mn"], dz=1)`

dpanici:
add as error metric G' + iota I' = p' V

desc/compute/_omnigenity.py
```diff
@@ -284,18 +355,69 @@ def _sqrtg_B(params, transforms, profiles, data, **kwargs):
...
+    profiles=[],
+    coordinates="rtz",
+    data=[
+        "sqrt(g)_Boozer_DESC",
+        "Z",
+        "Boozer transform matrix",
+    ],
+)
+def _Z_mn(params, transforms, profiles, data, **kwargs):
+    norm = data["Boozer transform matrix"]
```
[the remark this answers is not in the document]
dpanici: renamed to `transform_matrix`

desc/compute/_omnigenity.py
```diff
@@ -319,6 +441,47 @@ def _B_modes(params, transforms, profiles, data, **kwargs):
...
+    params=[],
+    transforms={"B": [[0, 0, 0]]},
+    profiles=[],
+    coordinates="rtz",
+    data=["rho", "theta_B", "zeta_B", "Boozer transform modes norm"],
+)
+def _boozer_transform_matrix(params, transforms, profiles, data, **kwargs):
+    nodes = jnp.array([data["rho"], data["theta_B"], data["zeta_B"]]).T
+    data["Boozer transform matrix"] = (
+        data["Boozer transform modes norm"] * transforms["B"].basis.evaluate(nodes)
```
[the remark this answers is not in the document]
dpanici: I need to re-read the math and check, this seems to work now though

dpanici:
Everything matches well except for some reason the asymmetric (cos(mt - nz)) nu modes are the opposite sign from my boozm output compared to the Fortran/C++ output.

I currently do not understand why this is, as if I had a systematic sign error in how I save the boozmn this should crop up in the other asymmetric parts of the other quantities... but it does not, those all match well, as do the symmetric modes of nu.

sym matches fine

image
image

asym are opposite sign

image

image

dpanici:
- [ ]  check Psi sign of asym eq

dpanici:
check sign convention again

dpanici:
Remove fixme in vmec_utils for basis sym in zernike to fourier

dpanici:
See if anything changes now that #1369 is in? maybe booz_xform requires that lambda surf avg be zero?

desc/compute/_omnigenity.py
```diff
@@ -328,7 +385,83 @@ def _sqrtg_B(params, transforms, profiles, data, **kwargs):
 
 
 @register_compute_fun(
-    name="|B|_mn",
+    name="sqrt(g)_B",
```
unalmis, line 388: I suggest naming this sqrt(g)_Boozer. At least something different than the old name. Otherwise this PR redefines sqrt(g)_B and this may affect external code

desc/compute/_omnigenity.py
```diff
@@ -231,6 +231,62 @@ def _nu(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="nu_mn",
```
f0uriest, line 235: might be clearer as `nu_B_mn` or `nu_mn_Boozer`?

desc/compute/_omnigenity.py
```diff
@@ -306,20 +362,21 @@ def _zeta_B(params, transforms, profiles, data, **kwargs):
 
 
 @register_compute_fun(
-    name="sqrt(g)_B",
-    label="\\sqrt{g}_{B}",
+    name="sqrt(g)_Boozer_DESC",
```
f0uriest, line 365: Do we have a way to deprecate these names and throw a warning or something? We should consider doing that to allow a changeover period

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
     boundary_subspace = block_diag(Rb_subspace, Zb_subspace)
     opt_subspace = null_space(boundary_subspace)
     return opt_subspace
+
+
+def make_boozmn_output(  # noqa: 16 fxn too complex
```
f0uriest, line 487: think this should be `noqa: C901`

desc/vmec_utils.py
f0uriest: I think it's fine by itself. VMECIO is kind of a weird class anyways since it doesn't have any state or instance methods, its basically just an extra layer of namespace

f0uriest: should add this to the docs though

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+        File path of output data.
+    surfs: int
+        Number of flux surfaces to calculate Boozer transform at (Default = 128).
+        NOTE: because this is performed on the so-called "half-grid", this will
+        result in an output of size surfs-1, since for surfs number of surfaces
+        there is only surfs-1 surfaces on the half-grid, the first one being
+        the surface at s = 0.5 / surfs
+        where s = rho**2 is the normalized toroidal flux coordinate
+    M_booz : int, optional
+        poloidal resolution to use for Boozer transform.
```
f0uriest, line 508: mention default here

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    N_booz : int, optional
+        toroidal resolution to use for Boozer transform.
```
f0uriest, line 510: mention default here

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+        "nu",
+        "sqrt(g)_Boozer_DESC",
+    ]
+    data_vol = eq.compute(keys, grid=grid)
+
+    data_keys = ["|B|_mn_B", "R_mn_B", "sqrt(g)_B_mn", "psi_r"]
+    data_keys = data_keys + ["Z_mn_B", "nu_mn"] if not eq.sym else data_keys
+    data_keys_sin = ["Z_mn_B", "nu_mn"]
+
+    data = eq.compute(data_keys, grid=grid, data=data_vol.copy(), transforms=transforms)
```
f0uriest, line 608: does this have to be a separate step? didn't we just compute a bunch of stuff on the same grid?

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    if verbose > 1:
+        timer.disp("Boozer Transform")
+
+    file = Dataset(path, mode="w", format="NETCDF3_64BIT_OFFSET")
+    # dimensions
+    # a few of these are redundant, but are included for sake
+    # of matching the convention of the original booz_xform outputs
+    # and the hidden symmetries implementation:
+    # (below two lines are a single link)
+    # https://github.com/hiddenSymmetries/booz_xform/blob/main/src/...
```
f0uriest, line 667: better to keep these on the same line and throw a `# noqa:E501` at the end

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    lasym.long_name = "0 if the configuration is stellarator-symmetric, 1 if not"
+    lasym[:] = not eq.sym
+
+    ns = file.createVariable("ns_b", np.int32)
+    ns.long_name = "Number of radial surfaces at which data is outputted minus 1"
+    ns[:] = surfs
+
+    aspect = file.createVariable("aspect_b", np.float64)
+    aspect.long_name = "Aspect Ratio"
+    aspect[:] = eq.compute("R0/a")["R0/a"]
```
f0uriest, line 705: can this be computed above along with everything else?

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+
+    grid_lcfs = LinearGrid(M=eq.M_grid, N=eq.N_grid, rho=1, NFP=NFP)
+    Rs = eq.compute("R", grid=grid_lcfs)["R"]
```
f0uriest, line 708: can this be computed above along with everything else?

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    Rmax.long_name = "Maximum Radius"
+    Rmax[:] = np.max(Rs)
+
+    Rmin = file.createVariable("rmin_b", np.int32)
+    Rmin.long_name = "Minimum Radius"
+    Rmin[:] = np.min(Rs)
+
+    # betaxis = beta_vol at the axis?
+    betaxis = file.createVariable("betaxis_b", np.float64)
+    betaxis[:] = 0
```
f0uriest, line 719: i think we figured out what this was in vmec?

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    betaxis.long_name = "Not Implemented: This output is hard-coded to 0!"
+
+    mboz = file.createVariable("mboz_b", np.int32)
+    mboz.long_name = (
+        "Maximum poloidal mode number m for which the Fourier"
+        "amplitudes rmnc, bmnc etc are stored"
+    )
+    mboz[:] = M_booz
```
f0uriest, line 727: shouldn't this be -1 or something?

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+        modes_sin = np.insert(modes_sin, 0, [-1, 0, 0], axis=0)
+
+    ## 1D Arrays
+    keys_1d = [
+        "I",
+        "G",
+    ]
+    # this should be compressed then to just the radial profiles
+    grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, rho=r_half, NFP=NFP, sym=eq.sym)
+    data_1d = eq.compute(keys_1d, grid=grid)
```
f0uriest, line 760: weren't these computed above?

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+
+    iotas = file.createVariable("iota_b", np.float64, ("radius",))
+    iotas.long_name = (
+        "Rotational transform. The radial grid corresponds to the"
+        "requested surfaces, and a 0 is prepended"
+    )
+    iotas.units = "None"
+
+    iotas[0:] = np.insert(
+        -grid.compress(eq.compute("iota", grid=grid, data=data_1d)["iota"]), 0, 0
```
f0uriest, line 794: wasn't this computed above already?

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    bvco_b.long_name = (
+        "Coefficient multiplying grad zeta_Boozer in the covariant"
+        "representation of the magnetic field vector, often denoted G(psi)."
+    )
+    bvco_b.units = "None"
+    bvco_b[0] = 0
+
+    bvco_b[1:] = grid.compress(data_1d["G"])
+
+    # TODO: assuming this is on full mesh
```
f0uriest, line 817: should verify this and remove this todo

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    presf = file.createVariable("pres_b", np.float64, ("radius",))
+    presf.long_name = "pressure on full mesh"
+    presf.units = "Pa"
+    presf[:] = eq.compute("p", grid=LinearGrid(rho=r_full, theta=0, zeta=0))["p"]
+
+    beta = file.createVariable("beta_b", np.float64, ("radius",))
+    beta.long_name = "Blank, only included for compatibility reasons"
```
f0uriest, line 824: is this our choice or is it also like this in the simsopt and fortran versions?

desc/vmec_utils.py
```diff
@@ -468,3 +482,451 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+    beta[:] = np.zeros_like(r_full)
+
+    phipf = file.createVariable("phip_b", np.float64, ("radius",))
+    phipf.long_name = "d(phi)/ds: toroidal flux derivative, not normalized by 2pi"
+    phipf[:] = Psi * np.ones((surfs,))
+
+    phi = file.createVariable("phi_b", np.float64, ("radius",))
+    phi.long_name = "toroidal flux"
+    phi.units = "Wb"
+    phi[:] = np.linspace(0, Psi, surfs)
```
f0uriest, line 835: is this supposed to be on the full or half grid?

docs/notebooks/tutorials/omnigenity.ipynb
f0uriest, line 1: does this need to be updated? If so it needs to be re-run again

tests/test_vmec.py
```diff
@@ -1278,6 +1280,487 @@ def test_vmec_boundary_subspace(DummyStellarator):
...
+    np.testing.assert_allclose(xn_cpp, xn, atol=1e-16)
+
+    # precompute some things to avoid redundant calculations
+    quant_mns = {}
+    quant_cpp_mns = {}
+    ms = {}
+    ns = {}
+    quant_bases = {}
+    quant_mode_idxs = {}
+    for quant_index in range(len(quantities)):
```
f0uriest, line 1584: this test feels overly complicated, shouldn't we be able to directly compare arrays from the 2 `.nc` files?

f0uriest (changes requested):

desc/vmec_utils.py
dpanici: it's separate at least from the below one because if stellarator symmetry is assumed, the sin-symmetric quantites Z_mn_b and nu_mn cannot be computed with the same transforms as the cos-symmetric quantities. So I separate out these steps so that I can control that. 

I guess this and the above can be combined and then I can just compute the sin stuff separately if needed though.

desc/compute/_omnigenity.py
dpanici: I don't think we currently have a way. I guess the place to add it would be to add an alias here as sqrt(g)_B, then inside the compute function (not the `_compute` function), check if the requested name is in deprecated_names and throw a warning? should not be too hard.

dpanici: Added a deprecation warning

desc/vmec_utils.py
dpanici: This currently matches the C++ version's value when the same arguments for mboz and nboz are passed in to their function as our function.

desc/vmec_utils.py
dpanici: simsopt has it as zeros, fortran doesnot. I will change to compute this correctly

desc/vmec_utils.py
dpanici: neither techincaly since it can be different from the chosen outputted surfaces in the boozmn, I just chose it as the full grid.
https://github.com/hiddenSymmetries/booz_xform/blob/2ceb1d7b688bf9e1a88847dfbf34c990ddd086e1/src/_booz_xform/write_boozmn.cpp#L100

desc/vmec_utils.py
```diff
@@ -468,3 +483,548 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+        numc[0:, :] = -Nu_mn[:, inds_cos].squeeze()
+
+    # calculate sqrt(g)
+    gmn = file.createVariable("gmn_b", np.float64, ("comput_surfs", "mn_mode"))
+    gmn.long_name = (
+        "cos(m * theta_Boozer - n * zeta_Boozer) Fourier amplitudes of"
+        "the Boozer coordinate Jacobian (G + iota * I) / B^2"
+    )
+    gmn.units = "m/T"
+    gmn[0:, :] = Sqrt_g_B_mn[:, inds_cos].squeeze()
```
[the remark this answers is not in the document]
dpanici: no this is fine

docs/notebooks/tutorials/omnigenity.ipynb
dpanici: that was an accidental commit, removing

f0uriest (dismissed):

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ New Features
 - Adds the output quantities ``wb`` and ``wp`` to ``VMECIO.save``.
 - Change implementation of Dommaschk potentials to use recursive algorithm and symbolic integration.
 - Changes hessian computation to use chunked ``jacfwd`` and ``jacrev``, allowing ``jac_chunk_size`` to now reduce hessian memory usage as well.
+- Adds ``desc.vmec_utils.make_boozmn_output `` for writing boozmn.nc style output files
+for compatibility with other codes which expect such files from the Booz_Xform code.
+- Renames compute quantity ``sqrt(g)_B`` to ``sqrt(g)_Boozer_DESC`` to more accurately reflect what the quantiy is (the jacobian from (rho,theta_B,zeta_B) to (rho,theta,zeta)), and adds a new function to compute ``sqrt(g)_B`` which is the jacobian from (rho,theta_B,zeta_B) to (R,phi,Z).
```
YigitElma, line 21: Isn't the name of the compute function for jacobian from boozer to R,phi,Z ``sqrt(g)_Boozer``? I think this PR keeps ``sqrt(g)_B`` as alias for the DESC coordinates to Boozer coordinates?

desc/compute/_omnigenity.py
```diff
@@ -328,7 +386,83 @@ def _sqrtg_B(params, transforms, profiles, data, **kwargs):
 
 
 @register_compute_fun(
-    name="|B|_mn",
+    name="sqrt(g)_Boozer",
```
YigitElma, line 389: In the changelog, this is refered as ``sqrt(g)_B``, am I missing something?

desc/vmec_utils.py
```diff
@@ -468,3 +483,461 @@ def vmec_boundary_subspace(eq, RBC=None, ZBS=None, RBS=None, ZBC=None):  # noqa:
...
+        poloidal resolution to use for Boozer transform. Default is `2 * eq.M`.
+    N_booz : int, optional
+        toroidal resolution to use for Boozer transform.Default is `2 * eq.N`.
+    verbose: int
+        Level of output (Default = 1).
+        * 0: no output
+        * 1: status of quantities computed
+        * 2: as above plus timing information
+
+    Returns
```
dpanici, line 518: add a warning if eq is asym that `nu` is opposite sign of that saved from SIMSOPT and thus not verified, though every other quantity is checked and agrees with the other versions.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ New Features
 - Adds a new objective ``desc.objectives.CoilIntegratedCurvature`` for targeting convex coils.
 - Adds batching feature to singular integrals.
 - ``desc.objectives.CoilSetMinDistance`` and ``desc.objectives.PlasmaCoilSetMinDistance`` now include the option to use a softmin which can give smoother gradients. They also both now have a ``dist_chunk_size`` option to break up the distance calculation into smaller pieces to save memory
+- Adds ``desc.vmec_utils.make_boozmn_output `` for writing boozmn.nc style output files
+for compatibility with other codes which expect such files from the Booz_Xform code.
+- Renames compute quantity ``sqrt(g)_Boozer`` to ``sqrt(g)_Boozer_DESC`` to more accurately reflect what the quantiy is (the jacobian from (rho,theta_B,zeta_B) to (rho,theta,zeta)), and adds a new function to compute ``sqrt(g)_Boozer`` which is the jacobian from (rho,theta_B,zeta_B) to (R,phi,Z).
```
YigitElma, line 25: The first one is ``sqrt(g)_B`` which is the old name and now the alias
```suggestion
- Renames compute quantity ``sqrt(g)_B`` to ``sqrt(g)_Boozer_DESC`` to more accurately reflect what the quantiy is (the jacobian from (rho,theta_B,zeta_B) to (rho,theta,zeta)), and adds a new function to compute ``sqrt(g)_Boozer`` which is the jacobian from (rho,theta_B,zeta_B) to (R,phi,Z).
```

YigitElma (approved):

f0uriest (dismissed):

tests/test_plotting.py
```diff
@@ -787,7 +790,11 @@ def test_plot_boozer_surface(self):
         fig, ax, data = plot_boozer_surface(
             eq, M_booz=eq.M, N_booz=eq.N, return_data=True, rho=0.5, fieldlines=4
         )
-        for string in ["|B|", "theta_B", "zeta_B"]:
+        for string in [
+            "|B|",
+            "theta_B",
+            "zeta_B",
+        ]:
```
ddudt, line 797: this annoys me lol

tests/test_compute_funs.py
```diff
@@ -1142,14 +1142,15 @@ def test_boozer_transform_multiple_surfaces():
     grid1 = LinearGrid(rho=0.6, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
     grid2 = LinearGrid(rho=0.8, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
     grid3 = LinearGrid(rho=np.array([0.6, 0.8]), M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
-    data1 = eq.compute("|B|_mn", grid=grid1, M_booz=eq.M, N_booz=eq.N)
-    data2 = eq.compute("|B|_mn", grid=grid2, M_booz=eq.M, N_booz=eq.N)
-    data3 = eq.compute("|B|_mn", grid=grid3, M_booz=eq.M, N_booz=eq.N)
+    with pytest.warns(DeprecationWarning):  # |B|_mn is a deprecated name
+        data1 = eq.compute("|B|_mn", grid=grid1, M_booz=eq.M, N_booz=eq.N)
```
ddudt, line 1146: I suggest using the new name here and add this to `test_compute_deprecation_warning` if you want to cover it. Because in theory we will drop support for the old name eventually and then this test will break.

desc/compute/_omnigenity.py
```diff
@@ -390,6 +650,26 @@ def _B_modes(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="Boozer transform modes norm",
+    label="",
+    units="~",
+    units_long="None",
+    description="Inner product norm for boozer modes basis.",
```
ddudt, line 658: I know what this is, but it's not obvious from the name/description. Is this something that has to be its own compute quantity? It's really just a weighting used for other calculations, and it's a constant determined by the basis modes. I think it would make more sense to have it passed in with the `transforms` (or even add it as an attribute to the Basis class?).

ddudt (commented):
A few minor comments

desc/compute/_omnigenity.py
dpanici: It is a constant which has to be re-calculated for a few other things, so to me makes sense to ahve it computed once and stored. Whether that is here or elsewhere does not really matter, but to pass it in transforms I would need to compute it in like get_transforms or something, and that will require other special logic as well. If possible I'd like to keep it as is here, if people are curious they can check the source code like the other stuff in our list of variables.

This also could be used to transform other quantities to Boozer as well, so it is useful to have as a compute quantiity I think.

YigitElma (approved):

f0uriest (approved):

dpanici:
For future reference, the `norm` thing comes from (from my quick check here)

image

If you consider the `m=0` `n=0` case and plug in $\Omega$ into eq 20 (the cosine eqn specifically), you find that

$$
\begin{align}
\bar{\Omega}^C_{00} &= norm\frac{1}{4\pi^2} \bar{\Omega}^C_{00}\int_0^{2\pi}\int_0^{2\pi} d\theta_B d\zeta_B\\
&= \bar{\Omega}^C_{00} norm
\end{align}
$$

So for the `m=0` `n=0` mode, `norm=1` is required to recover the Boozer Fourier coefficient.

For `n=0` `m!=0` (in first line here, all other modes are killed by the inner product and mutual orthogonality of the trig functions)


$$
\begin{align}
\bar{\Omega}^C_{m0} &= norm\frac{1}{4\pi^2} \bar{\Omega}^C_{m0}\int_0^{2\pi} d\zeta_B \int_0^{2\pi} d\theta_B cos^2(m\theta_B)\\ 
&= \bar{\Omega}^C_{m0} norm \frac{1}{4\pi^2} (2\pi) (\pi)\\
&= \bar{\Omega}^C_{m0} norm / 2
\end{align}
$$

So for the `m!=0` `n=0` mode, `norm=2` is required to recover the Boozer Fourier coefficient. Same thing for the `m=0` `n!=0` mode, and for the `m!=0` `n!=0` mode, you find you need `norm=4`

f0uriest (approved):

ddudt (approved):

PR #681, opened 2023-09-21
Implement Axis and O(rho) NAE Constraints on Poloidal Stream Function Lambda

dpanici:
The NAE tells us what the transformation to Boozer coordinates looks like as well as the flux surface positions, and we can use this to see what lambda should be based off of the near-axis. This PR implements this as a linear constraint to constrain on-axis lambda and/or the O(rho) lambda behavior to be what the NAE dictates.

When we use the NAE constraint, fitting from the NAE surfaces, we have $\theta=\vartheta_B$, but we also know that $\vartheta_\mathrm{PEST}=\vartheta_B-\iota\nu$ where $\nu=\varphi_B-\phi$. But, by definition $\vartheta_\mathrm{PEST}=\theta+\lambda$. If $\theta=\vartheta_B$ imposed through the near-axis constraint, then $\lambda=-\iota\nu$. Hence, the near-axis expansion does also impose a constraint on $\lambda$ when it comes to finding equilibrium fields. 

To find $\nu$, we simply need to find $\phi$, the cylindrical angle, as a function of $\phi_B$ and $\vartheta_B$ within the near-axis expansion. We may write $\phi=\phi^{(0)}(\varphi_B)+r\phi^{(1)}(\vartheta_B,\varphi_B)+\dots$, where the leading order is well known within the near-axis framework. This is nothing but the relation between the cylindrical angle and the Boozer angle along the magnetic axis, which is intimately related to the the length along the axis.
 From eqn. A20 in http://arxiv.org/abs/1908.10253, we can see that

$$\nu^{(0)}=\int_0^\phi\left(\frac{B_0}{G_0}\frac{\mathrm{d}\ell}{\mathrm{d}\phi}-1\right)\mathrm{d}\phi'$$

Of course, beyond the axis, over flux surfaces, the cylindrical angle of a point $\mathbf{x}$ does not generally correspond to the same cylindrical angle parameter on the magnetic axis. In fact, there is in general an offset that we found geometrically in the form of $\delta\phi$. Then, using $\delta\phi\approx\frac{|\mathbf{x}_1^\pi|}{R_0}\sin\phi_1=\frac{x_1^\phi}{R_0}$,
$$\nu^{(1)}=-\delta\phi_1=-\frac{x_1^\phi}{R_0}$$
parametrised with the cylindrical angle on axis.

image


desc/objectives/getters.py
```diff
@@ -212,9 +213,13 @@ def get_NAE_constraints(
         Whether to also fix kinetic profiles.
     normalize : bool
         Whether to apply constraints in normalized units.
-    N : int,
+    N : int
         max toroidal resolution to constrain.
         If None, defaults to equilibrium's toroidal resolution
+    fix_lambda : bool
```
f0uriest, line 219: if an int is allowed should add it here

desc/objectives/linear_objectives.py
```diff
@@ -1376,6 +1363,151 @@ def compute(self, Z_lmn, **kwargs):
...
+        len(target) = len(weight) = len(modes).
+        If True uses all of the Equilibrium's modes.
+        Must be either True or specified as an array
+    name : str
+        Name of the objective function.
+
+    """
+
+    _scalar = False
+    _linear = True
```
f0uriest, line 1399: since you're inheriting from `_FixedObjective` i don't think these are needed but you do need to specify `target_arg`

desc/objectives/linear_objectives.py
```diff
@@ -1689,6 +1819,173 @@ def compute(self, Z_lmn, **kwargs):
         return f
 
 
+class FixSumModesLambda(_FixedObjective):
```
f0uriest, line 1822: I think this shouldn't inherit from `_FixedObjective` because it doesn't give a diagonal constraint matrix

tests/test_examples.py
```diff
@@ -515,7 +515,7 @@ def test_NAE_QSC_solve():
 
     # this has all the constraints we need,
     #  iota=False specifies we want to fix current instead of iota
-    cs = get_NAE_constraints(eq, qsc, iota=False, order=1, N=eq.N)
+    cs = get_NAE_constraints(eq, qsc, iota=False, order=1, fix_lambda=True, N=eq.N)
```
f0uriest, line 518: as an additional test, can we try solving with and without fixed lambda and make sure they give the same lambda near axis?

desc/objectives/nae_utils.py
```diff
@@ -6,7 +6,79 @@
...
+
+    # from integrating eqn A20 in
+    # Constructing stellarators with quasisymmetry to high order 2019
+    #  Landreman and Sengupta
+    nu_0 = np.cumsum(qsc.B0 / qsc.G0 * qsc.d_l_d_phi - 1) * np.ones_like(phi) * dphi
+    nu_0_n = Ltrans_sin.fit(nu_0)
+
+    Lconstraints = ()
+    # lambda = -iota_0 * nu
+    L_0_n = -qsc.iota * nu_0_n
```
f0uriest, line 63: are you sure this is correct? Eduardo and I spent a bunch of time going back and forth about the sign and don't remember where we landed. I think the "wrong" sign actually gave better results?

dpanici: This is the correct sign, we initialize it with the opposite sign when we do from_near_axis (I need to try changing that), but the final, correct lambda has this sign on it.

desc/objectives/linear_objectives.py
dpanici: hm, so then should the FixSumModesR and FixSumModesZ also not be inheriting?

desc/objectives/linear_objectives.py
```diff
@@ -1376,6 +1363,149 @@ def compute(self, Z_lmn, **kwargs):
...
+        -------
+        f : ndarray
+            Fixed mode lambda errors.
+
+        """
+        fixed_params = L_lmn[self._idx]
+        return fixed_params
+
+    @property
+    def target_arg(self):
```
f0uriest, line 1504: this isn't needed since its set at the class level

desc/objectives/linear_objectives.py
f0uriest: wait nvm this is fine, the `_FixedObjective` base class really just defines the `update_target` method

desc/objectives/linear_objectives.py
```diff
@@ -1689,6 +1817,172 @@ def compute(self, Z_lmn, **kwargs):
...
+        -------
+        f : ndarray
+            Fixed sum mode lambda errors.
+
+        """
+        f = jnp.dot(self._A, L_lmn)
+        return f
+
+    @property
+    def target_arg(self):
```
f0uriest, line 1981: Same as above

f0uriest (approved):

f0uriest (approved):

PR #686, opened 2023-09-25
Update 3d plotting to use plotly for better images

### LINKED ISSUE #455: Mayavi Plotting Support
When plotting 3D things, matplotlib's 3D support is not great and [cannot render multiple 3D objects well ](https://stackoverflow.com/questions/13932150/matplotlib-wrong-overlapping-when-plotting-two-3d-surfaces-on-the-same-axes)(overlap is not in order of distance but rather in order of when matplotlib draws them, etc )

example: cannot tell if this coil is surrounding the plasma or not (it should be)

image

Other softwares (coilPy, pyQsc) use mayavi, which seems to handle this issue better. Might be worth looking into how hard it would be to add support for Mayavi plotting backend

f0uriest:
- Matplotlib kinda sucks for 3d stuff
- Mayavi is a pain to install on any recentish version of python
- VTK and Paraview are really their own programs, and probably overkill for what we need
- Plotly seems to strike a nice balance. Renders nicely in a notebook or browser, you can zoom, pan, rotate, etc.
- Seems easy enough to install with pip

Downsides:
- Can't easily test plot output like with matplotlib. For now we just test that the plotting function runs without error, and I've manually inspected the results for accuracy (see below)
- For some reason it has trouble rendering LaTeX in 3d plots. As a workaround, we use a utility to convert LaTeX to unicode which seems to cover most cases, though sometimes sub- and super-scripts don't look as pretty as they could

Main API differences:
- plotly doesn't have "axes" like matplotlib does, just figures. So the `ax` argument has been replaced by `fig` for `plot_3d` and `plot_coils`, and they no longer return `ax`
- Names of colormaps, line patterns, etc are different, so use caution when specifying those using kwargs. Thankfully the error messages plotly generates are usually pretty informative and list the available options.

Resolves #455


f0uriest:
Here are what the benchmark plots look like: 
Screenshot-20230925151854-749x638
Screenshot-20230925151921-755x637
Screenshot-20230925152303-726x536
Screenshot-20230925152441-751x666

desc/plotting.py
```diff
@@ -719,38 +776,29 @@ def plot_3d(
...
-        Axis to plot on.
+    fig : plotly.graph_objs._figure.Figure, optional
+        Figure to plot on
     return_data : bool
         if True, return the data plotted as well as fig,ax
     **kwargs : dict, optional
         Specify properties of the figure, axis, and plot appearance e.g.::
 
-            plot_X(figsize=(4,6),cmap="plasma")
+            plot_X(figsize=(4,6), cmap="RdBu")
```
ddudt, line 786: Can we please keep the default as `"Plasma"`?

ddudt (dismissed):
I tested this out a little with `plot_3d` and it's so nice to have an interactive figure! See my comment about the default colormap, otherwise it looks good. We just have to be careful about remembering to manually check things since we are removing the test images, but that should be easy since we use the plotting functions regularly.

dpanici:
This needs a bit of extra work to show the plots inside of jupyter notebooks I think, at least I had a bit of trouble, does this work fine in our notebook tests?

f0uriest:
> This needs a bit of extra work to show the plots inside of jupyter notebooks I think, at least I had a bit of trouble, does this work fine in our notebook tests?

what sort of extra work? It seems to work fine in the notebook tests.

desc/plotting.py
```diff
@@ -768,116 +816,131 @@ def plot_3d(
...
+    if grid.num_rho == 1:
+        n1, n2 = grid.num_theta, grid.num_zeta
+        p1, p2 = True, True
+    elif grid.num_theta == 1:
+        n1, n2 = grid.num_rho, grid.num_zeta
+        p1, p2 = False, True
+    elif grid.num_zeta == 1:
+        n1, n2 = grid.num_theta, grid.num_rho
+        p1, p2 = True, False
+    ijk = _trimesh_idx(n1, n2, p1, p2)
```
dpanici, line 864: so a difference internally btwn mpl and plotly is we need to do some of the work to tell it how to mesh the surfaces?

dpanici (dismissed):

dpanici (approved):

PR #689, opened 2023-09-29
update changelog for v0.10.1

dpanici:


dpanici:
only changed the changelog, so I will go ahead and merge this

PR #690, opened 2023-10-03
Improve Biot-Savart discretization, add new methods for converting between coil types

### LINKED ISSUE #602: Improve Coil discretization in Biot-Savart Law
For coils which we know the curvature of (so any coil but the `XYZCoil` which does not necessarily have an accurate curvature), discretizing the coil curves by shifting the line segments along the normal to the curve can double the order of accuracy, could be useful to investigate and implement:

https://www.osti.gov/pages/servlets/purl/1817216

### LINKED ISSUE #679: Allow comments in MAKEGRID coil files
currently we cannot read in MAKEGRID formatted coil files which have comments before the periods line, should skip comment lines, probably by seeking where the `periods`  line is

### LINKED ISSUE #682: Make Coil Representation work better with interpolation methods
Currently, XYZCoils with a `cubic2` spline interpolation will return NaN, maybe we do not need to close the curve when passing points to the interpolation algorithms

### LINKED ISSUE #687: Allow conversions between coil types
The curve classes have methods to convert between types, we should do the same for coil types (overriding the base class where appropriate)

Could also add the option to CoilSet to convert all the coils to the same type/resolution. This could in theory allow much faster compilation of coilset operations. Right now everything is done in python loops which is really slow to compile.

f0uriest:
This updates the Biot-Savart implementation for filamentary coils. The existing method discretized each coil into piecewise linear segments and used the Hanson-Hirshman method to compute B from each segment. The new method uses a direct quadrature of the Biot-Savart integral using the trapezoidal rule, which converges exponentially for smooth integrands.

In testing, I've found that the quadrature method works extremely well for Fourier curves, but not as well for spline curves, likely due to the splines being only C1 or C2:
image

Another important note is that the quadrature method doesn't guarantee curl(B) = 0 except in the limit, but in practice the error is extremely small as long as there are enough points on the curve: 
image

Due to the slow convergence of the quadrature method for spline coils, I've left those using Hanson-Hirshman, and only used the quadrature method for Fourier coils. In general it seems that even for extremely complicated curves it should never need more than ~200 points to get both B and curl(B) errors down to machine precision.

- Also fixes the `SplineXYZ` classes to not include the endpoint at 2pi, making it consistent with other discretizations we have. The calculation of length has been modified to account for this.
- The base `CoilSet` is now specifically for coils of the same type and resolution, and calculations have been vectorized for speed. For combining coils of different types there is a new class `MixedCoilSet`.
- Changes `curve.to_FourierXYZCurve` etc to `curve.to_FourierXYZ`, and then overrides these methods for coils so that we can convert coil -> coil instead of having it do coil -> curve. Also adds methods for converting all coils in a coilset to a given type.

Resolves #602 
Resolves #682 
Resolves #687 
Resolves #679


desc/backend.py
```diff
@@ -118,6 +118,44 @@ def sign(x):
         y = jnp.where(x == 0, 1, jnp.sign(x))
         return y
 
+    @jit
+    def tree_stack(trees):
```
f0uriest, line 122: Reminder to include the link to original source for this

desc/coils.py
```diff
@@ -351,14 +557,160 @@ def __init__(
...
+        data = self.compute(["x"], grid=grid, params=params, basis="xyz")
+        # need to make sure the curve is closed. If it's already closed, this doesn't
+        # do anything (effectively just adds a segment of zero length which has no
+        # effect on the overall result)
+        coil_pts_start = data["x"]
+        coil_pts_end = jnp.concatenate([data["x"][1:], data["x"][:1]])
+        # could get up to 4th order accuracy by shifting points outward as in
+        # (McGreivy, Zhu, Gunderson, Hudson 2021), however that requires knowing the
+        # coils curvature which is a 2nd derivative of the position, and doing that
+        # with only possibly c1 cubic splines is inaccurate, so we don't do it
```
dpanici, line 609: we could have a conditional where if the spline interpolation method is `c2` we use the shift, but if not we use the current method.  I agree, this is a future improvement though

desc/geometry/curve.py
```diff
@@ -220,7 +220,7 @@ def from_input_file(cls, path):
             inputs = InputReader.parse_vmec_inputs(f)[-1]
         else:
             inputs = InputReader().parse_inputs(f)[-1]
-        curve = cls(
+        curve = FourierRZCurve(
```
dpanici, line 223: is there a reason to avoid using cls?

desc/geometry/curve.py
```diff
@@ -637,29 +657,34 @@ def __init__(
     ):
         super().__init__(name)
         X, Y, Z = np.atleast_1d(X), np.atleast_1d(Y), np.atleast_1d(Z)
+        X, Y, Z = np.broadcast_arrays(X, Y, Z)
```
dpanici, line 660: what does this call accomplish?

dpanici (approved):

desc/geometry/curve.py
f0uriest: The coil class inherits this method but I wasn't sure if we wanted to treat the axis as a coil

desc/geometry/curve.py
f0uriest: The main use case is if one of the coordinates is constant, ie X=[1,2,3...], Y=[4,5,6...], Z=0. We broadcast everything together to make sure they're the same length

desc/geometry/curve.py
dpanici: Ah ok I see

desc/coils.py
```diff
@@ -6,16 +6,107 @@
...
+    L = jnp.linalg.norm(d_vec, axis=-1)
+
+    Ri_vec = eval_pts[jnp.newaxis, :] - coil_pts_start[:, jnp.newaxis, :]
+    Ri = jnp.linalg.norm(Ri_vec, axis=-1)
+    Rf = jnp.linalg.norm(
+        eval_pts[jnp.newaxis, :] - coil_pts_end[:, jnp.newaxis, :], axis=-1
+    )
+    Ri_p_Rf = Ri + Rf
+
+    # 1.0e-7 == mu_0/(4 pi)
```
ddudt, line 58: Is this just more numerically stable? If you are going to hard-code numbers, you might as well combine it with the factor of 2 and use `2.0e-7`

desc/coils.py
```diff
@@ -6,16 +6,107 @@
...
+    scales the same as the error in B itself, so will only be zero when fully
+    converged. However in practice, for smooth curves described by Fourier series,
+    this method converges exponentially in the number of coil points.
+    """
+    dl = tangents
+    R_vec = eval_pts[jnp.newaxis, :] - coil_pts[:, jnp.newaxis, :]
+    R_mag = jnp.linalg.norm(R_vec, axis=-1)
+
+    vec = jnp.cross(dl[:, jnp.newaxis, :], R_vec, axis=-1)
+    denom = R_mag * R_mag * R_mag
```
ddudt, line 106: Is this any better than `R_mag**3`? The latter is more readable in my opinion.

ddudt (approved):

desc/coils.py
f0uriest: in regular numpy powers that aren't `2` get passed to a floating point `pow` routine which is relatively slow (because its designed for stuff like `x**2.12039850984509324`) while stuff like `x**2` gets evaluated as `x*x` which is super fast.

It looks like jax is smart enough to evaluate `x**3` as `x*x*x` if its jitted so I'll replace with that.

desc/coils.py
f0uriest: yeah its just to save some flops

ddudt (approved):

PR #691, opened 2023-10-04
Remove `node_pattern` from Equilibrium

### LINKED ISSUE #662: Remove `node_pattern` as attribute of `Equilibrium`
It's really not used, and each Objective should have its own default grid with whatever node pattern makes sense, if somebody wants a specific grid they can create one for a given objective.

f0uriest:
Resolves #662


desc/input_reader.py
```diff
@@ -374,7 +373,10 @@ def parse_inputs(self, fname=None):  # noqa: C901 - FIXME: simplify this
                 flag = True
             match = re.search(r"node_pattern", argument, re.IGNORECASE)
             if match:
-                inputs["node_pattern"] = words[0].lower()
+                warnings.warn(
+                    "node_pattern parameter has been deprecated. To use a custom node "
+                    + "pattern, pass in the desired grid when creating the objective."
```
dpanici, line 378: say to use a script instead of an input file

dpanici:
https://github.com/PlasmaControl/DESC/pull/691/files?short_path=79f4f5d#diff-bb03519bd8f2c335e804a49791bb33c7866bae2d6e47888d857ae9e1ad6d2ada

this is a bit of a difference, did the boundary change somehow? I don't why else g_tz would be so different

desc/objectives/_equilibrium.py
```diff
@@ -103,29 +99,14 @@ def build(self, eq=None, use_jit=True, verbose=1):
         """
         eq = eq or self._eq
         if self._grid is None:
```
rahulgaur104, line 101: So, if we run DESC from an input file, the equilibrium grid will be concentric?

f0uriest:
> https://github.com/PlasmaControl/DESC/pull/691/files?short_path=79f4f5d#diff-bb03519bd8f2c335e804a49791bb33c7866bae2d6e47888d857ae9e1ad6d2ada image
> 
> this is a bit of a difference, did the boundary change somehow? I don't why else g_tz would be so different

The name is a bit misleading, its actually plotting `sqrt(g)` as a function of theta and zeta. `sqrt(g)` depends on `e_rho` as well, and due to different node pattern we end up with a slightly different theta inside the surface, so we get a slightly different d/drho at constant theta

desc/objectives/_equilibrium.py
f0uriest: yes, `ConcentricGrid` with `jacobi` node pattern. This was already the default for all of our examples in `desc/examples` so shouldn't be any differences.

f0uriest: The `Energy` objective defaults to a quadrature grid though.

rahulgaur104: What is the difference between quadrature grid and jacobi grid? I thought jacobi node pattern was part of another quadrature grid.

f0uriest: The main difference is the poloidal resolution. a `QuadratureGrid` uses 2*M+1 nodes on each surface to get exact poloidal integrals. The radial spacing is based on jacobi polynomials. a `ConcentricGrid` by default still uses jacobi spacing radially, but uses 2*M+1 nodes on the outermost surface, and fewer as you go inwards, usually down to 1 or 3 nodes near the axis. This is because we know things scale like rho**m so the theta variation near the axis is minimal, so we can get away with using roughly 1/2 as many nodes there. There's a more detailed explainer here: https://desc-docs.readthedocs.io/en/stable/notebooks/basis_grid.html#Collocation-grids

rahulgaur104 (approved):

dpanici (approved):

PR #695, opened 2023-10-05
rescale equilibria

### LINKED ISSUE #526: Utility for rescaling equilibria to given major radius and field strength
It's common in publications to compare equilibria after re-scaling them to the same major radius and field strength. It would be nice to have a utility to do this automatically.

I think it's fairly straightforward. If we define 2 scale factors:

$c_R = R_{0, desired} / R_{0}$

$c_B = B_{desired} / B$

then we should just have to multiply `R_lmn` and `Z_lmn` by $c_R$, and multiply `Psi` by $c_B c_R^2$.

$J \times B \sim \Psi^2$ so to preserve force balance and beta, we'd also need to multiply $\nabla p$ by $c_B^2c_R^4$, which works out to multiplying `pressure` by $c_B^2c_R^5$ (the extra factor of $c_R$ is due to the $\nabla$ operator, which goes like $1/R$). For temperature and density its trickier since they can vary independently but a simple one would just be to multiply both by the square root.

For fixed current we'd also need to multiply `current` by $c_R c_B$ (I think, based on $I = \int B \cdot dl$)

ddudt:
Resolves #526 

Adds a new function `rescale(eq, R0, B0)` to `desc/compat.py` to rescale equilibria to a new major radius and/or magnetic field strength.

### DIFF
--- a/desc/compat.py
+++ b/desc/compat.py
@@ -2,7 +2,7 @@
 
 import numpy as np
 
-from desc.grid import Grid
+from desc.grid import Grid, LinearGrid, QuadratureGrid
 
 
 def ensure_positive_jacobian(eq):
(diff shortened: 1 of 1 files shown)


f0uriest:
Might be more useful if there were a few options for defining the length and field scale factors, ie, scaling to a particular `R0` or `a` or `V` or a particular `<|B|>_vol`, `B_axis`, `B_max` etc

f0uriest (approved):

dpanici (changes requested):
looks good but need to change the match case stuff to if/else so we can work with python 3.9 and 3.8 still

edit: my bad was looking at an old commit of this, this is fine now

dpanici (approved):

PR #711, opened 2023-10-18
Continuation _solve_axisym and Surface from_input_file hotfix

### LINKED ISSUE #710: Error when loading in a surface from a VMEC input file
when running

```python
surface = FourierRZToroidalSurface.from_input_file("input_LAYSM_not_present.txt")
```

I get an error
```python
/tmp/ipykernel_28279/386266570.py in <module>
----> 1 surface = FourierRZToroidalS
[input_LAYSM_not_present.txt](https://github.com/PlasmaControl/DESC/files/13031483/input_LAYSM_not_present.txt)
urface.from_input_file("input.qi-nfp5-w7x-standard-pruned-mpol3-ntor3-beta4")

~/DESC/desc/geometry/surface.py in from_input_file(cls, path)
    276             inputs = InputReader.parse_vmec_inputs(f)[-1]
    277         else:
--> 278             inputs = InputReader().parse_inputs(f)[-1]
    279         if (inputs["bdry_ratio"] is not None) and (inputs["bdry_ratio"] != 1):
    280             warnings.warn(

~/DESC/desc/input_reader.py in parse_inputs(self, fname)
    215             if isVMEC:
    216                 print("Converting VMEC input to DESC input")
--> 217                 path = self.input_path + "_desc"
    218                 InputReader.vmec_to_desc_input(self.input_path, path)
    219                 print("Generated DESC input file {}:".format(path))

TypeError: unsupported operand type(s) for +: '_io.TextIOWrapper' and 'str'
```
occurs bc the surface.from_input_file method only checks the first line for `&INDATA`, so if comment lines exist above it tries to parse it as a DESC file

### LINKED ISSUE #712: Solving DESC from input file when poloidal resolution is low with 0 bdry ratio results in incorrect boundary shape
Running DESC from the command line on an input file where the boundary is low resolution and so is the equilibrium causes the DESC solution resulting boundary surface to be incorrect. here is an input file with boundary M=N=3 and when solved from the command line with L=M=N=3 spectral resolution, result in an equilibrium whose surface lacks the M=2 N=0 R mode and M=-2 N=0 Z modes, even though they are present in the input file and present in the surface if one loads the surface from the file with `FourierRZToroidalSurface.from_input_file`

Issue probably because the first step of the continuation is very low resolution (like L=M=1) and so it cuts out all of the axisymmetric modes with M>1, then when we change resolution, because we are not changing the non-axisymmetry (i.e. bdry ratio is fixed), we don't add back those modes to the surface, resulting in a different boundary at the end which lacks the axisymmetric boundary modes with M>1

Solution is to just set the surface at the beginning of each continuation step in the axisymmetric solve back to the full surface, then the eq.change_resolution call will truncate it down at each step to that step's resolution.

```python
from desc.io import load
surface = FourierRZToroidalSurface.from_input_file("input_DESC_M3_bdry_M3_res_bug.txt")
print(surface.R_lmn[surface.R_basis.get_idx(M=2)]) # is nonzero, this is correct
eq = load("input_DESC_M3_bdry_M3_res_bug.txt_output.h5")[-1] # load eq after running input from cmd line
print(eq.surface.R_lmn[eq.surface.R_basis.get_idx(M=2)]) # this is zero! when it should be the same surface as above
```
[input_DESC_M3_bdry_M3_res_bug.txt](https://github.com/PlasmaControl/DESC/files/13031794/input_DESC_M3_bdry_M3_res_bug.txt)

dpanici:
- let surface from_input_file method find a VMEC file if INDATA is not first line, change input.DSHAPE to have a comment line
- fix bug in `_solve_axisym` method where if the resolution is low enough, the method does not run to the full resolution of the desired axisymmetric vacuum equilibrium

resolves #710 
resolves #712


tests/test_examples.py
```diff
@@ -52,6 +52,15 @@ def test_SOLOVEV_vacuum(SOLOVEV_vac):
     np.testing.assert_allclose(data["iota"], 0, atol=1e-16)
     np.testing.assert_allclose(data["|J|"], 0, atol=3e-3)
 
+    # test that solving with the continuation method works correctly
+    # when eq resolution is lower than the mres_step
+    eq.change_resolution(L=3, M=3)
```
f0uriest, line 57: I feel like in this sort of case it would be faster/better to just solve at the full resolution from the start, instead of doing multigrid in M

Like if we did
```python
Mi = min(M, mres_step) if mres_step > 0 else M
mres_steps = int(max(np.ceil(M / mres_step), 1)) if mres_step > 0 else 0

dpanici:
@f0uriest do we want to make a similar change to the `_add_pressure` function, where if `pres_step` is zero it will just solve with the entire pressure? right now it will just run through the function and not run anything if `pres_step` is zero, effectively skipping that step, which is fine but it is now different than the behavior of `_solve_axisym` if `mres_step` is 0 (where that function will now run the equilibrium at the full poloidal resolutoin in one step if `mres_step` is 0)

we could revert that behavior to skipping the axisym solve if `mres_step` is 0 by reverting the `<=` to `<` in that function's `while` loop, I think

f0uriest:
> @f0uriest do we want to make a similar change to the `_add_pressure` function, where if `pres_step` is zero it will just solve with the entire pressure? right now it will just run through the function and not run anything if `pres_step` is zero, effectively skipping that step, which is fine but it is now different than the behavior of `_solve_axisym` if `mres_step` is 0 (where that function will now run the equilibrium at the full poloidal resolutoin in one step if `mres_step` is 0)
> 
> we could revert that behavior to skipping the axisym solve if `mres_step` is 0 by reverting the `<=` to `<` in that function's `while` loop, I think

I don't think that's necessary. In general neither step should ever be zero.

f0uriest (approved):

PR #721, opened 2023-10-22
fix some typos and try to increase readability of equations

YigitElma:
There were 2 typos:
- R, Z and $\lambda$ were called dependent variables. But they are independent and they are called independent in the previous page.
- radial part of the Zernike polynomial is shown by Z. Although it makes sense if you just put $\rho$ inside the brackets, still we should be more consistent with the names.

For some of us using the website to look up the formulas, I tried to make the equations more readable. You might not want to insert higher fractions inside a text line but most of the quantities are just defined inside the text. So, it might be more visible this way.

For some basis functions, the absolute value for m or n is unnecessary (maybe you intentionally put it for aligning purposes then no problem). Just wanted to mention.


f0uriest:
Personally I'd vote for keeping the absolute value signs, I think it makes it clearer.

YigitElma:
@f0uriest Ok I changed it back

f0uriest (approved):

f0uriest (approved):

PR #725, opened 2023-10-25
Add free boundary equilibrium objectives

### LINKED ISSUE #778: Tutorial: Free boundary

f0uriest:
- Adds objectives for performing free boundary equilibrium calculations, `BoundaryError` for finite beta and `VacuumBoundaryError` for vacuum fields.
- Adds utility functions for performing singular integral transforms over a surface using the high order method by Malhotra et al.
- New tutorial demonstrating the basics of free boundary equilibrium.
- Also adds python/JAX implementation of NESTOR algorithm used by VMEC, though this is mainly for comparison as the Malhotra method is both faster and more accurate (though likely speed could be improved by refactoring the code).

Resolves #778


dpanici:
Trying Patrick's spectral condensation

krystophny:
Hey, @f0uriest great work! Would DESC free boundary work on devices without stellarator symmetry i.e. tokamaks with perturbations already? Do you think we could have a chat on sheet currents some time?

f0uriest:
> Hey, @f0uriest great work! Would DESC free boundary work on devices without stellarator symmetry i.e. tokamaks with perturbations already? Do you think we could have a chat on sheet currents some time?

Yes it should work fine for non-stellarator symmetric cases, we've been working on testing it with perturbed equilibria from d3d. I'd be happy to chat later this week or next. Will you be in canberra for the simons meeting? We could also chat then.

desc/nestor.py
```diff
@@ -0,0 +1,1287 @@
...
+from scipy.constants import mu_0
+
+from desc.backend import fori_loop, jnp, put
+from desc.grid import LinearGrid
+from desc.io import IOAble
+from desc.transform import Transform
+from desc.utils import Index
+
+
+def copy_vector_periods(vec, zetas):
```
dpanici, line 18: This is Jonathan's code so maybe we won't change it, but don't we have util functions for this already?

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,840 @@
...
+        grid=None,
+        field_grid=None,
+        name="Vacuum boundary error",
+    ):
+        if target is None and bounds is None:
+            target = 0
+        self._grid = grid
+        self._ext_field = ext_field
+        self._field_grid = field_grid
+        things = [eq]
```
dpanici, line 81: If we were to have  single stage, where the external field is also allowed to vary, would we have a separate quadratic flux objective with the external field having DOF for it, or would we allow it to be inside of `things` for the free boundary objectives? I guess the former probably

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,840 @@
...
+        if self._src_grid is None:
+            src_grid = LinearGrid(
+                rho=np.array([1.0]),
+                M=M_grid,
+                N=N_grid,
+                NFP=int(eq.NFP),
+                sym=False,
+            )
+        else:
+            src_grid = self._src_grid
```
dpanici, line 395: any check for making sure src and eval grid is only on the rho=1 surface?

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,840 @@
...
+                "desc.magnetic_fields.FourierCurrentPotentialField",
+                self._sheet_data_keys,
+                params=sheet_params,
+                transforms=constants["sheet_eval_transforms"],
+                profiles={},
+            )
+            src_data["K_vc"] += sheet_src_data["K"]
+
+        # this is in cartesian
+        Bplasma = -virtual_casing_biot_savart(
```
dpanici, line 560: reason for the negative sign?

I'd like the sign conventions to be clearly laid out at least in the objective so it is clear what is what i.e. that K = n x B or B x n, n is an outward facing normal, etc

desc/singularities.py
```diff
@@ -0,0 +1,730 @@
...
+    # vmap found more efficient than fori_loop, esp on gpu, but uses more memory
+    if loop:
+        f = fori_loop(0, v.size, polar_pt_loop, f)
+    else:
+        f = vmap(polar_pt_vmap)(jnp.arange(v.size)).sum(axis=0)
+
+    return f
+
+
+def singular_integral(
```
dpanici, line 521: seems like the eval data can be in `rpz` but the output quantity will be in `xyz`? should document this if this is the case and possibly add basis argument as well

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,840 @@
...
+
+        bsq_out = jnp.sum(Bex_total * Bex_total, axis=-1)
+        bsq_in = jnp.sum(Bin_total * Bin_total, axis=-1)
+
+        g = eval_data["|e_theta x e_zeta|"]
+        Bn_err = Bn * g
+        Bsq_err = (bsq_in + eval_data["p"] * (2 * mu_0) - bsq_out) * g
+        Bjump = Bex_total - Bin_total
+        if self._sheet_current:
+            Kerr = mu_0 * sheet_eval_data["K"] - jnp.cross(eval_data["n_rho"], Bjump)
```
dpanici, line 586: this Kerr just ensures that the sheet current is the one due to n x B_jump right?

desc/perturbations.py
```diff
@@ -427,6 +427,7 @@ def optimal_perturb(  # noqa: C901 - FIXME: break this up into simpler pieces
     dPsi=False,
     dRb=False,
     dZb=False,
+    dIGPhi=False,
```
dpanici, line 430: not in docstring, and is not included in the `argmap`, is this necessary/used in this PR?

desc/singularities.py
```diff
@@ -0,0 +1,730 @@
...
+    r, w = jnp.meshgrid(r, w)
+    r = r.flatten()
+    w = w.flatten()
+    dr, dw = jnp.meshgrid(dr, dw)
+    dr = dr.flatten()
+    dw = dw.flatten()
+    return r, w, dr, dw
+
+
+class _BIESTInterpolator(IOAble, ABC):
```
dpanici, line 91: Do you have your own writeup of this stuff (like for your free bdry paper)? Would be nice to have (for me) when reviewing this file

tests/test_nestor.py
```diff
@@ -0,0 +1,234 @@
+"""Tests for NESTOR algorithm, benchmarked against fortran version.
+
+Created on Mon Apr 12 21:14:09 2021
+
+@author: jonathan
+"""
+
```
dpanici, line 7: This is not a real `pytest` test? or does pytest handle having a file like this?

tests/test_nestor.py
```diff
@@ -0,0 +1,234 @@
...
+        Btot,
+        nestor.M,
+        nestor.N,
+        nestor.ntheta,
+        nestor.nzeta,
+        nestor.NFP,
+    )
+
+
+def test_same_outputs(tmpdir_factory):
```
dpanici, line 192: needs a pytest mark for this to do anything

tests/test_objective_funs.py
```diff
@@ -429,6 +433,57 @@ def test(eq):
...
+        eq = Equilibrium(L=3, M=3, N=3, Psi=np.pi)
+        sheet_current = FourierCurrentPotentialField()
+        sheet_current.change_resolution(M=eq.M, N=eq.N, NFP=eq.NFP, sym=eq.sym)
+        eq.solve()
+        obj = BoundaryError(eq, coilset, sheet_current, field_grid=coil_grid)
+        obj.build()
+        f = obj.compute_scaled_error(*obj.xs(eq, sheet_current))
+        n = len(f) // 3
+        # first n should be B*n errors
+        np.testing.assert_allclose(f[:n], 0, atol=1e-4)
```
dpanici, line 451: can you explain this test a bit? we expect a low Bn bc it is axisymmetric in a dense solenoid? and we expect 0 K bc it is 0 pressure at edge?

dpanici (changes requested):
- a tutorial on using the free bdry solve would be good (you wrote one for simons so could even just use that), even for a simple case like a tokamak in ripple or something, or a vacuum field.
- a check to ensure that the `src_grid` has only a surface at `rho=1` if it is passed into the bdry error objective
- clearly lay out sign conventions for the normal vector and K, in the `virtual_casing_biot_savart` and also inside the bdry error objectives
- document the basis expected and output by the `singular_integral` function
- and some other miscellaneous comments

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,840 @@
...
+            UserWarning,
+            f"Boundary pressure is nonzero (max {edge_pres} Pa), "
+            + "a sheet current should be included.",
+        )
+
+        self._eq_data_keys = [
+            "K_vc",
+            "B",
+            "|B|^2",
+            "B",
```
dpanici, line 441: `"B"` is listed here twice

desc/magnetic_fields.py
```diff
@@ -964,10 +1110,29 @@ def from_field(
...
-            method,
-            extrap,
-            period,
+            currents=1.0,
+            NFP=NFP,
+            method=method,
+            extrap=extrap,
+        )
+
+    def tree_flatten(self):
```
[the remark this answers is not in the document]
f0uriest: yeah its because the default flattening thing treats all floats as variables, but we want `period` to be static

desc/nestor.py
f0uriest: Not that I know of? This is basically rotating a vector from a single field period to the other periods.

desc/objectives/_free_boundary.py
f0uriest: yeah i need to add field here

desc/objectives/_free_boundary.py
f0uriest: yep

tests/test_objective_funs.py
f0uriest: Yeah its basically just a 1/R toroidal field, so it should be tangent to the surface, and its scaled such that the pressure matches so all the errors should be small.

desc/compute/utils.py
```diff
@@ -1305,35 +1305,3 @@ def body(i, mins):
     # The above implementation was benchmarked to be more efficient than
     # alternatives without explicit loops in GitHub pull request #501.
     return grid.expand(mins, surface_label)
```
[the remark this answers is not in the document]
f0uriest: this should have been removed a while ago, its just a leftover

desc/perturbations.py
f0uriest: no this is leftover from an earlier version, will remove.

desc/objectives/_free_boundary.py
f0uriest: I was using the opposite sign convention to the current potential stuff, it should be fixed now. I've added a more complete explanation of the virtual casing to the docsting of that function, for the objective I'm just focusing on what equations its actually minimizing rather than how its calculated.

desc/objectives/_free_boundary.py
dpanici: I guess we can leave that to a future PR though? I will make an issue just to keep a tab on this #855

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,882 @@
...
+
+        timer.stop("Precomputing transforms")
+        if verbose > 1:
+            timer.disp("Precomputing transforms")
+
+        self._dim_f = 2 * grid.num_nodes
+
+        if self._normalize:
+            scales = compute_scaling_factors(eq)
+            self._normalization = scales["B"] * scales["R0"] * scales["a"]
```
dpanici, line 180: shouldn't the normalization be this for the first half of the returned value Bn, but should have a `scales["B"]**2` for the 2nd half where it is the difference between the squared B magnitudes?

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,882 @@
...
+        eq_params : dict
+            Dictionary of equilibrium degrees of freedom, eg Equilibrium.params_dict
+        constants : dict
+            Dictionary of constant data, eg transforms, profiles etc. Defaults to
+            self.constants
+
+        Returns
+        -------
+        f : ndarray
+            Boundary force error (N).
```
dpanici, line 198: are these in `N`? but the print statement says it is `T` and `T^2`?

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,882 @@
...
+            target = 0
+        self._src_grid = src_grid
+        self._eval_grid = eval_grid
+        self._s = s
+        self._q = q
+        self._ext_field = ext_field
+        self._field_grid = field_grid
+        self._loop = loop
+        self._sheet_current = None
+        things = [eq]
```
dpanici, line 390: same note as above on including `ext_field` in things, if it is easy to do now it might be worth doing that in this PR

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,882 @@
...
+        Whether to compute the error in physical units or non-dimensionalize.
+    normalize_target : bool
+        Whether target and bounds should be normalized before comparing to computed
+        values. If `normalize` is `True` and the target is in physical units,
+        this should also be set to True.
+    s, q : integer
+        Hyperparameters for singular integration scheme, s is roughly equal to the size
+        of the local singular grid with respect to the global grid, q is the order of
+        integration on the local grid
+    src_grid, eval_grid : Grid, optional
```
dpanici, line 343: may be worth specifying in the doc that the `src_grid` must be `sym=False`, and that the `src_grid` is specifically for evaluating the singular integral on the plasma surface

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,882 @@
...
+
+        timer.stop("Precomputing transforms")
+        if verbose > 1:
+            timer.disp("Precomputing transforms")
+
+        self._dim_f = neq * eval_grid.num_nodes
+
+        if self._normalize:
+            scales = compute_scaling_factors(eq)
+            self._normalization = scales["B"] * scales["R0"] * scales["a"]
```
dpanici, line 543: same comment as above, is this normalization actually the correct one for the other errors (in `K` and in `Bsq`)?

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,882 @@
...
+            curr > 1e-8,
+            UserWarning,
+            f"Current is non-zero (max {curr} A), "
+            + "VacuumBoundaryError will be incorrect.",
+        )
+
+        self._eq_data_keys = [
+            "B",
+            "R",
+            "zeta",
```
dpanici, line 151: should this be `phi`, not `zeta`? For when general toroidal angle is implemented `zeta` won't be the same as `phi` but we need this to be the toroidal cylindrical angle `phi` for `compute_magnetic_field`

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,882 @@
...
+            f"Boundary pressure is nonzero (max {edge_pres} Pa), "
+            + "a sheet current should be included.",
+        )
+
+        self._eq_data_keys = [
+            "K_vc",
+            "B",
+            "|B|^2",
+            "R",
+            "zeta",
```
dpanici, line 487: same comment on `zeta` versus `phi` from above, this should probably be `phi`

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,882 @@
...
+        )
+        eval_data = compute_fun(
+            "desc.equilibrium.equilibrium.Equilibrium",
+            self._eq_data_keys,
+            params=eq_params,
+            transforms=constants["eval_transforms"],
+            profiles=constants["eval_profiles"],
+        )
+        if self._sheet_current:
+            assert sheet_params is not None
```
dpanici, line 583: is this a debug statement? Is it still needed? Shouldn't this get caught somewhere else first, like in the optimizer when it checks that the set of passed-in things is the same as obj.things, if the sheet current is not passed in with `things`?

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+
+
+class FFTInterpolator(_BIESTInterpolator):
+    """FFT interpolation operator required for high order singular integration.
+
+    Parameters
+    ----------
+    eval_grid, src_grid : Grid
+        Evaluation and source points for the integral transform.
+        src_grid should be a LinearGrid
```
dpanici, line 109: state that it needs to be sym=False in the description

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+
+    def __call__(self, f, i):
+        """Interpolate data to polar grid points.
+
+        Parameters
+        ----------
+        f : ndarray
+            Data at source grid points to interpolate.
+        i : int
+            Index of polar node.
```
dpanici, line 201: How do these correspond to the `eval_grid`? Or are they separate?

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+
+
+class DFTInterpolator(_BIESTInterpolator):
+    """Fourier interpolation matrix required for high order singular integration.
+
+    Parameters
+    ----------
+    eval_grid, src_grid : Grid
+        Evaluation and source points for the integral transform.
+        src_grid should be a LinearGrid
```
dpanici, line 231: for `dft` the grid can be stellarator symmetric?

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+    eval_grid,
+    src_data,
+    src_grid,
+    kernel,
+    interpolator,
+    loop=False,
+):
+    """Evaluate a singular integral transform on a surface.
+
+    eg f(θ, ζ) = ∫ ∫ K(θ, ζ, θ', ζ') g(θ', ζ') dθ' dζ'
```
dpanici, line 497: I know this is in the paper but can you also define here to specify that K is the kernel and `g` is the metric (right?)

dpanici: Or can you cite the equation? (this one seems like eqn 3.7 in [1], though without the $\sigma` function)

this would be helpful down the line for when we are deceased and others are maintaining the code in a 100 years.

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+        Returns
+        -------
+        fi : ndarray
+            Source data interpolated to ith polar node.
+        """
+        return self._mat[:, i] @ f
+
+
+def _chi(rho):
+    """Partition of unity function."""
```
dpanici, line 326: eqn 39 in [2]

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+    return jnp.exp(-36 * jnp.abs(rho) ** 8)
+
+
+def _rho(theta, zeta, theta0, zeta0, dtheta, dzeta, s):
+    """Polar grid radial coordinate."""
```
dpanici, line 331: defined in eqn 36 (?) in [2]

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+                src_theta,
+                src_data["zeta"],  # to account for different field periods
+                eval_theta[i],
+                eval_zeta[i],
+                h_t,
+                h_z,
+                s,
+            )
+
+            eta = _chi(rho)
```
dpanici, line 380: eta from eqn 36 of [2]

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+        else:
+            fj = vmap(eval_pt_vmap)(jnp.arange(eval_grid.num_nodes))
+
+        f += fj.reshape((eval_grid.num_nodes, kernel.ndim))
+        return f, src_data
+
+    f = jnp.zeros((eval_grid.num_nodes, kernel.ndim))
+    f, _ = fori_loop(0, src_grid.NFP, nfp_loop, (f, src_data))
+
+    # we sum distance vectors, so they need to be in xyz for that to work
```
dpanici, line 404: where is this sum happening? in the kernel evaluations?

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+
+    eta = _chi(r)
+    v = eta * s**2 * h_t * h_z / 4 * abs(r) * dr * dw
+    keys = list(set(["|e_theta x e_zeta|"] + kernel.keys))
+    fsrc = [src_data[key] for key in keys]
+
+    def polar_pt_vmap(i):
+        # evaluate the effect from a single polar node around each singular point
+        # on that singular point. Polar grids from other singularities have no effect
+        dt = s / 2 * h_t * r[i] * jnp.sin(w[i])
```
dpanici, line 441: in section 3.2.2 of [2] above eqn 38

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+    eval_theta = jnp.asarray(eval_grid.nodes[:, 1])
+    eval_zeta = jnp.asarray(eval_grid.nodes[:, 2])
+
+    r, w, dr, dw = _get_quadrature_nodes(q)
+
+    h_t = jnp.mean(src_dtheta)
+    h_z = jnp.mean(src_dzeta)
+
+    eta = _chi(r)
+    v = eta * s**2 * h_t * h_z / 4 * abs(r) * dr * dw
```
dpanici, line 434: this is the all the terms in the integrad in eqn 38 of [2] except for the G, f, and metric

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+    eval_grid : Grid
+        Points where integral transform is to be evaluated (eg unprimed coordinates).
+    src_data : dict
+        Dictionary of data at source points. Keys should be those required by
+        kernel as kernel.keys. Vector data should be in rpz basis.
+    src_grid : LinearGrid
+        Source points for integral (eg primed coordinates). Should be linearly spaced
+        rectangular grid in both theta, zeta.
+    kernel : str or callable
+        Kernel function to evaluate. Should take 3 arguments:
```
dpanici, line 515: add list of available kernels 
```
{
    "1_over_r": _kernel_1_over_r,
    "nr_over_r3": _kernel_nr_over_r3,
    "biot_savart": _kernel_biot_savart,
    "biot_savart_A": _kernel_biot_savart_A,
}
```

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+def _kernel_biot_savart(eval_data, src_data, diag=False):
+    # K x r / |r|^3
+    src_x = jnp.atleast_2d(
+        rpz2xyz(jnp.array([src_data["R"], src_data["zeta"], src_data["Z"]]).T)
+    )
+    eval_x = jnp.atleast_2d(
+        rpz2xyz(jnp.array([eval_data["R"], eval_data["zeta"], eval_data["Z"]]).T)
+    )
+    if diag:
+        dx = eval_x - src_x
```
dpanici, line 616: I get that `diag=False` means it will find the B (for this kernel at least) at each point in eval_data with contributions from every other point in `src_data`...

But what does `diag=True` do? it seems like it just takes everypoint in `eval_data` and then subtracts from the corresponding points (`src_x` and `eval_x` must be same size?)
Sorry, I know from the comments it seems that this should be just evaluating the contribution at one eval point from the local polar grid and not the other points' polar grids, but it is not clear to me how that is enforced with the code, can you clarify it for me?

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+    ----------
+    eval_data : dict
+        Dictionary of data at evaluation points. Keys should be those required by
+        kernel as kernel.keys. Vector data should be in rpz basis.
+    eval_grid : Grid
+        Points where integral transform is to be evaluated (eg unprimed coordinates).
+    src_data : dict
+        Dictionary of data at source points. Keys should be those required by
+        kernel as kernel.keys. Vector data should be in rpz basis.
+    src_grid : LinearGrid
```
dpanici, line 703: Any check that the source and eval grids are on a single surface. and that is the same surface? that is required right? we check in the objectives but would be good to have somewhere in this one as well since this is a public function

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+        assert islinspaced(
+            src_theta, axis=1
+        ), "singular integration requires source nodes be equally spaced in theta"
+        assert islinspaced(
+            src_zeta, axis=0
+        ), "singular integration requires source nodes be equally spaced in zeta"
+
+        # need eval_grid to be linearly spaced in theta, zeta,
+        # and contain only 1 rho value
+        assert isalmostequal(
```
dpanici, line 152: could add check here that the 2 grids' `rho` values are the same

desc/singularities.py
```diff
@@ -0,0 +1,733 @@
...
+            diag : boolean, whether to evaluate full cross interactions or just diagonal
+        If a callable, should also have the attributes ``ndim`` and ``keys`` defined.
+        ``ndim`` is an integer representing the dimensionality of the output function f,
+        1 if f is scalar, 3 if f is a vector, etc.
+        ``keys`` is a list of strings of what data is required to evaluate the kernel.
+        The kernel will be called with dictionaries containing this data at source and
+        evaluation points.
+        If vector valued, the input to the kernel function will be in rpz and output
+        should be in xyz.
+    interpolator : callable
```
dpanici, line 527: these themselves have their own grids, which I assume should be the same as the `eval_grid` and `src_grid` being  passed into the singular integral call. Should add a check that the interpolator grids are the same as the passed-in grids

dpanici (changes requested):
Real close to being ready to merge I think, I had a few changes and left some comments
- [ ] #855 We could resolve this in this PR if you want, but if you think that is too much change for this and want to get it merged in first, we can wait and do as a separate PR, I am fine with waiting so I'll leave it up to you
- [x] fix normalization for `VacuumBoundaryError` and `BoundaryError` to normalize by `B^2 * Area` for the B squared difference part of the objective (right now both are normalized by just `B * Area`)
- [ ] Use `phi` instead of `zeta` key for the toroidal coordinate calculation in the free boundary objectives
- [ ] Add references in `singularities.py` to equations in Malhotra's 2 papers so it is easier to understand (at least for me it would help if I could refer to the papers by eqn number to see what each part of the file is doing, I added a few comments with them already)
- [ ] add check that the eval and src grid have the same `rho` value
- [ ] The interpolators themselves have their own grids, which I assume should be the same as the `eval_grid` and `src_grid` being  passed into the singular integral call. Should add a check that the interpolator grids are the same as the passed-in grids
- [ ] Test with the sheet current? Even just a super simple one where it is included in a vacuum case we know to be nested, so that K should be small, or even a skeleton script in the tutorial notebook for if a sheet current is needed, how you need to add K as an optiimzable thing and use `optimizer.optimize` instead of `eq.optimize`

desc/singularities.py
f0uriest: this corresponds to the `i`th polar node ($\rho_i$, $\omega_i$) around each ($\theta$, $\zeta$) point in eval grid

desc/objectives/_free_boundary.py
f0uriest: Yeah wasn't sure how to handle this. As it currently stands `normalization` is a single scalar for each objective, assuming that each objective only returns one thing. But for the free boundary stuff its super wasteful to split the different outputs into different objectives.

desc/objectives/_free_boundary.py
f0uriest: I think that will need some extra testing etc to make sure it works for arbitrary nested collections of `MagneticField` types.

We would also need some non-trivial constraints/objectives for coils/fields etc.

desc/objectives/_free_boundary.py
f0uriest: The singular integral stuff current assumes zeta==phi. It could be generalized but I'd have to think about it more so I just kept everything as zeta for now.

desc/objectives/_free_boundary.py
f0uriest: i think its still a useful check, in the event `obj.compute` is called by itself.

desc/singularities.py
f0uriest: yeah the eval grid can be anything, but the source grid should still be linear (for accuracy, not for logic)

desc/singularities.py
f0uriest: if diag is False, then dx is a NxNx3 array, so coupling all src pts to all eval pts. If diag is True its just Nx3 (the diagonal only), so the polar points around each eval point only contribute to their particular eval pt, not neighbors.

desc/singularities.py
f0uriest: It happens when we sum the contribution from different source points

desc/singularities.py
f0uriest: This isn't formally a public function (not listed in the docs page), more like the compute utils that are only for devs. We can't do runtime checking here because this function is jit compiled, so we have to do any checking in the objective, but I can add more to the docstring.

desc/singularities.py
f0uriest: Thats a good point, we can actually avoid having to pass in the grids directly themselves

desc/singularities.py
```diff
@@ -0,0 +1,758 @@
...
+
+        # data interpolated to each eval pt offset by dt,dz
+        src_data_polar = {key: interpolator(val, i) for key, val in zip(keys, fsrc)}
+
+        # can't interpolate phi directly since its not periodic, so we interpolate
+        # omega and add it back in
+        src_data_polar["zeta"] = zeta_i
+        src_data_polar["theta"] = theta_i
+        if "phi" in keys:
+            src_data_polar["phi"] = zeta_i + src_data_polar["omega"]
```
dpanici, line 478: for now though this effectively does nothing since we dont use omega?

desc/singularities.py
```diff
@@ -0,0 +1,758 @@
...
+        should be those required by kernel as kernel.keys. Vector data should be in
+        rpz basis.
+    kernel : str or callable
+        Kernel function to evaluate. If str, one of the following:
+            '1_over_r' : 1 / |𝐫 − 𝐫'|
+            'nr_over_r3' : 𝐧⋅(𝐫 − 𝐫') / |𝐫 − 𝐫'|³
+            'biot_savart' : μ₀/4π 𝐊×(𝐫 − 𝐫') / |𝐫 − 𝐫'|³
+            'biot_savart_A' : μ₀/4π 𝐊 / |𝐫 − 𝐫'|
+        If callable, should take 3 arguments:
+            eval_data : dict of data at evaluation points (primed)
```
dpanici, line 547: is the primed and unprimed here flipped? unprimed is source right since that is what the integral is over?

desc/singularities.py
```diff
@@ -0,0 +1,758 @@
...
+            src_data : dict of data at source points (unprimed)
+            diag : boolean, whether to evaluate full cross interactions or just diagonal
+        If a callable, should also have the attributes ``ndim`` and ``keys`` defined.
+        ``ndim`` is an integer representing the dimensionality of the output function f,
+        1 if f is scalar, 3 if f is a vector, etc.
+        ``keys`` is a list of strings of what data is required to evaluate the kernel.
+        The kernel will be called with dictionaries containing this data at source and
+        evaluation points.
+        If vector valued, the input to the kernel function will be in rpz and output
```
dpanici, line 556: this is true even if it is not vector valued (like the `1/r` kernel you have), right? the src_data needs to be `rpz`?

desc/singularities.py
dpanici: so for `diag=True`:
`dx` is N x 3 array
`num = K x dx` is also an `N x 3` array 
`r` before the `if diag` statement is an `(N,) ` array
`r = r[:, None]` makes it an `(N,1)` array again
then the final result is just doing the kernel, evaluating once the K at each node?

This is called in the _singular_part as
```python
k = kernel(
            {key: val for key, val in eval_data.items() if key in keys},
            src_data_polar,
            diag=True,
        ).reshape((eval_grid.num_nodes, kernel.ndim))
```
here src_data_polar is the polar grid points around a single point on the eval_grid? I still don't quite see how `diag` makes this do what we want it to do

dpanici (changes requested):
Couple tiny docs comments, and I still don't think I saw a test where the sheet current was shown in an optimization? Did I miss it?

desc/singularities.py
```diff
@@ -0,0 +1,758 @@
...
+    dz = jnp.minimum(dz, 2 * np.pi - dz)
+    return 2 / s * jnp.sqrt((dt / dtheta) ** 2 + (dz / dzeta) ** 2)
+
+
+def _nonsingular_part(eval_data, eval_grid, src_data, src_grid, s, kernel, loop=False):
+    """Integrate kernel over non-singular points.
+
+    Generally follows sec 3.2.1 of [2].
+    """
+    assert isinstance(src_grid.NFP, int)
```
ddudt, line 350: I'm getting errors here in some examples where NFP is of type `numpy.int64`

desc/singularities.py
f0uriest: no, here `src_data_polar` contains one polar point around _each_ evaluation point, and we loop over the different polar offsets `dtheta[i]` and `dzeta[i]`

If we have N eval pts and M polar points around each eval pt, this `src_data_polar` is size N, not size M. We do it this way because we can then vectorize the fft interpolation, since the offsets around each eval pt are the same.

The diag part is basically making sure that polar node i around evaluation point j only contributes to the result at eval pt j, since they're in theory entirely independent calculations, we just vectorize them for speed and efficiency.

desc/singularities.py
f0uriest: Correct, but I wanted to include the logic here for future proofing since naively interpolating phi the same way we do R,Z wouldn't work since phi itself has a secular term

desc/singularities.py
f0uriest: yeah ill flip this

desc/singularities.py
f0uriest: I guess technically, but if there arent't any vectors than it doesnt matter. The eval/src pts themselves are always built from 1d arrays R, phi, Z

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,945 @@
...
+        else:
+            src_grid = self._src_grid
+
+        if self._eval_grid is None:
+            eval_grid = LinearGrid(
+                rho=np.array([1.0]),
+                M=eq.M_grid,
+                N=eq.N_grid,
+                NFP=int(eq.NFP),
+                sym=False,
```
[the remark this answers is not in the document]
f0uriest: In theory the eval grid can be anything, but to work with the FFT interpolation it has to be sym=False, so thats the default. Using sym=True would require the DFT interpolation which is way slower but still formally allowed if you need to evaluate at arbitrary points

desc/singularities.py
```diff
@@ -0,0 +1,758 @@
...
+    dz = jnp.minimum(dz, 2 * np.pi - dz)
+    return 2 / s * jnp.sqrt((dt / dtheta) ** 2 + (dz / dzeta) ** 2)
+
+
+def _nonsingular_part(eval_data, eval_grid, src_data, src_grid, s, kernel, loop=False):
+    """Integrate kernel over non-singular points.
+
+    Generally follows sec 3.2.1 of [2].
+    """
+    assert src_grid.NFP == int(src_grid.NFP)
```
ddudt, line 350: I'm getting errors here when NFP is of type `numpy.int64`

f0uriest: Can you provide an example? That should work fine for any integer type

ddudt: Actually it is initially type `numpy.int64` but then during the step: 
`Compiling objective function and derivatives: ['Boundary error']`
it somehow becomes type `numpy.ndarray` and that throws the error. 

I'm trying to reproduce it with an equilibrium I can share...

desc/singularities.py
dpanici: Ok so this function then will calculate, in one call, the contribution to the integral from one polar point around the intended eval point, and `diag=True` makes sure of that, then we loop over all the offsets to then form the full polar contribution for each eval point. That makes more sense, thanks for the clarification

desc/singularities.py
dpanici: wait I am confusing myself after looking at this comment again, I meant I thought conventionally primed is used for source and unprimed for eval... though this is mionr I guess and now consistent with your kernel docstring above

dpanici (dismissed):

dpanici (changes requested):
Not to be annoying but now we have `source_grid` for our coils and `src_grid` for the free boundary objectives... can we just uniformly use `source_grid` everywhere?

dpanici (changes requested):
There is a typo in the free bdry notebook, and I had one question, which is can we do one of the fields in that tutorial using DESC MagneticFIelds/coils to then make the SplineMagneticField, instead of both using mgrid?

desc/singularities.py
f0uriest: I think it should be consistent now, primed is the source point, unprimed is eval pt

f0uriest:
> There is a typo in the free bdry notebook, and I had one question, which is can we do one of the fields in that tutorial using DESC MagneticFIelds/coils to then make the SplineMagneticField, instead of both using mgrid?

In principle yes, I didn't do that for a few reasons:
- using mgrid is probably the most relevant for users since thats what they already have for vmec
- its usually the most efficient way to evaluate a field, rather than doing biot savart
- I don't have the coil files for either of these test cases, so we'd have to generate our own coils or something

That said, I'm certainly open to the idea, I think it would be a good thing to show for one of the examples, but it can probably wait for another PR. Ideally we'd also figure out the API for coil optimization since that might affect things (ie, for regular free boundary do we need some sort of `coil_fixed` parameter)

dpanici (approved):

dpanici (approved):

tests/test_examples.py
```diff
@@ -875,6 +878,113 @@ def test_only_non_eq_optimization():
...
+        -3.560639108717275e05,
+        -6.588434719283084e04,
+        -1.154387774712987e04,
+        -1.153546510755219e04,
+        -6.588300858364606e04,
+        -3.560589388468855e05,
+    ]
+    ext_field = SplineMagneticField.from_mgrid(
+        r"tests/inputs/mgrid_solovev.nc", extcur=extcur
+    )
```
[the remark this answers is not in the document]
ddudt: The `raw_coil_cur` array in the mgrid file itself is all ones, as is normal for raw mode

tests/test_examples.py
```diff
@@ -875,6 +878,113 @@ def test_only_non_eq_optimization():
...
 
+@pytest.mark.regression
+@pytest.mark.solve
+@pytest.mark.slow
+def test_freeb_vacuum():
+    """Test for free boundary vacuum stellarator."""
+    extcur = [4700.0, 1000.0]
+    ext_field = SplineMagneticField.from_mgrid(
+        "tests/inputs/mgrid_test.nc", extcur=extcur
+    )
```
ddudt, line 889: Where did you get these coil currents from? This mgrid file is in "raw" mode, but the coil currents given in the file itself are 96 and 124

docs/notebooks/tutorials/mgrid_test.nc
ddudt, line 1: Is this file a duplicate of the one we already have in `tests/inputs`? Don't need to add it again here

tests/test_singularities.py
```diff
@@ -0,0 +1,160 @@
...
+        loop=True,
+    )
+    # need extra factor of B/2 bc we're evaluating on plasma surface
+    Bplasma += eval_data["B"] / 2
+    Bplasma = np.linalg.norm(Bplasma, axis=-1)
+    # scale by total field magnitude
+    B = Bplasma / np.mean(np.linalg.norm(eval_data["B"], axis=-1))
+    # this isn't a perfect vacuum equilibrium (|J| ~ 1e3 A/m^2), so increasing
+    # resolution of singular integral won't really make Bplasma less.
+    np.testing.assert_array_less(B, 0.05)
```
ddudt, line 124: Why is `Bplasma` normalized here? And what is it normalized to, the local value of |B|? Is it checking that normalized value is <5%?

tests/test_objective_funs.py
```diff
@@ -1630,6 +1970,47 @@ def test_objective_no_nangrad():
...
+        -3.560639108717275e05,
+        -6.588434719283084e04,
+        -1.154387774712987e04,
+        -1.153546510755219e04,
+        -6.588300858364606e04,
+        -3.560589388468855e05,
+    ]
+    ext_field = SplineMagneticField.from_mgrid(
+        r"tests/inputs/mgrid_solovev.nc", extcur=extcur
+    )
```
ddudt, line 1990: you don't actually need to give `extcur` for this test, the default values will be fine

tests/test_objective_funs.py
ddudt, line 1: General comment for the new unit tests: the magnetic fields you are using from `CoilSet` are probably really small, because you are using the default coil currents of 1 Amp. So this makes it difficult to tell if your absolute tolerances are actually good tolerances or not, because it's not clear what the magnitude of the magnetic fields should actually be. I would recommend setting the coil currents so that |B| at the evaluation points are on the order of 1 T.

desc/singularities.py
```diff
@@ -0,0 +1,768 @@
...
+        original Malhotra papers.
+    q : int
+        Order of quadrature in polar domain
+
+    """
+
+    _io_attrs_ = _BIESTInterpolator._io_attrs_ + ["_h_t", "_h_z", "_st", "_sz"]
+
+    def __init__(self, eval_grid, source_grid, s, q):
+        # current fft interpolating can't handle symmetric grids correctly
```
ddudt, line 121: Is this something we could fix in the future? If so, make sure we have an Issue for it so we don't forget. I assume it could be a substantial speedup.

desc/singularities.py
```diff
@@ -0,0 +1,768 @@
...
+    if kernel.ndim == 3:
+        f = xyz2rpz_vec(f, phi=eval_data["phi"])
+    return f
+
+
+def _singular_part(
+    eval_data,
+    eval_grid,
+    src_data,
+    source_grid,
```
ddudt, line 437: Let's be consistent with either `src` or `source` in our variable names. I think `source` is what we are using most other places already.

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,945 @@
...
+from .normalization import compute_scaling_factors
+
+
+class VacuumBoundaryError(_Objective):
+    """Target for free boundary conditions on LCFS for vacuum equilibrium.
+
+    Computes the residuals of the following:
+
+    𝐁ₒᵤₜ ⋅ 𝐧 = 0
+    𝐁ₒᵤₜ² - 𝐁ᵢₙ² = 0
```
ddudt, line 30: The actual compute function adds an area element weighting. We document that similar thing in the equations here for the force balance objectives, and I think it would be good to include here too. So that it's obvious what the units of the residuals are.

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,945 @@
...
+        Grid used to discretize ext_field.
+    name : str
+        Name of the objective function.
+
+    """
+
+    _scalar = False
+    _linear = False
+    _print_value_fmt = "Boundary Error: {:10.3e} "
+    _units = "(mixed units)"
```
ddudt, line 81: List the different units, like `("(T)", "(T^2)")` or whatever they are. I don't think this gets used since you override it in the custom `print` method anyways, but it's nice as documentation of what the units are.

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,945 @@
...
+        less memory.
+    name : str
+        Name of the objective function.
+
+    """
+
+    _scalar = False
+    _linear = False
+    _print_value_fmt = "Boundary Error: {:10.3e} "
+    _units = "(mixed units)"
```
ddudt, line 388: See previous comments about units, both here and in the doc string.

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,945 @@
...
+        eval_profiles = get_profiles(self._eq_data_keys, obj=eq, grid=eval_grid)
+        eval_transforms = get_transforms(self._eq_data_keys, obj=eq, grid=eval_grid)
+
+        neq = 3 if self._sheet_current else 2  # number of equations we're using
+
+        self._constants = {
+            "eval_transforms": eval_transforms,
+            "eval_profiles": eval_profiles,
+            "src_transforms": src_transforms,
+            "src_profiles": src_profiles,
```
ddudt, line 544: See other comment about consistently using `source_` instead of `src_`. But I'm being a little picky here, since these are internal variables that the user shouldn't be interacting with.

desc/objectives/_free_boundary.py
```diff
@@ -0,0 +1,945 @@
...
+            # mean abs makes more sense than mean
+            fi = jnp.abs(fi) if abserr else fi
+            wi = w[i * nn : (i + 1) * nn]
+            fmax = jnp.max(fi)
+            fmin = jnp.min(fi)
+            fmean = jnp.mean(fi * wi) / jnp.mean(wi)
+            _print(fmt, fmax, fmin, fmean, norm, unit)
+
+
+class BoundaryErrorNESTOR(_Objective):
```
ddudt, line 755: We don't have any examples of using this objective, right? I think I only see it covered in a single unit test. That might be fine if we don't really want people to use it, but I want to understand how well it is tested.

ddudt (changes requested):

desc/nestor.py
```diff
@@ -0,0 +1,1211 @@
...
+    if vec.shape[0] == 3:
+        zz = jnp.broadcast_to(z.reshape((*shp, 1)), (*shp, zetas.size))
+        return jnp.array((xx, yy, zz))
+    return jnp.array((xx, yy))
+
+
+def eval_surface_geometry(
+    R_lmn, Z_lmn, Rb_transform, Zb_transform, ntheta, nzeta, NFP, sym
+):
+    """Evaluate coordinates and derivatives on the surface.
```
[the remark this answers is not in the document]
f0uriest: These are mostly from Jonathan. We could replace most of these with our compute stuff but I'd prefer minimize the changes to all the nestor stuff

desc/nestor.py
```diff
@@ -0,0 +1,1211 @@
...
+        * (
+            normal["R_n"] * (coords["R_zz"] - coords["R"])
+            + normal["Z_n"] * coords["Z_zz"]
+        )
+    ) / NFP**2
+    return jacobian
+
+
+def biot_savart(eval_pts, coil_pts, current):
+    """Biot-Savart law following [1].
```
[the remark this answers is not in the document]
f0uriest: Not really, this one is specific to the way the plasma current in nestor is modelled as a filament. The one for BIEST is a volume biot-savart

desc/nestor.py
```diff
@@ -0,0 +1,1211 @@
...
+    Btot["Bn"] = (
+        normal["R_n"] * Btot["BR"]
+        + normal["phi_n"] * Btot["Bphi"]
+        + normal["Z_n"] * Btot["BZ"]
+    )
+
+    return Btot
+
+
+class Nestor(IOAble):
```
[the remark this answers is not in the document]
f0uriest: I agree we could simplify this, but it's generally only used for comparisons so don't really feel like it

tests/test_examples.py
```diff
@@ -875,6 +878,113 @@ def test_only_non_eq_optimization():
...
+    eq, out = eq.optimize(
+        objective,
+        constraints,
+        optimizer="proximal-lsq-exact",
+        verbose=3,
+        options={},
+    )
+    rho_err, _ = area_difference_vmec(eq, "tests/inputs/wout_test_freeb.nc")
+
+    np.testing.assert_allclose(rho_err[:, -1], 0, atol=4e-2)  # only check rho=1
```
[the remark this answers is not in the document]
f0uriest: We're just checking that the boundary shape is the same, the resolution is a bit too low to fully resolve the volume.

docs/notebooks/tutorials/mgrid_solovev.nc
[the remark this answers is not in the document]
f0uriest: They are the same. I chose to put them in both places to keep them independent and self contained, since we may change the tutorial in the future, or change the tests and don't want them linked together.

tests/test_examples.py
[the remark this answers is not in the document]
f0uriest: These are the currents from the VMEC input which gives the correct field and boundary shape. This is why I was concerned about the raw vs scaled stuff, because even though this one is "raw" the extcur is still needed to get the right field

tests/test_examples.py
f0uriest: See above

tests/test_singularities.py
f0uriest: Yes its normalizing by the average |B| on the surface. This is very close to vacuum so we expect the plasma part to be a very small fraction of the total B

tests/test_objective_funs.py
f0uriest: see above

desc/objectives/_free_boundary.py
f0uriest: Yeah its both slower and less accurate so it's not really used. It's also somewhat memory intensive so it's hard to come up with a real example test that can run on CI

tests/test_objective_funs.py
f0uriest: The initial coil has a current of 5e5 amps, so thats what the coilset uses. It's scaled so that the total field from all the coils is ~1T on axis, same as the field from the equilibrium.

tests/test_examples.py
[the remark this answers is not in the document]
ddudt: I don't think this checks the boundary shape, I believe it only compares the volumes/areas within that surface. If you want to check the boundary you should compare the boundary coefficients.

docs/notebooks/tutorials/mgrid_solovev.nc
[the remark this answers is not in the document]
ddudt: I don't like the idea of having duplicate files. If we do end up changing them in the future, then they definitely don't need to refer to the same file so that won't be an issue anyways. And the tests for the notebooks would catch if something breaks.

tests/test_examples.py
ddudt: Again, how did you choose the numbers 4700 and 1000? They feel arbitrary. If you copied it from an example somewhere, reference that example.

tests/test_objective_funs.py
ddudt: Oh I missed that. Great

desc/nestor.py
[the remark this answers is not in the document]
ddudt: OK this might be controversial, but I think it's worth a discussion: What if we don't include the NESTOR stuff in this PR? 

My reasoning:

1. The code is messy. 
2. It's worse than the BIEST implementation and we don't recommend using it. 
3. We don't have any tutorials, example, or tests for the NESTOR code. 

We are adding this whole new file of duplicate code that does not conform to the API of the rest of the DESC code. Moreover, _we don't even use this code_! It sounds like there is no use case for the NESTOR stuff in the near future, but if we ever decide that we do need it we can easily add it back.

tests/test_examples.py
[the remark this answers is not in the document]
f0uriest: It checks the area of the overlap, so it is effectively checking the shape of the boundary. We can't compare coefficients due to different theta angles etc.

tests/test_examples.py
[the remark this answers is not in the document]
f0uriest: I mean these were the values in the VMEC input that came with the mgrid file, I think from jonathan.

tests/test_examples.py
f0uriest: See above. They're not published anywhere afaik, but these are the values that were in the vmec input along with the mgrid

desc/singularities.py
f0uriest: did you figure out the source of this issue?

dpanici (dismissed):

desc/nestor.py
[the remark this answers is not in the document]
dpanici: Since it is used in the paper I agree it is worth keeping in the code repo. I am fine with it not being in the API documentation maybe with a note in the docstring saying it is not meant for public use or something

dpanici (approved):

ddudt (approved):

PR #726, opened 2023-10-25
ko/coils quad flux

### LINKED ISSUE #901: Objective function for Bnormal

kianorr:
Adds quadratic flux objective function.

Resolves #901


desc/objectives/_free_boundary.py
```diff
@@ -686,57 +713,36 @@ def build(self, use_jit=True, verbose=1):
...
         """Compute boundary force error.
 
         Parameters
         ----------
-        params : dict
-            Dictionary of equilibrium degrees of freedom, eg Equilibrium.params_dict
-        constants : dict
-            Dictionary of constant data, eg transforms, profiles etc. Defaults to
-            self.constants
+        params :
```
YigitElma, line 721: Even though you set the default value to None, you should describe what that stands for. Also, you seem to be not using 'params', consider deleting it completely.
```suggestion
        params : dict
                some explanation about the usage of params
```

desc/objectives/_free_boundary.py
```diff
@@ -686,57 +713,36 @@ def build(self, use_jit=True, verbose=1):
 
         super().build(use_jit=use_jit, verbose=verbose)
 
-    def compute(self, params, constants=None):
+    def compute(self, params=None, constants=None):
```
YigitElma, line 716: Apparently you don't use params, you can get rid of it completely.
```suggestion
    def compute(self, constants=None):
```

desc/objectives/_free_boundary.py
```diff
@@ -686,57 +713,36 @@ def build(self, use_jit=True, verbose=1):
...
-        params : dict
-            Dictionary of equilibrium degrees of freedom, eg Equilibrium.params_dict
-        constants : dict
-            Dictionary of constant data, eg transforms, profiles etc. Defaults to
-            self.constants
+        params :
 
         Returns
         -------
         f : ndarray
```
YigitElma, line 725: You don't return f anymore. Change the name to Bnorm and give description for what that stands for.
```suggestion
        Bnorm: float
                 Some description
```

YigitElma (commented):
Review for APC524 homework, pull request review part.

tests/test_objective_funs.py
```diff
@@ -460,14 +461,44 @@ def test_boundary_error_nestor(self):
...
-        np.testing.assert_allclose(f, 0, atol=1e-3)
+        obj = QuadraticFlux(t_field, eq)
+        obj.build(eq)
+        f = obj.compute()
+        np.testing.assert_allclose(f, 0)
+
+        # nonaxisymmetric surface
+
+        # need nonaxisymmetric with Bplasma ~ 0
+        eq = Equilibrium()
```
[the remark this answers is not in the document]
kianorr: "precise_QA" still has a high J^rho (maybe I'm doing something wrong?). I was planning on checking with a higher resolution and getting back to you but I've had no time.

kianorr:
Not sure if having QuadraticFlux in _free_boundary.py is appropriate but I'm not sure where you all would prefer to put it

desc/objectives/_free_boundary.py
```diff
@@ -572,8 +572,6 @@ def __init__(
         field_grid=None,
         name="Quadratic flux",
     ):
-        if target is None and bounds is None:
-            target = 0
         self._src_grid = src_grid
         self._eval_grid = eval_grid
         self._s = s
```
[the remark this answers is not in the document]
dpanici: Adding a flag like `eq_fixed=False` to this would be good, for cases where we are only optimizing over the external field (coilset i.e. stage 2 optimization) and not the equilibrium, in this case we can precompute all equilibrium quantities once and store them, and also reduces the DOFs to just the external field.

 See e.g. [PlasmaVesselDistance](https://github.com/PlasmaControl/DESC/blob/5d8c1384824133596f09931592fac58504bd3aa3/desc/objectives/_geometry.py#L534) for an example

desc/objectives/_free_boundary.py
```diff
@@ -589,74 +590,240 @@ def compute(self, eq_params, sheet_params=None, constants=None):
...
+
+        self._dim_f = eval_grid.num_nodes
+
+        if self._normalize:
+            scales = compute_scaling_factors(eq)
+            self._normalization = scales["B"]
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params=None, constants=None):
```
dpanici, line 796: add docstrings here

dpanici (changes requested):
- Add docstring to the compute function
- Add an `eq_fixed` option to this so that you can only optiimze an external field given a fixed equilibrium
- Add a test where you optimize a field with quadratic flux, something like this might work I think (since the equilibrum is axisymmetric the optimizer should put the non-axisymmetric part of the external field to zero, i.e. zero out the `Phi_mn`, though if this does not work lmk)
```python
eq = Equilibrium()
surf = eq.surface.copy()
surf.R_lmn[surf.R_basis.get_idx(M=1,N=0)] = 2
surf.Z_lmn[surf.Z_basis.get_idx(M=-1,N=0)] = 2

field = FourierCurrentPotentialField.from_surface(surface=surf, Phi_mn=np.array([1e4]),modes_Phi=np.array([[1,4]])),G=1e4)

constraints=(FixParameter(eq), # this line is unneeded when eq_fixed is implemented
                     FixParameter(field,["I","G","R_lmn","Z_lmn"]))
obj = ObjectiveFunction(QuadraticFlux(ext_field=field.eq=eq))
(eq,field) = optimizer.optimize((eq,field),objective=obj,constraints=constraints,ftol=1e-15,xtol=1e-15)
np.testing.assert_allclose(field.Phi_mn,0,atol=1e-14)
```

tests/test_objective_funs.py
```diff
@@ -535,6 +537,74 @@ def test_quadratic_flux(self):
         # for this to pass with rtol=1e-3, the source resolution needs to be quite high
         np.testing.assert_allclose(f, Bnorm, rtol=1e-2, atol=1e-2)
 
+    def test_quadratic_flux_optimization(self):
+        """Test optimization using quadratic flux."""
+        eq = Equilibrium(M=2, L=2)
```
dpanici, line 542: would be good to comment this a bit to say what the test is doing (optimizing to reduce the Phi_mn which produce non-axisymmetric fields)

dpanici:
- one test where eq_fixed is False, and FixParameter(eq), and do the same test you already have
- copy Daniel's code for having both eq fixed and field fixed flags, and then have field fixed with a ToroidalFIeld and start with nonaxisymmetric eq and optimize it so it becomes axisymmetric

ddudt:
> * one test where eq_fixed is False, and FixParameter(eq), and do the same test you already have
> * copy Daniel's code for having both eq fixed and field fixed flags, and then have field fixed with a ToroidalFIeld and start with nonaxisymmetric eq and optimize it so it becomes axisymmetric

The code to copy for `eq_fixed` and `field_fixed` can be found here: [Omnigenity objective](https://github.com/PlasmaControl/DESC/blob/dd/omnigenity/desc/objectives/_omnigenity.py#L544)

To summarize/clarify the tests:

1. `eq_fixed=True` with an axisymmetric equilibrium target. Start from a non-axisymmetric field, and check that it becomes axisymmetric after optimization. 
2. `field_fixed=True` with an axisymmetric field target. Start from a non-axisymmetric equilibrium, and check that it becomes axisymmetric after optimization. 
3. `eq_fixed=False` and `field_fixed=False`. I would recommend making this a regression test, where you just check that the code runs and the optimization errors decreased.

kianorr:
> To summarize/clarify the tests:
> 
> 1. `eq_fixed=True` with an axisymmetric equilibrium target. Start from a non-axisymmetric field, and check that it becomes axisymmetric after optimization. 
> 2. `field_fixed=True` with an axisymmetric field target. Start from a non-axisymmetric equilibrium, and check that it becomes axisymmetric after optimization. 
> 3. `eq_fixed=False` and `field_fixed=False`. I would recommend making this a regression test, where you just check that the code runs and the optimization errors decreased. 

For dev meeting: first two tests are passing and the third passes but I'm making sure the third makes sense.

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+        new_Zb_lmn = things_with_field_fixed[0].Rb_lmn[
+            np.where(eq.surface.R_basis.modes[:, 2] != 0)
+        ]
+
+        np.testing.assert_allclose(new_Rb_lmn, 0, atol=1e-10)
+        np.testing.assert_allclose(new_Zb_lmn, 0, atol=1e-10)
+
+        # test with eq_fixed = field_fixed = False: nonaxisymmetric field and
+        # nonaxisymmetric equilibrium
+        eq = Equilibrium(M=2, L=2, N=4)
```
ddudt, line 699: This is an axisymmetric equilibrium with the default surface. Also it is not in force balance. Why not use `eq = nonaxisym_eq.copy()`?

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+
+        # for this to pass with rtol=1e-3, the source resolution needs to be quite high
+        np.testing.assert_allclose(f, Bnorm, rtol=1e-2, atol=1e-2)
+
+    def test_quadratic_flux_optimization(self):
+        """Test optimization using quadratic flux."""
+
+        def get_grids(eq, field):
+            N_grid = 20
+            M_grid = 10
```
ddudt, line 551: I suggest making these input parameters so they can be chosen separately for each test

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+        plasma_vessel_obj = PlasmaVesselDistance(
+            surface=field,
+            eq=eq,
+            plasma_grid=source_grid,
+            surface_grid=surface_grid,
+            bounds=(0.2, 2),
+        )
+        mean_curvature_obj = MeanCurvature(eq, bounds=(-3, 3), grid=surface_grid)
+        objective = ObjectiveFunction(
+            (quadflux_obj, plasma_vessel_obj, mean_curvature_obj)
```
ddudt, line 726: `plasma_vessel_obj` is to prevent the equilibrium from intersecting the `field`? Why is `mean_curvature_obj` needed?

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+        things, __ = nonlinear_optimizer.optimize(
+            (field, eq),
+            objective=objective,
+            constraints=constraints,
+            ftol=0,
+            gtol=0,
+            copy=True,
+        )
+
+        np.testing.assert_allclose(things[0].Phi_mn, 0, atol=1e-7)
```
ddudt, line 745: This is basically just a repeat of test 1. I suggest evaluating $B \cdot n$ before and after the optimization, and checking that the errors decreased sufficiently. We don't want to force the solution to be axisymmetric for this test.

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+        objective = ObjectiveFunction(
+            (quadflux_obj, plasma_vessel_obj, mean_curvature_obj)
+        )
+
+        constraints = (
+            ForceBalance(eq),
+            FixPressure(eq),
+            FixCurrent(eq),
+            FixPsi(eq),
+            FixParameter(field, ["I", "G", "R_lmn", "Z_lmn"]),
```
ddudt, line 734: This constraint shouldn't be necessary for this test, since we don't want to force the field to be axisymmetric. Maybe fixing the surface is helpful, but at least remove the constraints on `"I"` and `"G"`

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+            objective=objective,
+            constraints=constraints,
+            ftol=0,
+            gtol=0,
+            copy=True,
+        )
+
+        np.testing.assert_allclose(things[0].Phi_mn, 0, atol=1e-7)
+
+        # test 4: testing a different magnetic field when field_fixed=False.
```
ddudt, line 747: I don't like this test. You are trying to make an axisymmetric field match a stellarator, which is impossible. I think the other 3 tests will be sufficient

tests/test_objective_funs.py
```diff
@@ -526,6 +793,27 @@ def test(eq):
         test(get("DSHAPE"))
         test(get("HELIOTRON"))
 
+    @pytest.mark.unit
+    def test_target_mean_iota(self):
```
ddudt, line 797: Why is this test in this PR?

ddudt (changes requested):

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
-            mean_iota = jnp.mean(eq.compute("iota", grid=grid)["iota"])
-            obj = RotationalTransform(
-                target=mean_iota, weight=1, eq=eq, loss_function="mean", grid=grid
+        # test that torus (axisymmetric) Bnorm is exactly 0
+        eq = Equilibrium()
+        eq.solve()
+        obj = QuadraticFlux(t_field, eq, eq_fixed=True)
+        obj.build(eq)
+        f = obj.compute(params_1=t_field.params_dict)
+        np.testing.assert_allclose(f, 0)
```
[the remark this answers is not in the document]
dpanici: R0 also does not matter since in the end the function is B = B0 * R0 / R

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+            eq_fixed=True,
+        )
+        Bnorm = t_field.compute_Bnormal(
+            eq.surface, eval_grid=eval_grid, source_grid=source_grid
+        )[0]
+        obj.build(eq, verbose=0)
+        f = obj.compute(params_1=t_field.params_dict)
+
+        # for this to pass with rtol=1e-3, the source resolution needs to be quite high
+        np.testing.assert_allclose(f, Bnorm, rtol=1e-2, atol=1e-2)
```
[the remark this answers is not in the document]
dpanici: Bnorm has values near zero that we are comparing, so  using only rtol fails, here we can use atol but lower like 1e-4 or so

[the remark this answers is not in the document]
dpanici: i.e. it has values between 0 and 0.4 or so

desc/objectives/_free_boundary.py
```diff
@@ -589,74 +590,326 @@ def compute(self, eq_params, sheet_params=None, constants=None):
...
+        integration on the local grid
+    src_grid : Grid, optional
+        Collocation grid containing the nodes for plasma source terms and where
+        to evaluate errors.
+    eval_grid : Grid, optional
+        Collocation grid containing the nodes at which the magnetic field is being
+        calculated and where to evaluate errors.
+    field_grid : Grid, optional
+        Grid used to discretize ext_field (e.g. grid for the magnetic field source from
+        coils).
```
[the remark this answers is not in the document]
dpanici: `src_grid` is used as the source term for the singular integral for the plasma. `eval_grid` is where to evaluate the Bn errors on the plasma surface (That one can be definitely clarified @kianorr ). `field_grid` discretizes the magnetic field object

We don't want these to all be the same, as the singular integral requires a 2D grid which might not be the exact points we want to evaluate the field errors at, and the MagneticField object might be a coil in which case its grid is 1-D

desc/objectives/_free_boundary.py
```diff
@@ -589,74 +590,326 @@ def compute(self, eq_params, sheet_params=None, constants=None):
...
+        target=0,
+        bounds=None,
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        s=None,
+        q=None,
+        src_grid=None,
+        eval_grid=None,
+        field_grid=None,
```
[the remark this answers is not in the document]
dpanici: Yes, as the MagneticField object needs different grids depending on if it is a current potential field (2D grid), a coil (1D grid) or a analytic field e.g. a ToroidalMagneticField (in which case no grid is needed).

A default would be messy to assign given these differences so we defer to the field's default grid

tests/test_objective_funs.py
dpanici: mean curvature is needed because when you use the plasma vessel distance objective the errors are only evaluated point-wise, and no cost is given to just contorting the boundary shape to have the correct distances at the grid points but having a wild curvature or self-intersections in between.

tests/test_objective_funs.py
dpanici: It is not a repeat because the equilibrium is included in the optimization (as in, it is passed in and so this exercises the multiple object optimization framework)

tests/test_objective_funs.py
dpanici: this test is making sure it works with the analytic magnetic field objects (as there was a bug before that we fixed once we wrote this test and realized it did not work). 

We should clarify in the comment that it is testing eq/field optimization with an analytic magnetic field instead of a current potential field.

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+        eq = nonaxisym_eq.copy()
+        field_grid, eval_grid, source_grid = get_grids(eq, field)
+
+        # fix modes where Rb_lmn = 0 so optimizer only pays attention to
+        # nonaxisymmetric modes
+        zero_indices = np.where(eq.Rb_lmn == 0)
+        fixed_R_modes = eq.surface.R_basis.modes[zero_indices]
+
+        zero_indices = np.where(eq.Zb_lmn == 0)
+        fixed_Z_modes = eq.surface.Z_basis.modes[zero_indices]
```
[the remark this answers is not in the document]
dpanici: That is fair, this is maybe a bit too constraining as it also is fixing axisymmetric modes that initially happen to be zero, for the purposes of this test that is fine as we just want to make the problem well-defined so we can easily test the outcome

tests/test_objective_funs.py
```diff
@@ -485,25 +493,284 @@ def test_boundary_error_nestor(self):
...
+            NFP=4,
+        )
+        nonaxisym_eq = Equilibrium(L=4, M=4, N=4, surface=surface)
+        nonaxisym_eq.solve(verbose=0)
+
+        optimizer = Optimizer("lsq-exact")
+        nonlinear_optimizer = Optimizer("proximal-lsq-exact")
+
+        # test 1 with eq_fixed = True: the optimizer will zero out the nonaxisymmetric
+        # part of the field (Phi_mn) since equilibrium is axisymmetric.
```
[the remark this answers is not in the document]
dpanici: @kianorr  I agree, splitting it into 4 tests will make it clearer when something goes wrong which specific test it is related to

desc/objectives/_free_boundary.py
```diff
@@ -589,74 +590,326 @@ def compute(self, eq_params, sheet_params=None, constants=None):
...
+                eval_data["R"],
+                eval_data["zeta"],
+                eval_data["Z"],
+            ]
+        ).T
+
+        # can't pre-compute Bext because it is dependent on eval_grid
+        Bext = self._ext_field.compute_magnetic_field(
+            x, grid=self._field_grid, basis="rpz", params=field_params
+        )
```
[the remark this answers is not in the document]
dpanici: correct

desc/objectives/_free_boundary.py
```diff
@@ -603,74 +604,326 @@ def compute(self, eq_params, sheet_params=None, constants=None):
...
+
+            # don't need extra B/2 since we only care about normal component
+            Bplasma = -singular_integral(
+                eval_data,
+                self._eval_grid,
+                src_data,
+                self._src_grid,
+                "biot_savart",
+                constants["interpolator"],
+            )
```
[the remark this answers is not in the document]
dpanici: surprised the objective runs at all with this forgotten import

tests/test_objective_funs.py
dpanici: and the test is not matching the field to the stellarator, it is intentiaonally seeking the trivial solution of zeroing out the external field strength. Again the test is to test the framework and the optimization of analytic field objects

desc/magnetic_fields.py
```diff
@@ -613,13 +613,20 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
             magnetic field at specified points
 
         """
+        if params is None:
```
dpanici, line 616: you should merge `rc/freeb` into this again so that the changes from  the PR that got merged into master already disappear from this one

desc/objectives/_free_boundary.py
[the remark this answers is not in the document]
kianorr: Actually the import seems to be there so I'm not sure why this popped up

tests/test_objective_funs.py
dpanici: if we remove those constraints, then the minimum Bnormal will be just making `I` and `G` and `Phi_mn` equal to zero, which is not what we are trying to test here

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +755,338 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+        integration on the local grid
+    src_grid : Grid, optional
+        Collocation grid containing the nodes for plasma source terms.
+    eval_grid : Grid, optional
+        Collocation grid containing the nodes on the plasma surface at which the
+        magnetic field is being calculated and where to evaluate Bn errors.
+    field_grid : Grid, optional
+        Grid used to discretize ext_field (e.g. grid for the magnetic field source from
+        coils).
+    eq_fixed : bool
```
f0uriest, line 797: I think for this objective we should always have the equilibrium fixed. If the equilibrium is allowed to change the regular BoundaryError objective will be more accurate since it can include the sheet current term.

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +755,338 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+        elif self._field_fixed:
+            eq = self.things[0]
+        else:
+            eq = self.things[1]
+
+        if eq != self._eq:
+            self._eq = eq
+
+        if self._src_grid is None:
+            src_grid = LinearGrid(
```
f0uriest, line 886: if the equilibrium is always fixed we can use a higher resolution grid here to ensure the plasma field is computed accurately.

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +755,338 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+            normalize_target=normalize_target,
+            name=name,
+        )
+
+    def build(self, eq=None, use_jit=True, verbose=1):
+        """Build constant arrays.
+
+        Parameters
+        ----------
+        eq : Equilibrium, optional
```
f0uriest, line 867: eq shouldn't be passed in here

desc/objectives/_free_boundary.py
```diff
@@ -589,74 +590,326 @@ def compute(self, eq_params, sheet_params=None, constants=None):
...
             )
+            self._eval_grid = eval_grid
+        else:
+            eval_grid = self._eval_grid
+        if self._s is None:
+            k = min(src_grid.num_theta, src_grid.num_zeta)
+            self._s = k // 2 + int(np.sqrt(k))
+        if self._q is None:
+            k = min(src_grid.num_theta, src_grid.num_zeta)
+            self._q = k // 2 + int(np.sqrt(k))
```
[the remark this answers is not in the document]
f0uriest: see the BoundaryError objective, I updated the default values

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +755,338 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+            src_data = compute_fun(
+                "desc.equilibrium.equilibrium.Equilibrium",
+                self._data_keys,
+                params=eq.params_dict,
+                transforms=src_transforms,
+                profiles=src_profiles,
+            )
+
+            # don't need extra B/2 since we only care about normal component
+            Bplasma = -singular_integral(
```
f0uriest, line 988: see the BoundaryError objective, I added a convenience wrapper for the biot-savart integral

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +755,338 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+                eval_data["Z"],
+            ]
+        ).T
+
+        # can't pre-compute Bext because it is dependent on eval_grid
+        Bext = self._ext_field.compute_magnetic_field(
+            x, grid=self._field_grid, basis="rpz", params=field_params
+        )
+
+        f = jnp.sum((Bext + Bplasma) * eval_data["n_rho"], axis=-1)
```
f0uriest, line 1086: do we want to weight this by the area element?

desc/objectives/_free_boundary.py
kianorr: Okay so we're getting rid of the `eq_fixed` option completely right? Including the tests where `eq_fixed = False`?

f0uriest: Yeah. If we have any tests we want to keep we can switch them to using `BoundaryError` or `VacuumBoundaryError`

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +754,341 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+        Grid used to discretize ext_field (e.g. grid for the magnetic field source from
+        coils).
+    eq_fixed : bool
+        Whether or not the equilibrium's DOF (params) change during optimization.
+        If true, Bplasma will be precomputed in build().
+    field_fixed : bool
+        Whether or not the external field's DOF (params) change during optimization.
+    vacuum : bool
+        If true, Bplasma is set to zero.
+    loop : bool
```
f0uriest, line 803: if we're only evaluating the integral once I think we can just fix `loop=True` to avoid memory issues

desc/objectives/_free_boundary.py
kianorr: Okay so this means that the field should always be unfixed right? Since the field and eq can't both be fixed?

f0uriest: exactly

desc/objectives/_free_boundary.py
kianorr: I set it to N=M=15

f0uriest: It should still be based on `eq.M_grid` etc, but we can do `3*eq.M_grid` or something

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +754,283 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+            has_axis=False,
+        )
+
+        timer = Timer()
+        if verbose > 0:
+            print("Precomputing transforms")
+        timer.start("Precomputing transforms")
+
+        src_profiles = get_profiles(self._data_keys, obj=eq, grid=src_grid)
+        src_transforms = get_transforms(self._data_keys, obj=eq, grid=src_grid)
```
f0uriest, line 926: all the source stuff (grid, profiles, transforms, interpolator etc) are only needed if `vacuum=False` so we should move it under the conditional so we don't calculate a bunch of stuff that's not needed.

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +754,283 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+        self._constants = {"interpolator": interpolator}
+        self._data_keys = [
+            "R",
+            "zeta",
+            "Z",
+            "n_rho",
+            "K_vc",
+            "phi",
+        ]
+        self._args = get_params(
```
f0uriest, line 914: I don't think this is needed anymore? It's from an old API

desc/objectives/_free_boundary.py
```diff
@@ -589,74 +590,326 @@ def compute(self, eq_params, sheet_params=None, constants=None):
...
+
+        if eq != self._eq:
+            self._eq = eq
+
+        if self._src_grid is None:
+            src_grid = LinearGrid(
+                rho=np.array([1.0]),
+                M=eq.M_grid,
+                N=eq.N_grid,
+                NFP=int(eq.NFP),
```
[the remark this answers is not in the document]
f0uriest: It needs to be an int for the singular integral stuff, and I've had issues (especially when loading old solutions) where it was a float. I think it's fine to cast it as a safeguard

desc/objectives/_free_boundary.py
kianorr: I'm not sure what this means

desc/objectives/_free_boundary.py
```diff
@@ -754,6 +754,258 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+    _scalar = False
+    _linear = False
+    _print_value_fmt = "Boundary normal field Error: {:10.3e} "
+    _units = "(T)"
+    _coordinates = "rtz"
+
+    def __init__(
+        self,
+        ext_field,
+        eq=None,
```
f0uriest, line 813: eq should be required so no default value. Also I think with other objectives the convention has been eq is the first argument, then the field

desc/objectives/_free_boundary.py
f0uriest: basically right now this is an approximation to in the integral 

$$\int B\cdot n d\theta d\zeta$$

which has units of Tesla.

Another option is to have it approximate the integral

$$\int B\cdot n dA = \int B\cdot n |e_\theta \times e_\zeta| d\theta d\zeta$$

with units of T*m^2, where $|e_\theta \times e_\zeta|$ is the area element that accounts for the fact that different chunks of $\theta, \zeta$ cover different areas in real space.

I'm not sure which is more conventional/correct, @dpanici might know more.

tests/test_objective_funs.py
```diff
@@ -569,6 +576,176 @@ def test(eq):
...
+
+        np.testing.assert_allclose(things[0].Phi_mn, 0, atol=1e-8)
+
+    @pytest.mark.unit
+    def test_quadratic_flux_with_analytic_field(self):
+        """Test analytic field optimization when eq_fixed=True, field_fixed=False.
+
+        Checks that B goes to zero for axisymmetric eq and field.
+        """
+        eq = desc.examples.get("precise_QA")
```
f0uriest, line 701: can reduce resolution a lot after loading to save time

`eq.change_resolution(3,3,3,6,6,6)` or something

tests/test_objective_funs.py
```diff
@@ -569,6 +576,176 @@ def test(eq):
...
+
+        # test that torus (axisymmetric) Bnorm is exactly 0
+        __, eq = quadratic_flux_equilibriums
+        obj = QuadraticFlux(t_field, eq)
+        obj.build(eq, verbose=2)
+        f = obj.compute(field_params=t_field.params_dict)
+        np.testing.assert_allclose(f, 0, rtol=1e-15, atol=1e-15)
+
+        # test nonaxisymmetric surface
+        eq = desc.io.load("desc/examples/precise_QA_output.h5")[0]
```
f0uriest, line 638: use `desc.examples.get` here, and see other comment about reducing resolution

desc/objectives/_free_boundary.py
dpanici: I think it should include the area element, yes. I just checked and simsopt defines it that way too.

tests/test_objective_funs.py
kianorr: I didn't use `desc.examples.get` because this equilibrium worked better than the final equilibrium. Now that I think about it, it doesn't make much sense to me why the final equilibrium doesn't pass this test

f0uriest: The "family" of the precise QA/QH are from the multigrid optimization, so each will have a different boundary shape, so the flux will be different.

you get get the first member using `get("precise_QA", "all")[0]` but we might also want to think if this is a good test for it since we don't expect the toroidal field to match well with it. Maybe @dpanici can make a coilset for this that will give low field error?

desc/objectives/_free_boundary.py
kianorr: Okay thank you. So would the code look something like `f = jnp.sum((Bext + Bplasma) * eval_data["n_rho"] * eval_data["S(r)"], axis=-1)`?

kianorr: also do I need to do anything else when comparing with the computed Bnorm? i.e. `field.compute_Bnormal`

tests/test_objective_funs.py
```diff
@@ -570,6 +577,178 @@ def test(eq):
...
+        f = obj.compute(field_params=t_field.params_dict)
+
+        np.testing.assert_allclose(f, Bnorm, atol=1e-3)
+
+    @pytest.mark.unit
+    def test_quadratic_flux_with_eq_fixed(self, quadratic_flux_equilibriums):
+        """Test with eq_fixed = True, field_fixed = False.
+
+        Checks that a non-axisymmetric field becomes axisymmetric
+        given an axisymmetric field.
```
ddudt, line 663: This says is uses `eq_fixed=True` but it doesn't actually. 

Also the second sentence of the description doesn't make sense.

desc/objectives/_free_boundary.py
```diff
@@ -757,6 +757,260 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+    _scalar = False
+    _linear = False
+    _print_value_fmt = "Boundary normal field Error: {:10.3e} "
+    _units = "(T)"
+    _coordinates = "rtz"
+
+    def __init__(
+        self,
+        eq,
+        ext_field,
```
ddudt, line 816: Can we rename this to `field` instead of `ext_field` to match the free-boundary objectives?

desc/objectives/_free_boundary.py
```diff
@@ -757,6 +757,260 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+            profiles=eval_profiles,
+        )
+
+        self._constants = {}
+
+        # pre-compute Bplasma because we are assuming eq is fixed
+        if self._vacuum:
+            Bplasma = np.zeros(np.shape(eval_grid.nodes))
+
+        else:
```
f0uriest, line 911: see wrapper added in #950 for doing this, can replace most of this block

desc/objectives/_free_boundary.py
f0uriest: it should be the local area element `"|e_theta x e_zeta|"`:

```python
f = jnp.sum((Bext + Bplasma) * eval_data["n_rho"], axis=-1) * eval_data["|e_theta x e_zeta|"]

File: tests/conftest.py [hunk 40]
```diff
@@ -325,3 +326,22 @@ def VMEC_save(SOLOVEV, tmpdir_factory):
...
+    """Create equilibriums for quadratic flux tests."""
+    surface = FourierRZToroidalSurface(
+        R_lmn=[1, 0.125, 0.1],
+        Z_lmn=[-0.125, -0.1],
+        modes_R=[[0, 0], [1, 0], [0, 1]],
+        modes_Z=[[-1, 0], [0, -1]],
+        NFP=4,
+    )
+    nonaxisym_eq_with_surface = Equilibrium(L=4, M=4, N=4, surface=surface)
+    nonaxisym_eq_with_surface.solve(verbose=0)
```
f0uriest, line 342: would be better to compute these once and store them in `tests/inputs` rather than recomputing on each test run

tests/test_objective_funs.py
kianorr: was there a resolution to this?

f0uriest:
@kianorr @dpanici @ddudt Right now there are 4 tests for the quadratic flux objective, do we need all of them? I would vote for reducing down to 1, combining `test_quadratic_flux` and `test_quadratic_flux_vacuum` and moving the others to regression tests in #956

dpanici:
> @kianorr @dpanici @ddudt Right now there are 4 tests for the quadratic flux objective, do we need all of them? I would vote for reducing down to 1, combining `test_quadratic_flux` and `test_quadratic_flux_vacuum` and moving the others to regression tests in #956

I think yes to combining those two tests (that just test the quad flux calculation),and maybe getting rid of the one that optimized the nonaxisymmetric current potential field. But the one doing optimization of the toroidal fieldsI think is fine to keep, as it is quick and had caught errors in the objective when used in optimization (maybe we can set the vacuum flag to true for it to make it even quicker).

tests/test_objective_funs.py
```diff
@@ -655,6 +660,180 @@ def test(coil, grid=None):
...
+            gtol=1e-14,
+            copy=True,
+        )
+
+        # optimizer should zero out field since that's the easiest way
+        # to get to Bnorm = 0
+        np.testing.assert_allclose(things[0].B0, 0, atol=2e-3)
+
+    @pytest.mark.unit
+    def test_quadratic_flux_vacuum(self):
```
dpanici, line 817: Rory is right, let's combine this test with the first test

tests/test_objective_funs.py
```diff
@@ -655,6 +660,180 @@ def test(coil, grid=None):
...
+            eq.surface, eval_grid=eval_grid, source_grid=source_grid
+        )[0]
+        obj.build(eq)
+        dA = eq.compute("|e_theta x e_zeta|", grid=eval_grid)["|e_theta x e_zeta|"]
+        f = obj.compute(field_params=t_field.params_dict)
+
+        np.testing.assert_allclose(f, Bnorm * dA, atol=1e-3)
+
+    @pytest.mark.unit
+    def test_quadratic_flux_with_eq_fixed(self):
```
dpanici, line 743: we can get rid of this test, as the eq fixed part is basically tested with the following test

dpanici (changes requested):
changes requested in comments

dpanici (dismissed):

desc/objectives/_free_boundary.py
```diff
@@ -780,6 +781,208 @@ def _print(fmt, fmax, fmin, fmean, norm, units):
...
+class QuadraticFlux(_Objective):
+    """Target B*n = 0 on LCFS.
+
+    Uses virtual casing to find plasma component of B and penalizes
+    (B_coil + B_plasma)*n. The equilibrium is kept fixed while the
+    field is unfixed.
+
+    Parameters
+    ----------
+    field : MagneticField
```
dpanici, line 793: order of arguments in docstring does not match that in the actual init

dpanici (approved):

f0uriest (dismissed):

f0uriest (approved):

dpanici (approved):

tests/test_objective_funs.py
```diff
@@ -816,6 +819,64 @@ def test(coil, grid=None):
...
+
+        # test that torus (axisymmetric) Bnorm is exactly 0
+        eq = load("./tests/inputs/vacuum_circular_tokamak.h5")
+        obj = QuadraticFlux(eq, t_field)
+        obj.build(eq, verbose=2)
+        f = obj.compute(field_params=t_field.params_dict)
+        np.testing.assert_allclose(f, 0, rtol=1e-14, atol=1e-14)
+
+        # test non-axisymmetric surface
+        eq = desc.examples.get("precise_QA", "all")[0]
```
ddudt, line 835: Why use the initial equilibrium from this solve and not the "real" precise QA equilibrium?

ddudt (changes requested):
Since the `QuadraticFlux` objective assumes the Equilibrium is held fixed during the optimization, I think we need to add a check for that. Like if `QuadraticFlux` is in `optimizer.objective`, then we should throw an error if `eq` is in `optimizer.things`.

f0uriest:
> Since the `QuadraticFlux` objective assumes the Equilibrium is held fixed during the optimization, I think we need to add a check for that. Like if `QuadraticFlux` is in `optimizer.objective`, then we should throw an error if `eq` is in `optimizer.things`.

This is a good idea, I'll add this.

ddudt (approved):

f0uriest (approved):

PR #732, opened 2023-10-27
Add util for reading efit gfile

f0uriest:
- [ ] needs tests
- [ ] check units and factors of 2pi

### DIFF
--- a/desc/io/efit_io.py
+++ b/desc/io/efit_io.py
@@ -1,0 +1,197 @@
+"""Utilities for reading EFIT geqdsk files."""
+
+import time
+
+import numpy as np
+from scipy.interpolate import CubicSpline
+
+from desc.basis import DoubleFourierSeries
+from desc.equilibrium import Equilibrium
+from desc.geometry import FourierRZToroidalSurface
+from desc.grid import LinearGrid
+from desc.profiles import SplineProfile
+from desc.transform import Transform
+
+
+def efit_to_desc(g, M, N=0, L=None):
+    """Create a DESC equilibrium from an EFIT geqdsk file.
+
+    The equilibrium will not be solved, this just creates an equilibirum
+    with the correct boundary and profiles.
+
+    Parameters
+    ----------
+    g : path-like or dict
+        Path to geqdsk file, or dict of already read parameters.
+    M : int
+        Poloidal resolution of the equilibrium. Used for fitting the boundary.
+    N : int, optional
+        Toroidal resolution of the equilibrium.
+    L : int, optional
+        Radial resolution of the equilibrium. Used for fitting profiles.
+        If None, spline profiles will be used. Otherwise they are fit
+        to even power series in rho.
+
+    Returns
+    -------
+    eq : Equilibrium
+        DESC equilibrium with boundary shape, pressure, iota, and total toroidal flux
+        from geqdsk file.
+    """
+    if not isinstance(g, dict):
+        g = read_gfile(g)
+    ns = len(g["q"])
+    chiN = np.linspace(0, 1, ns)
+    chi_scale = g["chi_boundary"] - g["chi_axis"]
+    # chi = poloidal flux
+    # psi = toroidal flux
+
+    # iota = dchi / dpsi = dchi / dchiN * dchiN / dpsi = dchiN / dpsi * (chi_scale)
+    # dpsi = 1/iota dchi = q dchiN
+    # psi = integral(q dchiN) * (chi_scale)
+
+    # q * (chi(1) - chi(0)) as function of chiN
+    q_spline = CubicSpline(chiN, g["q"] * chi_scale)
+    # this gives psi(chiN)
+    psi_spline = q_spline.antiderivative()
+
+    psi = psi_spline(chiN)
+    Psi_b = psi[-1]  # boundary flux
+    psiN = psi / Psi_b
+    rho = np.sqrt(psiN)
+
+    pressure = SplineProfile(g["pres"], rho)
+    iota = SplineProfile(1 / g["q"], rho)
+    if L is not None:
+        pressure = pressure.to_powerseries(order=L, sym=True)
+        iota = iota.to_powerseries(order=L, sym=True)
+
+    nbdry = g["nbbbs"]
+    rbdry = g["rbbbs"]
+    zbdry = g["zbbbs"]
+    theta = np.linspace(0, 2 * np.pi, nbdry)  # rbbbs includes endpoints
+
+    grid = LinearGrid(theta=theta)
+    basis = DoubleFourierSeries(M=M, N=N, sym=False)
+    transform = Transform(grid, basis, build_pinv=True)
+    Rb_lmn = transform.fit(rbdry)
+    Zb_lmn = transform.fit(zbdry)
+
+    surf = FourierRZToroidalSurface(
+        R_lmn=Rb_lmn,
+        Z_lmn=Zb_lmn,
+        modes_R=basis.modes[:, 1:],
+        modes_Z=basis.modes[:, 1:],
+    )
+
+    eq = Equilibrium(
+        L=L, M=M, N=N, surface=surf, pressure=pressure, iota=iota, Psi=Psi_b
+    )
+    return eq
+
+
+def read_gfile(filename):
+    """Read an EFIT geqdsk file into a dict of ndarray.
+
+    Parameters
+    ----------
+    filename : path-like
+        Path to geqdsk file.
+
+    Returns
+    -------
+    g : dict
+        dict of ndarray containing data from the geqdsk file.
+    """
+    with open(filename) as f:
+        lines = f.readlines()
+    g = {}
+
+    has_spaces = all(
+        [elem == 5 for elem in [len(lines[i].split()) for i in range(1, 5)]]
+    )
+    if has_spaces:
+
+        def splitline(s):
+            return s.split()
+
+        def splitarr(l):
+            return " ".join(l).split()
+
+    else:
+
+        def splitline(s):
+            ss = s.split("\n")[0]
+            idx = np.arange(0, len(ss) + 1, 16)
+            return [ss[i:j] for i, j in zip(idx[:-1], idx[1:])]
+
+        def splitarr(l):
+            return splitline("".join([line.split("\n")[0] for line in l]))
+
+    line0 = lines[0].split()
+    g["date"] = time.strptime(line0[1], "%m/%d/%Y")
+    g["shot"] = int("".join([c for c in line0[2] if c.isdigit()]))
+    g["time"] = int("".join([c for c in line0[3] if c.isdigit()]))
+    g["nw"] = int("".join([c for c in line0[-2] if c.isdigit()]))
+    g["nh"] = int("".join([c for c in line0[-1] if c.isdigit()]))
+    if len(line0) == 8:
+        g["efit"] = line0[4]
+    else:
+        g["efit"] = None
+
+    [g["rdim"], g["zdim"], g["rcentr"], g["rleft"], g["zmid"]] = [
+        float(foo) for foo in splitline(lines[1])
+    ]
+    [g["rmaxis"], g["zmaxis"], g["chi_axis"], g["chi_boundary"], g["bcentr"]] = [
+        float(foo) for foo in splitline(lines[2])
+    ]
+    [g["current"], g["chi_axis"], _, g["rmaxis"], _] = [
+        float(foo) for foo in splitline(lines[3])
+    ]
+    [g["zmaxis"], _, g["chi_boundary"], _, _] = [
+        float(foo) for foo in splitline(lines[4])
+    ]
+    lines_per_profile = int(np.ceil(g["nw"] / 5))
+    lines_psirz = int(np.ceil(g["nw"] * g["nh"] / 5))
+    lines = lines[5:]
+    g["fpol"] = np.array([float(foo) for foo in splitarr(lines[:lines_per_profile])])
+    lines = lines[lines_per_profile:]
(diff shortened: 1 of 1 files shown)


f0uriest:
> Currently, the boundary coefficients from the efit_to_desc function and the ones from SangKyeun's file do not match because his coefficients are not from the last surface. @f0uriest I think it would be better to take SangKyeun's contour extractor instead of yours.

I just used the LCFS that efit identifies, it's fine to change it but we should make it an option where to truncate (and also may need to modify profiles accordingly)

tests/desc_from_eqdsk_cocos_sep_dev
ddudt, line 1: this should be moved to the directory `tests/inputs/`

tests/desc_from_eqdsk_cocos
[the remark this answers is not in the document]
ddudt: or deleted, since it is redundant with the other file

ddudt (changes requested):

ddudt:
We should clean up the code organization here. 

If this PR is meant to convert an EFIT input file into a DESC input file, then it should be a method of the `InputReader` class, like [vmec_to_desc_input](https://github.com/PlasmaControl/DESC/blob/master/desc/input_reader.py#L932). 

If this PR is a larger collection of I/O utilities for EFIT files, then it should be a new class called `EFITIO` modeled after [VMECIO](https://github.com/PlasmaControl/DESC/blob/master/desc/vmec.py#L36).

desc/io/efit_io.py
```diff
@@ -0,0 +1,234 @@
...
+    psi_spline = q_spline.antiderivative()
+
+    psi = psi_spline(chiN)
+    Psi_b = psi[-1]  # boundary flux
+    psiN = psi / Psi_b
+    rho = np.sqrt(psiN)
+
+    pressure = SplineProfile(g["pres"], rho)
+    iota = SplineProfile(1 / g["q"], rho)
+    if L is not None:
```
dpanici, line 68: I think that having a non-None `L` is not the same as asking for the pressure and iota to be power series, if using a spline profile is a better fit of the efit data then it should still be a SplineProfile. So I'd prefer to not have this if statement at all, if this funciton is just making an `Equilibrium` object from an EFIT. 

if the purpose of this is to later write an input file, put this logic of converting to a power series into the `Equilibrium` to input file converter, not this function. This function I'd prefer to just return an `Equilibrium` object with the boundary desired and the profiles from the eqdsk file

desc/io/efit_io.py
```diff
@@ -0,0 +1,234 @@
...
+        Path to geqdsk file, or dict of already read parameters.
+    M : int
+        Poloidal resolution of the equilibrium. Used for fitting the boundary.
+    N : int, optional
+        Toroidal resolution of the equilibrium.
+    L : int, optional
+        Radial resolution of the equilibrium. Used for fitting profiles.
+        If None, spline profiles will be used. Otherwise they are fit
+        to even power series in rho.
+    sep_dev : float
```
dpanici, line 26: sep_dev is not very intuitive, is this a usual term in the tokamak community? 

I think something more like `fraction_of_bdry` or something better where it defualts to `1.0` if the separatrix is desired is better, especially since you can ask for `0.95` and get the `q95` surface which is (I think) used a lot in cases where a hard separatrix is not desired.

desc/io/efit_io.py
```diff
@@ -0,0 +1,234 @@
...
+        Deviation from separatrix. If sep_dev == 0, then the boundary
+         is read from the file.sep_dev = 1 corresponds to the magnetic axis.
```
dpanici, line 28: `file.sep_dev` sounds like a file structure, putting a space between would make this sentence clearer

desc/io/efit_io.py
```diff
@@ -0,0 +1,234 @@
...
+    # this gives psi(chiN)
+    psi_spline = q_spline.antiderivative()
+
+    psi = psi_spline(chiN)
+    Psi_b = psi[-1]  # boundary flux
+    psiN = psi / Psi_b
+    rho = np.sqrt(psiN)
+
+    pressure = SplineProfile(g["pres"], rho)
+    iota = SplineProfile(1 / g["q"], rho)
```
dpanici, line 67: is `1/q` the same as iotabar = $/iota/ 2\pi$ or is it just $/iota$?

dpanici: we use iotabar

tests/test_input_output.py
```diff
@@ -580,3 +580,121 @@ def test_io_SplineMagneticField(tmpdir_factory):
     for key in derivs1.keys():
         for key2 in derivs1[key].keys():
             np.testing.assert_allclose(derivs1[key][key2], derivs2[key][key2])
+
+
```
dpanici, line 584: adding a test where you compare the DESC Equilibrium quantities to quantities from the gfile directly would be good, maybe for some simple case (can't the SOLOVEV be ran with EFIT? DESC compares very well with that case)

desc/io/efit_io.py
```diff
@@ -0,0 +1,234 @@
+"""Utilities for reading EFIT geqdsk files."""
+
+import numpy as np
+from matplotlib import pyplot as plt
+from scipy.interpolate import CubicSpline
+
+
+def efit_to_desc(g, M, N=0, L=None, sep_dev=0):
```
dpanici, line 8: I would prefer if there were an option to specify current or iota profile, if not then specify in docstring it willl return a Fixed-iota equiibrium

dpanici (changes requested):
Changes in comments, and I second the organizational comment from daniel

desc/io/efit_io.py
rahulgaur104: It should just be iota.

tests/test_input_output.py
rahulgaur104: I don't have access to any Grad-Shafranov solvers currently, including the ones in EFIT.

desc/io/efit_io.py
rahulgaur104: Right now it's just fixed iota because the eqdsk doesn't explicitly contain the current values. What we can do in the long run is to calculate the current numerically by using the p' and F' values.

desc/io/efit_io.py
rahulgaur104: Ok, I'll do that.

desc/efit.py
```diff
@@ -0,0 +1,374 @@
...
+        M : int
+            Poloidal resolution of the equilibrium. Used for fitting the boundary.
+        N : int, optional
+            Toroidal resolution of the equilibrium.
+        L : int, optional
+            Radial resolution of the equilibrium. Used for fitting profiles.
+            If None, spline profiles will be used. Otherwise they are fit
+            to even power series in rho.
+        bdry_dist : float
+            Deviation from separatrix. If bdry_dist == 1, then the boundary
```
dpanici, line 45: is this related to like $\rho$ or $\psi_n$?

desc/efit.py
```diff
@@ -0,0 +1,374 @@
...
+                def splitarr(l):
+                    return splitline("".join([line.split("\n")[0] for line in l]))
+
+            line0 = lines[0].split()
+
+            # points in the horizontal direction (width) and vertical direction (height)
+            g["nw"] = int("".join([c for c in line0[-2] if c.isdigit()]))
+            g["nh"] = int("".join([c for c in line0[-1] if c.isdigit()]))
+            if len(line0) == 8:
+                g["efit"] = line0[4]
```
dpanici, line 88: what is this key? wouuld be good to say what this is (is it the EFIT version?)

desc/efit.py
```diff
@@ -0,0 +1,374 @@
...
+            ).reshape((g["nh"], g["nw"]))
+            lines = lines[lines_psirz:]
+
+            # Safety factor
+            g["q"] = np.array(
+                [float(foo) for foo in splitarr(lines[:lines_per_profile])]
+            )
+            lines = lines[lines_per_profile:]
+
+            if bdry_dist == 1:  # distance of the boundary from the axis
```
dpanici, line 144: what is this distance in? some normalized unit? $\rho$ or $\psi_n$?

desc/efit.py
```diff
@@ -0,0 +1,374 @@
...
+        outfile,
+        objective="force",
+        optimizer="lsq-exact",
+        ftol=1e-2,
+        xtol=1e-2,
+        gtol=1e-6,
+        maxiter=100,
+        threshold=1e-8,
+    ):
+        """Generate a DESC input file from a DESC equilibrium object.
```
dpanici, line 274: this docstring sohuld mention efit?? what is the purpose of the function?

desc/efit.py
```diff
@@ -0,0 +1,374 @@
...
+
+        DESC will automatically choose continuation parameters
+
+        Parameters
```
dpanici, line 278: some of these parameters are not actually in the init, like infile. And `eq` is not in the parameters but isin the init.

desc/efit.py
```diff
@@ -0,0 +1,374 @@
...
+    @classmethod
+    def load(  # noqa: C901 - fxn too complex
+        cls,
+        g,
+        M,
+        N=0,
+        L=None,
+        bdry_dist=1,
+    ):
+        """Create a DESC equilibrium object from an EFIT geqdsk file.
```
dpanici, line 27: probably a stupid quesiton, but in the inputs you have a `eqdsk_cocos.out` file, is that the same as a `geqdsk` file?

dpanici (changes requested):
Instead of making a separate `efit_to_desc_input` fxn, use your existing function (perhaps modified so you can specify the string printed at the top of the file)

desc/efit.py
rahulgaur104: It's normalized poloidal flux. Adding that to the docstring.

desc/efit.py
rahulgaur104: It's the efit file information. I have added a comment. It doesn't affect the calculation in any way.

desc/efit.py
rahulgaur104: It's normalized poloidal flux. Adding that to the docstring.

desc/efit.py
rahulgaur104: You are right. This function is more general purpose than an input file writer for EFIT files. Right now, I don't know how to move it into another script.

desc/efit.py
rahulgaur104: Yes, it's an eqdsk file. The data may have been preprocesses but the format is exactly the same as an eqdsk file.

desc/efit.py
rahulgaur104: Ok, I have made changes to the docstring so that it is the same as the inputs to the function.

desc/efit.py
```diff
@@ -0,0 +1,373 @@
...
+            relative tolerance of the state vector x
+        gtol : float
+            absolute tolerance of the projected gradient g
+        maxiter : int
+            maximum number of optimizer iterations per continuation step
+        threshold : float
+            Fourier coefficients below this value will be set to 0.
+        """
+        with open(outfile, "w+") as f:
+            header = "# DESC input file generated from an geqdsk file\n# "
```
dpanici, line 303: would be better to allow header to be passed into this file, have the default be "generated from a DESC equilibrium on XXX date with DESC version COMMIT_TAG"

as the fxn does not need efit it should not be mentioned

tests/test_efit.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    efit_file_path = "./tests/inputs/eqdsk_cocos.out"
+
+    tmpdir = tmpdir_factory.mktemp("efit_to_desc_inputs")
+    tmp_path1 = tmpdir.join("desc_from_eqdsk_true")
+    tmp_path2 = tmpdir.join("desc_from_eqdsk_bdry_dist_true")
+
+    with pytest.warns(UserWarning):
+        eq = EFITIO.load(str(efit_file_path), M=20)
+        ir1 = EFITIO.efit_to_desc_input(eq, tmp_path1)
```
dpanici, line 26: change this once the method is put as a separate fxn in a separate file

dpanici (changes requested):
- Move the `efit_to_desc_input` to a separate file (maybe somewhere in `io`) and also change the name to just `write_desc_input` as the fxn itself does not care about efit right? instead of move, just use the existing function if you have one already

desc/efit.py
```diff
@@ -0,0 +1,373 @@
...
+            f.write("\n# pressure and rotational transform/current profiles\n")
+
+            grid = LinearGrid(L=eq.L_grid, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
+            rho = grid.nodes[grid._unique_rho_idx, 0]
+
+            pressure = grid.compress(eq.compute("p", grid=grid)["p"])
+            iota = grid.compress(eq.compute("iota", grid=grid)["iota"])
+
+            pres_profile = PowerSeriesProfile.from_values(
+                rho, pressure, order=eq.L, sym=False
```
dpanici, line 342: we probably want to enforce sym=True? since physically it should be an even function of rho? in the input file you have, the coeffs for the odd parts are super small so maybe it does not matter but things will throw warnings with sym=False

desc/efit.py
```diff
@@ -0,0 +1,373 @@
...
+            ).params
+            iota_profile = PowerSeriesProfile.from_values(
+                rho, iota, order=eq.L, sym=False
+            ).params
+
```
dpanici, line 347: do we not want an option for a current profile instead of an iota profile?

dpanici:
Would it be better to instead use an external package like https://github.com/freegs-plasma/FreeQDSK?

Instead of us writing our own which invariably has less features

dpanici:
Only do the `iota` profile and if `current` is specified, point to the other repo you have that can do this

PR #733, opened 2023-11-03
Fixes to allow linear objectives to work as bounds

f0uriest:
Currently all linear constraints are passed to `factorize_linear_constraints`, even those with bounds instead of targets, which causes things to fail when using linear inequality constraints. This fixes how linear constraints are parsed and also how linear constraints set bounds/targets to allow linear inequality constraints.


desc/objectives/linear_objectives.py
```diff
@@ -38,9 +38,28 @@ def update_target(self, eq):
...
+        self, target_from_user, default_target, default_bounds, idx
+    ):
+        if target_from_user is None:
+            target = default_target
+            bounds = default_bounds
+        elif isinstance(target_from_user, tuple) and (
+            len(target_from_user) == 2
+        ):  # treat as bounds
+            target = None
+            bounds = (
```
kianorr, line 51: Why do you broadcast bounds to a different shape? I guess it's needed for somewhere else?

desc/optimize/optimizer.py
```diff
@@ -345,11 +345,15 @@ def _parse_constraints(constraints):
...
+    # factorized like linear equality constraints
     linear_constraints = tuple(
-        constraint for constraint in constraints if constraint.linear
+        constraint
+        for constraint in constraints
+        if (constraint.linear and (constraint.bounds is None))
     )
     nonlinear_constraints = tuple(
-        constraint for constraint in constraints if not constraint.linear
+        constraint for constraint in constraints if constraint not in linear_constraints
```
kianorr, line 356: this doesn't need (constraint.bounds is not None)? I guess that would be redundant?

desc/objectives/linear_objectives.py
```diff
@@ -38,9 +38,28 @@ def update_target(self, eq):
         if self._use_jit:
             self.jit()
 
+    def _parse_target_from_user(
+        self, target_from_user, default_target, default_bounds, idx
```
kianorr, line 42: what is `idx`? It comes from below I see but I'm not sure what it is

desc/objectives/linear_objectives.py
f0uriest: its because bounds is a tuple of two arrays (lower, upper), rather than a single array. Each bound should have the same shape as target, which is the same as dim_f

desc/objectives/linear_objectives.py
f0uriest: its the indices of which modes/etc are being constrained, in the case that not all of them are. Generally all the modes are constrained, so idx is basically just `np.arange(target.size)`

desc/optimize/optimizer.py
f0uriest: yeah we already get all the non-bounded constraints into `linear_constraints` and this just gets whatever is leftover

desc/objectives/linear_objectives.py
```diff
@@ -148,14 +167,24 @@ def build(self, use_jit=False, verbose=1):
...
+        }
+        default_bounds = None
+        target, bounds = self._parse_target_from_user(
+            self._target_from_user, default_target, default_bounds, indices
+        )
+        if target:
+            self.target = jnp.concatenate([target[par] for par in params])
+            self.bounds = None
+        else:
+            self.target = None
```
dpanici, line 183: test for the use of this with bounds?

dpanici (changes requested):
add test for FixParameter with bounds and I'm happy

kianorr (approved):

dpanici (approved):

ddudt (approved):

PR #734, opened 2023-11-03
Rc/coil tracing

f0uriest:


YigitElma:
If you want to get rid of some PRs, I can take this one. It looks interesting and I am bored:D

f0uriest:
> If you want to get rid of some PRs, I can take this one. It looks interesting and I am bored:D

Sure, main thing it needs is some tests, so probably just need to come up with some simple magnetic fields for which we know the trajectory analytically and make sure they agree. Some ideas:
- Uniform B (using `VerticalField`) to test basic motion (guiding center should move in straight line, full orbit should be a helix)
- B with gradient (`MagneticFieldFromUser` with B = B0*x or something, to test grad B drift
- Purely toroidal field to test gradB and curvature drift

rahulgaur104:
Do you have a symplectic orbit solver to compare with? I have one that I can add to the test.
It calculates the the exact particle position instead of just guiding center. We only have to run it once, gyroaverage and save the data.

f0uriest:
> Do you have a symplectic orbit solver to compare with? I have one that I can add to the test. It calculates the the exact particle position instead of just guiding center. We only have to run it once, gyroaverage and save the data.

Yeah I benchmarked against SIMPLE with rogerio, I'll try to dig up the data

dpanici:
Close once #1820  is merged in

PR #735, opened 2023-11-04
Remove unneeded args from objective getters

f0uriest:
- When using stuff like ``get_fixed_boundary_constraints`` we have to specify both whether to fix profiles, and also which profiles the equilibrium has.
- Now that ``eq`` is a required argument to the get_* functions, this now redundant, since we can just look and see what profiles the equilibrium has.
- This removes the unneeded args from all the get_* methods, so that the only things needed are the equilibrium and a bool for profiles, the specific profiles will be determined from the equilibrium.

### DIFF
--- a/desc/objectives/getters.py
+++ b/desc/objectives/getters.py
@@ -87,9 +87,9 @@ def get_fixed_axis_constraints(eq, profiles=True, normalize=True):
     Parameters
     ----------
     eq : Equilibrium
-        Equilibrium being constrained.
+        Equilibrium to constrain.
     profiles : bool
-        Whether to also return constraints to fix input profiles.
+        If True, also include constraints to fix all profiles assigned to equilibrium.
     normalize : bool
         Whether to apply constraints in normalized units.
 
@@ -120,9 +120,9 @@ def get_fixed_boundary_constraints(eq, profiles=True, normalize=True):
     Parameters
     ----------
     eq : Equilibrium
-        Equilibrium to constraint.
+        Equilibrium to constrain.
     profiles : bool
-        Whether to also return constraints to fix input profiles.
+        If True, also include constraints to fix all profiles assigned to equilibrium.
     normalize : bool
         Whether to apply constraints in normalized units.
 
@@ -168,7 +168,7 @@ def get_NAE_constraints(
     order : int
         order (in rho) of near-axis behavior to constrain
     profiles : bool
-        Whether to also return constraints to fix input profiles.
+        If True, also include constraints to fix all profiles assigned to equilibrium.
     normalize : bool
         Whether to apply constraints in normalized units.
     N : int
--- a/desc/objectives/getters.py
+++ b/desc/objectives/getters.py
@@ -30,6 +30,17 @@ from .linear_objectives import (
 from .nae_utils import calc_zeroth_order_lambda, make_RZ_cons_1st_order
 from .objective_funs import ObjectiveFunction
 
+_PROFILE_CONSTRAINTS = {
+    "pressure": FixPressure,
+    "iota": FixIota,
+    "current": FixCurrent,
+    "electron_density": FixElectronDensity,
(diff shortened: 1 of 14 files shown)


tests/test_compute_funs.py
```diff
@@ -1292,7 +1292,8 @@ def test_compute_everything():
                     np.testing.assert_allclose(
                         actual=this_branch_data[p][name],
                         desired=master_data[p][name],
-                        atol=1e-12,
+                        atol=1e-10,
```
f0uriest, line 1295: This is needed due to minor differences caused by CPU types. Previously pretty much all the gh runners used intel xeon cpus, but recently they've upgraded some to the newer AMD Ryzen ones. This seems to lead to tiny differences in some computed quantities. For most things the difference is on the order of machine precision so no problem, but for ``current_r`` at the axis the difference is ~1e-11. The difference in ``I_r`` is ~1e-17, but when we scale by mu0 we get larger differences for ``current_r``

tests/test_examples.py
```diff
@@ -548,12 +548,12 @@ def test_NAE_QSC_solve():
 
     # this has all the constraints we need,
     #  iota=False specifies we want to fix current instead of iota
```
dpanici, line 550: comment does not correspond to code now

tests/test_examples.py
```diff
@@ -636,7 +636,7 @@ def test_NAE_QIC_solve():
...
```
dpanici, line 638: comment does not correspond to code now

dpanici (changes requested):
2 minor comments
 - add to docstring of getter that it will use whatever profiles are present in the eq
 - a couple of comments in the code refer to iota=False when it is no longer present

ddudt (approved):
Looks good, besides the comments @dpanici already made.

desc/objectives/getters.py
```diff
@@ -102,58 +105,24 @@ def get_fixed_axis_constraints(
...
-                ),
-                FixAtomicNumber(eq=eq, normalize=normalize, normalize_target=normalize),
-            )
-        else:
-            constraints += (
-                FixPressure(eq=eq, normalize=normalize, normalize_target=normalize),
-            )
-            if anisotropy:
+        for name, con in _PROFILE_CONSTRAINTS.items():
+            if getattr(eq, name) is not None:
```
rahulgaur104, line 109: So if the attribute is not in the dictionary, would it be free?

rahulgaur104 (approved):

desc/objectives/getters.py
```diff
@@ -70,23 +81,15 @@ def get_equilibrium_objective(eq, mode="force", normalize=True):
...
+        Equilibrium to constrain.
     profiles : bool
-        Whether to also return constraints to fix input profiles.
-    iota : bool
-        Whether to add FixIota or FixCurrent as a constraint.
-    kinetic : bool
-        Whether to add constraints to fix kinetic profiles or pressure
-    anisotropy : bool
-        Whether to add constraint to fix anisotropic pressure.
+        If True, also include constraints to fix all profiles assigned to equilibrium.
```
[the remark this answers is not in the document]
dpanici: I think you should just create the constraints yourself then for something like that, this getter is mainly for the typical use case of fixing profiles, Psi, and boundary

desc/objectives/getters.py
f0uriest: the dictionary covers all profiles types, so it should fix anything that the equilibrium has

dpanici: Not free, if an equilibrium does not have that profile assigned, then it does not have that attribute at all

i.e. the `FixIota` constraint will not be given to an equilibrium which does not have an `eq.iota` attribute, but if it lacks that attribute that means that instead it has a `eq.current` attribute instead and that one will be fixed

same for kinetic profiles(n and T) and pressure, if the former are present then the latter must not be present, and vice-versa

PR #736, opened 2023-11-04
Add objective for constructing "good" coordinates, add refinement step to initial guess

f0uriest:
Implements the coordinate mapping thing from Zeno Tecchiolli for finding a good (ie non-degenerate) initial guess for strongly shaped equilibria

- Adds a new objective for minimizing the cost function from Zeno's work
- Adds a step in `set_initial_guess` to run a short optimization with that objective if the default initial guess isn't nested.


desc/objectives/_geometry.py
```diff
@@ -1019,3 +1019,148 @@ def compute(self, params, constants=None):
...
+    1/ρ² |√g|² + ω |𝐞ᵨ|²
+
+    where √g is the jacobian of the coordinate system and 𝐞ᵨ is the covariant radial
+    basis vector.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    omega : float
```
dpanici, line 1038: Is this in their paper/presentation? or could we change this name? as `omega` is already the name of the (currently unused) toroidal stream function, so this could get confusing. Although if we want to stick to their convention then it is probably fine

f0uriest: it was used in slides he showed at spectaculars but i don't think its that important, we can change it

dpanici (dismissed):

dpanici (approved):

PR #739, opened 2023-11-07
fix inconsistency with the equations on the paper, code and website

YigitElma:
The following equation on the general theory page is not consistent with the one on the paper,

image

Maybe in SFL, they might be equivalent, but I am not sure about the Jacobian definition.

image

### DIFF
--- a/docs/theory_general.rst
+++ b/docs/theory_general.rst
@@ -120,8 +120,8 @@ When written in flux coordinates there are only two independent components:
   \begin{aligned}
     \mathbf{F} &= F_\rho \mathbf{e}^\rho + F_\beta \mathbf{\beta}_{DESC} \\[.3cm]
     F_\rho &=  -\frac{\partial p}{\partial \rho} + \sqrt{g} (J^\theta B^\zeta - J^\zeta B^\theta)\\[.3cm]
-    F_\beta &=  J^\rho \\[.3cm]
-    \mathbf{\beta_{DESC}} &=  \sqrt{g} (B^\zeta\mathbf{e}^\theta - B^\theta\mathbf{e}^\zeta)
+    F_\beta &=  \sqrt{g} J^\rho \\[.3cm]
+    \mathbf{\beta_{DESC}} &=  B^\zeta\mathbf{e}^\theta - B^\theta\mathbf{e}^\zeta
   \end{aligned}
 
 where :math:`\mathbf{e}^u = \nabla u` is the contravariant basis vector of parameter u. These forces in both the radial and helical directions must vanish in equilibrium.
--- a/docs/theory_general.rst
+++ b/docs/theory_general.rst
@@ -120,8 +120,8 @@ When written in flux coordinates there are only two independent components:
   \begin{aligned}
     \mathbf{F} &= F_\rho \mathbf{e}^\rho + F_\beta \mathbf{\beta}_{DESC} \\[.3cm]
     F_\rho &=  \frac{\partial p}{\partial \rho} + \sqrt{g} (J^\zeta B^\theta - J^\theta B^\zeta)\\[.3cm]
-    F_\beta &=  \sqrt{g} J^\rho \\[.3cm]
-    \mathbf{\beta_{DESC}} &=  (B^\zeta\mathbf{e}^\theta - B^\theta\mathbf{e}^\zeta)
+    F_\beta &=  J^\rho \\[.3cm]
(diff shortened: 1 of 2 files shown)


ddudt:
The equations in my 2020 paper are outdated, but we should update the documentation if it no longer matches the code. It looks like they do need to be updated now that we are not using SFL coordinates in the code. 

Please keep the sign convention the same, where $F \equiv J \times B - \nabla p$. In your notes it looks like you are using $F \equiv -J \times B + \nabla p$

YigitElma:
I was looking at DESC code suite part 1.
image

But ok, you are correct in the code, the force error is calculated in the other way.
image

I will change it. Thanks for pointing it out!

YigitElma:
Btw, 
image
This section says that F_helical has jacobian in it, but in the actual calculation of F_helical, there is no jacobian.
image
In fact, there is no jacobian in the e^helical either,
image
This part has Jacobian but I assume it is for dV volume thing
image

Probably, I miss something. If you want to change the place of jacobian in this PR, let me know please.

dpanici:
You are looking in my paper Appendix which uses the VMEC notation. in DESC, we define beta without the sqrt(g) (see section 3):
image

dpanici (changes requested):
change to use the notation of section 3 of DESC suite part I (maybe need to change a minus sign to use whatever convention we have for F in the code, I don't remember if my paper is consistent with what we use in the code currently as far as F = JxB - grad(p) or F = grad(p) - JxB

YigitElma:
@dpanici In the code,  F = JxB - grad(p) is used. I fixed the Jacobian thing.

docs/theory_general.rst
```diff
@@ -96,27 +108,34 @@ The ideal magnetohydrodynamic equilibrium force balance is defined as
...
 .. math::
   \begin{aligned}
-    \mathbf{F} &= F_\rho \nabla \rho + F_\beta \mathbf{\beta} \\[.2cm]
-    F_\rho &= \sqrt{g} (B^\zeta J^\theta - B^\theta J^\zeta) - \partial_\rho p \\[.2cm]
-    F_\beta &= \sqrt{g} B^\zeta J^\rho \\[.2cm]
-    \mathbf{\beta} &= \nabla \theta - \iota \nabla \zeta
+    \mathbf{F} &= F_\rho \mathbf{e}^\rho + F_{helical} \mathbf{e}^{helical} \\[.3cm]
+    F_\rho &= \sqrt{g} (J^\theta B^\zeta - J^\zeta B^\theta) - \frac{\partial p}{\partial \rho} \\[.3cm]
+    F_{helical} &= \sqrt{g} J^\rho \\[.3cm]
+    \mathbf{e}^{helical} &= B^\zeta\mathbf{e}^\theta - B^\theta\mathbf{e}^\zeta
```
ddudt, line 124: I changed our notation from "beta" to "helical" to match the code. There is a note later on that explains this difference from the original paper.

docs/theory_general.rst
```diff
@@ -96,27 +108,34 @@ The ideal magnetohydrodynamic equilibrium force balance is defined as
...
 When written in flux coordinates there are only two independent components:
 
 .. math::
   \begin{aligned}
-    \mathbf{F} &= F_\rho \nabla \rho + F_\beta \mathbf{\beta} \\[.2cm]
-    F_\rho &= \sqrt{g} (B^\zeta J^\theta - B^\theta J^\zeta) - \partial_\rho p \\[.2cm]
-    F_\beta &= \sqrt{g} B^\zeta J^\rho \\[.2cm]
-    \mathbf{\beta} &= \nabla \theta - \iota \nabla \zeta
+    \mathbf{F} &= F_\rho \mathbf{e}^\rho + F_{helical} \mathbf{e}^{helical} \\[.3cm]
+    F_\rho &= \sqrt{g} (J^\theta B^\zeta - J^\zeta B^\theta) - \frac{\partial p}{\partial \rho} \\[.3cm]
```
ddudt, line 122: Equation (3.26a) in @dpanici's paper has $(J^\zeta B^\theta - J^\theta B^\zeta)$, but I think that is wrong and this equation is correct.

ddudt (commented):
I made a few edits and I think all of the equations are consistent now.

dpanici:
very confused as to how the unit test compute_everything is failing when this only changed docs...

YigitElma:
I didn't understand either.

f0uriest:
> very confused as to how the unit test compute_everything is failing when this only changed docs...

I think its a hardware issue, I fixed it here: https://github.com/PlasmaControl/DESC/pull/735/files#r1382344659

dpanici (approved):

PR #740, opened 2023-11-07
Move interpolate functions to interpax package

f0uriest:
Several people had expressed a need for interpolation routines compatible with JAX so I broke our stuff from ``desc.interpolate`` into its own package ``interpax`` https://github.com/f0uriest/interpax

In the process I also fixed some bugs for 2d and 3d interpolation.

This PR removes the ``desc.interpolate`` module in favor of using ``interpax``.


ddudt (commented):
I don't love the idea of making DESC dependent on an external package that is a personal project. It's not that I don't trust @f0uriest, but it doesn't seem like a smart idea for the longevity of the code if it needs to be updated in the future when Rory might not be around. I would feel more comfortable either keeping all of the interpolation code within the DESC repo, or at least transferring ownership of the `interpax` package to the Plasma Control group for future development.

dpanici (approved):

dpanici (approved):

ddudt (approved):
fine

PR #741, opened 2023-11-08
Compatibility of DESC with Apple Silicon Devices (jax-metal)

rogeriojorge:
This PR introduces the JAX device METAL as available in Apple Silicon devices (M1, M2 or M3) that have the jax-metal installed
https://developer.apple.com/metal/jax/
However, not all functionalities are available yet.

- Only 32-bit functionality is available.
- As jax.scipy.special functions have limited availability, the functions gammaln and logsumexp have special definitions when METAL is the device being used.
- An equilibrium cannot be created yet due to the following error when calling the Equilibrium class:
`jaxlib.xla_extension.XlaRuntimeError: UNKNOWN: DESC/desc/backend.py:116:15: error: failed to legalize operation 'mhlo.scatter'
        return jnp.asarray(arr).at[inds].set(vals)`
It appears that `mhlo.scatter` will be implemented in the next version of jax-metal, as per this thread here
https://developer.apple.com/forums/thread/738697
Therefore, I will wait for the. next update to see if this PR can go through.


YigitElma:
Dropping this here for reference https://github.com/tillahoffmann/jax-mps

PR #742, opened 2023-11-08
Allow Objectives to target Objects other than Equilibrium

dpanici:
Generalize the changes in the #604 PR to allow objectives to target `things` that are not an `eq`, which allows for optimizing objects other than `Equilibrium` objects (e.g. surfaces, coils, MagneticFields)

To optimize one of these other objects, the corresponding objective must be generalized to not expect only an `Equilibrium`. `normalization.py` must also be modified to make normalization scales based off the passed-in object, which may not be an `Equilibrium`

Objectives which could use a FourierRZToroidalSurface or an Equilibrium:

- [x] MeanCurvature
- [x] Volume (same note as above, for "V")
- [x] PrincipalCurvature

For Future PR (as need other physics quantities, not really needed for this  functionality to start working) #795 
-------------------------------------------------------------------------------------------------------------------------
- AspectRatio (must make `"R0/a"` parametrization with FourierRZToroidalSurface  first)
- Elongation (must make "a_major/a_minor" with FourierRZToroidalSurface  first)


dpanici:
As it is now, I changed the `init` keyword arg from `eq` to `thing` only for those objectives which can take more than just an eq for that. I think it would be best to do this for all objectives, and then to maybe use [type hints](https://docs.python.org/3/library/typing.html)

desc/objectives/_geometry.py
```diff
@@ -376,7 +379,12 @@ def build(self, use_jit=True, verbose=1):
         """
         eq = self.things[0]
         if self._grid is None:
-            grid = QuadratureGrid(L=eq.L_grid, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
+            grid = QuadratureGrid(
```
f0uriest, line 382: Do we want a quadrature grid for surface classes? I think quadrature grid with L=1 doesn't have a node at rho=1

dpanici: good catch

desc/objectives/_geometry.py
```diff
@@ -376,7 +379,20 @@ def build(self, use_jit=True, verbose=1):
...
+            if hasattr(eq, "L_grid"):
+                grid = QuadratureGrid(
+                    L=getattr(eq, "L_grid", 1),
+                    M=getattr(eq, "M_grid", eq.M * 2),
+                    N=getattr(eq, "N_grid", eq.N * 2),
+                    NFP=eq.NFP,
+                )
+            else:
+                grid = LinearGrid(
+                    L=1,
```
f0uriest, line 391: might be clearer to put rho=1.0 instead of L=1

f0uriest (approved):

desc/objectives/_geometry.py
```diff
@@ -297,8 +299,9 @@ class Volume(_Objective):
 
     Parameters
     ----------
-    eq : Equilibrium
-        Equilibrium that will be optimized to satisfy the Objective.
+    eq : Equilibrium or FourierRZToroidalSurface
```
ddudt, line 302: Should we generalize the name to `thing`? In theory this should be able to be anything with other toroidal surface parameterizations.

docs/notebooks/tutorials/advanced_optimization.ipynb
```diff
@@ -1222,7 +1222,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.11"
+   "version": "3.8.10"
```
ddudt, line 1225: Is this change necessary?

tests/test_examples.py
```diff
@@ -807,6 +809,72 @@ def test_multiobject_optimization_prox():
...
+
+    obj = PrincipalCurvature(surf, target=1)
+
+    objective = ObjectiveFunction((obj,))
+    optimizer = Optimizer("lsq-exact")
+    (surf), result = optimizer.optimize(
+        (surf), objective, constraints, verbose=3, maxiter=100
+    )
+    surf = surf[0]
+    np.testing.assert_allclose(obj.compute(*obj.xs(surf)), 1, atol=1e-5)
```
ddudt, line 875: This is fine here, but remember that in general it is almost always better to use relative tolerances.

ddudt (commented):
My main comment is we might want to rename `eq` to `thing` to reflect the generalized input parameters.

desc/objectives/_geometry.py
dpanici: talked with @f0uriest and we wanna minimize API breaking changes

docs/notebooks/tutorials/advanced_optimization.ipynb
dpanici: that is just a byproduct of jupyter, this is not a real change

ddudt: So why is it included in this PR? It's fine, but in the future don't add extra files if you don't need to

ddudt (approved):

docs/notebooks/tutorials/advanced_optimization.ipynb
dpanici: will do

PR #744, opened 2023-11-11
fix installation documentation for using cluster GPUs

YigitElma:
Note: This change is already on the "stable" version of the website, but that has older JAX version I guess. This one uses the command from Princeton Jax page.

"Latest" version of the installation page, for DELLA and STELLAR clusters with CPU+GPU support, tells to use 

`module load anaconda3/2023.3
conda create --name desc-env python=3.9
conda activate desc-env
pip install --upgrade "jax[cuda11_pip]<0.4.15" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html`

However, this results in following error, and the simulation only uses CPU,

` Loaded runtime CuDNN library: 8.2.0 but source was compiled with: 8.6.0.  CuDNN library needs to have matching major version and equal or higher minor version. If using a binary install, upgrade your CuDNN library.  If building from sources, make sure the library loaded at runtime is compatible with the version specified during compile configuration.`

The Princeton cluster repo for JAX suggests to use following command, and it works for using GPU,

`$ module load anaconda3/2023.9
$ CONDA_OVERRIDE_CUDA="11.2" conda create --name jax-gpu jax "jaxlib==0.4.14=cuda112*" -c conda-forge`

I added this command in addition to the previous one (because apparently for someone else this doesn't work, but I didn't understand how this can happen since the cluster has the same configuration for every user)

EDIT: The stable version command gave an error for me too. Apparently it installs older JAX and Cuda version. Maybe thats the reason. My change:
`CONDA_OVERRIDE_CUDA="11.2" conda create --name desc-env jax "jaxlib==0.4.14=cuda112*" -c conda-forge`
Stable version commanda:
`CONDA_OVERRIDE_CUDA="11.2" conda create --name desc-env jax "jaxlib==0.4.10=cuda112*" -c conda-forge`

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -123,7 +123,6 @@ specific JAX GPU installation instructions, as that is the main installation dif
 
 Della and Stellar Clusters (Princeton)
 ++++++++++++++++++++++++++++++++++++++
-NOTE: For GPU jobs in Della, you should ssh into <NetID>@della-gpu.princeton.edu.
 
 **Option 1:** First, install JAX (check `this tutorial <https://github.com/PrincetonUniversity/intro_ml_libs/tree/master/jax>`__ ) for the latest version of `jaxlib` available on the Princeton clusters):
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -123,9 +123,9 @@ specific JAX GPU installation instructions, as that is the main installation dif
 
 Della and Stellar Clusters (Princeton)
 ++++++++++++++++++++++++++++++++++++++
-These instructions were tested and confirmed to work on the Della and Stellar clusters at Princeton as of 11-6-2023.
+NOTE: For GPU jobs in Della, you should ssh into <NetID>@della-gpu.princeton.edu.
 
-First, install JAX (check `this tutorial <https://github.com/PrincetonUniversity/intro_ml_libs/tree/master/jax>`__ ) for the latest version of `jaxlib` available on the Princeton clusters):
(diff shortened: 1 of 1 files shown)


dpanici (changes requested):
Can you update the date on which the instructions were last checked by you?

And did you check these for both stellar and della? which one does the pip install work and whicih needs the conda install?

YigitElma:
> Can you update the date on which the instructions were last checked by you?
> 
> And did you check these for both stellar and della? which one does the pip install work and whicih needs the conda install?

@dpanici I added the date. Yes, I checked it both for Stellar and Della. For me, it is working on both clusters (I ssh to della-gpu not della8). Today @kianorr had a similar issue with GPU jobs, but he was using della8 login. I told him to use updated commands on della-gpu. Can you give update here @kianorr ?

kianorr:
@YigitElma the issues I was having were my fault but I haven't tried with pip install since after fixing what I was doing wrong.

docs/installation.rst
```diff
@@ -123,9 +123,9 @@ specific JAX GPU installation instructions, as that is the main installation dif
 
 Della and Stellar Clusters (Princeton)
 ++++++++++++++++++++++++++++++++++++++
-These instructions were tested and confirmed to work on the Della and Stellar clusters at Princeton as of 11-6-2023.
+NOTE: For GPU jobs in Della, you should ssh into <NetID>@della-gpu.princeton.edu.
```
dpanici, line 126: I am able to work just fine on `della8` so I am unsure if this is necessary, I also can use `sbatch` to launch GPU jobs from `della8` without issue

dpanici: is this for if you want to use the GPU on the login node?

YigitElma: Ok, changed it.

dpanici (approved):

PR #746, opened 2023-11-13
add flag to PlasmaVesselDistance to avoid recomputing surface coordinates when the surface is fixed during optimization

dpanici:
- Adds new keyword `surface_fixed` to `PlasmaVesselDistance` which says whether or not the surface comparing the distance from the plasma to is fixed or not. If `True`, then the surface coordinates can be precomputed, saving on computation during optimization. Set to `False` by default.

### DIFF
--- a/desc/objectives/_geometry.py
+++ b/desc/objectives/_geometry.py
@@ -401,6 +401,11 @@ class PlasmaVesselDistance(_Objective):
     will only be an upper bound on the minimum separation between the plasma and the
     surrounding surface.
 
+    NOTE: By default, assumes the surface is fixed and its coordinates are precomputed,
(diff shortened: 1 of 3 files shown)


f0uriest:
I think it would be safer to assume the surface isn't fixed, this is also in line with the rest of the objective stuff which assumes things are free unless explicitly constrained

desc/objectives/_geometry.py
```diff
@@ -401,6 +401,11 @@ class PlasmaVesselDistance(_Objective):
     will only be an upper bound on the minimum separation between the plasma and the
     surrounding surface.
 
+    NOTE: By default, assumes the surface is fixed and its coordinates are precomputed,
```
f0uriest, line 404: should update this

f0uriest (approved):

PR #752, opened 2023-11-15
fix for using PlasmaVesselDistance with ProximalProjection

dpanici:
using `ProximalProjection` with `surface_fixed=True` should be possible, this fixes the bug preventing that


dpanici:
check if things only contains eq

desc/objectives/_geometry.py
```diff
@@ -451,7 +451,7 @@ class PlasmaVesselDistance(_Objective):
         is fixed or not. If True, the surface is fixed and its coordinates are
         precomputed, which saves on computation time during optimization.
         If False, the surface coordinates are computed at every iteration.
-        False by default.
+        False by default, so that self.things = [eq] only, not [eq,surface]
```
f0uriest, line 454: I think this is the opposite?

desc/optimize/optimizer.py
```diff
@@ -168,6 +168,15 @@ def optimize(  # noqa: C901 - FIXME: simplify this
             objective.build(verbose=verbose)
         if nonlinear_constraint is not None and not nonlinear_constraint.built:
             nonlinear_constraint.build(verbose=verbose)
+        # check and ensure ProximalProjection is only used with eq as an optimizable
+        if isinstance(objective, ProximalProjection):
+            if len(objective.things) > 1:
```
f0uriest, line 173: should also check that the one thing is an equilibrium

desc/objectives/_geometry.py
ddudt: Oh yes, this needs to be changed. It should say something like: 

> If True, the surface is fixed and its coordinates are precomputed, which saves on computation time during optimization. In this case `self.things = [eq]` only. If False (default), the surface coordinates are computed at every iteration and `self.things = [eq, surface]`.

ddudt (changes requested):

f0uriest (approved):

f0uriest:
I think we should be able to get this to work even with `fixed_surface=False` (or with any other `thing`), the proximal part is more limited by what constraints than what objectives/things. I'll try to take a look at it.

ddudt (approved):

f0uriest:
Going to merge this for now. I have some stuff that seems to work with `fixed_surface=False` but will need more testing and also required rolling in some changes from other PRs, so probably better to wait on that for a bit.

PR #758, opened 2023-11-20
ko/faster_zernike

### LINKED ISSUE #720: Faster Zernike evaluation
We've been adding more stuff that needs computational grids that change during an optimization like particle tracing, field aligned stuff etc. This can be a significant slowdown because we can't pre-compute the transform matrices.

The primary bottleneck is in evaluating the Zernike polynomials (the matmul and fourier stuff is relatively quick), so any speedup in evaluating the Zernike stuff could be a huge speed improvement for those sorts of things.

Right now we're using a recurrence relation based on jacobi polynomials, which may not be the best option. There are other recurrence relations for the Zernike polynomials directly that may be more efficient.

Another thing is that the function is currently written to evaluate for a single L, M at a time and then vectorized. However in most cases we're evaluating an array of L, M which leads to some redundant calculations (ie, L=1, M=1 gets recalculated when computing L=2, M=2). We could likely change the function to get rid of these redundant calculations for some significant speedup. There are similar recurrence relations for the derivatives so we could likely have some more savings there.

Might be a good project for an undergrad or one of the new folks if they're interested in the math/numerics side of things

@JaxCrocker @YigitElma @kianorr @unalmis

kianorr:
resolves #720


dpanici:
I think the initial commit might have included a free boundary file on accident?

kianorr:
@dpanici my bad

YigitElma:
I have added a new method that doesn't calculate already calculated alpha, n pairs in Jacobi recursion relations. I put 2 cells at the end of the zernike_eval notebook for you to see the results. It is around 10% faster than the current algorithm. The accuracy is the same. I didn't work on derivatives, but if you are happy with the current version, I can include them too. @f0uriest @dpanici

YigitElma:
image
image

YigitElma:
My method basically tries to calculate only last row of this table. The current version calculates every mode without making use of the recursion relation
image

YigitElma:
Fully JIT able version is done, and it is fast.
```
zernike_radial, 0th derivative with DESC method
2.36 ms ± 116 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
zernike_radial, 0th derivative with optimized method
476 µs ± 84.5 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

I will work on the derivatives later.

YigitElma:
The results for the derivatives. (Previous one was on my computer, this one is on cluster, there might be slight difference)
```
Zernike Radial with DESC method
5.58 ms ± 29.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
First derivative with DESC method
11 ms ± 42 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
Second derivative with DESC method
16 ms ± 21 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
Third derivative with DESC method
21.1 ms ± 52.4 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
Fourth derivative with DESC method
26.1 ms ± 66.8 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

Zernike Radial with optimized method
604 µs ± 1.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
First derivative with optimized method
1.06 ms ± 3.09 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Second derivative with optimized method
1.44 ms ± 1.48 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Third derivative with optimized method
1.84 ms ± 3.32 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Fourth derivative with optimized method
2.17 ms ± 4.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

f0uriest:
> The results for the derivatives. (Previous one was on my computer, this one is on cluster, there might be slight difference)
> ```
> Zernike Radial with DESC method
> 5.58 ms ± 29.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> First derivative with DESC method
> 11 ms ± 42 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> Second derivative with DESC method
> 16 ms ± 21 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> Third derivative with DESC method
> 21.1 ms ± 52.4 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
> Fourth derivative with DESC method
> 26.1 ms ± 66.8 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
> 
> Zernike Radial with optimized method
> 604 µs ± 1.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> First derivative with optimized method
> 1.06 ms ± 3.09 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> Second derivative with optimized method
> 1.44 ms ± 1.48 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> Third derivative with optimized method
> 1.84 ms ± 3.32 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> Fourth derivative with optimized method
> 2.17 ms ± 4.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> ```

This looks great! Was this for the super high res case? Would be good to also try more moderate resolution (L,M = 6, 8, 10, 12) to get a better idea of how it scales.

desc/basis.py
```diff
@@ -1413,6 +1413,424 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+
+        # Find the index corresponding to the original array
+        index = jnp.where(
+            jnp.logical_and(m == alpha, n == N),
+            jnp.arange(m.size),
+            0,
+        )
+        # jnp.where() will result in 1D array with 1 value which
+        # is not 0. Sum it to find the index
+        idx = jnp.sum(index)
```
f0uriest, line 1491: Are we sure there will only be one nonzero value? Might need to also test for stuff with duplicate modes etc

desc/basis.py
```diff
@@ -1413,6 +1413,424 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+    m : ndarray of int, shape(K,)
+        azimuthal mode number(s)
+
+    Returns
+    -------
+    out : ndarray, shape(N,K)
+        basis function(s) evaluated at specified points
+
+    """
+    if dr > 4:
```
f0uriest, line 1471: I'm wondering if it's possible to also do something clever for the derivative. I think the k derivative can be written in terms of different l,m for the k-1 derivative, so we could possibly do all the derivatives in one big loop?

desc/basis.py
```diff
@@ -1413,6 +1413,424 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+    This is completely JITable version of zernike_radial_optimized
+    function.
+
+    Parameters
+    ----------
+    r : ndarray, shape(N,)
+        radial coordinates to evaluate basis
+    l : ndarray of int, shape(K,)
+        radial mode number(s)
+    m : ndarray of int, shape(K,)
```
f0uriest, line 1462: Missing dr in docstring

desc/basis.py
```diff
@@ -1413,6 +1413,424 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+        out = out.at[:, idx].set((-1) ** N * result)
+
+        P_n2 = jnp.where(N >= 2, P_n1, P_n2)
+        P_n1 = jnp.where(N >= 2, P_n, P_n1)
+        dP_n2 = jnp.where(N >= 3, dP_n1, dP_n2)
+        dP_n1 = jnp.where(N >= 3, dP_n, dP_n1)
+
+        return (r_jacobi, alpha, r, out, (P_n1, P_n2, dP_n1, dP_n2), m, n)
+
+    # USE DOT PRODUCT TO GENERALIZE DERIVATIVES
```
f0uriest, line 1531: Can you explain this a bit more?

desc/basis.py
```diff
@@ -1413,6 +1413,424 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+                ddddP_n2,
+            ),
+            m,
+            n,
+        )
+
+    def body(alpha, args):
+        r, r_jacobi, L_max, m, n, out = args
+        # Maximum possible value for n for loop bound
+        N_max = (L_max - alpha) // 2
```
f0uriest, line 1728: Are we sure this only depends on Lmax and not Mmax?

desc/basis.py
YigitElma: For the derivatives, I used the [formula](https://en.wikipedia.org/wiki/Jacobi_polynomials#Derivatives). I am not sure if we can relate other derivatives to first derivative since alpha and beta also change. But we can definitely be more compact with coding I guess.

desc/basis.py
YigitElma: Yeah I assumed there are no duplicate modes. And also I assumed that mode exists. Since I am constructing in between modes from the max M, if for example for Mmax=10, there is actually no m=5th mode, this code will change the 0th index of the out array. It's a good idea to some checks before that

desc/basis.py
YigitElma: For the derivatives, we have derivative order+1 terms in result formula (i.e. a1*f + a2*df + a3*ddf + a4 * dddf, we can write this by calculatinf some A=[a1 a2 a3 a4] and F=[f df ddf dddf] and taking jnp.dot(A,F) ). I wanted to get some results first, before finalizing the algorithm.

desc/basis.py
YigitElma: The purpose of body function is to loop over every m value from 0 to Mmax, assuming that we have every mode. However, again by assuming we have every mode, the highest value corresponding to any m is Lmax. I am not very familiar with other basis uses, so I hoped my assumptions are valid, but if not I can chane Lmax with jnp.max(l) kind of thing

YigitElma:
> > The results for the derivatives. (Previous one was on my computer, this one is on cluster, there might be slight difference)
> > ```
> > Zernike Radial with DESC method
> > 5.58 ms ± 29.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> > First derivative with DESC method
> > 11 ms ± 42 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> > Second derivative with DESC method
> > 16 ms ± 21 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> > Third derivative with DESC method
> > 21.1 ms ± 52.4 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
> > Fourth derivative with DESC method
> > 26.1 ms ± 66.8 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
> > 
> > Zernike Radial with optimized method
> > 604 µs ± 1.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> > First derivative with optimized method
> > 1.06 ms ± 3.09 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> > Second derivative with optimized method
> > 1.44 ms ± 1.48 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> > Third derivative with optimized method
> > 1.84 ms ± 3.32 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> > Fourth derivative with optimized method
> > 2.17 ms ± 4.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> > ```
> 
> This looks great! Was this for the super high res case? Would be good to also try more moderate resolution (L,M = 6, 8, 10, 12) to get a better idea of how it scales.

Yes, it is for 50 (included the basis at the top now). I can test and send others to here too

YigitElma:
```
basis = ZernikePolynomial(L=5, M=5, spectral_indexing="ansi", sym="cos")
r = np.linspace(0, 1, 100)
```

```
Zernike Radial with DESC method
38.1 µs ± 112 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
First derivative with DESC method
62.3 µs ± 278 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Second derivative with DESC method
95.3 µs ± 1.06 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Third derivative with DESC method
125 µs ± 254 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Fourth derivative with DESC method
158 µs ± 398 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

Zernike Radial with optimized method
28.7 µs ± 218 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
First derivative with optimized method
46.9 µs ± 155 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Second derivative with optimized method
64.7 µs ± 247 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Third derivative with optimized method
83.9 µs ± 469 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Fourth derivative with optimized method
98.9 µs ± 535 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

YigitElma:
```
basis = ZernikePolynomial(L=10, M=10, spectral_indexing="ansi", sym="cos")
r = np.linspace(0, 1, 100)
```

```
Zernike Radial with DESC method
133 µs ± 408 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
First derivative with DESC method
253 µs ± 800 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Second derivative with DESC method
376 µs ± 1.12 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Third derivative with DESC method
490 µs ± 2.24 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Fourth derivative with DESC method
585 µs ± 1.07 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

Zernike Radial with optimized method
56.1 µs ± 194 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
First derivative with optimized method
104 µs ± 534 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Second derivative with optimized method
150 µs ± 614 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Third derivative with optimized method
196 µs ± 1.81 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
Fourth derivative with optimized method
232 µs ± 2.35 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

desc/basis.py
YigitElma: Actually, these assumptions caused problems for existing tests with specified modes
l = np.array([3, 4, 6])
m = np.array([1, 2, 2])
I will look for it in detail.

f0uriest: I think this indexing along with the assignment below could be replaced by

```python
out = jnp.where(jnp.logical_and(m == alpha, n == N), (-1) ** N * r**alpha * P_n, out)
```

Which I think should also work for duplicate modes or non-sequential stuff

YigitElma: I believe I tried something like this before (and tried again now). The problem is out is the full result array, so it is 2D with (r.size, m.size). However, jnp.logical_and(m == alpha, n == N) is 1D with m.size and (-1) ** N * r**alpha * P_n is 1D with r.size . Dimensions cause broadcasting issue.

f0uriest: Can't you just broadcast things like normal?

```python
mask = jnp.logical_and(m == alpha, n == N)
val = (-1) ** N * r**alpha * P_n
out = jnp.where(mask[None,:], val[:,None], out)
```

f0uriest: In practice there will often be duplicate modes. The most common reason is if N (toroidal resolution) is nonzero, in which case there will be a L,M pair for each N, but the zernike radial part will see that as duplicates for L,M

YigitElma:
I have changed the naming and the old version is now called zernike_radial_old. Here are the final speed comparison of the 2. For the basis (Rory said it is most common case),
```python
basis = FourierZernikeBasis(L=10, M=10, N=10)
r = np.linspace(0, 1, 100)
```
image

YigitElma:
@dpanici @f0uriest I changed all the function calls that I can find to what my function accepts (without np.newaxis) . Please let me know if there are any problems.

YigitElma:
In initial_guess.py and linear_objectives.py files, there some other ways which you called zernike_radial function.
```python
scale = zernike_radial(coord, abs(m), m)
self._A[j, i] = zernike_radial(surf, l, m)
```
These are probably causing issue for some tests.

YigitElma:
It seems like jit compilation time of the new function is longer than previous one. Also, the one which is very long and has many different functions for derivatives seemed to be compiled faster compared to my lastest attempt to combine all derivative operations to a single body function.

YigitElma:
> It seems like jit compilation time of the new function is longer than previous one. Also, the one which is very long and has many different functions for derivatives seemed to be compiled faster compared to my lastest attempt to combine all derivative operations to a single body function.

I change the initial guess function a little bit to make it faster. Actually the problem was not compilation (yes, it is slow but not that much), but initial guess function was calling zernike_radial with all scalar values inside a fori_loop, hence for no computational advantage. I changed it to calculate all modes at the same time, so make use of recurrence relation.

desc/basis.py
```diff
@@ -1414,9 +1413,584 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
 
 
 @functools.partial(jit, static_argnums=3)
+def zernike_radial_separate(r, l, m, dr=0):  # noqa: C901
```
f0uriest, line 1416: Is this function needed?

desc/basis.py
```diff
@@ -1414,9 +1413,584 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+    P_n = jnp.where(n == 1, (alpha + 1) + (alpha + beta + 2) * (x - 1) / 2, P_n)
+    return P_n
+
+
+@custom_jvp
+@jit
 def zernike_radial(r, l, m, dr=0):
     """Radial part of zernike polynomials.
 
+    Calculates Radial part of Zernike Polynomials using Jacobi recursion relation
```
f0uriest, line 1791: Should cleanup this docstring

desc/basis.py
```diff
@@ -1414,9 +1413,584 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+
+    """
+    return _zernike_radial_vectorized(r, l, m, dr)
+
+
+@functools.partial(jnp.vectorize, excluded=(1, 2, 3), signature="()->(k)")
+def _zernike_radial_vectorized(r, l, m, dr):
+    # need this to be separate so that we can have a default value for dr and not
+    # vectorize over it.
+    def update(i, args):
```
f0uriest, line 1843: Can you add some comments describing what the different parts of this function are doing

desc/basis.py
```diff
@@ -1414,9 +1413,584 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+        out = cond(idx >= 0, trueFun, falseFun, (idx, result, out))
+        return (alpha, N, result, out)
+
+    def body_inner(N, args):
+        alpha, out, P_past = args
+        P_n2 = P_past[0]
+        P_n1 = P_past[1]
+        P_n = jnp.zeros(MAXDR + 1)
+
+        def find_inter_jacobi(dx, args):
```
f0uriest, line 1865: Add comment explaining what this is doing

desc/basis.py
```diff
@@ -1414,9 +1413,584 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+            + coef[2] * 48 * (2 * alpha**2 + 2 * alpha + 1) * r**alpha * P_n[2]
+            - coef[3] * 128 * (2 * alpha + 3) * r ** (alpha + 2) * P_n[3]
+            + coef[4] * 256 * r ** (alpha + 4) * P_n[4]
+        )
+        branch5 = lambda: jnp.nan
+        branches = [branch0, branch1, branch2, branch3, branch4, branch5]
+        result = switch(dr, branches)
+        _, _, _, out = fori_loop(0, m.size, update, (alpha, N, result, out))
+
+        # Shift past values if needed
```
f0uriest, line 1918: Explain what this is doing?

desc/basis.py
```diff
@@ -1414,9 +1413,584 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+    df = zernike_radial(r, l, m, dr + 1)
+    # in theory l, m, dr aren't differentiable (they're integers)
+    # but marking them as non-diff argnums seems to cause escaped tracer values.
+    # probably a more elegant fix, but just setting those derivatives to zero seems
+    # to work fine.
+    return f, (df.T * rdot).T + 0 * ldot + 0 * mdot + 0 * drdot
+
+
+@functools.partial(jit, static_argnums=3)
+def zernike_radial_old(r, l, m, dr=0):
```
f0uriest, line 1991: Can probably remove this if we're confident the new one is better in all ways

f0uriest (changes requested):

YigitElma:
Some changes had to be made to allow using reverse mode AD. Moreover, additional checks added to the first version of this PR had a substantial effect on the performance. Still, this is faster than previous version for upto L=M=20, but extreme resolution is a bit slower. What we can do is add new functions for derivatives, such that zeroth derivative doesn't suffer from extra calculations etc. That function would still be encapsulated by zernike_radial, so the syntax would be the same.

YigitElma:
@f0uriest I couldn't make switch work, my solution is using couple jax.lax.cond . It doesn't look good but it works. If you don't like it, I can revert back

f0uriest:
> @f0uriest I couldn't make switch work, my solution is using couple jax.lax.cond . It doesn't look good but it works. If you don't like it, I can revert back

How much of an improvement is this version over the past? There's a lot of duplicated code that would be annoying to maintain.

YigitElma:
> > @f0uriest I couldn't make switch work, my solution is using couple jax.lax.cond . It doesn't look good but it works. If you don't like it, I can revert back
> 
> How much of an improvement is this version over the past? There's a lot of duplicated code that would be annoying to maintain.

For L=M=N=12 basis,
image

For L=M=50 Zernike basis,
image

It is around 2 times faster for LMN=12 case.  And 3-4 times faster for higher resolutions. My main reason for this is that, for low resolutions around 12, even zernike_radial_poly function is fast (as you have pointed out in old function, the real performance effect comes for resolution greater than 24). So, until we find a better solution, this might be the fastest.

f0uriest (approved):
Approving this for now, but should get some other folks to look at it first as I've been staring at it for a while so I might have missed things

There is still a bunch of duplicated code, not sure how much that can be reduced while still working with JAX.

YigitElma:
> Approving this for now, but should get some other folks to look at it first as I've been staring at it for a while so I might have missed things
> 
> There is still a bunch of duplicated code, not sure how much that can be reduced while still working with JAX.

Yes I agree. It's been a while for us.

dpanici:
I will review this tomorrow night, sorry

desc/basis.py
```diff
@@ -1413,83 +1412,150 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+    Calculates Radial part of Zernike Polynomials using Jacobi recursion relation
+    by getting rid of the redundant calculations for appropriate modes.
+    https://en.wikipedia.org/wiki/Jacobi_polynomials#Recurrence_relations
+
+    For the derivatives, the following formula is used with above recursion relation,
+    https://en.wikipedia.org/wiki/Jacobi_polynomials#Derivatives
+
+    Used formulas are also in the zerike_eval.ipynb notebook in docs.
+
+    This function can be made faster. However, JAX reverse mode AD causes problems.
```
dpanici, line 1429: Should make an issue for this instead of including as a docstring, or put as a TODO comment somewhere where the problems occur

desc/basis.py
```diff
@@ -1413,83 +1412,150 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+def _zernike_radial_vectorized(r, l, m, dr):
+    """Calculation of Radial part of Zernike polynomials."""
+
+    def body_inner(N, args):
+        alpha, out, P_past = args
+        P_n2 = P_past[0]  # Jacobi at N-2
+        P_n1 = P_past[1]  # Jacobi at N-1
+        P_n = jacobi_poly_single(r_jacobi, N, alpha, 0, P_n1, P_n2)
+
+        # Only calculate the function at dr th index with input r
```
dpanici, line 1472: what is dr th index?

desc/basis.py
YigitElma: #862 and #881 are related to this PR

desc/basis.py
YigitElma: Ah my bad, at some point we had the switch function here, so it was to execute the intended derivative function

desc/equilibrium/equilibrium.py
```diff
@@ -623,6 +623,8 @@ def get_surface_at(self, rho=None, theta=None, zeta=None):
             AR = np.zeros((surface.R_basis.num_modes, self.R_basis.num_modes))
             AZ = np.zeros((surface.Z_basis.num_modes, self.Z_basis.num_modes))
 
+            # TODO: Don't call zernike_radial for each mode, instead compute all
```
dpanici, line 626: is there an issue for this so we don't lose track of it? and for this in the linear objectives too

desc/basis.py
```diff
@@ -1413,83 +1412,150 @@ def zernike_radial_poly(r, l, m, dr=0, exact="auto"):
...
+
+def jacobi_poly_single(x, n, alpha, beta=0, P_n1=0, P_n2=0):
+    """Evaluate Jacobi for single alpha and n pair."""
+    c = 2 * n + alpha + beta
+    a1 = 2 * n * (c - n) * (c - 2)
+    a2 = (c - 1) * (c * (c - 2) * x + (alpha - beta) * (alpha + beta))
+    a3 = 2 * (n + alpha - 1) * (n + beta - 1) * c
+
+    # Check if a1 is 0, to prevent division by 0
+    a1 = jnp.where(a1 == 0, 1e-6, a1)
```
dpanici, line 1539: can `safediv` be used instead of this?

desc/basis.py
```diff
@@ -1595,134 +1661,440 @@ def fourier(theta, m, NFP=1, dt=0):
...
-    val : int, float, array-like
-        number of possible combinations
+    norm : float
+        the integral (Z^m_l)^2 r dr dt, r=[0,1], t=[0,2*pi]
+
     """
-    # adapted from scipy:
-    # https://github.com/scipy/scipy/blob/701ffcc8a6f04509d115aac5e5681c538b5265a2/
-    # scipy/special/orthogonal_eval.pxd#L68
+    return np.sqrt((2 * (l + 1)) / (np.pi * (1 + int(m == 0))))
```
dpanici, line 1678: test for this function?

dpanici (changes requested):
Add a test for the norm of a zernike function, and use `safediv` instead of your own version of `safediv`

desc/basis.py
dpanici: So this is an old comment that is no longer needed?

YigitElma: I changed this with the last commit

desc/basis.py
YigitElma: I don't even remember coding this :D

f0uriest: This is (very) old code, it just got moved around. Iirc there is already a test for it

YigitElma: Oh yeah it is 3 years old. Definitely not me!

desc/basis.py
YigitElma: Hmm, safediv has extra checks, I don't think we need that everytime this function is called. jnp.where for single argument is fast but still takes time, especially if the intended value is in the last argument, (for larger array conditions, I realized if 3rd instead of 2nd entry is used most of the time, it takes significantly more time). I am not sure about how much difference would it make for this case, but still this line is just for not throwing error in vain, only possible a1==0 cases are n=0 or 1, and they are replaced right at the end, it is dummy calculation anyway

YigitElma:
```python
c = 2 * n + alpha + beta
a1 = 2 * n * (c - n) * (c - 2)
```
a1=0 for n=0, c=n or c=2,
c=n :
    n = -(alpha+beta) where alpha and beta both positive so impossible

n=2:
   n=0 and alpha+beta=2 or n=1 and alpha=beta=0

dpanici: so are you saying this line is not needed? But anyhow If this is faster than safediv then ok that is fine, we can keep it as is, thanks for the explanaiton

dpanici (approved):

desc/basis.py
YigitElma: yeah, it is kind of not needed, some tests were failing, so I added it.

PR #760, opened 2023-11-21
Fixes to ensure reverse mode AD works

### LINKED ISSUE #738: Check all objectives for NaN gradients
In several places we inadventently divide by zero, but then replace the value with something sensible. This works fine in forward mode AD, but can cause issues in reverse mode (like `jax.grad`)

f0uriest:
Resolves #738


desc/compute/utils.py
```diff
@@ -489,6 +489,30 @@ def cross(a, b, axis=-1):
...
+    fill : float, ndarray, optional
+        Value to return where x is zero.
+    threshold : float >= 0
+        How small is x allowed to be.
+
+    """
+    is_zero = (jnp.abs(x) <= threshold).all(axis=axis, keepdims=True)
+    x = jnp.where(is_zero, jnp.ones_like(x), x)  # replace x with ones if is_zero
+    n = jnp.linalg.norm(x, ord=ord, axis=axis)
+    n = jnp.where(is_zero.squeeze(), fill, n)  # replace norm with zero if is_zero
```
ddudt, line 512: why `squeeze()` here and not in line 510?

ddudt (approved):

dpanici (approved):

PR #761, opened 2023-11-21
Allow specification of AD mode

### LINKED ISSUE #666: Allow specification or auto detection of AD mode
Right now we use forward mode AD for pretty much everything, since it tends to be the most efficient in our use cases. However, some future objectives (ie, ODE based ones) may be more efficient in reverse mode (or may only support reverse mode). We should figure out some way to select which mode to use, or possibly autodetect based on shape of the jacobian and type of objective. There may also be a need to combine forward and reverse mode within 1 objective.

f0uriest:
Resolves #666

### DIFF
--- a/desc/objectives/objective_funs.py
+++ b/desc/objectives/objective_funs.py
@@ -1,6 +1,7 @@
 """Base classes for objectives."""
 
 from abc import ABC, abstractmethod
+from functools import partial
 
 import numpy as np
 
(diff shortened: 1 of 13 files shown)


desc/objectives/objective_funs.py
```diff
@@ -907,22 +890,34 @@ def build(self, use_jit=True, verbose=1):
     def compute(self, *args, **kwargs):
         """Compute the objective function."""
 
+    def _maybe_array_to_params(self, *args):
```
ddudt, line 893: Why is this needed when it wasn't before?

ddudt (approved):

PR #763, opened 2023-11-21
Experimental support for optimization using multiple devices

f0uriest:
This doesn't work as is, will likely need to wait until we get MPI working.


f0uriest:
Closing this in favor of #1071

PR #767, opened 2023-11-21
Make magnetic field classes optimizable

f0uriest:
- Adds optimizable parameters to various magnetic field classes.
- Adds base class for collections of optimizable objects like coilsets
- Refactors SplineMagneticField to be closer to mgrid stuff, with independent coil groups


desc/magnetic_fields.py
```diff
@@ -383,15 +385,44 @@ class ScaledMagneticField(_MagneticField):
...
+    def __init__(self, scale, field):
+        assert np.isscalar(scale) or len(scale) == 1, "scale must be a scalar value"
         assert isinstance(
             field, _MagneticField
         ), "field should be a subclass of MagneticField, got type {}".format(
             type(field)
         )
-        self._scalar = scalar
-        self._field = field
+        object.__setattr__(self, "_scale", scale)
```
dpanici, line 395: why this syntax and not the usual `self._scale = scale` with a default value for `scale` in the init?

f0uriest: It's needed because I'm overriding the `getattr` and `setattr` behavior so that the scaled field "acts like" the underlying field for most purposes (ie the optimizable attributes are seen as a pass through)

desc/magnetic_fields.py
```diff
@@ -861,18 +1018,19 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
...
                 self._extrap,
-                (None, self._period, None),
+                (None, 2 * np.pi / self.NFP, None),
                 **self._derivs["BZ"],
             )
-        B = jnp.array([BRq, Bphiq, BZq]).T
+        # BRq etc shape(nq, ngroups)
+        B = jnp.stack([BRq, Bphiq, BZq], axis=1)
+        # B shape(nq, 3, ngroups)
+        B = jnp.sum(B * currents, axis=-1)
```
dpanici, line 1027: Why the Ngroups parameter added? So that you can pass in different sources of fields, which may have different scaling factors associated with them, without having to manualy scale/sum them before passing into the class?

dpanici (approved):
Excited to use this with the regcoil stuff

tests/test_magnetic_fields.py
```diff
@@ -60,6 +60,85 @@ def test_basic_fields(self):
             (tfield + vfield - pfield)([1, 0, 0.1]), [[0.4, 2, 1]]
         )
 
+    @pytest.mark.unit
+    def test_combined_fields(self):
+        """Tests for sum/scaled fields."""
+        tfield = ToroidalMagneticField(2, 1)
+        vfield = VerticalMagneticField(1)
+        pfield = PoloidalMagneticField(2, 1, 2)
```
rahulgaur104, line 68: What are these indices? What is this test doing? Can you add more info in the docstring?

dpanici: they are not indices, they are the arguments to the init of each. e.g. the B0 and R0 for the `ToroidalMagneticField`, the B0 for `VerticalMagneticField`

dpanici (dismissed):

desc/magnetic_fields.py
```diff
@@ -787,6 +943,7 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
 
         """
         assert basis.lower() in ["rpz", "xyz"]
+        currents = self.currents if params is None else params["currents"]
```
ddudt, line 946: Should update the docs, it says `params` is unused by this method.

ddudt (changes requested):
need to update the docs for the new current parameter

dpanici (dismissed):

ddudt (approved):

dpanici (approved):

PR #781, opened 2023-12-01
remove the default legend from plot_surfaces, since plot_comparison w…

dpanici:
…ill make a legend as well


YigitElma (approved):
I was actually thinking about fixing this today, thanks!

ddudt (approved):

PR #782, opened 2023-12-03
Poincare Boundary Condition

### LINKED ISSUE #204: Add Utility Functions for Poincare XS BC
Add utility fxns for poincare XS BC, as methods of the bases they convert

- [ ] FourierZernike 3D to Zernike 2D
- [ ] FourierZernike 3D to FourierZernike 3D without N modes
- [ ]  FourierZernike 3D to DoubleFourier 2D on LCFS (or specified rho surface)

### LINKED ISSUE #773: Tutorial: Poincare BC

YigitElma:
> [!note]
> I will use this PR only for CI, and once it works properly, I will close it and reference only the branch in the paper. In a follow-up small PR, I can clean up the Poincaré-related things in master, and add some of the useful code from this PR.

This PR focuses on solving equilibrium by giving a Poincare section at zeta 0 surfaces instead of giving the last closed flux surface, LCFS. 

This method gives exactly same results for high NFP >12, but different results for low NFP cases. Note: as far as my observation, the difference is caused by the NFP, but it might caused by something else that I am missing.

- Adds `Rp_lmn`, `Zp_lmn` and `Lp_lmn` to `Equilibrium` class
These are not `optimizable_parameter` by default, instead when you call `build` method of one of the `FixSection...` constraints, they become optimizable and `Rb_lmn` and `Zb_lmn` become un-optimizable. This behavior can be changed by `remove_optimizables` argument.
- Adds `make_optimizable` and `make_nonoptimizable` for this conversion. These are called automatically in `FixSection..` and `FixBoundary..` constraints.
- Alters `ProximalProjection` to solve inner equilibrium with Poincare boundary condition. This is done automatically is a user gives `FixSection..` constraint (they shouldn't mix any `FixBoundary..`)
- `get_surface_at` method of `Equilibrium` can now return any arbitrary zeta cross-section including the lambda values. If the asked cross-section is not 0 or pi/NFP, then the return `ZernikeRZTorodialSection` object doesn't have symmetry.
- Adds tutorial on how to use Poincare BC in different contexts. Also, adds an example to use in optimization without running it for long.
- Removes `bdry_mode` attribute
- Adds `update_all_equilibria_backwards_compatibility.py` script to devtools that load and then saves (almost) every equilibria in desc/examples and tests/inputs to run `_set_up` method.  

Resolves #204
Resolves #773


YigitElma:
Example usage of 2 boundary surface

```python
def get_poin2(eq, zeta=0):
    surface = eq.get_surface_at(zeta=zeta/eq.NFP)
    Lb_lmn, Lb_basis = FourierZernike_to_FourierZernike_no_N_modes(eq.L_lmn, eq.L_basis, zeta)

    eq_poin = Equilibrium(surface=surface,
                pressure=eq.pressure,
                iota=eq.iota,
                Psi=eq.Psi, # flux (in Webers) within the last closed flux surface
                NFP=eq.NFP, # number of field periods
                L=eq.L, # radial spectral resolution
                M=eq.M, # poloidal spectral resolution
                N=eq.N, # toroidal spectral resolution 
                L_grid=eq.L_grid, # real space radial resolution, slightly oversampled
                M_grid=eq.M_grid, # real space poloidal resolution, slightly oversampled
                N_grid=eq.N_grid, # real space toroidal resolution
                sym=True, # explicitly enforce stellarator symmetry
                bdry_mode='poincare',
                spectral_indexing=eq._spectral_indexing,
                )
    eq_poin.L_lmn = (
        Lb_lmn  # initialize the poincare eq with the lambda of the original eq
    )
    eq_poin.axis = eq_poin.get_axis()
    return eq_poin

eq = get("HELIOTRON")
zeta = 0
zeta2 = np.pi

# Take second surface to be fixed
surface2 = eq.get_surface_at(zeta=zeta2/eq.NFP)
surface2.change_resolution(eq.L,eq.M,eq.N)

eq_poin = get_poin2(eq, zeta=zeta)
eq_poin.change_resolution(eq_poin.L,eq_poin.M,eq.N) 

Lb2_lmn, Lb2_basis = FourierZernike_to_FourierZernike_no_N_modes(eq.L_lmn, eq.L_basis, zeta=zeta2)

constraints = get_fixed_boundary_constraints(eq=eq_poin,poincare_lambda=True, zeta=zeta)
constraints += (
    SecondBoundaryRSelfConsistency(eq=eq_poin, zeta=zeta2, surface=surface2), 
    SecondBoundaryZSelfConsistency(eq=eq_poin, zeta=zeta2, surface=surface2),
    SecondBoundaryLambdaSelfConsistency(eq=eq_poin, zeta=zeta2, Lb2_lmn=Lb2_lmn, Lb2_basis=Lb2_basis),
    )
objective = ObjectiveFunction(ForceBalance(eq_poin))

eq_poin.solve(
    verbose=3, 
    ftol=0, 
    objective=objective, 
    constraints=constraints, 
    optimizer=optimizer,
    maxiter=20, 
    xtol=0,
    zeta=zeta, 
    zeta2=zeta2
    )
plot_comparison(eqs=[eq,eq_poin],labels=['LCFS',f'2 BC Poincare']);
````

File: desc/objectives/linear_objectives.py [hunk 1]
```diff
@@ -403,6 +474,460 @@ def compute(self, params, constants=None):
...
+            self._dim_f = self._A.shape[0]
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params, constants=None):
+        """Compute deviation from desired boundary."""
+        return jnp.dot(self._A, params["L_lmn"])
+
+
+class SecondBoundaryRSelfConsistency(_Objective):
```
f0uriest, line 592: what are these needed for?

desc/basis.py
```diff
@@ -1726,3 +1726,180 @@ def zernike_norm(l, m):
 
     """
     return np.sqrt((2 * (l + 1)) / (np.pi * (1 + int(m == 0))))
+
+
+def FourierZernike_to_PoincareZernikePolynomial(X_lmn_3D, basis_3D, zeta=0):
```
f0uriest, line 1731: would these be better as methods of the relevant class?

like `FourierZernikeBasis.to_ZernikePolynomial`

f0uriest (changes requested):
I think the better API would be to introduce new degrees of freedom Rp, Zp etc like we have for the LCFS and axis, and then constrain those to be self consistent.

This would eliminate the `boundary_mode` thing, and allow more flexibility for setting up constraints.

Each equilibrium, in addition to having a `.surface` and `.axis` attribute would have a `.poincare` or `.section` property with the ZernikeRZToroidalSurface object. This would then have an assigned zeta value so we wouldn't need to pass around zeta to different functions.

desc/objectives/linear_objectives.py
YigitElma: I was trying to see if giving 2 Poincare sections (one at zeta=0 and one at zeta=pi) would help or not. Probably these won't be needed in the final PR.

YigitElma:
> I think the better API would be to introduce new degrees of freedom Rp, Zp etc like we have for the LCFS and axis, and then constrain those to be self consistent.
> 
> This would eliminate the `boundary_mode` thing, and allow more flexibility for setting up constraints.
> 
> Each equilibrium, in addition to having a `.surface` and `.axis` attribute would have a `.poincare` or `.section` property with the ZernikeRZToroidalSurface object. This would then have an assigned zeta value so we wouldn't need to pass around zeta to different functions.

@f0uriest I made some changes to the API. First of all, I created a new class called PoincareSurface (which is an inherited class of ZernikeRzToroidalSection) to be able to store lambda basis and coefficients in the section. That class needs more cleaning but my main aim is to decide which type of objectives to use by checking the eq.surface type. This can be either FourierRZToroidalSurface or PoincareSurface. I changed the previous eq.bdry_mode stuff to that. Also, I am using the zeta of the surface now, so no need to pass zeta value to optimizer and objectives anymore.

desc/optimize/optimizer.py
```diff
@@ -83,6 +83,7 @@ def optimize(  # noqa: C901 - FIXME: simplify this
         maxiter=None,
         options=None,
         copy=False,
+        zeta=None,
```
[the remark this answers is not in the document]
YigitElma: I made some changes regarding that.

ddudt:
Let's get rid of support for the Poincare BC from a text input file in this PR

desc/examples/W7X_output.h5
dpanici, line 1: should remove this, it is unneeded as we have W7-X instead, this must have been a mistaken upload

desc/basis.py
```diff
@@ -2092,3 +2092,180 @@ def body(alpha, out):
     # loop which will execute necessary n values.
     out = fori_loop(0, (M_max + 1).astype(int), body, (out))
     return out
+
+
+def get_basis_poincare(X_lmn_3D, basis_3D, zeta=0):
```
f0uriest, line 2097: are these new functions needed?

desc/compute/utils.py
```diff
@@ -1365,3 +1365,38 @@ def body(i, mins):
     # The above implementation was benchmarked to be more efficient than
     # alternatives without explicit loops in GitHub pull request #501.
     return grid.expand(mins, surface_label)
+
+
+# defines the order in which objective arguments get concatenated into the state vector
+arg_order = (
```
f0uriest, line 1371: This was deleted a while ago but seems to keep coming back. Nothing references `arg_order` anymore so it should be removed

desc/compute/utils.py
```diff
@@ -1365,3 +1365,38 @@ def body(i, mins):
...
+    "Ra_n",
+    "Za_n",
+    "Rb_lmn",
+    "Zb_lmn",
+    "Rp_lmn",
+    "Zp_lmn",
+    "Lp_lmn",
+)
+# map from profile name to equilibrium parameter name
+profile_names = {
```
f0uriest, line 1393: i think this is already defined at the top of this file

desc/equilibrium/equilibrium.py
```diff
@@ -221,12 +229,15 @@ def __init__(
         self._Z_sym = "sin" if self.sym else False
 
         # surface
-        self._surface, self._bdry_mode = parse_surface(
+        self._surface = parse_surface(
```
f0uriest, line 232: should add a check to make sure the user didnt supply both surface and section

desc/equilibrium/equilibrium.py
```diff
@@ -764,6 +826,87 @@ def get_axis(self):
         axis = FourierRZCurve(R_n, Z_n, modes_R, modes_Z, NFP=self.NFP, sym=self.sym)
         return axis
 
+    def set_poincare_equilibrium(self, zeta=0):
```
f0uriest, line 829: is this needed?

desc/equilibrium/equilibrium.py
```diff
@@ -221,12 +229,15 @@ def __init__(
...
 
         # surface
-        self._surface, self._bdry_mode = parse_surface(
+        self._surface = parse_surface(
             surface, self.NFP, self.sym, self.spectral_indexing
         )
 
         # magnetic axis
-        self._axis = parse_axis(axis, self.NFP, self.sym, self.surface)
+        self._axis = parse_axis(axis, self.NFP, self.sym, self.surface, xsection)
```
f0uriest, line 237: shouldnt we parse the section first if it is needed to parse the axis?

desc/equilibrium/equilibrium.py
```diff
@@ -237,12 +248,13 @@ def __init__(
...
 
         self._N = int(setdefault(N, self.surface.N))
-        self._M = int(setdefault(M, self.surface.M))
+        self._M = int(setdefault(M, setdefault(self.surface.M, self.xsection.M)))
         self._L = int(
             setdefault(
                 L,
                 max(
                     self.surface.L,
+                    self.xsection.L,
```
f0uriest, line 257: i think you can get rid of `self.surface.L` here since its always 0 for a toroidal surface. This was a leftover from when `surface` could also be a cross section

desc/equilibrium/equilibrium.py
```diff
@@ -375,6 +388,11 @@ def __init__(
             self.L_lmn = kwargs.pop("L_lmn", jnp.zeros(self.L_basis.num_modes))
         else:
             self.set_initial_guess(ensure_nested=ensure_nested)
+            if not self.xsection.isgiven:
```
f0uriest, line 391: what is this `isgiven` attribute?

desc/equilibrium/equilibrium.py
```diff
@@ -1311,6 +1468,45 @@ def Zb_lmn(self):
...
+    def Lp_lmn(self):
+        """ndarray: Spectral coefficients of Lambda at the cross-section."""
+        return self.xsection.L_lmn
+
+    @Lp_lmn.setter
+    def Lp_lmn(self, Lp_lmn):
+        self.xsection.L_lmn = Lp_lmn
+
+    @property
+    def Lp_basis(self):
```
f0uriest, line 1502: we dont have pass through attributes for the other basis stuff, are these needed? does `eq.xsection.L_basis` not work?

desc/examples/update_all_equilibria_backwards_compatibility.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
```
f0uriest, line 1: this file might be better placed in `devtools` rather than in the desc source. 

in either case should also add it to `setup.cfg` to have coverage ignore it

desc/examples/update_all_equilibria_backwards_compatibility.py
```diff
@@ -0,0 +1,35 @@
...
+for fname in glob.glob(pwd + "/*.h5"):
+    foutputname = fname.split(".")[0].split("/")[-1]
+    print(f"\nUpdating the output file {foutputname} \n")
+    # load old output file
+    eqfam = load(fname)
+    print(eqfam)
+    for i, eq in enumerate(eqfam):
+        print(eq)
+        # update the output file
+        eq._set_up()
```
f0uriest, line 33: `_set_up()` automatically gets called after loading an object so you don't need to call it explicitly here

desc/geometry/surface.py
```diff
@@ -890,3 +893,315 @@ def set_coeffs(self, l, m=0, R=None, Z=None):
             if ZZ is not None:
                 idxZ = self.Z_basis.get_idx(ll, mm, 0)
                 self.Z_lmn = put(self.Z_lmn, idxZ, ZZ)
+
+
+class PoincareSurface(ZernikeRZToroidalSection):
```
f0uriest, line 898: im still not convinced we need a separate class for this vs just adding lambda to the existing cross section object. would like to hear other's opinions

desc/geometry/surface.py
```diff
@@ -890,3 +893,315 @@ def set_coeffs(self, l, m=0, R=None, Z=None):
...
+    def L_lmn(self):
+        """ndarray: Spectral coefficients for Lambda."""
+        return self._L_lmn
+
+    @L_lmn.setter
+    def L_lmn(self, L_lmn):
+        self._L_lmn = L_lmn
+
+    @property
+    def isgiven(self):
```
f0uriest, line 1099: this feels really weird to have this be an attribute of the class? Like, the section object shouldnt have to know how its being used

desc/objectives/linear_objectives.py
```diff
@@ -408,6 +397,329 @@ def compute(self, params, constants=None):
...
+        if eq.xsection.zeta == 0:
+            for i, (l, m, n) in enumerate(modes):
+                j = np.argwhere(
+                    np.logical_and(
+                        (eq.R_basis.modes[:, :-1] == [l, m]).all(axis=1),
+                        eq.R_basis.modes[:, 2] >= 0,
+                    )
+                )
+                self._A[i, j] = 1
+        elif eq.xsection.zeta == np.pi:
```
f0uriest, line 463: does it have to be literally pi? or pi/NFP?

ie, if NFP > 1 then zeta=pi might be in a different field period and might mean something different than the "half period" cross section

desc/objectives/normalization.py
```diff
@@ -12,6 +12,11 @@ def compute_scaling_factors(thing):
 
     scales = {}
 
+    # change this to also work with Poincare XS where bdry is  ZernikePolynomial
```
f0uriest, line 15: i dont think we need to do anything to the normalizations. If the LCFS and xsection are compatible than the scale factors that are computed here will be the same for both, so its easier to just compute them from the LCFS

setup.cfg
```diff
@@ -48,11 +48,15 @@ markers=
 filterwarnings=
     error
     ignore::pytest.PytestUnraisableExceptionWarning
+    ignore:Save attribute '_xsection' was not loaded.:RuntimeWarning
```
f0uriest, line 51: we shouldn't ignore this globally, better to fix the saved files to have the correct attribute or ignore it locally in the particular tests that cause it

setup.cfg
```diff
@@ -48,11 +48,15 @@ markers=
...
+    ignore::RuntimeWarning:desc.compute
```
f0uriest, line 52: i dont think we need to ignore this anymore? We used to divide by zero and then overwrite the bad value but i think that was fixed

tests/inputs/update_all_equilibria_backwards_compatibility.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+"""
```
f0uriest, line 2: see previous note about moving this stuff to `devtools`

we could probably also combine the scripts, or just have a single script that takes a directory as an argument and updates all the files in that directory, then we just call it twice

desc/basis.py
YigitElma: Yes, I use get_basis_poincare for setting lambda at the cross-section. Before xsection changes, I was also using FourierZernike_to_FourierZernike_no_N_modes, but I might delete it since I don't think I am using it anymore.

desc/compute/utils.py
YigitElma: Ok, that part always causes merge conflicts, probably that's why I didn't realize it.

desc/objectives/linear_objectives.py
YigitElma: It is actually pi/NFP but I made the functions take in zeta as just angle and then divide by eq.NFP inside, so the calculations are correct. For the version to be merged, I can change it back to proper zeta/eq.NFP .

setup.cfg
dpanici: Also should add a default to the `set_up` method like we had talked about in a previous PR? Or include a default at the class level where the zernike cross section is defined?

desc/objectives/linear_objectives.py
YigitElma: I tried that once but since zeta=0 and pi (pi/nfp actually) has proper lambda gauge thing, it was being weird. So I just chose 0 and pi as possible values for poincare.

tests/inputs/update_all_equilibria_backwards_compatibility.py
YigitElma: I combined the 2 files and moved them to devtools/

setup.cfg
YigitElma: Deleted it

desc/objectives/normalization.py
YigitElma: I changed it

desc/geometry/surface.py
YigitElma: I got rid of isgiven completely

desc/equilibrium/equilibrium.py
YigitElma: I got rid of taht attribute

YigitElma:
Only failing regression tests are Near axis ones. @dpanici will look into those. Probably related to #839

dpanici:
> Only failing regression tests are Near axis ones. @dpanici will look into those. Probably related to #839 

Some plot tests seem to fail too?

YigitElma:
> > Only failing regression tests are Near axis ones. @dpanici will look into those. Probably related to #839 
> 
> Some plot tests seem to fail too?

Yeah, I mean among regression tests😁

desc/examples/update_all_equilibria_backwards_compatibility.py
YigitElma: @f0uriest About this script, I have a question. Some of the tests expect some old files to have lacking attributes, and use
```python
with pytest.warns(RuntimeWarning):
    eq = load(desc_no_current_path)[-1]
```
type of loading. but since the new script adds every attribute with a default value, there is no RuntimeWarning thrown and pytest complains. Should I add exception for old files, so that it won't update them or change the tests?

f0uriest: Ahh yeah we should add an exception for those, or possibly remove those tests if we're confident they're not needed. Should check with @dpanici and @ddudt

desc/continuation.py
```diff
@@ -350,6 +350,8 @@ def _add_shaping(
             deltas["Rb_lmn"] *= bdry_step
         if "Zb_lmn" in deltas:
             deltas["Zb_lmn"] *= bdry_step
+        if "Rp_lmn" in deltas or "Zp_lmn" in deltas or "Lp_lmn" in deltas:
+            print("WARNING: Poincare BC is not yet supported for continuation")
```
f0uriest, line 354: should raise an error here, not just print a warning (and if you actually want a warning use ``warnings.warn`` or ``desc.utils.warnif`` not just a print statement)

desc/basis.py
f0uriest: ok id definitely vote for removing things that arent strictly necessary to reduce the amount of code we have to maintain

desc/equilibrium/equilibrium.py
f0uriest: I think this should now happen automatically during init right?

desc/continuation.py
YigitElma: Yes, it should be better. I wrote it this way to fall back to just perturbing the last closed flux surface boundary.

desc/geometry/surface.py
YigitElma: Personally, I think any cross-section object should have lambda in it. But I don't know the other use cases of this class. What do you think @dpanici @ddudt ?

f0uriest: I'm pretty sure we don't use the base `ZernikeRZToroidalSection` anywhere right now, it was added for use with the poincare stuff so if we need to modify it for that it should be fine

desc/equilibrium/equilibrium.py
YigitElma: Yes, if you create a PoincareSurface() object, you can directly create a Poincare compatible equilibrium. But if you mean, "you updated every example file with the new script, so they should have proper xsxection. Creating additional equilibrium is unnecessary!", you are correct.

YigitElma: One advantage of this function is that it returns a copy, but yeah for the final version, we can get rid of this. It helped for setting xsection.isgiven stuff (that we don't have anymore)

desc/equilibrium/equilibrium.py
```diff
@@ -216,17 +224,33 @@ def __init__(
...
-            surface, self.NFP, self.sym, self.spectral_indexing
+        errorif(
+            surface is not None and xsection is not None,
+            ValueError,
+            "Cannot specify both surface and xsection",
         )
-
-        # magnetic axis
-        self._axis = parse_axis(axis, self.NFP, self.sym, self.surface)
+        # Parse surface, magnetic axis and cross-section
```
f0uriest, line 241: I feel like this is still a bit clunky and could lead to some weird inconsistencies. I think what we want is for at the end of this block, once surface, xsection, and axis are assigned, they should all be self-consistent (eg evaluating the xsection at rho=0 should be the same as the axis, evaluating the surface at zeta=0 should be the same as the xsection at rho=1 etc)

Right now I don't think that's the case, because the one thing (surface/xsection) that isn't given by the user is just set to the default for that class which generally won't match with the other object supplied by the user.

If we ensure that they're all self consistent here then we don't need to do anything with `set_initial_guess` since it will all be the same.

Here's a sketch of what I'm thinking:
```python
errorif(surface is not None and xsection is not None)
errorif(xsection is not None and axis is not None) # axis is redundant with xsection right?
if xsection is None and surface is None:
    self.surface = default_surface
    self.xsection = default_xsection
    # maybe need some extra logic if surface and xsection are None but axis is given, ie from NAE?
if xsection is not None:
    self.xsection = parse_xsection(xsection)
    self.surface = get_surface_from_section(self.xsection)
    self.axis = get_axis_from_xsection(self.xsection)
if surface is not None:
    self.surface = parse_surface(surface)
    if axis is None:
        self.axis = get_axis_from_surface(self.surface) # basically getting centroid of surface, M=0 modes
    else:
        self.axis = self.parse_axis(axis)
    self.xsection = get_xsection_from_surface_and_axis(self.surface, self.axis)
```

dpanici:
> ```diff
>  test_build_transform_fft_lowres         |     -2.28 +/- 1.42     | -2.12e-04 +/- 1.32e-04 |  9.08e-03 +/- 1.2e-04  |  9.30e-03 +/- 5.1e-05  |
>  test_build_transform_fft_midres         |     -0.25 +/- 1.26     | -2.16e-04 +/- 1.10e-03 |  8.74e-02 +/- 9.1e-04  |  8.76e-02 +/- 6.1e-04  |
>  test_build_transform_fft_highres        |     +1.58 +/- 1.19     | +7.22e-03 +/- 5.41e-03 |  4.63e-01 +/- 4.6e-03  |  4.56e-01 +/- 2.9e-03  |
>  test_equilibrium_init_lowres            |     +7.39 +/- 2.51     | +1.97e-02 +/- 6.70e-03 |  2.87e-01 +/- 6.6e-03  |  2.67e-01 +/- 1.2e-03  |
> -test_equilibrium_init_medres            |    +53.88 +/- 1.66     | +3.26e-01 +/- 1.01e-02 |  9.31e-01 +/- 8.7e-03  |  6.05e-01 +/- 5.1e-03  |
> -test_equilibrium_init_highres           |    +68.07 +/- 1.70     | +1.48e+00 +/- 3.71e-02 |  3.67e+00 +/- 3.4e-02  |  2.18e+00 +/- 1.5e-02  |
>  test_objective_compile_dshape_current   |     +5.67 +/- 8.54     | +2.20e-01 +/- 3.31e-01 |  4.10e+00 +/- 2.3e-01  |  3.88e+00 +/- 2.3e-01  |
>  test_objective_compile_atf              |    +10.54 +/- 5.19     | +8.04e-01 +/- 3.96e-01 |  8.44e+00 +/- 2.8e-01  |  7.63e+00 +/- 2.8e-01  |
>  test_objective_compute_dshape_current   |     +8.62 +/- 6.90     | +4.57e-04 +/- 3.66e-04 |  5.76e-03 +/- 3.6e-04  |  5.30e-03 +/- 6.4e-05  |
>  test_objective_compute_atf              |     +6.38 +/- 3.93     | +7.20e-04 +/- 4.44e-04 |  1.20e-02 +/- 4.1e-04  |  1.13e-02 +/- 1.7e-04  |
> -test_objective_jac_dshape_current       |   +110.41 +/- 10.48    | +5.20e-02 +/- 4.93e-03 |  9.90e-02 +/- 4.4e-03  |  4.71e-02 +/- 2.2e-03  |
> -test_objective_jac_atf                  |    +16.31 +/- 3.40     | +3.48e-01 +/- 7.24e-02 |  2.48e+00 +/- 4.3e-02  |  2.13e+00 +/- 5.8e-02  |
> -test_perturb_1                          |    +22.90 +/- 6.79     | +3.62e+00 +/- 1.07e+00 |  1.94e+01 +/- 9.7e-01  |  1.58e+01 +/- 4.5e-01  |
> -test_perturb_2                          |    +20.06 +/- 1.61     | +4.19e+00 +/- 3.37e-01 |  2.51e+01 +/- 9.8e-02  |  2.09e+01 +/- 3.2e-01  |
> -test_proximal_jac_atf                   |     +7.35 +/- 1.07     | +5.28e-01 +/- 7.67e-02 |  7.72e+00 +/- 6.5e-02  |  7.19e+00 +/- 4.1e-02  |
>  test_proximal_freeb_compute             |     +1.89 +/- 1.07     | +2.34e-03 +/- 1.33e-03 |  1.26e-01 +/- 6.2e-04  |  1.24e-01 +/- 1.2e-03  |
>  test_proximal_freeb_jac                 |     -0.16 +/- 1.24     | -1.12e-02 +/- 8.82e-02 |  7.08e+00 +/- 6.0e-02  |  7.09e+00 +/- 6.5e-02  |
> ```

These are some pretty large slowdowns, mainly for the init due to poincare surface being innitialized (OK), but for the rest likely just due to the addition of Zp Rp Lp attributes which make the jacobian larger for the constraint even if it then gets factored out in the null space. Might be worth thinking of ways to improve this before we merge

desc/equilibrium/equilibrium.py
```diff
@@ -237,12 +261,12 @@ def __init__(
         _assert_nonnegint(N_grid, "N_grid")
 
         self._N = int(setdefault(N, self.surface.N))
-        self._M = int(setdefault(M, self.surface.M))
+        self._M = int(setdefault(M, setdefault(self.surface.M, self.xsection.M)))
```
f0uriest, line 264: This might give weird results for the reasons above. IE, if the user passes in a xsection with M=10, then self.surface will be the default surface with M=1, so eq.M will be 1 not 10

desc/equilibrium/equilibrium.py
```diff
@@ -364,6 +389,11 @@ def __init__(
             ValueError,
             "Surface and Equilibrium must have the same symmetry",
         )
+        errorif(
```
f0uriest, line 392: I think this should get checked above where the other symmetry stuff is?

desc/equilibrium/equilibrium.py
YigitElma: I can update parse_surface function to use xsection too. I think the rest is already implemented. What you mean by get_xsection_from_surface would be basically rewriting the initial_guess stuff. Since we already do that at the end, I didn't spend time reimplementing it.

f0uriest:
> > ```diff
> > |             benchmark_name             |         dt(%)          |         dt(s)          |        t_new(s)        |        t_old(s)        | 
> > | -------------------------------------- | ---------------------- | ---------------------- | ---------------------- | ---------------------- |
> >  test_build_transform_fft_lowres         |     -2.28 +/- 1.42     | -2.12e-04 +/- 1.32e-04 |  9.08e-03 +/- 1.2e-04  |  9.30e-03 +/- 5.1e-05  |
> >  test_build_transform_fft_midres         |     -0.25 +/- 1.26     | -2.16e-04 +/- 1.10e-03 |  8.74e-02 +/- 9.1e-04  |  8.76e-02 +/- 6.1e-04  |
> >  test_build_transform_fft_highres        |     +1.58 +/- 1.19     | +7.22e-03 +/- 5.41e-03 |  4.63e-01 +/- 4.6e-03  |  4.56e-01 +/- 2.9e-03  |
> >  test_equilibrium_init_lowres            |     +7.39 +/- 2.51     | +1.97e-02 +/- 6.70e-03 |  2.87e-01 +/- 6.6e-03  |  2.67e-01 +/- 1.2e-03  |
> > -test_equilibrium_init_medres            |    +53.88 +/- 1.66     | +3.26e-01 +/- 1.01e-02 |  9.31e-01 +/- 8.7e-03  |  6.05e-01 +/- 5.1e-03  |
> > -test_equilibrium_init_highres           |    +68.07 +/- 1.70     | +1.48e+00 +/- 3.71e-02 |  3.67e+00 +/- 3.4e-02  |  2.18e+00 +/- 1.5e-02  |
> >  test_objective_compile_dshape_current   |     +5.67 +/- 8.54     | +2.20e-01 +/- 3.31e-01 |  4.10e+00 +/- 2.3e-01  |  3.88e+00 +/- 2.3e-01  |
> >  test_objective_compile_atf              |    +10.54 +/- 5.19     | +8.04e-01 +/- 3.96e-01 |  8.44e+00 +/- 2.8e-01  |  7.63e+00 +/- 2.8e-01  |
> >  test_objective_compute_dshape_current   |     +8.62 +/- 6.90     | +4.57e-04 +/- 3.66e-04 |  5.76e-03 +/- 3.6e-04  |  5.30e-03 +/- 6.4e-05  |
> >  test_objective_compute_atf              |     +6.38 +/- 3.93     | +7.20e-04 +/- 4.44e-04 |  1.20e-02 +/- 4.1e-04  |  1.13e-02 +/- 1.7e-04  |
> > -test_objective_jac_dshape_current       |   +110.41 +/- 10.48    | +5.20e-02 +/- 4.93e-03 |  9.90e-02 +/- 4.4e-03  |  4.71e-02 +/- 2.2e-03  |
> > -test_objective_jac_atf                  |    +16.31 +/- 3.40     | +3.48e-01 +/- 7.24e-02 |  2.48e+00 +/- 4.3e-02  |  2.13e+00 +/- 5.8e-02  |
> > -test_perturb_1                          |    +22.90 +/- 6.79     | +3.62e+00 +/- 1.07e+00 |  1.94e+01 +/- 9.7e-01  |  1.58e+01 +/- 4.5e-01  |
> > -test_perturb_2                          |    +20.06 +/- 1.61     | +4.19e+00 +/- 3.37e-01 |  2.51e+01 +/- 9.8e-02  |  2.09e+01 +/- 3.2e-01  |
> > -test_proximal_jac_atf                   |     +7.35 +/- 1.07     | +5.28e-01 +/- 7.67e-02 |  7.72e+00 +/- 6.5e-02  |  7.19e+00 +/- 4.1e-02  |
> >  test_proximal_freeb_compute             |     +1.89 +/- 1.07     | +2.34e-03 +/- 1.33e-03 |  1.26e-01 +/- 6.2e-04  |  1.24e-01 +/- 1.2e-03  |
> >  test_proximal_freeb_jac                 |     -0.16 +/- 1.24     | -1.12e-02 +/- 8.82e-02 |  7.08e+00 +/- 6.0e-02  |  7.09e+00 +/- 6.5e-02  |
> > ```
> 
> These are some pretty large slowdowns, mainly for the init due to poincare surface being innitialized (OK), but for the rest likely just due to the addition of Zp Rp Lp attributes which make the jacobian larger for the constraint even if it then gets factored out in the null space. Might be worth thinking of ways to improve this before we merge

I think it's because the benchmarks aren't including the self-consistency constraints that remove the extra DoFs, will fix

YigitElma:
```diff
|             benchmark_name             |         dt(%)          |         dt(s)          |        t_new(s)        |        t_old(s)        | 
| -------------------------------------- | ---------------------- | ---------------------- | ---------------------- | ---------------------- |
 test_build_transform_fft_lowres         |     +1.07 +/- 4.59     | +9.06e-03 +/- 3.89e-02 |  8.58e-01 +/- 2.5e-02  |  8.49e-01 +/- 3.0e-02  |
 test_equilibrium_init_medres            |    +10.08 +/- 3.53     | +6.97e-01 +/- 2.44e-01 |  7.61e+00 +/- 2.2e-01  |  6.91e+00 +/- 1.1e-01  |
-test_equilibrium_init_highres           |    +25.09 +/- 2.59     | +1.95e+00 +/- 2.02e-01 |  9.73e+00 +/- 1.6e-01  |  7.78e+00 +/- 1.3e-01  |
 test_objective_compile_dshape_current   |     -1.35 +/- 2.25     | -5.45e-02 +/- 9.09e-02 |  3.99e+00 +/- 7.2e-02  |  4.05e+00 +/- 5.5e-02  |
 test_objective_compute_dshape_current   |    +13.99 +/- 11.95    | +8.51e-05 +/- 7.27e-05 |  6.93e-04 +/- 6.2e-05  |  6.08e-04 +/- 3.7e-05  |
 test_objective_jac_dshape_current       |     +4.60 +/- 30.55    | +1.12e-03 +/- 7.40e-03 |  2.54e-02 +/- 5.7e-03  |  2.42e-02 +/- 4.7e-03  |
 test_perturb_2                          |     -0.05 +/- 3.83     | -8.97e-03 +/- 7.45e-01 |  1.94e+01 +/- 7.2e-01  |  1.94e+01 +/- 1.7e-01  |
 test_proximal_jac_atf_with_eq_update    |     -0.06 +/- 1.03     | -7.72e-03 +/- 1.32e-01 |  1.27e+01 +/- 5.9e-02  |  1.27e+01 +/- 1.2e-01  |
 test_proximal_freeb_jac                 |     -2.51 +/- 8.19     | -1.21e-01 +/- 3.96e-01 |  4.71e+00 +/- 7.9e-02  |  4.83e+00 +/- 3.9e-01  |
 test_solve_fixed_iter_compiled          |     -1.23 +/- 2.23     | -8.57e-02 +/- 1.55e-01 |  6.86e+00 +/- 1.3e-01  |  6.94e+00 +/- 8.3e-02  |
 test_LinearConstraintProjection_build   |     -1.52 +/- 4.36     | -1.06e-01 +/- 3.05e-01 |  6.89e+00 +/- 2.6e-01  |  6.99e+00 +/- 1.7e-01  |
 test_objective_compute_ripple_bounce1d  |     +2.07 +/- 5.13     | +5.81e-03 +/- 1.44e-02 |  2.87e-01 +/- 9.9e-03  |  2.81e-01 +/- 1.0e-02  |
 test_objective_grad_ripple_bounce1d     |     -0.44 +/- 2.96     | -4.31e-03 +/- 2.92e-02 |  9.81e-01 +/- 1.1e-02  |  9.85e-01 +/- 2.7e-02  |
 test_build_transform_fft_midres         |     -0.65 +/- 2.08     | -5.53e-03 +/- 1.78e-02 |  8.47e-01 +/- 1.1e-02  |  8.53e-01 +/- 1.4e-02  |
 test_build_transform_fft_highres        |     -0.01 +/- 2.83     | -1.27e-04 +/- 3.22e-02 |  1.14e+00 +/- 2.1e-02  |  1.14e+00 +/- 2.4e-02  |
 test_equilibrium_init_lowres            |     +3.55 +/- 3.50     | +2.22e-01 +/- 2.19e-01 |  6.49e+00 +/- 1.5e-01  |  6.27e+00 +/- 1.6e-01  |
 test_objective_compile_atf              |     -0.03 +/- 3.67     | -1.77e-03 +/- 2.13e-01 |  5.81e+00 +/- 1.6e-01  |  5.82e+00 +/- 1.5e-01  |
 test_objective_compute_atf              |     +4.67 +/- 13.28    | +9.39e-05 +/- 2.67e-04 |  2.11e-03 +/- 8.8e-05  |  2.01e-03 +/- 2.5e-04  |
 test_objective_jac_atf                  |     -1.34 +/- 5.42     | -2.18e-02 +/- 8.78e-02 |  1.60e+00 +/- 3.5e-02  |  1.62e+00 +/- 8.1e-02  |
 test_perturb_1                          |     +0.88 +/- 1.93     | +1.30e-01 +/- 2.86e-01 |  1.50e+01 +/- 2.6e-01  |  1.48e+01 +/- 1.2e-01  |
 test_proximal_jac_atf                   |     -0.08 +/- 1.59     | -4.14e-03 +/- 8.35e-02 |  5.26e+00 +/- 5.3e-02  |  5.26e+00 +/- 6.4e-02  |
 test_proximal_freeb_compute             |     +0.48 +/- 2.46     | +6.59e-04 +/- 3.35e-03 |  1.37e-01 +/- 2.8e-03  |  1.36e-01 +/- 1.9e-03  |
 test_solve_fixed_iter                   |     +0.14 +/- 2.17     | +3.38e-02 +/- 5.11e-01 |  2.36e+01 +/- 4.1e-01  |  2.35e+01 +/- 3.0e-01  |
 test_objective_compute_ripple           |     -2.66 +/- 3.14     | -5.67e-03 +/- 6.69e-03 |  2.07e-01 +/- 4.4e-03  |  2.13e-01 +/- 5.1e-03  |
 test_objective_grad_ripple              |     -0.59 +/- 2.54     | -5.21e-03 +/- 2.26e-02 |  8.84e-01 +/- 1.4e-02  |  8.89e-01 +/- 1.8e-02  |
```


Github CI performance can be noisy. When evaluating the benchmarks, developers should take this into account.

YigitElma:
#931 changed the error in factorize_linear_constraints to a warning but still pytest fails some test due to the warning, i.e.
```python
=========================== short test summary info ============================
FAILED tests/test_equilibrium.py::test_poincare_as_bc - UserWarning: 
Not equal to tolerance rtol=3e-14, atol=3e-14
Incompatible constraints detected, cannot satisfy constraint <desc.objectives.linear_objectives.BoundaryRSelfConsistency object at 0x7f5982199220>.
Mismatched elements: 2 / 38 (5.26%)
Max absolute difference: 6.2394534e-14
Max relative difference: inf
 x: array([ 3.063705e-16, -1.948655e-16, -1.146194e-17, -4.111364e-18,
        4.678379e-17,  2.974076e-17, -1.001699e-15,  2.379133e-16,
        1.181742e-16,  3.608323e-17,  7.501427e-16, -1.843074e-16,...
 y: array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
 This may indicate incompatible constraints, or be due to floating point error.
```

Either we should add manual flags like `@pytest.mark.filterwarnings("ignore::UserWarning")` to relevant tests or we can ignore all `UserWarnings` in `setup.cfg`.

I found out that,
```python
with pytest.raises(UserWarning):
...
```
is not a proper solution due to numerical tolerances of different computers. For example, Github machines throw warning but my laptop doesn't. So, running test with `with` clause cause test to fail on my laptop, because the warning is not thrown.

dpanici:
- check master and maybe raise tols for the factorize assertion
- if needed, throw an ignore simple filter on the test that throws this

YigitElma:
I tried to change the order of the arguments,
```python
arg_order = (
            "R_lmn",
            "Z_lmn",
            "L_lmn",
            "p_l",
            "i_l",
            "c_l",
            "Psi",
            "Te_l",
            "ne_l",
            "Ti_l",
            "Zeff_l",
            "a_lmn",
            "Ra_n",
            "Za_n",
            "Rb_lmn",
            "Zb_lmn",
            "Rp_lmn",
            "Zp_lmn",
            "Lp_lmn",
            "I",
            "G",
            "Phi_mn",
        )
```
I put *p_lmn parameters before axis parameters, after axis and to the very end (after Phi_mn). Only putting it to the end changed the results a little bit (in a bad way). I will try some other orders too, but I think we should find some other solution to this, since it affects the accuracy noticably.

YigitElma:
The numerical tolerance issue is probably caused by this. Here is the dimensions of matrix A for both master and Poincare branch,
```python
constraints = maybe_add_self_consistency(eq, constraints)
constraints = ObjectiveFunction(constraints)

objective = ObjectiveFunction(ForceBalance(eq))
objective = LinearConstraintProjection(objective, constraints)
objective.build(verbose=3)
print(objective._A.shape)
```
```bash
master:      (650, 6356)
poincare:   (897, 6603)
```

desc/geometry/surface.py
YigitElma: Ok, with the latest changes, I merged the PoincareSurface and ZernikeRZToroidaSection classes into new one called PoincareRZLSection (the naming is to specify the basis)

dpanici:
change tol to be eps * size(A) or something

YigitElma:
Failing Heliotron vacuum test results,
HELIOTRON_vacuum_output_poincare_vs_master

YigitElma:
This is ready for review. Only the Heliotron vacuum test fails. I guess it is some numerical tolerance issue that is somehow more prominent for that equilibrium. For reference, that test again solves the fixed boundary problem, and the newly added variables `Rp`, `Zp` and `Lp` are supposed to be taken out by the self-consistency constraints. My hunch is that for that specific problem, `factorize_linear_constraints` cannot do a good job of finding a particular solution or null space. I checked the errors in self-consistency constraints and they were around 1e-15 and 1e-16 for some of them. I would assume that should be fine, but somehow the final result is not the same as master even if I increase the `maxiter`. I will be expecting your feedback for the PR and try to find some bugs or improve the accuracy of `factorize_linear_constrains`.

dpanici:
write what the actual test is checking (area diff), check if iota goes thru rational, also can make coiils and check against field tracing, one must be correct and one must be incorrect

desc/equilibrium/utils.py
```diff
@@ -77,9 +77,7 @@ def parse_surface(surface, NFP=1, sym=True, spectral_indexing="ansi"):
     -------
     surface : Surface
         Parsed surface object, either FourierRZToroidalSurface or
-        ZernikeRZToroidalSection.
-    bdry_mode : str
-        Either "lcfs" or "poincare"
+        PoincareRZLSection.
```
f0uriest, line 80: looks like now it only ever returns a FourierRZToroidalSurface

desc/equilibrium/utils.py
```diff
@@ -181,3 +175,31 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
...
+        Parsed Poincare surface object.
+    """
+    if isinstance(xsection, PoincareRZLSection):
+        _xsection = xsection
+    elif isinstance(xsection, (np.ndarray, jnp.ndarray)):
+        # This is temporary, until we have a proper PoincareRZLSection constructor
+        # from input file
+        raise NotImplementedError("PoincareRZLSection from input file not implemented")
+    else:
+        _xsection = PoincareRZLSection(L=surface.M, M=surface.M, sym=sym)
```
f0uriest, line 204: can we make sure that the shapes match as well?

desc/geometry/surface.py
```diff
@@ -758,17 +758,19 @@ def fun_jax(zeta_hat, theta, zeta):
             return offset_surface
 
 
-class ZernikeRZToroidalSection(Surface):
-    """A toroidal cross section represented by a Zernike polynomial in R,Z.
+class PoincareRZLSection(Surface):
```
f0uriest, line 761: This deviates from our existing naming convention of `{Basis}{Parameterization}{Type}`. I think `ZernikeRZToroidalSection` was fine, I thought we were just going to add lambda to that?

tests/conftest.py
```diff
@@ -335,3 +335,34 @@ def VMEC_save(SOLOVEV, tmpdir_factory):
     )
     desc = Dataset(str(SOLOVEV["desc_nc_path"]), mode="r")
     return vmec, desc
+
+
+@pytest.fixture(scope="session")
+def SOLOVEV_Poincare(tmpdir_factory):
```
f0uriest, line 341: what makes this use poincare? I thought we didn't support it from the input file?

tests/test_optimizer.py
```diff
@@ -1116,6 +1119,7 @@ def test_optimize_with_single_constraint():
 
 @pytest.mark.slow
 @pytest.mark.regression
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
f0uriest, line 1122: what warnings is this throwing now?

f0uriest (commented):
A few comments, still need to go through things in more detail.

Also looks like there are still a number of unresolved comments from previous reviews.

tests/conftest.py
YigitElma: You are correct there was nothing that made it Poincare. It was unused, I guess Dario put it to use later just incase we support poincare from input files. I deleted it

YigitElma:
I have finally got to see the actual coverage of this PR 🥲🥲🥲 How I manage to do this a bit controversial but a good example of the problem with null space accuracy!

desc/equilibrium/equilibrium.py
YigitElma: What do you mean by this?

YigitElma:
Do we ever want to fix the cross-section at zeta=Pi/NFP ? I don't test them and those were mainly for fixing 2 cross-sections together. I am pretty sure if I delete them the coverage will pass too. @f0uriest @dpanici

dpanici:
dont do pi/nfp otion for zeta, make issue to make compat function to change zeta=0 definition

desc/objectives/getters.py
```diff
@@ -306,11 +356,45 @@ def _get_NAE_constraints(
...
+def maybe_add_self_consistency(thing, constraints):  # noqa: C901
     """Add self consistency constraints if needed."""
     params = set(unique_list(flatten_list(thing.optimizable_params))[0])
 
-    # Equilibrium
+    # check which boundary mode we are in
+    if (
+        is_any_instance(constraints, FixBoundaryR)
+        and is_any_instance(constraints, FixSectionR)
+        and is_any_instance(constraints, BoundaryRSelfConsistency)
```
dpanici, line 367: what if someone has bdryR selv consistency and FixBoundaryR? techincally that is an allowed combination, though it is rarely manually passed in

desc/objectives/getters.py
```diff
@@ -306,11 +356,45 @@ def _get_NAE_constraints(
...
+
+    if is_any_instance(constraints, FixBoundaryR) and is_any_instance(
+        constraints, BoundaryRSelfConsistency
+    ):
+        mode = "lcfs"
+    elif is_any_instance(constraints, FixSectionR) and is_any_instance(
+        constraints, SectionRSelfConsistency
+    ):
+        mode = "poincare"
+    elif is_any_instance(constraints, FixBoundaryR) and not is_any_instance(
```
dpanici, line 383: isn't this sort of unnecessary since you will eror above if FixbdryR and FixSectionR are in cnostraints?

desc/objectives/getters.py
```diff
@@ -306,11 +356,45 @@ def _get_NAE_constraints(
...
+        mode = "lcfs"
+    elif is_any_instance(constraints, FixSectionR) and not is_any_instance(
+        constraints, FixBoundaryR
+    ):
+        mode = "poincare"
+    elif not is_any_instance(constraints, FixBoundaryR) and not is_any_instance(
+        constraints, FixSectionR
+    ):
+        mode = "lcfs"
+    else:
```
dpanici, line 395: you could simplify this section by just having two checks, 

```python
if is_any_instance(constraints, FixBoundaryR) or is_any_instance(constraints, BoundaryRSelfConsistency):
    mode="lcfs"
elif  is_any_instance(constraints, FixSectionR) or is_any_instance(constraints, SectionRSelfConsistency):
    mode="poincare"
else:
    mode="other"
```

dpanici: since any of the other combinations would error out in the first if statement you have

desc/objectives/utils.py
```diff
@@ -224,8 +224,9 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
         # else check with tighter tols and throw an error, these tolerances
         # could be tripped due to just numerical roundoff or poor scaling between
         # constraints, so don't want to error out but we do want to warn the user.
-        atol = 3e-14
-        rtol = 3e-14
+        size = max(A.shape)
+        atol = 1e-15 * size
```
dpanici, line 228: does this help a lot? if so we should put it in a separate PR instead of this on

desc/objectives/linear_objectives.py
```diff
@@ -774,6 +1019,249 @@ def build(self, use_jit=False, verbose=1):
...
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    target : {float, ndarray}, optional
+        Target value(s) of the objective. Only used if bounds is None.
+        Must be broadcastable to Objective.dim_f. Defaults to ``target=eq.Rb_lmn``.
+    bounds : tuple of {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
+        Both bounds must be broadcastable to to Objective.dim_f.
+        Defaults to ``target=eq.Rb_lmn``.
```
dpanici, line 1197: wrong default

desc/objectives/linear_objectives.py
```diff
@@ -774,6 +1019,249 @@ def build(self, use_jit=False, verbose=1):
...
```
dpanici, line 1116: wrong default

desc/objectives/linear_objectives.py
```diff
@@ -774,6 +1019,249 @@ def build(self, use_jit=False, verbose=1):
...
```
dpanici, line 1035: wrong default

desc/equilibrium/equilibrium.py
```diff
@@ -463,7 +504,7 @@ def __repr__(self):
             )
         )
 
-    def set_initial_guess(self, *args, ensure_nested=True):
+    def set_initial_guess(self, *args, ensure_nested=True, lcfs_surface=True):
         """Set the initial guess for the flux surfaces, eg R_lmn, Z_lmn, L_lmn.
```
dpanici, line 508: should  have some comment for what happens if lcfs_surface=True

desc/equilibrium/equilibrium.py
```diff
@@ -698,8 +742,17 @@ def get_surface_at(self, rho=None, theta=None, zeta=None):
 
         if zeta is not None:
             assert (zeta >= 0) and (zeta <= 2 * np.pi)
-            surface = ZernikeRZToroidalSection(sym=self.sym, zeta=zeta)
+            warnif(
+                zeta != 0,
+                UserWarning,
+                "The lambda value of the returned surface will not be correct "
```
dpanici, line 748: why does this happen?

desc/objectives/utils.py
YigitElma: I don't remember but we discussed this in one of the dev meetings. I am down for putting in a new PR.

desc/geometry/surface.py
```diff
@@ -841,8 +845,10 @@ def __init__(
         self,
         R_lmn=None,
         Z_lmn=None,
+        L_lmn=None,
```
f0uriest, line 848: move the new args to the end for backwards compatibility

desc/objectives/getters.py
```diff
@@ -150,6 +160,46 @@ def get_fixed_boundary_constraints(eq, profiles=True, normalize=True):
     return constraints
 
 
+def get_fixed_xsection_constraints(
+    eq,
+    profiles=True,
+    normalize=True,
+    fix_lambda=True,
```
f0uriest, line 167: include this in docstring

desc/objectives/linear_objectives.py
```diff
@@ -436,6 +425,262 @@ def compute(self, params, constants=None):
         return jnp.dot(self._A, params["Z_lmn"]) - params["Zb_lmn"]
 
 
+class SectionRSelfConsistency(_Objective):
+    """Ensure that the boundary and interior surfaces are self-consistent.
```
f0uriest, line 429: should mention section here to distinguish it from the lcfs boundary

desc/objectives/linear_objectives.py
```diff
@@ -436,6 +425,262 @@ def compute(self, params, constants=None):
...
+
+    Note: this constraint is automatically applied when needed, and does not need to be
+    included by the user.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    surface_label : float, optional
+        Surface to enforce boundary conditions on. Defaults to Equilibrium.surface.rho
```
f0uriest, line 439: change to zeta? or remove if unused

desc/objectives/linear_objectives.py
```diff
@@ -436,6 +425,262 @@ def compute(self, params, constants=None):
...
+                )
+            )
+            self._A[i, j] = 1
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params, constants=None):
+        """Compute boundary R self-consistency errors.
+
+        IE, the mismatch between the Fourier-Zernike basis evaluated at rho=1 and the
```
f0uriest, line 502: update this

desc/objectives/linear_objectives.py
```diff
@@ -436,6 +425,262 @@ def compute(self, params, constants=None):
...
+
+    Note: this constraint is automatically applied when needed, and does not need to be
+    included by the user.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    surface_label : float, optional
+        Surface to enforce boundary conditions on. Defaults to Equilibrium.surface.rho
```
f0uriest, line 533: update this

desc/objectives/linear_objectives.py
```diff
@@ -436,6 +425,262 @@ def compute(self, params, constants=None):
...
+                    (eq.Z_basis.modes[:, :-1] == [l, m]).all(axis=1),
+                    eq.Z_basis.modes[:, 2] >= 0,
+                )
+            )
+            self._A[i, j] = 1
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params, constants=None):
+        """Compute deviation from desired boundary."""
```
f0uriest, line 596: make this a full docstring

desc/objectives/linear_objectives.py
```diff
@@ -436,6 +425,262 @@ def compute(self, params, constants=None):
...
+        return jnp.dot(self._A, params["Z_lmn"]) - params["Zp_lmn"]
+
+
+class SectionLambdaSelfConsistency(_Objective):
+    """Enforces lambda values at cross-section.
```
f0uriest, line 601: enforces self consistency, not values

desc/objectives/linear_objectives.py
```diff
@@ -436,6 +425,262 @@ def compute(self, params, constants=None):
...
+
+    Parameters
+    ----------
+    eq : Equilibrium, optional
+        Equilibrium that will be optimized to satisfy the Objective.
+    target : float, ndarray, optional
```
f0uriest, line 607: many of these args are unused, and its missing the surface_label arg

desc/objectives/linear_objectives.py
```diff
@@ -436,6 +425,262 @@ def compute(self, params, constants=None):
...
+            )
+            self._A[i, j] = 1
+
+        if self.target is not None:
+            self._dim_f = self._A.shape[0]
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params, constants=None):
+        """Compute deviation from desired boundary."""
```
f0uriest, line 680: full docstring here

docs/notebooks/tutorials/basic_optimization.ipynb
f0uriest, line 1: these changes look like they may have been committed accidentally?

desc/equilibrium/equilibrium.py
f0uriest: I'd recommend doing the loop with R and Z, then we can get rid of `get_basis_poincare` entirely i think

f0uriest (dismissed):
still need to go through a lot of it, here are some short term comments

YigitElma:
@dpanici @f0uriest I made the `maybe_add_self_consistency` function cleaner. For your point on why do we need to check extra cases, the problem starts at this line https://github.com/PlasmaControl/DESC/blob/9ef0b386c2bbffa2fc52ee9dfa31da9f7faa7ee0/desc/optimize/_constraint_wrappers.py#L580

This is the set of constraints used for `ProximalProjection`s equilibrium update. We add self consistency constraints at the beginning. Then, we pass these constraints to `perturb` and `solve` which again call `maybe_add_self_consistency` function. This is the main problem why we need to check. My previous implementation was really bad. Now it is a bit cleaner.

The idea here is to add self consistency only if there is a corresponding fixed constraint, AND there is no opposing self consistency constraint. For example,
{FixBoundaryR} -> add BoundaryRSelfConsistency and FixSectionR
{FixBoundaryR, BoundaryRSelfConsistency, FixSectionR} -> don't add anything

Similarly for FixSectionR constraints.

YigitElma:
By the way, I am aware of the full red benchmarks, and don't want to affect 99% of the users which won't use this. I will take a closer look at it once the main code works as expected.

YigitElma:
I think these are the all related cases.

YigitElma:
> We have added lots of stuff since the first time we noticed things were different here, can we check now again to see if the equilibrium solve without any hacky removal of constraints results in the same answer?

I think it is better than it used to be but the `test_HELITRON_vac_results` still fails. If I remember correctly, that was the most problematic one for some reason. I tried to change tolerances but putting lower tolerances cause the flux surfaces to be not nested after the first step of continuation and perturb. And it quits. I can achieve the intended result like this tho,

```python
args = ["../tests/inputs/HELIOTRON_vacuum", "-vvv", "-o", "heli_vac.h5"]
main(args)

vmec_path = "..//tests//inputs//wout_HELIOTRON_vacuum.nc"
eq = Equilibrium.load("heli_vac.h5")[-1]
VMECIO.plot_vmec_comparison(eq, vmec_path);
plt.savefig("vmec_comp.png", dpi=1000)

eq.solve(maxiter=100, ftol=2e-5, gtol=0, verbose=3);

rho_err, theta_err = area_difference_vmec(eq, vmec_path)
np.testing.assert_allclose(rho_err.mean(), 0, atol=1e-2)
np.testing.assert_allclose(theta_err.mean(), 0, atol=2e-2)
curr = eq.get_profile("current")
np.testing.assert_allclose(curr(np.linspace(0, 1, 20)), 0, atol=1e-8)
```

YigitElma:
I changed the hacky way with another approach where I declare some parameters non-optimizable. To do so, I added a new function inside `desc.optimizable` called `make_nonoptimizable` and a setter for `cls.optimizable_params`. If this works, it should solve the speed issue.

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ New Features
 - Changes hessian computation to use chunked ``jacfwd`` and ``jacrev``, allowing ``jac_chunk_size`` to now reduce hessian memory usage as well.
 - Adds an option to ``VMECIO.save`` to specify the grid resolution in real space.
 - Adds a new objective ``desc.objectives.CoilIntegratedCurvature`` for targeting convex coils.
+- Adds ``Rp_lmn``, ``Zp_lmn`` and ``Lp_lmn`` atributes to the ``Equilibrium`` class. They are optimizable parameters will be used if users tries to solve equilibrium problem with fixed Poincare section boundary condition.
+- Adds option to solve equilibrium problem using fixed Poincare cross-section boundary condition. Adds a new helper function to ``desc.objectives.getters`` called ``get_fixed_xsection_constraints`` which will automatically adds the required ``FixSectionR``, ``FixSectionZ`` adn optionally ``FixSectionLambda`` (if the argument ``fix_lambda=True`` default, it doesn't add it if set to ``fix_lambda=False``).
```
YigitElma, line 22:
```suggestion
- Adds option to solve equilibrium problem using fixed Poincare cross-section boundary condition. Adds a new helper function to ``desc.objectives.getters`` called ``get_fixed_xsection_constraints`` which will automatically adds the required ``FixSectionR``, ``FixSectionZ``and optionally ``FixSectionLambda`` (if the argument ``fix_lambda=True`` default, it doesn't add it if set to ``fix_lambda=False``).
```

desc/continuation.py
```diff
@@ -693,6 +695,8 @@ def solve_continuation(  # noqa: C901
         NotImplementedError,
         "Continuation method with anisotropic pressure is not currently supported",
     )
+    # TODO: add warning for cases with Poincare BC. This needs a better
```
dpanici, line 698: is this done above?

YigitElma:
@f0uriest @dpanici, one final offer about this. I am pretty convinced that a user cannot solve or optimize an equilibrium (with cold-start) with just Poincare (minor changes in the section cause big changes in the overall equilibrium). Given this, I don't think we need to have `Rp_lmn` etc as an optimizable variable. What I suggest is to enforce the constraint by computing the Rp_lmn values during the build method of the objective. This way you can prevent the cross-section from changing during the optimization. In this PR, I had added some logic to different parts of the code like perturb etc, I can remove all of them if you are ok to this? I assume we do something similar in NAE.

desc/continuation.py
YigitElma: that is a separate part of the code (300 lines above), but I don't think this is needed anymore, I deleted this

desc/equilibrium/equilibrium.py
YigitElma: Every zeta cross-section should be possible now

YigitElma:
Couple notes on this. 
- After some local changes, I realized that during stage 1 optimization, having `Lp_lmn` doesn't make sense. Only having `Rp_lmn` and `Zp_lmn` as optimizable then solving for the corresponding lambda gives much better results.
- LCFS boundary optimization usually fixes the Rb00 mode and target aspect ratio which basically enforces a rough estimate for the volume. For poincare variables, there is no immediate linear constraint, because Rp00 only fixes the cross sections average R position not the whole major radius. So, instead I put volume and aspect ratio objectives, and it seems to work fine.
- For the eq.solve(), I still think having option to fix lambda or not is useful. I might change the default to not fixed.

YigitElma:
The state of the code that is used for the results of the possible paper is tagged with name `poincare-bc-paper`. Closing this PR since making both these options (lcfs and section optimization) in the current master is an unnecessary burden to maintain. Interested people can use the tagged commit and build on it.

PR #786, opened 2023-12-04
Rc/nangrad

### LINKED ISSUE #738: Check all objectives for NaN gradients

f0uriest:
Fixes to prevent NaNs in reverse mode differentiation for qtys at the axis.

The issue is that stuff like this
```python
jnp.where(b==0, a_r/b_r, a/b)
```
is not reverse mode differentiable, since you end up propagating a NaN adjoint variable.

This switches stuff like the above to be more like
```python
mask = b==0
jnp.where(mask, a_r, a)/jnp.where(mask, b_r, b)  
```
Side effect: some qtys that are sort of undefined for surface classes that used to evaluates to NaN now evaluate to zero.

Resolves #738 (again)

### DIFF
--- a/tests/test_axis_limits.py
+++ b/tests/test_axis_limits.py
@@ -5,6 +5,9 @@ If a new quantity is added to the compute functions whose limit is not finite
 If the limit has yet to be derived, add it to the ``not_implemented_limits`` set.
 """
 
+import functools
+import inspect
+
 import numpy as np
 import pytest
 
@@ -13,6 +16,7 @@ from desc.compute.utils import dot, surface_integrals_map
 from desc.equilibrium import Equilibrium
 from desc.examples import get
 from desc.grid import LinearGrid
+from desc.objectives import GenericObjective, ObjectiveFunction
 
 # Unless mentioned in the source code of the compute function, the assumptions
 # made to compute the magnetic axis limit can be reduced to assuming that these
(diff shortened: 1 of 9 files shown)


ddudt (approved):

tests/test_axis_limits.py
```diff
@@ -332,3 +336,57 @@ def test(eq):
...
+
+@pytest.mark.parametrize("name", _reverse_mode_unsafe_names())
+def test_reverse_mode_ad_axis(name):
+    """Asserts that the rho=0 axis limits are reverse mode differentiable."""
+    eq = get("ESTELL")
+    grid = LinearGrid(rho=0.0, M=2, N=2, NFP=eq.NFP, sym=eq.sym)
+    eq.change_resolution(2, 2, 2, 4, 4, 4)
+
+    obj = ObjectiveFunction(GenericObjective(name, eq, grid=grid), verbose=0)
+    obj.build(verbose=0)
```
rahulgaur104, line 390: We only need to check if the reverse mode diff works for all quantities that are using safediv.  When you build a GenericObjective, you are not specifiying any specific quantity to compute. Is jax automatically building a graph across all the registered quantities?

rahulgaur104 (approved):

tests/test_axis_limits.py
f0uriest: the `name` here is a parameter of the test, generated by the helper function above. We basically check every quantity that uses the `replace_at_axis` method, which usually means we're dividing by zero and then overwriting it.

PR #777, opened 2023-11-28
Bootstrap

ddudt:
Tutorial and improved functionality for bootstrap current optimization

PR #789, opened 2023-12-07
enforce O(r^2) in 'current Redl'

ddudt:
This is an update to PR #777. It ensures that the `"current Redl"` quantity has $\mathcal{O}(\rho^2)$ behavior near the magnetic axis, and this is achieved by changing `polyfit` of the current derivative to a manual polynomial fit without the constant term.


tests/test_bootstrap.py
```diff
@@ -1656,10 +1654,10 @@ def test_bootstrap_optimization_comparison_qa():
     data2 = eq2.compute(["<J*B> Redl", "<J*B>"], grid)
 
     np.testing.assert_allclose(
-        grid.compress(data1["<J*B>"]), grid.compress(data1["<J*B> Redl"]), rtol=3e-2
+        grid.compress(data1["<J*B>"]), grid.compress(data1["<J*B> Redl"]), rtol=4e-2
```
dpanici, line 1657: don't reduce tolerance if unneeded

dpanici (changes requested):
test tolerance comment, don't reduce unless needed

dpanici (approved):

f0uriest (approved):

PR #792, opened 2023-12-07
LinearObjectiveFromUser

ddudt:
Adds a new class `LinearObjectiveFromUser` which is very similar to `ObjectiveFromUser` but meant for linear functions. This allows for custom constraints on parameters of optimizable things. 

For example, this can be used to constrain the mirror ratio in the omnigenity PR. 

I did not change `GenericObjective` or `ObjectiveFromUser`, only reorganized the order of the classes in that file.


desc/objectives/_generic.py
```diff
@@ -190,22 +153,155 @@ def compute(self, params, constants=None):
...
+        self._dim_f = self._dim_f = jax.eval_shape(self._fun, thing.params_dict).size
+
+        # check that fun is linear
+        J1 = jax.jacrev(self._fun)(thing.params_dict)
+        params = thing.params_dict.copy()
+        for key, value in params.items():
+            params[key] = value + np.random.rand(value.size) * 10
+        J2 = jax.jacrev(self._fun)(params)
+        for key in J1.keys():
+            np.testing.assert_allclose(
```
f0uriest, line 258: i think this could be `assert np.all(A==B), "message"`

desc/objectives/_generic.py
```diff
@@ -190,22 +153,155 @@ def compute(self, params, constants=None):
...
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        thing = self.things[0]
+
+        import jax
+
+        self._dim_f = self._dim_f = jax.eval_shape(self._fun, thing.params_dict).size
```
f0uriest, line 249: duplicated statement

desc/objectives/_generic.py
```diff
@@ -190,22 +153,155 @@ def compute(self, params, constants=None):
...
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params, constants=None):
+        """Compute fixed degree of freedom errors.
+
+        Parameters
+        ----------
+        params : dict
+            Dictionary of equilibrium degrees of freedom, eg Equilibrium.params_dict
```
f0uriest, line 270: equilibrium -> thing

f0uriest (changes requested):
Few typos

f0uriest (approved):

PR #800, opened 2023-12-10
Make plotting functions work for surface and curve objects

### LINKED ISSUE #764: Make plotting routines work with objects besides Equilibrium
IE, 
```python
plot_2d(surf, "H_rho")
plot_1d(curve, "torsion")
plot_boundaries([eq, surface])
plot_boozer_surface(omnigenous_field)
```

Some of these may already work, in which case we just need to document them and maybe include examples.

f0uriest:
Resolves #764


desc/plotting.py
```diff
@@ -273,7 +274,8 @@ def _compute(eq, name, grid, component=None, reshape=True):
         Computed quantity.
 
     """
-    if name not in data_index["desc.equilibrium.equilibrium.Equilibrium"]:
+    parameterization = _parse_parameterization(eq)
+    if name not in data_index[parameterization]:
         raise ValueError("Unrecognized value '{}'.".format(name))
```
dpanici, line 279: can we add to this something like `f"for object of type {parametrization}"`

dpanici (approved):

dpanici (changes requested):
Add more info about the object to the error message when an incorrect compute quantity is entered in

dpanici (approved):

PR #803, opened 2023-12-12
Update W7-X Example Input File

dpanici:
- fix W7X input file to use auto continuation (as manual continuation method gave unnested surfaces), update notebook to refer to NCSX input file for example of manual continuation steps
- allow nested surface optimization when manual continuation is specified in input file


dpanici:
Interestingly, I uploaded one master pickle with the computed data from when I was on a della GPU node, and then I tested locally against it and there are some of the higher order derivative quantities which vary by an rtol of about `1e-10`, so something to be wary of when testing on GPU:
[test_against_data_from_gpu.txt](https://github.com/PlasmaControl/DESC/files/13671230/test_against_data_from_gpu.txt)

rahulgaur104 (approved):

PR #805, opened 2023-12-12
Rc/proximal

f0uriest:


desc/objectives/objective_funs.py
```diff
@@ -38,14 +38,16 @@ class ObjectiveFunction(IOAble):
         Jacobian column by column. Generally the slowest, but most memory efficient.
         "auto" defaults to "batched" if all sub-objectives are set to "fwd",
         otherwise "blocked".
-    verbose : int, optional
```
f0uriest, line 41: verbose was already unused here so figured we might as well remove it

dpanici:
What sort of memory/time difference do you see with this?

f0uriest:
> What sort of memory/time difference do you see with this?

I added a new benchmark and for a "standard" QS optimization its about 35% faster without any linear constraints - it should be even faster with linear constraints since we now use JVPs for that as well.

Once free boundary is merged in I'll add another benchmark for that, we should see a much bigger speedup for cases where the objective is much more expensive to evaluate than the constraint.

desc/objectives/objective_funs.py
```diff
@@ -980,6 +1012,54 @@ def jac_unscaled(self, *args, **kwargs):
         """Compute Jacobian matrix of vector form of the objective wrt x, unweighted."""
         return self._jac_unscaled(*args, **kwargs)
 
+    def jvp_scaled(self, v, x, constants=None):
+        """Compute Jacobian-vector product of the objective function.
+
```
rahulgaur104, line 1017: Not necessary but it would be super helpful to add the matrix product operation in the docstring. Something like (A B) C^T or A (B C^T) and add a few words about why one is faster etc.

desc/optimize/_constraint_wrappers.py
```diff
@@ -11,9 +11,9 @@
     maybe_add_self_consistency,
 )
 from desc.objectives.utils import factorize_linear_constraints
-from desc.utils import Timer, get_instance
+from desc.utils import Timer, get_instance, setdefault
 
-from .utils import compute_jac_scale, f_where_x
+from .utils import f_where_x
```
rahulgaur104, line 16: What is f_where_x?

desc/optimize/_constraint_wrappers.py
```diff
@@ -98,26 +99,16 @@ def build(self, use_jit=None, verbose=1):
         self._dim_x = self._objective.dim_x
         self._dim_x_reduced = self._Z.shape[1]
 
+        # equivalent matrix for A[unfixed_idx]@Z == A@unfixed_idx_mat
+        self._unfixed_idx_mat = (
```
rahulgaur104, line 103: So what is _unfixed_idx_mat in the above equation. Is it like a mask that picks up certain indices in A?

desc/optimize/_constraint_wrappers.py
```diff
@@ -292,8 +283,14 @@ def jac_unscaled(self, x_reduced, constants):
             Jacobian matrix.
         """
         x = self.recover(x_reduced)
-        df = self._objective.jac_unscaled(x, constants)
-        return df[:, self._unfixed_idx] @ self._Z
+        if self._objective._deriv_mode == "blocked":
```
rahulgaur104, line 286: What is deriv_mode == "blocked"? How is it dfferent from fwd and rev?

desc/optimize/_constraint_wrappers.py
```diff
@@ -846,35 +858,118 @@ def hess(self, x, constants=None):
...
+        sf += sf[-1]  # add a tiny bit of regularization
+        sfi = jnp.where(sf < cutoff * sf[0], 0, 1 / sf)
+        Fxh_inv = vtf.T @ (sfi[..., None] * uf.T)
+        return Fxh_inv @ Fc
+
+    def _jvp(self, v, xf, xg, constants, op):
+        # we're replacing stuff like this with jvps
+        # Fx_reduced = Fx[:, unfixed_idx] @ Z               # noqa: E800
+        # Gx_reduced = Gx[:, unfixed_idx] @ Z               # noqa: E800
+        # Fc = Fx @ dxdc @ v                                # noqa: E800
```
rahulgaur104, line 921: What is the order of products here? Maybe put parentheses?

tests/test_optimizer.py
```diff
@@ -1127,3 +1128,198 @@ def test_optimize_with_single_constraint():
 
     # test depends on verbose > 0
     optimizer.optimize(eq, objective=objectective, constraints=constraints, verbose=2)
+
+
+@pytest.mark.unit
+def test_proximal_jacobian():
+    """Test that JVPs and manual concatenation give the same result as full jac."""
```
rahulgaur104, line 1135: Can you add some info about speed and memory in the docstring or somewhere else?

rahulgaur104 (changes requested):

desc/objectives/objective_funs.py
```diff
@@ -387,7 +390,7 @@ def x(self, *things):
         """Return the full state vector from the Optimizable objects things."""
         # TODO: also check resolution etc?
         things = things or self.things
-        assert all([type(t1) == type(t2) for t1, t2 in zip(things, self.things)])
+        assert all([type(t1) is type(t2) for t1, t2 in zip(things, self.things)])
```
dpanici, line 393: is this to allow subclasses to match against their parent class types?

desc/optimize/_constraint_wrappers.py
dpanici: https://github.com/PlasmaControl/DESC/blob/16c89acc531655e225954cea1788f8f436838bae/desc/optimize/utils.py#L499

desc/objectives/objective_funs.py
```diff
@@ -454,22 +465,30 @@ def jvp_unscaled(self, v, x):
...
         """
-        if not isinstance(v, tuple):
-            v = (v,)
+        v = v if isinstance(v, (tuple, list)) else (v,)
+
+        compute_unscaled = lambda x: self.compute_unscaled(x, constants)
         if len(v) == 1:
-            return Derivative.compute_jvp(self.compute_unscaled, 0, v[0], x)
+            jvpfun = lambda dx: Derivative.compute_jvp(compute_unscaled, 0, dx, x)
+            return jnp.vectorize(jvpfun, signature="(n)->(k)")(v[0])
```
dpanici, line 477: "jvp methods for objectives are now vectorized over the tangents for propagating derivatives" what does this mean? as far as propagating derivatives?

desc/objectives/objective_funs.py
```diff
@@ -980,6 +1012,54 @@ def jac_unscaled(self, *args, **kwargs):
...
+        Parameters
+        ----------
+        v : tuple of ndarray
+            Vectors to right-multiply the Jacobian by.
+        x : ndarray
+            Optimization variables.
+        constants : list
+            Constant parameters passed to sub-objectives.
+
+        """
```
dpanici, line 1029: probably a stupid question but there are jvp_scaled methods for _Objective now and the one already in ObjectiveFunction, is the idea that if you want a single objective's derivative you'd call the _Objective method, but the ObjectiveFunction is still separate and the jvp computed using its own compute_scaled method?

tests/test_objective_funs.py
```diff
@@ -271,7 +271,7 @@ def test(eq):
             """Ensure compilation without any errors from JAX, related to issue #625."""
             obj = ObjectiveFunction(QuasisymmetryBoozer(eq=eq))
             obj.build()
-            obj.compile()
+            obj.compile("all")
```
dpanici, line 274: why the change to all here?

desc/optimize/_constraint_wrappers.py
dpanici: `A[unfixed_idx]` are the parts of the linear constraint matrix `A` that are not just a row of all zeros with a single 1 (which are the `fixed_idx`, the indices of `x` which are just directly fixed by constraints and therefore can be just put in the particular solution and removed from the overall `x` vector, I think)

so we use `A[unfixed_idx]` a lot as it then is used to find the reduced optimization vector, I think this varibale here is just a convenience definition as we use it a lot

desc/optimize/_constraint_wrappers.py
dpanici: https://github.com/PlasmaControl/DESC/blob/d4ab2f58ff16c388a432a859016ba1c4122ed31e/desc/objectives/objective_funs.py#L34

desc/objectives/objective_funs.py
f0uriest: This was basically just to shut up my linter. Using `is` is the more pythonic/correct way to check types like this, since types are singleton objects so comparing identity is the same as equality

desc/objectives/objective_funs.py
f0uriest: This is what allows us to use these methods in `ProximalProjection` and `LinearConstraintProjection`, where they're used instead of calculating full jacobian matrices and then doing a matmul, that's all I meant.

desc/optimize/_constraint_wrappers.py
dpanici: though with this change is batched still the most memory intensive? as now we can do the jvp's when batched is the method, while blocked is still making the whole jacobian, so it strikes me as that might be more intensive

tests/test_objective_funs.py
f0uriest: I removed the default explicit compile step so this is just to keep coverage the same (and to make sure all derivatives work)

desc/optimize/_constraint_wrappers.py
f0uriest: blocked still uses JVPs, but does them on the individual objectives rather than on the whole. The only exception is objectives that specify reverse mode, in which case we calculate that full jacobian. This was done because some objectives may _only_ support reverse mode (like ODE integration). But in those cases its usually a scalar or very low dimensional objective so doing the full reverse mode jacobian is pretty cheap (and possibly cheaper to doing multiple jvps, depending on how many variables we're optimizing)

desc/optimize/_constraint_wrappers.py
f0uriest: matmul is associative so its whatever order the compiler decides

desc/optimize/_constraint_wrappers.py
```diff
@@ -311,10 +308,55 @@ def jac_scaled(self, x_reduced, constants=None):
...
+        return df.T
+
+    def jvp_scaled(self, v, x_reduced, constants=None):
+        """Compute Jacobian-vector product of the objective function.
+
+        Uses the scaled form of the objective.
+
+        Parameters
+        ----------
+        v : tuple of ndarray
```
dpanici, line 326: v itself here is a full length vector right, not the reduced length like x_reduced? Would it maybe be helpful to mention sizes or would that get too confusing with the use of this for the proximal boundary stuff too?

desc/optimize/_constraint_wrappers.py
```diff
@@ -502,6 +549,26 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
             self._scalar = False
 
         self._set_eq_state_vector()
+
+        # map from eq c to full c
+        self._dimc_per_thing = [t.dim_x for t in self.things]
```
dpanici, line 554: This is just saying that every optimizable param of things other than eqs is fair game to be changed, but for the equilibrium the inputs c are actually different than dim_x ?

dpanici (approved):

desc/objectives/objective_funs.py
f0uriest: yeah exactly. The `_Objective` jvp methods get used in the proximal thing when the `ObjectiveFunction` deriv mode is blocked.

desc/optimize/_constraint_wrappers.py
f0uriest: no, v should be the same size as x_reduced, we multiply by the `unfixed_idx_mat` to make it the correct size against the full x

desc/optimize/_constraint_wrappers.py
f0uriest: yeah the equilibrium only considers the boundary dofs, other objects have everything counted

rahulgaur104 (dismissed):

dpanici (approved):

rahulgaur104 (approved):

PR #813, opened 2023-12-30
change assert statement for I,G so it can accept size 1 0D numpy arrays correctly

dpanici:


f0uriest:
I remember there was a reason I didn't do this to begin with, something with broadcasting weird. But I think as long as it's stored as a float and not a 1d array it should work

desc/magnetic_fields.py
```diff
@@ -389,7 +389,9 @@ class ScaledMagneticField(_MagneticField, Optimizable):
     _io_attrs = _MagneticField._io_attrs_ + ["_field", "_scalar"]
 
     def __init__(self, scale, field):
-        assert np.isscalar(scale) or len(scale) == 1, "scale must be a scalar value"
+        assert (
+            np.isscalar(scale) or np.asarray(scale).size == 1
+        ), "scale must be a scalar value"
```
f0uriest, line 394: Should also cast scale to a float I think?

f0uriest (changes requested):

f0uriest (approved):

PR #814, opened 2024-01-04
VMECIO.write_vmec_input

ddudt:
Adds a class method to `VMECIO` to write a VMEC input file. This is useful for other codes that take more than the wout file as an input, and re-running VMEC will generate all the files.


ddudt:
In a manual test I performed, the generated input file ran properly in VMEC and converged to the expected solution from DESC. I'm not sure what the best automatic test for this is, since we don't want to be calling VMEC in our CI tests.

desc/input_reader.py
```diff
@@ -876,7 +876,7 @@ def desc_output_to_input(  # noqa: C901 - fxn too complex
         current = grid.compress(eq.compute("current", grid=grid)["current"])
 
         pres_profile = PowerSeriesProfile.from_values(
-            rho, pressure, order=10, sym=False
+            rho, pressure, order=eq.L, sym=False
```
dpanici, line 879: don't refit if already a powerseries (same for curent and iota)

ddudt:
Code coverage on the patch is low because there are a lot of different equilibrium options (iota vs current profile, pressure vs kinetic profiles, stellarator symmetry, etc.) Since we aren't publicizing this new feature to the broader public I am comfortable with having <95% coverage.

desc/input_reader.py
```diff
@@ -867,23 +867,35 @@ def desc_output_to_input(  # noqa: C901 - fxn too complex
...
-            rho, pressure, order=10, sym=False
-        ).params
-        iota_profile = PowerSeriesProfile.from_values(
-            rho, iota, order=eq.L, sym=False
-        ).params
-        curr_profile = PowerSeriesProfile.from_values(
-            rho, current, order=eq.L, sym=False
-        ).params
+        if isinstance(eq.pressure, PowerSeriesProfile):
+            eq.pressure.change_resolution(eq.L)
```
f0uriest, line 874: What if the profile already has higher resolution?

desc/vmec.py
```diff
@@ -1338,6 +1339,209 @@ def read_vmec_output(cls, fname):
...
+        else:
+            current = eq.current
+            f.write("  NCURR = 1\n")  # current profile specified
+            f.write("  CURTOR = {:+14.8E}\n".format(float(current(1))))  # AC scale
+            if isinstance(current, PowerSeriesProfile) and current.sym:
+                f.write("  AC =")  # current power series coefficients
+                for ac in current.params:
+                    f.write(" {:+14.8E}".format(ac))
+                f.write("\n  PCURR_TYPE = 'power_series_I'\n")
+            else:
```
f0uriest, line 1460: Probably shouldn't use bare else here to account for possibly different profile types in the future

f0uriest (changes requested):

tests/inputs/input.DSHAPE_desc
```diff
@@ -8,11 +8,11 @@ NFP =   1
 Psi = 1.00000000
 
 # spectral resolution
-L_rad = 12
```
dpanici, line 11: why the changes to these resolutions?

tests/test_vmec.py
```diff
@@ -913,3 +915,47 @@ def test_vmec_boundary_subspace(DummyStellarator):
...
+    np.testing.assert_allclose(eq0.L, eq1.L)
+    np.testing.assert_allclose(eq0.M, eq1.M)
+    np.testing.assert_allclose(eq0.N, eq1.N)
+    np.testing.assert_allclose(eq0.NFP, eq1.NFP)
+    np.testing.assert_allclose(eq0.Psi, eq1.Psi)
+    np.testing.assert_allclose(eq0.p_l, eq1.p_l)
+    np.testing.assert_allclose(eq0.i_l, eq1.i_l)
+    np.testing.assert_allclose(eq0.Rb_lmn, eq1.Rb_lmn)
+    np.testing.assert_allclose(eq0.Zb_lmn, eq1.Zb_lmn)
+    np.testing.assert_allclose(W0, W1)
```
dpanici, line 961: this is the only part that is a "solution", the rest are just inputs. If the solution is being tested, testing R_lmn and Z_lmn and L_lmn or something would be appropriate

tests/inputs/input.DSHAPE_desc
ddudt: This change was needed so that `tests/inputs/input.DSHAPE` is consistent with `tests/inputs/input.DHSAPE_desc`. The VMEC convention is such that `MPOL = eq.M + 1`, so I decreased the resolution in the DESC input file to match the resolution of the VMEC input file.

tests/test_vmec.py
ddudt: I initially tried comparing the coefficients R_lmn, Z_lmn, and L_lmn but they did not match well. I think this is because of the freedom in $\lambda$, and it happens to converge to a different but equally valid solution. I think I checked the flux surfaces and they agreed -- should I add a plot comparison test to check that? Any ideas on other tests to add?

dpanici: maybe not a plot_comparison but a comparison of the area enclosed by each, like we do currently when we compare VMEC and DESC solutions, or a comparison of |B| strength at points on a grid (since the poloidal angles are the same right?)

And maybe I am a bit confused but what is different btwn these two eqs? are the surfaces not the same, and the profiles as well?

ddudt: I like the suggestion to compare the enclosed areas -- I'll try that, thanks. 

The two equilibria are supposed to be the same. I take the original equilibrium, generate a VMEC input file for it, and then solve in DESC from that VMEC input. So I am testing to see if that equilibrium matches the original one. But there is a freedom on lambda for the internal surfaces during the solve, so comparing the coefficients themselves isn't sufficient.

dpanici: Sure but I guess I am saying that this should be equivalent to running DESC twice with the same inputs? since despite the internal freedom we should get the same answer since the code is deterministic (barring numerical precision/hardware differences). 

You could just re-solve the SOLOVEV example then check it against the other one. It is really not necessary if you do area difference of B strength instead, and I get that if the SOLOVEV example is from an earlier version of the code or used a different objective than it may have slight differences, but just wanted to point it out that they should theoretically be the same, even the coefficients, if the solution method and inputs were the same

ddudt: The code is deterministic, but different continuation methods can lead to slightly different results. In our SOLOVEV example, we do a manual continuation method with 2 steps of different resolutions. When we convert from a VMEC input file in this test, we solve it in a single step for the final resolution. So that's why the coefficients don't match exactly. 

I added the area difference to the test and the solutions agree to a low tolerance.

desc/equilibrium/equilibrium.py
```diff
@@ -638,9 +638,6 @@ def get_surface_at(self, rho=None, theta=None, zeta=None):
             Zb = AZ @ self.Z_lmn
             surface.R_lmn = Rb
             surface.Z_lmn = Zb
-            surface.grid = LinearGrid(
-                rho=rho, M=2 * surface.M, N=2 * surface.N, endpoint=True, NFP=self.NFP
-            )
```
ddudt, line 643: The surface classes no longer have grid attributes. This is a relic that we forgot to remove before.

desc/geometry/surface.py
ddudt, line 1: These changes were to made to be consistent with our structure of having "private" attributes accessible by getter/setter methods. 

This change also necessitated updating all of the example files. I didn't re-run all of the equilibria, but simply reset `eq.surface.rho = 1` for each eq in each family.

f0uriest (approved):

dpanici (approved):

f0uriest (approved):

PR #816, opened 2024-01-05
Dp/surface from pts

### LINKED ISSUE #815: Make method of creating surface object from points of R,Z (and theta, phi)

dpanici:
- Adds a class method `from_values` to `FourierRZToroidalSurface` to create a surface by fitting `(R,phi,Z)` points, along with a user-defined poloidal angle `theta` which sets the poloidal angle for the created surface

Resolves #815

### DIFF
--- a/desc/geometry/surface.py
+++ b/desc/geometry/surface.py
@@ -4,6 +4,7 @@ import numbers
 import warnings
 
 import numpy as np
+import scipy
 
 from desc.backend import jnp, put, sign
 from desc.basis import DoubleFourierSeries, ZernikePolynomial
@@ -447,25 +448,29 @@ class FourierRZToroidalSurface(Surface):
             )
             Zb_lmn = transform.fit(Z)
         else:  # perform weighted fit
-            # solves system A^T W A x = A^T W b
-            # where A is the transform matrix, W is the diagonal weight matrix,
-            # and b is the vector of data points
+            # solves system W A x = W b
+            # where A is the transform matrix, W is the diagonal weight matrix
+            # of weights w, and b is the vector of data points
             w = np.asarray(w)
-            W = np.diag(w**2)
+            W = np.diag(w)
             assert w.size == R.size, "w must same length as number of points being fit"
 
             transform = Transform(
                 nodes, R_basis, build=True, build_pinv=False, method="direct1"
             )
-            A = transform.matrices[transform.method][0][0][0]
-            # multiple w*R since W @ R is just elementwise multiplication
-            Rb_lmn = np.linalg.lstsq(A.T @ W @ A, A.T @ (w * R), rcond=rcond)[0]
+            AR = transform.matrices[transform.method][0][0][0]
 
             transform = Transform(
                 nodes, Z_basis, build=True, build_pinv=False, method="direct1"
             )
-            A = transform.matrices[transform.method][0][0][0]
-            Zb_lmn = np.linalg.lstsq(A.T @ W @ A, A.T @ (w * Z), rcond=rcond)[0]
+            AZ = transform.matrices[transform.method][0][0][0]
+
+            A = scipy.linalg.block_diag(W @ AR, W @ AZ)
+            b = np.concatenate([w * R, w * Z])
+            x_lmn = np.linalg.lstsq(A, b, rcond=rcond)[0]
+
+            Rb_lmn = x_lmn[0 : R_basis.num_modes]
+            Zb_lmn = x_lmn[R_basis.num_modes :]
 
         surf = cls(
             Rb_lmn,
--- a/desc/geometry/surface.py
+++ b/desc/geometry/surface.py
@@ -361,7 +361,17 @@ class FourierRZToroidalSurface(Surface):
 
     @classmethod
     def from_values(
-        cls, coords, theta, zeta=None, M=6, N=6, NFP=1, sym=True, check_orientation=True
+        cls,
+        coords,
+        theta,
+        zeta=None,
+        M=6,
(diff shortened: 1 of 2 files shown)


rahulgaur104:
I am trying to use this to get the Fourier series for a surface. Here's my script.

```
#!/usr/bin/env python3
"""
In this script, I want to use the parametrization of an umbilic torus as defined in the file Umbilic_reserach.nb
to create a 3D MHD equilibrium using DESC. I will start with the modified Releaux torus.
rg6256
"""
import pdb
import numpy as np

# Define the parametrization of the umbilic torus
ntheta = int(300)
nphi = int(100)

phi_arr = np.linspace(0, 2 * np.pi, nphi)
t = np.linspace(0, 2 * np.pi, ntheta)

n = 3

r1 = 1.5
r2 = 1.
r3 = 1.

fig = plt.figure()
plt.axis('equal')

markers = ['o','s','^','*', 'x', 'D', 'v', 'p', 'h', '+', '1', '2', '3', '4', '8', 'H', 'D', 'd', '|', '_']

data = np.zeros((ntheta*nphi, 3))

for k, phi in enumerate(phi_arr):
    R = 3 + 2 * np.cos(np.pi / (2 * n)) * np.cos((t + np.pi * (2 * np.floor(n * t / (2 * np.pi)) + 1)/n)/2 + (phi+r1)/n) - np.cos(np.pi * (2 * np.floor(n * t/(2 * np.pi)) + 1)/ n + (phi+r1)/n) + r2*np.cos(phi)
    Z = 2 * np.cos(np.pi / (2 * n)) * np.sin((t + np.pi * (2 * np.floor(n * t / (2 * np.pi)) + 1)/n)/2 + phi/n) - np.sin(np.pi * (2 * np.floor(n * t/(2 * np.pi)) + 1)/ n + phi/n) + r3*np.sin(phi)
    
    arr = np.array([R, phi*np.ones((ntheta,)), Z]).T
    data[:ntheta, :] = arr 

theta = t
surface2 = FourierRZToroidalSurface.from_values(data, theta, M=20, N=10, NFP=1, sym=True)

pdb.set_trace()

```

However, I get the following error:

`TypeError: dot_general requires contracting dimensions to have the same shape, got (100,) and (30000,).
`

rahulgaur104:
Apart from code coverage, everything looks great!

desc/geometry/surface.py
```diff
@@ -357,6 +359,67 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
         surf = cls(R_lmn=R_lmn, Z_lmn=Z_lmn, modes_R=modes_R, modes_Z=modes_Z, NFP=NFP)
         return surf
 
+    @classmethod
+    def from_values(cls, coords, theta, M, N, NFP, sym=True, check_orientation=True):
```
f0uriest, line 363: do we want default values for M,N,NFP?

Also do we want to include a dummy argument for zeta for future-proofing with the toroidal angle stuff? (and for now just enforce that zeta=phi)?

f0uriest, line 363: It might also be good to include an option for weighted/uncertain least squares for fitting experimental data etc, we also do that for the profile classes.

desc/geometry/surface.py
```diff
@@ -357,6 +359,67 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
...
+
+        Parameters
+        ----------
+        coords : array-like shape(num_points,3) or Grid
+            cylindrical coordinates (R,phi,Z) to fit as a FourierRZToroidalSurface
+        theta : ndarray, shape(num_points,)
+            Locations in poloidal angle theta where real space coordinates are given.
+            Expects same number of angles as coords (num_points),
+            This determines the poloidal angle for the resulting surface.
+        M : int
```
f0uriest, line 374: may want to include notes here that fitting works better if M<ntheta, N<nzeta etc.

desc/geometry/surface.py
f0uriest, line 363: also just a reminder for @rahulgaur104 that you'll need to add the umbilic factor to this as well

f0uriest: (not in this PR)

ddudt (dismissed):

desc/geometry/surface.py
```diff
@@ -357,6 +359,125 @@ def from_near_axis(cls, aspect_ratio, elongation, mirror_ratio, axis_Z, NFP=1):
...
+            w = np.asarray(w)
+            W = np.diag(w**2)
+            assert w.size == R.size, "w must same length as number of points being fit"
+
+            transform = Transform(
+                nodes, R_basis, build=True, build_pinv=False, method="direct1"
+            )
+            A = transform.matrices[transform.method][0][0][0]
+            # multiple w*R since W @ R is just elementwise multiplication
+            Rb_lmn = np.linalg.lstsq(A.T @ W @ A, A.T @ (w * R), rcond=rcond)[0]
```
f0uriest, line 462: I don't think this is right. The least squares call already implicitly does the `A.T @ A` stuff, i think you just need to multiply `A` and `R` by `w`

You could also save a factorization by doing it for R and Z at the same time

ddudt: this looks right if the formula on line 450 is correct

dpanici: I think Rory is right here, my bad, this might be right but is forming a square system and giving to lstsq when it can already do that for us in a smarter way

f0uriest (approved):

PR #817, opened 2024-01-07
Dp/signed distance

### LINKED ISSUE #1070: Add `eq_fixed` as option for `PlasmaVesselDistance`
Add in #817

dpanici:
- Add ``use_signed_distance`` flag to ``PlasmaVesselDistance`` which will use a signed distance as the target, which is positive when the plasma is inside of the vessel surface and negative if the plasma is outside of the vessel surface, to allow optimizer to distinguish if the equilbrium surface exits the vessel surface and guard against it by targeting a positive signed distance.
- Deprecates `alpha` in favor of `softmin_alpha` to avoid confusion with other quantities similarly-named in the code

Resolves #1070 
- [x] vectorize the signed distance algorithm
- [x] use signed dist with softmin
- [x] remove redundant or unneeded tests
- [x] decide on keeping Circular dist objective


desc/objectives/_geometry.py
```diff
@@ -707,22 +728,265 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
+                    data_surf["n_rho"], diff_vecs_at_mins, axis=-1
+                ).squeeze()
+                # we mult by -1 because diff_vec points from surface to
+                # the plasma, so
+                # if plasma is inside the surface, n_rho dot diff vec <0,
+                # but we define "plasma inside surface" as >0
+                return min_ds * jnp.sign(n_rho_dot_diff_vec) * -1
+
+
+class PlasmaVesselDistanceCircular(_Objective):
```
f0uriest, line 767: How much more efficient/effective is this that just using the regular one? Do we need a separate objective just for this case? It seems like this might be specific to NTtau stuff.

desc/objectives/_geometry.py
```diff
@@ -567,6 +576,12 @@ def __init__(
         self._surface_grid = surface_grid
         self._plasma_grid = plasma_grid
         self._use_softmin = use_softmin
+        self._use_signed_distance = use_signed_distance
+        if use_softmin and use_signed_distance:
```
f0uriest, line 580: can you explain a bit more why these two don't work together? I'm not sure I totally understand

desc/objectives/_geometry.py
```diff
@@ -707,22 +728,265 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
 
         if self._use_softmin:  # do softmin
             return jnp.apply_along_axis(softmin, 0, d, self._alpha)
         else:  # do hardmin
-            return d.min(axis=0)
+            if not self._use_signed_distance:
+                return d.min(axis=0)
+            else:
+                min_inds = d.argmin(axis=0, keepdims=True)
+                min_ds = jnp.take_along_axis(d, min_inds, axis=0).squeeze()
```
f0uriest, line 753: this is the smallest magnitude distance, is that what we want? IE, suppose the plasma surface is bulging really far outside of the bounding surface, then this might still be small (as there is somewhere where the two surfaces touch) but what we want to penalize is the large negative distance where its bulging out the most.

desc/objectives/_geometry.py
dpanici: It is more effective in that it definitely assigns signs correctly, but yea it is a bit constrained to the NT tao stuff.

desc/objectives/_geometry.py
dpanici: it is the smallest magnitude distance per point on the surface that we are considering, so even if the surfaces touch somewhere.. hm yea I see what you are saying. What I am doing would make sense/work if we were taking the minimum distances from the points on the plasma to the surface (i.e. if `dim_f = plas_grid.num_nodes`) as then we would be looking at the minimum distances per point on the plasma, and so if the plasma bulges far we would still have a point whose min distance is large and then it would be assigned a negative value based off the normal vector dot product.

desc/objectives/_geometry.py
ddudt: I agree this is limited in application to small-scale experiments, like nTtau and Muse. It might not be worth maintaining an extra objective function for such a niche use case, if the original one works fine.

desc/objectives/_geometry.py
```diff
@@ -567,6 +576,12 @@ def __init__(
         self._surface_grid = surface_grid
         self._plasma_grid = plasma_grid
         self._use_softmin = use_softmin
+        self._use_signed_distance = use_signed_distance
+        if use_softmin and use_signed_distance:
+            warnings.warn(
+                "signed distance cannot currently" " be used with use_softmin=True!",
```
ddudt, line 582: string syntax issue? Either make the whole this a single string, or I think you are missing a `+` between the two strings

desc/objectives/_geometry.py
dpanici: use argmin on distances to determine sign after using softmin

dpanici:
Or just change to force grids to be the same toroidal angle, then look at things plane by plane and check distances like in the circular objective (find dist to magnetic axis at that phi, and then compare those...)

dpanici:
Algorithm implementing the winding number approach to find if a point is inside a surface
can use this to assign signs to the min distancesof each point on the plasma, if we restrict the plasma and surf grids to use the same zeta values

based off of https://www.sciencedirect.com/science/article/abs/pii/B9780123361561500124?via%3Dihub

```python
import numpy as np
from desc.backend import jnp


def find_angle_vec(R,Z,Rtest,Ztest):
    """R Z and surf points, Rtest Ztest are the point we wanna check is inside or not"""
    # R Z can be vecs for sure
    Rbool = R>Rtest
    Zbool = Z>Ztest
    return_data = np.zeros_like(R)
    return_data = np.where(np.logical_and(Rbool, Zbool), 
                           0,
                           return_data)
    return_data = np.where(np.logical_and(np.logical_not( Rbool), Zbool), 
                           1,
                           return_data)
    return_data = np.where(np.logical_and(np.logical_not( Rbool) , np.logical_not( Zbool)), 
                           2,
                           return_data)
    return_data = np.where(np.logical_and(Rbool , np.logical_not( Zbool)), 
                           3,
                           return_data)
    return return_data
            

circ = FourierRZToroidalSurface()
pt_in_R = np.array([10])
pt_in_Z = np.array([0])
data_circ = circ.compute(["R","Z"],grid=LinearGrid(M=10,L=0,N=0,endpoint=True))
# vectorized over the surf pts
# NOTE: you loop over every VERTEX, so the last vertex you consider is the last point to the first point, so here we must add a ghost last point to account
# for that
R_circ = np.append(data_circ["R"],data_circ["R"][0])
Z_circ=np.append(data_circ["Z"],data_circ["Z"][0])
R=R_circ
Z=Z_circ
quads = find_angle_vec(R_circ,Z_circ,pt_in_R[0],pt_in_Z[0]) 
deltas = quads[1:] - quads[0:-1]
deltas = np.where(deltas==3,-1,deltas)
deltas = np.where(deltas==-3,1,deltas)
# then flip sign if the R intercept is > Rtest and the 
# quadrant flipped over a diagonal
b = (Z[1:]/R[1:] - Z[0:-1]/R[0:-1]) / (Z[1:]-Z[0:-1])
Rint = pt_in_Z[0] - b*(R[1:]-R[0:-1])/(Z[1:]-Z[0:-1])
deltas = np.where(np.logical_and(np.abs(deltas)==2, Rint>pt_in_R[0]),-deltas,deltas)

theta = np.sum(deltas)
print(theta) # this is +/- 4 if inside the shape, and 0 if outside

```

desc/objectives/_geometry.py
dpanici: black moved things around, but two strings will be assumed to be concatenated even without a + in between so it is only an aesthetic issue

tests/test_optimizer.py
```diff
@@ -983,6 +984,141 @@ def test_constrained_AL_scalar():
     np.testing.assert_array_less(-Dwell, ctol)
 
 
+@pytest.mark.slow
+@pytest.mark.unit
+@pytest.mark.optimize
+def test_proximal_with_PlasmaVesselDistance():
```
f0uriest, line 990: is this test new? I thought we already had one like this?

desc/objectives/_geometry.py
```diff
@@ -567,6 +567,16 @@ class PlasmaVesselDistance(_Objective):
...
+    use_signed_distance: bool, optional
+        Whether to use absolute value of distance or a signed distance, with d
+        being positive if the plasma is inside of the bounding surface, and
+        negative if outside of the bounding surface.
+        NOTE: ``plasma_grid`` and ``surface_grid`` must have the same
+        toroidal angle values for signed distance to be used.
+        NOTE: this convention assumes that both surface and equilibrium have
+        poloidal angles defined such that they are in a right-handed coordinate
+        system with the surface normal vector pointing outwards
+        NOTE: only works with use_softmin=False currently
```
f0uriest, line 579: is this still the case? if so should add a check for it

desc/objectives/_geometry.py
```diff
@@ -698,18 +728,33 @@ def build(self, use_jit=True, verbose=1):
             "quad_weights": w,
         }
 
+        if self._use_signed_distance:
+            # get the indices corresponding to the grid points
+            # at each distinct zeta plane, so that can be used
+            # in compute to separate the computed pts by zeta plane
+            zetas = plasma_grid.nodes[plasma_grid.unique_zeta_idx, 2]
+            plasma_zeta_indices = [
```
f0uriest, line 736: isnt this basically `grid.inverse_zeta_idx`?

desc/objectives/_geometry.py
```diff
@@ -750,24 +795,324 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
+                # negative distance is assigned
+                # pt_sign = 0 : Means SURFACE is OUTSIDE of the PLASMA,
+                #               assign positive distance
+                # pt_sign = +/-4: Means SURFACE is INSIDE PLASMA, so
+                #                 assign negative distance
+                pt_sign = jnp.where(jnp.isclose(pt_sign, 0), 1, -1)
+                return pt_sign
+
+            # loop over zeta planes
+            for plasma_zeta_idx, surface_zeta_idx in zip(
```
f0uriest, line 863: can this be done with vmap or a jax loop?

desc/objectives/_geometry.py
```diff
@@ -750,24 +795,324 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
+
+        surface_coords = data_surf["x"]
+        diff_vec = plasma_coords[:, None, :] - surface_coords[None, :, :]
+        d = safenorm(diff_vec, axis=-1)
+
+        point_signs = jnp.ones(surface_coords.shape[0])
+        if self._use_signed_distance:
+            surface_coords_rpz = xyz2rpz(surface_coords)
+
+            def _find_angle_vec(R, Z, Rtest, Ztest):
```
f0uriest, line 820: can you include a reference for this algorithm?

f0uriest (changes requested):

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ New Features
 
 - Add method ``from_values`` to ``FourierRZCurve`` to allow fitting of data points
 to a ``FourierRZCurve`` object, and ``to_FourierRZCurve`` methods to ``Curve`` class.
+- Adds ``PlasmaVesselDistanceCircular`` as an objective, which is a simpler
+way to calculate the plasma-vessel separation when the vessel is a circular
+axisymmetric torus
```
ddudt, line 10: If you remove this class, also remove this here

desc/objectives/_geometry.py
ddudt: Yeah I don't think we need all this logic because it is already done in the `Grid` class (and we already checked the grids have the same zeta values)

desc/objectives/_geometry.py
```diff
@@ -567,6 +567,12 @@ class PlasmaVesselDistance(_Objective):
         Defaults to ``LinearGrid(M=eq.M_grid, N=eq.N_grid)``.
     use_softmin: bool, optional
         Use softmin or hard min.
+    use_signed_distance: bool, optional
```
ddudt, line 570: Is the signed algorithm much slower? Or is it common to use different grids for the two different surfaces? I'm wondering if we should just always use the signed version (or at least default to the signed version)

desc/objectives/_geometry.py
```diff
@@ -750,24 +791,324 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
+        diff_vec = plasma_coords[:, None, :] - surface_coords[None, :, :]
+        d = safenorm(diff_vec, axis=-1)
+
+        point_signs = jnp.ones(surface_coords.shape[0])
+        if self._use_signed_distance:
+            surface_coords_rpz = xyz2rpz(surface_coords)
+
+            def _find_angle_vec(R, Z, Rtest, Ztest):
+                # R Z and surface points,
+                # Rtest Ztest are the point we wanna check is inside
```
ddudt, line 818: horrendous grammar lol

desc/objectives/_geometry.py
```diff
@@ -750,24 +795,324 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
 
         if self._use_softmin:  # do softmin
-            return jnp.apply_along_axis(softmin, 0, d, self._alpha)
+            return jnp.apply_along_axis(softmin, 0, d, self._alpha) * point_signs
         else:  # do hardmin
-            return d.min(axis=0)
+            return d.min(axis=0) * point_signs
+
+
+class PlasmaVesselDistanceCircular(_Objective):
```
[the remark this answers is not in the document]
ddudt: I vote to remove it

desc/objectives/_geometry.py
```diff
@@ -750,24 +791,324 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
+            # positive meaning the surface pt
+            # is outside the plasma and -1 if the surface pt is
+            # inside the plasma
 
         if self._use_softmin:  # do softmin
-            return jnp.apply_along_axis(softmin, 0, d, self._alpha)
+            return jnp.apply_along_axis(softmin, 0, d, self._alpha) * point_signs
         else:  # do hardmin
-            return d.min(axis=0)
+            return d.min(axis=0) * point_signs
```
ddudt, line 886: Does hard min makes this whole objective a scalar? In that case we also have to take the sign from that minimum point, because I think `point_signs` here is still a vector. 

What axis is soft min taking the minimum along? In other words, what is the dimension of the output with soft min? 

This objective always uses `dim_f = surface_grid.num_nodes` so I think there are some inconsistencies here.

ddudt (changes requested):

desc/objectives/_geometry.py
dpanici: likely yes, I will refactor

desc/objectives/_geometry.py
dpanici: yea I will clean up

tests/test_optimizer.py
dpanici: Not new, I think you might have removed it in your test cleanup PR though. Probably this is unneeded and I can remove it now as I assume elsewhere something similar tests it

desc/objectives/_geometry.py
dpanici: No, this PR does not change any dimensions. The `d` object here in both cases is a 2D array of dimensions [N_plasma, N_surf] (or vice versa, though I think it is this one), and it contains the matrix of distances between every plasma point and every surface point. The reduction operation to a "min" distance can either be, for each surface point, the min of the dist btwn that point and every point on the plasma, or the softmin function of the dist btwn that point and every point on the plasma.

both will return `surface_grid.num_nodes` sized array, so there are no inconsistencies.

desc/objectives/_geometry.py
dpanici: Currently it is yes, as it has to loop over the zeta planes to assign the signs. I have not tested it extensively though but I think it should be optional for now

desc/objectives/_geometry.py
```diff
@@ -750,24 +771,113 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
+
+        surface_coords = data_surf["x"]
+        diff_vec = plasma_coords[:, None, :] - surface_coords[None, :, :]
+        d = safenorm(diff_vec, axis=-1)
+
+        point_signs = jnp.ones(surface_coords.shape[0])
+        if self._use_signed_distance:
+            surface_coords_rpz = xyz2rpz(surface_coords)
+
+            def _find_angle_vec(R, Z, Rsurf, Zsurf):
```
rahulgaur104, line 796: Can this algorithm be used to calculate the distance between a curve and a surface?

dpanici: yes if at same zeta

desc/objectives/_geometry.py
```diff
@@ -3,6 +3,7 @@
 import warnings
 
 import numpy as np
+from jax import vmap
```
f0uriest, line 6: import this from `desc.backend`

desc/objectives/_geometry.py
```diff
@@ -603,6 +610,7 @@ def __init__(
         surface_grid=None,
         plasma_grid=None,
         use_softmin=False,
+        use_signed_distance=False,
         surface_fixed=False,
         alpha=1.0,
```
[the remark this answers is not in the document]
f0uriest: maybe `softmin_alpha` to make it clear where its used

tests/test_optimizer.py
```diff
@@ -983,6 +983,97 @@ def test_constrained_AL_scalar():
...
+    optimizer = Optimizer("lsq-exact")
+    (eq, surf), _ = optimizer.optimize(
+        (eq, surf), objective, constraints, verbose=3, maxiter=60, ftol=1e-8, xtol=1e-9
+    )
+
+    np.testing.assert_allclose(
+        obj.compute(*obj.xs(eq, surf)), -0.25, atol=1e-2, err_msg="Using hardmin"
+    )
+
+    # with softmin
```
kianorr, line 1017: Do you mind explaining what softmin is? Or is there a resource with a good definition?

desc/objectives/_geometry.py
```diff
@@ -656,9 +664,21 @@ def build(self, use_jit=True, verbose=1):
         if not np.allclose(plasma_grid.nodes[:, 0], 1):
             warnings.warn("Plasma grid includes interior points, should be rho=1")
 
+        # TODO: How to use with generalized toroidal angle?
+        errorif(
+            self._use_signed_distance
+            and not np.allclose(
```
kianorr, line 670: what is this checking?

desc/objectives/_geometry.py
```diff
@@ -750,24 +770,115 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
+                # or not they are inside the plasma
+
+                # algorithm based off of "An Incremental Angle Point in Polygon Test",
+                # K. Weiler, https://doi.org/10.1016/B978-0-12-336156-1.50012-4
+
+                Rbool = R[:, None] > Rsurf
+                Zbool = Z[:, None] > Zsurf
+                # these are now size (Nplasma, Nsurf)
+                quads = jnp.zeros_like(Rbool)
+                quads = jnp.where(jnp.logical_and(Rbool, Zbool), 0, quads)
```
kianorr, line 807: what is `quads`? 

Also it seems like this is checking where `Rbool` and `Zbool` are both true and making those cases 0, otherwise `quads` (which is already all zero) so doesn't that mean everything stays 0?

tests/test_optimizer.py
dpanici: https://en.wikipedia.org/wiki/Smooth_maximum

https://github.com/PlasmaControl/DESC/blob/8c4051a555365f87c34871cf073a35f2c0d06e93/desc/objectives/utils.py#L215

basically is a way to smoothly approximate the max/min function, where the parameter alpha determines the degree of approximation, for large alpha the functions tend to yield the max value of the input array. This is useful as it allows the gradient of the output of this function to be nonzero for every entry, not just for the entry that is the largest/smallest (which is what the max/min function would do, as the function ONLY depends on whatever the max/min value is and not on the other parts of the array)

desc/objectives/_geometry.py
dpanici: the signed distance algorithm I use only works if the points lie in the same toroidal plane, this is checking that. Check the link to the algorithm: https://doi.org/10.1016/B978-0-12-336156-1.50012-4

desc/objectives/_geometry.py
dpanici: it is checking in which "quadrant" each point on the plasma is in relation to each point on the surface. The algorithm works by counting the number of points in each quadrant and assigning a value based on which quadrant, then adding them up, and that determines if the point is inside of outside the given perimeter.

ah true, we can skip this check, good catch

desc/objectives/_geometry.py
```diff
@@ -613,8 +620,9 @@ def __init__(
         self._surface_grid = surface_grid
         self._plasma_grid = plasma_grid
         self._use_softmin = use_softmin
+        self._use_signed_distance = use_signed_distance
         self._surface_fixed = surface_fixed
-        self._alpha = alpha
+        self._softmin_alpha = softmin_alpha
```
f0uriest, line 625: should add a deprecation warning for `alpha` if we're changing the name: 
https://github.com/PlasmaControl/DESC/blob/8c4051a555365f87c34871cf073a35f2c0d06e93/desc/utils.py#L492-L493

desc/objectives/_geometry.py
```diff
@@ -603,8 +609,9 @@ def __init__(
         surface_grid=None,
         plasma_grid=None,
         use_softmin=False,
+        use_signed_distance=False,
```
f0uriest, line 612: should add new positional args to the end to avoid mismatches in existing code.

tests/test_optimizer.py
```diff
@@ -983,6 +983,97 @@ def test_constrained_AL_scalar():
     np.testing.assert_array_less(-Dwell, ctol)
 
 
+@pytest.mark.slow
+@pytest.mark.unit
+@pytest.mark.optimize
+def test_signed_PlasmaVesselDistance():
```
f0uriest, line 989: how slow is this? does it need to be a unit test?

tests/test_objective_funs.py
```diff
@@ -1122,6 +1124,70 @@ def test_target_profiles():
     )
 
 
+@pytest.mark.unit
+def test_signed_plasma_vessel_distance():
```
f0uriest, line 1128: can this be moved to the `TestObjectiveFunction` class to keep things organized?

f0uriest (changes requested):

tests/test_optimizer.py
dpanici: about 20 seconds, I can make it a a regression test

kianorr (dismissed):

f0uriest (approved):
Update the description at the top of the PR before merging, it still references some old stuff.

desc/objectives/_geometry.py
ddudt: Could we spin this out into it's own utility function? Because we might have use for this other applications (for example, checking if a point is inside the plasma boundary for field line tracing)

ddudt (changes requested):

tests/test_objective_funs.py
```diff
@@ -956,6 +958,88 @@ def test(eq, field, correct_value, rtol=1e-14, grid=None):
...
+            surface_grid=grid,
+            plasma_grid=LinearGrid(M=10, N=6),
+            surface=surface,
+            use_signed_distance=True,
+        )
+        obj.build()
+        d = obj.compute_unscaled(*obj.xs(eq, surface))
+        assert obj.dim_f == d.size
+        assert abs(d.max() - (-0.5 * a_p)) < 1e-14
+        assert abs(d.min() - (-0.5 * a_p)) < grid.spacing[0, 1] * a_p * 0.5
```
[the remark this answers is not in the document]
dpanici: they are scalar quantities, Ihad just copied this from PlasmaVesselDistance.

desc/objectives/_geometry.py
```diff
@@ -604,17 +610,25 @@ def __init__(
         plasma_grid=None,
         use_softmin=False,
         surface_fixed=False,
-        alpha=1.0,
+        softmin_alpha=1.0,
```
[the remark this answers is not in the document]
dpanici: there is a check later on in the code for the size of softmin_alpha compared to the size of the arrays passed in, and it increases the size automatically if it is too small.

tests/test_objective_funs.py
```diff
@@ -956,6 +958,88 @@ def test(eq, field, correct_value, rtol=1e-14, grid=None):
...
         test(eq, field, psi_from_field)
 
+    @pytest.mark.unit
+    def test_signed_plasma_vessel_distance(self):
+        """Test calculation of signed distance from plasma to vessel."""
+        R0 = 10.0
+        a_p = 1.0
+        a_s = 2.0
+        # default eq has R0=10, a=1
+        eq = Equilibrium(M=3, N=2)
```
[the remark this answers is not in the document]
dpanici: Elsewhere in our tests we do things based off the default geometry, and it has not changed since the code's inception. I would rather keep this as it is.

tests/test_optimizer.py
dpanici: yes

f0uriest (dismissed):

f0uriest (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,18 @@
 Changelog
 =========
 
+New Features
```
f0uriest, line 4: Needs to be cleaned up

f0uriest (approved):

desc/objectives/_geometry.py
```diff
@@ -766,12 +830,55 @@ def compute(self, equil_params, surface_params=None, constants=None):
...
+                3,
+            )
+
+            # loop over zeta planes
+            def fun(plasma_pts_at_zeta_plane, surface_pts_at_zeta_plane):
+                plasma_pts_at_zeta_plane = jnp.vstack(
+                    (plasma_pts_at_zeta_plane, plasma_pts_at_zeta_plane[0, :])
+                )
+
+                surface_pts_at_zeta_plane
```
rahulgaur104, line 858: Is there a missing comma here or is surface_pts_at_zeta_planept_sign a single variable?
The formatting looks weird.

rahulgaur104 (dismissed):

desc/objectives/_geometry.py
```diff
@@ -601,20 +607,42 @@ def __init__(
...
         self._surface_fixed = surface_fixed
-        self._alpha = alpha
+        self._eq_fixed = eq_fixed
+        self._eq = eq
+        errorif(
+            eq_fixed and surface_fixed, ValueError, "Cannot fix both eq and surface"
+        )
+
+        self._softmin_alpha = parse_argname_change(
+            softmin_alpha, kwargs, "alpha", "softmin_alpha"
```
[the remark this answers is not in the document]
dpanici: we've relabelled in `softmin_alpha` now, to avoid this confusion, and in a future version can deprecate `alpha` as a kwarg for this

desc/objectives/_geometry.py
dpanici: ah good catch idt this line is doing anything rn

desc/objectives/utils.py
```diff
@@ -295,3 +295,78 @@ def _parse_callable_target_bounds(target, bounds, x):
...
+        jnp.logical_and(jnp.abs(deltas) == 2, Rint.T > Rcheck),
+        -deltas,
+        deltas,
+    )
+    pt_sign = jnp.sum(deltas, axis=0)
+    # positive distance if the check pt is inside the perimeter, else
+    # negative distance is assigned
+    # pt_sign = 0 : Means point is OUTSIDE of the perimeter,
+    #               assign positive distance
+    # pt_sign = +/-4: Means point is INSIDE perimeter, so
```
rahulgaur104, line 369: What is the relevance of -4?

rahulgaur104 (approved):

desc/objectives/utils.py
dpanici: It basically is a net count of how many quadrants around the given point have been passed through by the perimeter. If you are a point inside the perimeter, then this number will be nonzero, as the perimeter goes all around the point, but if the point is outside the perimeter, then this number is zero as the perimeter does not encompass the point

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
...
+
+- Add ``use_signed_distance`` flag to ``PlasmaVesselDistance`` which will use a signed distance as the target, which is positive when the plasma is inside of the vessel surface and negative if the plasma is outside of the vessel surface, to allow optimizer to distinguish if the equilbrium surface exits the vessel surface and guard against it by targeting a positive signed distance.
+- Add utility function ``check_if_points_are_inside_perimeter`` to ``desc.objectives.utils`` which will check whether a given set
```
f0uriest, line 7: if this isn't part of the public API it doesn't need to be in the changelog

f0uriest (dismissed):

f0uriest (approved):

rahulgaur104 (approved):

desc/objectives/_geometry.py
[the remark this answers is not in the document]
ddudt: @dpanici Where was this?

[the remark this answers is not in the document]
dpanici: https://github.com/PlasmaControl/DESC/blob/5c309d85ba16a9e0629ddc59cca6edc79f5c33cc/desc/objectives/utils.py#L231

PR #818, opened 2024-01-08
fix bug where FourierCurrentPotentialField would not initialize Phi_mn coefficients properly

dpanici:
- also adds asserts to ensure passed-in modes match the coefficients size to `FourierCurrentPotentialField` and `FourierRZToroidalSurface`

### DIFF
--- a/desc/geometry/curve.py
+++ b/desc/geometry/curve.py
@@ -72,6 +72,9 @@ class FourierRZCurve(Curve):
 
         modes_R, modes_Z = np.asarray(modes_R), np.asarray(modes_Z)
 
+        assert R_n.size == modes_R.size, "R_n size and modes_R must be the same size!"
+        assert Z_n.size == modes_Z.size, "Z_n size and modes_Z must be the same size!"
+
         assert issubclass(modes_R.dtype.type, np.integer)
         assert issubclass(modes_Z.dtype.type, np.integer)
         assert isposint(NFP)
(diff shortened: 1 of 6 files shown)


dpanici:
> Was the bug was caused by a size mismatch between `Phi_mn` and `modes_Phi`? It might be good to include an explicit check for that, with a warning/error for the user. We should also probably do that check for other classes that are initialized with both coefficient values and mode numbers, like `FourierRZToroidalSurface`. Otherwise this looks good.

It was not with a size mismatch from the provided modes, it was that the initial modes provided did not match the full modes array of the corresponding Basis created (i.e. the copy coeffs line was missing). I used code from that surface class you mention to fix the issue. Ensuring they are the same size initially is a good idea though too, I will add

desc/geometry/curve.py
```diff
@@ -533,6 +540,8 @@ def __init__(
         else:
             modes = np.asarray(modes)
         assert issubclass(modes.dtype.type, np.integer)
+        print(modes.shape)
```
f0uriest, line 543: extra print

desc/geometry/curve.py
```diff
@@ -72,6 +72,9 @@ def __init__(
 
         modes_R, modes_Z = np.asarray(modes_R), np.asarray(modes_Z)
 
+        assert R_n.size == modes_R.size, "R_n size and modes_R must be the same size!"
```
f0uriest, line 75: I don't think we need exclamation points on all of these

ddudt: we should write all of our error messages with a really angry tone so the users feel stupid when they do something wrong

ddudt (approved):
Looks good, besides Rory's comments

f0uriest (approved):

ddudt (approved):

PR #819, opened 2024-01-09
Rg/nfp fac Adding a separate factor for NFP to calculate umbilic configurations

rahulgaur104:
Our group has been trying to generate equilibria with umbilic shapes, i.e., shapes with a rational effective field period. These configurations are useful for a resonant divertor design and simplifies divertor placement.

The objective of this PR is to add an additional factor that replaces all instances of NFP to NFP/NFP_umbilic_factor and make sure that the code runs without any issues.

### DIFF
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -44,7 +44,7 @@ class _Basis(IOAble, ABC):
         self._M = int(self._M)
         self._N = int(self._N)
         self._NFP = int(self._NFP)
-        self._NFP_umbilic_factor = float(self._NFP_umbilic_factor)
+        self._NFP_umbilic_factor = int(self._NFP_umbilic_factor)
         self._modes = self._modes.astype(int)
 
     def _set_up(self):
@@ -59,7 +59,7 @@ class _Basis(IOAble, ABC):
         self._M = int(self._M)
         self._N = int(self._N)
         self._NFP = int(self._NFP)
-        self._NFP_umbilic_factor = float(self._NFP_umbilic_factor)
+        self._NFP_umbilic_factor = int(self._NFP_umbilic_factor)
         self._modes = self._modes.astype(int)
 
     def _enforce_symmetry(self):
@@ -204,7 +204,7 @@ class _Basis(IOAble, ABC):
 
     @property
     def NFP_umbilic_factor(self):
-        """float: Umbilic factor field periods."""
+        """int: Umbilic factor field periods."""
         return self.__dict__.setdefault("_NFP_umbilic_factor", 1)
 
     @property
@@ -365,8 +365,8 @@ class FourierSeries(_Basis):
         Maximum toroidal resolution.
     NFP : int
         number of field periods
-    NFP_umbilic_factor : float
-        Rational number of the form 1/integer with integer>=1.
+    NFP_umbilic_factor : int
+        Prefactor of the form 1/NFP_umbilic_fac.
 	This is needed for the umbilic torus design.
     sym : {``'cos'``, ``'sin'``, False}
         * ``'cos'`` for cos(m*t-n*z) symmetry
@@ -467,7 +467,7 @@ class FourierSeries(_Basis):
         NFP : int
             Number of field periods.
         NFP_umbilic_factor : float
-            Rational number of the form 1/integer with integer>=1.
+            Prefactor of the form 1/NFP_umbilic_factor.
             This is needed for the umbilic torus design.
         sym : bool
             Whether to enforce stellarator symmetry.
@@ -496,7 +496,7 @@ class DoubleFourierSeries(_Basis):
     NFP : int
         Number of field periods.
     NFP_umbilic_factor : float
-        Rational number of the form 1/integer with integer>=1.
+        Prefactor of the form 1/NFP_umbilic_factor. 
 	This is needed for the umbilic torus design.
     sym : {``'cos'``, ``'sin'``, ``False``}
         * ``'cos'`` for cos(m*t-n*z) symmetry
@@ -598,7 +598,7 @@ class DoubleFourierSeries(_Basis):
             n = n[nidx]
 
         poloidal = fourier(t[:, np.newaxis], m, 1, derivatives[1])
-        toroidal = fourier(z[:, np.newaxis], n, self.NFP*self.NFP_umbilic_factor, derivatives[2])
+        toroidal = fourier(z[:, np.newaxis], n, self.NFP/self.NFP_umbilic_factor, derivatives[2])
         if unique:
             poloidal = poloidal[toutidx][:, moutidx]
             toroidal = toroidal[zoutidx][:, noutidx]
@@ -617,7 +617,7 @@ class DoubleFourierSeries(_Basis):
         NFP : int
             Number of field periods.
         NFP_umbilic_factor : float
-            Rational number of the form 1/integer with integer>=1.
+            Prefactor of the form 1/NFP_umbilic_factor. 
             This is needed for the umbilic torus design.
         sym : bool
             Whether to enforce stellarator symmetry.
@@ -864,7 +864,7 @@ class ChebyshevDoubleFourierBasis(_Basis):
     NFP : int
         Number of field periods.
     NFP_umbilic_factor : float
-        Rational number of the form 1/integer with integer>=1.
+        Prefactor of the form 1/NFP_umbilic_factor. 
         This is needed for the umbilic torus design.
     sym : {``'cos'``, ``'sin'``, ``False``}
         * ``'cos'`` for cos(m*t-n*z) symmetry
@@ -950,7 +950,7 @@ class ChebyshevDoubleFourierBasis(_Basis):
 
         radial = chebyshev(r[:, np.newaxis], l, dr=derivatives[0])
         poloidal = fourier(t[:, np.newaxis], m, 1, derivatives[1])
-        toroidal = fourier(z[:, np.newaxis], n, self.NFP*self.NFP_umbilic_factor, derivatives[2])
+        toroidal = fourier(z[:, np.newaxis], n, self.NFP/self.NFP_umbilic_factor, derivatives[2])
 
         return radial * poloidal * toroidal
 
@@ -968,7 +968,7 @@ class ChebyshevDoubleFourierBasis(_Basis):
         NFP : int
             Number of field periods.
         NFP_umbilic_factor : float
-            Rational number of the form 1/integer with integer>=1.
+            Prefactor of the form 1/NFP_umbilic_factor. 
             This is needed for the umbilic torus design.
         sym : bool
             Whether to enforce stellarator symmetry.
@@ -1006,7 +1006,7 @@ class FourierZernikeBasis(_Basis):
     NFP : int
         Number of field periods.
     NFP_umbilic_factor : float
-        Rational number of the form 1/integer with integer>=1.
+        Prefactor of the form 1/NFP_umbilic_factor. 
         This is needed for the umbilic torus design.
     sym : {``'cos'``, ``'sin'``, ``False``}
         * ``'cos'`` for cos(m*t-n*z) symmetry
@@ -1190,7 +1190,7 @@ class FourierZernikeBasis(_Basis):
 
         radial = zernike_radial(r[:, np.newaxis], lm[:, 0], lm[:, 1], dr=derivatives[0])
         poloidal = fourier(t[:, np.newaxis], m, dt=derivatives[1])
-        toroidal = fourier(z[:, np.newaxis], n, NFP=self.NFP*self.NFP_umbilic_factor, dt=derivatives[2])
+        toroidal = fourier(z[:, np.newaxis], n, NFP=self.NFP/self.NFP_umbilic_factor, dt=derivatives[2])
         if unique:
             radial = radial[routidx][:, lmoutidx]
             poloidal = poloidal[toutidx][:, moutidx]
@@ -1212,7 +1212,7 @@ class FourierZernikeBasis(_Basis):
         NFP : int
             Number of field periods.
         NFP_umbilic_factor : float
-            Rational number of the form 1/integer with integer>=1.
+            Prefactor of the form 1/NFP_umbilic_factor. 
             This is needed for the umbilic torus design.
         sym : bool
             Whether to enforce stellarator symmetry.
@@ -1622,7 +1622,7 @@ def fourier(theta, m, NFP=1, NFP_umbilic_factor = 1, dt=0):
     NFP : int
         number of field periods (Default = 1)
     NFP_umbilic_factor : float
-        Rational number of the form 1/integer.
+        NFP prefactor of the form 1/NFP_umbilic_factor. 
 	This is needed for the umbilic torus design.
     dt : int
         order of derivative (Default = 0)
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -29,7 +29,7 @@ class _Basis(IOAble, ABC):
         "_M",
         "_N",
         "_NFP",
-	"_NFPufac",
+	"_NFP_umbilic_factor",
         "_modes",
         "_sym",
         "_spectral_indexing",
@@ -44,7 +44,7 @@ class _Basis(IOAble, ABC):
         self._M = int(self._M)
         self._N = int(self._N)
         self._NFP = int(self._NFP)
-        self._NFPufac = float(self._NFPufac)
+        self._NFP_umbilic_factor = float(self._NFP_umbilic_factor)
         self._modes = self._modes.astype(int)
 
     def _set_up(self):
@@ -59,7 +59,7 @@ class _Basis(IOAble, ABC):
         self._M = int(self._M)
         self._N = int(self._N)
         self._NFP = int(self._NFP)
-        self._NFPufac = float(self._NFPufac)
+        self._NFP_umbilic_factor = float(self._NFP_umbilic_factor)
         self._modes = self._modes.astype(int)
 
     def _enforce_symmetry(self):
@@ -203,9 +203,9 @@ class _Basis(IOAble, ABC):
         return self.__dict__.setdefault("_NFP", 1)
 
     @property
-    def NFPufac(self):
+    def NFP_umbilic_factor(self):
         """float: Umbilic factor field periods."""
-        return self.__dict__.setdefault("_NFPufac", 1)
+        return self.__dict__.setdefault("_NFP_umbilic_factor", 1)
 
     @property
     def sym(self):
@@ -237,8 +237,8 @@ class _Basis(IOAble, ABC):
             type(self).__name__
             + " at "
             + str(hex(id(self)))
-            + " (L={}, M={}, N={}, NFP={}, NFPufac={}, sym={}, spectral_indexing={})".format(
-                self.L, self.M, self.N, self.NFP, self.NFPufac, self.sym, self.spectral_indexing
+            + " (L={}, M={}, N={}, NFP={}, NFP_umbilic_factor={}, sym={}, spectral_indexing={})".format(
+                self.L, self.M, self.N, self.NFP, self.NFP_umbilic_factor, self.sym, self.spectral_indexing
             )
         )
 
@@ -263,7 +263,7 @@ class PowerSeries(_Basis):
         self.M = 0
         self.N = 0
         self._NFP = 1
-        self._NFPufac = 1
+        self._NFP_umbilic_factor = 1
         self._sym = sym
         self._spectral_indexing = "linear"
 
@@ -365,7 +365,7 @@ class FourierSeries(_Basis):
         Maximum toroidal resolution.
     NFP : int
         number of field periods
-    NFPufac : float
+    NFP_umbilic_factor : float
         Rational number of the form 1/integer with integer>=1.
 	This is needed for the umbilic torus design.
     sym : {``'cos'``, ``'sin'``, False}
@@ -375,12 +375,12 @@ class FourierSeries(_Basis):
 
     """
 
-    def __init__(self, N, NFP=1, NFPufac = 1, sym=False):
+    def __init__(self, N, NFP=1, NFP_umbilic_factor = 1, sym=False):
         self.L = 0
         self.M = 0
         self.N = N
         self._NFP = NFP
-        self._NFPufac = NFPufac
+        self._NFP_umbilic_factor = NFP_umbilic_factor
         self._sym = sym
         self._spectral_indexing = "linear"
 
@@ -451,13 +451,13 @@ class FourierSeries(_Basis):
             z = z[zidx]
             n = n[nidx]
 
-        toroidal = fourier(z[:, np.newaxis], n, self.NFP*self.NFPufac, derivatives[2])
+        toroidal = fourier(z[:, np.newaxis], n, self.NFP*self.NFP_umbilic_factor, derivatives[2])
         if unique:
             toroidal = toroidal[zoutidx][:, noutidx]
 
         return toroidal
 
-    def change_resolution(self, N, NFP=None, NFPufac=None, sym=None):
+    def change_resolution(self, N, NFP=None, NFP_umbilic_factor=None, sym=None):
         """Change resolution of the basis to the given resolutions.
 
         Parameters
@@ -466,7 +466,7 @@ class FourierSeries(_Basis):
             Maximum toroidal resolution.
         NFP : int
             Number of field periods.
-        NFPufac : float
+        NFP_umbilic_factor : float
             Rational number of the form 1/integer with integer>=1.
             This is needed for the umbilic torus design.
         sym : bool
@@ -474,7 +474,7 @@ class FourierSeries(_Basis):
 
         """
         self._NFP = NFP if NFP is not None else self.NFP
-        self._NFPufac = NFPufac if NFPufac is not None else self.NFPufac
+        self._NFP_umbilic_factor = NFP_umbilic_factor if NFP_umbilic_factor is not None else self.NFP_umbilic_factor
         if N != self.N:
             self.N = N
             self._sym = sym if sym is not None else self.sym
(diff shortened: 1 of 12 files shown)


desc/equilibrium/equilibrium.py
```diff
@@ -198,14 +202,32 @@ def __init__(
         )
 
         errorif(
-            (NFP is not None) and not isposint(NFP),
+            (NFP_umbilic_factor is not None) and not isposint(NFP_umbilic_factor),
```
f0uriest, line 205: this should be regular NFP, the umbilic one is checked below

desc/equilibrium/equilibrium.py
```diff
@@ -427,9 +473,9 @@ def __repr__(self):
             type(self).__name__
             + " at "
             + str(hex(id(self)))
-            + " (L={}, M={}, N={}, NFP={}, sym={}, spectral_indexing={})".format(
-                self.L, self.M, self.N, self.NFP, self.sym, self.spectral_indexing
-            )
+            + f" (L={self.L}, M={self.M}, N={self.N}, NFP={self.NFP},"
+            + f" NFP_umbilic_factor={self.NFP_umbilic_factor}, sym={self.sym},"
```
f0uriest, line 477: not sure if we need to include the umbilic stuff here, since this is mainly just used when printing very summarized information about the object

dpanici:
#844 will help with this

desc/equilibrium/equilibrium.py
```diff
@@ -64,6 +64,8 @@ class Equilibrium(IOAble, Optimizable):
         total toroidal flux (in Webers) within LCFS. Default 1.0
     NFP : int (optional)
         number of field periods Default ``surface.NFP`` or 1
+    NFP_umbilic_factor : int (optional)
+        field period prefactor Default ``surface.NFP_umbilic_factor`` or 1
```
dpanici, line 68: this is not a very good description, I would say "integer dividing the field period, such that the effective field period is NFP/NFP_umbilic_factor" or something.

Also out of curiosity, would this work with NFP that is not equal to one?

desc/equilibrium/equilibrium.py
```diff
@@ -353,15 +374,34 @@ def __init__(
...
         )
 
+        errorif(
+            not (
+                eq_NFP_umbilic_factor
+                == surf_NFP_umbilic_factor
+                == axis_NFP_umbilic_factor
+            ),
+            ValueError,
+            "Unequal number of umbilic field period factors for equilibrium "
```
dpanici, line 400: this should just be "unequal umbilici field period prefactor" or something, not number of

desc/equilibrium/equilibrium.py
```diff
@@ -554,6 +602,9 @@ def change_resolution(
             Toroidal real space grid resolution.
         NFP : int
             Number of field periods.
+        NFP_umbilic_factor : float
```
dpanici, line 605: is this not supposed to be an integer still?

desc/equilibrium/equilibrium.py
```diff
@@ -876,7 +985,13 @@ def compute(
...
         if calc0d and override_grid:
-            grid0d = QuadratureGrid(self.L_grid, self.M_grid, self.N_grid, self.NFP)
+            grid0d = QuadratureGrid(
+                self.L_grid,
+                self.M_grid,
+                self.N_grid,
+                self.NFP,
+                NFP_umbilic_factor=self.NFP_umbilic_factor,
+            )
             data0d = compute_fun(
```
dpanici, line 995: have you checked if quantities like volume are correct when the prefactor is used?

desc/geometry/curve.py
```diff
@@ -29,6 +29,9 @@ class FourierRZCurve(Curve):
         Mode numbers associated with Z_n, If not given defaults to [-n:n]].
     NFP : int
         Number of field periods.
+    NFP_umbilic_factor : float
```
dpanici, line 32: again this is an integer right? not a float?

desc/geometry/surface.py
```diff
@@ -31,6 +31,9 @@ class FourierRZToroidalSurface(Surface):
         mode numbers associated with Z_lmn, defaults to modes_R
     NFP : int
         number of field periods
+    NFP_umbilic_factor : float
```
dpanici, line 34: should be an int

desc/geometry/surface.py
```diff
@@ -116,14 +124,24 @@ def __init__(
                 sym = False
 
         self._R_basis = DoubleFourierSeries(
-            M=self._M, N=self._N, NFP=NFP, sym="cos" if sym else False
+            M=MR,
```
dpanici, line 127: why change these from the self._M and self._N that they were previously?

desc/geometry/surface.py
```diff
@@ -116,14 +124,24 @@ def __init__(
...
+            N=NR,
+            NFP=NFP,
+            NFP_umbilic_factor=NFP_umbilic_factor,
+            sym="cos" if sym else False,
         )
         self._Z_basis = DoubleFourierSeries(
```
dpanici, line 133: same as above

desc/geometry/surface.py
```diff
@@ -169,16 +203,18 @@ def change_resolution(self, *args, **kwargs):
         """Change the maximum poloidal and toroidal resolution."""
         assert (
             ((len(args) in [2, 3]) and len(kwargs) == 0)
-            or ((len(args) in [2, 3]) and len(kwargs) in [1, 2])
+            or ((len(args) in [2, 3]) and len(kwargs) in [1, 3])
```
dpanici, line 206: len kwargs could be 1,2, or 3 right? for sym, NFP and NFP umbilic? so this should be [1,2,3] instead of [1,3]

desc/objectives/_omnigenity.py
```diff
@@ -163,6 +169,7 @@ def build(self, use_jit=True, verbose=1):
             transforms["B"].basis.modes,
             helicity=self.helicity,
             NFP=transforms["B"].basis.NFP,
+            NFP_umbilic_factor=transforms["B"].basis.NFP_umbilic_factor,
```
dpanici, line 172: As a test to see if we can get ANY QS umbilic (meaning just having a umbilic factor greater than one, not caring about sharp edges as even with smooth edges it is still a shape that repeates only every 2pi * umbliic factor times), can you try optimizing without the extra umbilic objective you have?

this would give confidence in the code changes you've made and that everything works fine with the umbilic factor, before adding other objectives we aren't sure of

desc/grid.py
```diff
@@ -933,13 +967,30 @@ def change_resolution(self, L, M, N, NFP=None):
             new toroidal grid resolution (N toroidal nodes)
         NFP : int
             Number of field periods.
+        NFP_umbilic_factor : float
```
dpanici, line 970: int

desc/grid.py
```diff
@@ -976,17 +1027,25 @@ class QuadratureGrid(_Grid):
         toroidal grid resolution (exactly integrates toroidal modes up to order N)
     NFP : int
         number of field periods (Default = 1)
+    NFP_umbilic_factor : float
```
dpanici, line 1030: int

desc/grid.py
```diff
@@ -1014,6 +1073,9 @@ def _create_nodes(self, L=1, M=1, N=1, NFP=1):
             toroidal grid resolution (N toroidal nodes, Default = 1)
         NFP : int
             number of field periods (Default = 1)
+        NFP_umbilic_factor : float
```
dpanici, line 1076: int

desc/grid.py
```diff
@@ -1071,12 +1136,25 @@ def change_resolution(self, L, M, N, NFP=None):
             new toroidal grid resolution (N toroidal nodes)
         NFP : int
             Number of field periods.
+        NFP_umbilic_factor : float
```
dpanici, line 1139: int

desc/grid.py
```diff
@@ -1107,6 +1185,9 @@ class ConcentricGrid(_Grid):
         toroidal grid resolution
     NFP : int
         number of field periods (Default = 1)
+    NFP_umbilic_factor : float
```
dpanici, line 1188: int

desc/grid.py
```diff
@@ -1160,6 +1262,9 @@ def _create_nodes(self, L, M, N, NFP=1, axis=False, node_pattern="jacobi"):
             toroidal grid resolution
         NFP : int
             number of field periods (Default = 1)
+        NFPUfac : float
```
dpanici, line 1265: int and name is different now

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+
+
+class UmbilicCurvature(_Objective):
+    """Takes in an equilibrium object and an umbilic curve object.
+
+    Umbilic object corresponds to the sharp boundary
+    of the true (sharp) equilibrium.
+
+    Using the grid defined by the umbilic curve object,
+    this class returns the minimum gaussian curvature k2 along that curve
```
dpanici, line 1405: it actually returns the max absolute curveture not the min

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+    loss_function : {None, 'mean', 'min', 'max'}, optional
+        Loss function to apply to the objective values once computed. This loss function
+        is called on the raw compute value, before any shifting, scaling, or
+        normalization.
+    deriv_mode : {"auto", "fwd", "rev"}
+        Specify how to compute jacobian matrix, either forward mode or reverse mode AD.
+        "auto" selects forward or reverse mode based on the size of the input and output
+        of the objective. Has no effect on self.grad or self.hess which always use
+        reverse mode and forward over reverse mode respectively.
+    grid : Grid, optional
```
dpanici, line 1440: grids in init dont match docstring

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+    def __init__(
+        self,
+        eq,
+        curve,
+        target=None,
+        bounds=None,
+        weight=1,
+        curve_grid=None,
+        equil_grid=None,
+        equil_grid_axis=None,
```
dpanici, line 1462: what is equil_grid_axis?

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        curve_fixed=False,
```
dpanici, line 1463: not in docstring

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        deriv_mode="auto",
+        grid=None,
+        name="Umbilic curvature",
+    ):
+
+        if target is None and bounds is None:
+            target = -10
```
dpanici, line 1473: if it is returning the max absolute gaussian curvature, then this target is not possible to attain? or is the docstring in the compute incorrect? that probably is the case here

desc/objectives/_geometry.py
dpanici: Actually I stand corrected it returns k_2

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+
+        """
+        eq = self.things[0]
+        curve = self._curve if self._curve_fixed else self.things[1]
+        # if things[1] is different than self._curve, update self._curve
+        if curve != self._curve:
+            self._curve = curve
+        if self._curve_grid is None:
+            phi_arr = jnp.linspace(0, 2 * jnp.pi, 4 * curve.N)
+            phi_arr = jnp.concatenate(
```
dpanici, line 1511: what is going on here exactly? you ad 2pi and then mod it? does this only work for an umbilic factor of 3?

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        else:
+            curve_grid = self._curve_grid
+
+        if self._equil_grid_axis is None:
+            phi_arr = jnp.linspace(0, 2 * jnp.pi, 4 * curve.N)
+            phi_arr = jnp.concatenate(
+                (phi_arr, phi_arr + 2 * jnp.pi, phi_arr + 4 * jnp.pi)
+            )
+            phi_arr = jnp.mod(phi_arr, 2 * jnp.pi)
+            equil_grid_axis = LinearGrid(
```
dpanici, line 1526: what is going on here exactly? you add 2pi and 4pi and then mod it? does this only work for an umbilic factor of 3?

desc/objectives/_geometry.py
dpanici: if this needs to go to 2pi * NFP_umbilic_fac, then need to make this more general, not hardcoded for factor of 3

desc/objectives/_geometry.py
dpanici: this is the same thing as just repeating the original phi_arr 3 times because of the modding btw

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+                basis="rpz",
+            )
+            self._constants["curve_coords"] = curve_coords
+
+        timer.stop("Precomputing transforms")
+        if verbose > 1:
+            timer.disp("Precomputing transforms")
+
+        if self._normalize:
+            self._normalization = 1.0
```
dpanici, line 1588: we could normalize this right? it has units of 1/m

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        # now compute the curvature
+        data = compute_fun(
+            "desc.equilibrium.equilibrium.Equilibrium",
+            self._equil_data_keys,
+            params=params_1,
+            profiles=equil_profiles,
+            transforms=equil_transforms,
+        )
+
+        return data["curvature_k2_rho"]
```
dpanici, line 1686: should we also multiply by like |e_theta x e_zeta| to weight by the local surface area? I am not sure

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        if verbose > 1:
+            timer.disp("Precomputing transforms")
+
+        if self._normalize:
+            self._normalization = 1.0
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params_1=None, params_2=None, constants=None):
+        """Compute max absolute principal curvature.
```
dpanici, line 1593: change this to the correct description

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        Z_axis = data_axis["Z"]
+
+        R_axis = R_axis.reshape(curve.NFP_umbilic_factor, 4 * curve.N).T
+        Z_axis = Z_axis.reshape(curve.NFP_umbilic_factor, 4 * curve.N).T
+
+        # calculate theta for each point in each section
+        theta = jnp.arctan2(Z - Z_axis, R - R_axis)
+        theta = jnp.mod(theta, 2 * jnp.pi)
+
+        phi_arr = jnp.mod(phi_arr, 2 * jnp.pi)
```
dpanici, line 1655: So the eq does not have the umbilic factor here, only the curve?

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+    _print_value_fmt = "Umbilic distance: {:10.3e} "
+
+    def __init__(
+        self,
+        eq,
+        curve,
+        target=None,
+        bounds=None,
+        weight=1,
+        curve_grid=None,
```
dpanici, line 1751: some of these not in docstring

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        loss_function=None,
+        deriv_mode="auto",
+        grid=None,
+        name="Umbilic curvature",
+    ):
+
+        if target is None and bounds is None:
+            target = 0.0
+
+        self._eq = curve
```
dpanici, line 1766: this is wrong (and unused?)

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        loss_function=None,
+        deriv_mode="auto",
+        grid=None,
+        name="Umbilic curvature",
+    ):
+
+        if target is None and bounds is None:
+            target = -10
+
+        self._eq = curve
```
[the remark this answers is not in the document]
dpanici: And is this unused? no need?

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,593 @@ def compute(self, params, constants=None):
...
+        curve_grid=None,
+        equil_grid=None,
+        equil_grid_axis=None,
+        curve_fixed=False,
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        deriv_mode="auto",
+        grid=None,
+        name="Umbilic curvature",
```
dpanici, line 1760: distance

desc/geometry/surface.py
rahulgaur104: I don't remember making that change. Reverting back...

desc/geometry/surface.py
rahulgaur104: I don't remember making that change. Reverting back...

rahulgaur104:
I am now trying to optimize a curve along with an equilibrium. On each point on an equilibrium surface closest to this curve, I try to impose a sharp curvature.

In this figure I impose sharp curvature on green, an weak curvature on red and blue.

desc/objectives/_geometry.py
```diff
@@ -1393,3 +1393,1388 @@ def compute(self, params, constants=None):
...
+        )
+        curve_R = curve_data["R"]
+        curve_phi = curve_data["phi"]
+        curve_Z = curve_data["Z"]
+
+        # First, we obtain a grid of points on the umbilic curve
+        curve_pts = rpz2xyz(jnp.array([curve_R, curve_phi, curve_Z]).T)
+
+        curve_pts = rpz2xyz(
+            self._curve._compute_position(
```
dpanici, line 2205: pass basis arg

desc/basis.py
```diff
@@ -186,6 +187,11 @@ def NFP(self):
         """int: Number of field periods."""
         return self.__dict__.setdefault("_NFP", 1)
 
+    @property
+    def NFP_umbilic_factor(self):
+        """int: Umbilic factor field periods."""
+        return self.__dict__.setdefault("_NFP_umbilic_factor", 1)
```
unalmis, line 193: You may need to add logic to [def _set_up](https://github.com/PlasmaControl/DESC/blob/363826c3ee0cbb71f4a0d92966ccedbe6b20a24e/desc/basis.py#L50) to set this factor for old equilibrium. Sometimes it might be as simple as
`self._NFP_umbilic_factor = self.NFP_umbilic_factor`

The setup method will be called automatically when you load an old file with examples.get("W7-X"). (The code that does this is here https://github.com/PlasmaControl/DESC/blob/4745460ede4ad2e3e9d5d25c762e1f963b4ce34a/desc/io/optimizable_io.py#L75)

desc/geometry/curve.py
```diff
@@ -114,31 +132,52 @@ def NFP(self):
         """Number of field periods."""
         return self._NFP
 
+    @property
+    def NFP_umbilic_factor(self):
+        """Field period umbilic factor."""
+        return self._NFP_umbilic_factor
```
unalmis, line 138: See previous comment

dpanici:
https://github.com/PlasmaControl/DESC/blob/master/tests/test_objective_funs.py
ctrl f "specials" and add umbilic objectices to that list, and then add a specific unit test to test those objetives below

desc/objectives/_geometry.py
```diff
@@ -1397,3 +1397,1315 @@ def compute(self, params, constants=None):
...
+        verbose : int, optional
+            Level of output.
+
+        """
+        eq = self._eq
+        curve = self._curve  # if self._curve_fixed else self.things[1]
+        # if things[1] is different than self._curve, update self._curve
+        if curve != self._curve:
+            self._curve = curve
+        if self._curve_grid is None:
```
dpanici, line 2355: either generlize for NFP fac !=3 or add an assert statement to ensure that is true

desc/objectives/_geometry.py
```diff
@@ -1397,3 +1397,1315 @@ def compute(self, params, constants=None):
...
+            transforms=constants["curve_transforms"],
+            profiles={},
+        )
+        curve_R = curve_data["R"]
+        curve_phi = curve_data["phi"]
+
+        theta_points = (curve_phi + curve_R) / 3
+
+        umbilic_edge_grid = Grid(
+            jnp.array([jnp.ones([1.0]), theta_points, curve_phi]).T, jitable=True
```
dpanici, line 2438: use ones_like here

rahulgaur104:
FAILED tests/test_plotting.py::TestPlotBoundary::test_plot_boundary_surface - pluggy.PluggyTeardownRaisedWarning: A plugin raised an exception during an old-style hookwrapper teardown.

But the test works on my local system.

Mention this during the group meeting.

unalmis:
> FAILED tests/test_plotting.py::TestPlotBoundary::test_plot_boundary_surface - pluggy.PluggyTeardownRaisedWarning: A plugin raised an exception during an old-style hookwrapper teardown.
> 
> But the test works on my local system.
> 
> Mention this during the group meeting.

When running the test locally, are you specifying to test that the plots match? (`pytest --mpl -k ...`)

rahulgaur104:
No, I did not use the mpl flag. I just ran` pytest -k test_plot_boundary_surface`
Without mpl, is it not comparing the figures at all?

unalmis:
Figures are only compared with the `--mpl` flag

desc/grid.py
```diff
@@ -440,19 +450,18 @@ def __repr__(self):
...
+            + " (L={}, M={}, N={}, NFP={}, NFP_umbilic_factor = {},\
+               sym={}, node_pattern={}, coordinates={})"
+        ).format(
+            self.L,
+            self.M,
+            self.N,
+            self.NFP,
+            self.NFP_umbilic_factor,
+            self.sym,
+            self.is_meshgrid,
```
unalmis, line 462: `is_meshgrid={}` lost in string format above. Can either add back or just remove `self._is_meshgrid` here.

rahulgaur104: Ok, resolved.

desc/grid.py
```diff
@@ -838,6 +848,10 @@ def _create_nodes(self, nodes):
         """
         nodes = jnp.atleast_2d(jnp.asarray(nodes)).reshape((-1, 3)).astype(float)
         # Do not alter nodes given by the user for custom grids.
+        # In particular, do not modulo nodes by 2pi or 2pi/(NFP/NFP_umbilic_factor).
+        # This may cause the surface_integrals() function to fail recognizing
+        # surfaces outside the interval [0, 2pi] as duplicates. However, most
+        # surface integral computations are done with LinearGrid anyway.
```
unalmis, line 854: this comment isn't relevant anymore, so you could remove it

desc/grid.py
```diff
@@ -958,6 +985,9 @@ def _create_nodes(  # noqa: C901
             Toroidal grid resolution.
         NFP : int
             Number of field periods (Default = 1).
+        NFP_umbilic_factor : float
+            Rational number of the form 1/integer with integer>=1.
+            This is needed for the umbilic torus design.
```
unalmis, line 990: I liked the earlier description more

desc/grid.py
```diff
@@ -1393,11 +1468,19 @@ def _create_nodes(self, L, M, N, NFP=1, axis=False, node_pattern="jacobi"):
             node spacing, based on local volume around the node
 
         """
+        self._L = L
+        self._M = M
+        self._N = N
+        self._NFP = NFP
```
unalmis, line 1474: These look like artifacts from merge conflict, should remove them.

rahulgaur104: Resolved.

desc/grid.py
rahulgaur104: Removed

desc/grid.py
```diff
@@ -656,6 +664,8 @@ def __init__(
         # Python 3.3 (PEP 412) introduced key-sharing dictionaries.
         # This change measurably reduces memory usage of objects that
         # define all attributes in their __init__ method.
+        self._NFP = 1
+        self._NFP_umbilic_factor = 1
```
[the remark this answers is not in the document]
rahulgaur104: Done

desc/examples/W7-X_output.h5
unalmis, line 1: check if using master version of this file passes test_compute_everything

rahulgaur104: It does.

dpanici:
I think the objective you added makes sense, it tries to align the alpha=0 field line with the umbilic curve. Something I am not sure of is, do your curves have a given helicity to them? that is, as phi increases do they move clockwise/counterclockwise poloidally? might have to make sure that if they do, that it is consistent with the helicity of the equilibrium field as determined by the sign of iota

rahulgaur104:
>I think the objective you added makes sense, it tries to align the alpha=0 field line with the umbilic curve. Something I am not sure of is, do your curves have a given helicity to them? that is, as phi increases do they move clockwise/counterclockwise poloidally? might have to make sure that if they do, that it is consistent with the helicity of the equilibrium field as determined by the sign of iota

Excellent question! Yes, there can be a sign issue but helicity of the curve = |curve.NPF/curve.NFP_umbilic_factor|. Here's what I get for a 3/5 umbilic curve.

3-5-umbilic-fieldline

dpanici:
https://github.com/PlasmaControl/DESC/blob/master/CONTRIBUTING.rst#what-if-the-test_compute_everything-test-fails-or-there-is-a-conflict-in-master_compute_data_rpzpkl

PR #821, opened 2024-01-10
Normalization Hotfix

dpanici:
- Fix for possible `inf` in normalization scales for asymmetric eqs (or any bdry that may have R10=Z10=0)


tests/test_objective_funs.py
```diff
@@ -1661,3 +1662,26 @@ def test_objective_no_nangrad():
         obj.build()
         g = obj.grad(obj.x(eq))
         assert not np.any(np.isnan(g)), str(objective)
+
+
+@pytest.mark.unit
+def test_asymmetric_normalization():
+    """Tests normalizations for asymmetric equilibrium."""
+    # related to PR
+    a = 0.6
```
rahulgaur104, line 1671: Could there be a potential code coverage issue if a is not close to 0. 
The if condition in the fix that we discussed is used when a is close to 0.
If a is finite that if condition is not called.

desc/objectives/normalization.py
```diff
@@ -21,6 +19,12 @@ def compute_scaling_factors(thing):
 
         scales["R0"] = R00
         scales["a"] = np.sqrt(np.abs(R10 * Z10))
+        # TODO: also implement for FourierRZToroidalSurface,
+        # need to add "A" compute fxn for that FourierRZToroidalSurface
+        if np.isclose(scales["a"], 0):
+            # R10 and Z10 can be 0 for stellarator asymmetric equilibria,
+            # just use the computed minor radius in this case
+            scales["a"] = thing.compute("a")["a"]
```
ddudt, line 27: What if we always did:
`scales["a"] = np.sqrt(np.abs(Rc * Zs) + np.abs(Rs * Zc))`
where `Rc` is the `M=+1` mode and `Rs` is the `M=-1` mode, etc. 

That should always works unless if like `Rc` and `Zc` are both 0 for some reason, which is unlikely. Otherwise we might as well always compute the actual minor radius and all the other quantities.

tests/test_objective_funs.py
```diff
@@ -1661,3 +1662,26 @@ def test_objective_no_nangrad():
         obj.build()
         g = obj.grad(obj.x(eq))
         assert not np.any(np.isnan(g)), str(objective)
+
+
+@pytest.mark.unit
+def test_asymmetric_normalization():
+    """Tests normalizations for asymmetric equilibrium."""
+    # related to PR
```
ddudt, line 1670: which PR?

ddudt (changes requested):

tests/test_objective_funs.py
dpanici: the if condition is not on `eq.compute("a")`, it is on `R10` and `Z10` which are zero in this example

tests/test_objective_funs.py
dpanici: forgot to link this one

desc/objectives/normalization.py
dpanici: the reason for this PR (@rahulgaur104  correct me if I am wrong) was a "a" which was zero for a stellarator symmetric case, so it did not even have `Rs` or `Zc`

ddudt:
The logic in my latest commit covers asymmetric cases, and should also handle cases where the m=1 modes are not defined, like @rahulgaur104 mentioned he has seen.

ddudt (approved):

PR #825, opened 2024-01-17
add 'scale_pressure' and 'copy' options to rescale

ddudt:
Adds two new options to the `desc/compat/rescale` function: 
1. `scale_pressure` to flag whether or not to modify the pressure profile to maintain force balance. This adds the option to keep the pressure profile the same, since a common use case is to re-solve the equilibrium with the original profile. 
2. `copy` to modify the original equilibrium or a copy.

### DIFF
--- a/desc/compat.py
+++ b/desc/compat.py
@@ -109,7 +109,9 @@ def flip_helicity(eq):
     return eq
 
 
-def rescale(eq, L=("R0", None), B=("B0", None), verbose=0):
+def rescale(
+    eq, L=("R0", None), B=("B0", None), scale_pressure=True, copy=False, verbose=0
+):
     """Rescale an Equilibrium in size L and magnetic field strength B.
 
     Parameters
@@ -124,6 +126,10 @@ def rescale(eq, L=("R0", None), B=("B0", None), verbose=0):
         First element is a string denoting the magnetic field strength to scale. One of:
         {"B0", "<|B|>", "B_max"} for B on axis, volume averaged, or maximum on the LCFS.
         Second element is a float denoting the desired field. Default is no scaling.
+    scale_pressure : bool, optional
+        Whether or not to scale the pressure profile to maintain force balance.
+    copy : bool, optional
+        Whether to rescale the original equilibrium (default) or a copy.
     verbose : int
         Level of output.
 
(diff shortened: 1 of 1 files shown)


dpanici:
document compat.py in the API

dpanici (approved):

dpanici (changes requested):
will be approved when the docs build passes

ddudt:
> will be approved when the docs build passes, it has this error:
> 
> ```
> /home/docs/checkouts/readthedocs.org/user_builds/desc-docs/checkouts/825/desc/compat.py:docstring of desc.compat.rescale:9: ERROR: Undefined substitution referenced: "B".
> ```

@dpanici Do you have any idea what this error message means? I'm not sure how to fix it

f0uriest:
> > will be approved when the docs build passes, it has this error:
> > 
> > ```
> > /home/docs/checkouts/readthedocs.org/user_builds/desc-docs/checkouts/825/desc/compat.py:docstring of desc.compat.rescale:9: ERROR: Undefined substitution referenced: "B".
> > ```
> 
> @dpanici Do you have any idea what this error message means? I'm not sure how to fix it

Usually it's when you use single bars for vector magnitude instead of double. Sphinx interprets single bars as text substitution

ddudt:
> > > will be approved when the docs build passes, it has this error:
> > > ```
> > > /home/docs/checkouts/readthedocs.org/user_builds/desc-docs/checkouts/825/desc/compat.py:docstring of desc.compat.rescale:9: ERROR: Undefined substitution referenced: "B".
> > > ```
> > 
> > 
> > @dpanici Do you have any idea what this error message means? I'm not sure how to fix it
> 
> Usually it's when you use single bars for vector magnitude instead of double. Sphinx interprets single bars as text substitution

Ah OK so it's probably being caused by `"<|B|>"` here:
```
B : tuple, (str, float)
        First element is a string denoting the magnetic field strength to scale. One of:
        {"B0", "<|B|>", "B_max"} for B on axis, volume averaged, or maximum on the LCFS.
        Second element is a float denoting the desired field. Default is no scaling.
```
But what should we do about that? Because I want to reference the literal string the user should input.

f0uriest:
Could try defining a dummy substitution?

https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html#use-a-substitution

Like that just replaces |B| with |B|

ddudt:
I decided to just change the string from `"<|B|>"` to `"<B>"` because it doesn't have to match the compute quantity (we don't have compute quantities for the other B scaling options anyways)

dpanici (approved):

PR #832, opened 2024-01-25
Dp/hotfix optimize multiple things

### LINKED ISSUE #828: Tracking order of `things` when multiple objectives take in more than one optimizable `thing`
I've found that when optimizing with two objectives that take in `things` in different orders (for example, `PlasmaVesselDistance` takes in (eq,surf) while `QuadraticFlux` takes in (field, eq) and sometimes surf=field if using a `FourierCurrentPotentialField`), the order of `things` matters when passed into `optimizer.optimize`

empirically, (eq, field) seems to work, while (field,eq) throws this error after the optimization, when the params of the objects are being changed, so somewhere the order of `x` is assuming eq is first...
```python
Cell In[21], line 1
----> 1 (surface_current_field_opt,eq_opt), result = optimizer.optimize(
      2         (surface_current_field,eq), objective, constraints, verbose=3, maxiter=2,ftol=1e-8,gtol=0,xtol=1e-14,copy=True
      3     )

File ~/DESC/desc/optimize/optimizer.py:308, in Optimizer.optimize(self, things, objective, constraints, ftol, xtol, gtol, ctol, x_scale, verbose, maxiter, options, copy)
    306 # temporarily assign new stuff for printing, might get replaced later
    307 for thing, params in zip(things, result["history"][-1]):
--> 308     thing.params_dict = params
    310 if verbose > 0:
    311     print("Start of solver")

File ~/DESC/desc/optimizable.py:51, in Optimizable.params_dict(self, d)
     49 for key, val in d.items():
     50     if jnp.asarray(val).size:
---> 51         setattr(self, key, val)

File ~/DESC/desc/geometry/surface.py:215, in FourierRZToroidalSurface.R_lmn(self, new)
    213     self._R_lmn = jnp.asarray(new)
    214 else:
--> 215     raise ValueError(
    216         f"R_lmn should have the same size as the basis, got {len(new)} for "
    217         + f"basis with {self.R_basis.num_modes} modes."
    218     )

ValueError: R_lmn should have the same size as the basis, got 385 for basis with 145 modes
```

dpanici:
- Fix bug that occurs when `optimizer.optimize` is called with `things` in an order different than the order of `objective.things`, allowing it to run error-free and to correctly perform the optimization
- Fix bug inside of assert statement of `ObjectiveFunction.x`

Essentially, this fix works by using the ordering of `objective.things`, which was used to unpack the `x` after the optimization is done, and just finding where in `things` corresponds to the order of `objective.things`.

It also ensure that `objective.x` is called with the correct order of `things` inside of `optimize`
 - I thought about adding this logic inside of `objective.x` but we also call that function with the copy of `things`, so that would not allow us to pass in any `things` that were not the exact Python objects inside of `objective.things`
 - with the assert statement fixed this should at least ensure that the correct types are passed in now

I am unsure if there are other places where this pops up but this seems to be enough to fix the bug I was encountering with the optimization of multiple objects, I think this is the only place though as inside of `optimize` is the only time where  user-generated ordering of `things` is passed, which might differ from `objective.things`

Resolves #828

### DIFF
--- a/desc/optimize/optimizer.py
+++ b/desc/optimize/optimizer.py
@@ -231,8 +231,7 @@ class Optimizer(IOAble):
                     "yellow",
                 )
             )
-
-        x0 = objective.x(*things)
+        x0 = objective.x(*[things[things.index(t)] for t in objective.things])
 
         stoptol = _get_default_tols(
             method,
@@ -310,18 +309,29 @@ class Optimizer(IOAble):
 
         if verbose > 0:
             print("Start of solver")
-            objective.print_value(objective.x(*things0))
+
+            # need to check index of things bc things0 contains copies of
+            # things, so they are not the same exact Python objects
+            objective.print_value(
+                objective.x(*[things0[things.index(t)] for t in objective.things])
+            )
             for con in constraints:
-                con.print_value(
-                    *con.xs(
-                        *[t0 for (t0, t) in zip(things0, things) if t in con.things]
-                    )
-                )
+                arg_inds_for_this_con = [
+                    things.index(t) for t in things if t in con.things
+                ]
+                args_for_this_con = [things0[ind] for ind in arg_inds_for_this_con]
+                con.print_value(*con.xs(*args_for_this_con))
 
             print("End of solver")
-            objective.print_value(objective.x(*things))
+            objective.print_value(
+                objective.x(*[things[things.index(t)] for t in objective.things])
+            )
             for con in constraints:
-                con.print_value(*con.xs(*[t for t in things if t in con.things]))
+                arg_inds_for_this_con = [
+                    things.index(t) for t in things if t in con.things
+                ]
+                args_for_this_con = [things[ind] for ind in arg_inds_for_this_con]
+                con.print_value(*con.xs(*args_for_this_con))
 
         if copy:
             # need to swap things and things0, since things should be unchanged
--- a/desc/optimize/optimizer.py
+++ b/desc/optimize/optimizer.py
@@ -304,8 +304,8 @@ class Optimizer(IOAble):
             _ = result.pop(key, None)
 
         # temporarily assign new stuff for printing, might get replaced later
-        for thing, params in zip(things, result["history"][-1]):
-            thing.params_dict = params
+        for thing, params in zip(objective.things, result["history"][-1]):
+            get_instance(things, type(thing)).params_dict = params
 
         if verbose > 0:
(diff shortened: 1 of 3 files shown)


f0uriest (approved):

tests/test_optimizer.py
```diff
@@ -1017,3 +1017,98 @@ def test_proximal_with_PlasmaVesselDistance():
...
+    constraints = (
+        # don't let eq vary
+        FixParameter(eq),
+        # only let the minor radius of the surface vary
+        FixParameter(surf, params=["R_lmn"], indices=surf.R_basis.get_idx(M=0, N=0)),
+    )
+    obj = PlasmaVesselDistance(
+        surface=surf,
+        eq=eq,
+        target=target_dist,
```
rahulgaur104, line 1092: Isn't this objective identical to the objective defined above?
Are you repeating to ensure independence between the two calculations?

desc/optimize/optimizer.py
```diff
@@ -231,8 +231,11 @@ def optimize(  # noqa: C901 - FIXME: simplify this
                     "yellow",
                 )
             )
-
-        x0 = objective.x(*things)
+        # we have to use this cumbersome indexing in this method when passing things
+        # to objective to guard against the passed-in things having an ordering
+        # different from objective.things, to ensure the correct order is passed
+        # to the objective
+        x0 = objective.x(*[things[things.index(t)] for t in objective.things])
```
rahulgaur104, line 238: So "things" here can be a list of objects from different classes like Equilibrium and FourierRZToroidalSurface. What is meaning of the *before things? I see this a lot in the code.

rahulgaur104 (approved):

desc/optimize/optimizer.py
dpanici: In python `fun(*list)` will "unpack" the list , so that if say we have `list=[a,b,c]`, calling `fun(*list)` is the same thing as calling `fun(a,b,c)`  i.e. it unpacks the iterable and passes the arguments to the function being called

dpanici: and since you will see this elsewhere, the single `*` unpacks a list or tuple, while a double `**` will unpack a dict, i.e `fun(**dict)` where `dict={"a":1, "b":2, "c":3}` will be the same thing as calling `fun(a=1, b=2, c=3)` i.e calling the function using the dict keys as  keyword args and the corresponding dict values as the values passed to the keywords

PR #833, opened 2024-01-25
Ensure curve compute method is jittable when grid is provided

f0uriest:


desc/geometry/core.py
```diff
@@ -111,16 +111,18 @@ def compute(
...
         if calc0d and override_grid:
-            grid0d = LinearGrid(N=2 * N + 5, NFP=NFP, endpoint=True)
+            grid0d = LinearGrid(N=2 * self.N + 5, NFP=NFP, endpoint=True)
             data0d = compute_fun(
                 self,
                 dep0d,
                 params=params,
-                transforms=get_transforms(dep0d, obj=self, grid=grid0d, **kwargs),
+                transforms=get_transforms(
+                    dep0d, obj=self, grid=grid0d, jitable=True, **kwargs
```
rahulgaur104, line 124: Making it jitable should make it faster on average. Is there a way to test that?
In other words, is there a way to check if jitable is working?

f0uriest: making it jitable actually makes it slower to build the transforms, since it has to use a direct method rather than fft. But this allows you to call the curve.compute method from within a jitted function (like an objective). As it is currently you can't even do that, so the speed is a secondary issue. Once its compiled the speed is basically the same.

desc/geometry/core.py
```diff
@@ -111,16 +111,18 @@ def compute(
         ]
         calc0d = bool(len(dep0d))
         # see if the grid we're already using will work for desired qtys
-        if calc0d and (grid.N >= 2 * N + 5) and isinstance(grid, LinearGrid):
+        if calc0d and (grid.N >= 2 * self.N + 5) and isinstance(grid, LinearGrid):
             calc0d = False
 
         if calc0d and override_grid:
-            grid0d = LinearGrid(N=2 * N + 5, NFP=NFP, endpoint=True)
+            grid0d = LinearGrid(N=2 * self.N + 5, NFP=NFP, endpoint=True)
```
dpanici, line 118: is this grid creation jitable?

f0uriest: yeah, self.N and NFP are both static attributes so are compile time constants.

dpanici (approved):

PR #834, opened 2024-01-25
Vectorize geometry utils for converting coordinate systems

f0uriest:


tests/test_geometry.py
```diff
@@ -29,7 +29,7 @@ def test_xyz2rpz():
 
     xyz = np.array([0, 1, 1])
     rpz = xyz2rpz_vec(xyz, x=0, y=1)
-    np.testing.assert_allclose(rpz, np.array([[1, 0, 1]]), atol=1e-10)
+    np.testing.assert_allclose(rpz, np.array([1, 0, 1]), atol=1e-10)
```
rahulgaur104, line 32: A test to check if vectorize is faster?

f0uriest: This wasn't done for speed. In effect they were already manually vectorized, this is just using JAX utils to allow it to handle arbitrary leading dimensions.

IE, previously it could handle arrays of shape (N, 3), but now it can also handle stuff of shape (N1, N2, ..., 3)

f0uriest: its tested implicitly in the free boundary calculation, I don't think we need a separate test for it since its basically just adding a jax decorator.

dpanici (approved):

tests/test_geometry.py
ddudt: Where do the higher dimensional arrays get used in the free boundary code? This is being merged into `master`, so it's not actually tested anywhere yet, even if it does get used in the free boundary code

PR #835, opened 2024-01-25
Switch from vmap to scan for coilset compute

f0uriest:
Saves a lot of memory and works better for heterogeneous coil types


desc/backend.py
```diff
@@ -555,6 +556,45 @@ def fun_vmap(fun_inputs):
 
         return fun_vmap
 
+    def scan(f, init, xs, length=None, reverse=False, unroll=1):
```
rahulgaur104, line 559: I am assuming this is probably not vectorized.
So if we want to integrate an array of shape (L, M, N) along the last axis, will this work?

f0uriest: You'd need to transpose the array. This function is basically just a non-jax version of a jax function (for legacy reasons we've been trying to keep things compatible with pure numpy), so this is just trying to match the API of the jax version. In practice this never gets used, only the jax version `jax.lax.scan` (imported above)

dpanici (approved):

rahulgaur104 (approved):

desc/backend.py
ddudt: Haven't we discussed not supporting regular numpy anymore? I would be in favor of that in for a new release

ddudt (approved):

PR #839, opened 2024-01-26
Combine NAE objectives into a single one

### LINKED ISSUE #528: Combine near axis constraints into fewer number of `Objectives`
Right now when you `get_NAE_constraints` it returns a whole bunch of individual `FixSumXModes` objects, it would be cleaner if they were all packaged into a single object like `FixNAEX`, where you could instantiate it directly giving the target as some particular QSC equilibrium.

dpanici:
Add a new `FixNearAxisR`,`FixNearAxisZ`, `FixNearAxisLambda` objectives that combines all of the many `FixSumModes` objectives required for the fixed near axis behavior constraint into single objectives, which makes the output after a solve much more legible (only 3 errors reported instead of 20-40, one for each FixSumModes objective)

These objectives are now created when `get_NAE_constraints` util function is called

- [x] understand better why tolerance for QSC test had to increase

Resolves #528 

#829 should be merged in first before this one


desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3238,218 @@ def compute(self, params, constants=None):
 
         """
         return params["Psi"]
+
+
+class FixNearAxis(_FixedObjective):
```
f0uriest, line 3243: Would it make more sense to have separate constraints for R,Z,lambda like we do for the others?

desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3238,218 @@ def compute(self, params, constants=None):
...
+    """Fixes an equilibrium's near-axis behavior.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    nae_eq : Qsc
```
f0uriest, line 3250: it would be good if this was optional, so that you could either fix it to a specific NAE, or just fix it to the current equilibriums near axis behavior

desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3238,218 @@ def compute(self, params, constants=None):
...
+        pyQSC Qsc object describing the NAE solution to fix the equilibrium's
+        near-axis behavior to.
+    order : int
+        order (in rho) of near-axis behavior to constrain
+    fix_lambda : bool or int
```
f0uriest, line 3255: see above about maybe making this a separate constraint

desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3238,218 @@ def compute(self, params, constants=None):
...
+        Must be broadcastable to to Objective.dim_f
+    normalize : bool, optional
+        Whether to compute the error in physical units or non-dimensionalize.
+        Unused by this objective
+    normalize_target : bool, optional
+        Whether target and bounds should be normalized before comparing to computed
+        values. If `normalize` is `True` and the target is in physical units,
+        this should also be set to True.
+        Unused by this objective
+    sum_weights : float, ndarray, optional
```
f0uriest, line 3281: dont think this is needed here?

desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3238,218 @@ def compute(self, params, constants=None):
...
+        Weights on the coefficients in the sum, should be same length as modes.
+        Defaults to 1 i.e. target = 1*R_111 + 1*R_222...
+        Unused by this objective
+    modes : ndarray, optional
```
f0uriest, line 3285: don't think this is needed here?

desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3238,218 @@ def compute(self, params, constants=None):
 
         """
         return params["Psi"]
+
+
+class FixNearAxis(_FixedObjective):
+    """Fixes an equilibrium's near-axis behavior.
```
f0uriest, line 3244: would be good to clarify if this also fixes the axis itself.

IE, do users need to use this in addition to FixAxisR/Z, or are they redundant/conflicting?

f0uriest (changes requested):
Should also update the example notebook to use this new API

desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3234,393 @@ def compute(self, params, constants=None):
...
+    """Fixes an equilibrium's near-axis behavior in R to specified order.
+
+        NOTE: does not also fix the axis R , for that
+        use the FixAxisR objectives.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    nae_eq : Qsc, optonal
```
f0uriest, line 3249: would it make more sense to combine this with `target`?

desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3234,393 @@ def compute(self, params, constants=None):
...
+        pyQSC Qsc object describing the NAE solution to fix the equilibrium's
+        near-axis behavior to. If None, will fix the equilibrium's current near
+        axis behavior.
+    order : int
```
f0uriest, line 3253: I feel like if order is 0 this should fix the axis? So that for most cases you only need this not this + fixed axis

f0uriest: also if order can only be a discrete value it should be noted like 
`order : {0, 1, 2}`

desc/objectives/getters.py
```diff
@@ -191,14 +195,84 @@ def get_NAE_constraints(
...
+                N=N,
+                order=int(fix_lambda),
+                normalize=normalize,
+            ),
+        )
+
+    return constraints
+
+
+def _get_NAE_constraints(
```
ddudt, line 235: Why do we still have this function? I think this can be removed

desc/objectives/getters.py
```diff
@@ -191,14 +195,84 @@ def get_NAE_constraints(
...
+        ),
+        FixNearAxisZ(
+            eq=desc_eq,
+            nae_eq=qsc_eq,
+            N=N,
+            order=order,
+            normalize=normalize,
+        ),
+    )
+    if fix_lambda or (fix_lambda >= 0 and type(fix_lambda) is int):
```
ddudt, line 221: Why is `fix_lambda=False` by default? Shouldn't lambda be constrained to get the same NAE behavior?

desc/objectives/nae_utils.py
```diff
@@ -48,23 +46,31 @@ def calc_zeroth_order_lambda(qsc, desc_eq, N=None):
...
+        #  Landreman and Sengupta
+        nu_0 = np.cumsum(qsc.B0 / qsc.G0 * qsc.d_l_d_phi - 1) * np.ones_like(phi) * dphi
+        nu_0_n = Ltrans_sin.fit(nu_0)
+
+        # lambda = -iota_0 * nu
+        L_0_n = -qsc.iota * nu_0_n
+    else:
+        # we will fix to eq current on-axis lambda, so just set L_0_n to 0
+        # as it will be unused
+        L_0_n = np.zeros(Lbasis_sin.num_modes)
```
ddudt, line 67: Is this related to PR #829?

tests/test_examples.py
```diff
@@ -622,6 +622,92 @@ def test_NAE_QSC_solve():
...
+    eq_lambda_fixed_1st_order = eq.copy()
+
+    # this has all the constraints we need,
+    cs = get_NAE_constraints(eq, qsc_eq=None, order=1, fix_lambda=False, N=eq.N)
+    cs_lambda_fixed_0th_order = get_NAE_constraints(
+        eq_lambda_fixed_0th_order, qsc_eq=None, order=1, fix_lambda=0, N=eq.N
+    )
+    cs_lambda_fixed_1st_order = get_NAE_constraints(
+        eq_lambda_fixed_1st_order, qsc_eq=None, order=1, fix_lambda=True, N=eq.N
+    )
```
ddudt, line 650: Is `fix_lambda` a bool or an int? Is the int the order of the constraint? Does it use first order if True? That's misleading, if it is specifying the order it should always be an int.

ddudt (changes requested):

desc/objectives/getters.py
dpanici: I use it to create the constraints in `FixNearAxisXXX`

desc/objectives/getters.py
dpanici: No, we have found that it is a lot of the time not needed to fix lambda, and only at higher order have I seen that fixing lambda on-axis helps get the NAE behavior desired

desc/objectives/nae_utils.py
dpanici: no change was made to the zeroth order, I just rearranged stuff to make creating the near-axis constraint without a qsc eq (i..e so fixing to the equilibrium's current behavior) easier

desc/objectives/linear_objectives.py
dpanici: hm, yea maybe, so that target is either None or qsc_eq

desc/objectives/linear_objectives.py
dpanici: Yea I can do that but I just need to rearrange some of the logic in the utility functions

desc/objectives/getters.py
```diff
@@ -172,19 +176,107 @@ def get_NAE_constraints(
...
+    N=None,
+    fix_lambda=False,
+    normalize=True,
+):
+    """Get the constraints necessary for fixing NAE behavior in an equilibrium problem.
+
+    NOTE: This will return tuples of FixSumModes__, this is not intended to be directly
+    used by the user. Instead, call the ``get_NAE_constraints`` function, or use the
+    FixNearAxis{R,Z,Lambda} objectives along with the FixAxis{R,Z} objectives. This
+    instead is a helper function to get the needed constraints for the
```
VascoSch92, line 230: you didn't finish the sentence :-)

f0uriest (changes requested):

desc/objectives/getters.py
```diff
@@ -172,19 +176,107 @@ def get_NAE_constraints(
...
         L_axis_constraints, _, _ = calc_zeroth_order_lambda(
             qsc=qsc_eq, desc_eq=desc_eq, N=N
         )
         constraints += L_axis_constraints
+
+    # Axis constraints
+    constraints += (
+        FixAxisR(**kwargs),
+        FixAxisZ(**kwargs),
+        AxisRSelfConsistency(desc_eq),
```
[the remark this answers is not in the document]
dpanici: Yes, but in order for the `FixNearAxis` objectives to also fix the axis, they require the `A` matrix relating `R_lmn` and `Ra_n` which is actually created inside the self consistency objectives. So I include this here so that the `FixNearAxis` objective can have all the info it needs to fix the axis

desc/objectives/linear_objectives.py
```diff
@@ -2993,3 +2994,416 @@ def __init__(
...
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        name="Fix Near Axis R Behavior",
+    ):
+        self._nae_eq = target
+        self._eq = eq
+        self._order = order
+        self._N = N
+        self._target_from_user = setdefault(bounds, None)
```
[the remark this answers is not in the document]
dpanici: no, I can remove

f0uriest (approved):

ddudt (approved):
I didn't review it as closely as I usually like to, but I don't see anything glaringly wrong.

PR #840, opened 2024-01-26
fix curvy bugs

ddudt:
This PR fixes two issues related to the `Curve` class that are necessary for optimizing modular coils:
1. `Curve.shift` and `Curve.rotmat` need to be "optimizable_param"s because they are params that get used in the compute functions. 
2. The `_rotation_matrix_from_normal` function was giving incorrect results and is now replaced with calls to `rotation_matrix`. 

Things to note when reviewing:
- `master_compute_data.pkl` was changed because all of the old data for the `FourierPlanarCurve` class was incorrect. 
- All of the examples had to be updated to replace `eq.axis.shift` with `eq.axis._shift`, etc. so they can be loaded without I/O warnings.


ddudt:
patch coverage is 100% and I added a new test, but project coverage decreased because I deleted code

desc/compute/geom_utils.py
```diff
@@ -45,7 +46,7 @@ def rotation_matrix(axis, angle=None):
     R1 = jnp.cos(angle) * jnp.eye(3)
     R2 = jnp.sin(angle) * jnp.cross(axis, jnp.identity(axis.shape[0]) * -1)
     R3 = (1 - jnp.cos(angle)) * jnp.outer(axis, axis)
-    return R1 + R2 + R3
+    return R1 + jnp.nan_to_num(R2 + R3)  # if axis=[0, 0, 0], R2 & R3 are NaN
```
ddudt, line 49: `FourierPlanarCurve` is a 2D curve in the X-Y plane, and get rotated into another plane in 3D space. If the curve stays in the X-Y plane (as is the default parameters), then this function gets called with `axis=[0, 0, 0]` and `angle=0`.

f0uriest: Can you check that reverse mode ad works for this? It sometimes has weirdness with overwriting nans

ddudt: This actually doesn't work with forward or reverse mode AD. I spent a bunch of time thinking about this and couldn't come up with a good solution, but I welcome help if you have an idea. 

I still think this PR is worth merging even if we don't fix that issue because: 

1. The old code was returning incorrect values
2. The old code also did not properly handle the case `axis=[0, 0, 0]`
3. In practice this will only arise for vertical field coils, and I don't think that is a high-priority use case

desc/compute/geom_utils.py
```diff
@@ -37,11 +37,14 @@ def rotation_matrix(axis, angle=None):
     -------
     rot : ndarray, shape(3,3)
         Matrix to rotate points in cartesian (X,Y,Z) coordinates
+
     """
+    # FIXME: does not work with AD
```
f0uriest, line 42: You could probably add a cond here that checks if axis is all zero (or otherwise small) and just returns the identity matrix in that case? I think that might solve the AD issue

desc/geometry/core.py
```diff
@@ -17,18 +16,44 @@
...
 
     def __init__(self, name=""):
-        self.shift = jnp.array([0, 0, 0]).astype(float)
-        self.rotmat = jnp.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]]).astype(float)
-        self.name = name
+        self._shift = np.array([0, 0, 0])
+        self._rotmat = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]]).flatten()
+        self._name = name
+
+    @optimizable_parameter
```
f0uriest, line 32: When I implemented these the idea was they would not be optimized, just fixed at creation to make coilsets with shared geometry but fixed offset positions.

One main reason is that these parameters are redundant with the standard fourier coefficients so introduces some extra null space.

I'm open to making them optimizable but wanted to at least mention that

ddudt: Yeah I was going to bring this up at the dev meeting. They are redundant parameters, so making them optimizable means we have to be careful about the constraints. My first thought was to make these constants (like the transforms), but this option was less changes to the original code. 

I think the best solution would be to remove these completely. It might make it a little harder to initialize the geometry you want, but then everything else will be cleaner going forward.

desc/compute/geom_utils.py
ddudt: OK I got that to work, but I had to get rid of the optional `angle` argument and make it always use the norm of `axis` as the angle. The issue is that the derivative of `jax.linalg.norm(x)` is NaN when `x = 0`, in both forward and reverse mode. So we can't call `norm(axis)` outside of the `cond` branch where it is guaranteed to be nonzero.

ddudt:
@f0uriest @dpanici This is ready for review. We still need to make a decision about how to better handle the `Curve.shift` and `Curve.rotmat` attributes, but that can wait for a future PR. They are already treated as optimizable compute params on the `master` branch, and the bug fixes in this PR are the minimal changes necessary to get the existing API functional.

desc/compute/geom_utils.py
f0uriest: I think this is what you want: https://github.com/PlasmaControl/DESC/blob/c837cc544aaaafacec2f54c9f2ccf32156ba6b4c/desc/compute/utils.py#L492-L493

desc/geometry/core.py
```diff
@@ -142,20 +167,20 @@ def compute(
...
+    def translate(self, displacement):
+        """Translate the curve by a rigid displacement in X, Y, Z."""
+        self.shift = self.shift + np.asarray(displacement)
 
-    def rotate(self, axis=[0, 0, 1], angle=0):
-        """Rotate the curve by a fixed angle about axis in xyz coordinates."""
-        R = rotation_matrix(axis, angle)
-        self.rotmat = R @ self.rotmat
+    def rotate(self, axis):
+        """Rotate the curve about axis in X, Y, Z coordinates."""
```
f0uriest, line 175: should mention that the angle is determined by the norm of axis

desc/geometry/core.py
```diff
@@ -17,18 +16,44 @@
...
 
-    _io_attrs_ = ["_name", "shift", "rotmat"]
+    _io_attrs_ = ["_name", "_shift", "_rotmat"]
 
     def __init__(self, name=""):
-        self.shift = jnp.array([0, 0, 0]).astype(float)
-        self.rotmat = jnp.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]]).astype(float)
-        self.name = name
+        self._shift = np.array([0, 0, 0], dtype=float)
+        self._rotmat = np.eye(3, dtype=float).flatten()
```
f0uriest, line 29: any reason for making these all regular np arrays rather than jnp?

desc/geometry/core.py
dpanici: I am slightly confused, so now if I want to rotate about the vector (1,1,0) 90 degrees (pi/4), I should pass in for the axis (1,1,0) * np.sqrt(pi/4 / 2)?

desc/compute/geom_utils.py
```diff
@@ -23,29 +23,33 @@ def reflection_matrix(normal):
     return R
 
 
-def rotation_matrix(axis, angle=None):
+def rotation_matrix(axis):
     """Matrix to rotate points about axis by given angle.
```
dpanici, line 27: need to change this, the angle is not given it is determined by axis, it reads like we are missing a parameter from this description.

desc/compute/geom_utils.py
dpanici: If we can use `safediv` and keep the angle argument, I prefer that as it is much clearer than having the angle be the norm of the axis

dpanici (dismissed):
We should use `safediv` and allow the angle argument again for the rotation matrix, it is cleaner to have the arguments be both axis, angle

desc/compute/geom_utils.py
ddudt: I think I have this working now with the angle argument. I also created a new util function `safenormalize` to normalize vectors to unit length

desc/geometry/core.py
ddudt: I reverted back to jnp but I don't think it matters and we are inconsistent about np vs jnp in other classes. We should probably try to consistently use jnp everywhere.

dpanici: sometimes np is needed though to ensure certain things are treated as static with JAX so blanket jnp might not be what we want, but yea a lot of classes build methods it does not matter, but for the init  and some quantities from build I think it can matter

desc/geometry/core.py
```diff
@@ -143,19 +169,19 @@ def compute(
         return data
 
     def translate(self, displacement=[0, 0, 0]):
-        """Translate the curve by a rigid displacement in x, y, z."""
-        self.shift += jnp.asarray(displacement)
+        """Translate the curve by a rigid displacement in X, Y, Z."""
+        self.shift = self.shift + jnp.asarray(displacement)
 
     def rotate(self, axis=[0, 0, 1], angle=0):
```
ddudt, line 175: This syntax means that the angle kwarg has to be passed in, you can't do `Curve.rotate(axis)` and have the angle be the magnitude of the axis. That was the way the code was originally so I'm leaving it that way for now, but we might want to rethink this.

desc/geometry/core.py
f0uriest: They actually were originally part of `transforms`: https://github.com/PlasmaControl/DESC/commit/65e4001663163621d085ab1381758e7413e0821c

I had to change it to allow vectorized evaluation of `CoilSet` stuff, where the transforms are the same for all members but params are different. We should talk tomorrow about the best way to handle these or possibly remove them.

dpanici:
Add automatic constraint to fix `rotmat` and `shift` as another self consistency check

desc/objectives/linear_objectives.py
```diff
@@ -3238,3 +3238,193 @@ def compute(self, params, constants=None):
...
+        """
+        curve = self.things[0]
+        self._dim_f = curve.shift.size
+
+        self.target, self.bounds = self._parse_target_from_user(
+            self._target_from_user, curve.shift, None, np.arange(self._dim_f)
+        )
+
+        if self._normalize:
+            self._normalization = 1
```
ddudt, line 3314: Really we should add normalization options for `Curve` to the `compute_scaling_factors` function, but for these objectives the normalization will never really matter.

desc/objectives/getters.py
```diff
@@ -214,20 +216,30 @@ def get_NAE_constraints(
...
-    if not _is_any_instance(constraints, BoundaryZSelfConsistency):
-        constraints += (BoundaryZSelfConsistency(eq=eq),)
-    if not _is_any_instance(constraints, FixLambdaGauge):
-        constraints += (FixLambdaGauge(eq=eq),)
-    if not _is_any_instance(constraints, AxisRSelfConsistency):
-        constraints += (AxisRSelfConsistency(eq=eq),)
-    if not _is_any_instance(constraints, AxisZSelfConsistency):
-        constraints += (AxisZSelfConsistency(eq=eq),)
+    # Equilibrium
+    if hasattr(thing, "_Rb_lmn") and hasattr(thing, "_Zb_lmn"):
```
ddudt, line 226: Is this a good enough way to check for class types? I didn't want to use `isinstance` to avoid circular imports.

ddudt: Maybe checking for `"_Psi"` would be better for Equilibrium? Although I think the surface classes use `"_R_lmn"` instead of `"Rb_lmn"`

desc/geometry/core.py
dpanici: I think being explicit about the angle is my preferred API for this

desc/objectives/linear_objectives.py
dpanici: #863 can be easily done in a separate PR, this one is getting large

ddudt:
Code coverage is low because of the new self-consistency curve constraints that were added but are not able to be tested yet.

f0uriest (approved):

PR #842, opened 2024-01-29
Optimizer.optimize with a single constraint

ddudt:
There was a bug with using `Optimizer.optimize(verbose>0)` and a single constraint that was not part of a tuple or list, because after the optimization it iterates through the constraints to print their values.


f0uriest (approved):

ddudt:
Random jobs keeps failing that are completely unrelated to this very minimal change. Re-running to see if it eventually fixes itself...

dpanici:
I have been having the same issue on multiple of my PRs...

dpanici (approved):

dpanici:
Issue with the test fails might be linked to [this action](https://github.com/Wandalen/wretry.action/issues/126), I've chimed in on the issue and hopefully it will be fixed soon

f0uriest (approved):

PR #846, opened 2024-01-31
Fix tests to comply with latest `pytest` and `pytest-split`

dpanici:
The way pytest catches exceptions and multiple warnings changed, see from the changelog:

> [#9288](https://github.com/pytest-dev/pytest/issues/9288): [warns()](https://docs.pytest.org/en/latest/reference/reference.html#pytest.warns) now re-emits unmatched warnings when the context closes – previously it would consume all warnings, hiding those that were not matched by the function.

which is the reason why the fft warning test fails with `pytest==8.0.0` now (it returned multiple warnings, and now warns() catches only the desired warning thrown and emits any uncatched ones), and I think this is also the reason the `ascii_io` test fails now (the file was never closed and a ResourceWarning was raised, but was ignored, and now it gets emitted and so causes the test fail).

My local pytest version is 7.3.1 while the CI uses 8.0.0. If I upgrade my local pytest to 8.0.0 then my tests fail in the same way the CI does. The CI before last week would install 7.4.4 pytest, not 8.0.0, because `pytest-split` versions `<0.8.2` did not support `pytest==8.0.0`. However on Jan. 29 `pytest-split==0.8.2` was release and that one did support the `pytest==8.0.0` so our CI started installing that one, hence the errors.

Locally we should all update our `pytest-split`  to `0.8.2`and our `pytest` to `8.0.0` (you need to update both, `8.0.0` pytest is not supported by any `pytest-split` versions older than `0.8.2`) That way the way the CI tests behave and our local should match. 

This PR fixes the tests that were a little sloppy anyways, and then restricts the `pytest` in our requirements to be `8.0.0` to avoid these issues of the CI tests not matching our local tests,.

(this is one way to handle this, we could alternatively restrict the pytest and pytest split versions to be what they used to be, but I am in favor of just bumping up these dependency versions so we use newer versions, these are just for testing anyways so not that restrictive, and these updates to `pytest` make sense to me as far as improving our tests)
# Changes
- Restrict `pytest` version to be `~=8.0.0` and `pytest-split==0.8.2`
- add missing file close to ascii io, and make user warning in test more specific, to fix `ResourceWarning` that caused test to fail
- split the `test_fft_warnings` test where the grid and  basis NFP do not match to test for both warnings that are thrown (both `UserWarning` but one from the grid and basis not matching and the other for the nodes not completing 1 full field period), as in the past the test only caught the nodes not completing a full period warning and ignored the other warning, but now it throws errors as it sees the grid and basis warning which was previously discarded)


f0uriest (approved):

devtools/dev-requirements.txt
```diff
@@ -21,12 +21,12 @@ pre-commit
 # testing and benchmarking
 markupsafe == 2.0.1
 nbmake
-pytest >= 5.0.0
+pytest ~= 8.0.0
```
ddudt, line 24: Do we still need it >5.0.0?

dpanici: The `~=` means it must be `>=8.0.0, <=8.1.0`

dpanici:
@daniel-dudt why did you add a commit to this?

ddudt (approved):

PR #852, opened 2024-02-02
Fix for unused kwargs in Equilibrium init

### LINKED ISSUE #850: Catch unrecognized kwargs in `Equilibrium` init
Ran into some weird bugs because I was passing things to `Equilibrium` using the wrong names which meant they were silently ignored.

f0uriest:
Resolves #850


unalmis (approved):

rahulgaur104 (approved):

dpanici (changes requested):
update notebooks

dpanici (approved):

PR #853, opened 2024-02-03
add basic coil objectives

### LINKED ISSUE #863: Add normalization scales for `Curve` classes
To make normalizations work with curves, we have to add normalizations for `Curve` objects to `normalization.py`

### LINKED ISSUE #895: Objective for coil curvature

### LINKED ISSUE #896: Objective for coil torsion

### LINKED ISSUE #897: Objective for coil length

kianorr:
Adds coil length objective

Resolves #895 
Resolves #897 
Resolves #896 
Resolves #863


desc/objectives/_geometry.py
```diff
@@ -459,6 +459,81 @@ def compute(self, params, constants=None):
...
+        # seems like it because it gives the circumference of a circular coil?
+        self._data_keys = ["length"]
+
+        timer = Timer()
+        if verbose > 0:
+            print("Precomputing transforms")
+        timer.start("Precomputing transforms")
+
+        grid = LinearGrid(L=4, M=4)
+        # TODO: what grid is supposed to be used here?
```
dpanici, line 503: Coils use the `zeta` part of `grid.nodes` as the curve parameter `s` , so `grid` should only have toroidal resolution and no poloidal/radial resolution. 

And for `length` calculation if you look in the compute function for it, we calculate it as the integral of `dl/ds` basically

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,81 @@ def compute(self, params, constants=None):
...
+        # TODO: am I supposed to use obj=coil?
```
dpanici, line 504: Yes, that is correct, `obj` is the type of object (or the object itself) that you are getting the profiles (or transforms, etc) for, in order to have the necessary information to calculate the data keys

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,81 @@ def compute(self, params, constants=None):
...
+            timer.disp("Precomputing transforms")
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params, constants=None):
+        """Compute length of coil."""
+        if constants is None:
+            constants = self._constants
+
+        # TODO: gets KeyError with 'rotmat' when computing 'x_s'.
```
dpanici, line 523: if you get this when using the objective function, this is fixed  in #840 , try merging that one in and see if it works

dpanici (changes requested):
Merge #840 and see if it works

kianorr:
> Merge #840 and see if it works

I merged #840 in and the test that I have is passing now.

dpanici:
Might want to merge again, idt any tests should fail from the compute function side

kianorr:
I added a base class for coil objectives so I included curvature in this PR. If this good, then I'll delete #865 and rename this one.

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,290 @@ def compute(self, params, constants=None):
...
+
+        is_mixed_coils = False
+        if isinstance(self.things[0], CoilSet):
+            if isinstance(self.things[0], MixedCoilSet):
+                is_mixed_coils = True
+                coil = self.things[0].coils
+                self._dim_f = len(coil)
+            else:
+                coil = self.things[0].coils[0]
+                self._dim_f = 1
```
f0uriest, line 553: for possibly nested coilsets you can find the total number of coils with the `tree_flatten` util in JAX (imported in desc.backend):
```python
tree_flatten(things, isleaf = lambda x: isinstance(x, Coil) and not isinstance(x, CoilSet))
```

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,290 @@ def compute(self, params, constants=None):
...
+        else:
+            coil = self.things[0]
+            self._dim_f = 1
+
+        if self._grid is None:
+            # TODO: raise error if grid, transforms are not the same size as mixed coils
+            self._grid = (
+                LinearGrid(N=32)
```
f0uriest, line 561: see the default grid resolutions in the `curve.compute` method

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,290 @@ def compute(self, params, constants=None):
...
+        f : float or array of floats
+            Coil length.
+        """
+        if constants is None:
+            constants = self._constants
+
+        if constants["is_mixed_coils"]:
+            coils = self.things[0].coils
+            data = [
+                coils[i].compute(
```
f0uriest, line 612: you can still use the coilset.compute method for mixed coilsets, you don't need to manually loop here

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,290 @@ def compute(self, params, constants=None):
...
+        Both bounds must be broadcastable to to Objective.dim_f
+    """
+
+    def __init__(self, coil, target=None, bounds=None, **kwargs):
+        if target is None and bounds is None:
+            target = 2 * np.pi
+
+        super().__init__(coil, ["length"], target=target, bounds=bounds, **kwargs)
+
+    def build(self, use_jit=True, verbose=1):
```
f0uriest, line 653: don't need to redefine this if its just calling the base class method

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,290 @@ def compute(self, params, constants=None):
...
+                grid=self._grid,
+            )
+
+        return data
+
+
+class CoilLength(_CoilObjective):
+    """Coil length.
+
+    Parameters
```
f0uriest, line 635: should document all args here since this is the docstring that the users will see

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,290 @@ def compute(self, params, constants=None):
...
+        Coil(s) that are to be optimized
+    target : {float, ndarray}, optional
+        Target value(s) of the objective. Only used if bounds is None.
+        Must be broadcastable to Objective.dim_f.
+    bounds : tuple of {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
+        Both bounds must be broadcastable to to Objective.dim_f
+    """
+
+    def __init__(self, coil, target=None, bounds=None, **kwargs):
```
f0uriest, line 647: and include all args here

desc/objectives/_geometry.py
```diff
@@ -459,6 +459,290 @@ def compute(self, params, constants=None):
...
+            Dictionary of constant data, eg transforms, profiles etc. Defaults to
+            self._constants.
+
+        Returns
+        -------
+        f : float or array of floats
+            Coil length.
+        """
+        data = super().compute(params, constants=constants)
+        if isinstance(data, list):
```
f0uriest, line 683: In all cases data here will either be a dict or nested list of dict from possibly nested coilsets, so you can use the `tree_flatten` thing again:

```python
data = tree_flatten(data, is_leaf=lambda x: isinstance(x, dict))[0]
out = jnp.concatenate([dat[key] for dat in data])
```

f0uriest: You could also allow the target/bounds etc to be nested pytrees that match the structure of the coilset and use this same thing to flatten them down to 1d arrays

desc/objectives/_geometry.py
kianorr: I tried not to, but I manually looped because the compute function doesn't take a list of transforms as far as I know (and we want different transforms for each coil). Is there a way around that?

f0uriest: The compute method for a mixed coilset should accept a list of transforms, or a single one which it uses for all of them

kianorr: Oh weird. I must've been doing something wrong before because it's working now. Thanks!

dpanici (changes requested):
Current * Length objective (proxy for HTS cost)

desc/objectives/_geometry.py
f0uriest: If its a regular `CoilSet` that should work fine because you can re-use transforms for each coil. If it's a mixed coilset you'll have to get the transforms for each one by one

kianorr: Okay so then in compute(), should data for all coils in the regular `CoilSet` be computed or should it just be computed for one coil? I guess all of them should be included since each coil would be optimized independently right?

desc/objectives/_geometry.py
```diff
@@ -493,6 +493,346 @@ def compute(self, params, constants=None):
...
+            )
+
+        timer = Timer()
+        if verbose > 0:
+            print("Precomputing transforms")
+        timer.start("Precomputing transforms")
+
+        if is_mixed_coils:
+            transforms = [
+                get_transforms(self._data_keys, obj=coils[i], grid=self._grid[i])
```
dpanici, line 605: @f0uriest using this with mixed?

f0uriest: @kianorr I think something like this will work for getting transforms for nested coilsets. It assumes a `CoilSet` all use the same transform, but a `MixedCoilSet` will need different transforms for each member.

Right now this assumes the grid is the same for all of them, if we have different grids you can modify this to map over the trees of coils and grids simultaneously, see the `tree_map` documentation.

```python
transforms = jax.tree_util.tree_map(lambda x: get_transforms(self._data_keys, obj=x, grid=grid),
                                    self.coils,
                                    is_leaf=lambda x: isinstance(x, _Coil) and not isinstance(x, MixedCoilSet)
                                   )
```

kianorr:
current length and target/bounds as pytrees will be separate PRs

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
+import numbers
+
+import jax
```
f0uriest, line 3: we try not to import jax directly, in order to keep an interface layer in case they change things.

If you need specific functions (like the tree map stuff) import it into `desc.backend` and then here do `from desc.backend import tree_map`

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+            self.things[0],
+            is_leaf=lambda x: isinstance(x, _Coil) and not isinstance(x, CoilSet),
+        )[0]
+
+        # if using single coil, make coils and grid a list so they can be
+        # used with tree_map
+        coils = [coils[0]] if not is_coil_set else coils
+
+        if self._grid is None:
+            NFP = self.NFP if hasattr(self, "NFP") else 1
```
f0uriest, line 112: `self` wouldn't have the NFP attribute, but the coils might, so you probably want to do this for each coil, you can modify the next line to 
```python
get_grid = lambda x: LinearGrid(N=2*x.N+5, NFP=getattr(x, "NFP", 1), endpoint=False)
self._grid = tree_map(get_grid, coils)
```

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+            grid = lambda x: LinearGrid(N=2 * x + 5, NFP=NFP, endpoint=False)
+            self._grid = [grid(coil.N) for coil in coils]
+        elif self._grid.num_rho > 1 or self._grid.num_theta > 1:
+            raise TypeError("Only use toroidal resolution for coil grids.")
```
f0uriest, line 116: this should be a `ValueError` since it's indicating that a value is wrong, a `TypeError` would be if they passed in something besides a grid like an array or int etc.

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+        elif isinstance(self._grid, numbers.Integral):
+            raise TypeError("The inputted grid should be of type `Grid`.")
```
f0uriest, line 118: not sure if we need to throw an error for this? could just do
`self._grid = LinearGrid(N=self._grid)` in this case

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+
+        """
+        from desc.coils import CoilSet
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+        if self._normalize:
+            self._normalization = self._scales["a"]
+
+        self._dim_f = len(self._coils.coils) if isinstance(self._coils, CoilSet) else 1
```
f0uriest, line 269: if `self._coils` is a `MixedCoilSet` made up of other `CoilSet`s then `len(self._coils.coils)` won't give you the number of actual coils.

It might be better to either use the tree flatten thing again here, or just record the length of the flattened coils in the superclass build and reference that here

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+            Coil length.
+        """
+        data = super().compute(params, constants=constants)
+        data = tree_flatten(data, is_leaf=lambda x: isinstance(x, dict))[0]
+        out = jnp.array([dat["length"] for dat in data])
+        return out
+
+
+class CoilCurvature(_CoilObjective):
+    """Coil curvature.
```
f0uriest, line 294: Might want to give a bit more info here, like that it's the local curvature at each point along the coil, and that smaller values mean straighter coils etc, and that its always positive

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+            Coil curvature.
+        """
+        data = super().compute(params, constants=constants)
+        data = tree_flatten(data, is_leaf=lambda x: isinstance(x, dict))[0]
+        out = jnp.concatenate([dat["curvature"] for dat in data])
+        return out
+
+
+class CoilTorsion(_CoilObjective):
+    """Coil torsion.
```
f0uriest, line 401: same as above, should mention that its local to each point and maybe a sentence explaining what torsion is and what it means

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+
+    Parameters
+    ----------
+    coil : CoilSet or Coil
+        Coil(s) that are to be optimized
+    target : {float, ndarray}, optional
```
f0uriest, line 407: the `{a, b}` in a type annotation means that it has to be one of those literal values. IE, in this case it means that `target` should be the literal python object `float`, rather than a number of type `float`. 

In this case, we just care that `target` has a particular type, so the notation would be `target : float, ndarray, optional`

see https://numpydoc.readthedocs.io/en/latest/format.html#parameters

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+        Weighting to apply to the Objective, relative to other Objectives.
+        Must be broadcastable to to Objective.dim_f
+    normalize : bool, optional
+        Whether to compute the error in physical units or non-dimensionalize.
+    normalize_target : bool, optional
+        Whether target and bounds should be normalized before comparing to computed
+        values. If `normalize` is `True` and the target is in physical units,
+        this should also be set to True.
+        be set to True.
+    loss_function : {None, 'mean', 'min', 'max'}, optional
```
f0uriest, line 424: here the curly braces are correct because `loss_function` must be one of the literal strings `"mean", "max", "min"` or `None`

desc/objectives/_coils.py
```diff
@@ -0,0 +1,504 @@
...
+        use_jit : bool, optional
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        super().build(use_jit=use_jit, verbose=verbose)
+
+        if self._normalize:
+            self._normalization = 1 / self._scales["a"] ** 2
```
f0uriest, line 483: i think the units of torsion are just 1/m not 1/m**2

tests/test_objective_funs.py
```diff
@@ -569,6 +572,66 @@ def test(eq):
...
 
+    @pytest.mark.unit
+    def test_coil_length(self):
+        """Tests coil length."""
+
+        def test(coil):
+            obj = CoilLength(coil, weight=1)
+            obj.build()
+            f = obj.compute(params=coil.params_dict)
+            np.testing.assert_allclose(f, 2 * np.pi, rtol=1e-8)
```
f0uriest, line 583: might also want to check that the output has the shape/size we expect it to?

tests/test_objective_funs.py
```diff
@@ -569,6 +572,66 @@ def test(eq):
...
+
+    @pytest.mark.unit
+    def test_coil_curvature(self):
+        """Tests coil curvature."""
+
+        def test(coil):
+            obj = CoilCurvature(coil, weight=1)
+            obj.build()
+            f = obj.compute(params=coil.params_dict)
+            np.testing.assert_allclose(f, 1 / 2, rtol=1e-8)
```
f0uriest, line 603: see above about checking output size to make sure we didn't make any mistakes with the tree flatten/map stuff

tests/test_objective_funs.py
```diff
@@ -569,6 +572,66 @@ def test(eq):
...
+        test(coils)
+        test(mixed_coils)
+        test(nested_coils)
+
+    @pytest.mark.unit
+    def test_coil_curvature(self):
+        """Tests coil curvature."""
+
+        def test(coil):
+            obj = CoilCurvature(coil, weight=1)
```
f0uriest, line 600: might also want to test passing in a grid/list of grids?

desc/objectives/_coils.py
kianorr: That's true. I think using `self._dim_f = len(self._grid)` would accomplish this too, right? 

But also, I'm confused because this passes the nested coils test, unless `nested_coils = MixedCoilSet.from_symmetry(mixed_coils, NFP=4)` doesn't actually create nested coils?

f0uriest: yeah `from_symmetry` i think flattens everything down to a single list: https://github.com/PlasmaControl/DESC/blob/acba92f229c4a28134792707092b588fe56df71b/desc/coils.py#L1052-L1074 To properly test it you might need to do `MixedCoilSet(coilset, coilset)`. It's fine if they're the same coilset without any shift etc, since all we're testing is length/cuvature/torsion which are local to each coil.

desc/objectives/_coils.py
```diff
@@ -0,0 +1,513 @@
...
+            self._grid = [self._grid]
+
+        if self._grid is None:
+            get_grid = lambda x: LinearGrid(
+                N=2 * x.N + 5, NFP=getattr(x, "NFP", 1), endpoint=False
+            )
+            self._grid = [get_grid(coil) for coil in coils]
+        elif np.any([grid.num_rho > 1 or grid.num_theta > 1 for grid in self._grid]):
+            raise ValueError("Only use toroidal resolution for coil grids.")
+        elif isinstance(self._grid, numbers.Integral):
```
f0uriest, line 120: After line 110, self._grid is always either a list or None, so I this will never trigger. I think this check needs to get moved first.
Something like 
```python
# first handle int or None
if isinstance(self._grid, numbers.Integral):
     self._grid = LinearGrid(N=self._grid, endpoint=False)
if self._grid is None:
    get_grid = lambda x: LinearGrid(
                N=2 * x.N + 5, NFP=getattr(x, "NFP", 1), endpoint=False
            )
    self._grid = [get_grid(coil) for coil in coils]   
# then cast to list 
if not isinstance(self._grid, (tuple, list)):
    self._grid = [self._grid]
# then do checks
errorif(any(grid.num_rho>1 ...)

File: desc/objectives/_coils.py (hunk 16 above)
Comment (kianorr) [reply desc/objectives/_coils.py:271]: Okay you're right. So now, with `MixedCoilSet(coils, coils)`, I run into a problem in `compute()`.  Since the default `self._grid` is created based on the flattened `coils`, the grid won't match up with what's required for `self.things`.  e.g. if you have a nested coil set that's like `[CoilSet(n=2), CoilSet(n=2)]` then I think the only grid that works is a list of 2 grids or a single grid. A list of 4 grids doesn't seem to work but maybe I'm doing something wrong

Comment (f0uriest) [reply desc/objectives/_coils.py:271]: I think for the case where the default grid is used you can use tree map over self.coils instead of over the flattened list and that should make sure that it has the right structure

For the case of a user supplied list of grid I think there's another tree util to convert a list to a given tree structure

Comment (kianorr) [reply desc/objectives/_coils.py:271]: By using tree map over self.coils, do you mean when getting the transforms? Or within `if self._grid is None`?

Comment (f0uriest) [reply desc/objectives/_coils.py:271]: Both

Comment (kianorr) [reply desc/objectives/_coils.py:271]:
```python
self._grid = tree_map(
                lambda x: LinearGrid(
                    N=2 * x.N + 5, NFP=getattr(x, "NFP", 1), endpoint=False
                ),
                self.things[0],
                is_leaf=lambda x: isinstance(x, _Coil) and not isinstance(x, CoilSet),
            )
print(self._grid)```

For whatever reason, this returns `self.things[0]` and not a list of `LinearGrid`. Is this some weird jax thing that I'm not understanding?

Comment (f0uriest) [reply desc/objectives/_coils.py:271]: Ahhh I think I see the issue. It's actually returning a `MixedCoilSet` where instead of coils it's made up of grids, because the output of `tree_map` has the same structure as the input. Let me think about this.

Comment (kianorr) [reply desc/objectives/_coils.py:271]: ohhh yeah you're right. lol
image

Comment (f0uriest) [reply desc/objectives/_coils.py:271]: ok I think this works:
```python

def get_grid(coilset):
    if isinstance(coilset, desc.coils.MixedCoilSet):
        return [get_grid(x) for x in coilset]
    elif isinstance(coilset, desc.coils.CoilSet): # must be regular coilset
        return get_grid(coilset.coils[0]) # use same grid for all members
    else: # must be base coil
        return LinearGrid(
                    N=2 * coilset.N + 5, NFP=getattr(coilset, "NFP", 1), endpoint=False
                )

self._grid = get_grid(self.things[0])

```

the same basic approach should work for transforms

kianorr: wow that is so smart. I'll try it

kianorr: Okay it's a mess but it's working

dpanici (changes requested):
I still owe this a proper review but could a unit test be added that will just optimize a single coil to for a certain length (and maybe one for curvature and torsion too)? Should be a simple optimization so it should be quick especially if the coil resolution is kept very small (so the correct answer can be tested for)

kianorr:
@dpanici okay so maybe the tests could be (with a torus equilibrium) a coil with additional length, curvature, or torsion that is not characteristic of a circle and then this coil becomes a circle after optimization?

I'm not sure how to actually make coils with the params in `params_dict` so i should probably look into that

dpanici:
> @dpanici okay so maybe the tests could be (with a torus equilibrium) a coil with additional length, curvature, or torsion that is not characteristic of a circle and then this coil becomes a circle after optimization?
> 
> I'm not sure how to actually make coils with the params in `params_dict` so i should probably look into that

There is no need for an equilibrium, you should be able to optimize non-equilibrium objects directly like a coil (I believe at least)

f0uriest:
I'd vote on not adding any more tests for this right now. Probably better to wait until #857 and #726 are in and add a realistic test.

f0uriest:
Maybe just manually run what @dpanici suggested to make sure it works but no need to add it to our regular test suite

dpanici:
> Maybe just manually run what @dpanici suggested to make sure it works but no need to add it to our regular test suite

They are pretty simple though and take 5 seconds to run, I think it is worth having in this PR (not to mention I made them before I saw your comment...)

tests/test_examples.py
```diff
@@ -1050,3 +1054,37 @@ def test_example_get_current(self):
...
+    grid = LinearGrid(N=10)
+    target = 2 * coil.compute("curvature", grid=grid)["curvature"]
+    obj = ObjectiveFunction(CoilCurvature(coil, target=target, grid=grid))
+    opt.optimize([coil], obj, maxiter=300)
+    np.testing.assert_allclose(
+        coil.compute("curvature", grid=grid)["curvature"], target, rtol=1e-3
+    )
+
+    # torsion
+    # this test is not working very well... unsure of why
```
f0uriest, line 1083: constant torsion and curvature is a helix, but if the pitch isn't right then it might not be periodic, so that might be the issue. 

A simpler test would be to initialize with some non-planar shape and target zero torsion, which should make it planar

kianorr:
thanks @dpanici :) that would have taken me much longer to get correctly I think

tests/test_examples.py
dpanici: True, thanks

kianorr:
@dpanici @f0uriest anything else I need to do with this PR?

desc/objectives/_coils.py
```diff
@@ -0,0 +1,591 @@
...
+
+        def get_dim_f(coilset):
+            """Get dim_f."""
+            if isinstance(coilset, list):
+                [get_dim_f(x) for x in coilset]
+            elif isinstance(coilset, MixedCoilSet):
+                [get_dim_f(x) for x in coilset]
+            elif isinstance(coilset, CoilSet):
+                get_dim_f(coilset.coils)
+            elif isinstance(coilset, LinearGrid):
```
dpanici, line 116: does this ever get called on a coilset? seems like it would just return None? Since the bottom of the recursion is a `Coil` object, not a `LinearGrid`

kianorr: No this only gets called on a `MixedCoilSet` or `CoilSet` container containing LinearGrids (`self._grid`). Since we're using `tree_map` to map the structure of `self.things[0]` to the inputted grid, the inputted grid is contained within a `CoilSet` or `MixedCoilSet`. e.g. if I input `CoilSet(coil, n=2)`, then `coilset.coils` is `[LinearGrid, LinearGrid]`

dpanici (approved):

desc/objectives/_coils.py
```diff
@@ -0,0 +1,591 @@
...
+            )
+        elif isinstance(self._grid, LinearGrid):
+            # map inputted single LinearGrid to structure of inputted coils
+            self._grid = [self._grid] * len(coil_leaves)
+            self._grid = tree_unflatten(coil_structure, self._grid)
+        else:
+            # this case covers an inputted list of grids that matches the size
+            # of the inputted coils. Can be a 1D list or nested list.
+            flattened_grid = tree_flatten(
+                self._grid, is_leaf=lambda x: isinstance(x, LinearGrid)
```
f0uriest, line 162: I think this `isinstance` check and the one in the elif above should check if it's a base `_Grid`, not just `LinearGrid`

desc/objectives/_coils.py
```diff
@@ -0,0 +1,591 @@
...
+            )[0]
+            self._grid = tree_unflatten(coil_structure, flattened_grid)
+
+        timer = Timer()
+        if verbose > 0:
+            print("Precomputing transforms")
+        timer.start("Precomputing transforms")
+
+        transforms = tree_map(
```
f0uriest, line 171: can this be moved to after line 180? For the case where we only have 1 unique grid for a coilset we probably don't have to compute all the extra transforms

desc/objectives/_coils.py
```diff
@@ -0,0 +1,591 @@
...
+        Whether to compute the error in physical units or non-dimensionalize.
+    normalize_target : bool, optional
+        Whether target and bounds should be normalized before comparing to computed
+        values. If `normalize` is `True` and the target is in physical units,
+        this should also be set to True.
+        be set to True.
+    loss_function : {None, 'mean', 'min', 'max'}, optional
+        Loss function to apply to the objective values once computed. This loss function
+        is called on the raw compute value, before any shifting, scaling, or
+        normalization. Note: Has no effect for this objective.
```
f0uriest, line 395: Loss function can have an effect for all of these

f0uriest: though maybe include a note that its the mean/min/max over all coils, not per coil

f0uriest (changes requested):
Few minor fixes

desc/objectives/_coils.py
kianorr: I'm not sure it could be since tree_map requires self._grid and self.things[0] to be of the same structure, right? so if self._grid is just one of the leaves then tree_map wouldn't work I think

kianorr:
Not sure why this popped up now but tests weren't passing because transforms were in a list and there were no quad_weights. So I added `"quad_weights": None` to fix it. Not sure what the quad_weights should be

f0uriest:
> Not sure why this popped up now but tests weren't passing because transforms were in a list and there were no quad_weights. So I added `"quad_weights": None` to fix it. Not sure what the quad_weights should be

quad_weights are basically the ds along each curve. Not needed for length since it's a scale, I'm not sure if they should be included for torsion and curvature, we can discuss tomorrow

f0uriest (approved):
Should get another review from @dpanici before merging

ddudt:
> Should get another review from @dpanici before merging

I would like to look at it too, if you don't mind waiting a day

kianorr:
> I would like to look at it too, if you don't mind waiting a day

Of course, go for it but I'd like to merge it sooner rather than later

desc/objectives/_coils.py
```diff
@@ -0,0 +1,599 @@
...
```
dpanici, line 403: doc still says it has no effect even though it does

dpanici (changes requested):
change the docstring to address Rory's previous comment on loss functions

kianorr:
> change the docstring to address Rory's previous comment on loss functions

my bad, forgot to change it for all of them. thanks

dpanici:
Instead of "Affects" can you say "operates over all coils, not each individial coil"? to make it more clear that it will make the error a scalar value, not a single value per coil

dpanici (approved):

desc/objectives/_coils.py
```diff
@@ -0,0 +1,599 @@
...
+        "auto" selects forward or reverse mode based on the size of the input and output
+        of the objective. Has no effect on self.grad or self.hess which always use
+        reverse mode and forward over reverse mode respectively.
+    grid : Grid, optional
+        Collocation grid containing the nodes to evaluate at.
+    name : str, optional
+        Name of the objective function.
+    """
+
+    _scalar = False  # Not always a scalar, if a coilset is passed in
```
ddudt, line 415: This is never a scalar, even for a single coil, right? Because it's the local curvature

desc/objectives/_coils.py
```diff
@@ -0,0 +1,599 @@
...
```
ddudt, line 530: This is never a scalar, even for a single coil, right? Because it's the local torsion

tests/test_examples.py
```diff
@@ -1301,3 +1305,42 @@ def test_example_get_current(self):
...
+    np.testing.assert_allclose(
+        coil.compute("length")["length"], target_length, rtol=1e-4
+    )
+    np.testing.assert_allclose(
+        coil.compute("curvature", grid=grid)["curvature"], target_curvature, rtol=1e-4
+    )
+
+    # torsion
+    # initialize with some torsion
+    coil.Z_n = coil.Z_n.at[0].set(0.1)
```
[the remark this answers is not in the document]
f0uriest: this is the jax version of `coil.Z_n[0] = 0.1`

tests/test_examples.py
```diff
@@ -1301,3 +1305,42 @@ def test_example_get_current(self):
...
+
+@pytest.mark.unit
+def test_single_coil_optimization():
+    """Test that single coil (not coilset) optimization works."""
+    # testing that the objectives work and that the optimization framework
+    # works when a single coil is passed in.
+
+    opt = Optimizer("fmintr")
+    coil = FourierRZCoil()
+    coil.change_resolution(N=1)
```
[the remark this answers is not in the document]
f0uriest: `FourierRZCurve` is non-planar, so there are multiple shapes that have constant curvature, so I don't think there is an "expected" value for any of the coefficients.

desc/objectives/_coils.py
kianorr: Yeah you're correct. You're just referring to the comment right?

ddudt: Yes, so just remove this comment (and same for the torsion objective)

tests/test_examples.py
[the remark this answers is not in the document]
ddudt: The coil is guaranteed to be planar since we are only allowing it to have a single Fourier mode. And in this test the `coil.Z_n` stays exactly at 0 and `coil.R_n` changes from `[10, 0]` to approximately `[9, 0]` (see picture: blue is the original coil, red is optimized)

image

dpanici (approved):

f0uriest (approved):

ddudt (approved):

PR #856, opened 2024-02-04
Fix `NaN` on-axis when a grid with an axis point is passed to  `ForceBalance` Objective

dpanici:


f0uriest:
Why is it Nan on axis? I thought it was well defined

unalmis:
> Why is it Nan on axis? I thought it was well defined

just need to refactor `fb`
https://github.com/PlasmaControl/DESC/blob/7240052096c51ef882b23c0d7370aaef5f2b0d53/desc/objectives/_equilibrium.py#L189
pushing the Jacobian into helical basis with `e^helical*sqrt(g)` at axis `= (data["B^zeta"] * data["e^theta*sqrt(g)"].T).T` should work

dpanici:
@unalmis Thanks!

desc/compute/_equil.py
```diff
@@ -562,6 +562,26 @@ def _e_sup_helical(params, transforms, profiles, data, **kwargs):
...
+    dim=3,
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["B^theta", "B^zeta", "e^theta*sqrt(g)", "e^zeta"],
+)
+def _e_sup_helical_times_sqrt_g(params, transforms, profiles, data, **kwargs):
+    data["e^helical*sqrt(g)"] = (
+        data["B^zeta"] * data["e^theta*sqrt(g)"].T - data["B^theta"] * data["e^zeta"].T
```
unalmis, line 580: Don't forget to multiply `this_term = data["B^theta"] * data["e^zeta"].T` by Jacobian. `this_term * sqrt(g)` will be computed as zero at magnetic axis, but you want it to be valid everywhere. Then it's good to merge.

Might be good to `git cherry-pick eb09288fc54c31a08a3d51630819ec80b9dca071` so that eb09288fc54c31a08a3d51630819ec80b9dca071 is merged in this PR.

dpanici: ah good point thanks

unalmis (approved):

f0uriest (approved):

PR #857, opened 2024-02-05
Allow Optimization with OptimizableCollections to Work

### LINKED ISSUE #860: Make optimization framework work with `OptimizableCollection`
In order to make the framework compatible with `OptimizableCollection`
-  https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/objectives/objective_funs.py#L206 change `is_leaf` to be `and not isinstance(x, OptimizableCollection)` in order to flatten to base Optimizable objects (maybe need to ensure that OptimizableCollection has no params to itself)
- Keep FixParameter as is to only be for a fixing a single object
- make a FixCollectionParameter for use in fixing a colletion of objects which share the same parameter (enforce they are the same subthings i.e. all same coiil type in a coilset?)
- change the things inside of optimizable.py  for everything leading up to the `split_idx` at [here](https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/optimizable.py#L197) e.g. at https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/optimizable.py#L57 and all downstream lines from that to the `jnp.split` call in the `OptimizableCollection`
  - to avoid jax complaining about non-static indices

dpanici:
Resolves #860 
This is an inelegant attempt to get OptimizableCollection optimization working, it currently does work for the simple test case I have in here so it could be a possible way towards implementing this?

- modified `FixParameter` to work with OptimizableCollection, though the API needs to be hammered out and still need to decide how to handle all the different cases that things may be passed into FixParameter for an OptimizableCollcetion (i.e. a list of lists of params of len(collection), or maybe just one param? I think we should enforce that it is a list of lists of length(collection) though as parsing without that assumption is hard), currently I think it should accept a list of lists of length collection, and if nothing is being fixed for a particular subthing in the collection, then that corresponding entry (say, the ith subthing) params[i] = [ ]  (an empty list). 
- Modified Factorize_linear_constraints to work with at least this one example of optimizing with OptimizableCollection
 - The jacobian calculation from the constraint seemed to return a tuple though which was weird... so I must be missing something there
- modified `OptimizableCollection.unpack_params` to work with `jit`, I think the best solution would be to have the init of an `OptimizableCollection` subclass (or really, it should be in the `init` of the `OptimizableColleciton` itself) create the `self._split_idx` array as a numpy array (to be considered static by JAX during the jit and thus allowed to be used as indexing in `unpack_params`, currently I have this attribute created in the `SumMagneticField` init which feels not the neatest way, as it really only needs to know about `OptimizableCollection` stuff. 

I have a dummy objective that just takes in a sum of fields in this PR as every other PR with a sensible objective is not yet in master.

Some of the issues to be addressed (feel free to add to these)
- factorize_linear_constraints the `A` matrix construction assumes things.dimensions is not a list
- How to use `FixParameter` objects when `thing` is actually an `OptimizableCollection`? Should it be passed the `thing` then for the params a list of the usual argument for each `subthing` in `things` e.g. `FixParameter(sumfield,params=[["B0",R0"], ["Phi_mn"]])` for `sumfield` being a `SumMagneticField` of a `ToroidalMagneticField` and a `FourierCurrentPotentialField`?
  - And fixing anywhere that assumes FixParameter is not acting on a list of things
  - Or making a new objective for fixing collections of objects?
- `unpack_params` for `OptimizableCollection` as implementing in master currently is not jittable, I think there is a way around it but we'd need to define the `split_idx` in the init of the `OptimizableCollection` objects as an `np.array` so JAX treats as static array when it is jitted

@f0uriest @ddudt I am not claiming this entirely as my own PR, just wanted to get the conversation started and see if I could make any progress

### DIFF
--- a/desc/magnetic_fields.py
+++ b/desc/magnetic_fields.py
@@ -475,6 +475,12 @@ class SumMagneticField(_MagneticField, MutableSequence, OptimizableCollection):
             [type(field) for field in fields]
         )
         self._fields = fields
+        offset = jnp.concatenate(
+            [jnp.array([0]), jnp.cumsum(jnp.array([s.dim_x for s in self]))[:-1]]
+        )
+        # store split indices for unpacking
+        # as a numpy array to avoid jax issues later
+        self._split_idx = np.asarray(offset)
 
     def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
         """Compute magnetic field at a set of points.
--- a/desc/objectives/utils.py
+++ b/desc/objectives/utils.py
@@ -101,8 +101,9 @@ def factorize_linear_constraints(constraints, objective):  # noqa: C901
                 A_per_subthing = []
                 # it is an optimizable collection, loop through its sub-objects
                 for subthing in thing:
-                    if subthing in con.things:
-                        A_ = con.jac_scaled(*xz)[con.things.index(subthing)]
+                    if subthing in con._subthings:
+                        # this returns a tuple of length 1 for some reason?
+                        A_ = con.jac_scaled(*xz)[0][con._subthings.index(subthing)]
                     else:
                         A_ = {
                             arg: jnp.zeros((con.dim_f, dimx))
--- a/desc/objectives/utils.py
+++ b/desc/objectives/utils.py
@@ -178,8 +178,6 @@ def factorize_linear_constraints(constraints, objective):  # noqa: C901
     xp_ = objective.unpack_state(xp, False)
     for con in constraints:
         xpi = [xp_[i] for i, t in enumerate(objective.things) if t in con.things]
-        if not xpi:
-            continue  # if no particular solution bc things is not constrained
         y1 = con.compute_unscaled(*xpi)
         y2 = con.target
         y1, y2 = np.broadcast_arrays(y1, y2)
--- a/desc/objectives/utils.py
+++ b/desc/objectives/utils.py
@@ -8,6 +8,7 @@ import warnings
 import numpy as np
 
 from desc.backend import cond, jnp, logsumexp, put
+from desc.optimizable import OptimizableCollection
 from desc.utils import Index, flatten_list, svd_inv_null
 
 
@@ -81,19 +82,42 @@ def factorize_linear_constraints(constraints, objective):  # noqa: C901
         xz = _tree_zeros_like([t.params_dict for t in con.things])
         # computing A matrix for each constraint for each thing in the optimization
         for thing in objective.things:
-            if thing in con.things:
-                A_ = con.jac_scaled(*xz)[con.things.index(thing)]
+            if not isinstance(thing, OptimizableCollection):
+                if thing in con.things:
+                    A_ = con.jac_scaled(*xz)[con.things.index(thing)]
+                else:
+                    A_ = {
+                        arg: jnp.zeros((con.dim_f, dimx))
+                        for arg, dimx in thing.dimensions.items()
+                    }
+                args = (
+                    objective._args
+                    if prox_flag and isinstance(thing, Equilibrium)
+                    else thing.optimizable_params
+                )
+                A_per_thing.append(jnp.hstack([A_[arg] for arg in args]))
+
             else:
-                A_ = {
-                    arg: jnp.zeros((con.dim_f, dimx))
-                    for arg, dimx in thing.dimensions.items()
-                }
-            args = (
-                objective._args
-                if prox_flag and isinstance(thing, Equilibrium)
-                else thing.optimizable_params
-            )
-            A_per_thing.append(jnp.hstack([A_[arg] for arg in args]))
+                A_per_subthing = []
+                # it is an optimizable collection, loop through its sub-objects
+                for subthing in thing:
+                    if subthing in con.things:
+                        A_ = con.jac_scaled(*xz)[con.things.index(subthing)]
+                    else:
+                        A_ = {
+                            arg: jnp.zeros((con.dim_f, dimx))
+                            for arg, dimx in subthing.dimensions.items()
+                        }
+                    args = (
+                        objective._args
+                        if prox_flag and isinstance(subthing, Equilibrium)
+                        else subthing.optimizable_params
+                    )
+                    # form the A per each sub-object in the collection
+                    A_per_subthing.append(jnp.hstack([A_[arg] for arg in args]))
+                # the A for the full object will just be the stacking of each sub-A
+                A_per_thing.append(jnp.hstack([A__ for A__ in A_per_subthing]))
+
         # using obj.compute instead of obj.target to allow for correct scale/weight
         b_ = -con.compute_scaled_error(*xz)
         A.append(A_per_thing)
@@ -153,6 +177,8 @@ def factorize_linear_constraints(constraints, objective):  # noqa: C901
     xp_ = objective.unpack_state(xp, False)
     for con in constraints:
         xpi = [xp_[i] for i, t in enumerate(objective.things) if t in con.things]
+        if not xpi:
+            continue  # if no particular solution bc things is not constrained
         y1 = con.compute_unscaled(*xpi)
         y2 = con.target
         y1, y2 = np.broadcast_arrays(y1, y2)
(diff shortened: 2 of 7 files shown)


desc/magnetic_fields.py
```diff
@@ -478,6 +478,12 @@ def __init__(self, *fields):
             [type(field) for field in fields]
         )
         self._fields = fields
+        offset = jnp.concatenate(
+            [jnp.array([0]), jnp.cumsum(jnp.array([s.dim_x for s in self]))[:-1]]
+        )
+        # store split indices for unpacking
+        # as a numpy array to avoid jax issues later
+        self._split_idx = np.asarray(offset)
```
dpanici, line 486: Way to put this in a super init for `OptimizableCollection`? otherwise this would have to be done in every init of subclasses of `OptimizableCollection`

tests/test_examples.py
```diff
@@ -943,3 +945,32 @@ def test_example_get_current(self):
...
+    objective = ObjectiveFunction(
+        DummyFields(
+            field,
+            eq,
+            target=2,
+            eval_grid=LinearGrid(rho=np.array(1.0), theta=np.array(0.0), N=1),
+        )
+    )
+    optimizer = Optimizer("lsq-exact")
+    constraints = (FixParameter(field, [["R0"], []]),)
```
dpanici, line 969: Currently I think the API I have setup (have not changed docs yet) is that if it is an optimizableCollection, you MUST pass in params that is a list of lists, the outer list being `len(OptimizableCollection)`, to avoid ambiguity in what is being input. Then if you say only want to fix one subthing's param, like in the case here, you leave the other list of params empty. 

This case is fixing only the `"R0"` of the Tf field but leaving its `"B0"` and the vertical field's `"B0"` free

desc/objectives/utils.py
```diff
@@ -81,19 +82,43 @@ def factorize_linear_constraints(constraints, objective):  # noqa: C901
...
-                if prox_flag and isinstance(thing, Equilibrium)
-                else thing.optimizable_params
-            )
-            A_per_thing.append(jnp.hstack([A_[arg] for arg in args]))
+                A_per_subthing = []
+                # it is an optimizable collection, loop through its sub-objects
+                for subthing in thing:
+                    if subthing in con._subthings:
+                        # this returns a tuple of length 1 for some reason?
+                        A_ = con.jac_scaled(*xz)[0][con._subthings.index(subthing)]
```
dpanici, line 106: not sure why I needed a 0 index here, but it `con.jac_scaled(*xz)` was returning a tuple of length 1 that contained a list of dicts of length(con._subthings), so I had to index into the tuple here. I am really tired so likely i just am missing something stupid here

desc/objectives/utils.py
```diff
@@ -81,19 +82,43 @@ def factorize_linear_constraints(constraints, objective):  # noqa: C901
...
+                    else:
+                        A_ = {
```
dpanici, line 108: This is completely untested right now (way I have it setup as of commit 237da24f3a is that `FixParameter` has in its `self._subthings` list every single object contained in the `OptimizableCollection` it has as its `self.things[0]`, so even though I am not fixing anything for the VF field in the example, it is still counted in the first logic branch above this line and the else is not triggered

desc/objectives/utils.py
f0uriest: iirc `con.jac_scaled` returns a tuple of dict of array, where the tuple corresponds to how many sets of params the constraint takes, and each dict is the derivative wrt that set of params

tests/test_examples.py
f0uriest: something like this might be easier? https://jax.readthedocs.io/en/latest/jax-101/05.1-pytrees.html#key-paths

desc/magnetic_fields.py
f0uriest: what is this needed for? I think there already is something similar in `OptimizableCollection`

f0uriest:
Just as a general comment for this type of stuff, the jax pytree utils are usually super helpful and a lot easier than rolling our own versions of similar stuff:
https://jax.readthedocs.io/en/latest/jax.tree_util.html

desc/magnetic_fields.py
dpanici: There was, but it used `jnp.split` and jax threw an error complaining that it could not deal with non-static indices, and it seems that [`jnp.split` cannot be jitted, but if you use an np array as the indices it works ](https://github.com/google/jax/issues/13769#issuecomment-1496709173)

dpanici:
> Just as a general comment for this type of stuff, the jax pytree utils are usually super helpful and a lot easier than rolling our own versions of similar stuff: https://jax.readthedocs.io/en/latest/jax.tree_util.html

I will give this a look tomorrow, thanks

desc/magnetic_fields.py
f0uriest: hmmm ok. We could add an init method to do this, and then call super init from all the subclasses. It might also be possible to do some metaclass stuff like we do to register things as pytrees:
https://github.com/PlasmaControl/DESC/blob/969774a4ef082689bf6a9f1a222be6e3f3a47d34/desc/io/equilibrium_io.py#L107

 Another option might be something like we do in the `Optimizable` class where it gets built the first time its called:
https://github.com/PlasmaControl/DESC/blob/969774a4ef082689bf6a9f1a222be6e3f3a47d34/desc/optimizable.py#L18

dpanici:
- make FixParameter only act on single objects 0> neater that way
- make a separate class to fix all currents of  a given coilset, for instance (FixParameters, FixCollectionParameters...)
 - Fix all subthings that have the parameter we are asking to constrain
 - have it basically call FixParameter for each single object, and then stack all its A matrices like #528 
- ShareParameter constraint?
 - One idea is could  have as its thing the optimizable collection, and therefore when it makes the A it knows of all the params of the subthings and its orderings, it could make the A matrix like we do now, and then inside of factorize linear when we check subthing in things we would also then check for thing in things and that would create the right A matrix... but no that is not for the right arg? or maybe this does work I need to think more
 - How to handle indexing? 
- How does optimizer know which subthings are actually being constrained?
 - https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/objectives/objective_funs.py#L206 change `is_leaf` to be `and not isinstance(x, OptimizableCollection)` in order to flatten to base Optimizable objects (maybe need to ensure that OptimizableCollection has no params to itself)
 - Rory wants to ignore factorize linear constraints, instead put all linear constraints into a single ObjectiveFunction, and get x vector from it(ensure it only has unique parts)
  - total A = jac(objective)
  - b = jac(objective)(zeros)
  - and add a `combine_args` call for the objective and the linear constraint ObjectiveFunction as well (move before the linear constraints get factorized in the projection)

So action items @ddudt 

- create `ObjectiveFunction` inside of `optimizer.optimize` that takes in the linear constriants
- call `combine_args` on it and objective before we do the linear constraint projection
- now `factorize_linear_constraints` argument `self._constraints` in (https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/optimize/_constraint_wrappers.py#L94) is now just an objective function, use its jac_scaled argument to get `A_full` in (https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/objectives/utils.py#L102) and call obj.compute(zeros) to get `b_full`
- keep the indexing stuff after that line above, and everything should work fine for optimizable optimization

for OptimizableCollection stuff
-  https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/objectives/objective_funs.py#L206 change `is_leaf` to be `and not isinstance(x, OptimizableCollection)` in order to flatten to base Optimizable objects (maybe need to ensure that OptimizableCollection has no params to itself)
- Make FixParameter only be for a single object (so reverting what I have done in this PR)
- make a FixCollectionParameter for use in fixing a colletion of objects which share the same parameter (enforce they are the same subthings i.e. all same coiil type in a coilset?)
- change the things inside of optimizable.py  for everything leading up to the `split_idx` at [here](https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/optimizable.py#L197)e.g. at https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/optimizable.py#L57
 - to avoid jax complaining about non-static indices

dpanici:
#956 takes care of this

PR #864, opened 2024-02-07
CoilSet with NFP and sym

### LINKED ISSUE #858: `CoilSet` Stellarator symmetry and NFP Symmetry logic
- Add `NFP` and `sym` as attributes to `CoilSet`
- Make `compute_magnetic_field` function of `CoilSet` use these to then handle these symmetries (maybe start with just the `NFP` symmetry)

ddudt:
Resolves #858 

TODO:

- [x] Get coil set working with multiple field period symmetry (`NFP > 1`)
- [x] Get coil set working with stellarator symmetyr (`sym = True`)
- [x] Get it working in both $(R,\phi,Z)$ and $(X,Y,Z)$ coordinates


desc/magnetic_fields.py
```diff
@@ -170,21 +170,21 @@ def __sub__(self, x):
         return self.__add__(-x)
 
     @abstractmethod
-    def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
+    def compute_magnetic_field(self, grid, params=None, basis="rpz", source_grid=None):
```
ddudt, line 173: I thought our old notation of `coords` vs `grid` was confusing, so I updated it for every `compute_magnetic_field` function. "coords" sounds like coordinate system, but that is what the `basis` argument is for. Now `grid` is the grid where the magnetic field is being evaluated at, and `source_grid` is the optional grid for discretizing the current source. Both can be either `Grid` types or an array of the grid nodes.

desc/coils.py
```diff
@@ -829,43 +853,74 @@ def flip(self, *args, **kwargs):
...
+                B += self[0].compute_magnetic_field(
+                    grid_nfp, params=x, basis="rpz", source_grid=source_grid
+                )
+                return B, None
+
+            B += scan(body, jnp.zeros(grid_nfp.shape), tree_stack(params))[0]
+
+        # sum the magnetic fields from both halves of the symmetric field period
+        if self.sym:
+            B = B[: grid.shape[0], :] + B[grid.shape[0] :, :] * jnp.array([-1, 1, 1])
```
ddudt, line 920: TBH I don't understand why I have to subtract the B_R component instead of summing it like the B_phi and B_Z components. Maybe it would change to a sum with a different reflection for the symmetric nodes? But this seems to work.

desc/coils.py
```diff
@@ -829,43 +853,74 @@ def flip(self, *args, **kwargs):
...
             for par, coil in zip(params, self):
                 par["current"] = coil.current
-        if hasattr(coords, "nodes"):
-            coords = coords.nodes
-        coords = jnp.atleast_2d(coords)
 
-        def body(B, x):
-            B += self[0].compute_magnetic_field(
-                coords, params=x, basis=basis, grid=grid
+        # stellarator symmetry is easiest in [X,Y,Z] coordinates
```
ddudt, line 886: I found it was easiest to convert back and forth between the two different coordinate systems to use whichever one is most convenient in the moment, rather than doing all the computations in whichever basis the user input. These conversions are small constant matrix multiplications so I figure there isn't much downside to this approach.

tests/test_coils.py
```diff
@@ -24,51 +26,110 @@ class TestCoil:
...
-        grid = LinearGrid(zeta=100, endpoint=False)
-        B_approx = coil.compute_magnetic_field(
-            Grid([[10, y, 0], [10, -y, 0]]), basis="xyz", grid=grid
-        )[0]
+        B_xyz = coil.compute_magnetic_field(
+            grid_xyz, basis="xyz", source_grid=coil_grid
+        )
+        B_rpz = coil.compute_magnetic_field(
+            grid_rpz, basis="rpz", source_grid=coil_grid
+        )
```
ddudt, line 53: I realized we didn't have any test to check that `compute_magnetic_field(basis="rpz")` gives the same result as `compute_magnetic_field(basis="xyz")`, so I added it here for each coil.

tests/test_coils.py
```diff
@@ -214,9 +287,56 @@ def test_from_symmetry(self):
...
         )
         coils2 = MixedCoilSet.from_symmetry(coils, NFP, True)
-        B_approx = coils2.compute_magnetic_field([10, 0, 0], basis="rpz", grid=32)[0]
+        B_approx = coils2.compute_magnetic_field(
+            [10, 0, 0], basis="rpz", source_grid=32
+        )[0]
         np.testing.assert_allclose(B_true, B_approx, rtol=1e-3, atol=1e-10)
 
+    @pytest.mark.unit
+    def test_symmetry_magnetic_field(self):
```
ddudt, line 296: Summary of what this new test is checking:

`sym_coilset` contains only 3 unique coils in half of one field period. It computes it's magnetic field assuming field period symmetry and stellarator symmetry, using the new code that was added. 

`asym_coilset` contains 24 coils but should be equivalent to the other coil set (3 unique coils, x2 for stellarator symmetry, x4 field periods). It computes it's magnetic field the "old school" way from each coil. 

This test checks that both coil sets give the same magnetic fields, in both coordinate systems.

desc/coils.py
```diff
@@ -716,21 +722,29 @@ class CoilSet(OptimizableCollection, _Coil, MutableSequence):
     Parameters
     ----------
     coils : Coil or array-like of Coils
-        collection of coils. Must all be the same type and resolution.
-    currents : float or array-like of float
-        currents in each coil, or a single current shared by all coils in the set
+        Collection of coils. Must all be the same type and resolution.
+    currents : float or array-like of float (optional)
```
f0uriest, line 726: Don't think we need currents as an input since each coil should already have its own current

desc/coils.py
```diff
@@ -716,21 +722,29 @@ class CoilSet(OptimizableCollection, _Coil, MutableSequence):
...
+        Currents in each coil, or a single current shared by all coils in the set.
+    NFP : int (optional)
```
f0uriest, line 728: Should include some more explanation of what it means if NFP>1 or sym=True. IE, that the set should only include the unique coils and they will be duplicated etc.

desc/coils.py
```diff
@@ -146,21 +147,21 @@ def current(self, new):
...
         The coil current may be overridden by including `current`
         in the `params` dictionary.
 
         Parameters
         ----------
-        coords : array-like shape(n,3) or Grid
-            coordinates to evaluate field at [R,phi,Z] or [x,y,z]
-        params : dict, optional
-            parameters to pass to curve
+        grid : array-like shape(n,3) or Grid
```
f0uriest, line 158: personally I prefer to not use `grid` here since that usually means a grid in computational coordinates but I could be convinced otherwise

desc/coils.py
```diff
@@ -829,43 +853,74 @@ def flip(self, *args, **kwargs):
...
-            return B, None
+            grid_xyz = jnp.vstack((grid_xyz, grid_sym))
 
-        B = scan(body, jnp.zeros(coords.shape), tree_stack(params))[0]
+        # field period rotation is easiest in [R,phi,Z] coordinates
+        grid_rpz = xyz2rpz(grid_xyz)
 
+        # sum the magnetic fields from each field period
+        B = jnp.zeros_like(grid_rpz)
+        for k in range(self.NFP):
```
f0uriest, line 907: could also do a scan over NFP which will be faster to compile than a python loop

desc/coils.py
f0uriest: I think it makes sense. When you mirror the coil across the symmetry plane, you reverse all the field components. When you then flip it upside down you reverse Z and phi back to the correct way, but R is still backwards

desc/coils.py
```diff
@@ -950,24 +1005,39 @@ def from_symmetry(cls, coils, NFP, sym=False):
         Parameters
         ----------
         coils : Coil, CoilGroup, Coilset
-            base coil or collection of coils to repeat
-        NFP : int
-            number of field periods
-        sym : bool
-            whether coils should be stellarator symmetric
+            Coil or collection of coils in one field period or half field period.
+        NFP : int (optional)
```
f0uriest, line 1009: Should clarify that this explicitly creates extra coils, not "virtual" coils like the regular init method

desc/coils.py
```diff
@@ -1361,34 +1432,32 @@ def compute(
...
             magnetic field at specified points, in either rpz or xyz coordinates
         """
         params = self._make_arraylike(params)
-        grid = self._make_arraylike(grid)
+        source_grid = self._make_arraylike(source_grid)
 
         B = 0
-        for coil, par, grd in zip(self.coils, params, grid):
-            B += coil.compute_magnetic_field(coords, par, basis, grd)
+        for coil, par, grd in zip(self.coils, params, source_grid):
```
f0uriest, line 1459: dont we need logic here as well to handle NFP and sym?

desc/magnetic_fields.py
f0uriest: I think source grid always be either Grid or None, not an array, since some magnetic field classes need the other info from the grid class.

Also if we want to rename things I'd prefer eval_grid and source_grid to make them both clear.

tests/test_coils.py
```diff
@@ -214,9 +287,56 @@ def test_from_symmetry(self):
...
+        """Tests that compute magnetic field is correct from symmetry."""
+        eq = get("precise_QH")
```
f0uriest, line 298: does this have a non-planar axis? Is that important?

tests/test_coils.py
```diff
@@ -372,7 +492,7 @@ def test_dunder_methods(self):
         del coils1[-2]
         assert len(coils1) == 4
         assert coils1[-1] is coil2
-        assert coils1[-2][0].__class__ is coil1.__class__
+        assert coils1[-2].__class__ is coil1.__class__
```
f0uriest, line 495: why did this change?

f0uriest (dismissed):

desc/coils.py
dpanici: I second the hesitation, I prefer `coords` as it makes it clear it is real space coordinates that are being used versus rho,theta,zeta or `s` or some other computational coordinate, we are pretty consistent with this elsewhere in the code

ddudt: I like `eval_grid`, since that clarifies the difference from `source_grid` and I think that's consistent with our free-boundary notation. Also down the road we have discussed generalizing Grids to be in other coordinate systems. Is everyone good with that?

tests/test_coils.py
ddudt: because before `CoilSet.from_symmetry` was making a `CoilSet` with each member another `CoilSet` for each field period, and now it is making a `CoilSet` where each member is a `Coil`

tests/test_coils.py
ddudt: Yes it does have a non-planar axis. It shouldn't matter, and if anything it is a more complicated/realistic test case.

desc/coils.py
ddudt: Well this depends on what the purpose of the `MixedCoilSet` is. I am of the opinion that it should only be used to combine different types of coils at the highest level, so that objectives expecting a single "field thing" can compute the total magnetic field from a combination of different coils. In other words, I think every sub-member should be responsible for it's own symmetry (if applicable), and the `MixedCoilSet` simply sums all the magnetic fields together. In which case this method can be left the way it is. 

Otherwise it could get weird. Say for example we had a `MixedCoilSet(NFP=4, sym=True)` that was a combination of a `CoilSet` with the same symmetry, another `CoilSet` without symmetry, and a couple of individual coils. How should we decide to apply the symmetry for each sub-member? So I think it's best if this class doesn't know anything about stellarator or field period symmetry.

dpanici: Hm I see the point, I think I agree with Daniel. We should make this explicit in the description of this class though that it does NOT have any symmetry etc and so if it is used you should have to provide either coils covering the entire 2pi in Phi range, or sub coilsets which account for their own symmetries

ddudt: If we go with this plan, we might need to rethink having `MixedCoilSet` be a subclass of `CoilSet`, or at least add checks for `self.NFP = 1` and `self.sym = False` (since it will inherit those attributes from `CoilSet`)

desc/magnetic_fields.py
dpanici: ^ yea I think the `FourierCurrentPotentialField` has to have a `Grid` object since it computes `K` using the `surface.compute` method

desc/coils.py
f0uriest: It can still be a subclass, but we just don't make NFP and sym args at init, so they're always set to 1/False. Should also make it clean in the docstring what this means etc.

desc/coils.py
f0uriest: Yeah I do something like that for the singular integral stuff: https://github.com/PlasmaControl/DESC/blob/b245323b4c177f20f0becea1c055536dfeaf43d8/desc/singularities.py#L364-L369

You'd have the regular base grid and just shift it by i*2pi/NFP where i is the loop step

desc/coils.py
f0uriest: Yes but on the free boundary branch `eval_grid` is always in computational coordinates, so I think its still a bit confusing here.

At the very least we should make it clear that it should be a regular `ndarray`, not a `Grid`, as currently `Grid`s are limited to computational coordinates.

desc/coils.py
```diff
@@ -1298,8 +1388,15 @@ class MixedCoilSet(CoilSet):
     ----------
     coils : Coil or array-like of Coils
         collection of coils
-    currents : float or array-like of float
-        currents in each coil, or a single current shared by all coils in the set
+    NFP : int (optional)
```
dpanici, line 1391: removed NFP and sym from docstring


call super init with the default values here

dpanici:
Change back to coords but make sure it cant accept Grid for that argument

f0uriest (approved):

desc/coils.py
```diff
@@ -563,21 +568,23 @@ def __init__(
...
 
         Parameters
         ----------
-        coords : array-like shape(n,3) or Grid
-            coordinates to evaluate field at [R,phi,Z] or [x,y,z]
-        params : dict, optional
-            parameters to pass to curve
+        coords : array-like shape(n,3)
+            Nodes to evaluate field at in [R,phi,Z] or [X,Y,Z] coordinates.
+        params : dict or array-like of dict, optional
```
dpanici, line 583: This also needs to be changed

desc/equilibrium/equilibrium.py
```diff
@@ -1975,7 +1975,7 @@ def _optimize(  # noqa: C901
                 print("Trust-Region ratio = {:9.3e}".format(tr_ratio[0]))
 
             # perturb + solve
-            (_, predicted_reduction, dc_opt, dc, c_norm, bound_hit,) = optimal_perturb(
+            (_, predicted_reduction, dc_opt, dc, c_norm, bound_hit) = optimal_perturb(
```
dpanici, line 1978: this was a really unnecessary change...

ddudt: I disagree. When you have that extra comma at the end, the black formatter expands the tuple onto multiple lines with each variable on a separate line (at least mine does). So this change reduced the same code from 6 lines down to 1 line.

PR #868, opened 2024-02-11
Setting rho in FourierRZSurface for old DESC equilibria to allow re-solve

rahulgaur104:
I think we should have a test for basic operations like loading and re-solving an old DESC file even if we are not releasing v1.0 anytime soon. 

This test can be used and modified in the future and used as a guide to make old DESC files compatible with the latest master.

All suggestions are welcome!


dpanici (changes requested):
I don't think we need to add an entirely new input file and equilibrium for this, can you just save the SOLOVEV output file, and have the test just load it, solve it with `maxiter=1` and save it to some temporary spot, to ensure all of these happen without errors or warnings?

f0uriest:
We do already include the version of DESC a file was saved with: 
https://github.com/PlasmaControl/DESC/blob/master/desc%2Fio%2Fhdf5_io.py#L301

f0uriest:
Also in this particular case we should probably have added a default value for rho to the `_set_up` method that gets called after loading, like here: https://github.com/PlasmaControl/DESC/blob/e246e9070719b72c7dd8201cfc8879593ab1a878/desc/equilibrium/equilibrium.py#L388-L401

tests/conftest.py
```diff
@@ -256,6 +256,31 @@ def HELIOTRON_vac2(tmpdir_factory):
...
+    cwd = os.path.dirname(__file__)
+    exec_dir = os.path.join(cwd, "..")
+    input_filename = os.path.join(exec_dir, input_path)
+
+    print("Running ESTELL_older test.")
+    print("exec_dir=", exec_dir)
+    print("cwd=", cwd)
+
+    args = ["-o", str(desc_h5_path), input_filename, "-vv"]
+    main(args)
```
f0uriest, line 275: this is re-running the current version of desc, is that what you want?

dpanici (dismissed):
add getter to FourierRZToroidalSurface `rho` property, see `Grid` properties for details

tests/conftest.py
rahulgaur104: This whole function has been removed from conftest.py

ddudt:
@rahulgaur104 cherry-pick this commit: f0514bf7779a542cd42603cf706a512807b5f28f

Also why is this a forked repo and not a regular branch? I was going to make the change myself, but I can't since it's on your other repo.

rahulgaur104:
Yeah, @ddudt  sorry about that and thanks for the rotmat and shift defaults. I created this branch in my fork and realized after creating this PR. Let's keep this one simple. You or I can add a new PR if you want more changes.

dpanici:
> Yeah, @ddudt sorry about that and thanks for the rotmat and shift defaults. I created this branch in my fork and realized after creating this PR. Let's keep this one simple. You or I can add a new PR if you want more changes.

He just meant he already made the change (add defaults to curve) elsewhere so to cherry pick that commit to keep the git history cleaner, but whatever this is also fine

tests/test_equilibrium.py
```diff
@@ -419,3 +419,16 @@ def test_equilibrium_unused_kwargs():
...
     _ = Equilibrium(pressure=pres, current=curr)
+
+
+@pytest.mark.slow
+@pytest.mark.unit
+@pytest.mark.solve
+def test_backward_compatible_load_and_resolve():
+    """Test backwards compatibility of load and re-solve."""
+    with pytest.warns(RuntimeWarning):
+        eq = EquilibriaFamily.load(load_from=".//tests//inputs//NCSX_older.h5")[-1]
```
f0uriest, line 430: minor thing but might want to reduce the resolution a lot after loading to make the compile/solve faster

f0uriest (approved):

tests/test_equilibrium.py
rahulgaur104: Resolved

tests/test_equilibrium.py
```diff
@@ -419,3 +419,20 @@ def test_equilibrium_unused_kwargs():
...
+
+    # reducing resolution since we only want to test eq.solve
+    eq.L = 4
+    eq.M = 4
+    eq.N = 4
```
ddudt, line 434: FYI we have a `change_resolution` function for this

desc/geometry/surface.py
```diff
@@ -158,6 +158,8 @@ def Z_basis(self):
     @property
     def rho(self):
         """float: Flux surface label."""
+        if not (hasattr(self, "_rho")) or self._rho is None:
+            self._rho = 1.0
```
ddudt, line 162: Any reason you didn't do `return self.__dict__.setdefault("_rho", 1)` like we do for all the other classes?

rahulgaur104: No reason.

tests/test_equilibrium.py
rahulgaur104: Thanks!

PR #875, opened 2024-02-13
mgrid I/O

ddudt:
Adds a method `MagneticField.save_mgrid` to write a magnetic field to an mgrid file. Useful for interfacing with other codes.

### DIFF
--- a/desc/magnetic_fields.py
+++ b/desc/magnetic_fields.py
@@ -5,7 +5,7 @@ from collections.abc import MutableSequence
 
 import numpy as np
 from interpax import approx_df, interp2d, interp3d
-from netCDF4 import Dataset, stringtochar
+from netCDF4 import Dataset, chartostring, stringtochar
 
 from desc.backend import fori_loop, jit, jnp, odeint, sign
 from desc.basis import DoubleFourierSeries
@@ -466,7 +466,7 @@ class _MagneticField(IOAble, ABC):
         kp[:] = nphi
 
         nfp = file.createVariable("nfp", np.int32)
-        nfp.long_name = "Number of toroidal field periods."
+        nfp.long_name = "Number of field periods."
         nfp[:] = NFP
 
         nextcur = file.createVariable("nextcur", np.int32)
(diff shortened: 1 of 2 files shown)


desc/magnetic_fields.py
```diff
@@ -374,6 +374,141 @@ def save_BNORM_file(
...
+        path,
+        Rmin,
+        Rmax,
+        Zmin,
+        Zmax,
+        nR=101,
+        nZ=101,
+        nphi=90,
+    ):
+        """Save the magnetic field to an mgrid NetCDF file.
```
f0uriest, line 388: The regular mgrid format allows you to have multiple coil groups with independent currents. We don't necessarily have to support that but we should mention that it only saves as one group with a single current.

Also i think the regular mgrid files also have a field for whether the currents the field was computed with are "raw" or "scaled"? I forget which is which bit might need to add that for codes that expect it

ddudt: Separating the coils into groups or all in one group is arbitrary if you only need the total magnetic field at the end of the day, so I put them all into a single group for simplicity.

ddudt: And the "raw" vs "scaled" is already in here -- it's the `mgrid_mode` variable and I have it hard-coded to "raw" (but the coil current `raw_coil_cur` is also hard-coded to 1)

desc/magnetic_fields.py
```diff
@@ -1058,53 +1197,64 @@ def compute_magnetic_field(self, coords, params=None, basis="rpz", grid=None):
         return B
 
     @classmethod
-    def from_mgrid(cls, mgrid_file, extcur=1, method="cubic", extrap=False):
+    def from_mgrid(cls, mgrid_file, extcur=None, method="cubic", extrap=False):
```
ddudt, line 1200: I changed this so that now by default it will use the coil currents from the mgrid file itself, without the user having to pass them in. This is the typical case for mgrid files in "raw" mode. 

I added a check so that mgrid files in "scaled" mode will throw a `UserWarning` if the currents are not given.

f0uriest (approved):

dpanici (dismissed):
Link to real docs of mgrid, and document specifically raw vs scaled, is B in mgrid with raw the full B or is it B/amps? what does extcur need to be for each?

- if mgrid file is raw, what are the extcur values used for? Should they be one if it is raw? if they are not one are they an additional scaling?
- If that is used in VMEC, and the extcur values are non-one, then does VMEC use B * extcur still? which basically scales the field twice if our understanding of raw vs scaled is correct

f0uriest (approved):

PR #880, opened 2024-02-15
Small Coil and Field Fixes

dpanici:
- fix coil docstrings to match supported inputs
- fix default grid setting for `CurrentPotentialField` and `FourierCurrentPotentialField`

### DIFF
--- a/desc/magnetic_fields.py
+++ b/desc/magnetic_fields.py
@@ -1917,9 +1917,6 @@ class FourierCurrentPotentialField(
             magnetic field at specified points
 
         """
-        coords = coords or LinearGrid(
-            M=self._M_Phi * 4 + 1, N=self._N_Phi * 4 + 1, NFP=self.NFP
-        )
         return _compute_magnetic_field_from_CurrentPotentialField(
             field=self,
             coords=coords,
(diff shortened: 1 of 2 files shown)


ddudt (approved):

desc/magnetic_fields.py
```diff
@@ -2029,7 +2026,7 @@ def _compute_magnetic_field_from_CurrentPotentialField(
     coords = jnp.atleast_2d(coords)
     if basis == "rpz":
         coords = rpz2xyz(coords)
-    surface_grid = source_grid or LinearGrid(M=30, N=30, NFP=field.NFP)
+    surface_grid = source_grid or LinearGrid(M=60, N=60, NFP=field.NFP)
```
f0uriest, line 2029: would it be better to have this depend on M,N of the surface/potential in some way?

ddudt: we might as well replace instances of `surface_grid` with `source_grid` to be consistent, no need for the variable name change here

dpanici: @f0uriest The problem with doing Rory's suggestion is that a surface current whose potential is purely the secular terms can have a M,N of 0 or 1 or something, but to accurately get the biot savart law you still need a lot of points (at least in my experience), so I figured a large default should be used

f0uriest: what about something like `30 + 2*max(self.M, self.M_phi)`? that way you get a reasonable resolution plus it scales as you increase the complexity of the surface

f0uriest (approved):

PR #883, opened 2024-02-16
Bump dependency versions, add automatic test action

### LINKED ISSUE #557: Support new scipy versions
``scipy 1.11.0`` changed the way optimizer callbacks are handled which breaks some of our wrapper code, but would ultimately make it much simpler once we update our stuff to handle it. Might want to wait though so we don't need to support multiple versions.

### LINKED ISSUE #790: Update dependency versions
Jax,  numpy, scipy, matplotlib etc all have newer versions we should try to support, and remove support for older Jax versions

f0uriest:
Right now we have dependency versions pegged fairly conservatively to prevent possible issues with new versions getting released that we don't test.

This PR bumps most of the dependency versions to their next major release and adds a new github action that will run the unit tests with the latest version of all the dependencies every week so we will be notified if there are any breaking changes.

This should allow us to leave the dependency versions a lot looser and add specific fixes or guards if we run into issues.

Resolves #790 
Resolves #557


desc/geometry/curve.py
```diff
@@ -770,7 +770,7 @@ def knots(self):
     @knots.setter
     def knots(self, new):
         if len(new) == len(self.knots):
-            knots = jnp.atleast_1d(new)
+            knots = jnp.atleast_1d(jnp.asarray(new))
```
dpanici, line 773: jax requires the inputs to the atleast functions to be jnp arrays in newer versions?

dpanici (approved):

desc/geometry/curve.py
f0uriest: yeah they've been steadily enforcing jnp arrays to be used for all jnp functions for  a while

ddudt (approved):
Does this increase the jax version that will be used for the tests, and if so could we run the benchmarks to see if the newer versions changed the speed at all?

f0uriest:
> Does this increase the jax version that will be used for the tests, and if so could we run the benchmarks to see if the newer versions changed the speed at all?

Our existing benchmarks run different versions of our code against the same version of jax, we'd need to make a new one to run the same version of our code against different versions of jax, but anecdotally I don't see any major improvements.

PR #884, opened 2024-02-18
Add compute implementation for surface geometry terms

### LINKED ISSUE #795: Add compute implementations for `FourierRZToroidalSurface` to allow targeting by `AspectRatio` and `Elongation` Objectives
Related to #742 , need the following quantities implemented for `FourierRZToroidalSurface` for certain objectives to work:
 - `AspectRatio` requires `R0/a` to be implemented
 - `Elongation` requires `a_major/a_minor` to be implemented

These use surface/volume integrals or involve basis vectors which are not defined for `FourierRZToroidalSurface` but these should be easily computable for a `FourierRZToroidalSurface` object

f0uriest:
- Adds implementation for area, major/minor radii, elongation for `FourierRZToroidalSurface`
- Allows `Elongation` and `Aspect` ratio objectives to target surfaces as well as equilibria

Resolves #795


desc/compute/_geometry.py
```diff
@@ -355,29 +421,42 @@ def _a_major_over_a_minor(params, transforms, profiles, data, **kwargs):
...
+    description="Elongation at a toroidal cross-section",
+    dim=1,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="z",
+    data=["g_tt", "A"],
+    parameterization=["desc.geometry.surface.FourierRZToroidalSurface"],
+)
+def _a_major_over_a_minor_ToroidalSurface(params, transforms, profiles, data, **kwargs):
```
ddudt, line 451: It looks like we could easily combine this with the existing `_a_major_over_a_minor` function to reduce the duplicated code.

tests/test_compute_funs.py
```diff
@@ -1653,3 +1653,20 @@ def test_iota_components(HELIOTRON_vac):
...
+    data = ["A", "V", "a", "R0", "R0/a", "a_major/a_minor"]
+    for name in names:
+        eq = get(name)
+        for key in data:
+            x = eq.compute(key)[key].max()  # max needed for elongation broadcasting
+            y = eq.surface.compute(key)[key].max()
+            if key == "a_major/a_minor":
+                rtol, atol = 1e-2, 1e-4  # need looser tol here bc of different grids
+            else:
+                rtol, atol = 1e-8, 1e-8
```
ddudt, line 1671: Does `atol` need to be non-zero? As a best practice we should try to keep `atol=0` whenever possible, especially when comparing multiple quantities at different scales. But all of these geometric quantities are large enough that `atol` shouldn't make a difference.

ddudt (changes requested):

desc/compute/_geometry.py
f0uriest: I thought we couldn't bc the original implementation was only correct for quadrature grids?

dpanici (approved):

tests/test_compute_funs.py
```diff
@@ -1653,3 +1653,20 @@ def test_iota_components(HELIOTRON_vac):
...
+    """Test that computing stuff from surface gives same result as equilibrium."""
+    names = ["DSHAPE", "HELIOTRON", "NCSX"]
+    data = ["A", "V", "a", "R0", "R0/a", "a_major/a_minor"]
+    for name in names:
+        eq = get(name)
+        for key in data:
+            x = eq.compute(key)[key].max()  # max needed for elongation broadcasting
+            y = eq.surface.compute(key)[key].max()
+            if key == "a_major/a_minor":
+                rtol, atol = 1e-2, 0  # need looser tol here bc of different grids
```
rahulgaur104, line 1669: If I increase the resolution of both the grids, can I reduce rtol and still pass the test?

desc/compute/_geometry.py
```diff
@@ -324,44 +386,68 @@ def _a(params, transforms, profiles, data, **kwargs):
...
+    data=["A(z)", "perimeter(z)"],
+    parameterization=[
+        "desc.equilibrium.equilibrium.Equilibrium",
+        "desc.geometry.core.Surface",
+    ],
+)
+def _a_major_over_a_minor(params, transforms, profiles, data, **kwargs):
+    A = transforms["grid"].compress(data["A(z)"], surface_label="zeta")
+    P = transforms["grid"].compress(data["perimeter(z)"], surface_label="zeta")
     # derived from Ramanujan approximation for the perimeter of an ellipse
```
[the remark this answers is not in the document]
f0uriest: it's always an approximation for any non-elliptical cross section, since the definitions of major and minor axes and elongation aren't uniquely defined in those cases. In practice it still gives reasonable results in all the cases we've tried it. For a triangular cross section you'd basically be getting the elongation of an equivalent elipse which would likely be close to 1, which is what we'd expect.

tests/test_compute_funs.py
f0uriest: in theory yes. The main difference is that the eq.compute uses a quadrature grid to compute the perimeter, which doesn't include nodes at rho=1, so we extrapolate what the true perimeter is based on an approximation from rho~0.9x. As you increase the resolution, the outermost node in the quadrature grid will approach rho=1 but you'd need to go to really high resolution.

rahulgaur104 (approved):

ddudt (approved):

PR #885, opened 2024-02-18
Add radial weighting for near axis fit

### LINKED ISSUE #798: Add radial weighting when fitting to near axis
We trust the nae closest to the axis, so doing a weighted fit with weight like 1/rho or something would likely give better results than uniform weighting, especially at low aspect ratio

f0uriest:
By construction, the near axis expansion is most accurate near the axis, so this PR adds a radial weighting when we fit an equilibrium from a near axis solution. Default is `1/rho**2` which seems to work well on a few equilibria I tried.

Resolves #798

### DIFF
--- a/desc/equilibrium/equilibrium.py
+++ b/desc/equilibrium/equilibrium.py
@@ -1586,7 +1586,15 @@ class Equilibrium(IOAble, Optimizable):
 
     @classmethod
     def from_near_axis(
-        cls, na_eq, r=0.1, L=None, M=8, N=None, ntheta=None, spectral_indexing="ansi"
+        cls,
+        na_eq,
+        r=0.1,
+        L=None,
+        M=8,
+        N=None,
+        ntheta=None,
+        spectral_indexing="ansi",
+        w=2,
     ):
         """Initialize an Equilibrium from a near-axis solution.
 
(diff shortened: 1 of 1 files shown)


desc/equilibrium/equilibrium.py
```diff
@@ -1586,7 +1586,15 @@ def resolution_summary(self):
...
-        cls, na_eq, r=0.1, L=None, M=8, N=None, ntheta=None, spectral_indexing="ansi"
+        cls,
+        na_eq,
+        r=0.1,
+        L=None,
+        M=8,
+        N=None,
+        ntheta=None,
+        spectral_indexing="ansi",
+        w=2,
```
ddudt, line 1597: In theory, should this be related to the "order" of the NAE constraints we use? If so, I would recommend changing the variable name to `order` or whatever we use in those NAE constraint functions.

f0uriest: I don't think theres a direct relation. While the order of the expansion usually means the distance from the axis that we trust it, the accuracy also depends on the shape of the axis and r/r_crit etc. In all cases we know the axis is "most accurate" though the exact degree of falloff isn't clear

dpanici: I agree, the order of the NAE would not relate to the weighting near-axis that we are trying to do here.

dpanici (approved):
Not sure if I will notice much of a difference but is good to have. re-trying my paper results to see if anything changes, and would be very cool if this can let us, with agressive weights, fit NAE surfaces that are very crappy at rho=1 with more reasonable surfaces so our initial guess for NAE constraint isnt so bad

ddudt (approved):

PR #889, opened 2024-02-19
Simplify `factorize_linear_constraints`

### LINKED ISSUE #859: Replace cumbersome `A` matrix stacking logic for linear constraints with a single `ObjectiveFunction` for all linear objectives
Action items @ddudt 
-------------------------
- create `ObjectiveFunction` inside of `optimizer.optimize` that takes in the linear constriants
- call `combine_args` on it and objective before we do the linear constraint projection
- now `factorize_linear_constraints` argument `self._constraints` in (https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/optimize/_constraint_wrappers.py#L94) is now just an objective function, use its jac_scaled argument to get `A_full` in (https://github.com/PlasmaControl/DESC/blob/3707e08652021f631443aa290969f89ab11b243c/desc/objectives/utils.py#L102) and call obj.compute(zeros) to get `b_full`
- keep the indexing stuff after that line above, and everything should hopefully work fine for optimizable optimization (not yet for OptimizableCollection)

ddudt:
Resolves #859


ddudt:
override `LinearConstraintProjection.dim_x` when its objective is a `ProximalProjection`

desc/objectives/objective_funs.py
```diff
@@ -700,7 +701,7 @@ def things(self):
             RuntimeError,
             "ObjectiveFunction must be built with ObjectiveFunction.build() first",
         )
-        return self._flatten(self._all_things)
+        return self._flatten(self.all_things)
 
     @things.setter
     def things(self, new):
```
f0uriest, line 707: I think it might be worth it to just get rid of this setter method. Changing around what things are bound to an equilibrium is likely to cause lots of weird problems, better to just make a new objective.

desc/objectives/utils.py
```diff
@@ -32,112 +32,100 @@ def factorize_linear_constraints(constraints, objective):  # noqa: C901
...
     project, recover : function
-        functions to project full vector x into reduced vector y,
-        and recovering x from y.
+        Functions to project full vector x into reduced vector y,
+        and to recover x from y.
 
     """
-    for con in constraints:
+    for con in constraint.objectives:
+        if con.bounds is not None:
```
f0uriest, line 58: can use `errorif` here to improve coverage

f0uriest: and may also want to check that its linear?

f0uriest:
I think part of the issue is we need to keep track of two sets of things:
1. things that this objective function actually needs in order to evaluate the objective
2. other things that it technically doesn't need, but needs to accept and throw away those params so that different objectives take the same x

I propose we make `things` the former and `all_things` the latter. The `x` vector is based on `all_things`, but when we do `unpack_state` it should select only the things that are actually needed and discard the rest, like
```python
params = [par for par, thi in zip(params, all_things) if thi in things]
```
as long as `things` and `all_things` are sorted using the same key this should be fine, a simple one is
```python
stuff = sorted(stuff, key=lambda x: id(x))
```

We might need separate `flatten` and `unflatten` utils for both sets.

ddudt:
> I think part of the issue is we need to keep track of two sets of things:
> 
> 1. things that this objective function actually needs in order to evaluate the objective
> 2. other things that it technically doesn't need, but needs to accept and throw away those params so that different objectives take the same x
> 
> I propose we make `things` the former and `all_things` the latter. The `x` vector is based on `all_things`, but when we do `unpack_state` it should select only the things that are actually needed and discard the rest, like
> 
> ```python
> params = [par for par, thi in zip(params, all_things) if thi in things]
> ```
> 
> as long as `things` and `all_things` are sorted using the same key this should be fine, a simple one is
> 
> ```python
> stuff = sorted(stuff, key=lambda x: id(x))
> ```
> 
> We might need separate `flatten` and `unflatten` utils for both sets.

This is basically the solution I came up with, with the following differences: 

- I removed the variable `all_things`, and now `ObjectiveFunction.things` keeps track of "all things" from itself + any other objectives. We don't need a separate variable to keep track of the things it needs itself because that is simply `[obj.things for obj in self.objectives]`. 
- The order of the params in the state vector is determined by the `combine_args` util function. (The actual order doesn't matter, as long as it's the same for every objective).

desc/objectives/objective_funs.py
```diff
@@ -200,13 +193,21 @@ def _set_things(self, things=None):
         Sets ``self._flatten`` as a function to return unique flattened list of things
         and ``self._unflatten`` to recreate full nested list of things
         from unique flattened version.
+
         """
         from jax.tree_util import tree_flatten, tree_unflatten
```
f0uriest, line 198: we use these in enough places now that we can probably move it to backend?

desc/objectives/objective_funs.py
```diff
@@ -685,48 +692,10 @@ def weights(self):
...
 
-    @property
-    def _all_things(self):
-        """list: all things known to this objective, used and unused."""
-        if not hasattr(self, "_extra_things"):
-            self._extra_things = []
-        return [obj.things for obj in self.objectives] + self._extra_things
-
     @property
     def things(self):
```
f0uriest, line 696: maybe don't need a public attribute for this?

desc/objectives/objective_funs.py
```diff
@@ -383,7 +384,13 @@ def unpack_state(self, x, per_objective=True):
         xs = jnp.split(x, xs_splits)
         params = [t.unpack_params(xi) for t, xi in zip(self.things, xs)]
         if per_objective:
+            # params is a list of lists of dicts, for each thing and for each objective
             params = self._unflatten(params)
+            # this filters out the params of things that are unused by each objective
+            params = [
```
f0uriest, line 390: so after this is it still a list of lists? or just a list of dict?

desc/objectives/utils.py
```diff
@@ -150,22 +139,21 @@ def recover(x_reduced):
...
         y2 = con.target
         y1, y2 = np.broadcast_arrays(y1, y2)
         np.testing.assert_allclose(
             y1,
             y2,
-            atol=2e-14,
-            rtol=5e-14,
-            err_msg="Incompatible constraints detected, cannot satisfy "
-            + f"constraint {con}",
+            atol=3e-14,
```
f0uriest, line 151: I think we should keep an rtol here in case of units issues.

Also I wonder if this might be better as a warning than an error? I've run into cases where due to units or scale issues it fails with an error of 1e-12 or 1e-10 or something, not a showstopper but something the user should at least be aware of.

desc/optimize/_constraint_wrappers.py
```diff
@@ -35,25 +35,33 @@ class LinearConstraintProjection(ObjectiveFunction):
...
-                raise ValueError(
-                    "LinearConstraintProjection method "
-                    + "cannot handle nonlinear constraint {}.".format(con)
-                )
+        errorif(
+            not isinstance(constraint, ObjectiveFunction),
+            ValueError,
+            "constraint should be instance of ObjectiveFunction.",
+        )
+        for con in constraint.objectives:
```
f0uriest, line 55: could also add a check that the constraint isnt uses bounds

desc/optimize/_constraint_wrappers.py
```diff
@@ -535,12 +538,13 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
         for constraint in self._linear_constraints:
             constraint.build(use_jit=use_jit, verbose=verbose)
 
-        assert self._constraint.things == [
-            eq
-        ], "ProximalProjection can only handle constraints on the equilibrium."
+        errorif(
+            self._constraint.things != [eq],
+            "ProximalProjection can only handle constraints on the equilibrium.",
```
f0uriest, line 543: need to specify an error type here

desc/optimize/optimizer.py
```diff
@@ -148,7 +147,7 @@ def optimize(  # noqa: C901 - FIXME: simplify this
         """
         if not isinstance(constraints, (tuple, list)):
             constraints = (constraints,)
-        things = flatten_list(things, flatten_tuple=True)
+        things = unique_list(flatten_list(things, flatten_tuple=True))[0]
```
f0uriest, line 150: should maybe also check here that they didnt pass in duplicate things?

desc/optimize/optimizer.py
```diff
@@ -163,40 +162,46 @@ def optimize(  # noqa: C901 - FIXME: simplify this
...
+        if linear_constraint is not None and not linear_constraint.built:
+            linear_constraint.build(verbose=verbose)
         if nonlinear_constraint is not None and not nonlinear_constraint.built:
             nonlinear_constraint.build(verbose=verbose)
 
+        # combine arguments from all three objective functions
         if nonlinear_constraint is not None:
-            objective, nonlinear_constraint = combine_args(
-                objective, nonlinear_constraint
+            objective, linear_constraint, nonlinear_constraint = combine_args(
```
f0uriest, line 187: dont we still need to combine args for the objective and linear constraints, even if there aren't any nonlinear constraints?

desc/objectives/objective_funs.py
ddudt: We probably don't _need_ it, but we have it public for `Optimizable` and I don't think it's hurting anything.

desc/objectives/objective_funs.py
ddudt: It's still a list of lists. The outermost list is for each objective, and the inner list is for each thing. Then within that it's a dict of the params for that thing. Like:
`[[{thing1_params}, {thing2_params}], [{thing1_params}, {thing2_params}]]`

At first, `params` contains all of the things for every objective. This logic is removing the things that aren't needed by that objective. So if the first objective only needs thing1 and the second objective only needs thing2, it will become:
`[[{thing1_params}], [{thing2_params}]]`

desc/objectives/objective_funs.py
f0uriest: Not sure what you mean? `things` is only defined for objectives. My point was that it's not something we necessarily want the user messing with.

ddudt: Oh sorry, scratch what I said about `Optimizeable`. But this is the `ObjectiveFunction` class, and it is an important attribute. This is only a getter method -- I removed the setter method -- so the users shouldn't be able to mess with it.

desc/optimize/optimizer.py
ddudt: I don't think duplicate things will actually break anything, but I added this check as a warning.

desc/objectives/objective_funs.py
f0uriest: yeah, @kianorr also uses them for the coil objectives.

desc/objectives/utils.py
ddudt: What do you mean by unit issues? The only reason this won't be exactly satisfied is from numerical issues with inverting the `A` matrix, right? So I guess if different constraints have different orders of magnitude it could get messy.

desc/optimize/optimizer.py
f0uriest: Yeah I guess none with more than 1 thing and no nonlinear constraints?

desc/objectives/utils.py
ddudt: I can add the rtol back in. I'm on the fence between error vs warning

f0uriest: yeah "unit issues" being constraints with different units that are different orders of magnitude, like pressure/current/temperature etc. I've also run into the same issue if I just re-scale an equilibrium to something really big

f0uriest (dismissed):

f0uriest:
We may want to rerun the benchmarks to make sure the slowdown in the Jacobian test is just a fluke

desc/objectives/utils.py
```diff
@@ -32,112 +30,103 @@ def factorize_linear_constraints(constraints, objective):  # noqa: C901
...
     project, recover : function
-        functions to project full vector x into reduced vector y,
-        and recovering x from y.
+        Functions to project full vector x into reduced vector y,
+        and to recover x from y.
 
     """
-    for con in constraints:
+    for con in constraint.objectives:
+        errorif(not con.linear, ValueError, f"Cannot handle nonlinear onstraint {con}.")
```
dpanici, line 56: spelling error (onstraint instead of constraint)

desc/objectives/utils.py
dpanici: I think I am in favor of moving to a warning but let's do that in another PR

desc/optimize/optimizer.py
```diff
@@ -148,8 +147,18 @@ def optimize(  # noqa: C901 - FIXME: simplify this
...
-        things = flatten_list(things, flatten_tuple=True)
+
+        # get unique things
+        things, indices = unique_list(flatten_list(things, flatten_tuple=True))
+        counts = np.unique(indices, return_counts=True)[1]
+        duplicate_idx = np.where(counts > 1)[0]
+        warnif(
+            len(duplicate_idx),
+            UserWarning,
+            f"{[things[idx] for idx in duplicate_idx]} is duplicated in things.",
```
dpanici, line 158: this is sort of esoteric, and do we want to throw an error if this is the case or does the rest of the optimizer work fine still if an object is duplicated in the passed-in things? We only use the unique things anyways?

dpanici (commented):
will be approved pending the benchmarks results

desc/optimize/optimizer.py
f0uriest: Everything should still work but it might lead to some weird behavior if things get updated twice etc

f0uriest:
I think the benchmarks still might fail bc the API on master is different, but as long as they work for this branch that's fine

dpanici:
> I think the benchmarks still might fail bc the API on master is different, but as long as they work for this branch that's fine

Yea, they were failing on this branch too though so this will at least get it to work for this branch

f0uriest (dismissed):

dpanici (changes requested):
I am going to be annoying but I'd like to know the speed comparison for the benchmark cases between here and master, it wont run automatically bc the expected arguments changed in the benchmarks functions in this PR. I can do this tomorrow

dpanici:
```diff
|             benchmark_name             |         dt(%)          |         dt(s)          |        t_new(s)        |        t_old(s)        | 
| -------------------------------------- | ---------------------- | ---------------------- | ---------------------- | ---------------------- |
 test_build_transform_fft_lowres         |     -2.14 +/- 3.66     | -3.41e-04 +/- 5.83e-04 |  1.56e-02 +/- 4.8e-04  |  1.59e-02 +/- 3.4e-04  |
 test_build_transform_fft_midres         |     +0.05 +/- 1.08     | +5.26e-05 +/- 1.11e-03 |  1.03e-01 +/- 6.9e-04  |  1.03e-01 +/- 8.8e-04  |
 test_build_transform_fft_highres        |     -0.51 +/- 0.89     | -2.53e-03 +/- 4.40e-03 |  4.90e-01 +/- 2.5e-03  |  4.93e-01 +/- 3.6e-03  |
 test_equilibrium_init_lowres            |     -5.27 +/- 5.76     | -2.68e-02 +/- 2.93e-02 |  4.81e-01 +/- 2.1e-02  |  5.08e-01 +/- 2.0e-02  |
 test_equilibrium_init_medres            |     +0.23 +/- 6.78     | +2.18e-03 +/- 6.45e-02 |  9.54e-01 +/- 2.8e-02  |  9.52e-01 +/- 5.8e-02  |
 test_equilibrium_init_highres           |     +1.04 +/- 2.56     | +3.06e-02 +/- 7.54e-02 |  2.97e+00 +/- 6.9e-02  |  2.94e+00 +/- 3.1e-02  |
 test_objective_compile_dshape_current   |     +1.85 +/- 3.73     | +1.21e-01 +/- 2.45e-01 |  6.68e+00 +/- 2.0e-01  |  6.56e+00 +/- 1.4e-01  |
 test_objective_compile_atf              |     -0.39 +/- 4.24     | -7.83e-02 +/- 8.48e-01 |  1.99e+01 +/- 5.2e-01  |  2.00e+01 +/- 6.7e-01  |
 test_objective_compute_dshape_current   |     +3.67 +/- 2.83     | +3.14e-04 +/- 2.42e-04 |  8.88e-03 +/- 2.0e-04  |  8.56e-03 +/- 1.3e-04  |
 test_objective_compute_atf              |     -0.38 +/- 2.08     | -7.00e-05 +/- 3.85e-04 |  1.85e-02 +/- 3.1e-04  |  1.85e-02 +/- 2.3e-04  |
 test_objective_jac_dshape_current       |     +0.12 +/- 1.70     | +1.77e-04 +/- 2.56e-03 |  1.51e-01 +/- 2.4e-03  |  1.51e-01 +/- 1.0e-03  |
 test_objective_jac_atf                  |     +4.47 +/- 1.82     | +4.04e-01 +/- 1.64e-01 |  9.44e+00 +/- 8.2e-02  |  9.04e+00 +/- 1.4e-01  |
 test_perturb_1                          |     +5.75 +/- 3.20     | +1.08e+00 +/- 6.00e-01 |  1.99e+01 +/- 2.3e-01  |  1.88e+01 +/- 5.5e-01  |
 test_perturb_2                          |     +7.67 +/- 2.73     | +2.07e+00 +/- 7.35e-01 |  2.90e+01 +/- 3.7e-01  |  2.70e+01 +/- 6.4e-01  |
 test_proximal_jac_atf                   |     +1.33 +/- 2.37     | +3.58e-01 +/- 6.39e-01 |  2.73e+01 +/- 2.2e-01  |  2.69e+01 +/- 6.0e-01  |
-test_proximal_freeb_compute             |     +1.34 +/- 0.41     | +2.76e-03 +/- 8.47e-04 |  2.09e-01 +/- 5.3e-04  |  2.07e-01 +/- 6.6e-04  |
 test_proximal_freeb_jac                 |     -5.14 +/- 3.25     | -2.06e+00 +/- 1.30e+00 |  3.80e+01 +/- 7.9e-02  |  4.00e+01 +/- 1.3e+00  |
```

dpanici (approved):
very minor difference in speed with things being a couple percent slower with this PR, that is fine though.

f0uriest (approved):

PR #894, opened 2024-02-24
Clean up API for free boundary with surface current

### LINKED ISSUE #869: Use `FourierCurrentPotentialField` object as the `eq.surface` attribute when doing free boundary calculations with sheet current
A cleaner way @f0uriest had thought of for free boundary to simplify the logic is to allow the surface of the equilibrium to be a surface current object, that way the params dont need to be manually copied over and it can be directly changed by the optimizer.

f0uriest:
Resolves #869


dpanici:
I think that adding a `M_Phi` and `N_Phi` to the current potenital from_surface method makes sense, kind of an oversight on my part to not have that as an option. If you want to add it to this PR I think it would fit or I can commit it today to this as well

desc/equilibrium/equilibrium.py
```diff
@@ -1297,6 +1297,51 @@ def Zb_lmn(self):
     def Zb_lmn(self, Zb_lmn):
         self.surface.Z_lmn = Zb_lmn
 
+    @optimizable_parameter
+    @property
+    def I(self):  # noqa: E743
+        """float: Net current linking the plasma surface toroidally."""
```
dpanici, line 1303: Maybe specify that it is a surface current? as `I` itself is something that can be calculated for interior surfaces too

desc/equilibrium/equilibrium.py
```diff
@@ -1297,6 +1297,51 @@ def Zb_lmn(self):
...
+            not hasattr(self.surface, "I"),
+            ValueError,
+            "Attempt to set I on an equilibrium without sheet current",
+        )
+        self.surface.I = new
+
+    @optimizable_parameter
+    @property
+    def G(self):
+        """float: Net current linking the plasma surface poloidally."""
```
dpanici, line 1318: same as above, specify that it is the surface current specifically, not the `G` from `eq.compute`

dpanici (changes requested):
- small comment on docstring of the new eq properties accessing the surface current field

dpanici (dismissed):

dpanici (dismissed):

dpanici (approved):

PR #903, opened 2024-02-24
Update zernike_radial to work better on GPU

YigitElma:
I have been benchmarking different functions on my other repository, and I realized that zernike_radial function is not suitable for GPU at all. It is very slow compared to running it on CPU.  For comparison, here are the CPU bechmarks,
image

And here are the GPU benchmarks,
image

Here, The first 3 results are from new functions, the last PR changed it to use 3rd one. The last result is previous version. My guess is previous function was using jnp.vectorize for all modes, so using GPU was making performance improvement. But the new version only vectorizes over r and probably transferring the data between different GPU cores causes the extreme slowing down. The pure CPU speeds are still good, so I changed the decorator to jit compile the function on CPU. 

Sorry for the mistake, I wouldn't think the GPU speed would be soo bad.


f0uriest:
In the times you posted which are the one as currently implemented?

f0uriest (approved):

Issue #904, opened 2024-02-25
eq.compute doesn't correctly override grids to compute quantities that require evaluating functions along field lines


rahulgaur104:
In the compute function in equilibrium.py if the user provides a grid, the function should set override_grid = False.

https://github.com/PlasmaControl/DESC/blob/f1689c85c022d5e8073325b7c057f6ea253923b4/desc/equilibrium/equilibrium.py#L794C1-L800C14

However, override_grid still stays True. This is a simple fix.

unalmis:
I think this logic is intentional, see #440

f0uriest:
See also #688 and related PR

rahulgaur104:
Well, it definitely throws an error with the ballooning branch when I create a separate grid before calculating the growth rate using eq.compute.

I get assertion error and see this code

```
            data1dr = compute_fun(
                self,
                dep1dr,
                params=params,
                transforms=get_transforms(dep1dr, obj=self, grid=grid1dr, **kwargs),
                profiles=get_profiles(dep1dr, obj=self, grid=grid1dr),
                data=None,
                **kwargs,
            )
            # need to make this data broadcast with the data on the original grid
            data1dr = {
                key: grid.expand(
                    grid1dr.compress(val, surface_label="rho"), surface_label="rho"
                )
                for key, val in data1dr.items()
                if key in dep1dr
            }
```

f0uriest:
It sounds like you might just need to pass `override_grid=False` when you call `eq.compute` if you have some custom grid you want to use and don't care about averages etc (because you're precomputing those on a different grid). That's the purpose of the `override_grid` option.

If you can post some code that reproduces the issue I can look into it more.

rahulgaur104:
Yeah, with override_grid=False, it worked fine. I was just thinking that it could be a potential issue for people doing calculations on a custom grid. I also only becomes an issue when I do this calculation in a jupyter notebook. 

Ok, I guess we can close this for now.

unalmis:
At some point, can you show where else you have needed to use `override_grid=False` so that I can check your code? I believe it indicates that your code has a bug.

For example, I recently fixed a bug in `bounce` branch where this was an issue with an inconsistent use of normalization between the toroidal flux label and desc's radial label. (This can lead to computations being done across different flux surfaces than expected). In this case, setting this to `override_grid=False` will temporarily supress the error, but it causes bugs that will pop up elsewhere.

PR #905, opened 2024-02-25
Fixes override_grid logic in the compute function in equilibrium.py

rahulgaur104:
See issue #904


rahulgaur104:
Bizzare that test_plotting fails here but works on my laptop. Have you guys seen this before?

f0uriest:
> Bizzare that test_plotting fails here but works on my laptop. Have you guys seen this before?

did you run pytest with ``--mpl``? If not it just checks that the plotting tests run without errors, you need the ``mpl`` flag to have it actually compare the images to the baseline.

Also see Kaya's comment on the related issue, this is intended behavior. Can you give an example of what issue you were seeing?

rahulgaur104:
override_grid = True is the correct default behavior.

PR #908, opened 2024-02-28
Add Bounding Box to Field Line Integration

dpanici:
- Adds a bounding box to the `field_line_integrate` defined by `bounds_R` and `bounds_Z` keyword arguments, which form a hollow cylindrical bounding box. If the field line trajectory exits these bounds, the RHS will be multiplied by `exp(-(R**2+Z**2))` to stop the trajectory and prevent tracing the field line out to infinity, which is both costly and unnecessary when making a Poincare plot, which is the principle purpose of the function.

In one example I had when tracing the precise_QA with a coilset that had some diverging field lines, it took 1/3 of the time compared to without using bounds, so seems to work fine. Defaults to no bounds.

@ddudt this PR is not addressing anything that happens when a SplineMagneticField is used and the field lines exit its domain of validity when `extrap=False` unless you specify the bounding box to lie within its domain of validity, in which case it should stop the integration soon after it exits the bounding box and possibly before it exits the field's interpolation domain. If there are ideas of how to treat that then we can make an issue and a separate PR

### DIFF
--- a/desc/magnetic_fields.py
+++ b/desc/magnetic_fields.py
@@ -1442,19 +1442,30 @@ def field_line_integrate(
         # integrating the field line to infinity, which is costly
         # and not useful in most cases
         decay_factor = jnp.where(
-            jnp.any(
-                jnp.array(
-                    [
-                        jnp.less(r, bounds_R[0]),
-                        jnp.greater(r, bounds_R[1]),
-                        jnp.less(z, bounds_Z[0]),
-                        jnp.greater(z, bounds_Z[1]),
-                    ]
-                )
+            jnp.array(
+                [
+                    jnp.less(r, bounds_R[0]),
+                    jnp.greater(r, bounds_R[1]),
+                    jnp.less(z, bounds_Z[0]),
+                    jnp.greater(z, bounds_Z[1]),
+                ]
             ),
-            jnp.exp(-(r**2 + z**2)),
+            1e-2
+            * jnp.array(
+                [
+                    # we mult by 1e-2 to accelerate the decay so that the integration
+                    # is stopped soon after the bounds are exited.
+                    jnp.exp(-((r - bounds_R[0]) ** 2)),
+                    jnp.exp(-((r - bounds_R[1]) ** 2)),
+                    jnp.exp(-((z - bounds_Z[0]) ** 2)),
+                    jnp.exp(-((z - bounds_Z[1]) ** 2)),
+                ]
+            ),
             1.0,
         )
+        # multiply all together, the conditions that are not violated
+        # are just one while the violated ones are continous decaying exponentials
+        decay_factor = jnp.prod(decay_factor)
 
         br, bp, bz = field.compute_magnetic_field(
             rpz, params, basis="rpz", source_grid=source_grid
--- a/desc/magnetic_fields.py
+++ b/desc/magnetic_fields.py
@@ -1433,6 +1433,7 @@ def field_line_integrate(
     """
     r0, z0, phis = map(jnp.asarray, (r0, z0, phis))
     assert r0.shape == z0.shape, "r0 and z0 must have the same shape"
+    assert decay_accel > 0, "decay_accel must be positive"
     rshape = r0.shape
     r0 = r0.flatten()
     z0 = z0.flatten()
@@ -1458,8 +1459,8 @@ def field_line_integrate(
             ),
             jnp.array(
                 [
-                    # we mult by 1e6 to accelerate the decay so that the integration
-                    # is stopped soon after the bounds are exited.
+                    # we mult by decay_accel to accelerate the decay so that the
+                    # integration is stopped soon after the bounds are exited.
                     jnp.exp(-(decay_accel * (r - bounds_R[0]) ** 2)),
                     jnp.exp(-(decay_accel * (r - bounds_R[1]) ** 2)),
                     jnp.exp(-(decay_accel * (z - bounds_Z[0]) ** 2)),
--- a/desc/magnetic_fields.py
+++ b/desc/magnetic_fields.py
@@ -1380,6 +1380,8 @@ def field_line_integrate(
     rtol=1e-8,
     atol=1e-8,
     maxstep=1000,
+    bounds_R=(0, np.inf),
+    bounds_Z=(-np.inf, np.inf),
 ):
     """Trace field lines by integration.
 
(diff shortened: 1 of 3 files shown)


desc/magnetic_fields.py
```diff
@@ -1419,12 +1436,35 @@ def field_line_integrate(
...
+                jnp.array(
+                    [
+                        jnp.less(r, bounds_R[0]),
+                        jnp.greater(r, bounds_R[1]),
+                        jnp.less(z, bounds_Z[0]),
+                        jnp.greater(z, bounds_Z[1]),
+                    ]
+                )
+            ),
+            jnp.exp(-(r**2 + z**2)),
```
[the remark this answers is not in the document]
dpanici: I think what I have should be continuous now, the decay factor is now `exp((r-R0)**2) * exp((r-R1)**2) * exp((z-Z0)**2) * exp((z-Z1)**2) `

where each element of this product is only added if that constraint is violated, and the bounding box is $R \in [R0,R1], Z \in [Z0,Z1]$

I also added a `1e-2` factor outside of the exponential to accelerate the decay a bit, as without it the integration still goes on for a bit longer (made it hard to test that this was working)

desc/magnetic_fields.py
```diff
@@ -1419,12 +1436,46 @@ def field_line_integrate(
...
+                    jnp.exp(-((r - bounds_R[1]) ** 2)),
+                    jnp.exp(-((z - bounds_Z[0]) ** 2)),
+                    jnp.exp(-((z - bounds_Z[1]) ** 2)),
+                ]
+            ),
+            1.0,
+        )
+        # multiply all together, the conditions that are not violated
+        # are just one while the violated ones are continous decaying exponentials
+        decay_factor = jnp.prod(decay_factor)
```
f0uriest, line 1468: does this need to be along an axis? Will this affect all field lines or just the one that leaves?

desc/magnetic_fields.py
```diff
@@ -1419,12 +1436,46 @@ def field_line_integrate(
...
+        decay_factor = jnp.where(
+            jnp.array(
+                [
+                    jnp.less(r, bounds_R[0]),
+                    jnp.greater(r, bounds_R[1]),
+                    jnp.less(z, bounds_Z[0]),
+                    jnp.greater(z, bounds_Z[1]),
+                ]
+            ),
+            1e-2
```
f0uriest, line 1453: I think this makes it discontinuous as well. The decay factor should be 1 for points on the boundary. If we want it to decay quicker probably better to just weight the exponential

dpanici: dang you are right, thanks

tests/test_magnetic_fields.py
```diff
@@ -678,6 +678,18 @@ def test_field_line_integrate(self):
...
         np.testing.assert_allclose(r[-1], 10, rtol=1e-6, atol=1e-6)
         np.testing.assert_allclose(z[-1], 0.001, rtol=1e-6, atol=1e-6)
+        # test that bounds work correctly, and stop integration when trajectory
+        # hits the bounds
+        r0 = [10.1]
+        z0 = [0.0]
+        # this will hit the R bound
+        # (there is no Z bound, and R would go to 10.0 if not bounded)
+        r, z = field_line_integrate(r0, z0, phis, field, bounds_R=(10.05, np.inf))
+        np.testing.assert_allclose(r[-1], 10.05, rtol=3e-4)
```
dpanici, line 688: we don't expect the integration to stop immediately after the bound is exited, but it should stop soon after depending on how fast the RHS decays. I made the decay very fast so we can test it, but if for some reason we should lessen the decay rates, then we will need to make these test tols here less tight

f0uriest (dismissed):

desc/magnetic_fields.py
```diff
@@ -1419,12 +1436,45 @@ def field_line_integrate(
...
+                ]
+            ),
+            jnp.array(
+                [
+                    # we mult by 1e6 to accelerate the decay so that the integration
+                    # is stopped soon after the bounds are exited.
+                    jnp.exp(-(1e6 * (r - bounds_R[0]) ** 2)),
+                    jnp.exp(-(1e6 * (r - bounds_R[1]) ** 2)),
+                    jnp.exp(-(1e6 * (z - bounds_Z[0]) ** 2)),
+                    jnp.exp(-(1e6 * (z - bounds_Z[1]) ** 2)),
```
ddudt, line 1460: Can we make this decay rate an input argument? A default value of 1e6 is fine, but that way it isn't a hard-coded magic number

desc/magnetic_fields.py
```diff
@@ -1401,6 +1403,21 @@ def field_line_integrate(
         relative and absolute tolerances for ode integration
     maxstep : int
         maximum number of steps between different phis
+    bounds_R : tuple of (float,float), optional
+        R bounds for field line integration bounding box.
+        If supplied, the RHS of the field line equations will be
+        multipled by exp(-r) where r is the distance from the origin,
```
ddudt, line 1409: where r is the distance from the bounding box

desc/magnetic_fields.py
```diff
@@ -1401,6 +1403,21 @@ def field_line_integrate(
...
+        this is meant to prevent the field lines which escape to infinity from
+        slowing the integration down by being traced to infinity.
+        defaults to (0,np.inf)
+    bounds_Z : tuple of (float,float), optional
+        Z bounds for field line integration bounding box.
+        If supplied, the RHS of the field line equations will be
+        multipled by exp(-r) where r is the distance from the origin,
```
ddudt, line 1416: where r is the distance from the bounding box

tests/test_magnetic_fields.py
```diff
@@ -678,6 +678,18 @@ def test_field_line_integrate(self):
         r, z = field_line_integrate(r0, z0, phis, field)
         np.testing.assert_allclose(r[-1], 10, rtol=1e-6, atol=1e-6)
         np.testing.assert_allclose(z[-1], 0.001, rtol=1e-6, atol=1e-6)
+        # test that bounds work correctly, and stop integration when trajectory
+        # hits the bounds
```
ddudt, line 682: Maybe make this a new separate unit test instead of adding it onto this existing test?

ddudt (changes requested):

desc/magnetic_fields.py
dpanici: good point, I'll add

f0uriest (approved):

tests/test_magnetic_fields.py
ddudt: This is duplicated now that you copied it to a new unit test, so you can delete it.

ddudt (changes requested):

ddudt (approved):

PR #909, opened 2024-02-28
Fix jvp signature for Objective

f0uriest:


rahulgaur104 (approved):
Works on dd/omnigenity now. Merge if all tests have passed!

PR #912, opened 2024-02-29
Fix jitable grid indexing and weights

### LINKED ISSUE #911: Fix weights and indexing for `Grid` when `jitable=True`
Like the cause of https://github.com/PlasmaControl/DESC/pull/854#discussion_r1504948317

Basically, when we set `jitable=True` the resulting weights aren't scaled correctly which can throw off any averaging done with that grid. This is usually fine because any averaging is done on a structured grid, but we should still fix it (and using a structured grid doesn't always fix it, see below).

The other problem is that `Grid.unique_rho_idx` etc are set as `np.arange(num_nodes)` when `jitable=True`, but it uses `np.unique` when `jitable=False`. `np.unique` returns the _sorted_ unique values, which means that when we do `grid1.expand(grid2.compress(x))` things get all jumbled up because `grid1` has sorted indices and `grid2` doesn't.

f0uriest:
Resolves #911


unalmis (approved):

unalmis:
Thanks, but it looks like this doesn't fix my issue. Also https://github.com/PlasmaControl/DESC/blob/569d35ebd6e9d9f732af14ea478f8b3a0d54a34a/desc/equilibrium/coords.py#L257 can squeeze 1d arrays into 0d which raises an error. It should have jnp.atleast_1d().

unalmis (changes requested):
The `jnp.argsort`  method will return an array of size equal to the input, rather than an array of size equal to the number of unique nodes, and this will cause errors with the compress/expand methods.

The errors are currently hidden by the choice of grids used for the tests. We should use tests where there are multiple nodes on a surface. On this grid, the number of nodes does not equal the number of surfaces (unique nodes).
```python
rho = np.repeat(np.arange(10), 10)
theta = np.repeat(np.arange(10), 10) * 2 * np.pi
zeta = np.repeat(np.arange(10), 10) * 2 * np.pi / eq.NFP
```

f0uriest:
@unalmis I'm able to reproduce the issue using 
```python
rho = np.linspace(0.1, 0.2, 2)
alpha = np.linspace(0, 2 * np.pi, 2)
zeta = np.linspace(0, 4 * np.pi, 10)
grid, data = desc.compute.utils.field_line_to_desc_coords(eq, rho, alpha, zeta)
print(sorted(data.keys()))
data = eq.compute(["|B|", "|B|_z constant rho alpha"], grid=grid, data=data, override_grid=False)
print(sorted(data.keys()))
print(data["|B|"].max())
print(data["|B|"].min())
print(data["|B|"].mean())
```
but curiously it goes away if I only use 1 rho value instead of 2 or more.

unalmis:
> @unalmis I'm able to reproduce the issue using
> 
> ```python
> eq = desc.examples.get("W7-X")
> rho = np.linspace(0.1, 0.2, 2)
> alpha = np.linspace(0, 2 * np.pi, 2)
> zeta = np.linspace(0, 4 * np.pi, 10)
> grid, data = desc.compute.utils.field_line_to_desc_coords(eq, rho, alpha, zeta)
> print(sorted(data.keys()))
> data = eq.compute(["|B|", "|B|_z constant rho alpha"], grid=grid, data=data, override_grid=False)
> print(sorted(data.keys()))
> print(data["|B|"].max())
> print(data["|B|"].min())
> print(data["|B|"].mean())
> ```
> 
> but curiously it goes away if I only use 1 rho value instead of 2 or more.

My findings were the same

f0uriest:
I don't think there's an easy way around that due to the limitations of JAX array sizes. One possibility would be to allow you to pass in values for `unique_idx` and `inverse_idx` to the custom `Grid` class? Since in most cases we know that the rho values will be unique?

f0uriest:
Or maybe the better fix is to just not use compress/expand to transfer data between grids, since it's kind of funky for unstructured grids. We could make a custom utility that copies data based on matching node values?

f0uriest:
We don't necessarily need to know the structure of the grid, I think something like this would work: 
```python
def copy_data_grids(x, grid1, grid2, match="rpz", tol=1e-14):
    """Copy data from one grid to another at matching nodes.
    
    Given data x corresponding to nodes of grid1, copy data
    to a new array that corresponds to grid2.
    
    Parameters
    ----------
    x : ndarray, shape(grid1.num_nodes,...)
        Data to copy
    grid1, grid2: Grid
        Grid to copy from and copy to
    match {"r", "p", "z"}
        What coordinates to match. For local quantities,
        all three coordinates must match. For flux surface 
        qtys, only rho must match, etc.
    tol : float
        tolerance for considering nodes the same.
        
    Returns
    -------
    y : ndarray, shape(grid2.num_nodes, ...)
        Data copied to grid2
    """
    axes = ()
    if "r" in match:
        axes += (0,)
    if "t" in match:
        axes += (1,)
    if "z" in match:
        axes += (2,)
    y = jnp.zeros((grid2.num_nodes, *x.shape[1:]))
    
    def body(i, y):
        y = jnp.where(jnp.max(jnp.abs((grid2.nodes[:,axes] - grid1.nodes[i,axes])), axis=-1) < tol, x[i], y)
        return y
    
    y = fori_loop(0, grid1.num_nodes, body, y)
    return y
```

The loop might not be the most efficient option, it can probably be vmapped instead, but I think the basic idea is sound.

All that said, I think that can wait for another PR. This one addresses the immediate issues of things not being set correctly when `jitable=True`, I think copying data between grids should be a separate issue

unalmis (commented):
My preference is to override `Grid.compress` and `Grid.expand` like
```python
def expand():
    if not jitable: # if not (self.num_rho + self.num_theta + self.num_zeta)
        raise error
    super.expand()
```
rather than the argsort which only works for the grids used in the new tests.

desc/grid.py
```diff
@@ -422,12 +422,12 @@ def __init__(self, nodes, sort=False, jitable=False):
             r = jnp.where(r == 0, 1e-12, r)
             self._nodes = jnp.array([r, t, z]).T
             self._axis = np.array([], dtype=int)
-            self._unique_rho_idx = np.arange(self._nodes.shape[0])
-            self._unique_theta_idx = np.arange(self._nodes.shape[0])
-            self._unique_zeta_idx = np.arange(self._nodes.shape[0])
-            self._inverse_rho_idx = np.arange(self._nodes.shape[0])
-            self._inverse_theta_idx = np.arange(self._nodes.shape[0])
-            self._inverse_zeta_idx = np.arange(self._nodes.shape[0])
+            self._unique_rho_idx = jnp.argsort(self._nodes[:, 0])
```
dpanici, line 425: something I did not realize, when jitable is passed the unique rho indices are not actually computed?

dpanici (approved):

desc/grid.py
f0uriest: no, since the size of the array wouldn't be a compile time constant. This works assuming all the rho/theta/zeta values are unique but not if there are any on the same surface.

PR #916, opened 2024-03-01
Allow shared transforms between R, Z, lambda

### LINKED ISSUE #812: Re-use transforms for Z, lambda
When `eq.sym=True`, the Z and lambda bases are the same (up to a minor difference in required derivatives), and the grids are the same as well, so the Transform for these two could point to the same object instead of 2 different objects (we would just need to set the `derivs` to the max of the derivatives needed btwn Z and lambda).

This could bring a speedup (maybe minor) to the code, and maybe even a decrease in memory usage at higher resolutions? 

And in the case `sym=False` this could be extended so that we only have one transform object used for R, Z, and lambda

f0uriest:
Resolves #812

### DIFF
--- a/desc/compute/utils.py
+++ b/desc/compute/utils.py
@@ -336,7 +336,7 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
             basis = getattr(obj, c + "_basis")
             # first check if we already have a transform with a compatible basis
             for transform in transforms.values():
-                if basis.eq(getattr(transform, "basis", None)):
+                if basis.equiv(getattr(transform, "basis", None)):
                     ders = np.unique(
                         np.vstack([derivs[c], transform.derivatives]), axis=0
                     ).astype(int)
--- a/desc/compute/utils.py
+++ b/desc/compute/utils.py
@@ -332,14 +332,14 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
     derivs = get_derivs(keys, obj, has_axis=grid.axis.size)
     transforms = {"grid": grid}
     for c in derivs.keys():
-        if hasattr(obj, c + "_basis"):
+        if hasattr(obj, c + "_basis"):  # regular stuff like R, Z, lambda etc.
             basis = getattr(obj, c + "_basis")
             # first check if we already have a transform with a compatible basis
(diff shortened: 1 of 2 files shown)


desc/compute/utils.py
```diff
@@ -333,13 +333,26 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
...
+            else:  # if we didn't exit the loop early
+                c_transform = Transform(
+                    grid,
+                    basis,
+                    derivs=derivs[c],
+                    build=False,
+                    method=method,
+                )
+            transforms[c] = c_transform
         elif c == "B":
```
rahulgaur104, line 356: c is a key in the dictionary derivs which sounds like a dictionary of derivatives. c corresponds to the magnetic field "B". Why is the magnetic field in the dictionary containing derivatives?

Confusing naming convention. Add a few comments in the code to clarify.

rahulgaur104 (changes requested):

desc/compute/utils.py
dpanici: c is meant to be just a quantity for which a transform exists, like R or Z or L. B is a transform that may exist as well (used for the boozer transform stuff). The `derivs` is a dictionary whose keys are things for which trasnforms exist, and that we may want derivs of, so I think it is natural for `"B"` to be here in its keys. `derivs` dict is explained by the docstring of `get_derivs` (https://github.com/PlasmaControl/DESC/blob/70001bcc9bc38c4e23d7d457ce5f9bb4898741ed/desc/compute/utils.py#L184)

dpanici (dismissed):

desc/compute/utils.py
f0uriest: "B" here refers to the transform thats needed to fit the boozer harmonics

rahulgaur104: I thought you meant the derivatives needed for the magnetic field. I'd still stay it's confusing to people who have used something like eq.compute("B"). Maybe add a comment in front of the elif in 356 saying "B means Boozer"?

rahulgaur104 (dismissed):
Merge if all tests have passed.

dpanici (approved):

PR #922, opened 2024-03-03
add to CONTRIBUTING.rst how to update master compute data

### LINKED ISSUE #888: Add how to update `master_compute_data.pkl` to `CONTRIBUTING.rst`
When the outputs of the compute quantities tested by the[ `test_compute_everything` test](https://github.com/PlasmaControl/DESC/blob/master/tests/test_compute_funs.py) are changed in a PR, that test will fail. The three(? maybe more) reasons this could occur are:
- the PR was not intended to change how things are computed, but messed up something unexpected and now the compute quantities are incorrect, if you did not expect things to change in the PR then look into why these differences are happening and fix the PR
- The PR updated the way one of the existing compute index quantities are computed (either by a redefinition or perhaps fixing an error present in `master`)
- The PR added a new class parametrization (such as a new subclass of `Curve` like `LinearCurve` etc)

If the 2nd case is the reason, then you must update the `master_compute_data.pkl` file with the correct quantities being computed by your PR: 
- First , run the test with `pytest tests -k test_compute_everything` and inspect the compute quantities whose values are in error, to ensure that only the quantities you expect to be different are shown (and that the new values are indeed the correct ones, you should have a test elsewhere for that though).
- If the values are as expected and only the expected compute quantities are different, then change the block
```python
except AssertionError as e:
    error = True
    print(e)
```
with
```python
except AssertionError as e:
    error = False
    update_master_data = True
    print(e)
```
- rerun the test  `pytest tests -k test_compute_everything` , now any compute quantity that is different between the PR and master will be updated with the PR value
- `git restore tests/test_compute_funs.py` to remove the change you made to the test
- `git add tests/inputs/master_compute_data.pkl` and commit to commit the new data file

If the 3rd case is the reason, then you must simply add the new parametrization to the [ `test_compute_everything` test](https://github.com/PlasmaControl/DESC/blob/master/tests/test_compute_funs.py) 
- `things` dictionary with a sensible example instance of the class to use for the test, and
- to the `grid` dictionary with a sensible default grid to use when computing the compute quantities for the new class
- Then, rerunning the test  `pytest tests -k test_compute_everything` will add the compute quantities for the new class and save them to the `.pkl` file
- `git add tests/inputs/master_compute_data.pkl` and commit to commit the new data file

dpanici:
resolves #888


f0uriest (dismissed):

f0uriest (dismissed):

CONTRIBUTING.rst
```diff
@@ -156,6 +156,46 @@ Opening a PR will trigger a suite of tests and style/formatting checks that must
...
 or edits to your PR.
 
+What if the ``test_compute_everything`` test fails, or there is a conflict in ``master_compute_data.pkl``?
+----------------------------------------------------------------------------------------------------------
+When the outputs of the compute quantities tested by the`test_compute_everything` [test](https://github.com/PlasmaControl/DESC/blob/master/tests/test_compute_funs.py) are changed in a PR, that test will fail.
+The three main reasons this could occur are:
+
+-  The PR was not intended to change how things are computed, but messed up something unexpected and now the compute quantities are incorrect, if you did not expect these changes in the PR then look into why these differences are happening and fix the PR.
+-  The PR updated the way one of the existing compute index quantities are computed (either by a redefinition or perhaps fixing an error present in ``master``)
+-  The PR added a new class parametrization (such as a new subclass of ``Curve`` like ``LinearCurve`` etc)
```
ddudt, line 166: Does this actually make the test fail? I don't think it will, but you still have to remember to add that new parameterization to the test as you explained. Also this applies to adding entirely new classes, not just new parameterizations of an existing class

CONTRIBUTING.rst
```diff
@@ -156,6 +156,46 @@ Opening a PR will trigger a suite of tests and style/formatting checks that must
...
+
+If the 2nd case is the reason, then you must update the ``master_compute_data.pkl`` file with the correct quantities being computed by your PR:
+
+-  First , run the test with ``pytest tests -k test_compute_everything`` and inspect the compute quantities whose values are in error, to ensure that only the quantities you expect to be different are shown (and that the new values are indeed the correct ones, you should have a test elsewhere for that though).
```
ddudt, line 170: lonely comma floating in empty space

CONTRIBUTING.rst
```diff
@@ -156,6 +156,46 @@ Opening a PR will trigger a suite of tests and style/formatting checks that must
...
+
+.. code-block:: python
+
+   except AssertionError as e:
+      error = False
+      update_master_data = True
+       print(e)
+
+
+-  rerun the test  ``pytest tests -k test_compute_everything`` , now any compute quantity that is different between the PR and master will be updated with the PR value
```
ddudt, line 189: another floating comma in empty space

CONTRIBUTING.rst
ddudt: Also "parameterization" is missing a second "e"

ddudt (changes requested):

CONTRIBUTING.rst
dpanici: I think my spelling of parametrization is right or at least is an accepted spelling: https://en.wikipedia.org/wiki/Parametrization_(geometry)#:~:text=Parametrization%20is%20a%20mathematical%20process,some%20independent%20quantities%20called%20parameters.

dpanici: and yea it does fail bc of this line in `test_compute_everything`

```python
assert things.keys() == data_index.keys(), (
        f"Missing the parameterizations {data_index.keys() - things.keys()}"
        f" to test against master."
    )
```

f0uriest (dismissed):

ddudt (dismissed):
Look how awkward this is , when there is an extra space between the word and the comma! 

You marked my comma comments as resolved without actually resolving them...

dpanici:
> Look how awkward this is , when there is an extra space between the word and the comma!
> 
> You marked my comma comments as resolved without actually resolving them...

I have them fixed locally but forgot to push yesterday, will push now

dpanici:
Final time I rerequest the review I promise, my bad for the forgotten push

f0uriest (approved):

ddudt (approved):

PR #923, opened 2024-03-04
Bump python versions

f0uriest:
Python 3.8 will reach end of life later this year, and some packages (such as diffrax) already require Python 3.9

This PR moves us in that direction, removing testing against Python 3.8 and adds it for Python 3.12 (the most recent).

Also updates docs accordingly.


docs/installation.rst
```diff
@@ -39,7 +39,7 @@ Option 1: Using pip to install packages (this will only install DESC + JAX with
 
 .. code-block:: sh
 
-    conda create --name desc-env 'python>=3.8, <=3.11'
+    conda create --name desc-env 'python>=9, <=3.12'
```
YigitElma, line 42: Is this a typo? 
`python>=3.9`

YigitElma (approved):

docs/installation.rst
```diff
@@ -204,7 +204,7 @@ Commit `a2fe711ffa3f` (an older version of the `master` branch) was tested to wo
 
     module load anaconda3/2020.11 cudatoolkit/11.1 cudnn/cuda-11.1/8.0.4
 
-    conda create --name desc-env python=3.8
```
dpanici, line 207: was this tested to work? I guess it should not really matter but the power architectures were tricky to work with install-wise

dpanici (approved):

docs/installation.rst
f0uriest: i had 3.9 installed when i was running stuff on traverse

PR #924, opened 2024-03-04
Fix incorrect toroidal components of surface basis vectors

dpanici:
- This fixes an error in our surface basis vectors, the `e_theta_z` and `e_zeta_t` did not have the proper `phi` components (they were zero when they really should have been `R_t`
- Adds a test for these surface (i.e. the ones without any `rho` derivs) basis vectors to make sure the `eq.compute` and `eq.surface.compute` methods agree when evaluating them on the `rho=1` surface.
- Also adds the appropriate surfaces as a parametrization for  `R_XX` and `Z_xx` derivatives in `compute._core`,


f0uriest:
Refactor to use base equilibrium compute methods, just add new parameterization for surfaces

desc/compute/_basis_vectors.py
```diff
@@ -26,10 +27,13 @@
     transforms={},
     profiles=[],
     coordinates="rtz",
-    data=["B", "|B|"],
+    data=["B", "|B|", "phi"],
+    basis="basis",
 )
 def _b(params, transforms, profiles, data, **kwargs):
     data["b"] = (data["B"].T / data["|B|"]).T
+    if kwargs.get("basis", "rpz").lower() == "xyz":
```
f0uriest, line 35: rather than having a conditional in every compute function, would it be easier to put it in `desc.compute.compute_utils._compute`?

IE, after computing a qty we check if dim==3 and what the requested basis was?

desc/compute/utils.py
```diff
@@ -95,6 +95,9 @@ def compute(parameterization, names, params, transforms, profiles, data=None, **
     return data
 
 
+# TODO: can we do the basis change here possibly? instead of repeating it everywhere?
```
f0uriest, line 98: ah i see you had the same thought.

A few options:
- basically everything is natively in rpz, so we could probably just enforce that all the compute functions return things in rpz and then if need be convert them to xyz here, for the few things natively in xyz it means transforming and transforming back but its just a 3x3 matmul so basically free
- could possibly get rid of the basis option entirely? and just make everything return in rpz and leave it up to the caller to then do any basis changes on top of that for their particular application. this might be the easiest option

tests/test_compute_funs.py
```diff
@@ -1660,6 +1660,26 @@ def test_surface_equilibrium_geometry():
     """Test that computing stuff from surface gives same result as equilibrium."""
     names = ["DSHAPE", "HELIOTRON", "NCSX"]
     data = ["A", "V", "a", "R0", "R0/a", "a_major/a_minor"]
+    # TODO: expand this to include all angular derivatives once they are implemented
```
f0uriest, line 1663: in the spirit of decreasing test time we probably don't need to test this against 3 different equilibria, can probably pick just one (a stellarator to get the 3d effects) and can crank down the resolution since the geometry doesnt care if its actually in force balance

dpanici: True, I had just copied the prior test that computed these things

desc/compute/utils.py
dpanici: I was thinking more along the lines of the first option as I do like the convenience of being able to request in either XYZ or RPZ basis. I did not include it here because this PR is really just trying to fix a bug and I can open another PR to then do this change for the basis stuff.

f0uriest (approved):

unalmis (approved):

desc/compute/utils.py
dpanici: I will leave it to a future PR

desc/equilibrium/equilibrium.py
```diff
@@ -856,7 +856,9 @@ def compute(
         dep1dz = [
             dep
             for dep in deps
-            if (data_index[p][dep]["coordinates"] == "z") and (dep not in data)
+            if (data_index[p][dep]["coordinates"] == "z")
+            and (dep not in data)
+            and dep not in ["phi", "zeta"]  # these don't need a special grid
```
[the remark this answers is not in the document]
dpanici: It lead to some issues where inside the `dep1dz` logic block we use the `compress` and `expand` methods, which fail if `jitable` grids are used. Maybe there is a better way but this works fine and tests pass

tests/test_compute_funs.py
```diff
@@ -1620,8 +1620,28 @@ def test_iota_components():
...
+    # for surfaces
+    data_basis_vecs_fourierRZ = [
+        "e_theta",
+        "e_zeta",
+        "e_theta_t",
+        "e_theta_z",
+        "e_zeta_t",
+        "e_zeta_z",
+    ]
+    data_basis_vecs_ZernikeRZ = [
```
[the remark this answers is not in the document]
dpanici: I sleep very fitfully if you must know

ddudt (approved):
Approving, but please wait to merge until after the dev meeting because I want to discuss the basis stuff in the compute logic

PR #925, opened 2024-03-06
VMEC wout fixes

### LINKED ISSUE #915: Sign difference in buco when writing a vmec input file vs writing a vmec wout file
To reproduce

Load a desc equilibrium that contains finite toroidal current

write a vmec wout file using
`VMECIO.save(eq, "wout_from_desc.nc")`

write a vmec input file using
`VMECIO.write_vmec_input(eq, "input.vmec_from_desc")`

run VMEC externally

The sign of `buco` will be flipped in the two instances.  However this is not so bad because the value of `iota` will agree in the two cases.

ddudt:
1. Resolves #915 by flipping the sign of `buco`
2. Also flips the sign of `jcuru` (same issue as above)
3. Adds the quantity `vp`, which is dV/ds / 4pi^2 [as documented here](https://github.com/PrincetonUniversity/STELLOPT/blob/2d7b671d6abb80b06fe93e4c2afa39cbae79a057/THRIFT/Sources/thrift_load_vmec.f90#L189)
4. Adds the quantity `nextcur` for the number of coils, which is hard-coded to 0 for fixed-boundary solutions
5. Refactors some of the scalar computations to make it more efficient and save compute time

`VMECIO.save` assumes the equilibrium was a fixed-boundary solution (we hard-code some outputs to their fixed-boundary values). But now that we have free-boundary working we could think about accommodating those free-boundary values (in a future PR). I think we would need to pass in the MagneticField that was used in addition to the Equilibrium.

### DIFF
--- a/desc/vmec.py
+++ b/desc/vmec.py
@@ -552,15 +552,9 @@ class VMECIO:
             ["<|B|^2>", "<J*B>", "sqrt(g)", "J^theta*sqrt(g)", "J^zeta", "D_Mercier"],
             grid=grid_full,
         )
-        data_half = eq.compute(["I", "G"], grid=grid_half)
+        data_half = eq.compute(["I", "G", "V_r(r)"], grid=grid_half)
 
-        bdotb = file.createVariable("bdotb", np.float64, ("radius",))
-        bdotb.long_name = "flux surface average of magnetic field squared, on full mesh"
-        bdotb.units = "T^2"
-        bdotb[:] = grid_full.compress(data_full["<|B|^2>"])
-        bdotb[0] = 0
-
-        # currents
+        # half mesh quantities
         buco = file.createVariable("buco", np.float64, ("radius",))
         buco.long_name = "Boozer toroidal current I, on half mesh"
         buco.units = "T*m"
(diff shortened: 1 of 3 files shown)


dpanici (dismissed):

desc/vmec.py
```diff
@@ -552,19 +570,15 @@ def save(cls, eq, path, surfs=128, verbose=1):  # noqa: C901 - FIXME - simplify
             ["<|B|^2>", "<J*B>", "sqrt(g)", "J^theta*sqrt(g)", "J^zeta", "D_Mercier"],
             grid=grid_full,
         )
-        data_half = eq.compute(["I", "G"], grid=grid_half)
+        data_half = eq.compute(["I", "G", "<|B|^2>", "V_r(r)"], grid=grid_half)
```
f0uriest, line 573: We also compute the profiles on the half grid around line 195, can that be combined with this?

There's also another half and full grid data computed on line ~833 that could be combined here

f0uriest (changes requested):
See comment about combining duplicate calculations on half/full grid

f0uriest (approved):

desc/vmec.py
```diff
@@ -258,6 +255,62 @@ def save(cls, eq, path, surfs=128, verbose=1):  # noqa: C901 - FIXME - simplify
         file.createDimension("dim_00100", 100)
         file.createDimension("dim_00200", 200)
 
+        # compute data
+        timer.start("compute")
+        if verbose > 0:
+            print("Computing data")
+
+        # desc throws NaN for r == 0, so we use something small to approximate it
```
dpanici, line 263: I thought that we have axis limits implemented for most things?

ddudt: I think this was left over from before we implemented the axis limits. Updated now, thanks.

f0uriest (approved):

PR #926, opened 2024-03-06
Add warnings about grid resolution to objective build methods

f0uriest:
This is to try to prevent the footgun of users giving a grid for a profile-like objective that only has radial resolution.

I added warnings to all the objectives I could think of where it might make sense, but open to other suggestions.

### DIFF
--- a/desc/objectives/_profiles.py
+++ b/desc/objectives/_profiles.py
@@ -5,7 +5,7 @@ import numpy as np
 from desc.compute import compute as compute_fun
 from desc.compute.utils import get_profiles, get_transforms
 from desc.grid import LinearGrid
-from desc.utils import Timer
+from desc.utils import Timer, warnif
 
 from .normalization import compute_scaling_factors
 from .objective_funs import _Objective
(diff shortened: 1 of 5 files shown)


desc/objectives/_profiles.py
```diff
@@ -264,6 +264,19 @@ def build(self, use_jit=True, verbose=1):
...
+        warnif(
+            grid.num_theta < 2 * eq.M,
+            RuntimeWarning,
+            "RotationalTransform objective grid requires poloidal "
+            "resolution for surface averages",
+        )
+        warnif(
+            grid.num_zeta < 2 * eq.N,
+            RuntimeWarning,
+            "RotationalTransform objective grid requires toroidal "
```
dpanici, line 276: technically if an iota profile is assigned, then no surface averages are used right? Maybe add another conditional checking that eq.iota is None?

dpanici (changes requested):
for RotationalTransform objective, check if an iota profile is assigned before warning as if so, then no surface averages are needed

dpanici (approved):

PR #927, opened 2024-03-06
Fixes that were missed in #883

f0uriest:


dpanici (approved):

desc/vmec.py
```diff
@@ -469,7 +469,7 @@ def save(cls, eq, path, surfs=128, verbose=1):  # noqa: C901 - FIXME - simplify
         chi[:] = (
             -2  # negative sign for negative Jacobian
             * Psi
-            * integrate.cumtrapz(r_full * iotaf[:], r_full, initial=0)
+            * integrate.cumulative_trapezoid(r_full * iotaf[:], r_full, initial=0)
```
ddudt, line 472: Did they really deprecate this function because of the name? lol

ddudt (approved):

dpanici (changes requested):
Seems like at least in the versions used in the PR here
- there is an uncaught warning from the scipy constrained optimizers that fails a test
- the conversion to mtanh is also slightly off from the expected values now?

desc/vmec.py
f0uriest: yeah there was a whole github issue where they argued about it.

dpanici (approved):

PR #931, opened 2024-03-07
Change Incompatible constraints to warn, add error if too large of a difference

### LINKED ISSUE #921: Change error to warning when particular solution to linear constraints does not satisfy them to the specified error threshold
Right now we have a pretty tight error tolerance when we check that a set of linear constraints is satisfiable (https://github.com/PlasmaControl/DESC/blob/50e805dfdf4c2abe3d4bb0e7627b6c184d2cea43/desc/objectives/utils.py#L159) and we throw an error if this assert fails.

However there are cases where the error is on the order of say `rtol=1e-13`, which can happen just due to poor scaling when different constraints have highly varying orders of magnitude, or when the geometric scale of an equilibrium becomes very large. We probably want the solver to continue in these cases, so it might be better to use a `warnif` statement here instead of throwing an error, so that the user is aware then this is violated but then can use their judgement to see if it is because two fundamentally incompatible constraints were used (in which case error would be very large), or if it is just due to scaling issues between the variables (in which case the rtol would still be quite low)

dpanici:
Resolves #921

### DIFF
--- a/desc/objectives/utils.py
+++ b/desc/objectives/utils.py
@@ -169,17 +169,20 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
         rtol = 3e-14
 
         try:
-            np.testing.assert_allclose(y1, y2, atol=atol, rtol=rtol)
-        except AssertionError:
-            abs_diff = np.abs(y1 - y2)
-            nonzero = np.where(y2 != 0)
+            np.testing.assert_allclose(
+                y1,
+                y2,
+                atol=atol,
+                rtol=rtol,
+                err_msg="Incompatible constraints detected, cannot satisfy constraint "
+                + f"{con}.",
+            )
+        except AssertionError as e:
             warnif(
                 True,
                 UserWarning,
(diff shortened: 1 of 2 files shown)


desc/objectives/utils.py
```diff
@@ -149,15 +149,39 @@ def recover(x_reduced):
...
 
+        # else check with tighter tols and throw an error, these tolerances
+        # could be tripped due to just numerical roundoff or poor scaling between
+        # constraints, so don't want to error out but we do want to warn the user.
+        atol = 3e-14
+        rtol = 3e-14
+
+        try:
+            np.testing.assert_allclose(y1, y2, atol=atol, rtol=rtol)
+        except AssertionError:
```
f0uriest, line 173: might be cleaner to use the existing error message:
```python
except AssertionError as e:
    warnings.warn(str(e) + "\n This may indicate incompatible constraints, or be due to floating point error.")

f0uriest (approved):

PR #932, opened 2024-03-07
Take zernike_radial outside for loops

YigitElma:


desc/equilibrium/equilibrium.py
```diff
@@ -628,23 +628,39 @@ def get_surface_at(self, rho=None, theta=None, zeta=None):
 
             # TODO: Don't call zernike_radial for each mode, instead compute all
             # Populate i, j, l, m and call zernike_radial once
+            arg_zernike = []
+            Js = []
```
f0uriest, line 632: would it be easier to just evaluate
```python
x = zernike_radial(rho, self.R_basis.modes[:,0], self.R_basis.modes[:,1])
for i, (l,m,n)
...
AR[j,i] = x[i]
```

YigitElma: Probably but I was not sure how many nodes we really needed to calculate, so only evaluated the required ones.

desc/equilibrium/equilibrium.py
```diff
@@ -628,23 +628,39 @@ def get_surface_at(self, rho=None, theta=None, zeta=None):
 
             # TODO: Don't call zernike_radial for each mode, instead compute all
```
dpanici, line 629: remove this TODO if this is completed by this PR

f0uriest (approved):

PR #933, opened 2024-03-07
Remove vaccum objective option from ``Equilibrium.solve``

### LINKED ISSUE #809: Add warning for vacuum solve with iota profile

f0uriest:
- Removes `"vacuum"` option from ``Equilibrium.solve``, ``solve_continuation`` and ``get_equilibrium_objective``, since it is non-trivial to check that the user didn't make a mistake with assigning compatible profiles.
- Keeps the option from the input file, but now just switches to force balance and explicitly drops any user supplied profiles and warns.

Resolves #809


desc/input_reader.py
```diff
@@ -902,11 +908,15 @@ def desc_output_to_input(  # noqa: C901 - fxn too complex
...
+        prof_modes = np.zeros((L_profile, 3))
+        prof_modes[:, 0] = np.arange(L_profile)
+        p1 = copy_coeffs(pres_profile.params, pres_profile.basis.modes, prof_modes)
+        p2 = copy_coeffs(profile.params, profile.basis.modes, prof_modes)
         f.write("\n# pressure and rotational transform/current profiles\n")
         for l in range(L_profile):
             f.write(
                 "l: {:3d}  p = {:15.8E}  {} = {:15.8E}\n".format(
-                    int(l), pres_profile.params[l], char, profile.params[l]
+                    int(l), p1[l], char, p2[l]
```
f0uriest, line 919: previously this was writing the wrong mode numbers if the profiles were even power series. The logic above now copies them all to a non-symmetric basis before outputting

desc/input_reader.py
```diff
@@ -558,20 +563,20 @@ def parse_inputs(self, fname=None):  # noqa: C901 - FIXME: simplify this
...
-                del inputs["current"]
-            else:  # if vacuum objective from input file, use zero current
-                del inputs["iota"]
-        else:
-            del inputs["iota"]
-
-        if inputs["objective"] == "vacuum" and (pres_flag or iota_flag or curr_flag):
+        if vac_flag and (pres_flag or iota_flag or curr_flag):
             warnings.warn(
                 "Vacuum objective does not use any profiles, "
```
dpanici, line 568: Is it better to say that the vacuum objective flag assumes p=0 and I=0 as the profiles? Since we have changed it to now just be force balance with 0 p and 0 I?

dpanici (approved):
Change

dpanici (approved):

ddudt (approved):

PR #854, opened 2024-02-04
Bounce averaging

unalmis:
This PR adds functionality to compute bounce averages in DESC

- [x] Differentiable algorithm to compute bounce points and integrals.
- [x] Works with any numerical quadrature
- [x] Fixed bugs with numpy compatibility.

Related
- #1003 
- #1042 
- #1229 
- #1196

PR #934, opened 2024-03-07
Add new method for copying data between grids

### LINKED ISSUE #913: Utility for copying data between grids that share nodes
EG, when computing iota on specific surfaces we then need to copy those iota values to another grid with the same rho but different theta/zeta.

Right now we do this using `grid.compress` and `grid.expand` but that doesn't really work for the sort of unstructured grids that get created under jit that we need for field line following stuff.

Heres a rough draft:
```python
def copy_data_grids(x, grid1, grid2, match="rpz", tol=1e-14):
    """Copy data from one grid to another at matching nodes.
    
    Given data x corresponding to nodes of grid1, copy data
    to a new array that corresponds to grid2.
    
    Parameters
    ----------
    x : ndarray, shape(grid1.num_nodes,...)
        Data to copy
    grid1, grid2: Grid
        Grid to copy from and copy to
    match {"r", "p", "z"}
        What coordinates to match. For local quantities,
        all three coordinates must match. For flux surface 
        qtys, only rho must match, etc.
    tol : float
        tolerance for considering nodes the same.
        
    Returns
    -------
    y : ndarray, shape(grid2.num_nodes, ...)
        Data copied to grid2
    """
    axes = ()
    if "r" in match:
        axes += (0,)
    if "t" in match:
        axes += (1,)
    if "z" in match:
        axes += (2,)
    y = jnp.zeros((grid2.num_nodes, *x.shape[1:]))
    
    def body(i, y):
        y = jnp.where(jnp.max(jnp.abs((grid2.nodes[:,axes] - grid1.nodes[i,axes])), axis=-1) < tol, x[i], y)
        return y
    
    y = fori_loop(0, grid1.num_nodes, body, y)
    return y
    
```

From what @YigitElma learned doing the zernike radial stuff loops might not be the best option, so vmap or some other tricks might be better. Anybody want to try out a few options?

f0uriest:
- `unique_X_idx` and `inverse_X_idx` are now not set when `jitable=True`, instead the caller can pass in those arrays manually if the structure of the grid is known (ie, field line on a single surface)
- `grid.compress`/`grid.expand` will now error if unique/inverse indices are not known
- adds new method `grid.copy_data_from_other` for handling copying of data between grids. If unique indices are known this just uses `compress`/`expand`, otherwise it uses a loop

Resolves #913


unalmis:
It looks like the test is is failing because an average is attempted on an unstructured grid. I would be surprised if this went unnoticed on master since this line throws an error when `masks` has size 0 (as it would when the unique and inverse idx were empty arrays).  https://github.com/PlasmaControl/DESC/blob/68c971fe0979ef5dfb2c28791c8b7d198389c591/desc/compute/utils.py#L969
The issue with doing averages on unstructured grids is that we don't know how many sums to do (since we can't jit jnp.unique), so to make this jitable in general we'd need to assume there are `grid.num_nodes` sums to do.

Since we are unable to do this:
https://github.com/PlasmaControl/DESC/blob/c0512453d129e0bc30c5ed3775bcd18a6b058207/desc/compute/utils.py#L923

My suggestion is to call this loop to use on unstructured grids:

```python
# method in desc.compute.utils
@partial(vmap, in_axes=(None, None, 0))
def _integrate(nodes, integrands, i):
    mask = nodes[:, idx] == nodes[i, idx]
    return jnp.tensordot(mask, integrands, axes=0)

# the usual api returning a function 
def surface_integrals():
  idx = {"rho": 0, "theta": 1, "zeta": 2}
  def _surface_integrals(q=jnp.array([1.0])):
        """Compute a surface integral for each surface in the grid."""
        integrands = (spacing * jnp.nan_to_num(q).T).T
        # note that this result is already expanded.
        result = integrate(grid.nodes, integrands, jnp.arange(grid.num_nodes))

return _surface_integrals

**Comment by unalmis** [comment]
Rory I pushed a commit here to fix the issue. I chose the vectorized version:


```python
def body(i, y):
    mask = jnp.abs(nodes - nodes[i]) <= tol
    y = jnp.where(mask, jnp.tensordot(mask, integrands, axes=1), y)
    return y
        
integrals = fori_loop(0, grid.num_nodes, body, jnp.zeros(integrands.shape))
```

unalmis:
Rory I pushed a commit to fix the issue.
You'll notice that an array of size $N^6 / 8$ (where grid.nodes has size $3N^3$) is constructed as the first argument to this function
```python
integrals = jnp.tensordot(
    jnp.abs(nodes - nodes[:, jnp.newaxis]) <= tol, integrands, axes=1
)
```
Usually that's too big, but the reason I chose to do that over a looped implementation:
```python
def body(i, y):
    mask = jnp.abs(nodes - nodes[i]) <= tol
    y = jnp.where(mask, jnp.tensordot(mask, integrands, axes=1), y)
    return y

integrals = fori_loop(0, grid.num_nodes, body, jnp.zeros(integrands.shape))
```
is because the looped implementation, which uses a mask of size $N^3 / 8$ would freeze my computer when I tried to differentiate it, which is usually indicative of an out of memory error.

The vectorized version, which should have used more memory, differentiated fine. I've attached my script here if you wanted to try to reproduce this.
[timing and memory.zip](https://github.com/PlasmaControl/DESC/files/14734536/timing.and.memory.zip)

desc/grid.py
```diff
@@ -345,15 +390,74 @@ def expand(self, x, surface_label="rho"):
...
+            y = self.expand(xc, surface_label)
+        except AttributeError:
+            self_nodes = jnp.asarray(self.nodes)
+            other_nodes = jnp.asarray(other_grid.nodes)
+            axis = {"rho": 0, "theta": 1, "zeta": 2}[surface_label]
+            y = jnp.zeros((self.num_nodes, *x.shape[1:]))
+
+            def body(i, y):
+                y = jnp.where(
+                    jnp.abs(self_nodes[:, axis] - other_nodes[i, axis]) <= tol,
```
unalmis, line 452: I changed this from `<` to `<=` in case people pass in 0

unalmis:
I'll stop pushing to this branch now and hand it back to you in case you want make edits.

f0uriest:
@unalmis can you try merging this into #854 and see if it resolves #958 ?

unalmis (approved):
Looks good, but the changes did not (were not sufficient to?) resolve #958.

desc/grid.py
```diff
@@ -441,9 +550,10 @@ def __init__(self, nodes, sort=False, jitable=False):
             ) = self._find_unique_inverse_nodes()
             self._weights = self._scale_weights()
 
-        self._L = self.num_rho
-        self._M = self.num_theta
-        self._N = self.num_zeta
+        self._L = self.num_nodes
+        self._M = self.num_nodes
+        self._N = self.num_nodes
```
unalmis, line 555: Might be better to set it to 0 in case users think its a resolution. For example LinearGrid(L, M, N) would we very large if each were grid.num_nodes.

dpanici: This is probably fine, it is actually pretty rare that we use `grid.L` etc., and even rarer that we use it in order to make another grid

tests/test_grid.py
```diff
@@ -785,30 +785,53 @@ def test_find_least_rational_surfaces():
...
-        "unique_zeta_idx",
-        "inverse_zeta_idx",
-    ]
-
-    for attr in attrs:
-        np.testing.assert_array_equal(
-            getattr(grid1, attr), getattr(grid2, attr), err_msg=attr
-        )
+    eq.change_resolution(3, 3, 3, 6, 6, 6)
+    # field lines on two surfaces
```
dpanici, line 789: these are not actually data along 2 field lines though?

dpanici (approved):

PR #935, opened 2024-03-07
Add field to things for boundary error objectives

### LINKED ISSUE #855: Add `field` to `things` in free boundary objective to allow single-stage free-boundary calculations
The free boundary objectives consider the external field to be fixed, but for single stage we would need field to be an optimizable thing for the objective

f0uriest:
- Adds the field parameters as optional inputs to `{Vacuum}BoundaryError.compute`
- Adds `field_fixed` option to objectives - currently this doesn't do anything since we usually can't precompute field qtys for a changing boundary, but could add special logic for certain field types later. Mainly this flag is for use in regular free boundary solve so you don't need to include explicit constraints targeting the field.

Resolves #855 

Should ideally merge #726 and #853 first so we have some sensible coil constraints before doing single stage stuff


desc/io/equilibrium_io.py
ddudt, line 1: Why were these changes necessary for this PR?

f0uriest: `SplineMagneticField` requires some custom behavior to work correctly as a pytree, and the previous method of defining a custom flattening method didn't consider all the extra attributes that get added with the `Optimizable` base class. This refactors things a bit so that each class can use the default flattening method but also define specific attributes as static or dynamic without having to re-define the entire flattening/unflattening method (and then having to account for every attribute individually)

f0uriest:
Note benchmarks are expected to fail due to updated argument name (renamed `ext_field -> field`)

ddudt (approved):

dpanici (approved):

PR #950, opened 2024-03-21
Add Normal Field Error as a plot quantity to `plot_2d` and `plot_3d` and wrapper for Bplasma from virtual casing

### LINKED ISSUE #910: plotting function for B-normal
Basically a way to visualize the errors returned by `QuadraticFlux.compute` from #726 

Want 2D and 3D options, similar to `plot_2d` and `plot_3d`

### LINKED ISSUE #965: Make wrapper function for computing the virtual casing biot savart integral field on the plasma surface
Should have a higher level wrapper to do the following code (calulcating the keys, making the interpolator, then passing those into the virtual casing biot savart function) (taken from `compute_Bnormal`)

```python
   # do virtual casing to find plasma contribution to Bnormal
            data_keys = [
                "K_vc",
                "B",
                "R",
                "phi",
                "Z",
                "e^rho",
                "n_rho",
                "|e_theta x e_zeta|",
            ]
            vc_eval_data = eq.compute(data_keys, grid=eval_grid)
            vc_data = eq.compute(data_keys, grid=vc_source_grid)

            k = min(
                vc_source_grid.num_theta, vc_source_grid.num_zeta * vc_source_grid.NFP
            )
            s = k - 1
            k = min(
                vc_source_grid.num_theta, vc_source_grid.num_zeta * vc_source_grid.NFP
            )
            q = k // 2 + int(np.sqrt(k))
            try:
                interpolator = FFTInterpolator(eval_grid, vc_source_grid, s, q)
            except AssertionError as e:
                print(
                    f"Unable to create FFTInterpolator, got error {e},"
                    "falling back to DFT method which is much slower"
                )
                interpolator = DFTInterpolator(eval_grid, vc_source_grid, s, q)
            if hasattr(eq.surface, "Phi_mn"):
                vc_data["K_vc"] += eq.surface.compute("K", grid=vc_source_grid)["K"]
            Bplasma = virtual_casing_biot_savart(
                vc_eval_data,
                vc_data,
                interpolator,
            )
            # need extra factor of B/2 bc we're evaluating on plasma surface
            Bplasma = Bplasma + vc_eval_data["B"] / 2
            Bnorm += np.sum(Bplasma * vc_eval_data["n_rho"], axis=-1)
```

dpanici:
- Add `"B*n"` as a plottable quantity to `plot_2d` and `plot_3d`. If passed in, then you must also pass in a `field` kwarg that is a `MagneticField` object and optionally a `field_grid` kwarg as well for the source grid of the magnetic field.

Resolves #910 
Resolves #965


desc/plotting.py
```diff
@@ -626,8 +630,28 @@ def plot_2d(
     plot_axes = _get_plot_axes(grid)
     if len(plot_axes) != 2:
         return ValueError(colored("Grid must be 2D", "red"))
-
-    data, label = _compute(eq, name, grid, kwargs.pop("component", None))
+    component = kwargs.pop("component", None)
+    if name != "Bn":
```
ddudt, line 634: Can we change the name to `"B*n"` to better match our existing format for other compute quantities? For example, we already have stuff like `"B*grad(|B|)"`

desc/plotting.py
```diff
@@ -599,6 +599,10 @@ def plot_2d(
         * ``ylabel_fontsize``: float, fontsize of the ylabel
         * ``cmap``: str, matplotlib colormap scheme to use, passed to ax.contourf
         * ``levels``: int or array-like, passed to contourf
+        * ``field``: MagneticField, a magnetic field with which to calculate Bn on
```
ddudt, line 602: Idea: to better match the Objectives API, we could change the first argument `eq` to `things`, which would normally be the equilibrium but in this case would be the tuple `(eq, field)`. Instead of adding a new kwarg. And this would also be more general for adding new plot quantities in the future.

desc/plotting.py
```diff
@@ -626,8 +630,28 @@ def plot_2d(
...
+            ValueError,
+            "If Bn is entered as the variable to plot, a magnetic field"
+            " must be provided.",
+        )
+        field_grid = kwargs.pop("field_grid", None)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            data, _ = field.compute_Bnormal(eq, eval_grid=grid, source_grid=field_grid)
+        data = data.reshape((grid.num_theta, grid.num_rho, grid.num_zeta), order="F")
+        label = r"$|B_n| ~(\mathrm{T})$"
```
ddudt, line 654: similar to the above comment, change this label to `r"$|\mathbf{B}\cdot\hat{n}| ~(\mathrm{T})$"` to better match our existing compute quantities

desc/plotting.py
```diff
@@ -851,18 +879,32 @@ def plot_3d(
     levels = kwargs.pop("levels", None)
     component = kwargs.pop("component", None)
 
+    if name != "Bn":
```
ddudt, line 882: same comments for `plot_3d` as for `plot_2d`

ddudt (changes requested):

desc/plotting.py
```diff
@@ -626,8 +630,28 @@ def plot_2d(
...
+        errorif(
+            field is None,
+            ValueError,
+            "If Bn is entered as the variable to plot, a magnetic field"
+            " must be provided.",
+        )
+        field_grid = kwargs.pop("field_grid", None)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            data, _ = field.compute_Bnormal(eq, eval_grid=grid, source_grid=field_grid)
```
f0uriest, line 652: will this only be correct for vacuum equilibria? What if is a nonozero Bn from the plasma field?

f0uriest (commented):
should figure out what to do for finite plasma field.

Also might be worth adding an option to normalize by total B?

desc/plotting.py
dpanici: Right now have it using the same plot grid for eval and source, I figure most times it will be fine. Can change it to be separate grids if needed

desc/plotting.py
dpanici: This is a good idea, but not necessarily needed by this PR. I will make an issue for it

desc/plotting.py
f0uriest: I think same grid is probably fine, the plotting resolution is already usually much higher than what we use during optimization.

desc/plotting.py
```diff
@@ -626,8 +635,62 @@ def plot_2d(
...
+                    "R",
+                    "phi",
+                    "Z",
+                    "e^rho",
+                    "n_rho",
+                    "|e_theta x e_zeta|",
+                ],
+                grid=grid,
+            )
+            k = min(grid.num_theta, grid.num_zeta * grid.NFP)
```
f0uriest, line 671: remind me to make a higher level wrapper to do this sort of stuff for common use cases

desc/plotting.py
```diff
@@ -626,8 +635,62 @@ def plot_2d(
...
+            except AssertionError:
+                interpolator = DFTInterpolator(grid, grid, s, q)
+            if hasattr(eq.surface, "Phi_mn"):
+                vc_data["K_vc"] += eq.surface.compute("K", grid=grid)["K"]
+            Bplasma = virtual_casing_biot_savart(
+                vc_data,
+                vc_data,
+                interpolator,
+            )
+            # need extra factor of B/2 bc we're evaluating on plasma surface
```
f0uriest, line 686: technically we don't need this since the equilibrium B always has zero normal component

ddudt (changes requested):
I don't think this is working as intended for `plot_3d`. This is the plot I get for the new test function `test_3d_plot_Bn`: 

plot_3d_Bn

dpanici:
> I don't think this is working as intended for `plot_3d`. This is the plot I get for the new test function `test_3d_plot_Bn`:
> 
> plot_3d_Bn

THis is due to the default grid having endpoint=True, while the singular integral needs endpoint=False to work properly. I will fix this

dpanici:
put in compute bnormal function

desc/plotting.py
dpanici: Ah is this only needed for the pressure balance?

f0uriest: yeah

desc/magnetic_fields/_core.py
```diff
@@ -252,9 +273,64 @@ def compute_Bnormal(
...
+                    "K_vc",
+                    "B",
+                    "R",
+                    "phi",
+                    "Z",
+                    "e^rho",
+                    "n_rho",
+                    "|e_theta x e_zeta|",
+                ],
+                grid=vc_source_grid,
```
f0uriest, line 304: think we need an actual default for this, otherwise it will use a quadrature grid

desc/plotting.py
```diff
@@ -626,8 +630,55 @@ def plot_2d(
...
+                    NFP=grid.NFP,
+                    endpoint=False,
+                )
+            else:
+                vc_grid = grid
+            data, _ = field.compute_Bnormal(
+                eq, eval_grid=vc_grid, source_grid=field_grid, vc_source_grid=vc_grid
+            )
+        data = data.reshape((vc_grid.num_theta, vc_grid.num_zeta), order="F")
+        if grid.endpoint:
```
f0uriest, line 675: shouldnt this be `if not grid.endpoint`?

desc/plotting.py
```diff
@@ -626,8 +630,55 @@ def plot_2d(
...
+            "If B*n is entered as the variable to plot, "
+            "the grid nodes must be at rho=1.",
+        )
+
+        field_grid = kwargs.pop("field_grid", None)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+
+            if grid.endpoint:
+                print("changing grids")
```
f0uriest, line 661: rogue print statement

desc/plotting.py
```diff
@@ -851,18 +906,57 @@ def plot_3d(
...
+                    NFP=grid.NFP,
+                    endpoint=False,
+                )
+            else:
+                vc_grid = grid
+            data, _ = field.compute_Bnormal(
+                eq, eval_grid=vc_grid, source_grid=field_grid, vc_source_grid=vc_grid
+            )
+        data = data.reshape((vc_grid.num_theta, vc_grid.num_zeta), order="F")
+        if grid.endpoint:
```
f0uriest, line 949: see above

tests/test_plotting.py
```diff
@@ -1005,3 +1005,30 @@ def test_plot_omnigenous_field():
     )
     fig, ax = plot_boozer_surface(field, iota=0.6, fieldlines=4)
     return fig
+
+
+@pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_2d)
+@pytest.mark.unit
+def test_2d_plot_Bn():
+    """Test 2d plotting of Bn on equilibrium surface."""
+    eq = get("precise_QA")
```
f0uriest, line 1014: recommend using a lower resolution equilibrium here

desc/magnetic_fields/_core.py
```diff
@@ -238,23 +256,91 @@ def compute_Bnormal(
...
-            surface = surface.surface
+            calc_Bplasma = True
+            eq = surface
+            surface = eq.surface
         if eval_grid is None:
             eval_grid = LinearGrid(
                 rho=jnp.array(1.0), M=2 * surface.M, N=2 * surface.N, NFP=surface.NFP
             )
+        if vc_source_grid is None:
+            source_NFP = surface.NFP if surface.N > 0 else 64
```
dpanici, line 271: Question on this, when N of the eq is zero, we make the virtual casing source grid NFP be 64. But if the eq N is 0 then so is the eq N_grid right? so how is it, for example in the tutorial axisymmetric free boundary solve, that the toroidal effects are accounted for if the source grid has no toroidal grid nodes?

f0uriest: there's always still nodes at zeta=0, and NFP=64 means we repeat this 64 times around the torus. There's no toroidal coupling though so each of those can be done independently.

docs/notebooks/tutorials/free_boundary_equilibrium.ipynb
f0uriest, line 1: what was changed here?

f0uriest: nvm see the new plot

f0uriest (dismissed):

ddudt (approved):
I made a formatting change to reduce the number of lines of code, and also changed to `endpoint=True` in the 3d plot test since that's the more common use case and covers more of the code

f0uriest (dismissed):

f0uriest (approved):

PR #952, opened 2024-03-24
Add a `from_values` method to fit FourierRZCurves with data

dpanici:
- Add a `from_values` method to `FourierRZCurve` class
  - this method will only allow fitting if the provided points are monotonic in `phi`


desc/geometry/core.py
```diff
@@ -266,8 +266,41 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
+            NOTE: Some types of curves may not be representable in this basis.
+
+        Parameters
+        ----------
+        N : int
+            Fourier resolution of the new R,Z representation.
+        grid : Grid, int or None
+            Grid used to evaluate curve coordinates on to fit with FourierRZCurve.
+            If an integer, uses that many equally spaced points.
+        s : ndarray, optional
```
f0uriest, line 281: Isn't this redundant with grid?

desc/geometry/core.py
```diff
@@ -266,8 +266,41 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
             coords, knots=knots, method=method, name=name, basis="xyz"
         )
 
-    # TODO: to_rz method for converting to FourierRZCurve representation
-    # (might be impossible to parameterize some curves with toroidal angle phi)
+    def to_FourierRZ(self, N=None, grid=None, s=None, NFP=None, name=""):
+        """Convert Curve to FourierRZCurve representation.
+
+            NOTE: Some types of curves may not be representable in this basis.
```
f0uriest, line 272: See numpydoc for formatting notes

f0uriest (approved):

desc/geometry/curve.py
```diff
@@ -225,6 +225,59 @@ def from_input_file(cls, path):
...
+
+        grid = LinearGrid(zeta=phi, NFP=1, sym=sym)
+        basis = FourierSeries(N=N, NFP=NFP, sym=sym)
+        transform = Transform(grid, basis, build_pinv=True)
+        R_n = transform.fit(R)
+        Z_n = transform.fit(Z)
+        return FourierRZCurve(R_n=R_n, Z_n=Z_n, NFP=NFP, name=name, sym=sym)
+
 
 def _unclose_curve(X, Y, Z):
```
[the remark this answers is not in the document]
f0uriest: For helper functions like this I think it's better to have them near where they're used

f0uriest (dismissed):

desc/geometry/curve.py
```diff
@@ -225,6 +225,60 @@ def from_input_file(cls, path):
...
+        # easiest to check closure in XYZ coordinates
+        X, Y, Z, _, _, _, input_curve_was_closed = _unclose_curve(X, Y, Z)
+        if input_curve_was_closed:
+            R = R[0:-1]
+            phi = phi[0:-1]
+            Z = coords_rpz[0:-1, 2]
+
+        # assert the curve is not doubling back on itself in ph,
+        # which can't be represented with a curve parameterized by phi
+        assert np.all(np.diff(phi) > 0), "Supplied phi must be monotonic"
```
ddudt, line 273: spelling mistake for "phi"

also we typically use `errorif` to raise a ValueError instead of plain asserts and I like consistency

f0uriest (approved):

desc/geometry/curve.py
```diff
@@ -225,6 +225,66 @@ def from_input_file(cls, path):
...
+        # easiest to check closure in XYZ coordinates
+        X, Y, Z, _, _, _, input_curve_was_closed = _unclose_curve(X, Y, Z)
+        if input_curve_was_closed:
+            R = R[0:-1]
+            phi = phi[0:-1]
+            Z = coords_rpz[0:-1, 2]
+        # check if any phi are negative, and make them positive instead
+        # so we can more easily check if it is monotonic
+        inds_negative = np.where(phi < 0)
+        phi = phi.at[inds_negative].set(phi[inds_negative] + 2 * np.pi)
```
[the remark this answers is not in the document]
dpanici: Yes, because the `atan2` function in `xyz2rpz` does not give values from `0` to `2pi` but rather from `-pi` to `pi`, so if you convert an array of x,y values that corresponds to `0` to `2pi` in phi, the `xyz2rpz` function will give `phi` values from `0` to `pi` then from `-pi` back to `0`, which makes checking monotonicity hard

ddudt (approved):

PR #961, opened 2024-03-27
Dommaschk Potential Miscellaneous Fixes and Improvements

### LINKED ISSUE #946: Implement field periodicity for Dommaschk field fitting
the `m` mode in Dommaschk potentials is basically the field periodicity, this should be an option in the fitting to exclude the non-field-periodic modes (can either only include $NFP*m$ modes, or rescale the potentials themselves (to avoid large denominators at high `m` numbers)

### LINKED ISSUE #960: At Z=0, some issues with Dommaschk fitting
related to `0**0` derivative

dpanici:
- Adds `NFP` to fitting function, so that only dommaschk potential modes with the corresponding discrete symmetry are included in the fit
-Changes `0**0` derivative trick to use double where, as it seems that the prior fix was causing some issues with the evaluation at Z=0

image

Fitted toroidal B from an elliptical vacuum equilibrium at rho=0.5 along a curve of constant theta (Z=0 at the first point)
The first point B is way off for some reason when the check and replace for `0**0` was present, but when removed this error vanishes (really am unsure of why this fix works but the proof is in the result)

image

same plot but after removing those lines where we check for 0**0 and replace the exponent with 1.

Resolves #960 
Resolves #946


desc/magnetic_fields/_dommaschk.py
```diff
@@ -79,7 +81,7 @@ def __init__(
 
     @classmethod
     def fit_magnetic_field(  # noqa: C901 - FIXME - simplify
-        cls, field, coords, max_m, max_l, sym=False, verbose=1
+        cls, field, coords, max_m, max_l, sym=False, verbose=1, NFP=1
```
f0uriest, line 84: kind of wondering now, would it be worth re-naming `m` to `n` here to be consistent with notation used elsewhere in the code?

Also do we want `max_m` (or whatever we call it) to be the actual max, or the max of the "effective" mode number m/NFP?

eg, elsewhere in the code if you set N=5 and NFP=3 you get modes `[0,5,10,15]`, so Nmax is really 15 although we would call that Nmax=5

f0uriest: same for possibly renaming `l -> m`

desc/magnetic_fields/_dommaschk.py
```diff
@@ -412,8 +417,6 @@ def D_m_n(R, Z, m, n):
     def body_fun(k, val):
         coef = CD_m_k(R, m, k) / gamma(n - 2 * k + 1)
         exp = n - 2 * k
-        # derivative of 0**0 is ill defined, so we do this to enforce it being 0
-        exp = jnp.where((Z == 0) & (exp == 0), 1, exp)
```
f0uriest, line 416: it feels weird that this works without this fix. This should return nan when differentiated:

```python
@jit
def foo(x, p):
    return x**p

jax.jacrev(foo)(0.0, 0)
# > Array(nan, dtype=float64, weak_type=True)
```

f0uriest: Ahh maybe this was giving the wrong derivative too. 

$x^0 = 0~ \forall~ x$, so $\partial_x x^0 = 0$ but this was making the derivative 1:

```python
@jit
def foo(x, p):
    p = jnp.where((x == 0) & (p == 0), 1, p)
    return x**p

jax.jacrev(foo)(0.0, 0)
# > Array(1., dtype=float64, weak_type=True)
```

I think we need a double where trick to enforce that it should be 0 with 0 derivative:
```python
@jit
def foo(x, p):
    mask = (x == 0) & (p == 0)
    p = jnp.where(mask, 1, p)
    return jnp.where(mask, 0, x**p)

jax.jacrev(foo)(0.0, 0)
# > Array(0., dtype=float64, weak_type=True)
```

desc/magnetic_fields/_dommaschk.py
dpanici: - the point of max m is taken, I can change that and make it clear in the docstring what this is
- I am slightly hesitant to change the notation only because this is the exact notation that matches the paper.

desc/magnetic_fields/_dommaschk.py
f0uriest: that's if p is hard-coded, but in our case its still a traced value because it changes in a loop. I added that logic because I was getting nan values and gradients without it, so I think something is probably needed

tests/test_magnetic_fields.py
```diff
@@ -854,55 +854,6 @@ def test_dommaschk_vertical_field():
     np.testing.assert_allclose(B_dom[:, 2], ones, atol=5e-15)
 
 
-@pytest.mark.unit
```
dpanici, line 857: This test was redundant with the one below, so I've removed it

desc/magnetic_fields/_dommaschk.py
```diff
@@ -412,9 +417,9 @@ def D_m_n(R, Z, m, n):
...
-        return val + coef * Z**exp
+        mask = (Z == 0) & (exp == 0)
+        exp = jnp.where(max, 1, exp)
```
f0uriest, line 421: typo, max -> mask

f0uriest (approved):

ddudt (approved):

PR #963, opened 2024-03-28
A few small improvements to reduce compile time/memory

f0uriest:
- integer and bool arrays are no longer treated as static values when doing JAX transformations (jit, etc). This avoids having them baked into the jitted function, which seems to avoid some of the "constant folding" issues
- The `Objective.compute` method now uses the lower level `desc.compute.utils._compute` directly instead of the outer one, which avoids some largely unnecessary runtime checks in cases where we know we're passing in the correct transforms etc.


desc/compute/utils.py
```diff
@@ -407,6 +413,13 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
     return transforms
 
 
+def has_data_dependencies(parameterization, qty, data, axis=False):
```
dpanici, line 416: this just removes some unneeded checks on transform and grid during the recursion?

dpanici (approved):

ddudt (approved):

PR #964, opened 2024-03-28
low-beta shifted circle geometric coefficients

rahulgaur104:
This PR adds some compute quantities such as the geometric factors in front of grad-B drifts and curvature drifts and their analytical expressions in tests/test_equilibrium.py for a low-beta shifted circle tokamak equilibrium. These quantities are used for calculating geometric coefficients such as epsilon effective and Gamma_c. 

The idea is to merge this into bounce #854 and use these tests to compare analytical and numerical bounce-averaged drifts.


desc/compute/_metric.py
```diff
@@ -1823,3 +1823,74 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
 def _gradzeta(params, transforms, profiles, data, **kwargs):
     data["|grad(zeta)|"] = jnp.sqrt(data["g^zz"])
     return data
+
+
+@register_compute_fun(
+    name="gbdrift",
+    label="\\mathrm{gradB-drift}",
```
f0uriest, line 1830: can you write out the math formula here? like `\\mathrm{gradB-drift} = 1/B^2 \mathbf{b} ..."`

Same for the other qtys, so that when they're plotted it has the actual formula as well as the gs2/gx name

desc/compute/_metric.py
```diff
@@ -1823,3 +1823,74 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    units="~",
```
f0uriest, line 1831: it doesn't look like this is actually unitless? shouldn't it be 1/Tm^2 or something?

desc/compute/_metric.py
```diff
@@ -1823,3 +1823,74 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    units_long="None",
+    description="Binormal component of th geometric part of the gradB drift"
```
f0uriest, line 1833: typo, th -> the. Also probably don't need all the line breaks

desc/compute/_metric.py
```diff
@@ -1823,3 +1823,74 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    + " used for local stability analyses,"
+    + " energetic particle proxy,"
+    + " effective field ripple",
+    dim=1,
+    params=[],
+    transforms={"grid": []},
```
f0uriest, line 1839: doesn't look like grid is required for this

desc/compute/_metric.py
```diff
@@ -1823,3 +1823,74 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    profiles=[],
+    coordinates="rtz",
+    data=["p_r", "psi_r", "|B|", "gbdrift"],
+)
+def _cvdrift(params, transforms, profiles, data, **kwargs):
+    mu_0 = 4 * jnp.pi * 1.0e-7
```
f0uriest, line 1870: elsewhere we just do `from scipy.constants import mu_0` at the top of the file

tests/test_equilibrium.py
```diff
@@ -464,3 +466,156 @@ def test_backward_compatible_load_and_resolve():
...
+    )
+
+    pressure = PowerSeriesProfile([1.5e5, 0, -1.5e5])
+    iota = PowerSeriesProfile([0.5, 0, -0.2])
+
+    eq = Equilibrium(
+        L=10, M=10, N=0, surface=surf, Psi=0.1, pressure=pressure, iota=iota
+    )
+
+    eq, _ = eq.solve(objective="force", ftol=1e-3, gtol=1e-6, maxiter=30, verbose=3)
```
f0uriest, line 497: do we need to actually do the solve as part of this test? Or can we just load a saved version stored in `tests/inputs`?

tests/test_equilibrium.py
```diff
@@ -464,3 +466,156 @@ def test_backward_compatible_load_and_resolve():
...
+    cvdrift0 *= sign_iota * sign_psi * shat * 2 / modB**3 / np.sqrt(psi)
+
+    cvdrift_2 = data["cvdrift"]
+    gbdrift_2 = data["gbdrift"]
+
+    ## Comparing coefficient calculation here with coefficients from compute
+    cvdrift_2 = -2 * sign_psi * Bref * Lref**2 * np.sqrt(psi) * data["cvdrift"]
+    gbdrift_2 = -2 * sign_psi * Bref * Lref**2 * np.sqrt(psi) * data["gbdrift"]
+
+    np.testing.assert_allclose(gbdrift, gbdrift_2, atol=1e-4)
```
f0uriest, line 591: in most other tests we've been trying to use `rtol` instead of `atol` where possible, mainly so when reading the tests we can get a sense of the expected % error without having to know the scale of the data

tests/test_equilibrium.py
```diff
@@ -464,3 +466,156 @@ def test_backward_compatible_load_and_resolve():
...
+    shat = -x / iotas * shears / Lref
+
+    grad_psi = data["grad(psi)"]
+    grad_alpha = data["grad(alpha)"]
+
+    iota = data["iota"]
+
+    gradpar = Lref * data["B^zeta"] / modB
+
+    gds21 = -sign_iota * np.array(dot(grad_psi, grad_alpha)) * shat / Bref
```
f0uriest, line 566: might be worth adding compute functions for these as well (or aliases to existing ones)

tests/test_equilibrium.py
rahulgaur104: I use atol because these quantities cross the y = 0 line multiple time and rtol breaks at points where we have 0/0 or 1/0 type situation.

tests/test_equilibrium.py
rahulgaur104: I think the only one missing is gds2 = |grad(alpha)|^2 but that is in the ballooning branch.

tests/test_equilibrium.py
rahulgaur104: No, we don't really have to calculate it every time a test is run. I can add this equilibrium to tests/inputs.

desc/compute/_metric.py
rahulgaur104: Done!

desc/compute/_metric.py
rahulgaur104: Removed.

desc/compute/_metric.py
rahulgaur104: Corrected.

desc/compute/_metric.py
rahulgaur104: Yes. Corrected.

desc/compute/_metric.py
rahulgaur104: Added LaTex expression.

tests/test_equilibrium.py
rahulgaur104: This test is only needed so that Kaya can use these formulae for bounce averaging. A much more thorough tests for the coefficients is in the adjoint ballooning branch.

desc/compute/_metric.py
rahulgaur104: Do need line breaks to avoid 88 character limit.

tests/test_equilibrium.py
dpanici: Using both atol and rtol can alleviate this, so that the atol can be made like 1e-8 or something very small, and that only activates when the data is crossing zero

tests/test_equilibrium.py
dpanici: `s_hat` instead of `shat`?

dpanici (changes requested):
- Add example reference
- atol and rtol tighter for test if possible
- maybe rename to be slightly more descriptive (though if this is normal nomenclature this is fine as well)

desc/compute/_metric.py
```diff
@@ -1823,3 +1825,79 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
     data["|grad(zeta)|"] = jnp.sqrt(data["g^zz"])
     return data
+
+
+@register_compute_fun(
+    name="gbdrift",
+    # Exact definition of the magnetic drifts taken from
+    # eqn. 48 of Introduction to Quasisymmetry by Landreman
+    # https://tinyurl.com/54udvaa4
+    label="\\mathrm{gradB-drift} = 1/B^{2} * (\\bm{b}\\times\\nabla (B)) \\dot"
```
f0uriest, line 1835: `\\bm` doesn't work in matplotlib, need to use `\\mathbf`
also `\\dot` puts an overdot, i think you want `\\cdot`

desc/compute/_metric.py
```diff
@@ -1823,3 +1825,79 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    + "\\nabla(\\alpha)",
+    units="1/(T-m^{2})",
+    units_long="None",
```
f0uriest, line 1838: not None

desc/compute/_metric.py
```diff
@@ -1823,3 +1825,79 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    )
+    return data
+
+
+@register_compute_fun(
+    name="cvdrift",
+    # Exact definition of the magnetic drifts taken from
+    # eqn. 48 of Introduction to Quasisymmetry by Landreman
+    # https://tinyurl.com/54udvaa4
+    label="\\mathrm{curvature-drift} = 1/B^{3} * (\\bm{b}\\times\\nabla(p + B^2/2))"
```
f0uriest, line 1862: see above

desc/compute/_metric.py
```diff
@@ -1823,3 +1825,79 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    + "\\dot\\nabla(\\alpha)",
+    units="1/(T-m^{2})",
+    units_long="None",
```
f0uriest, line 1865: not None

desc/compute/_metric.py
```diff
@@ -1823,3 +1825,79 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    data["cvdrift"] = 1 / data["|B|"] ** 2 * dp_dpsi + data["gbdrift"]
+    return data
+
+
+@register_compute_fun(
+    name="cvdrift0",
+    # Exact definition of the magnetic drifts taken from
+    # eqn. 48 of Introduction to Quasisymmetry by Landreman
+    # https://tinyurl.com/54udvaa4
+    label="\\mathrm{curvature-drift-1} = 1/B^{2} * (\\bm{b}\\times\\nabla(B))"
```
f0uriest, line 1886: see above

desc/compute/_metric.py
```diff
@@ -1823,3 +1825,79 @@ def _gradtheta(params, transforms, profiles, data, **kwargs):
...
+    + "\\dot\\nabla(rho)",
+    units="1/(T-m^{2})",
+    units_long="None",
```
f0uriest, line 1889: not None

tests/test_equilibrium.py
```diff
@@ -464,3 +465,143 @@ def test_backward_compatible_load_and_resolve():
...
+
+@pytest.mark.unit
+def test_shifted_circle_geometry():
+    """
+    In this test, we calculate a low-beta shifted circle equilibrium with DESC.
+
+    We then compare the various geometric coefficients with their respective analytical
+    expressions.
+    """
+    eq = Equilibrium.load(".//tests//inputs//low-beta-shifted-circle.h5")
```
f0uriest, line 478: looks like this file isn't included in the PR

tests/test_equilibrium.py
```diff
@@ -464,3 +465,143 @@ def test_backward_compatible_load_and_resolve():
...
     obj = ObjectiveFunction(f_obj)
     eq.solve(maxiter=1, objective=obj)
+
+
+@pytest.mark.unit
+def test_shifted_circle_geometry():
+    """
+    In this test, we calculate a low-beta shifted circle equilibrium with DESC.
+
+    We then compare the various geometric coefficients with their respective analytical
```
f0uriest, line 475: is there a reference for where the analytical forms come from?

rahulgaur104: There are a couple of references here and there but each one has slight variations. The point here is not to obtain the exact analytical expression with the right factors but to have something to compare with. 

For this case, we'll do the bounce-averaged drifts and the numerical solutions will be a function of the pitch angle whereas the analytical values will be of the form analytical_geometric_coefficients * Elliptic_integrals(function of the pitch angle) (from scipy).

tests/test_equilibrium.py
rahulgaur104: Done.

desc/compute/_metric.py
rahulgaur104: Changed

desc/compute/_metric.py
rahulgaur104: Corrected

desc/compute/_metric.py
rahulgaur104: Changed

desc/compute/_metric.py
rahulgaur104: Corrected

desc/compute/_metric.py
rahulgaur104: Corrected

tests/test_equilibrium.py
```diff
@@ -464,3 +465,143 @@ def test_backward_compatible_load_and_resolve():
...
+    )
+
+    cvdrift0 = np.array(dot(cross(data["B"], data["grad(|B|)"]), grad_psi))
+    cvdrift0 *= sign_iota * sign_psi * s_hat * 2 / modB**3 / np.sqrt(psi)
+
+    cvdrift_2 = data["cvdrift"]
+    gbdrift_2 = data["gbdrift"]
+
+    ## Comparing coefficient calculation here with coefficients from compute/_mtric
+    cvdrift_2 = -2 * sign_psi * Bref * Lref**2 * np.sqrt(psi) * data["cvdrift"]
```
f0uriest, line 569: wasn't this just defined above?

tests/test_equilibrium.py
f0uriest: that's fine, I just wanted a reference for where the formulas implemented below come from.

tests/test_equilibrium.py
rahulgaur104: Thanks! Changed.

tests/test_equilibrium.py
rahulgaur104: Ok, I have added a reference

f0uriest (approved):

tests/test_equilibrium.py
```diff
@@ -464,3 +465,141 @@ def test_backward_compatible_load_and_resolve():
...
+    gbdrift_2 = -2 * sign_psi * Bref * Lref**2 * np.sqrt(psi) * data["gbdrift"]
+
+    # The error here should be of the same order as the max force error
+    np.testing.assert_allclose(gbdrift, gbdrift_2, atol=1e-5, rtol=1e-5)
+    np.testing.assert_allclose(cvdrift, cvdrift_2, atol=8e-4, rtol=9e-5)
+
+    a0_over_R0 = Lref * np.sqrt(psi)
+
+    # For the rest of the expressions, the error ~ a0_over_R0
+    fudge_factor1 = -3.8
```
dpanici, line 577: what is the fudge factor for?

rahulgaur104: Because the real factor is a combination of iota, R0, a0, B0, pi, sqrt(2) and several such variables.

Note that bounce averaging operator is not going to include any of the geometric coefficients with these fudge factors. These factors are only used for the analytical geometric coefficients. Note that all fudge factors are actually scalars.

Let's say all the analytical geometric coefficients are in a vector X and their actual numerical values are in a vector Y.

The test is that
the numerical bounce-averaged drift  = f(Y, pitch_angle) 
evaluated numerically using the bounce-averaging PR must match 
the analytical bounce-averaged drift  = X[0] * E(pitch_angle) + X[1]* K(pitch_angle) 
where E and K are elliptic integrals. 

So we will chose a fixed set of pitch_angles and calculate the numerical solution and compare it with the analytical solution. The fudge factors make sure that the geometric coefficients are close enough (with a few percent) so that if the bounce-averaged drifts differ a lot, then something is wrong in a bounce-average PR.

Does that make sense?

f0uriest (approved):

tests/test_equilibrium.py
dpanici: Ah so these are from the thesis, like a fit to make the analytic expression match numerics? Or are they just so that the magnitudes are close, then the test is really to ensure that the trends of the analytic expressions match the numerical values (which are the "ground truth")?

dpanici (approved):

PR #966, opened 2024-03-29
Second Dommaschk Fix PR

dpanici:
- I realized there was an issue when using sym=True in the fitting, I used  floor divide`//` when I meant to use modulus `%`... this fixes that and adds a test.
- I stupidly thought the `0**0` thing was fixed but I was using a jupyter notebook to test and must not have correctly restarted it when I checked if it worked, as it was still not working correctly for the case I have. The issue was the value when the `0**0` case occurs should not be 0, but rather 1. This fixes that, and adds a test for that as well in case we fiddle with the potential again in the future.


f0uriest (approved):

ddudt (approved):
So embarrassing! But also we should have helped catch it in the review, sorry

PR #967, opened 2024-03-29
Improve heuristic initial guess for `map_coordinates`

f0uriest:


dpanici (dismissed):

unalmis:
I needed to add some logic to my bounce integral PR to transform from field line coords to desc coords with an accurate computation of the rotational transform for the initial guess. Is this useful for this PR?
https://github.com/PlasmaControl/DESC/pull/854/files#r1545524478

f0uriest:
Yeah that was my intention. These changes should make it possible to go from field line to DESC coordinates without needing to give an initial guess

desc/equilibrium/coords.py
```diff
@@ -98,27 +98,32 @@ def map_coordinates(  # noqa: C901
...
         x = jnp.array([data[k] for k in basis]).T
         return x
 
     @jit
     def residual(y, coords):
         xk = compute(y, inbasis)
-        r = xk % period - coords % period
-        return jnp.where(r > period / 2, -period + r, r)
+        r = periodic(xk) - periodic(coords)
+        return jnp.where((r > period / 2) & jnp.isfinite(period), -period + r, r)
```
dpanici, line 126: how did this change fix the function for negative values? I don't quite follow

f0uriest: we're using `period=np.inf` to denote no period, which is fine for quantities that are always positive (like `R`) because `1.23 % np.inf == 1.23`. However, `-1.23 % np.inf == np.inf` so it breaks for quantities that can be negative (like `Z`). This only does the modulo division where period is finite, which fixes the issue.

desc/equilibrium/coords.py
```diff
@@ -181,6 +161,66 @@ def _distance_body(i, idx):
...
+
+    # omega usually small (zero for now)
+    zeta = coords[:, inbasis.index(toroidal)]
+
+    if poloidal == "theta" or poloidal == "theta_PEST":  # lambda usually small
+        theta = coords[:, inbasis.index(poloidal)]
+    elif poloidal == "alpha":
+        alpha = coords[:, inbasis.index("alpha")]
+        iota = profiles["iota"](rho)
+        theta = (alpha + iota * zeta) % (2 * jnp.pi)
```
unalmis, line 196: When we're setting straight field line theta as the initial guess for desc theta,
I think this formula is only accurate if `zeta` is the toroidal coordinate in DESC coords.
If it is field line `zeta` then $(\rho, \alpha, \zeta) \neq (\rho, \alpha, \zeta + 2 \pi)$.
Why doesn't it make more sense to not modulo by 2pi for the initial guess?

dpanici:
double check jittability of map_coords is tested

unalmis (approved):

unalmis:
> double check jittability of map_coords is tested

basically was wondering how its jitable when a new grid is constructed, and
https://github.com/PlasmaControl/DESC/blob/5801ea8238a45b6f6674cc0d2ebe67d9d598242f/desc/equilibrium/equilibrium.py#L725
requires
https://github.com/PlasmaControl/DESC/blob/5801ea8238a45b6f6674cc0d2ebe67d9d598242f/desc/grid.py#L119

f0uriest:
@unalmis the key is that the grids created only depend on `eq.L_grid` etc, not on the values of the nodes being transformed. Instead of computing iota for each rho in the input data we just compute the entire profile and then we don't need to worry about mapping stuff between grids. 

This is already tested in https://github.com/PlasmaControl/DESC/blob/4540398bf5f92d1b24b17938308f88001c7f3cf0/tests/test_equilibrium.py#L160-L175

unalmis (dismissed):

dpanici (dismissed):

PR #968, opened 2024-03-29
vanishing b field

### LINKED ISSUE #958: vanishing B-field after coordinate mapping
Example of issue: https://github.com/PlasmaControl/DESC/pull/854#discussion_r1505121975. There are also some discrepancies between computing things on jitable grids and non jitable grids. Should be resolved or partially resolved by #934 . https://en.wikipedia.org/wiki/Hairy_ball_theorem .

Edit: Source of issue found and described here https://github.com/PlasmaControl/DESC/pull/968#issuecomment-2028451491. (The bug does not exist on master).

unalmis:
Resolves #958 . ~Still figuring out why this happens for this routine I've written.~ Edit: Found the issue

### DIFF
--- a/tests/test_configuration.py
+++ b/tests/test_configuration.py
@@ -7,7 +7,9 @@ import pytest
 
 from desc.backend import put
 from desc.equilibrium import EquilibriaFamily, Equilibrium
+from desc.equilibrium.coords import _field_line_to_desc_coords
 from desc.equilibrium.initial_guess import _initial_guess_surface
+from desc.examples import get
 from desc.geometry import (
     FourierRZCurve,
     FourierRZToroidalSurface,
@@ -449,6 +451,42 @@ class TestInitialGuess:
             eq = Equilibrium(M=surf.M, N=surf.N, surface=surf)
 
         assert eq.is_nested()
+
+    # pytest -s -k test_hairy_ball_theorem tests/test_configuration.py
+    @pytest.mark.unit
+    def test_hairy_ball_theorem(self):
+        """Ensure |B| does not vanish after coordinate mapping."""
+        eq = get("W7-X")
+        rho = np.linspace(0, 1, 5)
+        alpha = np.linspace(0, 2 * np.pi, 2)
+        zeta = np.linspace(0, 4 * np.pi, 10)
+
+        grid_jit, data_jit = _field_line_to_desc_coords(
+            eq, rho, alpha, zeta, jitable=True
+        )
+        data_jit = eq.compute("|B|", grid=grid_jit, data=data_jit)
+        grid_no_jit, data_no_jit = _field_line_to_desc_coords(
+            eq, rho, alpha, zeta, jitable=False
+        )
+        data_no_jit = eq.compute("|B|", grid=grid_no_jit, data=data_no_jit)
+        print(data_jit["|B|"])
+        print()
+        print(data_no_jit["|B|"])
+        # I think there are two separate bugs.
+        # Bug 1:
+        # Regardless of any logic issues with the coordinate mapping etc., I think
+        # these two arrays should be approximately equal. However, the jitted |B|
+        # has varies between 0 and 10^11, while the no jit |B| varies from 0 to 3.
+        np.testing.assert_allclose(
+            data_jit["|B|"],
+            data_no_jit["|B|"],
+            err_msg="jitting issue? grid weights issue?"
+            " mixing grids with different weights to compute things?",
+        )
+        # Bug 2: vanishing |B|.
+        assert np.all(
+            ~np.isclose(data_no_jit["|B|"], 0)
+        ), "logic issue in field_line_to_desc_coords?"
 
 
 class TestGetSurfaces:
--- a/desc/backend.py
+++ b/desc/backend.py
@@ -73,6 +73,7 @@ if use_jax:  # noqa: C901 - FIXME: simplify this, define globally and then assig
     vmap = jax.vmap
     scan = jax.lax.scan
     bincount = jnp.bincount
+    repeat = jnp.repeat
     from jax import custom_jvp
     from jax.experimental.ode import odeint
     from jax.scipy.linalg import block_diag, cho_factor, cho_solve, qr, solve_triangular
@@ -629,6 +630,13 @@ else:  # pragma: no cover
     def bincount(x, weights=None, minlength=None, length=None):
         """Same as np.bincount but with a dummy parameter to match jnp.bincount API."""
         return np.bincount(x, weights, minlength)
+
+    def repeat(a, repeats, axis=None, total_repeat_length=None):
+        """A numpy implementation of jnp.repeat."""
+        out = np.repeat(a, repeats, axis)
+        if total_repeat_length is not None:
+            out = out[:total_repeat_length]
+        return out
 
     def custom_jvp(fun, *args, **kwargs):
         """Dummy function for custom_jvp without JAX."""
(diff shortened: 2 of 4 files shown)


tests/test_configuration.py
```diff
@@ -450,6 +452,42 @@ def test_guess_from_coordinate_mapping_no_sym(self):
...
+        grid_no_jit, data_no_jit = _field_line_to_desc_coords(
+            eq, rho, alpha, zeta, jitable=False
+        )
+        data_no_jit = eq.compute("|B|", grid=grid_no_jit, data=data_no_jit)
+        print(data_jit["|B|"])
+        print()
+        print(data_no_jit["|B|"])
+        # I think there are two separate bugs.
+        # Bug 1:
+        # Regardless of any logic issues with the coordinate mapping etc., I think
```
f0uriest, line 477: When `jitable=True` we don't scale the weights to account for duplicates, that could be causing some issue? Though if those weights are being used somewhere for a surface average that likely indicates something else is wrong?

unalmis:
Currently on this PR (which is `master` + #934 + code to run these tests), I have isolated the vanishing B issue and the jitable grid discrepancy issue to the same bug.

## Issues
**1.**  ~Vanishing $|B|$ after coordinate mapping.~ $\to |B|$  is computed incorrectly when mapping a multiple surface field line grid to desc coordinates. This causes it to vanish at some nodes throughout the volume.
**2.** Resolved by reduction to previous issue:  A $10^{11}$ magnitude discrepancy when computing $|B|$  between jitable and not-jitable grids.

## Notes:

1. Using the test added in this PR, when the coordinate mapping is done from a field line grid on a single surface, I don't find any issues.
    - The only exception is that `jitable` grids compute $|B|$ to be 0 when we specify `rho=np.array([0])`.
2. Using the test added in this PR, I find that there are nodes where $|B|$  vanishes throughout the volume when we pass in a field line grid that has a surface $\rho < 10^{-8}$ and any other surfaces, e.g. $|B|$  vanishes if `rho = np.array([1e-9, 0.5])`. $|B|$ does **not** vanish when the grid has all surfaces $\rho > 10^{-8}$, e.g. `rho = np.linspace(start=1e-8, stop=1, num=10)`, but the larger `start - stop` is the less accurate the computation of $|B|$ is. I have determined accuracy by eyeballing the values; but the statement also holds if we take the value of $|B|$ transformed from a straight field line grid with a single rho surface as truth. **So it seems some computation that should be local to a surface is getting mixed together.**
    - Above, I mentioned the values of $|B|$ are computed incorrectly if multiple surfaces are given. The error in these values grows nonlinearly as the closest surface approaches the magnetic axis. **So `rho=np.array([1e-12, 0.2])` will give higher error than `rho=np.array([1e-2, 0.2])`.**
3. The computed value of $|B|$ at the magnetic axis for `non-jitable` grids is correct regardless of how many surfaces are specified. From this, it seems the data used to compute the limit of $|B|$ are unaffected by the bug. **Also contrary to the pattern in 2, `rho=np.array([1e-12, 0.2])` gives higher error than `rho=np.array([0, 0.2])`**.
4. Combining the above, issue 2 reduces to issue 1 as follows. When a `jitable` grid is specified to be constructed with a node at the magnetic axis, we ignore this and silently move the node to $\rho = 10^{-12}$. [Here is the relevant code](https://github.com/PlasmaControl/DESC/blob/e670bcc3dcd44a1bb38a3806551a35d15d7fd935/desc/grid.py#L420).
Comparing the computed $|B|$ on a `jitable` and `non-jitable` grid, the `non-jitable` grid now has correct values at the axis and reasonable values elsewhere, while the `jitable` grid has incorrect values at the axis and unreasonable values almost everywhere.


## Recommendations
if I recall correctly, we remove the axis for two reasons
- Because we don't want to compute expensive derivatives at each step of an iterative routine just to evaluate quantities at the magnetic axis. By removing the axis from the grid, the compute functions will not automatically compute these derivatives.
  - I would prefer to add a warning whenever we change something specified by the user. At worst it educates them of our expected inputs, and at best it saves them time debugging.   
- Because we can't compute the indices of the axis under `jit`, we just remove the axis to make the empty `grid.axis` array consistent.
  - Overwriting `nan` resulting form the indeterminate form of $|B|$ at axis with `0` may be equally as invalid. Consider changing `grid.axis` to a boolean mask to allow computing limits.
  
The fact that the bug appears only when there are multiple surfaces suggests to me there is still a bug with grid indexing etc. perhaps in my code here:
https://github.com/PlasmaControl/DESC/blob/f33d9f93ce12f7bde30990278f162d31b374ad50/desc/equilibrium/coords.py#L482
The fact that the error seems to grows nonlinearly as the closest surface approaches the magnetic axis might mean there's some numerical issue.

tests/test_configuration.py
unalmis: I don't think it has to do with duplicate nodes. I commented my debugging results below

f0uriest:
Out of curiosity, what was the issue?

Also wrt 
>Because we can't compute the indices of the axis under jit, we just remove the axis to make the empty grid.axis array consistent.

>    Overwriting nan resulting form the indeterminate form of at axis with 0 may be equally as invalid. Consider changing grid.axis to a boolean mask to allow computing limits.

This is indeed the issue for silently shifting rho=0 to rho=small. Making `grid.axis` a boolean mask could alleviate this, however I think in that case it would always compute the extra axis dependencies. Currently we branch based off `grid.axis.size` which is statically known at compile time. However if we make it a mask, then it's a traced value so there wouldn't be a clean way to only do stuff if `all(grid.axis == False)`

unalmis:
> Out of curiosity, what was the issue?
> 
> Also wrt
> 
> > Because we can't compute the indices of the axis under jit, we just remove the axis to make the empty grid.axis array consistent.
> 
> > Overwriting nan resulting form the indeterminate form of at axis with 0 may be equally as invalid. Consider changing grid.axis to a boolean mask to allow computing limits.
> 
> This is indeed the reason for silently shifting rho=0 to rho=small. Making `grid.axis` a boolean mask could alleviate this, however I think in that case it would always compute the extra axis dependencies. Currently we branch based off `grid.axis.size` which is statically known at compile time. However if we make it a mask, then it's a traced value so there wouldn't be a clean way to only do stuff if `all(grid.axis == False)`

Yes that makes sense. And for the same reason it's not possible to add a warning either if the axis is removed.

The issue was my `_meshgrid_expand` function expanded all the flux surface quantities to the full grid incorrectly resulting in an incorrect $|B|$. Separately, the reason why $|B|$ was computed differently between grids constructed with `jitable=True` vs `jitable=False`  was because the nodes were different; `jitable=True` grids shift nodes from $\rho = 0$ to $\rho = 10^{-12}$.

As to why this shift caused a $10^{11}$ order magnitude discrepancy between $|B|$ on these two grids, I don't have a precise answer. By continuity, I would expect the wrong computations for $|B|$ to be close. Since they were not, we can conclude either
1. An axis limit is discontinuous, and it shouldn't be. This implies either the axis limit is wrong or the normal expression is wrong.
2. An axis limit is discontinuous, and it should be. e.g. $e^{\theta}$. Then we could expect the wrong values of $|B|$ to differ from each significantly.
3. Near the axis, i.e. at $\rho = 10^{-12}$, the expressions to compute quantities needed for $|B|$ are more susceptible to numerical round off errors than the expressions for the axis limits.

I ruled out 1 because I doubt it, and we have tests for continuity. 2 sounds reasonable, in which case nothing more needs to be fixed. 3, if true, doesn't seem actionable.

PR #970, opened 2024-03-29
C0 coils with splines

kianorr:
Add discontinuous knots for C0 capabilities.

### DIFF
--- a/desc/compute/_curve.py
+++ b/desc/compute/_curve.py
@@ -645,6 +645,7 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
     params=["X", "Y", "Z", "knots", "rotmat", "shift"],
     transforms={
         "method": [],
+        "intervals": [],
     },
     profiles=[],
     coordinates="s",
@@ -655,30 +656,46 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
 def _x_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
     xq = data["s"]
 
-    Xq = interp1d(
-        xq,
-        params["knots"],
-        params["X"],
-        method=transforms["method"],
-        derivative=0,
-        period=2 * jnp.pi,
-    )
-    Yq = interp1d(
-        xq,
-        params["knots"],
-        params["Y"],
-        method=transforms["method"],
-        derivative=0,
-        period=2 * jnp.pi,
-    )
-    Zq = interp1d(
-        xq,
-        params["knots"],
-        params["Z"],
-        method=transforms["method"],
-        derivative=0,
-        period=2 * jnp.pi,
-    )
+    for interval in transforms["intervals"]:
+
+        query_range = (xq > interval[0]) & (xq < interval[1])
+        interval_range = (params["knots"] > interval[0]) & (
+            params["knots"] < interval[1]
+        )
+
+        xq_in_interval = xq[query_range]
+        knots_in_interval = params["knots"][interval_range]
+        X_in_interval = params["X"][interval_range]
+        Y_in_interval = params["Y"][interval_range]
+        Z_in_interval = params["Z"][interval_range]
+
+        # TODO: hmm
+        period = 2 * jnp.pi if abs(interval[1] - interval[0]) == 2 * jnp.pi else None
+
+        Xq = interp1d(
+            xq_in_interval,
+            knots_in_interval,
+            X_in_interval,
+            method=transforms["method"],
+            derivative=0,
+            period=period,
+        )
+        Yq = interp1d(
+            xq_in_interval,
+            knots_in_interval,
+            Y_in_interval,
+            method=transforms["method"],
+            derivative=0,
+            period=period,
+        )
+        Zq = interp1d(
+            xq_in_interval,
+            knots_in_interval,
+            Z_in_interval,
+            method=transforms["method"],
+            derivative=0,
+            period=period,
+        )
 
     coords = jnp.stack([Xq, Yq, Zq], axis=1)
     coords = (
--- a/desc/geometry/curve.py
+++ b/desc/geometry/curve.py
@@ -675,6 +675,7 @@ class SplineXYZCurve(Curve):
         Y,
         Z,
         knots=None,
+        discontinuous_knots=None,
         method="cubic",
         name="",
     ):
@@ -711,8 +712,25 @@ class SplineXYZCurve(Curve):
             errorif(knots[-1] > 2 * np.pi, ValueError, "knots must lie in [0, 2pi]")
             knots = knots[:-1] if closed_flag else knots
 
+        if discontinuous_knots is None:
+            intervals = [[0, 2 * np.pi]]
+        else:
+            find_nearest = lambda array, val: (np.abs(array - val)).argmin()
+            discontinuous_indices = [
+                find_nearest(knots, val) for val in discontinuous_knots
+            ]
+            discontinuous_knots = knots[discontinuous_indices]
+
+            intervals = [
+                [discontinuous_knots[i - 1], 2 * np.pi]
+                # TODO: this doesn't seem like the best way to do this
+                if i == 0 else [discontinuous_knots[i - 1], discontinuous_knots[i]]
+                for i in range(len(discontinuous_knots))
+            ]
+
         self._knots = knots
         self.method = method
+        self.intervals = intervals
 
     @optimizable_parameter
     @property
@@ -857,54 +875,3 @@ class SplineXYZCurve(Curve):
         return SplineXYZCurve(
             coords[:, 0], coords[:, 1], coords[:, 2], knots, method, name
         )
-
-
-class DiscontinuousCurve(Curve):
-    def __init__(
-        self,
-        X,
-        Y,
-        Z,
-        knots=None,
-        method="cubic",
-        name="",
-    ):
-        pass
-
-    @optimizable_parameter
-    @property
-    def X(self):
-        """Coordinates for X."""
-        return self._X
-
-    @X.setter
-    def X(self, new):
-        if len(new) == len(self.knots):
-            self._X = jnp.asarray(new)
-        else:
-            raise ValueError(
-                "X should have the same size as the knots, "
-                + f"got {len(new)} X values for {len(self.knots)} knots"
-            )
-
-    @optimizable_parameter
-    @property
-    def Y(self):
-        """Coordinates for Y."""
-        return self._Y
-
-    @Y.setter
-    def Y(self, new):
-        if len(new) == len(self.knots):
-            self._Y = jnp.asarray(new)
-        else:
-            raise ValueError(
-                "Y should have the same size as the knots, "
-                + f"got {len(new)} Y values for {len(self.knots)} knots"
-            )
-
-    @optimizable_parameter
-    @property
-    def Z(self):
-        """Coordinates for Z."""
-        return self._Z
--- a/desc/geometry/curve.py
+++ b/desc/geometry/curve.py
@@ -857,3 +857,8 @@ class SplineXYZCurve(Curve):
         return SplineXYZCurve(
             coords[:, 0], coords[:, 1], coords[:, 2], knots, method, name
         )
+
+
+class DiscontinuousCurve:
+    def __init__(self):
+        pass
--- a/desc/geometry/curve.py
+++ b/desc/geometry/curve.py
@@ -859,6 +859,52 @@ class SplineXYZCurve(Curve):
         )
 
 
-class DiscontinuousCurve:
-    def __init__(self):
+class DiscontinuousCurve(Curve):
+    def __init__(
+        self,
+        X,
+        Y,
+        Z,
+        knots=None,
+        method="cubic",
+        name="",
+    ):
         pass
+
+    @optimizable_parameter
+    @property
+    def X(self):
+        """Coordinates for X."""
+        return self._X
+
+    @X.setter
+    def X(self, new):
+        if len(new) == len(self.knots):
+            self._X = jnp.asarray(new)
+        else:
+            raise ValueError(
+                "X should have the same size as the knots, "
+                + f"got {len(new)} X values for {len(self.knots)} knots"
+            )
+
+    @optimizable_parameter
+    @property
+    def Y(self):
+        """Coordinates for Y."""
+        return self._Y
+
+    @Y.setter
+    def Y(self, new):
+        if len(new) == len(self.knots):
+            self._Y = jnp.asarray(new)
+        else:
+            raise ValueError(
+                "Y should have the same size as the knots, "
+                + f"got {len(new)} Y values for {len(self.knots)} knots"
+            )
+
+    @optimizable_parameter
+    @property
+    def Z(self):
+        """Coordinates for Z."""
+        return self._Z
--- a/tests/test_curves.py
+++ b/tests/test_curves.py
@@ -761,3 +761,19 @@ class TestSplineXYZCurve:
         c = SplineXYZCurve(X=R * np.cos(phi), Y=R * np.sin(phi), Z=np.zeros_like(phi))
         with pytest.raises(TypeError):
             c.compute("length", grid=np.linspace(0, 1, 10))
+
+    @pytest.mark.unit
+    def test_discontinuous(self):
+        """Test with discontinuous knots."""
+        discontinuous_knots = [0, np.pi]
+        R = 2
+        phi = 2 * np.pi * np.linspace(0, 1, 20, endpoint=True) ** 2
+        d = SplineXYZCurve(
+            X=R * np.cos(phi),
+            Y=R * np.sin(phi),
+            Z=np.zeros_like(phi),
+            knots="arclength",
+            discontinuous_knots=discontinuous_knots,
+        )
+        print(d.compute("x"))
+        assert 1 == 1


desc/geometry/curve.py
```diff
@@ -671,6 +671,7 @@ def __init__(
         knots=None,
         method="cubic",
         name="",
+        discontinuous_knots=None,
```
f0uriest, line 674: instead of knot values probably easier to have the user give the indices where its allowed to break, then you can avoid the `find_nearest` thing below

desc/compute/_curve.py
```diff
@@ -653,34 +654,60 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+    Y = jnp.append(params["Y"], params["Y"][0])
+    Z = jnp.append(params["Z"], params["Z"][0])
+
+    Xq = jnp.zeros(len(xq))
+    Yq = jnp.zeros(len(xq))
+    Zq = jnp.zeros(len(xq))
+
+    # TODO: generalize for when each coil has different intervals
+    for istart, istop in transforms["intervals"]:
+        istop = -1 if istop == 0 else istop
```
f0uriest, line 676: `istop = jnp.where(istop==0, -1, istop)`

desc/compute/_curve.py
```diff
@@ -653,34 +654,60 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+    X = jnp.append(params["X"], params["X"][0])
+    Y = jnp.append(params["Y"], params["Y"][0])
+    Z = jnp.append(params["Z"], params["Z"][0])
+
+    Xq = jnp.zeros(len(xq))
+    Yq = jnp.zeros(len(xq))
+    Zq = jnp.zeros(len(xq))
+
+    # TODO: generalize for when each coil has different intervals
+    for istart, istop in transforms["intervals"]:
```
f0uriest, line 675: would be good to use a jax `fori_loop` here

desc/compute/_curve.py
```diff
@@ -653,34 +654,60 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+            method=transforms["method"],
+            derivative=0,
+        )
+
+        Xq = jnp.where((xq >= knots[istart]) & (xq <= knots[istop]), Xq_temp, Xq)
+        Yq = jnp.where((xq >= knots[istart]) & (xq <= knots[istop]), Yq_temp, Yq)
+        Zq = jnp.where((xq >= knots[istart]) & (xq <= knots[istop]), Zq_temp, Zq)
 
-    coords = jnp.stack([Xq, Yq, Zq], axis=1)
+    coords = jnp.stack(jnp.stack([Xq, Yq, Zq]), axis=1)
```
f0uriest, line 710: why the double stack here?

desc/geometry/curve.py
```diff
@@ -705,8 +706,24 @@ def __init__(
             errorif(knots[-1] > 2 * np.pi, ValueError, "knots must lie in [0, 2pi]")
             knots = knots[:-1] if closed_flag else knots
 
+        if discontinuous_knots is None:
+            interval_indices = [[0, -1]]
```
f0uriest, line 710: might be better to default to an empty array, and then in the compute function do something like 

```python
if len(interval_indices):
    # do the piecewise stuff
else:
    # just do regular interpolation over whole interval with period=2pi
```

desc/compute/_curve.py
kianorr: Accident

desc/geometry/curve.py
kianorr: something like the most recent push?

kianorr:
I think the names for Xq, Yq, Zq should be more general so we it's easier to copy-paste if we refactor something that applies to all of the compute function (although the bigger problem is that there is a bunch of repeated code, but not sure if that's fixable with compute functions).

desc/compute/_curve.py
```diff
@@ -653,34 +654,83 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
-    )
+    is_discontinuous = len(transforms["intervals"][0])
+
+    def get_f_in_interval(arr, knots, istart, istop):
+        arr_temp = jnp.where(knots > knots[istop], arr[istop], arr)
+        arr_temp = jnp.where(knots < knots[istart], arr[istart], arr_temp)
+
+        return arr_temp
+
+    def inner_body(x, y, z, knots, period=None):
```
dpanici, line 669: to reduce duplicate code, you could define this once, outside of any single compute function, with an additional kwarg for `derivative`. Then you can just call `inner_body` for each of the compute fxns with the correct `derivative` order

dpanici: you can also do the same for `get_f_in_interval` and add `derivative` as an argument to that too then you should only need one function call of it per compute function, which should redue a lot of thestuff

desc/coils.py
```diff
@@ -241,7 +241,9 @@ def to_FourierXYZ(self, N=10, grid=None, s=None, name=""):
             self.current, coords, N=N, s=s, basis="xyz", name=name
         )
 
-    def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
+    def to_SplineXYZ(
+        self, knots=None, grid=None, method="cubic", name="", discontinuous_indices=None
+    ):
```
dpanici, line 246: Add discontinuous indices in the docstring and explain it

desc/coils.py
```diff
@@ -628,7 +637,14 @@ def compute_magnetic_field(
...
+        cls,
+        current,
+        coords,
+        knots=None,
+        method="cubic",
+        name="",
+        basis="xyz",
+        discontinuous_indices=None,
     ):
         """Create SplineXYZCoil from coordinate values.
```
dpanici, line 650: Add discontinuous indices in the docstring and explain it

desc/coils.py
```diff
@@ -1329,7 +1346,9 @@ def to_FourierXYZ(self, N=10, grid=None, s=None, name=""):
         coils = [coil.to_FourierXYZ(N, grid, s) for coil in self]
         return self.__class__(*coils, NFP=self.NFP, sym=self.sym, name=name)
 
-    def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
+    def to_SplineXYZ(
+        self, knots=None, grid=None, method="cubic", name="", discontinuous_indices=None
+    ):
         """Convert all coils to SplineXYZCoil.
```
dpanici, line 1352: Add discontinuous indices in the docstring and explain it

desc/compute/_curve.py
```diff
@@ -635,6 +635,88 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
     return data
 
 
+def _splinexyz_helper(f, p_knots, transforms, xq, kwargs, derivative):
+    method = kwargs.get("method", "cubic")
```
dpanici, line 639: some comments to explain what this is doing (I know it is private but it is long) would be good

desc/geometry/curve.py
```diff
@@ -731,6 +731,7 @@ def __init__(
         knots=None,
         method="cubic",
         name="",
```
dpanici, line 733: Add discontinuous indices in the docstring and explain it

desc/compute/_curve.py
```diff
@@ -644,43 +726,24 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
     transforms={
-        "method": [],
+        "intervals": [],
     },
     profiles=[],
     coordinates="s",
     data=["s"],
     parameterization="desc.geometry.curve.SplineXYZCurve",
     basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
+    method="{'cubic', 'linear'}: Interpolation type, Default 'cubic'",
```
dpanici, line 722: I think there are a lot more than just cubic or linear right? there is a list on the SplineXYZCurve docstring. We could add all the options here (or just say refer to SplineXYZCurve for the full list or something), we should be accurate here since this description will end up [in the docs](https://desc-docs.readthedocs.io/en/latest/variables.html#id15)

desc/geometry/curve.py
```diff
@@ -765,8 +766,17 @@ def __init__(
             errorif(knots[-1] > 2 * np.pi, ValueError, "knots must lie in [0, 2pi]")
             knots = knots[:-1] if closed_flag else knots
 
+        if discontinuous_indices is None:
+            interval_indices = [[]]
+        else:
+            interval_indices = [
+                [discontinuous_indices[i - 1], discontinuous_indices[i]]
```
dpanici, line 773: Is a check to ensure that the discontinuous indices are monotonic necessary? what if I put in [0,40, 10, 100, 50]? 

I'd think that would be good to add to avoid some weird curves being made that double back on themselves

desc/geometry/curve.py
```diff
@@ -869,8 +879,62 @@ def method(self, new):
...
     @classmethod
-    def from_values(cls, coords, knots=None, method="cubic", name="", basis="xyz"):
+    def from_values(
+        cls,
+        coords,
+        knots=None,
+        method="cubic",
+        name="",
+        basis="xyz",
+        discontinuous_indices=None,
```
dpanici, line 936: Add this new argument to the docstring

tests/test_curves.py
```diff
@@ -820,3 +820,35 @@ def test_compute_ndarray_error(self):
...
+            method="linear",
+        )
+
+        assert discontinuous.method == "linear"
+        assert continuous.method == "linear"
+
+        discont_torsion = discontinuous.compute("torsion")["torsion"]
+        cont_torsion = continuous.compute("torsion")["torsion"]
+
+        np.testing.assert_allclose(discont_torsion, cont_torsion, rtol=1e-3, atol=1e-3)
```
dpanici, line 854: Since the method is linear, isnt the torsion zero, since the second derivative is 0 and torsion involves the curve's second derivative?

A better test would be length (which should be pretty close if not exactly identical if you are using linear interpolation). I am not sure of a good test with a derivative quantity but you could try using curvature I guess (which for linear might be the same again)

And a test where the inteprolation method is not linear might be good too to actually test the derivatives computation, though I dont have a good one off the top of my head

dpanici (changes requested):
Left some comments, mainly
- update docstrings
- add a check and possible sort (or throw an error if they are not monotonic) for discontinuous indices to ensure they are monotonic
- Improve test

dpanici:
Also update with master so the tests will run!

desc/compute/_curve.py
kianorr: Yeah there's a lot more, just forgot to add them. Okay gotcha

kianorr:
from dev meeting yesterday: might be best to see if the user inputted `int`s (for index input) or `float`s (for arclength input) (because if `knots="arclength"`, then the user has to guess at the indices for `discontinuous_knots`)

desc/compute/_curve.py
kianorr: I just said to look at the SplineXYZCurve docstring, but i can also add this link: https://desc-docs.readthedocs.io/en/latest/_api/geometry/desc.geometry.SplineXYZCurve.html (just not sure how to add without going over character limit)

desc/compute/_curve.py
```diff
@@ -639,6 +639,87 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+def _splinexyz_helper(f, p_knots, transforms, xq, kwargs, derivative):
+    """Used to compute XYZ coordinates for the SplineXYZCurve compute functions.
+
+    Parameters
+    ----------
+    f : list of ndarray
+        X, Y, Z function coords with shape (3, len(p_knots))
+    p_knots : ndarray
+        knots that come from params
+    transforms : dict
```
ddudt, line 651: Missing docstring for `transforms`

desc/compute/_curve.py
```diff
@@ -647,42 +728,25 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
     description="Position vector along curve",
     dim=3,
     params=["X", "Y", "Z", "knots", "rotmat", "shift"],
-    transforms={"method": []},
+    transforms={
+        "intervals": [],
```
ddudt, line 732: I know I'm OCD, but if you delete this unnecessary comma this becomes 1 line instead of 3 and is more readable

ddudt: Same for the similar functions below

desc/geometry/curve.py
```diff
@@ -817,8 +820,21 @@ def __init__(
...
+            ), "Indices must me monotonic."
+            interval_indices = [
+                [discontinuous_indices[i - 1], discontinuous_indices[i]]
+                for i in range(len(discontinuous_indices))
+            ]
+
         self._knots = knots
-        self.method = method
+        self._method = method
+        self.intervals = interval_indices
```
ddudt, line 837: Can you explain to me why you needed to add the new attributes `discontinuous_indices` and `intervals` and how they are actually used? I don't see where they get referenced in these changes.

tests/test_curves.py
```diff
@@ -844,3 +844,48 @@ def test_compute_ndarray_error(self):
...
+            assert discontinuous.method == method
+            assert continuous.method == method
+
+            discont_curvature = discontinuous.compute("curvature")["curvature"]
+            cont_curvature = continuous.compute("curvature")["curvature"]
+
+            return discont_curvature, cont_curvature
+
+        discont_curvature, cont_curvature = test("linear")
+        np.testing.assert_allclose(discont_curvature, cont_curvature, rtol=1e-3)
```
ddudt, line 883: Is it worth testing any other compute quantities besides just curvature?

ddudt (changes requested):

desc/geometry/curve.py
kianorr: The user inputs indices at which they want the curve to be discontinuous (`discontinuous_indices`). These indices are taken and are used to create the separate intervals of the curve (`self.intervals`). e.g. if `discontinuous_indices = [0, 7]`, then `self.intervals = [[0, 7], [7, 0]]`. The intervals are then iterated through in the compute functions  -- via transforms -- such that each interval is interpolated separately.

Is that what you're looking for?

tests/test_curves.py
kianorr: I could add one for length pretty length pretty easily but not sure what I would add for torsion. I guess checking that the torsion is zero for each of them could work?

desc/coils.py
```diff
@@ -267,6 +269,8 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
             - `'catmull-rom'`: C1 cubic centripetal "tension" splines
         name : str
             name for this coil
+        discontinuous_indices : ndarray
+            indices of knots at which the curve is discontinuous (i.e. only C0)
```
dpanici, line 273: Maybe there is a better name for discontinuous indices, something more like "section indices" or something idk. Or you can keep this name. but you should explain more in the docstring what it actually means for the spline, i.e. the spline will not be evaluated as a periodic spline curve with the given knots, but instead as a sequence of splines over the knots with start and endpoint given by discnontinuous_indices[i-1,i] for the ith spline (with 0: disc_index[0] as the first interval and knots[disc_index[-1]], knots[01] as the last interval) evaluated with whatever endpoint conditions it is, (maybe say default endpoint conditions and refer to interpax's docs), and at knots[discontinuous_indices], the spline

desc/geometry/curve.py
```diff
@@ -770,6 +770,8 @@ class SplineXYZCurve(Curve):
 
     name : str
         name for this curve
+    discontinuous_indices : ndarray
```
dpanici, line 773: see above comment, we should explain more I thin

desc/coils.py
ddudt: I agree, I think I was confused about what this PR is doing because of the misleading names. This isn't really a discontinuous C0 spline, it's a regular spline of some order >0 that is only C0 in some intervals. And the whole curve is still continuous (at least C0 everywhere). Right?

kianorr: You're right Daniel and it is confusing so I'll fix it. If you consider the whole coil then it is C0 continuous. The "discontinuous" refers to the fact that it's discontinuous in it's derivatives at those indices. And in between those indices, it is a spline whose interpolation is chosen by the user

dpanici:
define torsion as zero at corners

kianorr:
> define torsion as zero at corners

Was this specifically for the test or should this be defined in the torsion compute function? I realized I don't understand what was meant by this (@dpanici @f0uriest )

desc/coils.py
kianorr: I think I will use `break_indices`. And this is a possible docstring:

"Indices of knots at which the curve breaks and is only continuous in space (and not continuous in it's derivatives). In between each set of break points, there is an unbroken spline whose start and endpoints are given by `break_indices[i-1, i]`, where `i` indicates the ith spline. Each (the ith) spline is interpolated independently of all other unbroken splines (i+1th, i-1th, etc.) and does not consider their query points (knots) when interpolating. The boundary conditions are evaluated using interpax's default () where non-periodicity is assumed."

f0uriest: I'd say 
```
and is only C0 continuous (eg, continuous but with "corners" where the derivative jumps)
```

f0uriest:
> > define torsion as zero at corners
> 
> Was this specifically for the test or should this be defined in the torsion compute function? I realized I don't understand what was meant by this (@dpanici @f0uriest )

I think it can be defined as zero at the breakpoints in the compute function. At a corner the curve is really just 2 lines, which by construction lie in a plane so the torsion is zero.

desc/geometry/curve.py
```diff
@@ -817,8 +827,21 @@ def __init__(
             errorif(knots[-1] > 2 * np.pi, ValueError, "knots must lie in [0, 2pi]")
             knots = knots[:-1] if closed_flag else knots
 
+        if break_indices is None:
+            unbroken_spline_intervals = [[]]
+        else:
+            # check that input is monotonic
+            assert np.all(
```
f0uriest, line 834: minor point: if break_indices contains duplicates than np.unique will be a different length than the original, so the `==` won't broadcast and you won't ever get to the assert, and it might give an uninformative error message. Also, np.unique already sorts the values so you don't need sorted as well.

Might be better as
```python
unique = np.unique(break_indices)
errorif(len(unique) != len(break_indices), ValueError, "break_indices must not contain any duplicated")
errorif(np.any(unique != break_indices), ValueError, "break_indices must be monotonic")
```

desc/geometry/curve.py
```diff
@@ -868,6 +891,7 @@ def Z(self, new):
                 + f"got {len(new)} Z values for {len(self.knots)} knots"
             )
 
+    @optimizable_parameter
```
f0uriest, line 894: i don't think we want the knots themselves to be optimizable?

desc/geometry/curve.py
```diff
@@ -921,8 +945,57 @@ def method(self, new):
                 + f"instead got unknown method {new} "
             )
 
+    def compute(
```
f0uriest, line 948: why does this need to be redefined from the base class?

desc/compute/_curve.py
```diff
@@ -639,6 +639,88 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+    has_break_points = len(transforms["intervals"][0])
+    n_dim = 3
+
+    def inner_body(f, knots, period=None):
+        """Interpolation for spline curves."""
+        fq = [
+            interp1d(
+                xq,
+                knots,
+                f[i],
```
f0uriest, line 676: interpax should be able to handle interpolating multiple funcitons at the same time, so you can pass in f as a nx3 array here instead of looping

desc/compute/_curve.py
```diff
@@ -639,6 +639,88 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+    def body(i, fq):
+        istart, istop = transforms["intervals"][i]
+        # catch end-point
+        istop = jnp.where(istop == 0, -1, istop)
+
+        f_in_interval = [get_interval(f, full_knots, istart, istop) for f in full_f]
+        fq_temp = inner_body(f_in_interval, full_knots, period=None)
+
+        xq_range = (xq >= full_knots[istart]) & (xq <= full_knots[istop])
+        for i in range(n_dim):
```
f0uriest, line 701: i think this loop can also be vectorized

desc/compute/_curve.py
```diff
@@ -639,6 +639,88 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+        arr_temp = jnp.where(knots < knots[istart], arr[istart], arr_temp)
+
+        return arr_temp
+
+    def body(i, fq):
+        istart, istop = transforms["intervals"][i]
+        # catch end-point
+        istop = jnp.where(istop == 0, -1, istop)
+
+        f_in_interval = [get_interval(f, full_knots, istart, istop) for f in full_f]
```
f0uriest, line 697: can also be vectorized

desc/compute/_curve.py
```diff
@@ -639,6 +639,88 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+        for i in range(n_dim):
+            fq = fq.at[i].set(jnp.where(xq_range, fq_temp[i], fq[i]))
+
+        return fq
+
+    if has_break_points:
+        # full_f, knots used in body()
+        # manually add endpoint for broken splines
+        full_knots = jnp.append(p_knots, p_knots[0] + 2 * jnp.pi)
+        full_f = [jnp.append(f[i], f[i][0]) for i in range(3)]
```
f0uriest, line 710: can be vectorized

desc/compute/_curve.py
```diff
@@ -639,6 +639,88 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+        # query points for Xq, Yq, Zq
+        fq = jnp.zeros((n_dim, len(xq)))
+        fq = fori_loop(0, len(transforms["intervals"]), body, fq)
```
f0uriest, line 713: I think this could be replaced by vmap + sum if you want. 
The idea is instead of modifying fq each time, you return fq_temp, but with the points outside the domain replaced with 0. So when vmapped you'd get back an array of shape(3, num_broken_segments, num_knots), where along the broken_segment dimension, there is only ever 1 nonzero element, so you can just do arr.sum(axis=1) to get the expected output.

desc/compute/_curve.py
```diff
@@ -1040,14 +950,43 @@ def _curvature(params, transforms, profiles, data, **kwargs):
...
+    data=["x_s", "x_ss", "x_sss"],
+    parameterization="desc.geometry.curve.SplineXYZCurve",
+    method="Interpolation type, Default 'cubic'. See SplineXYZCurve docs for options.",
+)
+def _torsion_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
+    dxd2x = cross(data["x_s"], data["x_ss"])
+    data["torsion"] = dot(dxd2x, data["x_sss"]) / jnp.linalg.norm(dxd2x, axis=-1) ** 2
+    # set torsion to zero at break points
+    if len(transforms["intervals"][0]):
+        data["torsion"] = data["torsion"].at[transforms["intervals"][:, 1]].set(0.0)
```
f0uriest, line 985: intervals are indices into the knots vector, but torsion is aligned with the s vector. I think what you want is
```
jnp.where((data['torsion'] == params['knots'][transforms['intervals']][:,1:]).any(axis=0), 0.0, data['torsion'])
```

(or something like that)

tests/test_curves.py
```diff
@@ -843,3 +843,57 @@ def test_compute_ndarray_error(self):
...
+                np.testing.assert_allclose(discon_data, cont_data, rtol=1e-3)
+            else:
+                np.testing.assert_allclose(
+                    discon_data[~np.array(break_indices)],
+                    cont_data[~np.array(break_indices)],
+                    rtol=1e-3,
+                )
+            return discon_data, cont_data
+
+        break_indices = [0, 250, 500, 750]
```
f0uriest, line 889: minor thing but can you move this to above the `def test` so that its clear what it is when its referenced inside that function?

tests/test_curves.py
```diff
@@ -843,3 +843,57 @@ def test_compute_ndarray_error(self):
...
+            assert continuous.method == method
+
+            discon_data = discontinuous.compute(data_key)[data_key]
+            cont_data = continuous.compute(data_key)[data_key]
+
+            if include_breaks:
+                np.testing.assert_allclose(discon_data, cont_data, rtol=1e-3)
+            else:
+                np.testing.assert_allclose(
+                    discon_data[~np.array(break_indices)],
```
f0uriest, line 883: see comment above, I'm not sure the indexing is correct here if knots and the query points are different

desc/geometry/curve.py
kianorr: This was the whole problem with getting the `method` parameter to work as a kwarg because it didn't work when we added it to `transforms`. I followed what Daniel did in his rpz basis pr

desc/geometry/curve.py
kianorr: I realized we could also modify the error above this that I didn't see before

```
errorif(
        not np.all(np.diff(sorted(break_indices)) > 0),
        ValueError,
        "supplied break_indices must be monotonically increasing",
)
```
but maybe it's better to separate them like you did so errors are more clear

desc/geometry/curve.py
kianorr: Probably not, I have always been fixing them during optimization anyway. But not completely sure, @dpanici what do you think?

And if I do delete this line, then I should probably move `"knots"` from `params` to `transforms` right?

ddudt: Agreed, don't make this an optimizable parameter

tests/test_curves.py
ddudt: Yes I was confused where this variable was being set

tests/test_curves.py
ddudt: I'm actually confused why this works, because the default grid should have about 4x as many nodes as the curve has knots. It would be best to explicitly pass in the grid that you are computing on, and choose a grid that doesn't align with the spline knots.

desc/geometry/curve.py
```diff
@@ -817,8 +827,21 @@ def __init__(
...
+            ), "Indices must be monotonic."
+            unbroken_spline_intervals = [
+                [break_indices[i - 1], break_indices[i]]
+                for i in range(len(break_indices))
+            ]
+
         self._knots = knots
-        self.method = method
+        self._method = method
+        self.intervals = unbroken_spline_intervals
```
ddudt, line 844: Does this have to be a "public" attribute? (i.e., no leading underscore) Probably best to make it private and add a getter method like we do for everything else

Also why call it `unbroken_spline_intervals` and then rename the variable to `intervals`? Might be better to just use `intervals` the whole time, like we do for knots, method, etc.

desc/coils.py
```diff
@@ -267,6 +269,15 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
             - `'catmull-rom'`: C1 cubic centripetal "tension" splines
         name : str
             name for this coil
+        break_indices : ndarray
```
dpanici, line 272: can you add something about hte default behavior, like "If not given, then the usual periodic spline is assumed" or at the start say "If not None, then the spline will be a piecewise set of splines..." or something so the default behavior is known, and put that the break_indices can be None `"break_indices : ndarray or None"` or something

desc/coils.py
```diff
@@ -554,6 +571,15 @@ class SplineXYZCoil(_Coil, SplineXYZCurve):
 
     name : str
         name for this curve
+    break_indices : ndarray
```
dpanici, line 574: same as above comment

desc/coils.py
```diff
@@ -663,6 +697,15 @@ def from_values(
             name for this curve
         basis : {"rpz", "xyz"}
             basis for input coordinates. Defaults to "xyz"
+        break_indices : ndarray
```
dpanici, line 700: same as above comment

desc/coils.py
```diff
@@ -1357,14 +1403,26 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
             - `'catmull-rom'`: C1 cubic centripetal "tension" splines
         name : str
             name for the new CoilSet
+        break_indices : ndarray
```
dpanici, line 1406: same as above comment

desc/coils.py
```diff
@@ -1562,14 +1627,26 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
```
dpanici, line 1630: same as above comment

desc/compute/_curve.py
kianorr: did you mean to write

```python
data['torsion'] == params['knots'][transforms['intervals']][:,1:]
```?

File: desc/coils.py (hunk 34 above)
Comment (kianorr) [reply desc/coils.py]: Okay let me know if what I have now is good

File: desc/geometry/curve.py (hunk 23 above)
Comment (kianorr) [reply desc/geometry/curve.py]: I deleted that line and put it in `transforms` instead

File: desc/geometry/curve.py (hunk 22 above)
Comment (kianorr) [reply desc/geometry/curve.py]: I made them separate

File: tests/test_curves.py (hunk 32 above)
Comment (kianorr) [reply tests/test_curves.py:884]: How do you recommend I get the indices of break points in `s`? I tried doing
```python
is_break_point = np.isin(s, discontinuous.knots[break_indices])
```
but there is only 1 `True` in `is_break_point` when I was expected 4. Does this mean the break points are not the same value in `s`?

kianorr:
For some reason the test with `"curvature"` and `"cubic"` fails when `break_indices=[0, 500]` but passes when `break_indices=[0,10]`

desc/compute/_curve.py
kianorr: just passing in `f` isn't working for me, it gives me this error: `ValueError: x and f must be arrays of equal length`

f0uriest: i think you just need to transpose it (and maybe also the output).

`f` is shape (3, N), interpax expects it to be (N,3) and will return something (M,3)

kianorr: Ah my bad, thanks!

tests/test_curves.py
kianorr: @ddudt So to be clear, you're saying to pass in a grid to the compute functions that is not `LinearGrid(zeta=knots)`?

ddudt: @kianorr correct

tests/test_curves.py
```diff
@@ -843,3 +843,66 @@ def test_compute_ndarray_error(self):
...
+                np.testing.assert_allclose(
+                    discont_quantity[~is_break_point],
+                    cont_quantity[~is_break_point],
+                    rtol=1e-3,
+                )
+                # we know there is 1 break point in s (0th index)
+                assert len(discont_quantity[~is_break_point]) != len(discont_quantity)
+
+        test("linear", "length")
+        # TODO: not sure why this one doesn't pass, first index is nan for discont
```
kianorr, line 899: the first index for `discont_quantity`  (which is the first breakpoint) is nan, but I don't why

tests/test_curves.py
```diff
@@ -843,3 +843,66 @@ def test_compute_ndarray_error(self):
...
+        test("linear", "curvature", compare_breaks=False)
+        # torsion = 0 for breakpoints, but otherwise torsion = NaN, so can't compare
+        test("linear", "torsion", compare_breaks=False)
+
+        test("cubic", "length")
+        # TODO: test doesn't pass for break_indices = [0, 500]
```
kianorr, line 905: this test with "cubic" and "curvature" doesn't pass with `break_indices=[0, 500]` because values are not within the tolerance but does pass with `break_indices=[0, 10]`

kianorr (commented):
A couple TODOs for me that I have to investigate, but otherwise this PR is ready for another review

tests/test_curves.py
dpanici: expect differences within 2 knots of the break points

`cubic2` is globally coupled so may need looser tols everywhere

tests/test_curves.py
dpanici: might be dividing by zero at the breakpoint, need to add a special case for it if at a breakpt in compute fxn

kianorr: The reasoning for making torsion 0 at breakpoints was because the curve is just 2 lines that lie in the same plane. Is there a reason to make the curvature 0?

tests/test_curves.py
kianorr: This resolved itself, not sure how. But, now the same thing is happening with "cubic" and "length"

kianorr:
My vectorization commits messed some stuff up so I think that explains the error in length. I'll have to look into it

image

image

dpanici:
#1122

tests/test_curves.py
kianorr: I think it is because of this. That test passes with `grid = LinearGrid(N=100)` and not `grid = LinearGrid(N=500)`

f0uriest (commented):
I think there might be an issue with `CoilSet`s if there are spline curves that have different breakpoints. We only vmap over params, not transforms, since its assumed that every coil in a `CoilSet` can use the same transforms. I think this is also an issue if they have different knots after #1099 

Possible fixes:
- move `knots` and `intervals` back to `params` instead of `transforms` (not sure if this will work, what was the reason they were moved in the first place?)
- Add logic to `desc.coils._check_type` to ensure that `knots` and `intervals` are the same for all `SplineXYZCoil` in a `CoilSet`

kianorr:
> I think there might be an issue with `CoilSet`s if there are spline curves that have different breakpoints. We only vmap over params, not transforms, since its assumed that every coil in a `CoilSet` can use the same transforms. I think this is also an issue if they have different knots after #1099 
> 
> Possible fixes:
> - move `knots` and `intervals` back to `params` instead of `transforms` (not sure if this will work, what was the reason they were moved in the first place?)
> - Add logic to `desc.coils._check_type` to ensure that `knots` and `intervals` are the same for all `SplineXYZCoil` in a `CoilSet`

I thought knots and intervals couldn't be in params because they're not optimizable?

desc/coils.py
```diff
@@ -331,6 +333,18 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
+            If supplied, the spline will be a piecewise set of splines.
+            Indices of knots at which the curve breaks and is only C0 continuous (e.g.
+            continuous but with "corners" where the derivative jumps). In between each
+            set of break points, there is an unbroken spline whose start and endpoints
+            are given by `break_indices[i-1, i]`, where `i` indicates the ith spline.
+            Each (the ith) spline is interpolated independently of all other unbroken
+            splines (i+1th, i-1th, etc.) and does not consider their query points
+            (knots) when interpolating. The boundary conditions are evaluated using
+            Interpax's default where non-periodicity is assumed.
+            If None (the default), the spline will be the usual periodic continuous
```
dpanici, line 346: "usual periodic with the continuity dictated by the spline method"

desc/coils.py
```diff
@@ -618,6 +638,17 @@ class SplineXYZCoil(_Coil, SplineXYZCurve):
...
+        If supplied, the spline will be a piecewise set of splines.
+        Indices are of knots at which the curve breaks and is only C0 continuous (e.g.
+        continuous but with "corners" where the derivative jumps). In between each set
+        of break points, there is an unbroken spline whose start and endpoints are
+        given by `break_indices[i-1, i]`, where `i` indicates the ith spline.
+        Each (the ith) spline is interpolated independently of all other unbroken
+        splines (i+1th, i-1th, etc.) and does not consider their query points
+        (knots) when interpolating. The boundary conditions are evaluated using
+        interpax's default where non-periodicity is assumed.
+        If None (the default), the spline will be the usual periodic continuous spline.
```
dpanici, line 651: same as above comment on continuity

desc/coils.py
```diff
@@ -729,6 +768,18 @@ def from_values(
...
+            If supplied, the spline will be a piecewise set of splines.
+            Indices of knots at which the curve breaks and is only C0 continuous (e.g.
+            continuous but with "corners" where the derivative jumps). In between each
+            set of break points, there is an unbroken spline whose start and endpoints
+            are given by `break_indices[i-1, i]`, where `i` indicates the ith spline.
+            Each (the ith) spline is interpolated independently of all other unbroken
+            splines (i+1th, i-1th, etc.) and does not consider their query points
+            (knots) when interpolating. The boundary conditions are evaluated using
+            Interpax's default where non-periodicity is assumed.
+            If None (the default), the spline will be the usual periodic continuous
```
dpanici, line 781: same as above

desc/coils.py
```diff
@@ -1496,14 +1550,29 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
```
dpanici, line 1563: same as above

desc/coils.py
```diff
@@ -1843,14 +1919,29 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
```
dpanici, line 1932: same as above

desc/compute/_curve.py
```diff
@@ -646,6 +646,83 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
+        return fq.T
+
+    def body(interval, full_f, full_knots):
+        """Body used if there are break points."""
+        istart, istop = interval
+        # catch end-point
+        istop = jnp.where(istop == 0, -1, istop)
+
+        # fill f values outside of interval with break point values so that
+        # interpolation only takes into consideration the interval
```
dpanici, line 694: Should note somewhere (maybe in a docstring) that the BC for a given spline is not enforced at the end of each interval but rather at the end of the 0 -> 2pi interval in `s`

desc/compute/_curve.py
```diff
@@ -729,43 +788,26 @@ def _x_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
     description="Position vector along curve, first derivative",
     dim=3,
     params=["X", "Y", "Z", "rotmat"],
-    transforms={"knots": []},
+    transforms={"intervals": [], "knots": []},
     profiles=[],
     coordinates="s",
     data=["s", "phi"],
     parameterization="desc.geometry.curve.SplineXYZCurve",
+    basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
```
dpanici, line 796: remove this per #1027

desc/compute/_curve.py
```diff
@@ -679,40 +756,22 @@ def _center_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
     description="Position vector along curve",
     dim=3,
     params=["X", "Y", "Z", "rotmat", "shift"],
-    transforms={"knots": []},
+    transforms={"intervals": [], "knots": []},
     profiles=[],
     coordinates="s",
     data=["s"],
     parameterization="desc.geometry.curve.SplineXYZCurve",
+    basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
```
dpanici, line 764: remove this per #1027

desc/compute/_curve.py
```diff
@@ -777,43 +819,25 @@ def _x_s_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
     description="Position vector along curve, second derivative",
     dim=3,
     params=["X", "Y", "Z", "rotmat"],
-    transforms={"knots": []},
+    transforms={"intervals": [], "knots": []},
     profiles=[],
     coordinates="s",
     data=["s", "phi"],
     parameterization="desc.geometry.curve.SplineXYZCurve",
+    basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
```
dpanici, line 827: remove this per #1027

desc/compute/_curve.py
```diff
@@ -825,43 +849,26 @@ def _x_ss_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
     description="Position vector along curve, third derivative",
     dim=3,
     params=["X", "Y", "Z", "rotmat"],
-    transforms={"knots": []},
+    transforms={"intervals": [], "knots": []},
     profiles=[],
     coordinates="s",
     data=["s", "phi"],
     parameterization="desc.geometry.curve.SplineXYZCurve",
+    basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
```
dpanici, line 857: remove this per #1027

desc/compute/_curve.py
```diff
@@ -729,43 +788,26 @@ def _x_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
...
-    coords = coords @ params["rotmat"].reshape((3, 3)).T
-    coords = xyz2rpz_vec(coords, phi=data["phi"])
-    data["x_s"] = coords
+    f = [params["X"], params["Y"], params["Z"]]
+
+    coords_s = _splinexyz_helper(f, transforms, xq, kwargs["method"], derivative)
+    coords_s = coords_s @ params["rotmat"].reshape((3, 3)).T
+
+    if kwargs.get("basis", "rpz").lower() == "rpz":
+        coords_s = xyz2rpz_vec(coords_s, phi=data["phi"])
```
dpanici, line 808: always return in `rpz` per #1027

desc/compute/_curve.py
```diff
@@ -777,43 +819,25 @@ def _x_s_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
...
-    coords = coords @ params["rotmat"].reshape((3, 3)).T
-    coords = xyz2rpz_vec(coords, phi=data["phi"])
-    data["x_ss"] = coords
+    f = [params["X"], params["Y"], params["Z"]]
+
+    coords_ss = _splinexyz_helper(f, transforms, xq, kwargs["method"], derivative)
+    coords_ss = coords_ss @ params["rotmat"].reshape((3, 3)).T
+
+    if kwargs.get("basis", "rpz").lower() == "rpz":
+        coords_ss = xyz2rpz_vec(coords_ss, phi=data["phi"])
```
dpanici, line 839: always return in `rpz` per #1027

desc/compute/_curve.py
```diff
@@ -825,43 +849,26 @@ def _x_ss_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
...
-    coords = coords @ params["rotmat"].reshape((3, 3)).T
-    coords = xyz2rpz_vec(coords, phi=data["phi"])
-    data["x_sss"] = coords
+    f = [params["X"], params["Y"], params["Z"]]
+
+    coords_sss = _splinexyz_helper(f, transforms, xq, kwargs["method"], derivative)
+    coords_sss = coords_sss @ params["rotmat"].reshape((3, 3)).T
+
+    if kwargs.get("basis", "rpz").lower() == "rpz":
+        coords_sss = xyz2rpz_vec(coords_sss, phi=data["phi"])
```
dpanici, line 869: always return in `rpz` per #1027

desc/geometry/curve.py
```diff
@@ -777,10 +777,21 @@ class SplineXYZCurve(Curve):
...
+        If supplied, the spline will be a piecewise set of splines.
+        Indices are of knots at which the curve breaks and is only C0 continuous (e.g.
+        continuous but with "corners" where the derivative jumps). In between each set
+        of break points, there is an unbroken spline whose start and endpoints are
+        given by `break_indices[i-1, i]`, where `i` indicates the ith spline.
+        Each (the ith) spline is interpolated independently of all other unbroken
+        splines (i+1th, i-1th, etc.) and does not consider their query points
+        (knots) when interpolating. The boundary conditions are evaluated using
+        interpax's default where non-periodicity is assumed.
+        If None (the default), the spline will be the usual periodic continuous spline.
```
dpanici, line 790: same as above on docstring comment

tests/test_curves.py
```diff
@@ -907,3 +907,64 @@ def test_compute_ndarray_error(self):
...
+                # we know there is 1 break point in s (0th index)
+                assert len(discont_quantity[~is_break_point]) != len(discont_quantity)
+
+        test("linear", "length")
+        test("linear", "curvature")
+        # torsion = 0 for breakpoints, but otherwise torsion = NaN, so can't compare
+        test("linear", "torsion", compare_breaks=False)
+
+        test("cubic", "length")
+        # don't include break points because of interpolator BCs
```
dpanici, line 968: re-think of why this fails, probably not due to the BC but the fact that the BC of the interval (besides at 0 and 2pi) is determined by whatever the spline does when it encounters the flat regions at the edges of the specific interval being considered

dpanici (changes requested):
- some docstring changes requested,
- the `basis` kwarg is deprecated now, the compute functions should always return things in `rpz` (per #1027 )
- since `intervals` is a new attribute of `SplineXYZCurve`/`Coil`, make sure in either `set_up` or wherever it needs to be (If forget exactly, search for like `_rho` or whatever for a `FourierRZTorioidalSurface` or `rotmat` for a `Curve` object, we do this for those), that `intervals` gets set to the correct value so that loading in old `SplineXYZ` objects does not cause issues where `coil.intervals` does not exist (just set to `None` is the thing to do)

kianorr:
image

image

The first grid exactly doubles the xyz coordinates of the query point at pi compared to the second grid.

dpanici:
check if evaluating a single point at the halfway pt yields the same error as the full grid evaluation

kianorr:
Having trouble with `test("cubic", "curvature", compare_breaks=False)` when `endpoint=True` in `grid` so probably have to rethink what's going on with that test like dario said months ago

dpanici:
@kianorr was this ready?

PR #973, opened 2024-04-01
Ensure things that should be ints are ints

f0uriest:
- Adds checking in a bunch of places to ensure stuff like L,M,N,NFP are ints
- Adds int casting to `_set_up` methods for backwards compatibility
- Removes some setter methods, as they're redundant with `change_resolution` and weren't implemented uniformly across classes/attributes.


ddudt (approved):

desc/basis.py
```diff
@@ -9,7 +9,7 @@
 
 from desc.backend import cond, custom_jvp, fori_loop, gammaln, jit, jnp, sign, switch
 from desc.io import IOAble
-from desc.utils import flatten_list
+from desc.utils import check_nonnegint, check_posint, flatten_list
```
rahulgaur104, line 12: Wasn't there something like isposint somewhere? Why not use that?

desc/basis.py
```diff
@@ -252,9 +233,9 @@ class PowerSeries(_Basis):
     """
 
     def __init__(self, L, sym="even"):
-        self.L = L
-        self.M = 0
-        self.N = 0
+        self._L = check_nonnegint(L, "L", False)
+        self._M = 0
```
rahulgaur104, line 237: Why no check_nonnegint for _M and _N?

desc/basis.py
```diff
@@ -647,9 +630,9 @@ class ZernikePolynomial(_Basis):
     """
 
     def __init__(self, L, M, sym=False, spectral_indexing="ansi"):
-        self.L = L
-        self.M = M
-        self.N = 0
+        self._L = check_nonnegint(L, "L", False)
+        self._M = check_nonnegint(M, "M", False)
+        self._N = 0
```
rahulgaur104, line 635: Why no check here?

tests/test_transform.py
```diff
@@ -197,9 +197,9 @@ def test_set_basis(self):
         """Tests the basis setter method."""
         grid = ConcentricGrid(L=4, M=2, N=1)
 
-        basis_20 = FourierZernikeBasis(L=-1, M=2, N=0)
```
dpanici, line 200: do you know why these were -1 before?

dpanici (approved):

tests/test_transform.py
dpanici: oh was it to set it to the default based on the indexing?

desc/basis.py
f0uriest: they're hardcoded to 0 for this class

desc/basis.py
f0uriest: see above

tests/test_transform.py
f0uriest: yeah `L=-1` meant `M if spectral_indexing=="ansi" else 2*M` but it wasn't really used anywhere and its better to be explicit.

PR #976, opened 2024-04-02
add toroidal flux objective

### LINKED ISSUE #954: Objective for `ToroidalFlux`
To avoid coils finding trivial solutions of Bn=0 by setting current (and hence tor flux thru the plasma) to zero.

I've implemented on a [messy side branch](https://github.com/PlasmaControl/DESC/tree/pyREGCOIL_and_CWS) but I dont have a good single commit to cherry pick from it, nor will this branch go into the master ever I think, so worth just looking at it then re-implementing cleanly in a new PR

dpanici:
Resolves #954


dpanici:
@kianorr feel free to change the API of this objective to match what you are doing in #726

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+    bounds : tuple of {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
+        Both bounds must be broadcastable to to Objective.dim_f
+    weight : {float, ndarray}, optional
+        Weighting to apply to the Objective, relative to other Objectives.
+        Must be broadcastable to to Objective.dim_f
+    normalize : bool, optional
+        Whether to compute the error in physical units or non-dimensionalize.
+        Note: has no effect on this objective.
+        FIXME: add normalization for the B part of this objective
```
f0uriest, line 625: to be fixed?

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+    loss_function : {None, 'mean', 'min', 'max'}, optional
+        Loss function to apply to the objective values once computed. This function
+        is called on the raw compute value, before any shifting, scaling, or
+        normalization. Note: has no effect for this objective
+    deriv_mode : {"auto", "fwd", "rev"}
+        Specify how to compute jacobian matrix, either forward mode or reverse mode AD.
+        "auto" selects forward or reverse mode based on the size of the input and output
+        of the objective. Has no effect on self.grad or self.hess which always use
+        reverse mode and forward over reverse mode respectively.
+    source_grid : Grid, optional
```
f0uriest, line 640: isnt this redundant with `external_field_source_grid`

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+        Collocation grid containing the nodes to evaluate the normal magnetic field at
+        plasma geometry at. Defaults to a LinearGrid(L=eq.L_grid, M=eq.M_grid,
+        zeta=jnp.array(0.0), NFP=eq.NFP).
+    external_field : MagneticField, optional
+        MagneticField object containing the external field to consider when
+        minimizing the Bn errors. If None, the external field is assumed to be zero.
+        e.g. this could be a 1/R field representing external TF coils, or
+        it could be set of discrete TF coils so that coil ripple is considered during
+        the optimization of the ``field`` object.
+    external_field_source_grid : Grid, optional
```
f0uriest, line 655: i think for the other objectives we call this `field_grid`

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+    source_grid : Grid, optional
+        Collocation grid containing the nodes to evaluate field source at on
+        the winding surface. (used if e.g. field is a CoilSet or
+        FourierCurrentPotentialField). Defaults to the default for the
+        given field, see the docstring of the field object for the specific default.
+    eval_grid : Grid, optional
+        Collocation grid containing the nodes to evaluate the normal magnetic field at
+        plasma geometry at. Defaults to a LinearGrid(L=eq.L_grid, M=eq.M_grid,
+        zeta=jnp.array(0.0), NFP=eq.NFP).
+    external_field : MagneticField, optional
```
f0uriest, line 649: rename to `field`

f0uriest: and shouldn't be optional

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+        Grid object used to discretize the external field source. Defaults to the
+        default for the given external_field, see the docstring of the field object
+        for the specific default.
+    name : str, optional
+        Name of the objective function.
+    """
+
+    _coordinates = "rtz"
+    _units = "T"
+    _print_value_fmt = "Quadratic Flux: {:10.3e} "
```
f0uriest, line 665: Toroidal flux

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+    external_field_source_grid : Grid, optional
+        Grid object used to discretize the external field source. Defaults to the
+        default for the given external_field, see the docstring of the field object
+        for the specific default.
+    name : str, optional
+        Name of the objective function.
+    """
+
+    _coordinates = "rtz"
+    _units = "T"
```
f0uriest, line 664: T m^2

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        deriv_mode="auto",
+        source_grid=None,
+        eval_grid=None,
+        external_field=None,
+        external_field_source_grid=None,
+        name="toroidal-flux",
+        eq_fixed=False,
```
f0uriest, line 683: I think this should always have eq fixed, similar to QuadraticFlux

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+        ----------
+        use_jit : bool, optional
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        eq = self._eq if self._eq_fixed else self.things[1]
+        field = self.things[0]
+        # if field is different than self._field, update
```
f0uriest, line 720: when would this happen?

desc/objectives/_coils.py
```diff
@@ -597,3 +598,274 @@ def compute(self, params, constants=None):
...
+        warnif(
+            not np.allclose(eval_grid.nodes[:, 2], eval_grid.nodes[0, 2]),
+            UserWarning,
+            "Evaluation grid should be at constant zeta",
+        )
+
+        # ensure vacuum eq, as we don't yet support finite beta
+        pres = np.max(np.abs(eq.compute("p")["p"]))
+        curr = np.max(np.abs(eq.compute("current")["current"]))
+        warnif(
```
f0uriest, line 743: would it be better to error out here, since the results might be meaningless if there is a plasma field?

Also, might be clearer to say that this objective isn't needed at finite beta

desc/objectives/_coils.py
dpanici: This was to allow the inclusion of a field that was not being optimized like a background TF field. But with the optimizable collection stuff this should not be needed anymore, we can just pass FixParameter to a SumMagneticField

desc/objectives/_coils.py
```diff
@@ -597,3 +597,207 @@ def compute(self, params, constants=None):
...
+    """Target the toroidal flux in an equilibrium from a magnetic field.
+
+    This objective is needed when performing stage-two coil optimization on
+    a vacuum equilibrium, to avoid the trivial solution of minimizing Bn
+    by making the coil currents zero. Instead, this objective ensures
+    the coils create the necessary toroidal flux for the equilibrium field.
+
+    Parameters
+    ----------
+    eq : Equilibrium, optional
```
f0uriest, line 612: not optional

desc/objectives/_coils.py
```diff
@@ -597,3 +597,207 @@ def compute(self, params, constants=None):
...
+    loss_function : {None, 'mean', 'min', 'max'}, optional
+        Loss function to apply to the objective values once computed. This function
+        is called on the raw compute value, before any shifting, scaling, or
+        normalization. Note: has no effect for this objective
+    deriv_mode : {"auto", "fwd", "rev"}
+        Specify how to compute jacobian matrix, either forward mode or reverse mode AD.
+        "auto" selects forward or reverse mode based on the size of the input and output
+        of the objective. Has no effect on self.grad or self.hess which always use
+        reverse mode and forward over reverse mode respectively.
+    source_grid : Grid, optional
```
f0uriest, line 641: I think for consistency with the other field objectives this should be `field_grid`?

desc/objectives/_coils.py
```diff
@@ -597,3 +597,207 @@ def compute(self, params, constants=None):
...
+        errorif(
+            curr > 1e-8,
+            UserWarning,
+            f"Current is non-zero (max {curr} A), "
+            + "this objective is unneeded at finite beta.",
+        )
+
+        # eval_grid.num_nodes for quad flux cost,
+        self._dim_f = 1
+        self._equil_data_keys = ["|e_rho x e_theta|", "R", "phi", "Z"]
```
f0uriest, line 737: i think this is unused

desc/objectives/_coils.py
```diff
@@ -597,3 +597,207 @@ def compute(self, params, constants=None):
...
+            UserWarning,
+            "Evaluation grid should be at constant zeta",
+        )
+        if self._normalize:
+            self._normalization = eq.Psi
+
+        # ensure vacuum eq, as is unneeded for finite beta
+        pres = np.max(np.abs(eq.compute("p")["p"]))
+        curr = np.max(np.abs(eq.compute("current")["current"]))
+        errorif(
```
f0uriest, line 722: i know i said erroring might be better here but I'm reconsidering.

Telling whether something is vacuum can be tricky because its always relative. IE pressure of 1e-8 might seem small, but if the whole problem is rescaled than it might be big. Similarly for current (especially since even if the loop integral of B is 0 to machine precision, when we divide by mu0 to get current its usually at least ~1e-8)

desc/objectives/_coils.py
```diff
@@ -597,3 +597,207 @@ def compute(self, params, constants=None):
...
+        """
+        eq = self._eq
+        if self._eval_grid is None:
+            eval_grid = LinearGrid(
+                L=eq.L_grid, M=eq.M_grid, zeta=jnp.array(0.0), NFP=eq.NFP
+            )
+            self._eval_grid = eval_grid
+        eval_grid = self._eval_grid
+
+        warnif(
```
f0uriest, line 711: this one might make more sense as an error since its easy to check without ambiguity, and the meaning of the objective for different zeta in the same grid isnt even defined

desc/objectives/_coils.py
```diff
@@ -597,3 +597,207 @@ def compute(self, params, constants=None):
...
+            Toroidal flux from coils and external field
+
+        """
+        if constants is None:
+            constants = self.constants
+
+        data = constants["equil_data"]
+        plasma_coords = constants["plasma_coords"]
+
+        B = self._field.compute_magnetic_field(
```
f0uriest, line 786: I think `field` and the grids should be in `constants`, not referenced via `self`

tests/test_objective_funs.py
```diff
@@ -1726,14 +1763,34 @@ def test_compute_scalar_resolution_vacuum_boundary_error(self):
...
+    def test_compute_scalar_resolution_toroidal_flux(self):
+        """ToroidalFlux."""
+        ext_field = ToroidalMagneticField(1, 1)
+        eq = get("precise_QA")
+        eq.change_resolution(4, 4, 4, 8, 8, 8)
+
+        obj = ObjectiveFunction(ToroidalFlux(eq, ext_field), use_jit=False)
+        obj.build()
+        g = obj.grad(obj.x(ext_field))
+        assert not np.any(np.isnan(g)), "toroidal flux"
```
f0uriest, line 1782: i think this is duplicated below

desc/objectives/_coils.py
f0uriest: I theory it can avoid the "constant folding" stuff but not sure how much that really helps.

f0uriest (approved):

tests/test_objective_funs.py
```diff
@@ -24,6 +24,8 @@
     FourierCurrentPotentialField,
     OmnigenousField,
     SplineMagneticField,
+    ToroidalMagneticField,
+    VerticalMagneticField,
```
rahulgaur104, line 28: What is a vertical magnetic field? Like in the Z_hat direction?

rahulgaur104 (approved):

tests/test_objective_funs.py
ddudt: yes, a magnetic field with only a B_Z component

desc/objectives/_coils.py
```diff
@@ -597,3 +597,208 @@ def compute(self, params, constants=None):
...
+
+    This objective is needed when performing stage-two coil optimization on
+    a vacuum equilibrium, to avoid the trivial solution of minimizing Bn
+    by making the coil currents zero. Instead, this objective ensures
+    the coils create the necessary toroidal flux for the equilibrium field.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
```
ddudt, line 613: The Equilibrium is assumed fixed and is not optimized, only the field, correct? Please state that here.

desc/objectives/_coils.py
```diff
@@ -597,3 +597,208 @@ def compute(self, params, constants=None):
...
+            self._eval_grid = eval_grid
+        eval_grid = self._eval_grid
+
+        errorif(
+            not np.allclose(eval_grid.nodes[:, 2], eval_grid.nodes[0, 2]),
+            ValueError,
+            "Evaluation grid should be at constant zeta",
+        )
+        if self._normalize:
+            self._normalization = eq.Psi
```
ddudt, line 717: use the `"Psi"` value from `compute_scaling_factors` instead

tests/test_objective_funs.py
```diff
@@ -816,6 +819,39 @@ def test(coil, grid=None):
...
+        grid1 = LinearGrid(L=10, M=10, zeta=np.array(0.0))
+
+        def test(eq, field, correct_value, rtol=1e-14, grid=None):
+            obj = ToroidalFlux(eq=eq, field=field, eval_grid=grid)
+            obj.build(verbose=2)
+            torflux = obj.compute_unscaled(*obj.xs(field))
+            np.testing.assert_allclose(torflux, correct_value, rtol=rtol)
+
+        eq = Equilibrium(iota=PowerSeriesProfile(0))
+        test(eq, VerticalMagneticField(B0=1), 0, grid=grid1)
```
ddudt, line 834: I'm confused by this test. The equilibrium is a purely toroidal field, but you are comparing to a purely vertical field? I would expect the equilibrium to have a flux of Psi=1 (the default) and the field to have Psi=0. So why is the "correct value" here 0?

desc/objectives/_coils.py
```diff
@@ -597,3 +597,208 @@ def compute(self, params, constants=None):
...
+    eval_grid : Grid, optional
+        Collocation grid containing the nodes to evaluate the normal magnetic field at
+        plasma geometry at. Defaults to a LinearGrid(L=eq.L_grid, M=eq.M_grid,
+        zeta=jnp.array(0.0), NFP=eq.NFP).
+    name : str, optional
+        Name of the objective function.
+    """
+
+    _coordinates = "rtz"
+    _units = "T*m^2"
```
ddudt, line 655: The units should be in parentheses. FYI our `FixPsi` objective uses units of `"(Wb)"`, and I vote to keep the notation consistent.

ddudt (changes requested):
Same comment I made about the `QuadraticFlux` objective: 

Since the `ToroidalFlux` objective assumes the Equilibrium is held fixed during the optimization, I think we need to add a check for that. Like if `ToroidalFlux` is in `optimizer.objective`, then we should throw an error if eq is in `optimizer.things`.

desc/objectives/_coils.py
dpanici: It is the exact same `eq.Psi` though, so I did not think there was a need to calculate the rest of the scalings.

tests/test_objective_funs.py
dpanici: I am testing the objective function, which calculates the toroidal flux from the given magnetic field. The given magnetic field here is a pure vertical field, thus it generates zero toroidal flux. This just is checking that the objective correctly calculates the toroidal flux as zero.

f0uriest (approved):

desc/objectives/_coils.py
ddudt: Technically it introduces a tiny bit of overhead, but I would argue it's worth it to follow the same format as all of the other objective normalizations.

ddudt (approved):

PR #977, opened 2024-04-02
Coil-Coil & Plasma-Coil Minimum Distance Objectives

### LINKED ISSUE #898: Objective for coil/coil distance

### LINKED ISSUE #900: Objective for coil/plasma distance

### LINKED ISSUE #947: Objective for coil/vessel distance
For CSX

dpanici:
- Adds the objective `CoilsetMinDistance`, which returns the minimum distance to another coil for each coil in a coilset. Resolves #898 
- Adds the objective `PlasmaCoilsetMinDistance`, which returns the minimum distance to the plasma surface for each coil in a coilset. Resolves #900, #947

Dependencies: 
- #1016 
- #1017 
-  #1018 

TODO: 
- [x] unit tests for `CoilsetMinDistance`
- [x] unit tests for `PlasmaCoilsetMinDistance`
- [x] regression test using both objectives in a coil optimization
- [x] make it work with surface and/or equilibrium to resolve #947


ddudt:
FYI Thomas is already working on this. Maybe we can combine his work with this.

desc/coils.py
```diff
@@ -930,6 +930,58 @@ def body(B, x):
...
+            "x",
+            grid=grid,
+            params=params,
+            basis="xyz",
+            transforms=transforms,
+            **kwargs,
+        )
+        # set some large value for initial separation
+        minSeparation2 = 1e20
+        for whichCoil1 in range(len(self)):
```
f0uriest, line 971: the double loop here will be really slow to compile, probably better to just stack the output data and do a vectorized operation. Assuming theres O(10) coils with O(100) pts per coil, thats only a 1000x1000 matrix which is fine. Also makes it easier to get min distance per coil and do softmin etc.

f0uriest:
I think this is probably a better implementation:

```python

npts = 100
ncoil = 10
pts = jnp.array(np.random.random((ncoil, npts,3)))
# pts is basically the output of coilset.compute("x")


def body(_, i):
    dx = pts[i] - pts
    dist = jnp.linalg.norm(dx, axis=-1)
    # ignore distance to pts within each coil
    mask = jnp.ones(ncoil)
    mask = mask.at[i].set(0)[:,None]
    return None, jnp.min(dist, where=mask, initial=jnp.inf)

min_dist_per_coil = jax.lax.scan(body, None, np.arange(ncoil))[1]

File: desc/objectives/_coils.py [hunk 2]
```diff
@@ -597,3 +598,151 @@ def compute(self, params, constants=None):
...
+            for whichCoil2 in range(len(data)):
+                if whichCoil1 == whichCoil2:
+                    continue
+                coords2 = data[whichCoil2]["x"]
+
+                d = jnp.linalg.norm(
+                    coords1[:, None, :] - coords2[None, :, :],
+                    axis=-1,
+                )
+                mindist = jnp.min(jnp.array([jnp.min(d), mindist]))
```
kianorr, line 744: I don't think I get why we're taking the minimum here. I thought if we set a low target then the distance would be optimized for that so why do we take the minimum?

dpanici:
TODO for me: make sure it caluculates the x's of EVERY coil (not just unique ones) so that it can properly avoid self intersection across the stellarator symmetry reflection plane

desc/objectives/_coils.py
dpanici: in the outer loop. I choose a coil, then in the inner loop we loop thru every other coil in the coilset. 

I want one number per coiil which is the min dist from that coil to the other coils in the coilset. the first min (`jnp.min(d)`) gives me the minimum distance from this coil to the current coil I am comparing against. Then the second `jnp.min` call checks if I am closer to this coil than I was to any of the other coils I've already checked previously. If yes, it updates the mindist for this coil to be the smaller value.

f0uriest:
For dealing with symmetry: I think the easiest option is to have `CoilSet.compute` return data for all the coils (ie via duplication, rotation, flipping etc), including virtual coils, similar to how `CoilSet.compute_magnetic_field` accounts for virtual coils. Then the existing logic should work fine.

desc/objectives/_coils.py
```diff
@@ -238,6 +239,7 @@ def compute(self, params, constants=None):
             params=params,
             transforms=constants["transforms"],
             grid=self._grid,
+            basis=kwargs.get("basis", None),
```
kianorr, line 242: why default to `None` here?

dpanici: I actually made a mistake here, if None gets passed in for basis, we get an error because the function is not equipped to handle it and does not have a default. I might just default to xyz here but not sure yet of which is best.

desc/coils.py
```diff
@@ -833,14 +838,208 @@ def compute(
...
+                        rotated_data = xyz2rpz_vec(
+                            rotated_data,
+                            phi=coords_rpz[:, 1] + i * 2 * np.pi / self.NFP,
+                        )
+                rotated_data_per_coil_this_FP = jnp.split(
+                    rotated_data, jnp.cumsum(jnp.array(data_sizes_one_FP))[:-1]
+                )
+                for k in range(N_coils_per_FP):
+                    data_this_FP[k][key] = rotated_data_per_coil_this_FP[k]
+            data_all_coils += copy.deepcopy(data_this_FP)
```
dpanici, line 1017: change  data_all_coils to a tuple and dont use copy

dpanici:
Just do this for "x" in the objective function

desc/objectives/_coils.py
```diff
@@ -602,11 +603,157 @@ def compute(self, params, constants=None):
...
+        f : array of floats
+            Distance to nearest coil for each coil in the coilset.
+
+        """
+        pts = self._coils.compute_position(source_grid=self._grid)
+
+        def body(i):
+            # FIXME: this is only element-wise distances! we need all permutations
+            dx = pts[i] - pts
+            dist = safenorm(dx, axis=-1)
```
ddudt, line 741: @f0uriest @dpanici I don't think this is actually computing what we want. This is the element-wise distances between the "same" grid points on each coil, i.e., the distances between nodes with the same index. But we also want to include distances between all nodes with other indices in each coil's grid. In other words, if there are 2 coils with 5 nodes along each coil, this will return 5 distances for each coil, but what we really want is 5^2=25 distances for each coil. 

Let me know if someone thinks of a way to fix this.

desc/coils.py
```diff
@@ -854,6 +882,64 @@ def flip(self, *args, **kwargs):
         """Flip the coils across a plane."""
         [coil.flip(*args, **kwargs) for coil in self.coils]
 
+    def compute_position(self, params=None, source_grid=None, **kwargs):
```
ddudt, line 885: I don't think we can easily move this logic into the objective function. We need to be able to compute the objective from an arbitrary `MixedCoilSet`, which could contain a mix of individual coils, coil sets, and other nested mixed coil sets. Only the `CoilSet` can have symmetry, not the `Coil` or `MixedCoilSet` classes. Because of this, it is much easier to have each class compute it's own coil positions. This solution mimics what we already do for `compute_magnetic_field` for each class

desc/objectives/_coils.py
ddudt: I think I fixed this now

dpanici: did you have a test where the old way failed but the new way works? I am unsure if I follow what you are saying

ddudt: This is covered in `test_coil_min_distance` by the last case with a `MixedCoilset`. In all the other cases, the minimum points on each coil all happen to be at the same grid point (like all at the inboard midplane, for example)

desc/coils.py
```diff
@@ -1474,6 +1572,34 @@ def compute(
...
+            Grid used to discretize coils. If an integer, uses that many equally spaced
+            points. Should NOT include endpoint at 2pi.
+
+        Returns
+        -------
+        x : ndarray, shape(len(self),source_grid.num_nodes,3)
+            Coil positions, in [R,phi,Z] or [X,Y,Z] coordinates.
+
+        """
+        # TODO: catch error if coils have different default grids
```
dpanici, line 1592: what is this error? we assume every sub coil is being computed with the same grid?

desc/coils.py
```diff
@@ -854,6 +887,64 @@ def flip(self, *args, **kwargs):
...
+
+        Returns
+        -------
+        x : ndarray, shape(len(self),source_grid.num_nodes,3)
+            Coil positions, in [R,phi,Z] or [X,Y,Z] coordinates.
+
+        """
+        if params is None:
+            params = [get_params("x", coil) for coil in self]
+        basis = kwargs.pop("basis", "xyz")
```
dpanici, line 909: something weird here can happen if `grid` is passed in, where it then is used as a kwarg and we get an error that `grid` was supplied twice to the compute function.

desc/coils.py
ddudt: We assume every coil in the coilset has the same compute grid (aka source grid). This happens when we stack together the big array of distances, which has shape (ncoils, num_nodes, 3), so if some coils have a grid with a different number of nodes then they won't concatenate together properly. 

This is only an issue for the `MixedCoilSet`, which can have different types of coils with different default grids. The objective `CoilsetMinDistance` takes care of this by making sure the same grid is used for every coil, but you can still run into problems if you call `MixedCoilSet.compute_position` directly.

tests/test_coils.py
```diff
@@ -499,17 +499,29 @@ def test_symmetry_position(DummyCoilSet):
...
+    # check that positions of CoilSets are the same
+    x_sym = coilset_sym.compute_position(basis="rpz")
+    x_asym = coilset_asym.compute_position(basis="rpz")
+    x_mixed = coilset_mixed.compute_position(basis="rpz")
+    # the phi is can be negative for the mixed coilset so make it positive to agree
+    # with the others
+    x_mixed = np.array(x_mixed)
+    x_mixed[:, :, 1] = np.where(
+        x_mixed[:, :, 1] < 0, x_mixed[:, :, 1] + 2 * np.pi, x_mixed[:, :, 1]
+    )
```
ddudt: Is this necessary? This was supposed to already be handled by this line: 
https://github.com/PlasmaControl/DESC/blob/dp/coil-coil-dist/desc/coils.py#L940

desc/coils.py
dpanici: should we make this a private function then?

tests/test_coils.py
dpanici: That line is for the CoilSet not the MixedCoilSet

dpanici: the MixedCoilSet does not have such a check

ddudt: But `MixedCoilSet` simply calls `compute_position` for all of its coils/coilsets, so I thought it would be covered already. If not, we should add a similar line so the phi position outputs are always positive.

dpanici: it calls its own compute_position which then calls the compute_position of the indiviudal coils. We could add something inside the compute function itself I suppose

desc/coils.py
dpanici: Don't love this but it works for now and is way less messy than before. Main reason why I dont love is that there are now two different ways to compute the positions for a coilset (compute(x) and compute_position ), even if we make it clear what the differences are and why the latter exists it still would be a point of confusion. If it is not a private function then we should make it clear the use case for it (you want to get all the coil positions possibly accounting for symmetry and NFP, not just unique coils)

f0uriest: I would strongly vote for making this a private method for now, since it's only used in the objective and we plan on making a better API in the future. We could also enforce that `source_grid` is only a single grid, the same for all coils.

desc/objectives/_coils.py
```diff
@@ -602,11 +603,154 @@ def compute(self, params, constants=None):
...
+            Minimum distance to another coil for each coil in the coilset.
+
+        """
+        # include all permutations of grid nodes between coils
+        xyz = self._coilset.compute_position(source_grid=self._grid)
+        pts1 = jnp.tile(xyz, (1, len(self._perms), 1))
+        pts2 = jnp.concatenate([xyz[:, p, :] for p in self._perms], axis=1)
+
+        def body(k):
+            dist = safenorm(pts1[k] - pts2, axis=-1)  # distances between pts
```
f0uriest, line 741: instead of doing all the permutation stuff I think we can just broadcast these differently. Assuming `pts` has shape `(num_coils, num_nodes, 3)`, we want `pts[i][None,:,None] - pts[:,None,:]` so that `dist` has shape `(num_coils, num_nodes, num_nodes)`

desc/coils.py
ddudt: renamed `source_grid` to `grid`, which should resolve this

tests/conftest.py
```diff
@@ -217,6 +227,67 @@ def DummyStellarator(tmpdir_factory):
...
+    tf_coil = FourierPlanarCoil(center=[2, 0, 0], normal=[0, 1, 0], r_n=[1])
+    tf_coilset = CoilSet.linspaced_angular(tf_coil, n=4)
+    vf_coil = FourierRZCoil(R_n=3, Z_n=-1)
+    vf_coilset = CoilSet.linspaced_linear(
+        vf_coil, displacement=[0, 0, 2], n=3, endpoint=True
+    )
+    xyz_coil = FourierXYZCoil()
+    _ = MixedCoilSet((tf_coilset, vf_coilset, xyz_coil))
+
+    # full_coilset.save(output_path)  FIXME: MixedCoilSet save error
```
ddudt, line 286: @f0uriest there is an issue preventing `MixedCoilSet.save` from working. I could use your help with this

f0uriest: What sort of issue?

ddudt: Error from running the above code, including the line that was commented out: 
```
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.
```

f0uriest: Can you post the full traceback?

ddudt: I made an issue for it: https://github.com/PlasmaControl/DESC/issues/1013

dpanici:
I would have preferred keeping this PR just coil-coil distance...

tests/conftest.py
ddudt: This will get resolved by #1016

tests/test_examples.py
```diff
@@ -1294,3 +1297,59 @@ def test_second_stage_optimization():
     np.testing.assert_allclose(field[0].R0, 3.5)  # this value was fixed
     np.testing.assert_allclose(field[0].B0, 1)  # toroidal field (no change)
     np.testing.assert_allclose(field[1].B0, 0, atol=1e-12)  # vertical field (vanishes)
+
+
+@pytest.mark.unit
+def test_coilset_geometry_optimization():
```
ddudt, line 1303: This is supposed to test the two new objectives in a simple second-stage tokamak coil optimization, but it could use improvement. 

The initial coil set is shown below: 
image

The idea is to reduce the radius of the coils (to match the target plasma-coil distance), and to move the coils so they are evenly spaced around the torus (to maximize coil-coil distance). But this is the optimized result: 
image

It would be ideal for this test if we could constrain the coils to remain in a constant $\phi$ plane. Could we add an option to `FourierPlanarCurve` to parameterize its `center` and `normal` in $(R,\phi,Z)$ coordinates instead of $(X,Y,Z)$ coordinates? Then that constraint would be easy to enforce.

ddudt:
This PR is ready except it is waiting on #1016 to finish adding/editing tests.

dpanici:
The test I added is failing at an unexpected place (when eq_fixed is False)...

desc/objectives/_coils.py
```diff
@@ -708,6 +716,362 @@ def compute(self, params, constants=None):
...
+                profiles=constants["eq_profiles"],
+            )
+            plasma_pts = rpz2xyz(jnp.array([data["R"], data["phi"], data["Z"]]).T)
+
+        def body(k):
+            # dist btwn all pts; shape(ncoils,plasma_grid.num_nodes,coil_grid.num_nodes)
+            dist = safenorm(coils_pts[k][None, :, :] - plasma_pts[:, None, :], axis=-1)
+            return jnp.min(dist, initial=jnp.inf)
+
+        min_dist_per_coil = fori_loop(
```
rahulgaur104, line 1066: Can this be vectorized for each coil?

desc/objectives/_coils.py
```diff
@@ -708,6 +716,362 @@ def compute(self, params, constants=None):
...
+                self._eq_data_keys,
+                params=eq_params,
+                transforms=constants["eq_transforms"],
+                profiles=constants["eq_profiles"],
+            )
+            plasma_pts = rpz2xyz(jnp.array([data["R"], data["phi"], data["Z"]]).T)
+
+        def body(k):
+            # dist btwn all pts; shape(ncoils,plasma_grid.num_nodes,coil_grid.num_nodes)
+            dist = safenorm(coils_pts[k][None, :, :] - plasma_pts[:, None, :], axis=-1)
```
rahulgaur104, line 1063: I think this could be useful for umbilic, where I want to find the distance between a curve and surface.

tests/test_examples.py
```diff
@@ -1313,3 +1314,239 @@ def test_optimize_with_fourier_planar_coil():
...
+            ),
+        )
+    )
+    # TODO: use FixCoilCurrent instead
+    # only 2 free optimization variables are coil center position phi and radius r_n
+    constraints = FixParameters(
+        coils, {"current": True, "center": np.array([0, 2]), "normal": True}
+    )
+    optimizer = Optimizer("scipy-trf")
+    [coils_opt_with_eq], _ = optimizer.optimize(
```
rahulgaur104, line 1378: So you don't have to pass the equilibrium object if the equilibrium is fixed?

Also, when the compute function inside objectives reads coil_params, does it read the params in the right order?

desc/objectives/_coils.py
```diff
@@ -708,6 +716,362 @@ def compute(self, params, constants=None):
...
+        -------
+        f : array of floats
+            Minimum distance from coil to surface for each coil in the coilset.
+
+        """
+        if constants is None:
+            constants = self.constants
+
+        # coil pts; shape(ncoils,coils_grid.num_nodes,3)
+        coils_pts = constants["coilset"]._compute_position(
```
rahulgaur104, line 1044: Why not use the grid to create a transform here and then compute the coil_points?
This seems more complicated than writing a transform in def compute()

tests/test_examples.py
dpanici: - Yes, if eq is fixed, then it is not among the optimizable things and does not need to be passed in
- Yea it does, the JAX pytree stuff makes sure of that I think

tests/test_coils.py
```diff
@@ -533,6 +489,66 @@ def test_coilset_convert(self):
...
+    # same coil sets with vs without symmetry
+    eq = get("precise_QH")
+    coilset_sym = load(
+        load_from=str(DummyCoilSet["output_path_sym"]), file_format="hdf5"
+    )
+    coilset_asym = load(
+        load_from=str(DummyCoilSet["output_path_asym"]), file_format="hdf5"
+    )
+
+    # test that both coil sets compute the same field on the plasma surface
```
kianorr, line 534: I'm confused what the actual difference between the coilsets is. If one doesn't have symmetry then why do we expect it produce the same field?

desc/objectives/_coils.py
```diff
@@ -708,6 +716,362 @@ def compute(self, params, constants=None):
...
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        coilset = self.things[0]
+        grid = self._grid or LinearGrid(N=16)
+
+        self._dim_f = coilset.num_coils
+        self._constants = {"coilset": coilset, "grid": grid, "quad_weights": 1.0}
```
kianorr, line 809: Why is `coilset` in constants? Does this prevent the geometry of the coils from changing?

tests/test_coils.py
ddudt: They are physically the same sets of coils. But the "asymmetric" one includes all of the coils explicitly, and the "symmetric" one only has coils in half of a field period and uses virtual coils from the symmetry assumption.

desc/objectives/_coils.py
ddudt: Technically putting the coilset in `self._constants` is equivalent to making a new attribute `self._coilset`. But as a design choice, we try to put everything "constant" needed for the `compute` function in `self._constants` rather than having a lot of different `self._` attributes.

dpanici:
> The test I added is failing at an unexpected place (when eq_fixed is False)...
> 
> It is a JAX error complaining about int I think, but JAX encounters an error when it tries to throw the error.
> 
> Issue is logged with JAX ([google/jax#20397 (comment)](https://github.com/google/jax/issues/20397#issuecomment-2143671178)) and a PR is in that should let us better see the error ([google/jax#21567](https://github.com/google/jax/pull/21567))

the bug on our part was using `compute` instead of `_compute` inside of `_coils.py`, changing that fixes the error (we had some jax-unfriendly checks inside of `compute` that we skip in `_compute`)

desc/objectives/_coils.py
dpanici: `_compute_position` accounts for any possibly stellarator symmetry and NFP, so we need to use that to get the actual positions of every coil in the coilset, not just the unique ones (which is what would happen if you did what you suggest if we have a `CoilSet` with symmetry/NFP).

tests/test_coils.py
kianorr: Hmm okay, so it's not that it's actually asymmetric when created right? It's more so that the optimizer doesn't know about any symmetries within the "asymmetric coilset"?

desc/objectives/_coils.py
```diff
@@ -708,6 +716,362 @@ def compute(self, params, constants=None):
...
+        "auto" selects forward or reverse mode based on the size of the input and output
+        of the objective. Has no effect on self.grad or self.hess which always use
+        reverse mode and forward over reverse mode respectively.
+    plasma_grid : Grid, optional
+        Collocation grid containing the nodes to evaluate plasma geometry at.
+        Defaults to ``LinearGrid(M=eq.M_grid, N=eq.N_grid)``.
+    coil_grid : Grid, optional
+        Collocation grid containing the nodes to evaluate coilset geometry at.
+        Defaults to ``LinearGrid(N=16)``.
+    eq_fixed: bool, optional
```
f0uriest, line 907: do we want an option to fix the coilset? IE sort of like plasma-vessel distance but with discrete curves instead of a surface?

desc/objectives/_coils.py
```diff
@@ -708,6 +716,362 @@ def compute(self, params, constants=None):
...
+        )
+
+        def body(k):
+            # dist btwn all pts; shape(ncoils,num_nodes,num_nodes)
+            dist = safenorm(pts[k][None, :, None] - pts[:, None, :], axis=-1)
+            # exclude distances between points on the same coil
+            mask = jnp.ones(self.dim_f).at[k].set(0)[:, None, None]
+            return jnp.min(dist, where=mask, initial=jnp.inf)
+
+        min_dist_per_coil = fori_loop(
```
f0uriest, line 848: does this need to be a loop or can we do vmap + reduction? It might be good to have both options with a flag like for the free boundary objective? or a batched vmap or something?

desc/objectives/_coils.py
```diff
@@ -708,6 +716,362 @@ def compute(self, params, constants=None):
...
+        pts = constants["coilset"]._compute_position(
+            params=params, grid=constants["grid"]
+        )
+
+        def body(k):
+            # dist btwn all pts; shape(ncoils,num_nodes,num_nodes)
+            dist = safenorm(pts[k][None, :, None] - pts[:, None, :], axis=-1)
+            # exclude distances between points on the same coil
+            mask = jnp.ones(self.dim_f).at[k].set(0)[:, None, None]
+            return jnp.min(dist, where=mask, initial=jnp.inf)
```
f0uriest, line 846: might also want an option for softmin?

desc/objectives/_coils.py
dpanici: make issue

tests/test_coils.py
dpanici: yea exactly

tests/test_examples.py
dpanici: other note on this, if one of the objectives takes one object and another takes the other, you still will pass in both (in the case of say one objective being just the eq and the other being just the coils)

f0uriest (approved):
Clean up PR description before merging

tests/test_examples.py
```diff
@@ -1293,3 +1295,177 @@ def test_optimize_with_fourier_planar_coil():
...
+        coils_opt[0].center,
+        [R0, np.pi / coils_opt.num_coils, 0],
+        rtol=1e-5,
+    )
+    np.testing.assert_allclose(coils_opt[0].normal, [0, 1, 0])  # normal was fixed
+    np.testing.assert_allclose(
+        coils_opt[0].r_n, a + offset, rtol=1e-2
+    )  # check coil radius
+
+    #### optimize surface with fixed coils ####
```
ddudt, line 1426: @dpanici I modified this test to include a linear constraint that the surface cross-section should remain circular, since the coilset is circular and before it was forcing the coils to become elliptical. Also I removed the 4th test you had because it was attempting to do an optimization with both the eq and coils, but it was actually fixing the coils. It would be nice to include a realistic test with both being optimized but I can't think of a simple/easy test case.

desc/objectives/_coils.py
```diff
@@ -708,6 +716,392 @@ def compute(self, params, constants=None):
...
+    NOTE: By default, assumes the plasma boundary is not fixed and its coordinates are
+    computed at every iteration, for example if the equilibrium is changing in a
+    single-stage optimization.
+    If the plasma boundary is fixed, set eq_fixed=True to precompute the last closed
+    flux surface coordinates and improve the efficiency of the calculation.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
```
ddudt, line 872: Should we edit this to mention that it also works for any `Surface` in addition to an `Equilibrium`? That is also missing from the documentation of `PlasmaVesselDistance`

ddudt (commented):
See minor comment about documentation. It would be nice to add another test with both the coils and surface changing during the optimization, but I don't think that's a dealbreaker. Also I would prefer to merge in PR #1030 before this one but it doesn't really matter.

desc/objectives/_coils.py
dpanici: Yes but `PlasmaVesselDistance` only works for an eq and a FourieRRZToroidalSurface. I don't know of a situation where you have two surfaces you care about (maybe blanket and vessel I guess...) but that is not going to be something we change in this PR, we can make an issue for it if needed

tests/test_examples.py
dpanici: The 4th test was just to test the logic when both the eq and the coils were in `things`. I know that the coil `r_n` was fixed, that was intentional. It was to check the logic in the objective function

f0uriest (approved):

kianorr (approved):

tests/test_examples.py
ddudt: That logic is already tested by `test_plasma_coil_min_distance` in `test_objective_functions.py`. We only need to check that `things` and `params_1` vs `params_2` get assigned in the correct order, which only requires testing that the objective can be computed without error and not a full optimization test. As you can see on [Codecov](https://app.codecov.io/github/PlasmaControl/DESC/commit/45d338af61673fe77c3b712177b2da7950870ee6), adding this test back did not change the coverage at all. I'm not trying to be annoying, I'm just trying to prevent us from adding redundant tests.

dpanici: Thank you for catching this, sorry for my hardheadedness

f0uriest (approved):

kianorr (approved):

PR #980, opened 2024-04-02
Fix `update_target` and a few other things

### LINKED ISSUE #979: Linear objective `update_target` method doesn't work when only certain modes are fixed
```python
eq = desc.examples.get("DSHAPE")

con = FixBoundaryR(eq, modes=[[0,0,0]])
con.build()
print(con.target)
# array([3.51])

con.update_target(eq)
print(con.target)
# array([3.51 , 1.   , 0.106, 0.   , 0.   , 0.   , 0.   , 0.   , 0.   ,
#        0.   , 0.   , 0.   , 0.   , 0.   ])
```

The problem seems to be that `update_target` always gets the full array corresponding to `self._target_arg`, rather than just the indices wanted: https://github.com/PlasmaControl/DESC/blob/5801ea8238a45b6f6674cc0d2ebe67d9d598242f/desc/objectives/linear_objectives.py#L37

On a related note, `FixParameter` inherits from `_FixedObjective` so should be able to call `FixParameter.update_target` but it doesn't work because there's no `target_arg` defined.

f0uriest:
- `Objective.update_target` now works correctly for all linear objectives.
- Some better printing from `FixParameter`
- Adds constraints for fixing sheet current terms in fixed boundary mode, to prevent column of zeros in the jacobian
- Adds guard against division by zero for weird edge case in trust region subproblem - when the Jacobian has a column of zeros, one of the singular values is identically zero. The trust region is usually able to handle this, but if the initial guess for the trust region regularization parameter is also zero, then you end up with `0/(0+0) = nan` and then things break.

Resolves #979


desc/objectives/getters.py
```diff
@@ -102,6 +105,9 @@ def get_fixed_axis_constraints(eq, profiles=True, normalize=True):
                 constraints += (
                     con(eq=eq, normalize=normalize, normalize_target=normalize),
                 )
+    for param in ["I", "G", "Phi_mn"]:
+        if np.array(getattr(eq, param, [])).size:
+            constraints += (FixParameter(eq, param),)
```
ddudt, line 110: It would probably be worth creating a new linear objective function like `FixSheetCurrent` or something that fixes these three parameters. Like how we have other fixed objectives that are redundant with `FixedParameter` for common use cases. 

But also you can do this with a single objective function as `FixParameter(eq, ["I", "G", "Phi_mn"])` instead of creating 3 separate constraints. And instead of the for loop with `gettattr`, you should be able to check `if {"I", "G", "Phi_mn"} <= set(eq.optimizable_params):`. If that doesn't work we need to add these sheet current variables to the equilibrium optimizable params.

ddudt: same for other instances below

desc/objectives/linear_objectives.py
```diff
@@ -108,11 +109,15 @@ def __init__(
...
+        name=None,
     ):
         self._target_from_user = target
-        self._params = params
+        self._params = params = setdefault(params, thing.optimizable_params)
         self._indices = indices
+        self._print_value_fmt = (
+            f"Fixed parameter ({self._params}) error: " + "{:10.3e} "
+        )
+        name = setdefault(name, f"Fixed parameter ({self._params})")
```
ddudt, line 120: Maybe lose the extra parentheses? I think this will print like `(["param1", "param2"])`

ddudt (changes requested):

desc/objectives/getters.py
f0uriest: > It would probably be worth creating a new linear objective function like `FixSheetCurrent` or something that fixes these three parameters. Like how we have other fixed objectives that are redundant with `FixedParameter` for common use cases.

I think that can be another PR. If the sheet current terms exist it's because the user explicitly included them which means they're usually not fixed. This is mainly just to handle the case of the fixed boundary solve inside the proximal optimizer, or other cases where you're doing a "fixed boundary" like solve with the output of a free boundary solve.

> But also you can do this with a single objective function as `FixParameter(eq, ["I", "G", "Phi_mn"])` instead of creating 3 separate constraints. And instead of the for loop with `gettattr`, you should be able to check `if {"I", "G", "Phi_mn"} <= set(eq.optimizable_params):`. If that doesn't work we need to add these sheet current variables to the equilibrium optimizable params.

`{"I", "G", "Phi_mn"} <= set(eq.optimizable_params)` doesn't do what we want. They are always listed under `optimiizable_params` but if no sheet current is assigned they're just empty arrays of length 0, in which case we don't need to fix them (similar to how we don't include constraints for profiles that don't exist, even though the attributes are still there and return an empty array)

ddudt (approved):
Approving, but see comment about removing the parentheses in the print statement

dpanici (approved):

PR #982, opened 2024-04-03
hotfix for `CoilSet` with NFP or stellarator symmetry

### LINKED ISSUE #959: Make plotting coils from a symmetric coilset plot all the coils
Add a flag like `plot_all_coils` to `plot_coils` so that we can have the whole coiliset visualized even when the coilset has `sym=True`. Probably easiest to just use `.from_symmetry` to make a new coiilset (containing every coil) to plot

ddudt:
Resolves #959. 

Also preserves `NFP` and `sym` attributes when converting a `CoilSet` to `FourierXYZ` or `SplineXYZ` curve types.


dpanici (changes requested):
`MixedCoilSet` don't have the `NFP` attribute iirc right? that is causing a test to fail, might want to make a separate conversion method for the subclass that avoids passing in `NFP`

desc/plotting.py
```diff
@@ -2253,6 +2254,11 @@ def plot_coils(coils, grid=None, fig=None, return_data=False, **kwargs):
 
     def flatten_coils(coilset):
         if hasattr(coilset, "__len__"):
+            # plot all coils for symmetric coil sets
+            if coilset.NFP > 1 or coilset.sym:
```
f0uriest, line 2258: do we want to leave the option for only plotting unique coils?

ddudt: I think no? We only keep track of the unique coils to simplify the optimization, but in reality there are coils in every field period (since the magnetic field calculation assumes that). And typically when visualizing the coils you are interested in how the coils are spaced relative to each other, which needs the full coil set. So I can't think of a common use case when you would want only the unique coils and think it's best to always plot the "real" coil set

f0uriest: Ok but I think this logic will break if you pass in a list of coils instead of a coilset, since it will have length but no NFP/sym attributes

ddudt: Hmm true, I'll fix that and a add a test

dpanici: I could imagine showing a figure that shows only the unique coiils in a coilset along with the rest of the plasma just to emphasize that it is only a few unique coils repeated, I have seen these before in past coil papers like for NCSX or in the regcoil paper.

dpanici (changes requested):
Adding a kwarg to plot only unique coils can't hurt I think, so I am in favor of adding that

ddudt:
patch coverage will be low but I think we should override that

f0uriest (approved):

dpanici (approved):
slightly uneasy with the lack of tests but I'll allow it

ddudt:
> slightly uneasy with the lack of tests but I'll allow it

This PR is just fixing existing bugs. The only code that isn't covered by tests is `MixedCoilSet.to_XYZ`. I could make a new test for it but I thought we were trying to minimize the number of tests

PR #984, opened 2024-04-04
`TwoPowerProfile`

ddudt:
New class for profiles of the form $f(\rho) = a(1-\rho^b)^c$ where $a$, $b$, $c$ are all floats. This is equivalent to the `"two_power"` profile input format in VMEC.


desc/profiles.py
```diff
@@ -687,6 +692,88 @@ def from_values(cls, x, y, order=6, rcond=None, w=None, sym="auto", name=""):
...
+        """
+        if params is None:
+            params = self.params
+        if (dt != 0) or (dz != 0):
+            return jnp.zeros(grid.num_nodes)
+        a, b, c = params
+        r = grid.nodes[:, 0]
+        if dr == 0:
+            f = a * (1 - r**b) ** c
+        # df/dr = inf at rho = 1 for all dr > 0
```
f0uriest, line 763: isnt this only true if b,c are floats <dr?

what is the use case for b,c being floats rather than ints?

my worry is also that if b<1 than the derivative blows up at r=0 as well.

ddudt: True, depending on the parameter values it may or may not blow up at $\rho=0$ and/or $\rho=1$. 

My personal use case is to use an integer `b` = 1 or 2, and `c` as a float. But if these numbers are optimizable params they all need to be floats. 

Options:

1. Add a check to make sure `b >= 1`. But there isn't much we can do about `c`. 
2. We don't need to merge this into the master branch. But I coded it up anyways so I figured I would share it.

f0uriest: My question is, do they all need to be optimizable? We could just make `b`,`c` constant integers and leave `a` optimizable?

Along a similar vein, VMEC has a bunch of other profiles types that might be worth looking at: https://github.com/jonathanschilling/educational_VMEC/blob/master/src/profile_functions.f

I have some VMEC input files that use the atan and two-lorentz profiles that would be good to try in DESC.

desc/profiles.py
```diff
@@ -687,6 +692,89 @@ def from_values(cls, x, y, order=6, rcond=None, w=None, sym="auto", name=""):
...
+        """
+        if params is None:
+            params = self.params
+        if (dt != 0) or (dz != 0):
+            return jnp.zeros(grid.num_nodes)
+        a, b, c = params
+        r = grid.nodes[:, 0]
+        if dr == 0:
+            f = a * (1 - r**b) ** c
+        # df/dr = inf at rho = 0 if b < 1
```
dpanici, line 763: is it even physical then if b < 1?

dpanici (changes requested):
- add higher derivs
- add to docstring that the params are only good for certain parameter ranges
- add warning in build to warn if the parameters make things be divergent or nonphysical
- check if it overlaps with a VMEC one and if so, then we can say it is similar to that

ddudt:
> * add higher derivs
> * add to docstring that the params are only good for certain parameter ranges
> * add warning in build to warn if the parameters make things be divergent or nonphysical
> * check if it overlaps with a VMEC one and if so, then we can say it is similar to that

- It turns out this is the same as the `"two_power"` profiles in VMEC, so I changed the name accordingly. 
- The `MTanhProfile` is also only implemented up to derivative order `dr=2`. Do we actually need higher order derivatives? 
- I added documentation and warnings for bad parameter values at initialization.

f0uriest (approved):

desc/profiles.py
```diff
@@ -687,6 +693,102 @@ def from_values(cls, x, y, order=6, rcond=None, w=None, sym="auto", name=""):
...
+        super().__init__(name)
+
+        if params is None:
+            params = [0, 1, 1]
+        self._params = np.atleast_1d(params)
+
+        errorif(
+            self._params.size != 3, ValueError, "params must be an array of size 3."
+        )
+        warnif(
```
dpanici, line 728: add why this is occuring (a[1]<1) in the warning string, I know this in the docstring but the warning is what will pop up first when this is encountered

desc/profiles.py
```diff
@@ -687,6 +693,102 @@ def from_values(cls, x, y, order=6, rcond=None, w=None, sym="auto", name=""):
...
+        elif dr == 1:
+            f = r ** (b - 1) * self.compute(grid, params=[-a * b * c, b, c - 1])
+        elif dr == 2:
+            f = (
+                r ** (b - 2)
+                * ((b * c - 1) * r**b - b + 1)
+                * self.compute(grid, params=[a * b * c, b, c - 2])
+            )
+        else:
+            raise NotImplementedError("dr > 2 not implemented for TwoPowerProfile!")
```
dpanici, line 788: add a test for this (is one line)

dpanici (changes requested):
see comments

f0uriest (approved):

dpanici (approved):

PR #985, opened 2024-04-05
Make users pass in spacing/weights to custom grids

### LINKED ISSUE #981: have custom `Grid` throw error if spacing or weights is called when it was not passed in

f0uriest:
Resolves #981


desc/equilibrium/coords.py
```diff
@@ -113,7 +113,9 @@ def periodic(x):
 
     @functools.partial(jit, static_argnums=1)
     def compute(y, basis):
-        grid = Grid(y, sort=False, jitable=True)
+        # need some dummy spacing values bc iota_num/iota_den still computed using FSA
```
f0uriest, line 116: this is kind of a kludge. We precompute iota using a good grid for the FSA, but the dependency structure stills tries to use the custom grid to compute iota_num etc. Not sure of a better option?

desc/grid.py
```diff
@@ -585,12 +611,7 @@ def _create_nodes(self, nodes):
         # This may cause the surface_integrals() function to fail recognizing
         # surfaces outside the interval [0, 2pi] as duplicates. However, most
         # surface integral computations are done with LinearGrid anyway.
-        spacing = (  # make weights sum to 4pi^2
```
unalmis, line 588: The three lines above about surface integrals can be deleted now. (But keep the note to not modulo custom nodes please).

desc/equilibrium/coords.py
unalmis: Have you thought of modifying the compute functions such that the quantity is computed only if grid.spacing exists?

(in general, the drawback of patching things via dummy spacing is that things may silently get computed incorrectly).

dpanici: could we just update the `data` with `iota_num` and `iota_den` etc so it knows it exists and does not try to recompute it?

f0uriest: @unalmis I think we have to return something from the compute function in both cases to prevent infinite recursion in `compute`. We could return nan if grid.spacing doesn't exist? But that might lead to some weird results down the line and be hard to tell where they're coming from.

@dpanici I think that might work. If we know ahead of time what keys to override we can create dummy values just for those, but again could be a source of weird bugs

dpanici: Maybe worth a try and then see if it works?

f0uriest:
@rahulgaur104 or @unalmis can you check the failing test ``tests/test_equilibrium.py::test_shifted_circle_geometry``?

It looks like you're doing `eq.compute(..., override_grid=False)` but generally that should be set to `True`

rahulgaur104:
test_shifted_circle is meant to be deleted after we approve the bounce-average PR because all the tests in test_shifted_circle have essentially been moved to test_bounce_average_driifts

dpanici:
@unalmis did you accidentally commit [2c20bb5](https://github.com/PlasmaControl/DESC/commit/2c20bb519dde2ef3cb8e832df0da8000f01a6009)? I am confused it says it is removing something that is done in this PR

desc/grid.py
```diff
@@ -515,14 +516,32 @@ class Grid(_Grid):
...
         self._node_pattern = "custom"
-        self._nodes, self._spacing = self._create_nodes(nodes)
+        self._nodes = self._create_nodes(nodes)
+        if spacing is not None:
+            spacing = (
+                jnp.atleast_2d(jnp.asarray(spacing))
+                .reshape(self.nodes.shape)
+                .astype(float)
+            )
+            self._spacing = spacing
```
unalmis, line 535: maybe just `self._spacing = jnp.asarray(spacing).astype(float)` with an assert statement for the shape

desc/grid.py
```diff
@@ -515,14 +516,32 @@ class Grid(_Grid):
...
+        if weights is None and spacing is not None:
+            self._weights = self._spacing.prod(axis=1)
```
unalmis, line 537: it seems like you're making an effort to ensure jax arrays. maybe because it's useful for [memory](https://github.com/PlasmaControl/DESC/blob/90d1ecb2599cb1127f09c838073e679a4dafbb40/desc/compute/utils.py#L924). in that case this should be jnp.prod(self.spacing, axis=1)

unalmis (approved):

dpanici (approved):

desc/grid.py
f0uriest: at this point `self._spacing` will already be a jax array so I think they should be the same.

PR #986, opened 2024-04-05
Add method for generating surface from shape parameters

### LINKED ISSUE #542: Create surface with specified shape parameters
Similar to the model we have for approximately QI surfaces, it would be neat to have a simple option for creating a surface based on simple scalar shape parameters:

- [ ] Major radius
- [ ] minor radius / aspect ratio
- [ ] elongation
- [ ] triangularity
- [ ] "twist" (something with n=1)?
- [ ] torsion

f0uriest:
Basically uses a Miller-like parameterization for the cross section, plus an axis with possible eccentricity and torsion, then a twist to get the cross section to rotate. 

Resolves #542 

Some code for playing around with it:
```python


import numpy as np
from ipywidgets import interact

@interact(major_radius=(1., 10.),
          aspect_ratio=(1., 10.),
          elongation=(-0.5,2.5), 
          triangularity=(-2.,2.),
          squareness=(-2.,2.),
          eccentricity=(0.,1.), 
          torsion=(-1.,1.), 
          twist=(-2., 2.), 
          NFP=(1,5),
         )
def plasma_XS(major_radius, 
              aspect_ratio, 
              elongation, 
              triangularity, 
              squareness, 
              eccentricity, 
              torsion, 
              twist, 
              NFP
             ):

    surf = FourierRZToroidalSurface.from_shape_parameters(
        major_radius,
        aspect_ratio,
        elongation,
        triangularity,
        squareness,
        eccentricity,
        torsion,
        twist,
        NFP,
        sym=True,
    )
    # uncomment below for 3d plot
    # fig = plot_3d(surf, "R")
    # fig.show()
    
    fig, ax = plt.subplots()
    plot_boundary(surf, ax=ax, phi=4)
    ax.set_xlabel('R')
    ax.set_ylabel('Z')
    plt.axis("equal")
```


desc/geometry/surface.py
```diff
@@ -519,6 +519,83 @@ def from_values(
...
+        NFP=1,
+        sym=True,
+    ):
+        """Create a surface using a generalized Miller parameterization.
+
+        Parameters
+        ----------
+        major_radius : float > 0
+            Average major radius. Constant term in the R coordinate.
+        minor_radius : float > 0
```
f0uriest, line 542: is it better to use aspect ratio here?

dpanici: instead of minor radius? yea maybe, they are basically the same but aspect ratio is something that would actually be scanned over and something people think of more than minor radius

dpanici:
image
I think this works pretty well right now, this looks remniscent of your stable optimized QA in your constrained optimization paper

I do notice though that at $\phi * NFP / 2\pi = 0.5$, the cross section seems to always be the same if `sym=True`, I think something in the toroidal variation enforces that if we have stellarator symmetry (probably bc of it being just a rigid rotation so it makes ellipse-like toroidal asymmetries). That is fine, just was pointing it out

dpanici (changes requested):
change minor radius to aspect ratio, but I think rest is fine. Only way I can think to alleviate the torsion thing is to add another parameter like torsions/torsion period per field period or something that is an integer that multiplies the inside of the torsion term

f0uriest:
So right now the "twist" isn't really a rigid rotation, it's just sort of some arbitrary N=1 perturbation applied in the same way to R and Z, so the cross section is always the same at zeta=0 and zeta=pi. We could make it a rigid rotation and make it a discrete integer, like how many times the cross section rotates per field period? that might be more intuitive and would give different cross sections at zeta=0,pi

rahulgaur104:
https://github.com/PlasmaControl/DESC/issues/542#issuecomment-1971612379

f0uriest:
OK changed `minor_radius` -> `aspect_ratio` and made `twist` more like a rigid rotation when it is an integer. For non-integer values it's more like rotation + deformation but seems to vary pretty smoothly. Only catch is that twist=0.5 gives zero volume, but twist=0.4 or 0.6 seem reasonable (makes it sort of sausage-like in addition to the rotation)

rahulgaur104 (approved):

dpanici (approved):

PR #988, opened 2024-04-05
Add documentation for default grid and target/bounds for objectives

### LINKED ISSUE #975: Define Default Grids and Targets for Objective Functions in docstring

f0uriest:
Resolves #975


desc/objectives/_free_boundary.py
```diff
@@ -68,8 +69,9 @@ class VacuumBoundaryError(_Objective):
         reverse mode and forward over reverse mode respectively.
     grid : Grid, optional
         Collocation grid containing the nodes to evaluate error at. Should be at rho=1.
+        Defaults to ``Linear(M=eq.M_grid, N=eq.N_grid)``
```
dpanici, line 72: *LinearGrid

desc/objectives/linear_objectives.py
```diff
@@ -594,10 +596,11 @@ class FixBoundaryR(_FixedObjective):
         Equilibrium that will be optimized to satisfy the Objective.
     target : {float, ndarray}, optional
         Target value(s) of the objective. Only used if bounds is None.
-        Must be broadcastable to Objective.dim_f.
+        Must be broadcastable to Objective.dim_f. Defaults to ``target=eq.Rb_lmn``.
     bounds : tuple of {float, ndarray}, optional
         Lower and upper bounds on the objective. Overrides target.
-        Both bounds must be broadcastable to to Objective.dim_f
+        Both bounds must be broadcastable to to Objective.dim_f.
+        Defaults to ``target=eq.Rb_lmn``.
```
dpanici, line 603: does it make sense to have this Defaults to sentence twice? These objectives default to using target not bounds

dpanici: nevermind this is fine

dpanici (approved):

desc/objectives/_coils.py
```diff
@@ -255,10 +255,11 @@ class CoilLength(_CoilObjective):
     target : float, ndarray, optional
         Target value(s) of the objective. Only used if bounds is None.
         Must be broadcastable to Objective.dim_f. If array, it has to
-        be flattened according to the number of inputs.
+        be flattened according to the number of inputs. Defaults to ``target=2*np.pi``.
```
rahulgaur104, line 258: Shouldn't this default depend on the diameter of the coil. What if the minor radius of the last surface is 10 ?

desc/objectives/_geometry.py
```diff
@@ -527,10 +536,11 @@ class PlasmaVesselDistance(_Objective):
         Bounding surface to penalize distance to.
     target : {float, ndarray}, optional
         Target value(s) of the objective. Only used if bounds is None.
-        Must be broadcastable to Objective.dim_f.
+        Must be broadcastable to Objective.dim_f. Defaults to ``bounds=(1,np.inf)``.
```
rahulgaur104, line 539: Same as my previous questions. Can we make defaults smarter?

rahulgaur104 (approved):

desc/objectives/_coils.py
f0uriest: We could maybe try to make better default values, but it should be in another PR. This one is just about documenting what is currently there.

desc/objectives/_geometry.py
f0uriest: This one is fairly sensible I think, it's just saying we want the vessel to be at least 1m away from the plasma. There will always be a tradeoff between how complicated the default is vs how useful it is. In 99% of cases, the user should be setting a target/bounds themselves anyways.

PR #990, opened 2024-04-05
Add class for user defined magnetic field

### LINKED ISSUE #938: Add ability for user defined ``MagneticField``
Basically allow a user to pass in some JAX transformable function that evaluates B at a point in space, and wrap it in a ``MagneticField`` class to work with the rest of the optimization stuff.

f0uriest:
Resolves #938


dpanici (approved):

desc/magnetic_fields/_core.py
```diff
@@ -540,6 +540,75 @@ def save_mgrid(
...
+
+    params : ndarray, optional
+        Default values for parameters. Defaults to an empty array.
+
+    """
+
+    def __init__(self, fun, params=None):
+        errorif(not callable(fun), ValueError, "fun must be callable")
+        self._params = setdefault(params, jnp.array([]))
+        self._fun = fun
```
ddudt, line 566: Would it be worth also checking that that `fun(coords, params)` returns an array of the expected shape for some dummy coordinates? So that the user doesn't get a mysterious error when they try to do `MagneticFieldFromUser.compute_magnetic_field(coords)`

dpanici (approved):

dpanici (approved):

rahulgaur104 (approved):
Please try to get the coverage up before merging.

PR #991, opened 2024-04-08
Change coilset loading to load in a nested coilset if coil groups are present in MAKEGRID file

dpanici:
- If multiple coil groups are present in a MAKEGRID coils file, will load in a nested coilset, with each sub coilset being named for the coil group.


desc/coils.py
```diff
@@ -1123,43 +1162,54 @@ def from_makegrid_coilfile(cls, coil_file, method="cubic"):
...
-        return cls(*coils)
+        # if it is a single group, then only return one coilset, not a
+        # nested CoilSet
+        groupinds = list(coils.keys())
+        class_init = cls
+        try:
+            if len(groupinds) == 1:
+                return class_init(*coils[groupinds[0]], name=groupnames[0])
+            else:  # return a nested coilset, containing one coilset per coilgroup
+                return MixedCoilSet(
```
f0uriest, line 1195: I'm not sure this is the best way. Since it's a class method, it should always return an instance of the class that called it, while this may sometimes return a `MixedCoilSet` even if called like `CoilSet.from_mgrid_file`

I think what we want is something more like 
- if called from `CoilSet`, try to return a single coilset with all the coils, ignoring groups. If we can't due to different number of knots etc, throw an error and tell user to try using MixedCoilSet
- If called from `MixedCoilSet`, break into groups and return a `MixedCoilSet` of where each subgroup is a `CoilSet` if possible, falling back to `MixedCoilSet` for the subgroups if necessary.
- May also want a an option to ignore groups when loading `MixedCoilSet`? Not as sure about that though.

dpanici: I think that `CoilSet` works if each of the individual CoilSets are `CoilSet` classes though.

A flatten option would be easy, to just return as a single coilset objet

dpanici: I can do this, but I don't love that the user might receive an error when they go to load a CoilSet, we could make this a separate method instead of a class method?

desc/coils.py
```diff
@@ -1613,3 +1587,163 @@ def insert(self, i, new_item):
...
+            # try making the coilgroups use a CoilSet
+            cset = cls(
+                *[
+                    CoilSet(*coils[groupind], name=groupname)
+                    for groupname, groupind in zip(groupnames, groupinds)
+                ]
+            )
+        except ValueError as e:  # can't load as a CoilSet if any of the coils have
+            # different length of knots, so load as MixedCoilSet instead
+            print(
```
f0uriest, line 1735: I feel like this should either be a warning or silent, not just a print statement. I think just silent would be fine, since the docstring says what happens in each case.

desc/coils.py
f0uriest: As it is right now there are coilsets that can't be loaded with `CoilSet.from_mgrid_file` so I don't think it's really an issue. As long as the error says why it happened and how to fix it (using `MixedCoilSet`) i think its fine

f0uriest (dismissed):

dpanici:
change so that it attempts to make CoilSet separately for each coiil group

f0uriest (approved):

kianorr:
What are coil groups? And do MAKEGRID files just contain information about coils?

f0uriest:
> What are coil groups? And do MAKEGRID files just contain information about coils?

a "coil group" is usually a set of coils that are all on the same circuit so share the same current, though there isn't a precise definition. Sometimes it also refers to the "unique" coils over one or 1/2 period

there are really two types of things called "makegrid files". The one referenced here is usually something like `coils.w7x` or `coils.ncsx`, and its a text file output by a coil design code that contains a list of points along each coil. These are read in by MAKEGRID and used to evaluate the magnetic field due to the coils on a grid of points in R,phi,Z which can then be used for free boundary VMEC, field line tracing, etc. This data is then saved in a binary format (usually netcdf), which is the other thing commonly called a "makegrid file".

ddudt (approved):

PR #994, opened 2024-04-11
Adding a piecewise omnigenity objective

### LINKED ISSUE #949: Implementing piecewise omnigenous target magnetic field as described by Velasco
Since the form of the target |B| doesn't depend on the Cary-Shasharina transformation and can be directly specified in Boozer coordinates, on a cursory glance, seems easy to implement.

rahulgaur104:
This PR add the piecewise omnigenity objective as described by Velasco during the Simons meeting.

This will allows us to do much faster an memory efficient optimizations while maintaining approximate omnigenity and maybe use reverse-mode differentiation.

The expression for the magnetic field is given in a screenshot of Velasco's Simons slide.
Screenshot from 2024-04-10 21-27-22

resolves #949


rahulgaur104:
When I try to import the PiecewiseOmnigenity objective, I get the following error:

```
DESC version 0.11.1+555.g6474d1e7b,using JAX backend, jax version=0.4.24, jaxlib version=0.4.24, dtype=float64
Using device: CPU, with 23.07 GB available memory
  File "/home/rgaur/pwO_import_test.py", line 2, in <module>
    from desc.objectives import PiecewiseOmnigenity
  File "/home/rgaur/DESC/desc/objectives/__init__.py", line 3, in <module>
    from ._bootstrap import BootstrapRedlConsistency
  File "/home/rgaur/DESC/desc/objectives/_bootstrap.py", line 6, in <module>
    from desc.compute import get_profiles, get_transforms
  File "/home/rgaur/DESC/desc/compute/__init__.py", line 31, in <module>
    from . import (
  File "/home/rgaur/DESC/desc/compute/_omnigenity.py", line 507, in <module>
    def _B_piecewise_omni(params, transforms, profiles, data, **kwargs):
  File "/home/rgaur/DESC/desc/compute/data_index.py", line 170, in _decorator
    raise ValueError(
ValueError: Can't register function with unknown parameterization: desc.magnetic_fields._core.PiecewiseOmnigenousField

```

f0uriest:
> When I try to import the PiecewiseOmnigenity objective, I get the following error:
> 
> ```
> DESC version 0.11.1+555.g6474d1e7b,using JAX backend, jax version=0.4.24, jaxlib version=0.4.24, dtype=float64
> Using device: CPU, with 23.07 GB available memory
>   File "/home/rgaur/pwO_import_test.py", line 2, in <module>
>     from desc.objectives import PiecewiseOmnigenity
>   File "/home/rgaur/DESC/desc/objectives/__init__.py", line 3, in <module>
>     from ._bootstrap import BootstrapRedlConsistency
>   File "/home/rgaur/DESC/desc/objectives/_bootstrap.py", line 6, in <module>
>     from desc.compute import get_profiles, get_transforms
>   File "/home/rgaur/DESC/desc/compute/__init__.py", line 31, in <module>
>     from . import (
>   File "/home/rgaur/DESC/desc/compute/_omnigenity.py", line 507, in <module>
>     def _B_piecewise_omni(params, transforms, profiles, data, **kwargs):
>   File "/home/rgaur/DESC/desc/compute/data_index.py", line 170, in _decorator
>     raise ValueError(
> ValueError: Can't register function with unknown parameterization: desc.magnetic_fields._core.PiecewiseOmnigenousField
> ```

You need to add the new class to `desc.compute.data_index._class_inheritance`

desc/compute/_omnigenity.py
```diff
@@ -634,6 +634,54 @@ def _interp(x, B):
...
+    dim=1,
+    params=["B_min", "B_max", "zeta_C", "theta_C", "t_1", "t_2", "w_2", "iota0", "NFP"],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="rtz",
+    data=[],  # Potential error, we want eq |B|
+    parameterization="desc.magnetic_fields._core.PiecewiseOmnigenousField",
+    parametrization=[],
+)
+def _B_piecewise_omni(params, transforms, profiles, data, **kwargs):
```
dpanici, line 658: have it take thetaB etc as data, add a compute param for it for this field, and pass it in with data in the objective using the eq-computed thetab etc

desc/compute/_omnigenity.py
```diff
@@ -634,6 +634,54 @@ def _interp(x, B):
...
+    theta_B = transforms["grid"].nodes[:, 1]
+    zeta_B = transforms["grid"].nodes[:, 2]
+
+    zeta_C = params["zeta_C"]
+    theta_C = params["theta_C"]
+    t_1 = params["t_1"]
+    t_2 = params["t_2"]
+    w_2 = params["w_2"]
+    iota0 = params["iota0"]
```
rahulgaur104, line 667: test comment. exponent should be asdfadfa

desc/compute/_omnigenity.py
rahulgaur104: I don't think that's possible because then data would have to be a part of the PiecewiseOmnigenity class.

desc/compute/_omnigenity.py
```diff
@@ -916,6 +916,101 @@ def _interp(x, B):
...
+                    zeta_pp - jnp.pi,
+                    zeta_pm - jnp.pi,
+                    theta_pp - jnp.pi,
+                    theta_pm - jnp.pi,
+                ],
+                axis=0,
+            )
+        )
+        / jnp.pi
+    )
```
victorfpalb, line 1010: There are some typos w.r.t the PDF J.L. Velasco sent us. The corrected version should be the following:

```
    Q = (
        jnp.max(
            jnp.stack(
                [
                    NFP * zeta_pp - jnp.pi,
                    NFP * zeta_pm - jnp.pi,
                    theta_pp - jnp.pi,
                    theta_pm + jnp.pi,
                ],
                axis=0,
            )
        )
        / jnp.pi
    )
```

rahulgaur104: It would be great if you c
an you make these changes using github?

desc/compute/_omnigenity.py
```diff
@@ -916,6 +916,101 @@ def _interp(x, B):
...
+    dim=1,
+    params=["B_min", "B_max", "zeta_C", "theta_C", "t_1", "t_2", "w_2", "iota0"],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="rtz",
+    data=[],  # Potential error, we want eq |B|
+    parameterization="desc.magnetic_fields._core.PiecewiseOmnigenousField",
+)
+def _B_piecewise_omni(params, transforms, profiles, data, **kwargs):
+    theta_B = transforms["grid"].nodes[:, 1]
```
dpanici, line 940: pass in as kwargs to this fxn instead of making a transform

desc/objectives/_omnigenity.py
```diff
@@ -849,6 +848,314 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+                profiles={},
+                helicity=constants["helicity"],
+                iota=jnp.mean(eq_data["iota"]),
+            )
+        else:
+            field_data = compute_fun(
+                "desc.magnetic_fields._core.PiecewiseOmnigenousField",
+                self._field_data_keys,
+                params=field_params,
+                transforms=field_transforms,
```
dpanici, line 1139: `={"grid":grid}`

desc/objectives/_omnigenity.py
```diff
@@ -849,6 +848,344 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+            )
+
+        Ntheta = constants["Ntheta_B"]
+
+        # Rolling ensures max(B) occurs in the corners
+        B_pwO = jnp.roll(field_data["|B|_pwO"], Ntheta / 2)
+        Q_pwO = field_data["Q_pwO"]
+
+        # ReLU operation
+        Q_pwO = (Q_pwO - 0.05) * (Q_pwO >= 0.05)
```
victorfpalb, line 1179: Just a small remark regarding this definition I think it is worth to mention.

 For a good pwO field, $Q \in [-1, 0]$ with $Q \sim 0$ being problematic. Therefore, we defined the upperbound of $Q = -0.05$ (i.e. 5% of margin) in order to be in the safe zone. So the definition of the previous line should be changed to the following to be consistent.

`Q_pwO = (Q_pwO + 0.05) * (Q_pwO >= -0.05)`

Therefore, `Q_pwO` is always positive for any $Q \geq -0.05$ and zero otherwise.

victorfpalb (changes requested):

desc/objectives/_omnigenity.py
```diff
@@ -849,6 +848,344 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+            "iota": jnp.unique(iota_data["iota"]),
+            "iota_r": jnp.unique(iota_data["iota_r"]),
+            "shear": jnp.unique(iota_data["shear"]),
+        }
+
+        ### Passing a Grid of Boozer coordinate values because for the same
+        # DESC grid changing the eq changes the Boozer grid
+        #### TODO: Have to account for NFP
+        field_grid = Grid(
+            jnp.array([eq_data["rho"], eq_data["theta_B"], eq_data["zeta_B"]]).T,
```
victorfpalb, line 1140: TODO fixed: take into account for NFP (that must come from equilibrium). For that the `field_grid` should be including the `NFP` like,

`        field_grid = Grid(
            jnp.array([eq_data["rho"], eq_data["theta_B"], eq_data["zeta_B"]]),
            NFP=constants["eq_transforms"]["grid"].NFP,
            jitable=True,
        )`

desc/objectives/_omnigenity.py
```diff
@@ -849,6 +848,344 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+
+        Ntheta = constants["Ntheta_B"]
+
+        # Rolling ensures max(B) occurs in the corners
+        B_pwO = jnp.roll(field_data["|B|_pwO"], Ntheta / 2)
+        Q_pwO = field_data["Q_pwO"]
+
+        # ReLU operation
+        Q_pwO = (Q_pwO - 0.05) * (Q_pwO >= 0.05)
+
```
victorfpalb, line 1180: ### Put penalty to cases in which `B_max - B_min < threshold`

After trying some optimization, the optimizer usually went to cases in which `B_max = B_min`. Therefore I thought that as a first approach, defining some penalty similar to the one used for `Q_pwO` is the solution. The changes where the following, 

 ```
        # Penalty for cases in which B_max = B_min
        Q_B = field_params["B_max"] - field_params["B_min"]
        
        Q_B = (1.2 - Q_B) * (Q_B <= 1.2)
        
        # Calculate pwO error

        pwO_error = (eq_data["|B|"] - B_pwO) + constants["overlap_penalty"] * (Q_pwO + Q_B)
```

With this, we force $B_{max} - B_{min} \geq value$, where the value can change depending the surface we are optimizing. 

For now, I defined manually the 1.2 value, but in a future this should be automatized so it takes the maximum difference of $|B|$ in a given surface. 

As I said before, this is preliminary and idk if its too restrictive for the optimizer. Pls @rahulgaur104 tell me if this makes sense to you.

**Review by victorfpalb — CHANGES_REQUESTED** [review]

File: desc/objectives/_omnigenity.py (hunk 8 above)
Comment (rahulgaur104) [reply desc/objectives/_omnigenity.py:1180]: A better objective would just be the mirror ratio. You want a mirror ratio = 0< (B_max-B_min)/(B_max + B_min) < 0.5 on the boundary. It's already a compute function and you can implement it using GenericObjective.

File: desc/objectives/_omnigenity.py (hunk 7 above)
Comment (rahulgaur104) [reply desc/objectives/_omnigenity.py:1140]: Thanks for figuring this out, Victor!

File: desc/objectives/_omnigenity.py (hunk 6 above)
Comment (rahulgaur104) [reply desc/objectives/_omnigenity.py:1179]: Ok, let me do that. I thought the sign was the other way around.

File: desc/objectives/_omnigenity.py (hunk 7 above)
Comment (victorfpalb) [reply desc/objectives/_omnigenity.py:1137]: image

Okay, so the grid seems that it returns the 3D grid with ($\rho, \theta_B, \zeta_B$). 

Anyways, I find it weird that its taking negative values for $\zeta_B$ coming directly from the equilibrium data. I guess this is something that should be fixed. 

Can you help me with this? Because I dont know exactly how to fix this or solve this because I dont know if the $|B|$ from the equilibrium is being calculated with the same grid. So either this is not a problem, or both of them are wrong, right?

Comment (rahulgaur104) [reply desc/objectives/_omnigenity.py:1137]: What is the max value of zeta_B that you obtain?
Is it less than 1.57? By how much. If it's less by ~ 0.012. Then it is possible that the Boozer transformation output can be fixed by doing jnp.mod(zeta_B, 2*jnp.pi/NFP).

Comment (rahulgaur104) [reply desc/objectives/_omnigenity.py:1137]: Another way to ensure that would be to plot zeta_B vs theta_B.

Comment (rahulgaur104) [reply desc/objectives/_omnigenity.py:1137]: Wait, the easiest thing to try is to do sym=True in the LinearGrid. If that gives you sensible values, then it's the stellarator symmetry.

File: desc/magnetic_fields/_core.py [hunk 9]
```diff
@@ -3024,3 +3024,194 @@ def helicity(self, helicity):
             and (int(helicity[1]) == helicity[1])
         )
         self._helicity = helicity
+
+
+class PiecewiseOmnigenousField(Optimizable, IOAble):
+    """A magnetic field with piecewise omnigenity.
+
+    Uses parameterization from Velasco et al.[2]
```
f0uriest, line 3032: reference should be `[1]`

desc/magnetic_fields/_core.py
```diff
@@ -3024,3 +3024,194 @@ def helicity(self, helicity):
...
+        modes in `x_basis`. If not supplied, `x_lmn` defaults to zero for all modes.
+
+    Notes
+    -----
+    Doesn't conform to MagneticField API, as it only knows about :math:`|B|` in
+    computational coordinates, not vector B in lab coordinates.
+
+    References
+    ----------
+    .. [1] Velasco, Jose L., et al. "Piecewise Omnigenity"
```
f0uriest, line 3051: is there a doi or arxiv number?

desc/magnetic_fields/_core.py
```diff
@@ -3024,3 +3024,194 @@ def helicity(self, helicity):
...
+    Parameters
+    ----------
+    NFP : int
+        Number of field periods.
+    helicity : tuple, optional
+        Type of pseudo-symmetry (M, N). Default = toroidal contours (1, 0).
+    params0 : ndarray, optional
+        Omnigenity parameters describing h(ρ,η,α). The coefficients correspond to the
+        modes in `x_basis`. If not supplied, `x_lmn` defaults to zero for all modes.
+
```
f0uriest, line 3043: these don't match the actual init signature

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    profiles=[],
+    coordinates="rtz",
+    data=[],
+    parameterization="desc.magnetic_fields._core.PiecewiseOmnigenousField",
+)
+def _B_piecewise_omni(params, transforms, profiles, data, **kwargs):
+    # RG:How does this objective change if iota < 0?
+    nsurfs = len(params["B_max"])
+    p = kwargs.get("p", 10)
+    iota0 = kwargs.get("iota", 0.6 * jnp.ones((nsurfs,)))[:, None]
```
f0uriest, line 942: i think we don't want a default here, it's safer to error if iota isn't supplied

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="rtz",
+    data=[],
+    parameterization="desc.magnetic_fields._core.PiecewiseOmnigenousField",
+)
+def _B_piecewise_omni(params, transforms, profiles, data, **kwargs):
+    # RG:How does this objective change if iota < 0?
+    nsurfs = len(params["B_max"])
+    p = kwargs.get("p", 10)
```
f0uriest, line 941: should add docs for these kwargs to the `register_compute_fun`

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    iota0 = kwargs.get("iota", 0.6 * jnp.ones((nsurfs,)))[:, None]
+
+    # RG: (theta_B, zeta_B) grid shape must be the same for all flux surfaces
```
f0uriest, line 944: could use `assert transforms['grid'].is_meshgrid` to ensure this

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    nodes = transforms["grid"].nodes
+    sort_idx = jnp.argsort(nodes[:, 0])
+    sorted_data = nodes[sort_idx]
+
+    N = jnp.shape(nodes)[0] // nsurfs
+    reshaped_data = sorted_data.reshape(nsurfs, N, 3)
+
+    theta_B = reshaped_data[:, :, 1]  # (nsurfs, N)
+    zeta_B = reshaped_data[:, :, 2]
```
f0uriest, line 953: I think it would be clearer/safer to use `grid.meshgrid_reshape` for this stuff.

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    theta_C = params["theta_C"][:, None]
+    t_1 = params["t_1"][:, None]
+    t_2 = params["t_2"][:, None]
+    w_1 = jnp.pi / NFP * (1 - t_1 * t_2) / (1 + t_2 / iota0)
+
+    w_2 = (
+        jnp.pi**2
+        * (1 - t_1 * t_2)
+        / (jnp.pi - (iota0 + t_2) * ((jnp.pi - NFP * w_1) / (iota0 + 1 / t_1)))
+    )  # Formula of w_2 to ensure Delta = 0. Enforces 0 BS current
```
f0uriest, line 968: does pwO always have zero bootstrap? is there a toroidal or helical version that doesn't?

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    theta_B = reshaped_data[:, :, 1]  # (nsurfs, N)
+    zeta_B = reshaped_data[:, :, 2]
+
+    # NFP can't be a parameter. Must come from equilibrium
+    NFP = transforms["grid"].NFP
+
+    zeta_C = params["zeta_C"][:, None]
+    theta_C = params["theta_C"][:, None]
+    t_1 = params["t_1"][:, None]
+    t_2 = params["t_2"][:, None]
```
f0uriest, line 961: im a bit confused by the radial mapping. eg, zeta_C is an array of size=num_surfs right? But what if the grid has a different number of surfaces? or same number of surfaces but with different spacing? I feel like there needs to be some sort of interpolation here?

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+
+    return data
+
+
+@register_compute_fun(
+    name="Delta_BS",
+    label="|\\Delta_{BS}|",
+    units="~",
+    units_long="None",
+    description="Delta proxi for zero pwO Bootstrap current",
```
f0uriest, line 1050: *proxy

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    zeta_B = reshaped_data[:, :, 2]
+
+    # NFP can't be a parameter. Must come from equilibrium
+    NFP = transforms["grid"].NFP
+
+    zeta_C = params["zeta_C"][:, None]
+    theta_C = params["theta_C"][:, None]
+    t_1 = params["t_1"][:, None]
+    t_2 = params["t_2"][:, None]
+    w_1 = jnp.pi / NFP * (1 - t_1 * t_2) / (1 + t_2 / iota0)
```
f0uriest, line 962: in veloasco's original paper w1 and w2 were parameters, where do these formulas come from? Does this assume anything else about the field?

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+
+    return data
+
+
+@register_compute_fun(
+    name="Q_pwO",
+    label="|\\Q_{pwO}|",
+    units="~",
+    units_long="None",
+    description="Self-overlap of the target field",
```
f0uriest, line 993: can you explain this a bit more? I don't remember this being defined in the paper

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    )  # Formula of w_2 to ensure Delta = 0
+
+    A1 = jnp.abs((4 * w_2 * (w_1 - jnp.pi / NFP)) / (1 - t_1 * t_2))
+    A2 = jnp.abs((4 * jnp.pi**2 / NFP) - (4 * w_2 * jnp.pi) / (NFP * (1 - t_1 * t_2)))
+
+    # RG: Need to check if averaging is correct
+    B_pwO_squared_averaged = surface_averages(
+        transforms["grid"],
+        (data["|B|_pwO"] ** 2).flatten(),
+    )[0]
```
f0uriest, line 1085: is this only for a single surface?

desc/magnetic_fields/_core.py
```diff
@@ -3024,3 +3024,194 @@ def helicity(self, helicity):
...
+    ):
+        """Compute the quantity given by name on grid.
+
+        Parameters
+        ----------
+        names : str or array-like of str
+            Name(s) of the quantity(s) to compute.
+        grid : Grid, optional
+            Grid of coordinates to evaluate at. The grid nodes are given in the usual
+            (ρ,θ,ζ) coordinates, but θ is mapped to η and ζ is mapped to α.
```
f0uriest, line 3101: i think the mapping stuff is leftover from the old class. in pwO the grid is assumed to be in boozer coordinates right?

desc/magnetic_fields/_core.py
```diff
@@ -3024,3 +3024,194 @@ def helicity(self, helicity):
...
+            Defaults to a linearly space grid on the rho=1 surface.
+        params : dict of ndarray
+            Parameters from the equilibrium, such as R_lmn, Z_lmn, i_l, p_l, etc
```
f0uriest, line 3104: Parameters from the field such as B_max, t1 etc.

desc/magnetic_fields/_core.py
```diff
@@ -3024,3 +3024,194 @@ def helicity(self, helicity):
...
+            Defaults to attributes of self.
+        transforms : dict of Transform
+            Transforms for R, Z, lambda, etc. Default is to build from grid
+        profiles : dict of Profile
```
f0uriest, line 3108: this is unused. We should keep it for a consistent API but change the description to note that

desc/objectives/_omnigenity.py
```diff
@@ -847,6 +846,379 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+        this should also be set to True.
+    loss_function : {None, 'mean', 'min', 'max'}, optional
+        Loss function to apply to the objective values once computed. This loss function
+        is called on the raw compute value, before any shifting, scaling, or
+        normalization.
+    deriv_mode : {"auto", "fwd", "rev"}
+        Specify how to compute jacobian matrix, either forward mode or reverse mode AD.
+        "auto" selects forward or reverse mode based on the size of the input and output
+        of the objective. Has no effect on self.grad or self.hess which always use
+        reverse mode and forward over reverse mode respectively.
```
f0uriest, line 884: these are handled by `collect_docs` so can be removed here

desc/objectives/_omnigenity.py
```diff
@@ -847,6 +846,379 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+
+    _coordinates = "rtz"
+    _units = "(T)"
+    _print_value_fmt = "Piecewise omnigenity error: "
+
+    def __init__(
+        self,
+        eq,
+        field,
+        target=None,
```
f0uriest, line 922: add a `*,` here to make everything except eq, field keyword only

desc/objectives/_omnigenity.py
```diff
@@ -847,6 +846,379 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+        bounds=None,
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        eq_grid=None,
+        field_grid=None,
+        deriv_mode="rev",
```
f0uriest: does this need to be reverse? If not better to leave this as "auto"

desc/objectives/_omnigenity.py
```diff
@@ -847,6 +846,379 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+        M_booz=None,
+        N_booz=None,
+        eq_fixed=False,
+        field_fixed=False,
+        overlap_penalty=1.0,
+        p=10,
```
f0uriest, line 936: need to add these to docstring

desc/objectives/_omnigenity.py
```diff
@@ -847,6 +846,379 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+        timer.start("Precomputing transforms")
+
+        profiles = get_profiles(self._eq_data_keys, obj=eq, grid=eq_grid)
+        eq_transforms = get_transforms(
+            self._eq_data_keys,
+            obj=eq,
+            grid=eq_grid,
+        )
+
+        # we need a uniform grid to get correct surface averages for iota
```
f0uriest, line 1025: iirc the boozer transform also needs a uniform grid so you can use the same one for both (and add a check that the eq grid is uniform)

desc/objectives/_omnigenity.py
```diff
@@ -847,6 +846,379 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+
+        # However, if field is fixed and eq changes, it changes the boozer grid
+        # field transform and field data
+
+        timer.stop("Precomputing transforms")
+        if verbose > 1:
+            timer.disp("Precomputing transforms")
+
+        if self._normalize:
+            # normalized with max |B|_pwO
```
f0uriest, line 1072: couldn't this use the same `scales['B']` that we use elsewhere?

desc/compute/_omnigenity.py
```diff
@@ -916,6 +917,182 @@ def _interp(x, B):
...
+    # shape (num surfaces, grid points)
+    exponent = -1 * (
+        ((zeta_B - zeta_C + t_1 * (theta_B - theta_C)) / w_1) ** (2 * p)
+        + ((theta_B - theta_C + t_2 * (zeta_B - zeta_C)) / w_2) ** (2 * p)
+    )
+
+    B_pwO = B_min + (B_max - B_min) * jnp.exp(exponent)
+
+    # Flattened array on each surface.
+    # Reshaping may cause jit-related issues
```
f0uriest, line 982: what issues? is this shape (nrho, ntheta*nzeta)?

desc/compute/_omnigenity.py
f0uriest: maybe not, i see in the objective that its not a meshgrid in boozer coordinates.

desc/objectives/_omnigenity.py
```diff
@@ -847,6 +846,379 @@ def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
...
+                transforms=field_transforms,
+                profiles={},
+                data={},
+                iota=data["iota"],
+                p=constants["p"],
+            )
+
+        Ntheta = constants["Ntheta_B"]
+
+        # Rolling ensures max(B) occurs in the corners
```
f0uriest, line 1193: can you explain this a bit?

tests/inputs/master_compute_data_rpz.pkl
f0uriest, line 1: seems weird that we added a new class and the data got smaller?

tests/test_objective_funs.py
```diff
@@ -3552,6 +3555,54 @@ def test_compute_scalar_resolution_omnigenity(self):
...
+            major_radius=1,
+            aspect_ratio=20,
+            elongation=6,
+            mirror_ratio=0.2,
+            torsion=0.1,
+            NFP=1,
+            sym=True,
+        )
+        eq = Equilibrium(Psi=6e-3, M=4, N=4, surface=surf)
+        eq, _ = eq.solve(objective="force", verbose=3)
```
f0uriest, line 3571: does it need to actually be in equilibrium for this test to work? if not can remove this to save some time

dpanici:
What was the status of this one?

PR #995, opened 2024-04-15
Add objective for coil current-length

### LINKED ISSUE #902: Objective for coil current length
ie, length * current, proxy for cost of HTS

kianorr:
Add coil current length objective, as it is useful for approximating HTS cost.

Resolves #902


desc/objectives/_coils.py
```diff
@@ -607,6 +607,122 @@ def compute(self, params, constants=None):
...
+            Dictionary of constant data, eg transforms, profiles etc. Defaults to
+            self._constants.
+
+        Returns
+        -------
+        f : float or array of floats
+        """
+        lengths = super().compute(params, constants=constants)
+        currents = tree_flatten(
+            self.things[0].current, is_leaf=lambda x: isinstance(x, float)
```
f0uriest, line 720: `current` can also be optimized, so should use the value from `params` rather than whatever the initial current was

desc/objectives/_coils.py
```diff
@@ -607,6 +607,121 @@ def compute(self, params, constants=None):
...
+        reverse mode and forward over reverse mode respectively.
+    grid : Grid, optional
+        Collocation grid containing the nodes to evaluate at.
+        Defaults to ``LinearGrid(N=2 * coil.N + 5)``
+    name : str, optional
+        Name of the objective function.
+    """
+
+    _scalar = False
+    _units = "(A)"
```
f0uriest, line 655: A*m

desc/objectives/_coils.py
```diff
@@ -607,6 +607,121 @@ def compute(self, params, constants=None):
...
+        use_jit : bool, optional
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        super().build(use_jit=use_jit, verbose=verbose)
+
+        if self._normalize:
+            self._normalization = 1 / self._scales["a"]
```
f0uriest, line 701: should include something with units of current here. Maybe just mu0? or maybe use whatever the average current in the coilset is at build time?

tests/test_objective_funs.py
```diff
@@ -2180,7 +2208,9 @@ def test_objective_no_nangrad(self, objective):
         assert not np.any(np.isnan(g)), str(objective)
 
     @pytest.mark.unit
-    @pytest.mark.parametrize("objective", [CoilLength, CoilTorsion, CoilCurvature])
+    @pytest.mark.parametrize(
+        "objective", [CoilLength, CoilTorsion, CoilCurvature, CoilCurrentLength]
```
f0uriest, line 2212: should also add to `test_compute_scalar_resolution_coils`

f0uriest (changes requested):
few minor comments. Default target of 0 is probably fine unless we can think of something better

desc/objectives/_coils.py
kianorr: I did average current but why would mu0 work? I thought mu0 had units other than A?

f0uriest: It's not exactly right, but it usually gives the right order of magnitude.

Generally, current ~ field/length/mu0. So assuming the field strength is order 1T, then a reasonable scale for the current is length*mu0. But for this average current is probably better.

f0uriest (approved):

dpanici (changes requested):
If the coils initially have zero current for some reason, the noramzliation would be infinite, can you add a check after you calculate the average current that it is nonzero, and if it is zero you set it to one or something?

dpanici:
> @dpanici want me to add a test where the current is zero or is it okay without it?

This is fine

desc/objectives/_coils.py
```diff
@@ -607,6 +607,128 @@ def compute(self, params, constants=None):
...
+
+        """
+        super().build(use_jit=use_jit, verbose=verbose)
+
+        if self._normalize:
+            coils = self.things[0]
+            params = tree_flatten(
+                coils.params_dict, is_leaf=lambda x: isinstance(x, dict)
+            )[0]
+            average_current = np.mean([param["current"] for param in params])
```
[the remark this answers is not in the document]
ddudt: FYI you can use `tree_leaves` instead of `tree_flatten()[0]` to get only the leaves

desc/objectives/_coils.py
```diff
@@ -607,6 +607,128 @@ def compute(self, params, constants=None):
...
+            if not average_current:
```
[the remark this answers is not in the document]
ddudt: another option: `average_current = max(average_current, 1)`

desc/objectives/_coils.py
```diff
@@ -607,6 +607,128 @@ def compute(self, params, constants=None):
...
+                average_current = 1
+            self._normalization = 1 / (average_current * self._scales["a"])
```
ddudt, line 708: The normalization should be the inverse of this (current * a) so that it has the same units as the objective compute values (A*m). The objective residuals are divided by the normalization, not multiplied.

desc/objectives/_coils.py
```diff
@@ -607,6 +607,128 @@ def compute(self, params, constants=None):
         return out
 
 
+class CoilCurrentLength(CoilLength):
```
ddudt, line 610: I don't think it makes sense to inherit from `CoilLength`, I think we should inherit from `_CoilObjective` instead. This doesn't save much code and it feels weird to overwrite the normalization already set by `CoilLength`. And thematically, I feel `CoilLength` and `CoilCurrentLength` should be the same "generation" in the family tree.

ddudt (changes requested):

desc/objectives/_coils.py
ddudt: Never mind

f0uriest (dismissed):

desc/objectives/_coils.py
f0uriest: I think this and the previous comment have been resolved?

dpanici (dismissed):

desc/objectives/_coils.py
```diff
@@ -103,11 +96,12 @@ def build(self, use_jit=True, verbose=1):  # noqa:C901
 
         """
         # local import to avoid circular import
-        from desc.coils import CoilSet, MixedCoilSet, _Coil
+        from desc.coils import CoilSet, MixedCoilSet
 
         self._dim_f = 0
         self._quad_weights = jnp.array([])
 
+        # FIXME: Is this necessary? It only gets called on a Grid, never coils/coilset
```
kianorr, line 104: @daniel-dudt Yeah I'm pretty certain it's necessary. the input is a pytree whose leaves are Grid -- it is in the same pytree structure as the user-inputted coilset so that it can go through each grid to get the total dim_f

ddudt: If it is a pytree of Grids, then won't only the first and last if statements ever trigger? I don't see when the `CoilSet` checks will ever trigger. 
```
if isinstance(coilset, list):
    # this will trigger at the branches
elif isinstance(coilset, MixedCoilSet):
    # I don't think this is necessary
elif isinstance(coilset, CoilSet):
    # I don't think this is necessary
elif isinstance(coilset, _Grid):
    # this will trigger at the leaves
```

I think this could be simplified with something like the following: 
```
grids = tree_leaves(tree_of_grids, is_leaf=lambda x: hasattr(x, "num_nodes"))
self._dim_f = np.sum([grid.num_nodes for grid in grids])
self._quad_weights = np.concatenate([grid.spacing[:, 2] for grid in grids])
```

kianorr: It's a pytree of objects that are of type `CoilSet` or `MixedCoilSet` but have Grids as leaves so every elif statement is used. I think Dario had the exact same question a while back: https://github.com/PlasmaControl/DESC/pull/853#discussion_r1526764746

And using `hasattr(x, "num_nodes")` is smart, I didn't think to use that at the time. I'll have to try it but it seems like it would work

kianorr: thank you! :)

dpanici (approved):

PR #996, opened 2024-04-16
Biot-Savart volume integral

ddudt:
Computes B_plasma in a volume as: 

𝐁ᵥ(𝐫) = μ₀/4π ∫ 𝐉(𝐫') × (𝐫 − 𝐫')/|𝐫 − 𝐫'|³ d³𝐫'

This is useful for when you need B_total = B_plasma + B_external in the full computational domain, not just on the boundary surface.


dpanici:
use this fxn https://github.com/PlasmaControl/DESC/blob/b10f8c8be9d3e751684e5d34abecaf5cd40b2205/desc/magnetic_fields/_core.py#L31

dpanici:
`scipy.integrate.tpl_quad` to check this

dpanici:
[IAEA2020_JGeiger_Manuscript_8p_finalversion (1).pdf](https://github.com/user-attachments/files/16090258/IAEA2020_JGeiger_Manuscript_8p_finalversion.1.pdf)
[Contributions to Plasma Physics - 2010 - Geiger - Effects of Net Currents on the Magnetic Configuration of W7‐X (2).pdf](https://github.com/user-attachments/files/16090260/Contributions.to.Plasma.Physics.-.2010.-.Geiger.-.Effects.of.Net.Currents.on.the.Magnetic.Configuration.of.W7.X.2.pdf)

f0uriest:
Several people at the scidac meeting today are using the BMW code: https://github.com/ORNL-Fusion/BMW/tree/master

refs: https://scipub.euro-fusion.org/wp-content/uploads/eurofusion/WPS1CPR17_17542_submitted-4.pdf
https://bpb-us-e2.wpmucdn.com/wordpress.auburn.edu/dist/5/118/files/2022/05/ISHW2017.pdf

basically it does a biot-savart volume integral to compute A (not B), then takes the finite difference curl of A to get B. As far as I can tell from the code the source grid is hardcoded to ntheta=101, ns, nzeta taken from the vmec/mgrid file. Allegedly it works well and gives accurate results?

Could try something similar (and maybe use spectral derivatives for going A->B)

ddudt:
Try implementing Eq. 17-18 of [Peraza-Rodriguez et al. _Phys. Plasmas_ 2017](https://doi-org.ezproxy.princeton.edu/10.1063/1.4986447)
It should be a simple linear system solve to get the Fourier-Zernike coefficients of the vector potential as: 
$\nabla \times (\nabla \times \mathbf{A}) = \mu_0 \mathbf{J}$
And then the magnetic field is: 
$\mathbf{B} = \nabla \times \mathbf{A}$
And all of this can be done in the native $(\rho,\theta,\zeta)$ coordinate system.

ddudt:
Update: I fixed a silly bug and now this is working. Direct Biot-Savart volume integral to compute B and the similar volume integral to compute A both agree well with BMW results at moderate resolution. But more benchmarking is needed.

rahulgaur104:
Thank you! Could be really useful for my umbilic stuff!

desc/integrals/singularities.py
```diff
@@ -1103,3 +1104,95 @@ def compute_B_plasma(
...
+    assert eval_grid.shape[1] == 3
+    if source_grid is None:
+        source_grid = QuadratureGrid(L=eq.L_grid, M=eq.M_grid, N=eq.N_grid * eq.NFP)
+
+    data_keys = ["J", "phi", "sqrt(g)", "x"]
+    data = eq.compute(data_keys, grid=source_grid)
+
+    # shape = (source_grid.num_nodes, eval_grid.num_nodes, 3)  # noqa: E800
+    dx = rpz2xyz(eval_grid)[None, :, :] - rpz2xyz(data["x"])[:, None, :]
+    B = np.sum(  # sum over source_grid nodes
```
dpanici: use biot_savart_general instead

dpanici:
- show plots of error
- errors may be due to large matrix dot product errors?

desc/equilibrium/equilibrium.py
```diff
@@ -1205,6 +1211,126 @@ def need_src(name):
...
+    def compute_magnetic_field(
+        self,
+        coords,
+        params=None,
+        basis="rpz",
+        source_grid=None,
+        transforms=None,
+        chunk_size=None,
+    ):
+        """Compute magnetic field at a set of points.
```
dpanici, line 1282: be more specific: this is the magnetic field generated by the plasma currents.

I think also we could add a flag for external evaluation or not, that will allow the use of the virtual casing principle if desired (if one always knows the evaluation points are far from the plasma surface, like is the case with a lot of diagnostic flux loops)

desc/equilibrium/equilibrium.py
```diff
@@ -1205,6 +1211,126 @@ def need_src(name):
...
+        -------
+        field : ndarray, shape(n,3)
+            magnetic field at specified points
+
+        """
+        coords = jnp.atleast_2d(coords)
+        eval_xyz = rpz2xyz(coords) if basis.lower() == "rpz" else coords
+        if source_grid is None:
+            source_grid = QuadratureGrid(
+                L=self.L_grid, M=self.M_grid, N=self.N_grid * self.NFP
```
dpanici: don't do this, instead use NFP = eq.NFP and then use the `nfp_loop` [helper function](https://github.com/PlasmaControl/DESC/blob/1d4806d280d11722839a73aee74f0888d07eeca1/desc/magnetic_fields/_current_potential.py#L1084) to rotate the current density vector around the field periods.

desc/equilibrium/equilibrium.py
```diff
@@ -1205,6 +1211,126 @@ def need_src(name):
...
+
+    def compute_magnetic_vector_potential(
+        self,
+        coords,
+        params=None,
+        basis="rpz",
+        source_grid=None,
+        transforms=None,
+        chunk_size=None,
+    ):
```
dpanici, line 1417: same comment on option for virtual casing principle

desc/equilibrium/equilibrium.py
```diff
@@ -1205,6 +1211,126 @@ def need_src(name):
...
+        -------
+        A : ndarray, shape(n,3)
+            magnetic vector potential at specified points
+
+        """
+        coords = jnp.atleast_2d(coords)
+        eval_xyz = rpz2xyz(coords) if basis.lower() == "rpz" else coords
+        if source_grid is None:
+            source_grid = QuadratureGrid(
+                L=self.L_grid, M=self.M_grid, N=self.N_grid * self.NFP
```
dpanici: same comment about nfp_loop

desc/equilibrium/equilibrium.py
```diff
@@ -1205,6 +1211,126 @@ def need_src(name):
...
+            )
+
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message="Unequal number of field periods")
+            data = self.compute(
```
dpanici, line 1337: will this be jitable if we use self.compute? I dont think so

dpanici (changes requested):
- use nfp_loop
- I think worth adding an option for virtual casing principle

dpanici:
> Possibly redundant with #1436, should merge them once we decide on a public API.

I dont think it is redundant as #1436 does not expose this API at all to the user, it is done opaquely inside of the objective compute function. So there is no dependency there

desc/basis.py
```diff
@@ -1118,6 +1120,196 @@ def change_resolution(self, L, M, N, NFP=None, sym=None):
             self._set_up()
 
 
+class DoubleChebyshevFourierBasis(_Basis):
```
dpanici, line 1123: Do we want to make things clearer that this is a real space basis (as opposed to flux coordinates) in the name/docstring?

desc/basis.py
```diff
@@ -1118,6 +1120,196 @@ def change_resolution(self, L, M, N, NFP=None, sym=None):
...
+    ----------
+    L : int
+        Maximum radial resolution.
+    M : int
+        Maximum toroidal resolution.
+    N : int
+        Maximum vertical resolution.
+    NFP : int
+        Number of field periods.
+    sym : {``'cos'``, ``'sin'``, ``False``}
```
dpanici, line 1138: there is not 't' in real space, what do these symmetries refer to now?

desc/basis.py
```diff
@@ -1118,6 +1120,196 @@ def change_resolution(self, L, M, N, NFP=None, sym=None):
...
+class DoubleChebyshevFourierBasis(_Basis):
+    """3D basis: tensor product of two Chebyshev polynomials and Fourier series.
+
+    Chebyshev polynomial in R and Z, and Fourier series in phi.
+
+    Parameters
+    ----------
+    L : int
+        Maximum radial resolution.
+    M : int
```
dpanici, line 1132: I don't love this, almost everywhere in the code `N` is toroidal, yet here we change it so `M` is toroidal. I would prefer if here we don't use `M`, and keep `N` as toroidal and use something else for the vertical resolution (Or be explicit and write `N_R` `N_phi` `N_Z` as the resolutions

desc/basis.py
```diff
@@ -1118,6 +1120,196 @@ def change_resolution(self, L, M, N, NFP=None, sym=None):
...
+        n = n.ravel()
+        return np.array([l, m, n]).T
+
+    def evaluate(self, grid, derivatives=np.array([0, 0, 0]), modes=None):
+        """Evaluate basis functions at specified nodes.
+
+        Parameters
+        ----------
+        grid : Grid or ndarray of float, size(num_nodes,3)
+            Node coordinates, in (rho,phi,Z).
```
dpanici, line 1194: rho or R?

desc/basis.py
```diff
@@ -1118,6 +1120,196 @@ def change_resolution(self, L, M, N, NFP=None, sym=None):
...
+        derivatives : ndarray of int, shape(num_derivatives,3)
+            Order of derivatives to compute in (rho,phi,Z).
```
dpanici, line 1196: rho or R?

desc/basis.py
```diff
@@ -1118,6 +1120,196 @@ def change_resolution(self, L, M, N, NFP=None, sym=None):
...
+            noutidx = self.inverse_N_idx
+        else:
+            lidx = loutidx = np.arange(len(modes))
+            midx = moutidx = np.arange(len(modes))
+            nidx = noutidx = np.arange(len(modes))
+        if not len(modes):
+            return np.array([]).reshape((grid.num_nodes, 0))
+
+        # Get the nodes of the grid
+        r, p, z = grid.nodes.T
```
dpanici, line 1233: nitpick but if r is really cylindrical R, use R, same for Z

desc/basis.py
```diff
@@ -1118,6 +1120,196 @@ def change_resolution(self, L, M, N, NFP=None, sym=None):
...
+        l, m, n = modes.T
+
+        # The coordinates in the grid are hardcoded to be called rho, theta, zeta,
+        # but these vars are just unique values for 1st, 2nd, and 3rd coordinates
```
dpanici, line 1237: This will change, presumedly, when we figure out our new grid stuff @ddudt #1840

desc/basis.py
```diff
@@ -1851,13 +2043,246 @@ def chebyshev(r, l, dr=0):
...
+        return dydr
+    else:
         raise NotImplementedError(
-            "Analytic radial derivatives of Chebyshev polynomials "
+            "Analytic third-degree derivatives of Chebyshev polynomials "
             + "have not been implemented."
         )
 
 
+def chebfit(y, axis):
```
dpanici, line 2093: we may aready have some utility functions that do this, see https://github.com/PlasmaControl/DESC/blob/702ae496397a3703d8b0c0a2586f6754016f5a10/desc/integrals/basis.py#L260

desc/basis.py
```diff
@@ -1851,13 +2043,246 @@ def chebyshev(r, l, dr=0):
...
+    f = f / (N - 1)
+
+    # Reshape to braodcast onto axis
+    f_shape = [1 if i != axis else -1 for i in range(y_c.ndim)]
+
+    # Adjust normalization and then inverse direct cosine transform
+    return idct(y_c / f.reshape(f_shape), axis=axis, type=1, norm=None)
+
+
+def fftfit(y, axis, n=None):
```
dpanici, line 2159: if these are to be public, some more safeguards should be added. If these are not meant to be public, add _ before the name

desc/basis.py
```diff
@@ -1851,13 +2043,246 @@ def chebyshev(r, l, dr=0):
...
+    """
+    Real fast fourier transform along axis.
+
+    Designed to convert coefficients into the form expected by the
```
dpanici, line 2163: what is meant here? takes in a function evaluated at linearly spaced nodes, then fourier transforms it, then adjusts the coefficients to fit what DESC's fourier function requires?

desc/basis.py
```diff
@@ -1851,13 +2043,246 @@ def chebyshev(r, l, dr=0):
...
+    """
+    Real fast inverse fourier transform along axis.
+
+    Designed to convert from form in basis classes to real space.
+    Evaluates on equally spaced nodes.
+
+    Parameters
+    ----------
+    y : ndarray, shape(...,2N+1,...)
+        Function to decompose into basis of Chebyshev functions.
```
dpanici, line 2234: is this docstring accurate?

desc/compute/nabla.py
```diff
@@ -0,0 +1,163 @@
...
+        the curl. out_R should never be 0.
+    in_transform: Transform
+        transform from the real grid on which A is defined to
+        a spectral basis in which partial derivatives of A can be evaluated.
+        the pseudoinverse of the transform must have been built.
+    out_transform: Transform
+        transform from the spectral basis on which A is calculated to the
+        real grid on which the curl is to be evaluated.
+        the transform must have been built, with derivs>=1.
+    scales: jnp.ndarray, shape (3,)
```
dpanici, line 35: shouldn't this be done in the transform itself, not here?

desc/compute/nabla.py
```diff
@@ -0,0 +1,163 @@
...
+        This parameter adjusts the dimensions of the partial derivatives so they are
+        taken with respect to the normal coordinates, not the coordinates in the
+        transform.
+
+    Returns
+    -------
+    curl_A : ndarray, shape(n,3)
+        The curl of the vector field, in cylindrical coordinates.
```
dpanici, line 43:
```suggestion
        The curl of the vector field, in cylindrical coordinates, at the points specified by `out_transform`.
```

desc/compute/nabla.py
```diff
@@ -0,0 +1,163 @@
...
+    Parameters
+    ----------
+    A : ndarray, shape(n,3)
+        vector field, in cylindrical (R,phi,Z) form
+    in_R : ndarray, shape(n,)
+        radial distance for each point of A
+    out_R : ndarray, shape(n,)
+        radial location for each point at which to calculate
+        the curl. out_R should never be 0.
+    in_transform: Transform
```
dpanici, line 27: specify these should be bases and grids of R, phi, Z

desc/compute/nabla.py
```diff
@@ -0,0 +1,163 @@
...
+        The curl of the vector field, in cylindrical coordinates.
+    """
+    # Calculate matrix of terms for the curl
+    # (dims: datapoint, component index, derivative index)
+    terms = jnp.zeros((out_R.shape[0], 3, 3))
+    for c in range(3):
+        for d in range(3):
+            if c == 1 and d == 0:
+                # partial (R*A_phi)/partial R instead of partial A_phi/partial R
+                term_cd = out_transform.transform(RA_phi_coeff, dr=1)
```
dpanici, line 83: cant you just use chain rule and do d/dR (R Aphi) = Aphi + R d(Aphi)/dR? then you only need to do one fit above, and all of the partial derivatives are "consistent" in that they are all derivaitves of A, instead of derivatives of A and some derivatves of `R*A` (which may have a different spectral width than A i.e. same resolution basis may be a worse for for `R*A` then for A)

desc/compute/nabla.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Returns
+    -------
+    div_A : ndarray, shape(n,)
+        The divergence of the vector field, in cylindrical coordinates.
+    """
+    A_coeff = in_transform.fit(
+        A * jnp.stack([in_R, jnp.ones_like(in_R), jnp.ones_like(in_R)], axis=-1)
+    )
+    # Calculate matrix of terms for the curl
```
dpanici, line 145:
```suggestion
    # Calculate matrix of terms for the div
```

desc/compute/nabla.py
```diff
@@ -0,0 +1,163 @@
+"""
+Functions for taking the curl and divergence in cylindrical coordinates.
```
dpanici, line 2: usually `desc.compute` has things for computing stuff in the compute dicts. This I could see as useful in the future when we have things in the compute dict using RpZ grids (#1805 ), for now let's call this like `_nabla_utils.py` maybe or `_real_space_deriv_utils.py`

desc/equilibrium/equilibrium.py
dpanici: This would also need to change if we commit to the RpZ api in #1805

desc/equilibrium/equilibrium.py
```diff
@@ -1224,6 +1236,217 @@ def need_src(name):
...
+        assert (
+            method in methods
+        ), f"""Method {method} unknown. Please choose one of the following methods:
+            {', '.join(methods)}"""
+
+        # Convert coords to a 2D Jax array
+        coords = jnp.atleast_2d(coords)
+
+        # Biot-Savart method
+        if method == methods[0]:
```
dpanici, line 1328:
```suggestion
        if method == "biot-savart":
```
I prefer to be verbose in the code here

desc/equilibrium/equilibrium.py
```diff
@@ -1224,6 +1236,217 @@ def need_src(name):
...
+                )
+                f += fj
+                return f
+
+            B = fori_loop(0, source_grid.NFP, nfp_loop, jnp.zeros_like(coords))
+            if basis.lower() == "rpz":
+                B = xyz2rpz_vec(B, phi=coords[:, 1])
+
+        # Virtual casing method
+        elif method == methods[1]:
```
dpanici, line 1374:
```suggestion
        elif method == "virtual casing":
```

desc/integrals/virtual_casing.py
```diff
@@ -0,0 +1,105 @@
+"""Virtual casing surface integrals for points strictly outside the surface."""
+
+from desc.backend import fori_loop, jnp
+from desc.batching import batch_map
+from desc.utils import errorif, xyz2rpz_vec
+
+
+def integrate_surface(coords, source_data, source_grid, kernel, chunk_size=None):
```
dpanici, line 8: maybe different name like integrate_surface_at_external_point

it does not need to be strictly outside per-say right? just needs to be off-surface, the kernels don't really care about outside or inside as far as the singularity goes

dpanici (changes requested):
partially reviewed but already some changes noted

desc/equilibrium/equilibrium.py
```diff
@@ -1257,6 +1269,217 @@ def need_src(name):
...
+                params=params,
+                transforms=transforms,
+                override_grid=False,
+            )
+
+            # Integrate the virtual casing integral over the plasma boundary
+            B = integrate_surface(
+                eval_rpz, source_data, source_grid, kernel, chunk_size=chunk_size
+            )
+            if basis.lower == "xyz":
```
dario-panici, line 1404:
```suggestion
            if basis.lower() == "xyz":
```

desc/equilibrium/equilibrium.py
```diff
@@ -1257,6 +1269,217 @@ def need_src(name):
...
+                B = rpz2xyz_vec(B, phi=coords[:, 1])
```
dario-panici, line 1405:
```suggestion
                B = rpz2xyz_vec(B, phi=eval_rpz[:, 1])
```
coords is xyz if basis is xyz

PR #998, opened 2024-04-17
Switch back to old version of the zernike radial function

### LINKED ISSUE #941: Regression in ``_initial_guess_surface`` from v0.10.4
On master now initializing an ``Equilibrium`` takes about 2x longer than it did on the last release v0.10.4, the culprit seems to be the ``_initial_guess_surface`` function.

In jupyter:
```python
jax.clear_caches()
%lprun -f Equilibrium.__init__ \
   -f desc.equilibrium.initial_guess.set_initial_guess \
   -f desc.equilibrium.initial_guess._initial_guess_surface \
   eq = Equilibrium(L=12, M=12, N=12)
```

I suspect we didn't see it because the benchmarks run the same code over and over so you get the advantage of JIT caching, but in practice you're usually only creating a single equilibrium so JIT doesn't really help. We should add `jax.clear_caches()` to those benchmarks to avoid this in the future.

YigitElma:
Until we decide to whether use new repo for zernike_radial, we switch back to the old version of the code. 
Resolves #941

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -174,7 +174,7 @@ If the 2nd case is the reason, then you must update the ``master_compute_data.pk
 
    except AssertionError as e:
       error = True
-       print(e)
+      print(e)
 
 with
 
@@ -183,7 +183,7 @@ with
    except AssertionError as e:
       error = False
       update_master_data = True
-       print(e)
+      print(e)
 
 
 -  rerun the test ``pytest tests -k test_compute_everything``, now any compute quantity that is different between the PR and master will be updated with the PR value
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -1683,6 +1683,42 @@ def zernike_norm(l, m):
     return np.sqrt((2 * (l + 1)) / (np.pi * (1 + int(m == 0))))
 
 
+@jit
+@jnp.vectorize
+def _binom(n, k):
+    """Binomial coefficient.
+
+    Implementation is only correct for positive integer n,k and n>=k
+
+    Parameters
+    ----------
+    n : int, array-like
+        number of things to choose from
+    k : int, array-like
+        number of things chosen
+
+    Returns
+    -------
+    val : int, float, array-like
+        number of possible combinations
+    """
+    # adapted from scipy:
+    # https://github.com/scipy/scipy/blob/701ffcc8a6f04509d115aac5e5681c538b5265a2/
+    # scipy/special/orthogonal_eval.pxd#L68
+
+    n, k = map(jnp.asarray, (n, k))
+
+    def _binom_body_fun(i, b_n):
+        b, n = b_n
+        num = n + 1 - i
+        den = i
+        return (b * num / den, n)
+
+    kx = k.astype(int)
+    b, n = fori_loop(1, 1 + kx, _binom_body_fun, (1.0, n))
+    return b
+
+
 @custom_jvp
 @jit
 @jnp.vectorize
@@ -1748,42 +1784,6 @@ def _jacobi(n, alpha, beta, x, dx=0):
     out = jnp.where(n == 0, 1.0, out)
     out = jnp.where(n == 1, 0.5 * (2 * (alpha + 1) + (alpha + beta + 2) * (x - 1)), out)
     return c * out
-
-
-@jit
-@jnp.vectorize
-def _binom(n, k):
-    """Binomial coefficient.
-
-    Implementation is only correct for positive integer n,k and n>=k
-
-    Parameters
-    ----------
-    n : int, array-like
-        number of things to choose from
-    k : int, array-like
-        number of things chosen
-
-    Returns
-    -------
-    val : int, float, array-like
-        number of possible combinations
-    """
-    # adapted from scipy:
-    # https://github.com/scipy/scipy/blob/701ffcc8a6f04509d115aac5e5681c538b5265a2/
-    # scipy/special/orthogonal_eval.pxd#L68
-
-    n, k = map(jnp.asarray, (n, k))
-
-    def _binom_body_fun(i, b_n):
-        b, n = b_n
-        num = n + 1 - i
-        den = i
-        return (b * num / den, n)
-
-    kx = k.astype(int)
-    b, n = fori_loop(1, 1 + kx, _binom_body_fun, (1.0, n))
-    return b
 
 
 @_jacobi.defjvp
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -1510,7 +1510,7 @@ def zernike_radial(r, l, m, dr=0):
         basis function(s) evaluated at specified points
 
     """
-    m = jnp.abs(m)
+    m = jnp.abs(m).astype(float)
     alpha = m
     beta = 0
     n = (l - m) // 2
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -7,7 +7,7 @@ from math import factorial
 import mpmath
 import numpy as np
 
-from desc.backend import cond, custom_jvp, fori_loop, gammaln, jit, jnp, sign, switch
+from desc.backend import custom_jvp, fori_loop, gammaln, jit, jnp, sign
 from desc.io import IOAble
 from desc.utils import check_nonnegint, check_posint, flatten_list
 
@@ -767,7 +767,7 @@ class ZernikePolynomial(_Basis):
             lm = lm[lmidx]
             m = m[midx]
 
-        radial = zernike_radial(r, lm[:, 0], lm[:, 1], dr=derivatives[0])
+        radial = zernike_radial(r[:, np.newaxis], lm[:, 0], lm[:, 1], dr=derivatives[0])
         poloidal = fourier(t[:, np.newaxis], m, 1, derivatives[1])
 
         if unique:
@@ -1129,7 +1129,7 @@ class FourierZernikeBasis(_Basis):
             m = m[midx]
             n = n[nidx]
 
-        radial = zernike_radial(r, lm[:, 0], lm[:, 1], dr=derivatives[0])
+        radial = zernike_radial(r[:, np.newaxis], lm[:, 0], lm[:, 1], dr=derivatives[0])
         poloidal = fourier(t[:, np.newaxis], m, dt=derivatives[1])
         toroidal = fourier(z[:, np.newaxis], n, NFP=self.NFP, dt=derivatives[2])
         if unique:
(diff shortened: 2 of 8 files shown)


YigitElma:
The compute_all test is failing due to small numerical differences between 2 methods, but I don't understand why map_coordinates_derivative test fails. I replaced the original function which was forward backward autodiff capable.

YigitElma:
It seems like the problem is `r**m` part, but I don't know why that was not an issue before.

```python
desc/compute/utils.py:405: in get_transforms
    t.build()
desc/transform.py:385: in build
    self.matrices["direct1"][d[0]][d[1]][d[2]] = self.basis.evaluate(
desc/basis.py:1132: in evaluate
    radial = zernike_radial(r[:, np.newaxis], lm[:, 0], lm[:, 1], dr=derivatives[0])
desc/basis.py:1520: in zernike_radial
    out = r**m * _jacobi(n, alpha, beta, jacobi_arg, 0)
../../miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/numpy/array_methods.py:736: in op
    return getattr(self.aval, f"_{name}")(self, *args)
../../miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/numpy/array_methods.py:264: in deferring_binary_op
    return binary_op(*args)
```

YigitElma:
If I just run jacfwd and jit on the zernike_radial function, I get following output,

```python
import jax
from desc.basis import zernike_radial, ZernikePolynomial

basis = ZernikePolynomial(L=10, M=10, spectral_indexing="ansi", sym="cos")
r = np.linspace(0, 1, 100)

J1 = jax.jacfwd(zernike_radial)(r[:, np.newaxis], basis.modes[:,0], basis.modes[:,1], 0)
J2 = jax.jit(jax.jacrev(zernike_radial))(r[:, np.newaxis], basis.modes[:,0], basis.modes[:,1])
J3 = jax.jit(jax.jacrev(zernike_radial))(r[:, np.newaxis], basis.modes[:,0], basis.modes[:,1], 0)
```

```
TracerBoolConversionError                 Traceback (most recent call last)
Cell In[52], [line 9](vscode-notebook-cell:?execution_count=52&line=9)
      [7](vscode-notebook-cell:?execution_count=52&line=7) J1 = jax.jacfwd(zernike_radial)(r[:, np.newaxis], basis.modes[:,0], basis.modes[:,1], 0)
      [8](vscode-notebook-cell:?execution_count=52&line=8) J2 = jax.jit(jax.jacrev(zernike_radial))(r[:, np.newaxis], basis.modes[:,0], basis.modes[:,1])
----> [9](vscode-notebook-cell:?execution_count=52&line=9) J3 = jax.jit(jax.jacrev(zernike_radial))(r[:, np.newaxis], basis.modes[:,0], basis.modes[:,1], 0)

    [... skipping hidden 31 frame]

File [~/Codes/DESC/desc/basis.py:1519](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/Codes/DESC/desc/basis.py:1519), in zernike_radial(r, l, m, dr)
   [1517](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/Codes/DESC/desc/basis.py:1517) s = (-1) ** n
   [1518](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/Codes/DESC/desc/basis.py:1518) jacobi_arg = 1 - 2 * r**2
-> [1519](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/Codes/DESC/desc/basis.py:1519) if dr == 0:
   [1520](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/Codes/DESC/desc/basis.py:1520)     out = jnp.pow(r, m) * _jacobi(n, alpha, beta, jacobi_arg, 0)
   [1521](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/Codes/DESC/desc/basis.py:1521) elif dr == 1:

    [... skipping hidden 1 frame]

File [~/miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/core.py:1492](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/core.py:1492), in concretization_function_error.<locals>.error(self, arg)
   [1491](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/core.py:1491) def error(self, arg):
-> [1492](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/core.py:1492)   raise TracerBoolConversionError(arg)

TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[]..
The error occurred while tracing the function zernike_radial at [/home/yigit/Codes/DESC/desc/basis.py:1487](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/desc/basis.py:1487) for jit. This value became a tracer due to JAX operations on these lines:

  operation a:bool[] = eq b c
    from line [/home/yigit/Codes/DESC/desc/basis.py:1519:7](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/desc/basis.py:1519:7) (zernike_radial)
See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.TracerBoolConversionError
```

which seems like if you try to jit a function which has partial(jit, static_argnum=...) decorator not by calling it by executing jax.jit(func), it doesn't care about the static_argnum part. Only if the parameter dr is not given, jit works as expected. 

Though, these 2 error messages are quite different from each other, the former complains about the r**m*_jacobi part, the latter says dr is not static, it is traced. I am confused.

<details>
Here, I am using directly on zernike_radial but I got similar behavior for nested functions. For example,

```python
import jax
import functools
import jax.numpy as jnp

@functools.partial(jax.jit, static_argnums=1)
def fun(x, dx=0):
    if dx == 0:
        return x
    else:
        return x**dx
    
@jax.jit
def main(t, dx=0):
    x = t**2
    return fun(x, dx)

J1 = jax.jit(main)(jnp.array([1, 2, 3]), 2)
```

```python
TracerBoolConversionError                 Traceback (most recent call last)
Cell In[53], [line 1](vscode-notebook-cell:?execution_count=53&line=1)
----> [1](vscode-notebook-cell:?execution_count=53&line=1) asd = jax.jit(main)(jnp.array([1, 2, 3]), 2)
      [2](vscode-notebook-cell:?execution_count=53&line=2) print(asd)

    [... skipping hidden 24 frame]

Cell In[44], [line 14](vscode-notebook-cell:?execution_count=44&line=14)
     [11](vscode-notebook-cell:?execution_count=44&line=11) @jax.jit
     [12](vscode-notebook-cell:?execution_count=44&line=12) def main(t, dx=0):
     [13](vscode-notebook-cell:?execution_count=44&line=13)     x = t**2
---> [14](vscode-notebook-cell:?execution_count=44&line=14)     return fun(x, dx)

    [... skipping hidden 12 frame]

Cell In[44], [line 6](vscode-notebook-cell:?execution_count=44&line=6)
      [4](vscode-notebook-cell:?execution_count=44&line=4) @functools.partial(jax.jit, static_argnums=1)
      [5](vscode-notebook-cell:?execution_count=44&line=5) def fun(x, dx=0):
----> [6](vscode-notebook-cell:?execution_count=44&line=6)     if dx == 0:
      [7](vscode-notebook-cell:?execution_count=44&line=7)         return x
      [8](vscode-notebook-cell:?execution_count=44&line=8)     else:

    [... skipping hidden 1 frame]

File [~/miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/core.py:1492](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/core.py:1492), in concretization_function_error.<locals>.error(self, arg)
   [1491](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/core.py:1491) def error(self, arg):
-> [1492](https://file+.vscode-resource.vscode-cdn.net/home/yigit/Codes/DESC/DEBUGGING/~/miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/core.py:1492)   raise TracerBoolConversionError(arg)

TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[]..
The error occurred while tracing the function fun at [/tmp/ipykernel_31334/1112096422.py:4](https://file+.vscode-resource.vscode-cdn.net/tmp/ipykernel_31334/1112096422.py:4) for jit. This value became a tracer due to JAX operations on these lines:

  operation a:bool[] = eq b c
    from line [/tmp/ipykernel_31334/1112096422.py:6:7](https://file+.vscode-resource.vscode-cdn.net/tmp/ipykernel_31334/1112096422.py:6:7) (fun)
See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.TracerBoolConversionError
```
</details>

f0uriest:
Yes, if you jit a function it basically ignores any other jit decorators that are inside that function. So you need to specify dr as static at the outermost level. But in our case its fine since the values of dr are always called statically.

The `r**m` thing might be related to this?: https://github.com/PlasmaControl/DESC/pull/961#discussion_r1542224500

YigitElma:
Hmm, I checked the branch with the old dependencies that are changed with PR #883. The tests pass as expected. I think either JAX, numpy or scipy changed something, which causes the issue. I only changed the current requirements to
```
jax[cpu] >= 0.3.2, <= 0.4.14
numpy >= 1.20.0, < 1.25.0
scipy >= 1.5.0, < 1.11.0
```
and it works. I can look into it in detail.

And also during the process, I realized that installing jax[cpu]==0.4.14 on Python 3.12 causes the issue even without the different dependency versions mentioned in #1004

YigitElma:
> Yes, if you jit a function it basically ignores any other jit decorators that are inside that function. So you need to specify dr as static at the outermost level. But in our case its fine since the values of dr are always called statically.
> 
> The `r**m` thing might be related to this?: [#961 (comment)](https://github.com/PlasmaControl/DESC/pull/961#discussion_r1542224500)

Could this be related to this change? https://github.com/google/jax/commit/69ad4df9a53ee6bc8d7e7f9cf589562cbffc2bd6 Apparently, this is implemented right after 0.4.14

dpanici:
How does this affect say an eq solve on cpu and GPU? just want to remember what the difference was

YigitElma:
Also, a reminder for myself. I tested running different derivative orders multiple times for the version in this PR. Since it has a static argument, for every different value of that argument, JAX has to jit the function again. For the previous version, all derivative compilations happen at the same time, so it doesn't require multiple compilations. My concern was if the JAX replaces the jitted version of the function with the new jitted one with a different static argument and hence compiles the function every time we change the argument. Fortunately, the answer is no. It compiles the function stores every version in the cache and calls the one with a matching static argument. I don't know the impact of this on the memory consumption but it is worth keeping in mind. YES, the static_argnum flag is cool, but it has an effect.

dpanici (approved):

PR #1000, opened 2024-04-18
Add ability to define a function parameteriztaion for a subclass of an already existing function parameterization

dpanici:
If we want to add a new compute function for a parameterization which is a subclass of a parametrization with an existing function (for example, if we had `length` use the `Curve` as its paramaterization then have a separate function for `length` with `SplineXYZCurve` as its parameterization), we can't simply just make one function with the superclass as a parameterization then one function with the one subclass as its parameterization.

This PR changes the logic to allow this, with the caveat that the subclass function must appear AFTER the superclass function (so if in the same file, it must appear after, and if they are in different files, the file where the superclass function is defined should be imported in `compute/__init__.py` before the file where the subclass function is defined)

I would like this change for a couple reasons
- it is intuitive and follows the superclass/subclass method behavior we are used to for Python classes
- It can reduce the amount of function definitions needed in `compute`, and allow us to subclass a class and make a new compute parameterization for it without needing to re-define every compute function  or needing to make a subclass for the original class. (as an example, if I subclassed `Equilibrium` with some thing else which requires a different kind of basis vector but otherwise can use all the downstream function definitions for the metrics etc, I would not be able to just add it currently, but with this change we can do so with minimal changes)

This is related to Xu trying to have a branch of DESC where both straight and toroidal equilibria can be computed, this would make that change easier


f0uriest:
#1000!

desc/compute/data_index.py
```diff
@@ -148,10 +148,30 @@ def _decorator(func):
...
+                        defined_from_superclass = False
+                    if (
+                        name in data_index[base_class]
+                        and not data_index[base_class][name]["defined_from_superclass"]
+                    ):
+                        # only throw an error if we are trying to redefine the same thing twice
+                        # but do not throw error if the definition was through the superclass,
+                        # as this means we defined a second compute function for the subclass
+                        # specifically, so we want to allow that to overwrite the previous
+                        # definition.
```
f0uriest, line 166: Another way to do this would be to add `parameterization` to `d` above (to keep track of the parameterization that defines the method), then if you get a second definition of a given function you check `_class_inheritance` and prefer subclasses over superclasses.

something like 
```python
if p in superclasses or p == base_class:
    if name in data_index[p] and superclass in data_index[p]['parameterization']:
        # overriding parent class
        data_index[p][name] = d
    else:
        raise ValueError(...)
```

dpanici: I believe it but the logic is not coming to me in the couple minutes I spent trying this way. If someone else wants to refactor it then go ahead

desc/compute/data_index.py
```diff
@@ -148,10 +148,30 @@ def _decorator(func):
...
             for base_class, superclasses in _class_inheritance.items():
                 if p in superclasses or p == base_class:
-                    if name in data_index[base_class]:
+                    if p != base_class:
+                        # track if the compute definition was from using
+                        # a superclass
+                        defined_from_superclass = True
+                    else:
+                        #
+                        defined_from_superclass = False
```
ddudt, line 157: can combine these 7 lines into a single line:
```
defined_from_superclass = p != base_class
```

desc/compute/data_index.py
ddudt: This idea seems better if we can get it working. Would that avoid the issue with the order the functions are defined in?

f0uriest (approved):

ddudt (approved):

desc/compute/_curve.py
```diff
@@ -1061,9 +1061,7 @@ def _torsion(params, transforms, profiles, data, **kwargs):
     coordinates="",
     data=["ds", "x_s"],
     parameterization=[
-        "desc.geometry.curve.FourierRZCurve",
-        "desc.geometry.curve.FourierXYZCurve",
-        "desc.geometry.curve.FourierPlanarCurve",
+        "desc.geometry.core.Curve",
```
ddudt, line 1064: I'm not going to make you change it, but the unnecessary comma really bothers me. Why use 3 line of code when it could fit on a single line?!

PR #1002, opened 2024-04-19
Fix JAX error with `FourierCurrentPotentialField` in Flux objectives

dpanici:
On master, an optimization with a `FourierCurrentPotentialField` and the `QuadraticFlux` objective would fail. This happens because in QuadraticFlux.compute, field.compute_magnetic_field is called. If the field needs transforms to evaluate, then these transforms will be created on the fly if they are not provided, resulting in an error.

This PR fixes that by adding `jitable=True` to the `.compute` call inside `.compute_magnetic_field`

It also adds `transform` as an argument to `MagneticField.compute_magnetic_field` methods, in preparation for when #1079 is resolved and objectives can pre-compute and pass in transform objects for all magnetic fields easily.

This PR kicks the can down the road of making `get_transforms` work with `MagneticField.compute_magnetic_field` objects (to allow us to pre-compute the transforms used for magnetic field computation #1079 ) and instead opts for the simple fix.


desc/objectives/_coils.py
```diff
@@ -753,13 +753,28 @@ def build(self, use_jit=True, verbose=1):
             Bplasma = compute_B_plasma(
                 eq, eval_grid, self._source_grid, normal_only=True
             )
+        field = self._field
+        if hasattr(field, "Phi_mn"):
```
f0uriest, line 757: would it be better to get this all to work with the existing `desc.compute.utils.get_transforms` helper? That would work for coils etc as well.

f0uriest (changes requested):
I think the reason coils worked while the current potential doesnt is that by default the coil classes create jitable transforms in their compute method, since they're only 1d its not really worth it to try to do the fft. 

I'd recommend adding the new classes to the existing `get_transforms` logic, I think most of it should "just work" assuming attributes are named correctly. (might need some logic somewhere for tree-like coilsets etc, but do-able.

dpanici:
> I think the reason coils worked while the current potential doesnt is that by default the coil classes create jitable transforms in their compute method, since they're only 1d its not really worth it to try to do the fft.
> 
> I'd recommend adding the new classes to the existing `get_transforms` logic, I think most of it should "just work" assuming attributes are named correctly. (might need some logic somewhere for tree-like coilsets etc, but do-able.

The problem is less the classes and more that not every magnetic field class needs the same keys to calculate the magnetic field. Coils need "I" and "x", the current potential fields need "K", toroidal fields dont need any keys as they just have their own compute magnetic field function.

dpanici:
add case for coils and 1D transforms

dpanici:
@f0uriest I don't exactly remember the issue with the current method/how putting the logic in `get_transforms` would improve it. 

We still would need to call `get_transforms` with the correct grid based off of what the underlying object is (which would still require a check in the build method to see if it is a coil or a field), and to ask for the correct keys ("x" and maybe "K" as well if it is a current potential field). 

I think I also found in this PR a fix to something causing a JAX error in a few places when current potential fields are being optimized, so if possible I'd like to get this one in. Unless there is a better case for putting more logic into get_transforms, which I think should wait until we make "B" a data index quantity for MagneticField objects and then can go through that route to get all the proper transforms etc based off of the specific MagneticField parameterization

desc/coils.py
```diff
@@ -148,7 +148,7 @@ def current(self, new):
         self._current = float(np.squeeze(new))
 
     def compute_magnetic_field(
-        self, coords, params=None, basis="rpz", source_grid=None
+        self, coords, params=None, basis="rpz", source_grid=None, transforms=None
```
kianorr, line 151: what's actually needed in transforms to compute the magnetic field?

desc/objectives/_coils.py
```diff
@@ -854,13 +854,28 @@ def build(self, use_jit=True, verbose=1):
...
+                self._field_grid = LinearGrid(
+                    M=30 + 2 * max(field.M, field.M_Phi),
+                    N=30 + 2 * max(field.N, field.N_Phi),
+                    NFP=field.NFP,
+                )
+            field_transforms = get_transforms(
+                ["K", "x"], obj=field, grid=self._field_grid
+            )
+        else:
+            field_transforms = None
```
kianorr, line 870: I don't think I understand this. Why is `field_transforms = None` when the field grid is `None`?

desc/objectives/_coils.py
```diff
@@ -1057,13 +1076,30 @@ def build(self, use_jit=True, verbose=1):
 
         plasma_coords = jnp.array([data["R"], data["phi"], data["Z"]]).T
 
+        field = self._field
+
+        if hasattr(field, "Phi_mn"):
+            # make the transform for the FourierCurrentPotentialField
+            if self._field_grid is None:
+                self._field_grid = LinearGrid(
+                    M=30 + 2 * max(field.M, field.M_Phi),
```
kianorr, line 1085: I guess the 30 is arbitrary and you want the larger of the two resolutions? What's the difference between `field.M` and `field.M_phi`?

tests/test_optimizer.py
```diff
@@ -1312,3 +1316,55 @@ def test_LinearConstraint_jacobian():
...
+    constraints = FixParameters(field, {"I": True, "G": True})
+    opt = Optimizer("lsq-exact")
+    # this should run without an error
+    (field_modular_opt,), result = opt.optimize(
+        field, objective=obj, constraints=constraints, maxiter=1, copy=True
+    )
+
+
+@pytest.mark.unit
+def test_tor_flux_with_surface_current_field():
```
kianorr, line 1351: Could these tests be one test with an inner `test()` function or is it better to keep them separate?

desc/coils.py
dpanici: If the field is `CurrentPotentialField`  or `FourierCurrentPotentialField`, then it needs geometry info of the surface so R and Z transforms are needed, and if it is FourierCurrentPotentialField it also needs Phi transform (bc Phi is itself a DoubleFourierSeries

desc/objectives/_coils.py
dpanici: Other field types (well other then CurrentPotentialField) don't need these pre-computed specially (well I guess a curve could but those are fine being created on the fly)

desc/objectives/_coils.py
dpanici: `field.M` is the `FourierRZToroidalSurface` resolution while `M_Phi` is the resolution of the current potential descriving the surface current, yea we have `30` as a base because even if the field and surface are simple, we need a certain baseline amt of resolution to get an accurate biot savart integral over the source

desc/coils.py
f0uriest, line 151: should add transforms to the docstrings for these

f0uriest: could also pass in transforms to `self.compute` I think?

desc/magnetic_fields/_current_potential.py
```diff
@@ -644,7 +647,23 @@ def _compute_magnetic_field_from_CurrentPotentialField(
     # compute surface current, and store grid quantities
     # needed for integration in class
     # TODO: does this have to be xyz, or can it be computed in rpz as well?
-    data = field.compute(["K", "x"], grid=source_grid, basis="xyz", params=params)
+    if not params and not transforms:
```
f0uriest, line 650: why this conditional?

dpanici:
Equinox for runtime error checking in jitted functions

dpanici:
@f0uriest check this again for placing the code in a util function

desc/magnetic_fields/_current_potential.py
dpanici: To avoid calling `field.compute` inside a compute function of an objective, as there is some logic there that `jit` does not like

dpanici:
@f0uriest  just noting that the test added in this PR does fail on `master` so #1069 did not fix it

f0uriest:
shouldn't we just need to add `jitable=True` somewhere?

dpanici:
> shouldn't we just need to add `jitable=True` somewhere?

#1092 Thanks Rory that seemed to fix it...

dpanici:
@f0uriest I think with this PR I will remove the hcanges I made in the objective build methods, leave the new transform kwargs I put in all the fields, and have that be the only change here. Then we can have a separate PR when we decide how to properly handle transforms for B with get_transforms

rahulgaur104 (dismissed):

rahulgaur104:
Merge after increasing the coverage.

kianorr (approved):

tests/test_optimizer.py
```diff
@@ -1318,3 +1321,31 @@ def test_LinearConstraint_jacobian():
     np.testing.assert_allclose(vjp_unscaled, vjp1, rtol=1e-12, atol=1e-12)
     np.testing.assert_allclose(vjp_unscaled, vjp2, rtol=1e-12, atol=1e-12)
     np.testing.assert_allclose(vjp_unscaled, vjp3, rtol=1e-12, atol=1e-12)
+
+
+@pytest.mark.unit
+def test_quad_flux_with_surface_current_field():
+    """Test that QuadraticFlux does not throw an error when field has transforms."""
+    # this happens because in QuadraticFlux.compute, field.compute_magnetic_field
```
f0uriest, line 1329: Reference GH issue here so we don't forget

tests/test_optimizer.py
```diff
@@ -1318,3 +1321,31 @@ def test_LinearConstraint_jacobian():
     np.testing.assert_allclose(vjp_unscaled, vjp1, rtol=1e-12, atol=1e-12)
     np.testing.assert_allclose(vjp_unscaled, vjp2, rtol=1e-12, atol=1e-12)
     np.testing.assert_allclose(vjp_unscaled, vjp3, rtol=1e-12, atol=1e-12)
+
+
+@pytest.mark.unit
+def test_quad_flux_with_surface_current_field():
```
f0uriest, line 1327: Is the plan to remove/combine this test with #1025 ?

dpanici: Not sure, as that one will add coil tests while this one is using surface current fields. If one test on that branch involves a sum magnetic field with coils and a surface current field then we can remove this in favor of that one

f0uriest (approved):

desc/objectives/_coils.py
```diff
@@ -1272,7 +1272,10 @@ def compute(self, field_params, constants=None):
 
         # B_ext is not pre-computed because field is not fixed
         B_ext = constants["field"].compute_magnetic_field(
-            x, source_grid=constants["field_grid"], basis="rpz", params=field_params
```
ddudt, line 1275: No real change to this file, right? You just like to make the code longer to bother me?

ddudt (approved):

desc/objectives/_coils.py
dpanici: I blame my IDE

dpanici: but yea no change here just formatting

PR #1004, opened 2024-04-21
Update requirements and fix conda install

unalmis:
There's still a dependency conflict when I add `quadax` which could be useful for effective ripple stuff.

### DIFF
--- a/devtools/dev-requirements_conda.yml
+++ b/devtools/dev-requirements_conda.yml
@@ -15,10 +15,9 @@ dependencies:
   - pip:
     # Conda only parses a single list of pip requirements.
     # If two pip lists are given, all but the last list is skipped.
-    - interpax >= 0.3.1
+    - interpax
     - jax[cpu] >= 0.3.2, < 0.5.0
     - nvgpu
-    - orthax
     - plotly >= 5.16, < 6.0
     - pylatexenc >= 2.0, < 3.0
     # building the docs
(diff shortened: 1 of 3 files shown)


f0uriest:
What is the conflict?

unalmis:
Sometimes the conflict message is uninformative, but at one point it gave
```
```

f0uriest:
Hmm yeah there should be compatible versions in that range. 

But also I should just update the dependents for quadax, they're pretty out of date right now

YigitElma:
I couldn't install `quadax` either. But if you change the conda environment creation to `conda create --name desc-env 'python>=3.9, <3.12'` instead of `conda create --name desc-env 'python>=3.9, <=3.12'`, it somehow works. I checked `quadax` dependencies, JAX related stuff is exactly the same, however the error message states `<=4.14.0`. I tried to clear pip cache with `pip cache purge`, no help.  Only method that worked was to use Python 3.11 instead of 3.12

dpanici:
@f0uriest I think you just need to do a new release so that the requirements get updated on `quadax`

dpanici:
Remove orthax and add to the eps eff branch where it is used

YigitElma:
Python 3.12 issue is probably related to [this](https://jax.readthedocs.io/en/latest/changelog.html#jaxlib-0-4-17-oct-3-2023), so any JAX >=0.4.17 should work. For our case, since `quadax` wanted 0.4.14, it was not compatible with Python 3.12. This is my understanding.

dpanici (approved):

PR #1005, opened 2024-04-28
Fix Tutorial `input.HELIOTRON` Solution 

dpanici:
Increases MPOL of `input.HELIOTRON` file to allow the solution to correctly capture shafranov shift. Without this increase, the auto continuation solve from the command line yields an equilibrium with barely any shafranov shift.

#1105 will investigate why this change is necessary.


ddudt (changes requested):
I think the following files also have to be updated to stay self-consistent with this change:

- input.HELIOTRON_desc_no_continuation
- input.HELIOTRON_desc_no_continuation_output.h5
- basic_equilibrium.ipynb
- advanced_equilibrium_continuation.ipynb

dpanici:
> I think the following files also have to be updated to stay self-consistent with this change:
> 
> * input.HELIOTRON_desc_no_continuation
> * input.HELIOTRON_desc_no_continuation_output.h5
> * basic_equilibrium.ipynb
> * advanced_equilibrium_continuation.ipynb

the files that need updating have ben updated. The no continuation input does not need updating as the resolution was not changed for that one.

f0uriest:
Feels really weird that going from m=8 to m=9 makes such a big difference?

ddudt:
> Feels really weird that going from m=8 to m=9 makes such a big difference?

@dpanici are you sure this PR is necessary? The notebooks on master look like they already have the correct Shafranov shift

dpanici:
> > Feels really weird that going from m=8 to m=9 makes such a big difference?
> 
> @dpanici are you sure this PR is necessary? The notebooks on master look like they already have the correct Shafranov shift

Yea I am sure. If you rerun the examples on master, you completely lose the shafranov shift. Try it for yourself and see

dpanici:
> Feels really weird that going from m=8 to m=9 makes such a big difference?

I also agree but this is the result I see, if the point of the notebook is to just use the automatic contiunation that by default is used when running on a vmec input file, it currently results in an equilibrium with barely any shafranov shift on master after running `python -m desc input.HELIOTRON`


image

f0uriest:
ok, im still a bit worried this is indicative of some larger issue. Can you check to see when it broke? Was it just because we changed the default m from vmec?

dpanici:
make as an issue then merge this in @dpanici

f0uriest (approved):

f0uriest:
Please clean up the pr title and description before merging

ddudt (approved):

PR #1003, opened 2024-04-21
Effective ripple ε

unalmis:
Resolves #519 . Will be replaced by #1290 soon, so users are advised to wait for that.

Issue #719, opened 2023-10-22
How to compute quantities along a fieldline


f0uriest:
There have been several applications recently that need data from the equilibrium evaluated along a fieldline (stability stuff from @rahulgaur104 , GX stuff from @pkim1818, gamma_c from @JaxCrocker)

Right now this isn't super easy to do so we have some hacky stuff for it, but it would be good to decide on a common API for this.

One possibility: 
- add an attribute to each `Grid` labelling what coordinates its using, `rtz` for regular DESC computational coordinates, `raz` for rho, alpha, zeta etc.
- When calling `desc.compute.compute` we check the grid type and do any coordinate mapping to computational coordinates there.
- quantities that use field aligned grids would need some logic in their individual compute functions to handle multiple fieldlines (ie multiple values of rho, alpha)

Issues:
- the low level `desc.compute.compute` currently expects pre-built transforms and profiles. If we're doing grid mapping inside that function we would need to add an option to just pass in the grid and build transforms from that and the equilibrium
- for grids based on alpha we would need to know iota to map it to computational coordinates which adds an extra layer of complication for fixed current equilibria

Any other thoughts? @ddudt @dpanici @unalmis

dpanici:
Source grid, eval grid, single one for each computeable, by default are the same (so i.e call compute twice if want things on same src but different eval grids, one in normal coords other in SFL along field line)

dpanici:
utilities covering common cases i.e. rtz -> raz and having routine to map back on field period etc.

unalmis:
also related: https://github.com/PlasmaControl/DESC/pull/1006#issuecomment-2091007113

PR #1006, opened 2024-04-29
Fix bug with recomputing quantities on incorrect grid

unalmis:
If you call `data = eq.compute(..., data=data)` the contract is that stuff already in `data` stays there. `master` violates this contract. At best it causes slow downs by recomputing things, at worst it computes quantities incorrectly.

### DIFF
--- a/desc/equilibrium/equilibrium.py
+++ b/desc/equilibrium/equilibrium.py
@@ -883,7 +883,13 @@ class Equilibrium(IOAble, Optimizable):
                 params=params,
                 transforms=get_transforms(dep0d, obj=self, grid=grid0d, **kwargs),
                 profiles=get_profiles(dep0d, obj=self, grid=grid0d),
-                data=None,
+                # If a dependency of something is already computed, use it
+                # instead of recomputing it on a potentially bad grid.
+                data={
+                    key: data[key]
+                    for key in data
+                    if data_index[p][key]["coordinates"] == ""
+                },
                 **kwargs,
             )
             # these should all be 0d quantities so don't need to compress/expand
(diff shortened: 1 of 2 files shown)


dpanici (commented):
Is there anyway to test this with existing code or is this too specific to the field line integral stuff?

unalmis:
If you call `data = eq.compute(..., data=data)` the contract is that stuff already in `data` stays there. `master` violates this contract. At best it causes slow downs by recomputing things, at worst it computes quantities incorrectly. Yes this is needed for #1003. It's related to but distinct from #719 .

desc/equilibrium/equilibrium.py
```diff
@@ -877,19 +877,31 @@ def compute(
...
             )
             # these should all be 0d quantities so don't need to compress/expand
             data0d = {key: val for key, val in data0d.items() if key in dep0d}
             data.update(data0d)
 
+        data0d_seed = (
+            {key: data[key] for key in data if data_index[p][key]["coordinates"] == ""}
+            if ((calc1dr or calc1dz) and override_grid)
+            else {}
+        )
```
f0uriest, line 904: isn't this only needed above?

unalmis: I think we should pass in seed data whenever we can (so if its already computed correctly and if it broadcasts with the calc1dx grids). For example, epsilon effective is function of rho and requires `R0` which is a scaler volume  quantity that would be computed wrong on a grid over a single surface.

dpanici:
update master data

f0uriest (dismissed):

f0uriest (commented):
I think changing this line: 
https://github.com/PlasmaControl/DESC/blob/rc/grids/desc/equilibrium/equilibrium.py#L844
to 
```python
        deps = list(
            set(get_data_deps(names, obj=p, has_axis=grid.axis.size) + names)
            - data.keys()
        )
```

would also fix this issue. If the user wants to compute `x` which depends on `y` which in turn depends on `z`, and they pass in `y` then we shouldn't need to compute `z` at all.

unalmis:
> I think changing this line: https://github.com/PlasmaControl/DESC/blob/rc/grids/desc/equilibrium/equilibrium.py#L844 to
> 
> ```python
>         deps = list(
>             set(get_data_deps(names, obj=p, has_axis=grid.axis.size) + names)
>             - data.keys()
>         )
> ```
> 
> would also fix this issue. If the user wants to compute `x` which depends on `y` which in turn depends on `z`, and they pass in `y` then we shouldn't need to compute `z` at all.

I added it, but subtracting out `data.keys()` as you suggest does:

If the user wants to compute `x` which depends on `y` ~~which in turn depends on `z`~~, and they pass in `y` then we shouldn't need to compute `y` at all.

The dependencies of `y` may still computed redundantly.

f0uriest:
Ahhh ok yeah. In `desc.compute.utils.compute` we do things recursively, so in that case priming `data` with just the first order dependencies will avoid computing any extra stuff. But in `eq.compute` we first get the *full* list of dependencies (beyond just first order). That could eventually be fixed but doesn't have to be now.

f0uriest (approved):

PR #1009, opened 2024-04-30
Fix map_coordinates errors for toroidal current equilibrium

### LINKED ISSUE #1008: map_coordinates errors for toroidal current equilibrium
After #967 , `map_coordinates` errors when trying to compute the rotational transform on equilibria with assigned toroidal current profiles. To reproduce replace the W7-X example in `test_map_coordinates` with `DSHAPE_current`.

unalmis:
Resolves #1008

### DIFF
--- a/desc/equilibrium/coords.py
+++ b/desc/equilibrium/coords.py
@@ -110,6 +110,7 @@ def map_coordinates(  # noqa: C901
     # do surface average to get iota once
     if "iota" in profiles and profiles["iota"] is None:
         profiles["iota"] = eq.get_profile("iota", params=params)
+        params["i_l"] = profiles["iota"].params
 
     @functools.partial(jit, static_argnums=1)
     def compute(y, basis):
--- a/tests/test_equilibrium.py
+++ b/tests/test_equilibrium.py
@@ -48,7 +48,7 @@ def test_compute_theta_coords():
 def test_map_coordinates():
     """Test root finding for (rho,theta,zeta) for common use cases."""
     # finding coordinates along a single field line
-    eq = get("DSHAPE_current")
+    eq = get("NCSX")
     eq.change_resolution(3, 3, 3, 6, 6, 6)
     n = 100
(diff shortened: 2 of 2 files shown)


tests/test_equilibrium.py
```diff
@@ -48,7 +48,7 @@ def test_compute_theta_coords():
 def test_map_coordinates():
     """Test root finding for (rho,theta,zeta) for common use cases."""
     # finding coordinates along a single field line
-    eq = get("W7-X")
+    eq = get("DSHAPE_current")
```
f0uriest, line 51: A better option would probably be to make the w7x example have a fixed current profile, since that's more realistic. I think the iota profile it has was a fit from a fixed current vmec solve

desc/equilibrium/coords.py
```diff
@@ -110,6 +110,7 @@ def periodic(x):
     # do surface average to get iota once
     if "iota" in profiles and profiles["iota"] is None:
         profiles["iota"] = eq.get_profile("iota", params=params)
+        params["i_l"] = profiles["iota"].params
```
f0uriest, line 113: can you explain how this fixes it?

unalmis:
```python
def _iota_num_current(params, transforms, profiles, data, **kwargs):
    """Current contribution to the numerator of rotational transform formula."""
    if profiles["iota"] is not None:
        # otherwise params is empty
        iota = profiles["iota"].compute(transforms["grid"], params["i_l"], dr=0)
        ..... # above function eventually calls the following line with f = params = empty list which errors
        fq = interp1d(xq, x, f, method=self._method, derivative=dr, extrap=True)
```

f0uriest (approved):

tests/test_equilibrium.py
dpanici: Or use precise_QA which has a fixed current (albeit at zero) profile?

dpanici (dismissed):

f0uriest (approved):

PR #1016, opened 2024-05-08
Fix io of MixedCoilSets

### LINKED ISSUE #1013: MixedCoilSet.save bug
There is an issue preventing a `MixedCoilSet` from saving when it contains a mix of both coil sets and individual coils. Minimal working example on `master`: 
```
from desc.coils import FourierPlanarCoil, CoilSet, MixedCoilSet

coil = FourierPlanarCoil()
coils = CoilSet.linspaced_angular(coil)
coilset = MixedCoilSet((coil, coils))

coilset.save("coilset.h5")
```
Traceback: 
```
/home/dudt/DESC/desc/io/hdf5_io.py:314: RuntimeWarning: Save attribute '_current' was not saved as it does not exist.
  warnings.warn(
  File "/home/dudt/DESC/debug.py", line 7, in <module>
    coilset.save("coilset.h5")
  File "/home/dudt/DESC/desc/io/equilibrium_io.py", line 248, in save
    writer.write_obj(self)
  File "/home/dudt/DESC/desc/io/hdf5_io.py", line 312, in write_obj
    loc.create_dataset(attr, data=data, compression=compression)
  File "/home/dudt/.local/lib/python3.10/site-packages/h5py/_hl/group.py", line 183, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
  File "/home/dudt/.local/lib/python3.10/site-packages/h5py/_hl/dataset.py", line 46, in make_new_dset
    data = array_for_new_object(data, specified_dtype=dtype)
  File "/home/dudt/.local/lib/python3.10/site-packages/h5py/_hl/base.py", line 118, in array_for_new_object
    data = np.asarray(data, order="C", dtype=as_dtype)
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part. 
```

Probably unrelated, but there is also the warning stating that the "_current" attribute is not saved. When saving a `CoilSet` this warning is also thrown, although it appears the currents are in fact saved properly.

f0uriest:
Resolves #1013


dpanici (changes requested):
Test?

ddudt:
> Test?

I can confirm this works, although it still throws the warning: 
`RuntimeWarning: Save attribute '_current' was not saved as it does not exist.`

As a test, maybe we move the `DummyMixedCoilSet` [from here](https://github.com/PlasmaControl/DESC/blob/dp/coil-coil-dist/tests/conftest.py#L272) to this PR?

dpanici:
The commit I pushed also gets rid of the runtime warning, I think this is fine as the coiilset `current` property is literally just the currents of the underlying coils, so there is really no need to save it at the CoilSet level, the underlying coils are the ones that need it to be saved.

ddudt (approved):
I moved the "Dummy" coil sets to `tests/conftest.py` so we can reference them in other tests in future PRs.

dpanici:
@ddudt it is not a real test if it is in conftest, we never check if the coilset when loaded in is equivalent to the  original coilset. Can you revert your commit? or re-add the test?

ddudt:
> @ddudt it is not a real test if it is in conftest, as then we never check if the coilset when loaded in is equivalent to the original coilset. Can you revert your commit? or re-add the test?

I think it still gets saved in conftest, and the original error was that it wasn't able to save the coilset at all, not that it was saved and loaded incorrectly. I have tests ready in other PRs that use this coilset and will effectively cover what you want.

dpanici (approved):

PR #1017, opened 2024-05-13
`FourierPlanarCurve` basis option

ddudt:
Adds an option for the `center` and `normal` vectors of `FourierPlanarCurve` to be specified in $(R,\phi,Z)$ coordinates, instead of the previous limitation to Cartesian coordinates.

### DIFF
--- a/desc/compute/_curve.py
+++ b/desc/compute/_curve.py
@@ -168,18 +168,15 @@ def _Z_Curve(params, transforms, profiles, data, **kwargs):
     description="Position vector along curve",
     dim=3,
     params=["r_n", "center", "normal", "rotmat", "shift"],
-    transforms={
-        "r": [[0, 0, 0]],
-    },
+    transforms={"r": [[0, 0, 0]], "basis": []},
     profiles=[],
     coordinates="s",
     data=["s"],
     parameterization="desc.geometry.curve.FourierPlanarCurve",
     basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
-    basis_in="{'rpz', 'xyz'}: Basis for input params vectors, Default 'xyz'",
 )
 def _x_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
-    if kwargs.get("basis_in", "xyz").lower() == "rpz":
+    if transforms["basis"].lower() == "rpz":
         center = rpz2xyz(params["center"])
         normal = rpz2xyz_vec(params["normal"], phi=params["center"][1])
     else:
@@ -212,18 +209,15 @@ def _x_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
     description="Position vector along curve, first derivative",
     dim=3,
     params=["r_n", "center", "normal", "rotmat", "shift"],
-    transforms={
-        "r": [[0, 0, 0], [0, 0, 1]],
-    },
+    transforms={"r": [[0, 0, 0], [0, 0, 1]], "basis": []},
     profiles=[],
     coordinates="s",
     data=["s"],
     parameterization="desc.geometry.curve.FourierPlanarCurve",
     basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
-    basis_in="{'rpz', 'xyz'}: Basis for input params vectors, Default 'xyz'",
 )
 def _x_s_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
-    if kwargs.get("basis_in", "xyz").lower() == "rpz":
+    if transforms["basis"].lower() == "rpz":
         center = rpz2xyz(params["center"])
         normal = rpz2xyz_vec(params["normal"], phi=params["center"][1])
     else:
@@ -265,18 +259,15 @@ def _x_s_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
     description="Position vector along curve, second derivative",
     dim=3,
     params=["r_n", "center", "normal", "rotmat", "shift"],
-    transforms={
-        "r": [[0, 0, 0], [0, 0, 1], [0, 0, 2]],
-    },
+    transforms={"r": [[0, 0, 0], [0, 0, 1], [0, 0, 2]], "basis": []},
     profiles=[],
     coordinates="s",
     data=["s"],
     parameterization="desc.geometry.curve.FourierPlanarCurve",
     basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
-    basis_in="{'rpz', 'xyz'}: Basis for input params vectors, Default 'xyz'",
 )
 def _x_ss_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
-    if kwargs.get("basis_in", "xyz").lower() == "rpz":
+    if transforms["basis"].lower() == "rpz":
         center = rpz2xyz(params["center"])
         normal = rpz2xyz_vec(params["normal"], phi=params["center"][1])
     else:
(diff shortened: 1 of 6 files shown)


dpanici (commented):
I feel like there is a better way to do this using a property of FourierPlanarCurve called basis_in and then having it be obtained through transforms in the compute method, instead of needing to redefine the compute method for this one curve.

dpanici (approved):

dpanici (changes requested):
Try this in a simple optimization (does not have to be meaningful or involve an equilibrium, just optimize the coil to change the curvature or something), @kianorr has encountered issues with JAX when a string is present in `transforms` and so I wonder if that occurs here too

dpanici:
@kianorr I added a test that I thought would trigger the JAX issue you had, but I got another error, can you help debug?
```python
    c = FourierPlanarCurve()
    objective = ObjectiveFunction(CoilLength(c, target=11))
    np.testing.assert_allclose(c.compute("length")["length"], 11, atol=1e-3)
```

has an error when the objective is built
```python
desc/optimize/optimizer.py:217: in optimize
desc/objectives/objective_funs.py:173: in build
desc/objectives/_coils.py:337: in build
desc/objectives/_coils.py:153: in build
    self._grid = tree_map(
../../miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/tree_util.py:312: in tree_map
    return treedef.unflatten(f(*xs) for xs in zip(*all_leaves))
../../miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/tree_util.py:312: in <genexpr>
    return treedef.unflatten(f(*xs) for xs in zip(*all_leaves))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = array([0., 1., 0.])

    lambda x: LinearGrid(
>       N=2 * x.N + 5, NFP=getattr(x, "NFP", 1), endpoint=False
    ),
    self.things[0],
    is_leaf=lambda x: is_single_coil(x),
E   AttributeError: 'numpy.ndarray' object has no attribute 'N'

desc/objectives/_coils.py:155: AttributeError
```

dpanici:
Ah it seems that the coil objectives cannot handle curves being passed. That is unfortunate, I will make a separate issue. The latest test I added shows the JAX issue though, sorry @ddudt seems my suggestion was a bad one as I had forgotten about this issue. The kwarg method you had originally might be the way to go

ddudt:
> Ah it seems that the coil objectives cannot handle curves being passed. That is unfortunate, I will make a separate issue. The latest test I added shows the JAX issue though, sorry @ddudt seems my suggestion was a bad one as I had forgotten about this issue. The kwarg method you had originally might be the way to go

See this test in PR #977 that uses this new planar curve parameterization in a coil optimization: https://github.com/PlasmaControl/DESC/blob/dp/coil-coil-dist/tests/test_examples.py#L1301

dpanici:
> > Ah it seems that the coil objectives cannot handle curves being passed. That is unfortunate, I will make a separate issue. The latest test I added shows the JAX issue though, sorry @ddudt seems my suggestion was a bad one as I had forgotten about this issue. The kwarg method you had originally might be the way to go
> 
> See this test in PR #977 that uses this new planar curve parameterization in a coil optimization: https://github.com/PlasmaControl/DESC/blob/dp/coil-coil-dist/tests/test_examples.py#L1301

The test I added here still fails on that PR (both when only the single coil is optimized, or if it is included in a MixedCoilSet). I think the way I suggested to do this was wrong and we might have to use your original way

ddudt:
> > > Ah it seems that the coil objectives cannot handle curves being passed. That is unfortunate, I will make a separate issue. The latest test I added shows the JAX issue though, sorry @ddudt seems my suggestion was a bad one as I had forgotten about this issue. The kwarg method you had originally might be the way to go
> > 
> > 
> > See this test in PR #977 that uses this new planar curve parameterization in a coil optimization: https://github.com/PlasmaControl/DESC/blob/dp/coil-coil-dist/tests/test_examples.py#L1301
> 
> The test I added here still fails on that PR (both when only the single coil is optimized, or if it is included in a MixedCoilSet). I think the way I suggested to do this was wrong and we might have to use your original way

But the way this `"basis"` option is implemented at the moment is equivalent to the `"method"` option for computing the length of a `SplineXYZCurve`. If a `SplineXYZCoil` works with the `CoilLength` objective, then I think what we are trying to do here should also work.

dpanici:
> > > > Ah it seems that the coil objectives cannot handle curves being passed. That is unfortunate, I will make a separate issue. The latest test I added shows the JAX issue though, sorry @ddudt seems my suggestion was a bad one as I had forgotten about this issue. The kwarg method you had originally might be the way to go
> > > 
> > > 
> > > See this test in PR #977 that uses this new planar curve parameterization in a coil optimization: https://github.com/PlasmaControl/DESC/blob/dp/coil-coil-dist/tests/test_examples.py#L1301
> > 
> > 
> > The test I added here still fails on that PR (both when only the single coil is optimized, or if it is included in a MixedCoilSet). I think the way I suggested to do this was wrong and we might have to use your original way
> 
> But the way this `"basis"` option is implemented at the moment is equivalent to the `"method"` option for computing the length of a `SplineXYZCurve`. If a `SplineXYZCoil` works with the `CoilLength` objective, then I think what we are trying to do here should also work.

`SplineXYZCoil` also do not work it seems, they throw the same JAX error, I just checked

ddudt:
> > > > > Ah it seems that the coil objectives cannot handle curves being passed. That is unfortunate, I will make a separate issue. The latest test I added shows the JAX issue though, sorry @ddudt seems my suggestion was a bad one as I had forgotten about this issue. The kwarg method you had originally might be the way to go
> > > > 
> > > > 
> > > > See this test in PR #977 that uses this new planar curve parameterization in a coil optimization: https://github.com/PlasmaControl/DESC/blob/dp/coil-coil-dist/tests/test_examples.py#L1301
> > > 
> > > 
> > > The test I added here still fails on that PR (both when only the single coil is optimized, or if it is included in a MixedCoilSet). I think the way I suggested to do this was wrong and we might have to use your original way
> > 
> > 
> > But the way this `"basis"` option is implemented at the moment is equivalent to the `"method"` option for computing the length of a `SplineXYZCurve`. If a `SplineXYZCoil` works with the `CoilLength` objective, then I think what we are trying to do here should also work.
> 
> `SplineXYZCoil` also do not work it seems, they throw the same JAX error, I just checked

@dpanici I reverted back to my original solution. 

@kianorr you can copy this solution for your spline method in #970. Also note that you can replace the new example tests in this PR when you resolve #1021.

dpanici (approved):

desc/compute/_curve.py
```diff
@@ -168,16 +168,21 @@ def _Z_Curve(params, transforms, profiles, data, **kwargs):
...
+    transforms={"r": [[0, 0, 0]]},
     profiles=[],
     coordinates="s",
     data=["s"],
     parameterization="desc.geometry.curve.FourierPlanarCurve",
     basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
+    basis_in="{'rpz', 'xyz'}: Basis for input params vectors, Default 'xyz'",
 )
 def _x_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
+    if kwargs.get("basis_in", "xyz").lower() == "rpz":
```
kianorr, line 180: How does this compute function know about kwargs["basis_in"]? Where is that happening in the code? I'm trying to replicate it in the spline branch but it's not working :( i just pushed it if either of you would like to try and help -- it's probably something simple I'm missing

ddudt: I overwrote the class's `compute` method so that it always passes in this kwarg:
https://github.com/PlasmaControl/DESC/blob/dd/rpz_curve/desc/geometry/curve.py#L736

f0uriest (approved):

desc/compute/_curve.py
kianorr: Ohh okay thanks daniel!

PR #1018, opened 2024-05-13
`FixCurrents` objective

ddudt:
New linear objective to fix the currents in a `Coil` or `CoilSet`. 

TO-DO:

- [ ] Add another linear objective to constraint the sum of coil currents within a coil set? This could be a separate PR
- [x] Tests? See comment below, this will be covered in #977

### DIFF
--- a/desc/objectives/linear_objectives.py
+++ b/desc/objectives/linear_objectives.py
@@ -2634,6 +2634,11 @@ class FixCoilCurrent(FixParameters):
         Whether target and bounds should be normalized before comparing to computed
         values. If `normalize` is `True` and the target is in physical units,
         this should also be set to True.
+    indices : ndarray or bool, optional
+        Indices of the coils within a CoilSet to fix.
+        For example, ``indices=[0, 2]`` will fix the current of the first and third
+        coils in `coil.coils`. Can only be specified if `coil` is of type `CoilSet`.
+        If True/False fixes all/none of the coil currents.
     name : str, optional
         Name of the objective function.
 
@@ -2650,11 +2655,26 @@ class FixCoilCurrent(FixParameters):
         weight=1,
         normalize=True,
         normalize_target=True,
+        indices=True,
         name="fixed coil current",
     ):
+        if hasattr(coil, "_coils"):  # CoilSet
+            if isinstance(indices, bool):
+                params = {"current": indices}
+            else:  # fix only coil currents specified by indices
+                params = [
+                    {"current": bool(k in list(indices))} for k in range(len(coil))
+                ]
+        else:  # Coil
+            errorif(
+                not isinstance(indices, bool),
+                ValueError,
+                "indices must be a bool for a single Coil (not a CoilSet)",
+            )
+            params = {"current": indices}
         super().__init__(
             thing=coil,
-            params={"current": True},
+            params=params,
             target=target,
             bounds=bounds,
             weight=weight,
--- a/desc/objectives/linear_objectives.py
+++ b/desc/objectives/linear_objectives.py
@@ -2621,16 +2621,17 @@ class FixCoilCurrent(FixParameters):
     ----------
     coil : Coil
         Coil(s) that will be optimized to satisfy the Objective.
-    target : {float, ndarray}, optional
+    target : nested list of bool, optional
         Target value(s) of the objective. Only used if bounds is None.
-        Must be broadcastable to Objective.dim_f. Default is ``target=coil.current``.
-    bounds : tuple of {float, ndarray}, optional
-        Lower and upper bounds on the objective. Overrides target.
-        Both bounds must be broadcastable to to Objective.dim_f.
+        Should have the same tree structure as coil.params.
         Default is ``target=coil.current``.
-    weight : {float, ndarray}, optional
+    bounds : tuple of dict {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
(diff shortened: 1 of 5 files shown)


f0uriest:
Maybe `FixCoilCurrent(s)` to make it more explicit and avoid confusion?

ddudt:
Similar to #1016, I already have plans to use this in coil optimization tests for #977. In an effort to not add redundant unit tests, I vote to skip testing on this PR and wait to cover the new objective in that PR.

dpanici (changes requested):
I think there still should be tests here as it is not clear to me if this can be used to fix only the currents in a subset of coils or to fix every current inside the given coil or coilset (which I think the latter is true). This does not allow the ability to, say, forgo adding the toroidal flux constraint for a vacuum coil optimization by fixing the current in a single coil and allowing the rest to vary

ddudt:
> I think there still should be tests here as it is not clear to me if this can be used to fix only the currents in a subset of coils or to fix every current inside the given coil or coilset (which I think the latter is true). This does not allow the ability to, say, forgo adding the toroidal flux constraint for a vacuum coil optimization by fixing the current in a single coil and allowing the rest to vary

Before it was always fixing all of the currents in a `CoilSet`. Now I added an `indices` option to specify fixing only a subset of the coil currents. Also added a basic test. If we want to get fancy, we could wait until PR #995 is merged and add a test optimizing for the coil currents in some coils with others fixed. But this will still also get used in an optimization test in #977.

desc/objectives/linear_objectives.py
```diff
@@ -2634,6 +2634,11 @@ class FixCoilCurrent(FixParameters):
         Whether target and bounds should be normalized before comparing to computed
         values. If `normalize` is `True` and the target is in physical units,
         this should also be set to True.
+    indices : ndarray or bool, optional
```
dpanici, line 2637: could this instead be more like the optimizable way where it is a pytree? Or could it be both? to allow for fixing currents when things are nested

desc/objectives/linear_objectives.py
```diff
@@ -2611,6 +2611,85 @@ def __init__(
...
+class FixCoilCurrent(FixParameters):
+    """Fixes current(s) in a Coil or CoilSet.
+
+    Parameters
+    ----------
+    coil : Coil
+        Coil(s) that will be optimized to satisfy the Objective.
+    target : {float, ndarray}, optional
+        Target value(s) of the objective. Only used if bounds is None.
+        Must be broadcastable to Objective.dim_f. Default is ``target=eq.Psi``.
```
f0uriest, line 2623: wrong default

tests/test_linear_objectives.py
```diff
@@ -186,7 +180,7 @@ def test_fixed_mode_solve():
     # and check that the mode stayed fix
     L = 1
     M = 1
-    eq = desc.examples.get("DSHAPE")
+    eq = get("DSHAPE")
```
f0uriest, line 183: personally I prefer the explicit `desc.examples.get`, since `get` by itself is pretty generic, and there are other methods in python with the same name

desc/objectives/linear_objectives.py
```diff
@@ -2611,6 +2611,85 @@ def __init__(
...
+    weight : {float, ndarray}, optional
+        Weighting to apply to the Objective, relative to other Objectives.
+        Must be broadcastable to to Objective.dim_f
+    normalize : bool, optional
+        Whether to compute the error in physical units or non-dimensionalize.
+    normalize_target : bool, optional
+        Whether target and bounds should be normalized before comparing to computed
+        values. If `normalize` is `True` and the target is in physical units,
+        this should also be set to True.
+    indices : nested list of bool, optional
```
f0uriest, line 2637: I think it would be good to include an example of this (can basically just use the one from the test). Also the part about `Must have the same pytree structure as coil.params_dict.` isn't quite right I think, because its a list of bools, not a list of dict of bool? Not sure the best way to describe that, so I think an example would be clearest.

f0uriest (changes requested):

desc/objectives/linear_objectives.py
```diff
@@ -2611,6 +2614,117 @@ def __init__(
...
 
 
+class FixCoilCurrent(FixParameters):
+    """Fixes current(s) in a Coil or CoilSet.
+
+    Parameters
+    ----------
+    coil : Coil
+        Coil(s) that will be optimized to satisfy the Objective.
+    target : {float, ndarray}, optional
```
dpanici, line 2624: does the target also need the same structure as the indices?

desc/objectives/linear_objectives.py
```diff
@@ -2611,6 +2614,118 @@ def __init__(
...
 
 
+class FixCoilCurrent(FixParameters):
+    """Fixes current(s) in a Coil or CoilSet.
+
+    Parameters
+    ----------
+    coil : Coil
+        Coil(s) that will be optimized to satisfy the Objective.
+    target : nested list of bool, optional
```
dpanici, line 2624: going to be annoying again, so this should not be a bool right? it should be either the desired current, or None in which case it will default to the coils current?

ddudt: Correct, thanks for catching this. `target` gets passed through to `FixParameters` so it should be the same pytree structure as `coil.params_dict`. I copied the docstring from that parent class

f0uriest (approved):

dpanici (approved):

PR #1024, opened 2024-05-20
compute quantities along a fieldline

### LINKED ISSUE #664: Error when computing Poloidal flux and Boozer toroidal stream function
There exists errors when computing ```Boozer toroidal stream function``` and ```Poloidal flux``` in the example shown below, which uses the DSHAPE_output from DESC packge.

```
import numpy as np
import desc.io
from desc.grid import Grid

#load the DSHAPE equilibrium
eq_fam = desc.io.load("DSHAPE_output.h5")

#node and node_list both contain the point (rho=0.5, theta=1.5PI, zeta=0)
node = Grid(np.array([0.5,1.5*np.pi,0]))
node_list = Grid(np.array([[0.5,1.5*np.pi,0], [0.5,1.7*np.pi,0]]))

#print their nodes to check if the points contained are correct
print('node:')
print(node.nodes)
print('node_list:')
print(node_list.nodes)

#computing the Boozer toroidal stream function and output
print('nu:')
print(eq_fam[0].compute('nu', node)['nu'])
print(eq_fam[0].compute('nu', node_list)['nu'])

#computing the Poloidal flux and output
print('chi:')
print(eq_fam[0].compute('chi', node)['chi'])
```
The output shown below hints some errors when computing  ```nu``` and ```chi```.
The ```nu``` should be the same at the same point (rho=0.5, theta=1.5PI, zeta=0), but it shows difference.
The ```chi``` can not be zero at the point (rho=0.5, theta=1.5PI, zeta=0) since rho ≠ 0.
```
node:
[[0.5        4.71238898 0.        ]]

node_list:
[[0.5        4.71238898 0.        ]
 [0.5        5.34070751 0.        ]]

nu:
[0.00400288]
[0.00665903 0.00363253]

chi:
[0.]
```

### LINKED ISSUE #692: Separate coordinate dependencies from grid dependencies for computable quantities
Related to #664, when computing `nu` (boozer stream function), it is a function of rho, theta, zeta, but really needs a grid over a whole surface to compute accurately due to needing the harmonics of B. Right now we assume that any quantity that is a function of rho,theta,zeta can be computed anywhere without need for averaging but this isn't quite the case. The main issue is that in the data index we use `coordinates` to mean both the coordinates the quantity varies in, and an indicator of whether volume or surface averages are required. A better solution would be to add a new flag to the data index to indicate quantities that need a specific grid to compute correctly.

### LINKED ISSUE #719: How to compute quantities along a fieldline

### LINKED ISSUE #904: eq.compute doesn't correctly override grids to compute quantities that require evaluating functions along field lines

### LINKED ISSUE #1077: Errors when compute things on custom grids
Pull request #985 made this change to prevent use of incorrect quadrature weights when integrating some quantities. This is good, but our current `eq.compute` isn't well-designed to handle grids without spacing, making it hard to compute stuff.

## Context
When a grid is constructed, we assign quadrature weights for surface integrals, `grid.spacing` , and volume integrals `grid.weights` . However, because we don't currently use fancy triangulation techniques to compute these quadrature weights, we can only compute them accurately if the grid has a special structure. It's unlikely for field line grids to have this structure.

When evaluating stuff along field lines, custom grids are used often, and it is tedious for the user to define `grid.spacing`. ( The first column of grid.spacing corresponds to drho (distance between rho surfaces, the second dtheta (distance between theta surface, and the third to dzeta (distance between zeta surfaces) for each node. Dtheta could be replaced with dalpha. As you may notice, assigning a distance to each coordinate between nodes is difficult if the grid is not "rectangular" (e.g. some tensor product /meshgrid of coordinates).

## Issue
Integration is attempted when it is not necessary, which will now error for almost all grids that do a coordinate mapping since they don't have `spacing` defined.

Example: after computing `a` correctly on a quadrature grid, `master` tries to compute a dependency of `a`, in particular `A(z)` on a `LinearGrid` which errors when doing field line stuff because the field line grids don't have `unique_zeta_idx` defined.

_Originally posted by @unalmis in https://github.com/PlasmaControl/DESC/pull/1024#discussion_r1653781790_

unalmis:
Resolves #719, resolves #692, resolves #664, resolves #904, resolves #1077. Needed to compute neoclassical stuff. Ready for review

Can look at the Gamma_c branch to see how these changes are used.

- According to jax, `jnp.take(x, idx, unique_indices=True)` is more efficient than `x[idx]`, so this PR makes this change in the `grid.compress()` method. (The `bounce` branch needs `jnp.take` for different reasons,  which is the actual motivation for adding `jnp.take` and its numpy equivalent to `desc.backend`. )
- Adds an optional resolution requirement string to the register compute function decorator to resolve 664.
- Adds stuff to use `eq.compute` method for things whose compute functions assume the input data is evaluated along a field line.
- Changes some grid attributes (backwards-compatible) for the grids to work better in different coordinate systems.
- Fixes a bug on `LinearGrid` with `grid.num_zeta = 1` and `grid.NFP!=1` if used for a surface integral (surface averages or others such as surface_integral(x) / surface_integral(y) were unaffected since the bug would net a constant factor difference).


dpanici:
LinearFieldLineGrid? To automatically obtain the rtz corresponding to its raz?

unalmis:
The `bounce` branch has a `create_meshgrid` method [here](https://github.com/PlasmaControl/DESC/pull/854/files#r1610773665) which along with `desc_grid_from_field_line_coords` method [here](https://github.com/PlasmaControl/DESC/pull/854/files#r1610774430) that generates an rtz grid paired to the raz source.

The `ripple` branch has some examples of how this is used in the compute functions in `desc.compute._neoclassical.py`.

I'll add the check in the compute.utils compute fun to do the coordinate mapping automatically if the supplied grid is not rtz.

desc/compute/utils.py
```diff
@@ -266,7 +307,7 @@ def get_profiles(keys, obj, grid=None, has_axis=False, jitable=False, **kwargs):
     profs = []
     for key in deps:
         profs += data_index[p][key]["dependencies"]["profiles"]
-    profs = sorted(list(set(profs)))
+    profs = sorted(set(profs))
```
unalmis, line 310: `sorted` takes an iterable and returns a list always.

desc/compute/utils.py
```diff
@@ -685,7 +726,7 @@ def _get_grid_surface(grid, surface_label):
     grid : Grid
         Collocation grid containing the nodes to evaluate at.
     surface_label : str
-        The surface label of rho, theta, or zeta.
+        The surface label of rho, poloidal, or zeta.
```
unalmis, line 729: Rest of changes in this file just make sure stuff works when different strings are given for the same surface label.

desc/equilibrium/equilibrium.py
```diff
@@ -777,7 +785,7 @@ def get_axis(self):
         axis = FourierRZCurve(R_n, Z_n, modes_R, modes_Z, NFP=self.NFP, sym=self.sym)
         return axis
 
-    def compute(
+    def compute(  # noqa: C901
```
unalmis, line 788: I can't add tests to see if `eq.compute("something that needs field line grid", grid)` works on this branch. I've merged this `ripple` branch and it works.

desc/equilibrium/equilibrium.py
```diff
@@ -964,17 +1050,16 @@ def compute(
...
+                    data1dz[key], grid1dz, surface_label="zeta"
+                )
+                for key in data1dz
+                if key in dep1dz and key not in data
             }
             data.update(data1dz)
 
-        # TODO: we can probably reduce the number of deps computed here if some are only
-        #   needed as inputs for 0d and 1d qtys, unless the user asks for them
-        #   specifically?
```
unalmis, line 977: We pass in `names` and not `deps` here, so nothing extra is computed.

tests/test_axis_limits.py
```diff
@@ -395,7 +364,8 @@ def get_source(name):
     return unsafe_names
 
 
-@pytest.mark.unit
+@pytest.mark.slow
+@pytest.mark.regression
```
unalmis, line 368: It's kind of demoralizing run `pytest -m unit`  and wait on `tests_axis_limits`  so I changed this to regression

desc/equilibrium/equilibrium.py
```diff
@@ -836,60 +888,100 @@ def compute(
...
+            return bool(data_index[p][name]["source_grid_requirement"])
+
+        need_src_deps = _grow_seeds(set(filter(need_src, deps)), deps)
+
+        dep0d = {dep for dep in deps if self.is_0d(dep) and dep not in need_src_deps}
+        # Unless user asks, don't try to recompute stuff which are only dependencies
+        # of dep0d. Example, need R0. R0 <- A <- A(z) computable on dep0d grid.
+        # But A(z) in dep1dz and attempt to recompute on dep1dz grid will error
+        # without unique zeta idx defined, which mapped coordinate grids lack.
+        dep0d_deps = set(get_data_deps(dep0d, obj=p, has_axis=grid.axis.size))
```
unalmis, line 917: @rahulgaur104 this is the issue you are running, except instead of `R0` you need `a`, and even after computing `a` correctly on a quadrature grid, `master` tries to compute a dependency of `a`, in particular `A(z)` on a `LinearGrid` which errors when doing field line stuff because the field line grids don't have `unique_zeta_idx` defined.

desc/equilibrium/coords.py
```diff
@@ -504,3 +503,65 @@ def to_sfl(
...
+    poloidal : ndarray
+        Sorted unique poloidal coordinates.
+    toroidal : ndarray
+        Sorted unique toroidal coordinates.
+    coordinates : str
+        Input coordinates that are specified by the arguments, respectively.
+        raz : rho, alpha, zeta
+        rpz : rho, theta_PEST, zeta
+        rtz : rho, theta, zeta
+    period : tuple of float
```
rahulgaur104, line 529: alpha is only 2pi periodic in theta_PEST not zeta. How does the user know which one you are talking about?
What does period in inbasis mean? In basis can have both theta and zeta.

desc/equilibrium/equilibrium.py
rahulgaur104: Okay, maybe we can merge it fast so I can fix ballooning because right now, ballooning calculations fail.
@dpanici  can you review this PR this week so we can merge it quickly.

desc/equilibrium/equilibrium.py
```diff
@@ -836,60 +888,100 @@ def compute(
...
+        # This filter is stronger than the name implies, but the false positives
+        # will still be computed correctly with the logic in compute.utils.compute.
+        just_dep0d_dep = lambda name: name in dep0d_deps and name not in names
+        dep1dr = {
             dep
             for dep in deps
-            if (data_index[p][dep]["coordinates"] == "r") and (dep not in data)
-        ]
-        dep1dz = [
+            if self.is_1dr(dep) and not just_dep0d_dep(dep) and dep not in need_src_deps
```
rahulgaur104, line 924: This part of gridding is always a mess to understand so I am completely ignoring it.

desc/grid.py
```diff
@@ -255,27 +320,48 @@ def inverse_rho_idx(self):
...
             AttributeError,
-            "jit compatible Grid objects do not have inverse indices assigned.",
+            f"{self} does not have inverse indices assigned. "
+            "It is not possible to do this automatically on grids made under JIT.",
         )
-        return self._inverse_theta_idx
+        return self._inverse_poloidal_idx
+
+    @property
+    def inverse_alpha_idx(self):
```
rahulgaur104, line 340: What purpose do these functions serve?

rahulgaur104 (approved):

desc/compute/utils.py
unalmis: I'm trying to make the grid stuff agnostic to the poloidal label, since grids may be composed of rho, alpha, zeta coordinates instead of rho, theta, zeta. I don't want to write `grid.num_theta` when I really mean `grid.num_alpha`, and likewise it gets confusing to write logic like `integrate over theta` when I may want to do `integrate over alpha`.

desc/backend.py
```diff
@@ -744,3 +753,37 @@ def root(
...
+        a,
+        indices,
+        axis=None,
+        out=None,
+        mode="fill",
+        unique_indices=False,
+        indices_are_sorted=False,
+        fill_value=None,
+    ):
+        """A numpy implementation of jnp.take."""
```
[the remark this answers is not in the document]
unalmis: If the user has `jax` installed, then the function `take` is an alias for `jnp.take`. If the user only has numpy installed, then `take` is aliased to this function which is just `np.take` generalized to support the API of `jnp.take`. In particular `jnp.take` supports additional arguments `unique_indices` and `indices_are_sorted` to execute more efficiently on certain hardware and an additional `mode="fill"` option which is often necessary in jitable algorithms due to JAX's dynamic size array limitations.

If we didn't do this, DESC would error for users without JAX installed. i think the current policy is to keep it compatible with numpy in case a jax is abandoned for a new ad library

desc/compute/utils.py
```diff
@@ -163,30 +175,65 @@ def get_data_deps(keys, obj, has_axis=False):
...
-
     out = []
     for key in keys:
-        out += _get_deps_1_key(key)
-    return sorted(list(set(out)))
+        out += _get_deps_1_key(key, p, has_axis)
+    return sorted(set(out))
+
+
+def _get_deps_1_key(key, p, has_axis):
```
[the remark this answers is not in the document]
unalmis: i don't want to since it's private

desc/compute/utils.py
```diff
@@ -163,30 +175,65 @@ def get_data_deps(keys, obj, has_axis=False):
...
+        for dep in axis_limit_deps:
+            out += _get_deps_1_key(dep, p, has_axis)
+
+    return sorted(set(out))
+
+
+def _grow_seeds(
+    seeds, search_space, p="desc.equilibrium.equilibrium.Equilibrium", has_axis=False
+):
+    """Traverse the dependency DAG for keys in search space dependent on seeds.
```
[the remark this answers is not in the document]
unalmis: [directed acycdiilic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph)

desc/equilibrium/coords.py
unalmis: I just copied the docstring from the `def map_coordinates` method in this file. I think it's more like this. 
```python
# want to make grid with three (long) field lines on each flux surface
eq = get("W7-X")
rho = [0, 0.5, 1]
alpha = [0, pi, 3pi/2]
zeta = np.linspace(0, 100000000, 1000000000).
grid = rtz_grid(eq, rho, alpha, zeta, coordinates="raz", period=(np.inf, 2 * np.pi, np.inf))
```
By specifying 2pi, you are telling the code that `alpha % (2pi) == alpha`.

desc/equilibrium/equilibrium.py
unalmis: understandable

desc/grid.py
unalmis: It's like inverse_rho_idx but for the poloidal coordinate. We usually use them to upsample back to the full grid after an integration so that we can compute things like `data["<|B|>"] * data["sqrt(g)"]` without broadcasting errors.

desc/equilibrium/coords.py
```diff
@@ -305,7 +304,7 @@ def fixup(x, *args):
     )
     theta_DESC, (res, niter) = vecroot(theta_star, theta_star, rho, zeta)
 
-    nodes = jnp.array([rho, theta_DESC.squeeze(), zeta]).T
+    nodes = jnp.array([rho, jnp.atleast_1d(theta_DESC.squeeze()), zeta]).T
```
unalmis, line 307: this would give errors on master if theta desc has size 1

desc/equilibrium/equilibrium.py
```diff
@@ -836,60 +888,100 @@ def compute(
...
-        calc1dz = bool(len(dep1dz))
-        if (  # see if the grid we're already using will work for desired qtys
-            (grid.L >= self.L_grid)
-            and (grid.M >= self.M_grid)
-            and (grid.N >= self.N_grid)
-        ):
+            if self.is_1dz(dep) and not just_dep0d_dep(dep) and dep not in need_src_deps
+            # These don't need a special grid, since the transforms are always
+            # built on the (rho, theta, zeta) coordinate grid.
+            and dep not in ["phi", "zeta"]
```
unalmis, line 932: think Dario added this way back. don't quite see the purpose but i left it

tests/test_axis_limits.py
```diff
@@ -220,7 +190,7 @@ def assert_is_continuous(
 
     num_points = 12
     rho = np.linspace(start=0, stop=delta, num=num_points)
-    grid = LinearGrid(rho=rho, M=5, N=5, NFP=eq.NFP, sym=eq.sym)
+    grid = LinearGrid(rho=rho, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=eq.sym)
```
unalmis, line 193: think we changed the resolution below so this is actually less resolution / the same

tests/test_compute_utils.py
```diff
@@ -69,7 +69,17 @@ def surface_integrals(grid, q=np.array([1.0]), surface_label="rho"):
             Surface integral of the input over each surface in the grid.
 
         """
-        _, _, spacing, has_endpoint_dupe = _get_grid_surface(grid, surface_label)
+        _, _, spacing, _, _ = _get_grid_surface(grid, grid.get_label(surface_label))
```
unalmis, line 72: this uses same logic as master, or what used to be in compute.utils to make sure i didn't make typos when relabling theta to poloidal in compute.utils

desc/grid.py
```diff
@@ -804,64 +1002,54 @@ def _create_nodes(  # noqa: C901
...
             if self.sym:
-                # cut domain to relevant subdomain: delete theta > pi nodes
+                # cut domain to relevant subdomain: delete theta > π nodes
                 t = t[: np.searchsorted(t, np.pi, side="right")]
-            dt = np.zeros_like(t)
             if t.size > 1:
-                # choose dt to be the cyclic distance of the surrounding two nodes
-                dt[1:-1] = t[2:] - t[:-2]
                 if not self.sym:
-                    dt[0] = t[1] + (THETA_ENDPOINT - t[-1]) % THETA_ENDPOINT
```
unalmis, line 818: this is all just moving logic to function `_periodic_spacing()` to not have repeated code

desc/equilibrium/equilibrium.py
unalmis:
```sh
# for security make new local branch that is copy of branch you want to merge
git switch branch_i_want_to_merge_this_PR
git checkout -b new_branch
git merge --squash fieldline_compute
```
The squash merge will merge the branch as one commit.  May be useful if you would like to unmerge with
```sh
git stash
```

rahulgaur104: Well there is another problem now. There is a change in this PR that allows the evaluation to work but throws an error when I optimize with the BallooningObjective. Note, that the current ballooning PR (not merged with this PR) works fine for optimization but throws an error during evaluation. 
So the error during optimization is definitely coming from this PR. You can find more details here: https://github.com/PlasmaControl/DESC/blob/370168ef6aa67934b645a472193fa1478a916f5d/docs/notebooks/tutorials/ideal_ballooning_stability.ipynb

unalmis: I pulled your fork `rg/adjoint_ballooning2` and ran the optimization under the cell block "Optimizing for ballooning stability". I was unable to reproduce the objective build error you linked. My machine is fedora linux with jax 0.4.28 on cpu.

This PR won't affect optimization; we don't call `eq.compute` in objectives. If your bug is reproducible, perhaps some other pull request since May 8th has affected numerics. Check #1022?

rahulgaur104: Yeah, I think my jax version was 0.4.23 which causes it to throw that that error. Updating to jax 0.4.28 gets rid of the error.

desc/equilibrium/equilibrium.py
```diff
@@ -776,7 +784,29 @@ def get_axis(self):
         axis = FourierRZCurve(R_n, Z_n, modes_R, modes_Z, NFP=self.NFP, sym=self.sym)
         return axis
 
-    def compute(
+    @staticmethod
+    def is_0d(name):
```
f0uriest, line 788: do these need to be methods? Can't it just be a util/local/inline function?

desc/grid.py
```diff
@@ -34,12 +34,16 @@ class _Grid(IOAble, ABC):
...
+        "_coordinates",
+        "_period",
+        "_source_grid",
         "_unique_rho_idx",
-        "_unique_theta_idx",
+        "_unique_poloidal_idx",
         "_unique_zeta_idx",
         "_inverse_rho_idx",
-        "_inverse_theta_idx",
+        "_inverse_poloidal_idx",
```
f0uriest, line 44: should add logic to `_set_up` to handle these new/renamed attributes when loading old solutions

desc/grid.py
```diff
@@ -199,6 +203,28 @@ def sym(self):
         """bool: True for stellarator symmetry, False otherwise."""
         return self.__dict__.setdefault("_sym", False)
 
+    @property
+    def is_meshgrid(self):
```
f0uriest, line 207: "meshgrid" is a matlab/numpy term. I think the more mathy term would be "tensor product grid" (but im fine with either)

desc/grid.py
```diff
@@ -508,23 +579,133 @@ class Grid(_Grid):
         Spacing between nodes in each direction.
     weights : ndarray of float, size(num_nodes, )
         Quadrature weights for each node.
+    coordinates : str
+        Coordinates that are specified by the nodes.
+        raz : rho, alpha, zeta
+        rpz : rho, theta_PEST, zeta
```
f0uriest, line 585: see previous note about possible confusion with R,phi,Z

desc/grid.py
```diff
@@ -508,23 +579,133 @@ class Grid(_Grid):
...
+        nodes.reshape(num_radial, num_poloidal, num_toroidal, 3).
     jitable : bool
         Whether to skip certain checks and conditionals that don't work under jit.
         Allows grid to be created on the fly with custom nodes, but weights, symmetry
-        etc may be wrong if grid contains duplicate nodes.
+        etc. may be wrong if grid contains duplicate nodes.
     """
 
+    @classmethod
+    def create_meshgrid(
```
f0uriest, line 607: move this after `__init__`

desc/compute/data_index.py
```diff
@@ -100,8 +101,12 @@ def register_compute_fun(
     parameterization : str or list of str
         Name of desc types the method is valid for. eg `'desc.geometry.FourierXYZCurve'`
         or `'desc.equilibrium.Equilibrium'`.
-    grid_type : str
-        Name of grid type the quantity must be computed with. eg `'quad'`.
+    resolution_requirement : str
+        Resolution requirements in coordinates. I.e. "r" expects radial resolution
+        in the grid, "rtz" expects grid to radial, poloidal, and toroidal resolution.
+    source_grid_requirement : dict
```
f0uriest, line 107: might want a bit more explanation here. IE, the source grid is generally not the grid things are being computed on, right? since all of the regular grid types don't have a source grid by default.

desc/grid.py
```diff
@@ -508,23 +579,133 @@ class Grid(_Grid):
...
+        spacing=None,
+        coordinates="rtz",
+        period=(np.inf, 2 * np.pi, 2 * np.pi),
+        **kwargs,
+    ):
+        """Create a meshgrid from the given coordinates in a jitable manner.
+
+        Parameters
+        ----------
+        nodes : Array, Array, Array
```
f0uriest, line 619: `nodes : iterable of [Array, Array, Array]`

desc/grid.py
```diff
@@ -508,23 +579,133 @@ class Grid(_Grid):
...
+            Sorted unique values of each coordinate.
+        spacing : Array, Array, Array
+            Weights for integration. Defaults to a midpoint rule.
+        coordinates : str
+            Coordinates that are specified by the nodes a, b, c, respectively.
```
f0uriest, line 624: what are a, b, c?

desc/grid.py
f0uriest: also might be better as separate args?

desc/grid.py
```diff
@@ -508,23 +579,133 @@ class Grid(_Grid):
...
+        period=(np.inf, 2 * np.pi, 2 * np.pi),
+        **kwargs,
+    ):
+        """Create a meshgrid from the given coordinates in a jitable manner.
+
+        Parameters
+        ----------
+        nodes : Array, Array, Array
+            Sorted unique values of each coordinate.
+        spacing : Array, Array, Array
```
f0uriest, line 621: see above

desc/grid.py
```diff
@@ -508,23 +579,133 @@ class Grid(_Grid):
...
+            Weights for integration. Defaults to a midpoint rule.
+        coordinates : str
+            Coordinates that are specified by the nodes a, b, c, respectively.
+            raz : rho, alpha, zeta
+            rpz : rho, theta_PEST, zeta
```
f0uriest, line 626: see previous note about possible confusion

desc/grid.py
```diff
@@ -508,23 +579,133 @@ class Grid(_Grid):
...
+            dc = _periodic_spacing(c, period[2])[1]
+        else:
+            da, db, dc = spacing
+        nodes = jnp.column_stack(
+            list(map(jnp.ravel, jnp.meshgrid(a, b, c, indexing="ij")))
+        )
+        spacing = jnp.column_stack(
+            list(map(jnp.ravel, jnp.meshgrid(da, db, dc, indexing="ij")))
+        )
+        weights = spacing.prod(axis=1) if np.prod(period) == 4 * jnp.pi**2 else None
```
f0uriest, line 652: can you explain this a bit?

desc/grid.py
```diff
@@ -1617,3 +1787,65 @@ def find_least_rational_surfaces(
     io = find_most_distant(io_rat, n, a, b, tol=atol, **kwargs)
     rho = _find_rho(iota, io, tol=atol)
     return rho, io
+
+
+def _periodic_spacing(x, period=2 * jnp.pi, sort=False, jnp=jnp):
```
f0uriest, line 1792: any reason for making `jnp` an argument?

desc/grid.py
```diff
@@ -1617,3 +1787,65 @@ def find_least_rational_surfaces(
...
+    """Compute dx between points in x assuming periodicity.
+
+    Parameters
+    ----------
+    x : Array
+        Points, assumed sorted in the cyclic domain [0, period], unless
+        specified otherwise.
+    period : float
+        Number such that x + period = x.
```
f0uriest, line 1801: i think you mean `f(x+period) = f(x)`

desc/grid.py
```diff
@@ -664,14 +856,22 @@ def __init__(
         rho=np.array(1.0),
         theta=np.array(0.0),
         zeta=np.array(0.0),
+        coordinates="rtz",
```
f0uriest, line 859: do we want this to be an option? I feel like for all our regular grids it might be better to hard-code the coordinates and period?

desc/equilibrium/equilibrium.py
f0uriest: can you explain the `just_dep0d_dep` part?

desc/equilibrium/equilibrium.py
```diff
@@ -836,60 +888,100 @@ def compute(
...
             if isinstance(grid, LinearGrid):
                 calc1dr = calc1dz = False
+        else:
+            # Warn if best way to compute accurately is increasing resolution.
+            for dep in deps:
+                req = data_index[p][dep]["resolution_requirement"]
+                coords = data_index[p][dep]["coordinates"]
+                msg = lambda direction: colored(
+                    f"Dependency {dep} may require more {direction}"
+                    f" resolution to compute.",
```
f0uriest, line 952: "to compute accurately"

desc/equilibrium/equilibrium.py
```diff
@@ -836,60 +888,100 @@ def compute(
...
+                    "yellow",
+                )
+                warnif(
+                    "r" in req and "r" in coords and grid.L < self.L_grid,
```
f0uriest, line 956: why do we need to check both `req` and `coords`? Shouldn't `req` be enough?

f0uriest (changes requested):

desc/compute/utils.py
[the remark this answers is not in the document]
unalmis: No change, just pulled this recursive function out of it's parent since it helped to use it when coding

desc/equilibrium/coords.py
```diff
@@ -504,3 +503,65 @@ def to_sfl(
...
+    radial : ndarray
+        Sorted unique radial coordinates.
+    poloidal : ndarray
+        Sorted unique poloidal coordinates.
+    toroidal : ndarray
+        Sorted unique toroidal coordinates.
+    coordinates : str
+        Input coordinates that are specified by the arguments, respectively.
+        raz : rho, alpha, zeta
+        rpz : rho, theta_PEST, zeta
```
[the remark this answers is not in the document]
unalmis: Ok I'll change. I think right now linear grid also has an option to specify coordinates in constructor. Should I remove that? I only see the use-case for custom grids using different coordiantes than rtz.

desc/grid.py
unalmis: I should have written `period[1] * period[2] == 4 pi^2`.
I think there's two uses for this constructor.

1. jitably Make some tensor product grid over a surface using nodes for some coordinate system, (e.g. either `rho, theta, zeta`, or maybe `rho, theta_pest, zeta`, or `rho, alpha, zeta`). In this case period=(np.inf, 2pi, 2pi).
2. Make a grid following a few field-lines. E.g.  rho, alpha, zeta coordinates with zeta in [0, large number] and periwod=(np.inf, 2pi, np.inf). Here I don't assign quadrature weights for volume integration on  a field line following grid; eights would be nan in this case, since the spacing for the last node is (finite number % np.inf = nan).

desc/grid.py
unalmis: I agree i'll change

desc/equilibrium/equilibrium.py
unalmis: Suppose you want to compute `A(z)` which has coords="z". Right now `A(z)` has empty string as resolution requirement, since `coordinates"z"` implicitly implies that you need radial and poloidal resolution, but we could just as well explicitly  specify that `resolution_requirement="rt"`.

So:
```python
name="A(z)"
coordinates="z"              # output is function of toroidal coordinate only
resolution_requirement="rt"  # radial and poloidal resolution needed

footgun_grid = Grid(np.array([[0, 0, zeta]]))  # 1d grid in zeta
# common footgun error to pass in grid that has only the desired output evaluation points
data = eq.compute(name, grid=footgun_grid)
```
Since the point of the logic in `eq.compute()` is to save the user from such errors, we will override the grid and compute `A(z)` on a grid with radial resolution and (losslessly)  interpolate back to the original grid. When we get to the following chunk of code
```python
# don't need to warn because we won't use this grid to calculate anyway.
warnif(
    # True and False and True
    "r" in req and "r" in coords and footgun_grid.L < eq.L_grid,
    ResolutionWarning,
    msg("radial"),
)
warnif(
    # True and False and True
    "t" in req and "t" in coords and footgun_grid.M < eq.M_grid,
    ResolutionWarning,
    msg("poloidal"),
)
```
The additional check that `"r" in coords`  or `"t" in coords` prevents unnecessary warnings from being raised. The reason this works is because the logical negation of those checks, `not ("r" in coords or "t" in coords)` is equivalent to `coords == "z"` which is precisely the condition for when we replace `footgun_grid` with a better grid.

unalmis: Without these extra checks you'd get warnings when computing anything when the input grid is not a high resolution quadrautre grid

unalmis: And it still works for things like
```python
@register_compute_fun(
    name="chi",
    label="\\chi",
    units="Wb",
    units_long="Webers",
    description="Poloidal flux (normalized by 2pi)",
    dim=1,
    params=[],
    transforms={"grid": []},
    profiles=[],
    coordinates="r",
    data=["chi_r", "rho"],
    resolution_requirement="r",
)
def _chi(params, transforms, profiles, data, **kwargs):
    chi_r = transforms["grid"].compress(data["chi_r"])
    chi = cumtrapz(chi_r, transforms["grid"].compress(data["rho"]), initial=0)
    data["chi"] = transforms["grid"].expand(chi)
    return data
```

unalmis: Having written this out, I now think it's probably better to change:
`"r" in req and "r" in coords` to `"r" in req and (coords not in {"z", ""})`

desc/compute/data_index.py
unalmis: Yes. One way to think about it is the source grid of any `LinearGrid`, `QuadratureGrid`, `ConcentricGrid` is itself

desc/compute/utils.py
```diff
@@ -1066,8 +1104,13 @@ def surface_averages_map(grid, surface_label="rho", expand_out=True, tol=1e-14):
         ``function(q, sqrt_g)``.
 
     """
-    if not hasattr(grid, f"num_{surface_label}"):
-        expand_out = False  # don't try to expand already expanded output
+    surface_label = grid.get_label(surface_label)
+    expand_out = (
```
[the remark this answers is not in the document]
unalmis: In an older PR, jax threw some missing indices error that seemed to suggest surface integration was done on grid without unique idx defined. So we generalized these methods to work without them, with the cavet that we always expand out

desc/compute/_geometry.py
```diff
@@ -196,6 +196,7 @@ def _A_of_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwarg
             transforms["grid"],
             data["Z"] * n[:, 2] * jnp.sqrt(data["g_tt"]),
             line_label="theta",
+            # This will only work if the grid has rho=1 surface.
```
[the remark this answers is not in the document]
unalmis: No.  Take the surface in `test_compute_everything` and add rho=0.8 to the constructor. `FourierRZToroidalSurface(**elliptic_cross_section_with_torsion, rho=0.8)`. Then the grid only has rho =0.8 nodes and A(z) is computed as zero. I think the fix is to compute this like we do `perimeter(z)`; fix_surface to the rho value of the surface and then divide by that rho value to get an approximation, but I'll take a closer look tomorrow.

desc/compute/utils.py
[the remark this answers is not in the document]
f0uriest: That's fine but I think then we should request the correct value for `expand_out` on the calling side and verify that we can expand and throw an error if the caller asked for expand_out but indices aren't given. As is this seems like it might silently change whether or not it expands out based on the type of grid.

desc/equilibrium/coords.py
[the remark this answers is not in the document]
f0uriest: Yeah I think for now it's probably safe to only allow `Grid` to have different coordinates, with all the others fixed to `rtz`.

desc/grid.py
f0uriest: ok, i think it might be better to spell out that logic, or maybe just check the coordinates to determine if weights should be assigned. IE, if the grid is rtz over a single period, then `period[1] * period[2] == 4*np.pi**2/NFP` but it would still make sense to assign weights.

Also fwiw `x % np.inf == x if x > 0 else np.inf`

desc/equilibrium/equilibrium.py
f0uriest: So right now, `coordinates` is used to mean two related but different things:
1. How the function varies as you move in coordinate space
2. What coordinates need to be well resolved to evaluate it

I like the idea of `resolution_requirements` replacing the second function of `coordinates` so that we only ever have to check `resolution_requirements` when computing stuff, and `coordinates` can just be for metadata etc.

unalmis: Ok I have updated all compute functions with resolution requirements

desc/equilibrium/equilibrium.py
unalmis: I've lengthened the explanation in the comments in the code

desc/grid.py
unalmis: If I just use jax.numpy for these functions, then the nodes and spacing of linear grid also become jax arrays. If i recall correctly, tihs caused an error with one of current potential classes that made a lineargrid inside a traced fori_loop,

unalmis (commented):
I think I resolved the requested comments, so it can be reviewed again.
Some thoughts

1.  `eq.compute` can be made much simpler, and more robust, if we just build the transforms we need for the different grids, and copy the recursion in `desc.compute.utils._compute_fun` into `eq.compute` to compute things on the correct grids and broadcast as we go. Right now we're trying to simulate a recursive algorithm with a clunky iteration.
2. I marked quantities that integrate over a toroidal cross section as requiring `rt` resolution, or rho and theta resolution. The quadrature grid is good for integrating over rho, but does `quadrature_grid.spacing[:, 0]` have the correct quadrature weights for that, or is just the radial distance between flux surfaces?

desc/equilibrium/equilibrium.py
unalmis: we wouldn't need crutches like filtering out just_dep0d_dep if we did just did the natural recursive algorithm

desc/equilibrium/equilibrium.py
unalmis: we can't just only check `resolution_requirements`, because we also need to know how we can interpolate back to the original grid. Tha latter information resides in `coordinates` metadata

desc/grid.py
```diff
@@ -199,6 +209,29 @@ def sym(self):
...
+
+        Let the tuple (r, p, t) ∈ R³ denote a radial, poloidal, and toroidal
+        coordinate value. The is_meshgrid flag denotes whether any coordinate
+        can be iterated over along the relevant axis of the reshaped grid:
+        nodes.reshape(num_radial, num_poloidal, num_toroidal, 3).
+        """
+        return self.__dict__.setdefault("_is_meshgrid", False)
+
+    @property
+    def coordinates(self):
```
dpanici, line 224: There are enough coordinates now that it maybe warrants a key in the docstring saying which letter corresponds to what i.e. `a: alpha, the field line label` etc

desc/grid.py
```diff
@@ -873,71 +1060,44 @@ def _create_nodes(  # noqa: C901
...
-                    dz[-1] = dz[0]
-                if z[0] == 0 and z[-1] == ZETA_ENDPOINT:
-                    # The cyclic distance algorithm above correctly weights
-                    # the duplicate node spacing at zeta = 0 and 2pi / NFP.
-                    # However, scale_weights() is not aware of this, so we
-                    # counteract the reduction that will be done there.
-                    dz[0] += dz[-1]
-                    dz[-1] = dz[0]
-            else:
-                dz = np.array([ZETA_ENDPOINT])
```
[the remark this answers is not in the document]
dpanici: does this mean some quantities are incorrect on master?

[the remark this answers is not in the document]
unalmis: I can say with complete certainty that this bug has affected no one ever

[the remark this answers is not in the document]
unalmis: (People aren't going to do a flux surface integral on a grid with just a single zeta cross-section)

desc/grid.py
dpanici: That error might be fixed with #1002 if it was related to `compute_magnetic_field` but not sure

desc/compute/_geometry.py
```diff
@@ -196,7 +204,7 @@ def _A_of_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwarg
             transforms["grid"],
             data["Z"] * n[:, 2] * jnp.sqrt(data["g_tt"]),
             line_label="theta",
-            fix_surface=("rho", 1.0),
+            fix_surface=("rho", jnp.max(transforms["grid"].nodes[:, 0])),
```
dpanici, line 207: should #1086 be merged first?

dpanici (changes requested):
left some comments to be addressed/answered

dpanici:
> I think I resolved the requested comments, so it can be reviewed again. Some thoughts
> 
> 1. ~`eq.compute` can be made much simpler, and more robust, if we just build the transforms we need for the different grids, and copy the recursion in `desc.compute.utils._compute_fun` into `eq.compute` to compute things on the correct grids and broadcast as we go. Right now we're trying to simulate a recursive algorithm with a clunky iteration.~
> 2. I marked quantities that integrate over a toroidal cross section as requiring `rt` resolution, or rho and theta resolution. The quadrature grid is good for integrating over rho, but does `quadrature_grid.spacing[:, 0]` have the correct quadrature weights for that, or is just the radial distance between flux surfaces?

Yes, the `grid.spacing[:,0]` is the radial quadrature weights, so it is good for integrating (i.e. `V = np.sum(grid.spacing[:,0]*np.sum(grid.spacing[:,1]*np.sum(grid.spacing[:,2] * data["sqrt(g)"]`)

desc/compute/utils.py
```diff
@@ -163,30 +187,136 @@ def get_data_deps(keys, obj, has_axis=False):
...
-            out += axis_limit_deps.copy()  # to be safe
-            for dep in axis_limit_deps:
-                out += _get_deps_1_key(dep)
-        return sorted(list(set(out)))
 
-    out = []
-    for key in keys:
-        out += _get_deps_1_key(key)
-    return sorted(list(set(out)))
+def _get_deps_1_key(p, key, has_axis):
```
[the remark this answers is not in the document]
unalmis: nothing changed

desc/equilibrium/equilibrium.py
```diff
@@ -843,60 +873,128 @@ def compute(
...
+        # grid. We don't need to additionally compute A(z) and interpolate it back;
+        # it was only needed to compute R0, so we should remove it from the dep1dz list.
+        # If we don't remove it from the dep1dz list, then the code would try to create
+        # a linear grid with cross-sections at all the unique zeta values in the
+        # user-supplied grids. Typically, the user-supplied grid lacks unique_zeta_idx
+        # attribute, so this would cause an error.
+        dep0d_deps = set(
+            get_data_deps(dep0d, obj=p, has_axis=grid.axis.size, data=data)
+        )
+        # This filter is stronger than the name implies, but the false positives
```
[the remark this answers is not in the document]
unalmis: It may return `True` for some quantities that are not just dependencies of 0d dependencies, in other words it can generate a false positive. This is fine though since the false positive that is removed must necessarily be  computed when the actual computation happens.

For example, say we want to compute
1. A flux surface quantity `F` where `C` is a dependency of `F`.
2.  and a volume quantity `V`  where `C` is a dependency of `V`.

Now the user calls `eq.compute(names=[F, V]`, and we gather everything that needs to be compute: `[F, C, V]`. This "filter" will notice that `C` is in the dependencies for volume quantity `V` and not in `names` and then it will be removed from `dep1r`, the list of stuff that should be computed on grid's with resolution over flux surfaces. Here `C` was not `just_dep0d_dep`, but it's okay to take it out of `dep1dr` because the compute function will compute `C`  before `F` is computed anyway.

desc/grid.py
```diff
@@ -426,11 +518,11 @@ def copy_data_from_other(self, x, other_grid, surface_label="rho", tol=1e-14):
         Parameters
         ----------
         x : ndarray, shape(other_grid.num_nodes,...)
-            Data to copy
+            Data to copy. Assumed to be constant over the specified surface.
```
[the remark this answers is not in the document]
unalmis: yes, except instead of first, it would be last piece of data on that surface

desc/grid.py
```diff
@@ -508,23 +606,57 @@ class Grid(_Grid):
...
+        raz : rho, alpha, zeta
+        rvz : rho, theta_PEST, zeta
+        rtz : rho, theta, zeta
+    period : tuple of float
+        Assumed periodicity for each coordinate.
+        Use np.inf to denote no periodicity.
+    NFP : int
+        Number of field periods (Default = 1).
+    source_grid : Grid
+        Grid from which coordinates were mapped from.
```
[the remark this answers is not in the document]
unalmis: yes, see  get_rtz_grid in coords.py

desc/compute/data_index.py
```diff
@@ -112,6 +124,8 @@ def register_compute_fun(
     """
     if aliases is None:
         aliases = []
+    if source_grid_requirement is None:
```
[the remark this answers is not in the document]
unalmis: https://github.com/PlasmaControl/DESC/pull/448/files#r1230093069

desc/compute/utils.py
```diff
@@ -79,6 +79,28 @@ def compute(parameterization, names, params, transforms, profiles, data=None, **
...
+        def check_fun(name):
+            reqs = data_index[p][name]["source_grid_requirement"]
+            errorif(
+                reqs and not hasattr(transforms["grid"], "source_grid"),
+                AttributeError,
+                f"Expected grid with attribute 'source_grid' to compute {name}. "
+                f"Source grid should have coordinates: {reqs.get('coordinates')}.",
+            )
+            for req in reqs:
+                errorif(
```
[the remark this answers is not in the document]
unalmis: It is yes

desc/compute/utils.py
```diff
@@ -79,6 +79,28 @@ def compute(parameterization, names, params, transforms, profiles, data=None, **
...
+                    not hasattr(transforms["grid"].source_grid, req)
+                    or reqs[req] != getattr(transforms["grid"].source_grid, req),
+                    AttributeError,
+                    f"Expected grid with '{req}:{reqs[req]}' to compute {name}.",
+                )
+
+        _ = _get_deps(
```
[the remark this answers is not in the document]
unalmis: It ensures you don't try to compute some quantity K whose dependency relies on a field line integration etc. on a linear grid

desc/equilibrium/coords.py
```diff
@@ -504,3 +503,65 @@ def to_sfl(
...
+    ----------
+    eq : Equilibrium
+        Equilibrium on which to perform coordinate mapping.
+    radial : ndarray
+        Sorted unique radial coordinates.
+    poloidal : ndarray
+        Sorted unique poloidal coordinates.
+    toroidal : ndarray
+        Sorted unique toroidal coordinates.
+    coordinates : str
```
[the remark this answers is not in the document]
unalmis: I'd like it to match with what the grid attribute is

desc/equilibrium/coords.py
```diff
@@ -504,3 +503,65 @@ def to_sfl(
     eq_sfl.L_lmn = L_lmn_sfl
 
     return eq_sfl
+
+
+def get_rtz_grid(eq, radial, poloidal, toroidal, coordinates, period, jitable=True):
+    """Return DESC coordinate grid from given coordinates.
```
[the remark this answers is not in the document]
unalmis: Is `get_rtz_grid` not sufficient for this?

desc/compute/utils.py
```diff
@@ -163,30 +187,136 @@ def get_data_deps(keys, obj, has_axis=False):
...
+                    data_index[p][name]["dependencies"]["axis_limit_data"],
+                    deps,
+                    data,
+                    has_axis,
+                    check_fun,
+                )
+    return deps
+
+
+def _grow_seeds(parameterization, seeds, search_space, has_axis=False):
```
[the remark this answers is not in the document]
unalmis: instead of collecting the dependencies of something, it goes the other way

desc/equilibrium/equilibrium.py
```diff
@@ -843,60 +873,128 @@ def compute(
...
+            # We exclude these from the depXdx sets below since the grids we will
+            # use to compute those dependencies are coordinate-blind.
+            # Example, "<L|r,a>" has coordinates="r", but requires computing on
+            # field line following source grid.
+            return bool(data_index[p][name]["source_grid_requirement"])
+
+        # Need to call _grow_seeds so that some other quantity like K = 2 * <L|r,a>,
+        # which does not need a source grid to evaluate, does not compute <L|r,a> on a
+        # grid that does not follow field lines.
+        need_src_deps = _grow_seeds(p, set(filter(need_src, deps)), deps)
```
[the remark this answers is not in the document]
unalmis: It's the other way.

> Is there somewhere in this function call that ensures that these dependencies themselves do not require source_grid

That's precisely `need_src_deps`.

`need_src_deps` is
1. anything that needs to get computed that also
2. has a dependency which requires a special non-DESC-coordinate grid to compute

We inform the code of 2. by registering the compute function with a `source_grid` `kwarg ` and include whatever requirements are expected from the `source_grid` in that `kwarg`.

The effective ripple needs to do integration along field lines, so it requires a source grid in rho, alpha, zeta coordinates. This is so that we know which nodes in the DESC grid are along a field line (fixed rho, alpha). We also typically need to access the original source grid so that we know how far apart the nodes along a field line are.

---
Suppose we want to compute the volume average of the effective ripple. The compute function
```python
data=["effective ripple"],
def _compute_volume_average_ripple(....):
    data["volume average ripple"] = jnp.sum(jacorbian weights * data["effective ripple"])
    return data
```
does not directly need any special grid, however the dependency "effective ripple", certainly does. We need to ensure that we try to compute `volume average ripple` on a QuadratureGrid, because that would cause the dependency "effective ripple" to be computed on QuadratureGrid as well, giving wrong results. So we collect all such quantities that should be computed on the special grid in `need_src_deps`.

(Whether or not volume average ripple is important to anyone is irrelevant, since there are other examples where this issue arises, so it needs to be resolved). There is also no way to resolve this in a recursive manner that is smart about swapping grids and broadcasting the dependencies etc. I have a local branch that attempted this, and it reduces to this one since in general one needs to walk up the entire dependency tree with a quadrature grid first to get all the quantities that are constant throuhout the plasma volume to compute the special grid quantities.

desc/compute/_geometry.py
```diff
@@ -277,13 +286,12 @@ def _A_of_r(params, transforms, profiles, data, **kwargs):
...
+    resolution_requirement="tz",
 )
 def _S(params, transforms, profiles, data, **kwargs):
     data["S"] = jnp.max(
         surface_integrals(
-            transforms["grid"],
-            data["|e_theta x e_zeta|"],
-            expand_out=False,
+            transforms["grid"], data["|e_theta x e_zeta|"], expand_out=False
         )
```
unalmis, line 295: We should scale this up and change description like we did to A(z) in different PR.

unalmis:
I marked the resolved the comments I addressed. Feel free to unresolve them/make commits

rahulgaur104 (approved):
Try increasing the coverage, if possible.
If not, merge at will!

desc/grid.py
```diff
@@ -508,23 +606,57 @@ class Grid(_Grid):
...
         # Python 3.3 (PEP 412) introduced key-sharing dictionaries.
         # This change measurably reduces memory usage of objects that
         # define all attributes in their __init__ method.
-        self._NFP = 1
+        self._NFP = check_posint(NFP, "NFP", False)
         self._sym = False
         self._node_pattern = "custom"
+        self._coordinates = coordinates
+        self._period = period
+        if source_grid is not None:
```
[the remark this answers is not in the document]
f0uriest: Could not having the attribute defined cause IO issues? Wouldn't this mean that we get io warnings every time we load a regular grid?

I agree a recursive definition would be bad, but setting it to None or some other sentinel value might be better.

f0uriest (approved):
Approving but wait till after the next release to merge so we can try to keep things organized a bit.

rahulgaur104 (approved):

f0uriest (approved):

unalmis:
> Approving but wait till after the next release to merge so we can try to keep things organized a bit.

That's fine but fyi, but if you release a new version without this PR, we're going to get a lot of complaints from #1077. Recall the current pip version does not have #985 .

f0uriest (approved):

f0uriest:
One question about this and related PRs:

Do we plan to add logic to `eq.compute` for "default" grids for field line quantities? IE, if a user just asks for `epsilon_effective` at `rho=[0.1, 0.2, ...]` but doesn't specify alpha/zeta, can we do that? Or do we require them to pass in a full grid in `rho, alpha, zeta` coordinates?

unalmis:
The latter. Feel free to create a new github issue if that'd be convenient

rahulgaur104 (approved):

desc/equilibrium/equilibrium.py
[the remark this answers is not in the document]
YigitElma: I just faced this issue. I am trying to write a compute function for magnitude of force balance error normalized by volume averaged magnetic pressure gradient magnitude, for plotting.

```python
@register_compute_fun(
    name="|F|_normalized",
    label="|\\mathbf{J} \\times \\mathbf{B} - \\nabla p|/\\langle "
    + "|\\nabla |B|^{2}/(2\\mu_0)| \\rangle_{vol}",
    units="",
    units_long="",
    description="Magnitude of force balance error normalized by volume averaged "
    + "magnetic pressure gradient",
    dim=1,
    params=[],
    transforms={},
    profiles=[],
    coordinates="rtz",
    data=["|F|", "<|grad(|B|^2)|/2mu0>_vol"],
)
def _Fmag_normalized(params, transforms, profiles, data, **kwargs):
    data["|F|_normalized"] = data["|F|"] / jnp.nanmean(
        jnp.abs(data["<|grad(|B|^2)|/2mu0>_vol"])
    )
    return data
```

If I call compute functions with the grid of plotting it throws this error
```
AttributeError: Custom grids must have spacing specified by user.
```
 Any suggestions @unalmis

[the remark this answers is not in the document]
YigitElma: If I just comment out `and not just_dep0d_dep(dep)`, it works fine. Why do we need that line exactly?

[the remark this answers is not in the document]
unalmis: and thanks for tracking this down

PR #1025, opened 2024-05-20
ko/test coils

### LINKED ISSUE #1021: Add test to Check that we can optimize with every coil type
- [x] CoilLength so that comput is tested
- [x] QuadraticFlux (eq can be fixed) so that compute magnetic field is tested
- [x] sym CoilSet versus non stellarator sym
- [x] nested CoilSets versus single CoilSet versus regular Coil 
- [x] save/reload coilsets

it was mentioned in #1025 that `PlasmaCoilsetMinDistance` should be tested with optimization but it seems that it is already tested in `test_coilset_geometry_optimization()`

kianorr:
resolves #1021


dpanici:
Post min example (or I assume the test is one) and point PR at C0 branch

ddudt:
Add tests that use [DummyCoilSet](https://github.com/PlasmaControl/DESC/blob/master/tests/conftest.py#L231) and [DummyMixedCoilSet](https://github.com/PlasmaControl/DESC/blob/master/tests/conftest.py#L231)

kianorr:
minimal example:
```
R = 2
phi = 2 * np.pi * np.linspace(0, 1, 1001, endpoint=True) ** 2
coil = SplineXYZCoil(
    current=1, X=R * np.cos(phi), Y=R * np.sin(phi), Z=np.zeros_like(phi)
)
obj = ObjectiveFunction(CoilLength(coil, target=11))
# fails here because jax doesn't like the string
(coil,), _ = optimizer.optimize(coil objective=obj, maxiter=200, ftol=0, xtol=0)
length = coil.compute("length")["length"]
np.testing.assert_allclose(length, 11, atol=1e-3)
```

kianorr:
> minimal example:
> 
> ```python
> R = 2
> phi = 2 * np.pi * np.linspace(0, 1, 1001, endpoint=True) ** 2
> coil = SplineXYZCoil(
>     current=1, X=R * np.cos(phi), Y=R * np.sin(phi), Z=np.zeros_like(phi)
> )
> obj = ObjectiveFunction(CoilLength(coil, target=11))
> optimizer = Optimizer("fmintr")
> # fails here because jax doesn't like the string
> (coil,), _ = optimizer.optimize(coil objective=obj, maxiter=200, ftol=0, xtol=0)
> ```
> 
> the compute function (`_x_SplineXYZCurve`) has the information about `method` in `kwargs` so that's no longer the problem but jax is still not happy. This is weird because daniel is doing the exact same thing and it is working for him so I must be missing something.
> 
> @daniel-dudt @dpanici I'll take another look too but if you see anything that I'm messing up then please lmk!!

Update: for some reason, `method` becomes a key anyway in `transforms`. I tried tracking down why for a while but was unsuccessful

kianorr:
> Update: for some reason, `method` becomes a key anyway in `transforms` even though I'm not adding it when registering the compute function. This doesn't happen with `basis_in`. I tried tracking down why for a while but was unsuccessful

Update 2: figured it out, I forgot to change `transforms` in `_length_SplineXYZCurve` 🙃

kianorr:
I found a couple bugs with using `DummyCoilSet` and `MixedCoilSet`:

```python
coilset = load(load_from=str(DummyCoilSet["output_path_sym"]), file_format="hdf5")
print(coilset)
```
This print statement gives an error about the CoilSet not having a `_name` attribute

```python
coilset = load(load_from=str(DummyCoilSet["output_path_sym"]), file_format="hdf5")
test(coilset) # gives error at optimization
```
Optimizing gives an error about `numpy.bool_` (similar to #1030)
image

tests/test_examples.py
```diff
@@ -1298,18 +1306,92 @@ def test_second_stage_optimization():
 
 # TODO: replace this with the solution to Issue #1021
 @pytest.mark.unit
-def test_optimize_with_fourier_planar_coil():
+def test_optimize_with_all_coil_types():
     """Test optimizing a FourierPlanarCoil."""
```
dpanici, line 1310: update docstring

ddudt:
See if these new tests pass with the changes from PR #1030 on branch `dd/hotfix`. Let's make sure we have at least one test where a coilset is saved, loaded, and then re-optimized (that was the underlying problem in Issue #1029).

kianorr:
> See if these new tests pass with the changes from PR #1030 on branch `dd/hotfix`. Let's make sure we have at least one test where a coilset is saved, loaded, and then re-optimized (that was the underlying problem in Issue #1029).

So should I have one of the dummy coil sets to load/save and then the other would just return the coil set? Or should both `DummyCoilSet` and `DummyMixedCoilSet` save/load the coil set?

dpanici:
Close this probably and open a new PR (and one for method fix PR)

dpanici:
Also, try to test `PlasmaCoilsetMinDistance` as well as `QuadraticFlux` for each coil type

kianorr:
PR #1099 replaced this one

PR #1026, opened 2024-05-20
Add Vector Potential Calculation to `Coil` classes and Most `MagneticField` Classes 

### LINKED ISSUE #1023: Add Biot-Savart law for vector potential
Add the biot savart law for the magnetic vector potential as a callable function, like `biot_savart_general` is, and also add one for coils

dpanici:
Add utilities for calculating the magnetic vector potential for fields and coils/currents (assuming the Coulomb gauge). These will be useful for reconstruction as the most efficient way to calculate flux through a diagnostic flux loop is to integrate the vector potential. The utilities here should allow for calculating the vector potential from most sources of interest: coils and the plasma contribution (through `K=nxB` and the general biot-savart law here, as the loops are located outside from the plasma)

- [x] Add general biot savart law for vector potential
- [x] Add Hanson-Hirhsman biot savart law for coil vector potential
- [x] Add quadrature biot savart law for coil vector potential 
- [x] add to current potential
- [x] add tests for current potential (tests loop integral against enclosed toroidal flux)
- [x] add to coils
- [x] add test for coils for simple loop on z-axis
- [x] add a test against enclosed toroidal flux doing a loop integral (like the one in [this paper perhaps](https://iopscience.iop.org/article/10.1088/0741-3335/55/2/025014/meta))
  - [x] understand why flipping orientation of the coil parameterization does not flip sign of A? Check signs of A and x_s, though even just the A does not change sign as I expect... but I'd expect xs to change sign when the parameterizwion is flipped
- [x] add to the analytic fields
- [x] add tests for the analytic fields
- [x] Add VectorPotentialField
- [x] Add to SplineMagneticField (from mgrid)
- [x] note all gauges (try to put Coulomb Gauge for all) TF and VF are, and the rest are either from AD or mgrid, so we can't really check that
- [x] Use Vector potential for ToroidalFlux objective (possibly as an option, since not all B fields have a vector potential field)

Resolves #1023


f0uriest:
div/grad/curl in jax (assumes cartesian coordinates, will need some extra stuff to work in cylindrical)
```python
def grad(fun, argnum):
    return jax.jacfwd(fun, argnum)

def div(fun, argnum):
    def divf(*args, **kwargs):
        return jnp.trace(jax.jacfwd(fun, argnum)(*args, **kwargs))
    return divf

def curl(fun, argnum):
    def curlf(*args, **kwargs):
        J = jax.jacfwd(fun, argnum)(*args, **kwargs)
        return jnp.array([J[2,1]-J[1,2],
                         J[0,2]-J[2,0],
                         J[1,0]-J[0,1]])
    return curlf

File: desc/magnetic_fields/_core.py [hunk 1]
```diff
@@ -451,6 +510,22 @@ def save_mgrid(
         B_phi = field[:, 1].reshape(nphi, nZ, nR)
         B_Z = field[:, 2].reshape(nphi, nZ, nR)
 
+        # evaluate magnetic vector potential on grid
+        try:
```
f0uriest, line 514: might be better to make saving the vector potential a user option (so that we can reduce file size and compute time if its not needed), default to True, and throw an actual error if it fails

desc/magnetic_fields/_core.py
```diff
@@ -886,6 +1081,45 @@ def compute_magnetic_field(
 
         return B
 
+    def compute_magnetic_vector_potential(
+        self, coords, params=None, basis="rpz", source_grid=None
+    ):
+        """Compute magnetic vector potential at a set of points.
+
+            The vector potential is specified assuming the Coulomb Gauge.
```
f0uriest, line 1089: remove indent

desc/magnetic_fields/_core.py
```diff
@@ -886,6 +1081,45 @@ def compute_magnetic_field(
 
         return B
 
+    def compute_magnetic_vector_potential(
+        self, coords, params=None, basis="rpz", source_grid=None
```
f0uriest, line 1085: should this also take `transforms` for consistency with `compute_magnetic_field`? (and in many other places as well)

desc/magnetic_fields/_core.py
```diff
@@ -1062,6 +1338,33 @@ def compute_magnetic_field(
...
+            Unused by this MagneticField class.
+
+        Returns
+        -------
+        A : ndarray, shape(N,3)
+            magnetic vector potential at specified points
+
+        """
+        raise NotImplementedError(
+            "PoloidalMagneticField has nonzero divergence, therefore it can't be "
```
f0uriest, line 1364: I'm wondering if we want to just remove this class entirely, or maybe we can figure out some form that actually obeys div B=0?

desc/magnetic_fields/_core.py
```diff
@@ -1152,6 +1477,15 @@ def _atleast_4d(x):
         self._derivs["BR"] = self._approx_derivs(self._BR)
         self._derivs["Bphi"] = self._approx_derivs(self._Bphi)
         self._derivs["BZ"] = self._approx_derivs(self._BZ)
+        if AR is not None and Aphi is not None and AZ is not None:
+            AR, Aphi, AZ = map(_atleast_4d, (AR, Aphi, AZ))
+            assert AR.shape == Aphi.shape == AZ.shape == shape
+            self._AR = AR
```
f0uriest, line 1483: if they only get assigned some of the time but they're listed under io_attrs, I tihnk that will cause a bunch of warnings every time you load/save a spline field without them. Probably better to assign None in the case.

desc/magnetic_fields/_core.py
```diff
@@ -1366,8 +1823,41 @@ def from_mgrid(cls, mgrid_file, extcur=None, method="cubic", extrap=False):
...
+
+            # shift axes to correct order
+            ar = np.moveaxis(ar, (0, 1, 2), (1, 2, 0))
+            ap = np.moveaxis(ap, (0, 1, 2), (1, 2, 0))
+            az = np.moveaxis(az, (0, 1, 2), (1, 2, 0))
+        except IndexError as e:
+            warnif(
+                True,
+                UserWarning,
+                "Encountered error:"
```
f0uriest, line 1851: instead of just `Encountered error` maybe we want something like `mgrid does not appear to contain vector potential information`

desc/magnetic_fields/_core.py
```diff
@@ -1474,6 +1976,148 @@ def compute_magnetic_field(
...
+        r, p, z = coords.T
+
+        A = self._potential(r, p, z, **params)
+
+        if basis == "xyz":
+            A = rpz2xyz_vec(A, phi=coords[:, 1])
+        return A
+
+
+# TODO: Implement a VectorPotentialField that uses AD to do curl(A) on the input
```
f0uriest, line 2118: isnt this what the above class is doing?

f0uriest (changes requested):
A few small comments, my main comment though is that there's a lot of code that's shared/duplicated between the B and A calculations for each class which can probably be simplified. maybe something like 

```python
def _compute_AB(self, coords, params, op, ...):
    data = ... # compute dependencies etc
    # do other pre-processing
    return op(coords, data['x'], data['...'])

def compute_magnetic_field(self, coords, params, ...):
    return self._compute_AB(coords, params, op=biot_savart_B)

def compute_magnetic_vector_potential(self, coords, params, ...):
    return self._compute_AB(coords, params, op=biot_savart_A)
```

desc/magnetic_fields/_core.py
dpanici: ah yes, good catch. Must've added this before I added transforms everywhere else for compute magnetic field

desc/magnetic_fields/_core.py
dpanici: Yep, this can be removed

desc/magnetic_fields/_core.py
dpanici: I am fine with keeping it for now, though yea I have never used this one

desc/magnetic_fields/_core.py
```diff
@@ -921,16 +938,57 @@ def compute_magnetic_field(
             # zip does not terminate early
             transforms = transforms * len(self._fields)
 
-        B = 0
-        for i, (field, g, tr) in enumerate(zip(self._fields, source_grid, transforms)):
-            B += field.compute_magnetic_field(
-                coords, params[i % len(params)], basis, source_grid=g, transforms=tr
-            )
+        if compute_A_or_B == "B":
```
f0uriest, line 941: could simplify this further into a single loop with `getattr(field, op)(coords...)`

desc/magnetic_fields/_core.py
```diff
@@ -1542,110 +1676,161 @@ def compute_magnetic_field(
...
+            in cylindrical form [BR, Bphi,BZ]
 
         """
+        errorif(
+            compute_A_or_B not in ["A", "B"],
+            ValueError,
+            f'Expected "A" or "B" for compute_A_or_B, instead got {compute_A_or_B}',
+        )
+        errorif(
+            compute_A_or_B == "A" and not hasattr(self, "_AR"),
```
f0uriest, line 1696: this should check if `self._AR is not None`

desc/magnetic_fields/_core.py
```diff
@@ -537,6 +613,28 @@ def save_mgrid(
         )
         bz_001[:] = B_Z
 
+        if A_R:
```
f0uriest, line 616: does this work? I thought using an array as a bool caused issues?

dpanici: looking at it I feel like it should not work and this part of the code is actually untested rn too. I will see if we can add a test and then fix this too

desc/objectives/_coils.py
```diff
@@ -1348,6 +1348,15 @@ class ToroidalFlux(_Objective):
...
         Name of the objective function.
+    use_vector_potential : True
+        Whether to use the vector potential method to calculate the toroidal flux
+        (Φ = ∮ 𝐀 ⋅ 𝐝𝐥 over the perimeter of a constant zeta plane)
+        instead of the brute force method using the magnetic field
+        (Φ = ∯ 𝐁 ⋅ 𝐝𝐒 over a constant zeta XS). The vector potential method
+        is much more efficient, however not every ``MagneticField`` object
+        has a vector potential available to compute, so this option should be
+        set to False in those cases or if the magnetic field method is preferred.
+        #TODO: add eq_fixed option so this can be used in single stage
```
f0uriest, line 1359: still todo? if so at least move the comment so its not in the docstring

desc/objectives/_coils.py
```diff
@@ -1348,6 +1348,15 @@ class ToroidalFlux(_Objective):
         zeta=jnp.array(0.0), NFP=eq.NFP).
     name : str, optional
         Name of the objective function.
+    use_vector_potential : True
```
f0uriest, line 1351: could we instead just do 
```python
try:
    A = field.compute_vector_potential(...)
    Phi = dot(A, e_theta)
except ValueError:
    B = field.compute_magnetic_field(...)
    Phi = dot(B, e_zeta)
```

desc/magnetic_fields/_current_potential.py
```diff
@@ -631,25 +769,37 @@ def _compute_magnetic_field_from_CurrentPotentialField(
...
     coords = jnp.atleast_2d(jnp.asarray(coords))
     if basis == "rpz":
         coords = rpz2xyz(coords)
-
+    op = {"B": biot_savart_general, "A": biot_savart_general_vector_potential}[
+        compute_A_or_B
+    ]
     # compute surface current, and store grid quantities
     # needed for integration in class
+    # TODO: does this have to be xyz, or can it be computed in rpz as well?
```
f0uriest, line 797: still todo? it looks like you're using rpz below?

dpanici: not todo, I can remove

desc/objectives/_coils.py
dpanici: yes but it affects which data needs to be computed, if we are fine with extra data being computed I can make it this way

desc/objectives/_coils.py
```diff
@@ -1412,6 +1411,11 @@ def build(self, use_jit=True, verbose=1):
 
         """
         eq = self._eq
+        self._use_vector_potential = True
+        try:
+            self._field.compute_magnetic_vector_potential([0, 0, 0])
+        except NotImplementedError:
```
f0uriest, line 1417: Is this the only error we'd expect? I thought if a spline field doesn't have A it raises a ValueError?

dpanici: good catch, thanks

f0uriest (approved):

f0uriest (dismissed):

desc/coils.py
```diff
@@ -837,11 +1028,85 @@ def compute_magnetic_field(
...
-        B = biot_savart_hh(coords, coil_pts_start, coil_pts_end, current)
+        AB = op(coords, coil_pts_start, coil_pts_end, current)
 
         if basis == "rpz":
-            B = xyz2rpz_vec(B, x=coords[:, 0], y=coords[:, 1])
-        return B
+            AB = xyz2rpz_vec(AB, x=coords[:, 0], y=coords[:, 1])
+        return AB
+
+    def compute_magnetic_field(
```
ddudt, line 1037: Isn't this already implemented by the parent `_Coil` class?

desc/coils.py
```diff
@@ -837,11 +1028,85 @@ def compute_magnetic_field(
...
+        Notes
+        -----
+        Discretizes the coil into straight segments between grid points, and uses the
+        Hanson-Hirshman expression for exact field from a straight segment. Convergence
+        is approximately quadratic in the number of coil points.
+
+        """
+        return self._compute_A_or_B(coords, params, basis, source_grid, transforms, "B")
+
+    def compute_magnetic_vector_potential(
```
ddudt, line 1073: Isn't this already implemented by the parent `_Coil` class?

desc/magnetic_fields/_core.py
```diff
@@ -62,6 +62,40 @@ def body(i, B):
...
+    A = jnp.zeros_like(re)
+
+    def body(i, A):
+        r = re - rs[i, :]
+        num = JdV[i, :]
+        den = jnp.linalg.norm(r, axis=-1)
+        A = A + jnp.where(den[:, None] == 0, 0, num / den[:, None])
+        return A
+
+    return 1e-7 * fori_loop(0, J.shape[0], body, A)
```
ddudt, line 96: Where does the 1e-7 come from?

desc/coils.py
```diff
@@ -187,10 +270,16 @@ def _compute_position(self, params=None, grid=None, **kwargs):
             x = x.at[:, :, 1].set(jnp.mod(x[:, :, 1], 2 * jnp.pi))
         return x
 
-    def compute_magnetic_field(
-        self, coords, params=None, basis="rpz", source_grid=None, transforms=None
+    def _compute_A_or_B(
```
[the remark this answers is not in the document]
ddudt: See comments on the subclasses, but if this is getting superseded by child classes then it should be an abstract method. Same for `compute_magnetic_field` and `compute_magnetic_vector_potential`

desc/magnetic_fields/_core.py
```diff
@@ -537,6 +613,28 @@ def save_mgrid(
         )
         bz_001[:] = B_Z
 
+        if A_R is not None:
```
ddudt, line 616: I think it's cleaner to check `if save_vector_potential` instead

desc/magnetic_fields/_core.py
```diff
@@ -845,10 +1074,16 @@ def B0(self):
     def B0(self, new):
         self._B0 = float(np.squeeze(new))
 
-    def compute_magnetic_field(
-        self, coords, params=None, basis="rpz", source_grid=None, transforms=None
+    def _compute_A_or_B(
```
ddudt, line 1077: Same comment as I already mentioned for the Coil classes: I think it will be fewer lines of code and easier to understand if you get rid of these methods and just put the logic for either A or B in their respective methods.

tests/test_objective_funs.py
```diff
@@ -2281,7 +2289,25 @@ def test_compute_scalar_resolution_quadratic_flux(self):
...
+        ext_field = ToroidalMagneticField(1, 1)
+        eq = get("precise_QA")
+        with pytest.warns(UserWarning, match="Reducing radial"):
+            eq.change_resolution(4, 4, 4, 8, 8, 8)
+
+        f = np.zeros_like(self.res_array, dtype=float)
+        for i, res in enumerate(self.res_array):
+            eq.change_resolution(
+                L_grid=int(eq.L * res), M_grid=int(eq.M * res), N_grid=int(eq.N * res)
+            )
```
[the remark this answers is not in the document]
f0uriest: we do this for all the other tests, since it avoids having to deal with possible different grid types for different objectives.

desc/coils.py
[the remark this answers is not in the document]
f0uriest: the point is to avoid duplicated code. A lot of the preprocessing is the same, the only thing that really changes is the integrand

desc/coils.py
f0uriest: the base method uses direct quadrature, which is best if the curve is analytic (eg all the fourier coils), but for the spline coils we use the hanson-hirshman method which is better for splines/non-analytic stuff

desc/magnetic_fields/_core.py
f0uriest: mu_0/4pi == 1e-7

tests/test_objective_funs.py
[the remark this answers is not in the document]
ddudt: No, most of the other `scalar_resolution` tests we specify the grid resolution explicitly. These tests only test a single objective so we know the type of grid to use. 

The purpose of these tests is to check that the scalar form of the objectives gives the same value regardless of the grid resolution. By changing the equilibrium resolution, we could be introducing other variation from the equilibrium params changing. So I actually think it's important that we update these few tests that were changing the equilibrium resolution to change the grid resolution instead.

desc/coils.py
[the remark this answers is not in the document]
ddudt: There are only about a dozen lines of duplicate code, and that is mostly just for checking "rpz" vs "xyz" basis which we already do in all the compute functions too. I would rather make a single util function to do that check and conversion that we can call everywhere. Because the header/docstring of this function alone is like 40 lines of code that also gets duplicated many times, even if it isn't "real code".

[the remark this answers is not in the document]
dpanici: for `SplineMagneticField` and `CurrentPotentialField` the duplicated code is 100+ lines

I don't have a strong preference either way but I do think this current way is cleaner for avoiding duplicated code, and this is all under the hood stuff anyways. Unless there is a stronger case for splitting this up I want to keep it as it is now.

desc/coils.py
dpanici: same comment as Rory below, for splined coils the Hanson-Hirshman formula is preferred so we override in the subclass here

desc/coils.py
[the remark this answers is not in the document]
ddudt: In other words, move the actual duplicate code to a separate util function, and then all the code that is specific to A vs B can be moved to their respective compute functions.

tests/test_objective_funs.py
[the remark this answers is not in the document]
dpanici: dbl check then resolve

dpanici:
remove _ fxn for toroidal field and vertical field and poloidal field

desc/objectives/_coils.py
```diff
@@ -1483,22 +1503,33 @@ def compute(self, field_params=None, constants=None):
...
+            )
 
-        Psi = jnp.sum(
-            grid.spacing[:, 0]
-            * grid.spacing[:, 1]
-            * data["|e_rho x e_theta|"]
-            * B_dot_n_zeta
-        )
+            A_dot_e_theta = jnp.sum(A * data["e_theta"], axis=1)
+            # TODO: use the line integral compute utilities
```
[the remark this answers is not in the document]
dpanici: it actually seems to be slightly slower (as it under the hood then calls surface integrals and divides the result by the spacing of the direction not being integrated over) so I will remove this TODO and leave as is

tests/test_objective_funs.py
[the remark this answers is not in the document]
dpanici: I checked and we never change L M or N inside of a loop when checking convergence of the compute scalar resolution with increasing grid resolution

f0uriest (approved):

f0uriest (approved):

PR #1027, opened 2024-05-20
Move basis change to main compute function

### LINKED ISSUE #992: Reduce amount of repeated code in `_compute` by converting bases `xyz` to/from `rpz` in the `_compute` method
First raised in #924 , there is lots of duplicate code to just call `rpz2xyz_vec`, so we could probably just move that into the recursive `_compute` method and check for vector dims and use the basis argument there. Might need to add a `native_basis` kwarg to the compute fun decorator to note the few items that are natively computed in `xyz` (like the `SplineXYZ` or `FourierXYZ` curves)

### LINKED ISSUE #1088: Quantities are computed wrong unless returned data is in cylindrical basis vectors
We only want to do a basis conversion on quantities that directly compute a vector's contravariant components in the cylindrical coordinate system. Otherwise we'll be converting basis vectors from cylindrical to Cartesian multiple times because the dependencies are already in cartesian.

YigitElma:
Resolves #992
Resolves #1088

All compute quantities are now returned in toroidal $(R,\phi,Z)$ coordinates.  

TODO:

- [x] Implement `_curve` computes too
- [x] Add `test_compute_everthing_xyz`


YigitElma:
So, the point of this PR is to get rid of repeated code. But if I do so, the test `test_data_index_deps` fails because we don't have `data["phi"]` inside compute functions anymore. I can try to add `phi` in that test. But I want to see all other tests pass first.

desc/compute/utils.py
```diff
@@ -140,6 +140,16 @@ def _compute(
         data = data_index[parameterization][name]["fun"](
             params=params, transforms=transforms, profiles=profiles, data=data, **kwargs
         )
+        # to reduce repeated code, for 3D and naturally in rtz but desired in xyz basis
+        # quantities can be converted to xyz basis here
+        if (
```
dpanici, line 145: move this outside of recursive, have _compute functions assume things come in in rpz

dpanici:
Enforce passed-in data is in rpz always (add to docstring)

tests/test_data_index.py
```diff
@@ -111,6 +111,13 @@ def test_data_index_deps(self):
...
                 # assert correct dependencies are queried
-                assert queried_deps[p][name]["data"] == data | axis_limit_data, err_msg
+                # TODO: conversion from rpz to xyz is taken out of actual function
+                #       registration because of this data["phi"] is not queried in
+                #       the source code but actually needed for the computation. This
+                #       is a temporary fix until we have a better way to automatically
+                #       handle this.
+                assert queried_deps[p][name]["data"].issubset(
+                    data | axis_limit_data
+                ), err_msg
```
unalmis, line 121: If you can't find a better way, then let's instead subtract phi from the later set. FYI we removed the tests that ensures everything can be computed individually, i.e. without computing the entire dependency tree, and replaced it with this test which explicitly checks for this via inspecting the source code. For functions like the third derivatives of the basis vectors it's useful to keep this check as strict as possible.

YigitElma: Substracting "phi" may cause problems since we only want to remove it if the quantity is one of the 3d vectors which changed basis in the main compute function. 
This was a temporary change. i will think of a better way to do it, once I finish the actual coding part.

desc/compute/utils.py
```diff
@@ -141,6 +141,19 @@ def _compute(
...
         )
 
+    # to reduce repeated code, 3D quantities can be converted to xyz basis here.
+    # Assumes every 3D quantity is in rpz basis.
+    for name in names:
+        if (
+            data_index[parameterization][name]["dim"] == 3  # it should be 3D
+            and kwargs.get("basis", "rpz") == "xyz"  # user should ask in xyz
+            and ("phi" in data)  # phi is needed for conversion
+            and data_index[parameterization][name]["coordinates"] == "rtz"
```
dpanici, line 151: post which tests fail when this is removed

YigitElma:
Make every compute function to return in rpz by default and make all of the basis changes in main compute function

YigitElma:
Apply basis conversion to _curve compute functions too

YigitElma:
Can someone who uses coils implement the same thing for _curve compute functions? @dpanici @ddudt @kianorr I couldn't figure out what caused it to fail and this is probably an easy thing for you guys. I don't know which parameter should be in which basis by default.

desc/compute/_surface.py
```diff
@@ -47,13 +46,10 @@ def _x_FourierRZToroidalSurface(params, transforms, profiles, data, **kwargs):
...
     data=["x"],
     parameterization="desc.geometry.core.Surface",
-    basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
 )
 def _X_Surface(params, transforms, profiles, data, **kwargs):
     coords = data["x"]
-    if kwargs.get("basis", "rpz").lower() == "rpz":
-        # if basis is rpz, then "x" is rpz and we must convert to xyz
-        coords = rpz2xyz(coords)
+    coords = rpz2xyz(coords)
```
ddudt, line 52: Why are we converting this to Cartesian coordinates? I thought the whole point of this PR is to return everything in $(R,\phi,Z)$ coordinates?

desc/compute/_surface.py
```diff
@@ -71,13 +67,10 @@ def _X_Surface(params, transforms, profiles, data, **kwargs):
...
     data=["x"],
     parameterization="desc.geometry.core.Surface",
-    basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
 )
 def _Y_Surface(params, transforms, profiles, data, **kwargs):
     coords = data["x"]
-    if kwargs.get("basis", "rpz").lower() == "rpz":
-        # if basis is rpz, then "x" is rpz and we must convert to xyz
-        coords = rpz2xyz(coords)
+    coords = rpz2xyz(coords)
```
ddudt, line 73: see previous comment

desc/compute/_surface.py
YigitElma: I thought it wouldn't make sense that `X` returns `R`, or `Y` returns `Phi`. We already have `R`, `Phi` and `Z` for `rpz` coordinates.

YigitElma: X, Y, R and Phi should be the only data type that have strict basis definition, if I remember correctly

ddudt: Oh I see now. Yes this makes sense

ddudt:
> Can someone who uses coils implement the same thing for _curve compute functions? @dpanici @ddudt @kianorr I couldn't figure out what caused it to fail and this is probably an easy thing for you guys. I don't know which parameter should be in which basis by default.

I will work on this

desc/compute/utils.py
```diff
@@ -91,6 +92,27 @@ def compute(parameterization, names, params, transforms, profiles, data=None, **
...
+
+            if name == "x":
+                data[name] = rpz2xyz(data[name])
+            else:
+                errorif(
+                    "phi" not in data.keys(),
+                    ValueError,
+                    "'phi' must be included in the compute data "
+                    + "to convert to 'xyz' basis.",
+                )
```
ddudt, line 113: It would be convenient if we could automatically add `"phi"` to the list of compute quantity `names` when `basis="xyz"` is requested. Unfortunately I think that logic would have to be added to each class's `compute` method, since that is typically where we get the params that are needed for the list of compute quantities.

ddudt:
> > Can someone who uses coils implement the same thing for _curve compute functions? @dpanici @ddudt @kianorr I couldn't figure out what caused it to fail and this is probably an easy thing for you guys. I don't know which parameter should be in which basis by default.
> 
> I will work on this

@YigitElma this should be implemented for curves now

desc/compute/utils.py
YigitElma: In some curve functions, this was previously done by using transforms. So, I had a different if statement for `phi` is not included in the `data`. But yeah just having `phi` always should help.

YigitElma:
> > > Can someone who uses coils implement the same thing for _curve compute functions? @dpanici @ddudt @kianorr I couldn't figure out what caused it to fail and this is probably an easy thing for you guys. I don't know which parameter should be in which basis by default.
> > 
> > 
> > I will work on this
> 
> @YigitElma this should be implemented for curves now

Thanks! Then only thing left for thisPR is the `test_compute_everything_xyz`

YigitElma:
I don't understand how the new test fails. I created the new pickle file using the (commented) test and pushed that file to use it in the actual test. Maybe the precision of the devices github actions use is different? Or it could also be that I made a mistake in the test.

desc/compute/utils.py
```diff
@@ -91,6 +95,35 @@ def compute(parameterization, names, params, transforms, profiles, data=None, **
         data=data,
         **kwargs,
     )
+
+    # convert data from default 'rpz' basis to 'xyz' basis, if requested by the user
+    parameterization = _parse_parameterization(parameterization)
+    for name in data.keys():
```
f0uriest, line 101: can we swap the order of the loop and the `if basis == xyz` part? That way we won't have to check every qty if we're not doing any conversion

desc/compute/utils.py
```diff
@@ -91,6 +95,35 @@ def compute(parameterization, names, params, transforms, profiles, data=None, **
...
+
+            if name == "x":
+                data[name] = rpz2xyz(data[name])
+            else:
+                if "phi" in data:
+                    data[name] = rpz2xyz_vec(data[name], phi=data["phi"])
+                else:
+                    # compute phi if it's not already in data. Adding phi
+                    # to data at the beginning cause matrix dimension problems
+                    # TODO: Maybe look into that later (why error?)
```
f0uriest, line 116: what error were you getting? was it just a test failing or an actual error?

YigitElma: It was a matrix dimension error. I didn't go too deep into this but when I append phi directly to names before recursive _compute function, dimension of phi was not proper for usage in rpz2xyz_vec function. I felt like appending it at the beginning cause it not to use given grid (that seemed the only reason for dimension difference), or something like that

tests/test_compute_funs.py
```diff
@@ -1298,6 +1299,159 @@ def test_compute_everything():
     assert not error
 
 
+@pytest.mark.unit
+def test_compute_everything_xyz():
```
f0uriest, line 1303: we could combine this with the previous one and use `pytest.mark.parametrize`

Would also be good to explicitly say `basis="rpz"` in the previous one.

tests/test_data_index.py
f0uriest: I think this should be ok? We're ensuring that the dependencies are a subset of what actually gets computed right? as opposed to ensuring equality?

desc/coils.py
```diff
@@ -224,7 +224,7 @@ def compute_magnetic_field(
             current = params.pop("current", self.current)
 
         data = self.compute(
-            ["x", "x_s", "ds"], grid=source_grid, params=params, basis="xyz"
+            ["phi", "x", "x_s", "ds"], grid=source_grid, params=params, basis="xyz"
```
f0uriest, line 227: do we need to add phi here? is this because of the issue with adding it automatically in the compute func?

desc/compute/utils.py
f0uriest: can you post the error? I would think if you just do 
```
if basis == "xyz" and "phi" not in names:
    names += ["phi"]
```
should work fine?

YigitElma: I am away from my laptop now, but I can try to send it around dev meeting

tests/test_compute_funs.py
YigitElma: Yeah I also thought of that. I will combine them

desc/coils.py
YigitElma: @ddudt added that one

desc/compute/utils.py
ddudt: I'm not certain this will always work, because the `params` and `transforms` were already set for the original list of names and might not contain the info needed to compute `"phi"`

desc/compute/_curve.py
```diff
@@ -173,10 +159,10 @@ def _Z_Curve(params, transforms, profiles, data, **kwargs):
     coordinates="s",
     data=["s"],
     parameterization="desc.geometry.curve.FourierPlanarCurve",
-    basis="{'rpz', 'xyz'}: Basis for returned vectors, Default 'rpz'",
     basis_in="{'rpz', 'xyz'}: Basis for input params vectors, Default 'xyz'",
 )
 def _x_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
+    # convert to xyz for displacement and rotation
     if kwargs.get("basis_in", "xyz").lower() == "rpz":
```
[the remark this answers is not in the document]
ddudt: `"basis_in"` is specific to the `FourierPlanarCurve` class. It allows the curve parameters to be defined in either basis, and we decided this was better than creating a whole new Curve class (like `FourierPlanarCurveXYZ` vs `FourierPlanarCurveRPZ`). We still need this and it should be correct as is.

desc/coils.py
ddudt: Yes I added this before we added the logic to automatically add phi when needed. I think this might still be necessary because computing `"phi"` can require additional `params` relative to the other compute quantities for some Curve classes. See my other comment about this

desc/compute/utils.py
dpanici: `"phi"` (currently, this will change with the general angle stuff) only needs `grid` to be included in `transforms` which I think is almost always the case for basically any compute quantity

desc/objectives/_geometry.py
```diff
@@ -711,8 +711,8 @@ def build(self, use_jit=True, verbose=1):
                 params=self._surface.params_dict,
                 transforms=surface_transforms,
                 profiles={},
-                basis="xyz",
             )["x"]
+            surface_coords = rpz2xyz(surface_coords)
```
dpanici, line 715: why is this needed? can't you still request `basis=xyz`?

tests/test_curves.py
```diff
@@ -55,19 +55,15 @@ def test_torsion(self):
     def test_frenet(self):
         """Test frenet-serret frame of circular curve."""
         c = FourierRZCurve()
-        data = c.compute(
-            ["frenet_tangent", "frenet_normal", "frenet_binormal"], basis="rpz", grid=0
-        )
+        data = c.compute(["frenet_tangent", "frenet_normal", "frenet_binormal"], grid=0)
```
dpanici, line 58: why remove the basis arguments in these tests?

tests/test_curves.py
```diff
@@ -55,19 +55,15 @@ def test_torsion(self):
...
         np.testing.assert_allclose(T, np.array([[0, 1, 0]]), atol=1e-12)
         np.testing.assert_allclose(N, np.array([[-1, 0, 0]]), atol=1e-12)
         np.testing.assert_allclose(B, np.array([[0, 0, 1]]), atol=1e-12)
         c.rotate(angle=np.pi)
         c.flip([0, 1, 0])
         c.translate([1, 1, 1])
-        data = c.compute(
-            ["frenet_tangent", "frenet_normal", "frenet_binormal"], basis="xyz", grid=0
-        )
+        data = c.compute(["frenet_tangent", "frenet_normal", "frenet_binormal"], grid=0)
```
dpanici, line 66: same as above comment

dpanici:
Make sure `basis` is removed from compute function decorators

desc/objectives/_geometry.py
YigitElma: Inside the objectives, we don't use `compute` but instead use `_compute` which is jittable. Since the latter is recursive, I put conversion to the former. So, only for objective cases, you should change basis manually

tests/test_curves.py
YigitElma: @ddudt

tests/test_curves.py
YigitElma: @ddudt

desc/objectives/_geometry.py
dpanici: Got it, thanks. Might be worth noting this in the [making an objective function tutorial in the docs](https://desc-docs.readthedocs.io/en/latest/adding_objectives.html) then

YigitElma: It is documented in the `compute` and `_compute` functions

tests/test_curves.py
ddudt: "rpz" is now the default basis

tests/test_curves.py
ddudt: @dpanici I reverted this change, but are you sure this test is supposed to use both bases? It's difficult for me to visualize the geometry that is being tested here, since there are flips/rotations/etc. Same for the other two Frenet tests in this file.

ddudt: FYI with this reversion the test is now failing, but I think that's an unrelated bug that I still need to track down.

desc/geometry/core.py
```diff
@@ -441,14 +453,18 @@ def compute(
...
+        method = kwargs.pop("method", "auto")
         if params is None:
-            params = get_params(names, obj=self)
+            params = get_params(names, obj=self, basis=kwargs.get("basis", "rpz"))
         if transforms is None:
             transforms = get_transforms(
                 names,
                 obj=self,
                 grid=grid,
                 jitable=kwargs.pop("jitable", False),
```
ddudt, line 465: Does anyone know why we use `False` as the default here for the `Surface` class? For the `Curve` class we hard-code this to `True`.

tests/test_curves.py
ddudt: Actually this change is needed to make the test pass (both compute calls need to return "rpz" basis). I'm about 80% sure the old test was incorrect and it is fixed now, but whoever wrote this original test should check it.

desc/geometry/core.py
f0uriest: `jitable=True` means it uses the direct transforms, which for 1d curves is pretty cheap. For a 2d surface its a bit more expensive (especially if you need high resolution)

tests/test_curves.py
f0uriest: yes they should both be in rpz. It's just testing the directions of the frenet frame before and after flip/rotate etc. It's not messing with any basis stuff.

desc/compute/utils.py
```diff
@@ -140,11 +164,10 @@ def _compute(
         data = data_index[parameterization][name]["fun"](
             params=params, transforms=transforms, profiles=profiles, data=data, **kwargs
         )
-
     return data
 
 
-def get_data_deps(keys, obj, has_axis=False):
+def get_data_deps(keys, obj, has_axis=False, basis="rpz"):
```
ddudt, line 170: For a future PR: The variable we call `keys` here and in the other util functions below is the same as what we call `names` in the `compute` functions above and in other files. We should probably rename for consistency across the code.

tests/test_curves.py
ddudt: @f0uriest there are 3 different `test_frenet` functions in this file for different curve parameterizations:

1. `FourierRZCurve` now uses `basis="rpz"` for both computes
2. `FourierXYZCurve` uses `basis="rpz"` for the first compute and `basis="xyz"` for the second compute
3. `FourierPlanarCurve` uses `basis="xyz" for both computes

All three tests now pass, but should they really all be using "rpz" throughout?

f0uriest: Iirc when I made them I was using whatever the "native" basis was for the curve, but I don't know how much has changed since then

desc/compute/__init__.py
```diff
@@ -63,10 +63,10 @@ def _build_data_index():
     for p in data_index:
         for key in data_index[p]:
             full = {
-                "data": get_data_deps(key, p, has_axis=False),
-                "transforms": get_derivs(key, p, has_axis=False),
-                "params": get_params(key, p, has_axis=False),
-                "profiles": get_profiles(key, p, has_axis=False),
+                "data": get_data_deps(key, p, has_axis=False, basis="rpz"),
```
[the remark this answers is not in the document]
f0uriest: the only thing different would be `phi` but i don't think we want to compute that for everything if its not needed (right now its free but in the future may require building additional transforms)

desc/compute/utils.py
```diff
@@ -91,16 +96,37 @@ def compute(parameterization, names, params, transforms, profiles, data=None, **
...
+                    data[name] = rpz2xyz_vec(data[name], phi=data["phi"])
+
     return data
 
 
-# TODO: can we do the basis change here possibly? instead of repeating it everywhere?
-#  Maybe give an "inherent basis" parameter to the compute decorator
-#  so we check what the natural basis is versus what the desired is?
 def _compute(
     parameterization, names, params, transforms, profiles, data=None, **kwargs
```
f0uriest, line 120: maybe add a check that basis isnt in kwargs here? (it gets popped in the other function so wouldn't normally be passed in here)

desc/magnetic_fields/_core.py
```diff
@@ -1835,9 +1835,16 @@ def compute(
...
-            params = get_params(names, obj=self)
+            params = get_params(names, obj=self, basis=kwargs.get("basis", "rpz"))
         if transforms is None:
-            transforms = get_transforms(names, obj=self, grid=grid, **kwargs)
+            transforms = get_transforms(
+                names,
+                obj=self,
+                grid=grid,
+                basis=kwargs.get("basis", "rpz"),
+                method=kwargs.pop("method", "auto"),
```
f0uriest, line 1845: could this cause confusion with the spline method?

tests/test_compute_funs.py
```diff
@@ -1257,45 +1258,77 @@ def test_compute_everything():
...
+    error_rpz = False
+    error_xyz = False
 
     for p in things:
-        this_branch_data[p] = things[p].compute(
-            list(data_index[p].keys()), **grid.get(p, {})
+        this_branch_data_rpz[p] = things[p].compute(
+            list(data_index[p].keys()), **grid.get(p, {}), basis="rpz"
+        )
+        this_branch_data_xyz[p] = things[p].compute(
```
f0uriest, line 1275: think you need to remove the tensor qtys before computing here

desc/compute/__init__.py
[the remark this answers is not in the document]
ddudt: But if we don't include `"phi"` here then won't this potentially not work when trying to compute some thing in the xyz basis?

f0uriest (approved):

desc/compute/utils.py
```diff
@@ -91,16 +96,38 @@ def compute(parameterization, names, params, transforms, profiles, data=None, **
...
+    # convert data from default 'rpz' basis to 'xyz' basis, if requested by the user
+    if basis == "xyz":
+        from .geom_utils import rpz2xyz, rpz2xyz_vec
+
+        for name in data.keys():
+            errorif(
+                data_index[p][name]["dim"] == (3, 3),
+                NotImplementedError,
+                "Tensor quantities cannot be converted to Cartesian coordinates.",
+            )
```
unalmis, line 109: [Yes they can](https://github.com/PlasmaControl/DESC/issues/1110#issuecomment-2221187485)

ddudt (dismissed):
We should wait to get an approval from at least one person who hasn't contributed to this PR.

desc/compute/utils.py
```diff
@@ -248,8 +282,8 @@ def get_profiles(keys, obj, grid=None, has_axis=False, jitable=False, **kwargs):
         Grid to compute quantity on.
     has_axis : bool
         Whether the grid to compute on has a node on the magnetic axis.
-    jitable: bool
```
dpanici, line 251: This was unused?

tests/test_curves.py
```diff
@@ -66,7 +66,7 @@ def test_frenet(self):
         c.flip([0, 1, 0])
         c.translate([1, 1, 1])
         data = c.compute(
-            ["frenet_tangent", "frenet_normal", "frenet_binormal"], basis="xyz", grid=0
```
dpanici, line 69: Interesting that these vectors are the same in xyz and rpz

dpanici (approved):

tests/test_data_index.py
YigitElma: Actually since we deleted `phi` from dependencies, the previous version of this test might still be passing.

dpanici (changes requested):
Just realized, we should update the change log to reflect this as it's pretty major (that by default the compute function will return things in rpz, as before I think it wasn't always the case?)

ddudt (approved):

dpanici (approved):

f0uriest (approved):

tests/test_data_index.py
unalmis: @YigitElma In that case can you change this back

unalmis: Nevermind I changed it back on my branch.

PR #1028, opened 2024-05-21
`ExternalObjective` function to wrap external codes

ddudt:
Creates an abstract base class for wrapping external codes with finite differences, like GX, TERPSICHORE, etc. 

TODO:

- [x] vectorize
- [x] reverse mode gradient (hard-coded to be forward mode, because using forward finite difference)
- [x] tests
- [x] tutorial (will be added by #1046)


desc/objectives/_generic.py
```diff
@@ -116,27 +112,42 @@ def build(self, use_jit=True, verbose=1):
...
-            + "`thing.optimizable_params`.",
-        )
-
-        # wrap keyword arguments, which may not be JAX compatable data types
-        no_kwargs_fun = lambda *args, **kwargs: self._fun(*args, **self._kwargs)
-        sig = inspect.signature(self._fun)
-        params = [p for p in sig.parameters.values() if p.default == p.empty]
-        no_kwargs_fun.__signature__ = sig.replace(parameters=params)
+        def fun_wrapped(
+            R_lmn,
```
dpanici, line 116: pass in params_dict instead

desc/objectives/_generic.py
```diff
@@ -15,6 +16,236 @@
...
+            Ra_n,
+            Za_n,
+            Rb_lmn,
+            Zb_lmn,
+            I,
+            G,
+            Phi_mn,
+        ):
+            """Wrap external function with optimiazable params arguments."""
+            for param in self._eq.optimizable_params:
```
dpanici, line 137: add a loop to manualy multiprocess over the user function

dpanici: be wary of pitfalls, maybe make user do this instead

tests/test_examples.py
```diff
@@ -1313,3 +1316,112 @@ def test_optimize_with_fourier_planar_coil():
     optimizer = Optimizer("fmintr")
     (c,), _ = optimizer.optimize(c, objective=objective, maxiter=200, ftol=0, xtol=0)
     np.testing.assert_allclose(c.compute("length")[1]["length"], 11, atol=1e-3)
+
+
+@pytest.mark.unit
+@pytest.mark.slow
```
ddudt, line 1322: This takes a minute or so to run on my laptop, since `VMECIO.save` is slow. If we really need to speed this up we could either reduce the equilibrium resolution or manually save only the VMEC quantities that are used.

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
...
+                        param_value = np.array(params[param_key][k, :])
+                        if len(param_value):
+                            setattr(eq, param_key, param_value)
+                # parallelize calls to external function
+                max_processes = (
+                    self._vectorized
+                    if isinstance(self._vectorized, int)
+                    else os.cpu_count()
+                )
+                with mp.Pool(processes=min(max_processes, num_eq)) as pool:
```
ddudt, line 157: There is a bit of overhead with creating the pool. This creates and then closes the pool at every Jacobian calculation, which is theoretically wasteful. I experimented with creating it only once when the objective is built and then re-using it, but it didn't seem to save much time if any.

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
 from .objective_funs import _Objective
 
 
+class _ExternalObjective(_Objective, ABC):
+    """Wrap an external code.
+
+    Similar to ``ObjectiveFromUser``, except derivatives of the objective function are
+    computed with finite differences instead of AD.
```
f0uriest, line 26: should also mention that the function need not be jax transformable/traceable etc.

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
...
+        eq,
+        fun,
+        dim_f,
+        target=None,
+        bounds=None,
+        weight=1,
+        normalize=False,
+        normalize_target=False,
+        loss_function=None,
+        fd_step=1e-4,  # TODO: generalize this to allow a vector of different scales
```
f0uriest, line 83: maybe just have an absolute and relative step size? so that `dx = abs_step + rel_step*abs(x)`

not sure we want to deal with different step sizes for each element in x and figuring out how they get mapped after projections etc.

f0uriest: actually I don't think a different fd step per element would even work with JVPs, since using different steps for different components effectively changes the direction of the JVP.

desc/objectives/_generic.py
f0uriest: i think its probably safer to do it locally each time. JAX still assumes that the user function is "pure" (no side effects) so maintaining a thread pool would violate that, potentially causing some weird bugs

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
...
+        self._scalar = self._dim_f == 1
+        self._constants = {"quad_weights": 1.0}
+
+        if self._vectorized:
+            max_processes = (
+                self._vectorized
+                if isinstance(self._vectorized, int)
+                else os.cpu_count()
+            )
+            self._pool = mp.Pool(processes=max_processes)
```
f0uriest, line 134: so this isn't used?

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
...
+            # number of equilibria for vectorized computations
+            param_shape = params["Psi"].shape
+            num_eq = param_shape[0] if len(param_shape) > 1 else 1
+
+            if self._vectorized and num_eq > 1:
+                # convert params to list of equilibria
+                eqs = [self._eq.copy() for _ in range(num_eq)]
+                for k, eq in enumerate(eqs):
+                    # update equilibria with new params
+                    for param_key in self._eq.optimizable_params:
```
f0uriest, line 147: instead of this I think you can use `tree_unstack` from backend and to turn a dict of arrays into a list of dict, and then just do `eq.params_dict = params[i]`

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
...
+                with mp.Pool(processes=min(max_processes, num_eq)) as pool:
+                    results = pool.map(
+                        functools.partial(self._fun, **self._kwargs), eqs
+                    )
+                    pool.close()
+                    pool.join()
+                    return jnp.vstack(results, dtype=float)
+            else:  # no vectorization
+                # update equilibrium with new params
+                for param_key in self._eq.optimizable_params:
```
f0uriest, line 166: can't this just be `eq.params_dict = params`?

desc/objectives/objective_funs.py
```diff
@@ -58,6 +58,7 @@ def __init__(
 
     def _set_derivatives(self):
         """Set up derivatives of the objective functions."""
+        # TODO: does deriv_mode have to be "blocked" if there is an ExternalObjective?
```
f0uriest, line 61: no I think as long as ExternalObjective has `deriv_mode="fwd"` then either batched or blocked will work correctly.

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
...
+        Returns
+        -------
+        f : ndarray
+            Computed quantity.
+
+        """
+        f = self._fun_wrapped(params)
+        return f
+
+    def _jaxify(self, func, abstract_eval):
```
f0uriest, line 198: this probably belongs in backend or utils or something

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
...
+                x, unflatx = jax.flatten_util.ravel_pytree(primals)
+                v, _______ = jax.flatten_util.ravel_pytree(tangents)
+                # scale to unit norm if nonzero
+                normv = jnp.linalg.norm(v)
+                vh = jnp.where(normv == 0, v, v / normv)
+
+                def f(x):
+                    return jax.flatten_util.ravel_pytree(func(*unflatx(x)))[0]
+
+                tangent_out = (f(x + self._fd_step * vh) - y) / self._fd_step * normv
```
f0uriest, line 268: we may want to also scale fd_step by `norm(x)` to ensure that dx << x

desc/objectives/_generic.py
```diff
@@ -15,6 +19,262 @@
 from .objective_funs import _Objective
 
 
+class _ExternalObjective(_Objective, ABC):
```
f0uriest, line 22: does this need to be an ABC? It doesn't look like the user needs to supply anything other than the function.

desc/backend.py
```diff
@@ -10,15 +11,23 @@
 from desc import config as desc_config
 from desc import set_device
 
+verbose = True
+
+# set child processes to use numpy backend and suppress print statements
+if not multiprocessing.current_process().name == "MainProcess":
```
f0uriest, line 17: can you explain again why we have to use numpy in the child processes? I feel like that could be a significant slowdown since all the jax loops will revert to python loops.

Also, could this break user code that is running DESC in multiple processes?

f0uriest (changes requested):

desc/objectives/_generic.py
ddudt: Correct, I forgot to remove it after testing

desc/objectives/_generic.py
ddudt: Unfortunately I think it does need to be an ABC. Python multiprocessing needs the function to be in the global scope -- it has something to do with it's use of pickle. Basically, the user-supplied function has to be defined in the same file as the objective function class. It's really annoying but this is the only way I could get it to work. There are workarounds we could consider but they all have their own significant drawbacks.

desc/backend.py
ddudt: JAX is not compatible with python multithreading, and I think it's because JAX is already multithreaded itself so that leads to conflicts. But only the spawned processes are using Numpy, the main code should still be using JAX. 

There might be ways to get around this that we could look into: https://github.com/google/jax/issues/743

But in general, I think the most common use case is to write a wout file and then call some Fortran code on that input. And you can pre-compute all the transforms with JAX in the main code and then cast them to Numpy arrays to pass to the external function.

ddudt: I'm not sure about your second question of breaking code that is running DESC in multiple processes. I'm having issues using these external objectives when running on a GPU, even though the same code runs fine on a CPU. Maybe that is related?

desc/objectives/_generic.py
dpanici: reference the stack or give an example of a not working setup?

dpanici: Have user subclass the ABC and then define compute_implementation or something instead of passing in a function

dpanici:
put parallelism on user have tutorial/example of how to use mp/subprocess to parallelize

desc/backend.py
dpanici: check local imports of jax for possible issues

dpanici: Try w/o multiprocessing on GPU and see if wout computation is faster

desc/objectives/_generic.py
ddudt: [Stack Overflow](https://stackoverflow.com/questions/72766345/attributeerror-cant-pickle-local-object-in-multiprocessing) explaining that `multiprocessing` uses pickle and can therefore only serialize top-module level functions. 

For this reason I think it might not work to have to function defined as a method of the objective class, but I will check that.

ddudt: Here is another related [SO example](https://stackoverflow.com/questions/52265120/python-multiprocessing-pool-attributeerror)

desc/objectives/_generic.py
ddudt: @f0uriest Do you mean: `dx = abs_step + rel_step * np.mean(np.abs(x))` or `dx = abs_step + rel_step * np.abs(x)`? The former is a scalar and the latter is a vector. 

Also would this be in addition to your other comment of scaling `self._fd_step` by `norm(x)` within the `_jaxify` function, or would this be instead of that?

desc/objectives/_generic.py
dpanici: normalize vh to be same norm as |x|

dpanici: (and undo at the end as well considering the scales)

desc/objectives/_generic.py
ddudt: Normally `params` is a dict of 1D arrays. When vectorized, it becomes a dict of 2D arrays, where each row corresponds to the param values for a different equilibrium. So I don't think the pytree utils will help here.

desc/objectives/_generic.py
ddudt: No, because when vectorized `params` is a dict of 2D arrays with each row the params for a different equilibrium

desc/utils.py
```diff
@@ -676,3 +677,94 @@ def broadcast_tree(tree_in, tree_out, dtype=int):
...
+            fd_step = abs_step + rel_step * jnp.mean(jnp.abs(x))
+
+            # scale tangents to unit norm if nonzero
+            normv = jnp.linalg.norm(v)
+            vh = jnp.where(normv == 0, v, v / normv)
+
+            def f(x):
+                return jax.flatten_util.ravel_pytree(func(*unflatx(x)))[0]
+
+            tangent_out = (f(x + fd_step * vh) - y) / fd_step
```
ddudt, line 763: @f0uriest I think this is correct and we don't need to multiply by a factor of `normv`, but please check this for me

desc/objectives/_generic.py
ddudt: `x` here is really `params`, which includes variables at very different scales like boundary coefficients vs current profile coefficients. We are using the same finite difference step size for everything, which means that step could be huge for some variables and tiny for others. Then when generating wout files and running external codes for those perturbed equilibria, the results might not be physically meaningful.

dpanici:
Make a blocked `grad` and `hess` to allow `rev` mode with this objective (or have it use FD for `grad` and `hess`)

desc/utils.py
dpanici: you need it

dpanici:
Add a basic tutorial (call `subprocess` etc but just to run a simple python script not an actual external code)

desc/utils.py
```diff
@@ -682,3 +683,94 @@ def broadcast_tree(tree_in, tree_out, dtype=int):
...
+        Total step size is ``abs_step + rel_step * mean(abs(x))``.
+
+    Returns
+    -------
+    func : callable
+        New function that behaves as func but works with jit/vmap/jacfwd etc.
+
+    """
+
+    def wrap_pure_callback(func):
```
YigitElma, line 733: Maybe we can add a standalone decorator to `desc/backend.py` for the `pure_callback`. It can help us run a specific portion of a jit function on the CPU. Our current decorator `execute_on_cpu` only works if the function is not called by another jitted function. [today I asked that question here](https://github.com/google/jax/discussions/23079 )

tests/test_examples.py
ddudt: I sped this up significantly by manually saving the few quantities instead of calling VMECIO.save

f0uriest:
Might be useful when we want to do multithreading https://github.com/jax-ml/jax/issues/24756

ddudt:
I'm having an issue with the new test in this PR after updating with `master` where the code gets stuck in some infinite loop. I can make the test pass locally if I add a print statement after the optimizer's first Jacobian call -- that seems to snap it back somehow and then it continues to run fine. Any ideas of what could be going on? This isn't using any multiprocessing

dpanici:
Put wrapper fxn in backend which checks JAX version and uses correct API

desc/utils.py
```diff
@@ -743,6 +745,97 @@ def atleast_nd(ndmin, ary):
...
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            result_shape_dtype = abstract_eval(*args, **kwargs)
+            return jax.pure_callback(
+                func,
+                result_shape_dtype,
+                *args,
+                vmap_method="expand_dims" if vectorized else "sequential",
```
ddudt, line 801: This works with JAX v0.4.35 and later. Need to add backwards compatibility. 

I think any other `vmap_method` besides `"expand_dims"` would also work since it only changes the dimensions for internal calculations and not the final result.

desc/optimize/least_squares.py
```diff
@@ -178,7 +178,7 @@ def lsqtr(  # noqa: C901
     f = fun(x, *args)
     nfev += 1
     cost = 0.5 * jnp.dot(f, f)
-    J = jac(x, *args)
+    J = jac(x, *args).block_until_ready()  # FIXME: block is needed for jaxify util
```
f0uriest, line 181: did you figure out what was going on here? This feels like a kludge, and an incomplete one at that (ie wouldn't we need to do similar things for all the other optimizers?)

desc/utils.py
f0uriest: should figure this out if possible, otherwise we need to bump our minimum jax version by a lot

desc/backend.py
```diff
@@ -1,6 +1,7 @@
 """Backend functions for DESC, with options for JAX or regular numpy."""
 
 import functools
+import multiprocessing as mp
```
f0uriest, line 4: is this needed? I thought you got rid of all the multiprocessing stuff?

If it's just to avoid printing extra stuff I'd rather not import it, sometimes just importing multiprocessing and jax can lead to random (mostly benign, but potentially confusing) warnings.

desc/objectives/_generic.py
```diff
@@ -11,12 +12,160 @@
...
 from desc.grid import QuadratureGrid
 from desc.optimizable import OptimizableCollection
-from desc.utils import errorif, parse_argname_change
+from desc.utils import errorif, jaxify, parse_argname_change
 
 from .linear_objectives import _FixedObjective
 from .objective_funs import _Objective, collect_docs
 
 
+class ExternalObjective(_Objective, ABC):
```
f0uriest, line 21: why is this an ABC? in the tests it looks like you instantiate it directly?

desc/objectives/_generic.py
```diff
@@ -11,12 +12,160 @@
...
+    """Wrap an external code.
+
+    Similar to ``ObjectiveFromUser``, except derivatives of the objective function are
+    computed with finite differences instead of AD. The function does not need not be
+    JAX transformable.
+
+    The user supplied function must take an Equilibrium or a list of Equilibria as its
```
f0uriest, line 28: would be good to include a simple example/mockup of what the function should look like as part of the docstring, could be something super simple like we do for `ObjectiveFromUser`

f0uriest: should also mention that the function needs to return a single 1d array of floats

desc/objectives/_generic.py
```diff
@@ -11,12 +12,160 @@
...
+        abs_step=1e-4,
+        rel_step=0,
+        target=None,
+        bounds=None,
+        weight=1,
+        normalize=False,
+        normalize_target=False,
+        loss_function=None,
+        name="external",
+        **kwargs,
```
f0uriest, line 79: This assumes that any kwargs are to be passed to the user function, which might cause issues down the road if we need kwargs for anything else (like handling an argument renaming). Would be cleaner if instead the kwargs to the user function were just passed as a dict like
```
fun_kwargs={"a": True, "b": 1.2}

File: desc/objectives/_generic.py [hunk 23]
```diff
@@ -11,12 +12,160 @@
...
+
+        """
+        self._scalar = self._dim_f == 1
+        self._constants = {"quad_weights": 1.0}
+
+        def fun_wrapped(params):
+            """Wrap external function with possibly vectorized params."""
+            # number of equilibria for vectorized computations
+            param_shape = params["Psi"].shape
+            num_eq = param_shape[0] if len(param_shape) > 1 else 1
```
f0uriest, line 120: isnt this just `len(jnp.atleast_1d(params['Psi']))`?

f0uriest (changes requested):

desc/backend.py
ddudt: There is no multiprocessing stuff in this PR, so I could move these changes to #1046. It is only to avoid printing extra stuff, but I argue that it is needed. Because otherwise with the TERPSICHORE objective (and any others that use multiprocessing) the optimizer output becomes illegible. It will print this backend stuff of "DESC version ... Using device ..." hundreds of times between each optimizer iteration.

desc/optimize/least_squares.py
ddudt: I am still not sure why this is needed. Last time I tested on all the optimizers (a couple of months ago) it was only an issue for our custom least-squares optimizer. The SciPy ones and our augmented-Lagrangian were fine. 

I will try testing again, but I might need help if we want to avoid this kludge.

desc/utils.py
ddudt: They deprecated the `vectorized` argument to `pure_callback` and replaced it with `vmap_method`. So we just need to add a wrapper to backend to check the JAX version and use the appropriate one.

ddudt: I would want help testing that if anyone is still using an old version of JAX

desc/backend.py
dpanici: make env info a standalone function, and do not auto-print it upon import

desc/utils.py
```diff
@@ -743,6 +745,97 @@ def atleast_nd(ndmin, ary):
...
+        Total step size is ``abs_step + rel_step * mean(abs(x))``.
+
+    Returns
+    -------
+    func : callable
+        New function that behaves as func but works with jit/vmap/jacfwd etc.
+
+    """
+
+    def wrap_pure_callback(func):
```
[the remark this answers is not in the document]
dpanici: bump

desc/utils.py
dpanici: put this in backend

desc/objectives/_generic.py
f0uriest: for a single equilibrium i think psi should be a scalar, not an array? so that when its vmapped it would be a 1d array not 2d.

desc/backend.py
```diff
@@ -43,40 +33,38 @@
...
     except ModuleNotFoundError:
         jnp = np
-        x = jnp.linspace(0, 5)
-        y = jnp.exp(x)
         use_jax = False
         set_device(kind="cpu")
         warnings.warn(colored("Failed to load JAX", "red"))
+
+
+def print_info():
```
f0uriest, line 44: `print_environment_info` or `print_backend_info` might be more descriptive.

Also should mention this in the docs, maybe along with `set_device` and mention it in the changelog that it will no longer print info by default.

Should also call this in `__main__` so that when running a simple command line input its the same

desc/objectives/_generic.py
ddudt: `eq.params_dict`, `_Objective.xs(eq)`, and `ObjectiveFunction.unpack_state(ObjectiveFunction.x(eq))` all return a dict of arrays. So Psi is never a scalar, unless of course the user passes in a custom set of parameters of different types.

ddudt:
`test_jaxify` in `tests/test_utils.py` is failing on the CI for Python 3.9 with the error message:
`ValueError: IO callbacks do not support JVP.`
But it passes for me locally with JAX 0.4.37. So maybe `io_callback` was also updated in a recent version of JAX? But I don't see anything about that in their Change Log...

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ New Features
 - Changes hessian computation to use chunked ``jacfwd`` and ``jacrev``, allowing ``jac_chunk_size`` to now reduce hessian memory usage as well.
 - Adds an option to ``VMECIO.save`` to specify the grid resolution in real space.
 - Adds a new objective ``desc.objectives.CoilIntegratedCurvature`` for targeting convex coils.
+- Adds a new objective ``desc.objectives.ExternalObjective`` for wrapping external codes with finite differences.
+- DESC/JAX version and device info is no longer printed by default, but can be accessed with the function `print_backend_info()`.
```
f0uriest, line 22: give full path to the function

docs/installation.rst
```diff
@@ -258,9 +259,10 @@ To check that you have properly installed DESC and its dependencies, try the fol
     python
     >>> from desc import set_device  # only needed if running on a GPU
     >>> set_device('gpu')  # only needed if running on a GPU
+    >>> from desc.backend import print_backend_info
+    >>> print_backend_info()
     >>> import desc.equilibrium
```
f0uriest, line 264: i think in theory we can get rid of this line, previously it was only meant to trigger the backend import to print the info, but thats its own call now

f0uriest:
> `test_jaxify` in `tests/test_utils.py` is failing on the CI for Python 3.9 with the error message: `ValueError: IO callbacks do not support JVP.` But it passes for me locally with JAX 0.4.37. So maybe `io_callback` was also updated in a recent version of JAX? But I don't see anything about that in their Change Log...

did you figure out the cause of this? it seems to be passing now?

ddudt:
> > `test_jaxify` in `tests/test_utils.py` is failing on the CI for Python 3.9 with the error message: `ValueError: IO callbacks do not support JVP.` But it passes for me locally with JAX 0.4.37. So maybe `io_callback` was also updated in a recent version of JAX? But I don't see anything about that in their Change Log...
> 
> did you figure out the cause of this? it seems to be passing now?

Are the tests deterministic on which ones get run on which Python versions? It looks like this time the test was run on v3.12 so that would explain why it passed.

ddudt:
> > > `test_jaxify` in `tests/test_utils.py` is failing on the CI for Python 3.9 with the error message: `ValueError: IO callbacks do not support JVP.` But it passes for me locally with JAX 0.4.37. So maybe `io_callback` was also updated in a recent version of JAX? But I don't see anything about that in their Change Log...
> > 
> > 
> > did you figure out the cause of this? it seems to be passing now?
> 
> Are the tests deterministic on which ones get run on which Python versions? It looks like this time the test was run on v3.12 so that would explain why it passed.

I switched back to using `jax.pure_callback` instead of `jax.experimental.io_callback` (basically reverting commit `2d561ee`). 

`io_callback` worked for the simple test on this PR, but it wasn't working for realistic tests on #1046 and I'm not certain why. Technically we are doing I/O operations, but the pure callback is better supported by JAX and seems to work for these cases. The only downside is that we need the backend check for which version of JAX is being used. 

All this is to say I think this issue is resolved now.

desc/utils.py
```diff
@@ -743,6 +745,93 @@ def atleast_nd(ndmin, ary):
...
+
+    Positional arguments to func can be differentiated,
+    use keyword args for static values and non-differentiable stuff.
+
+    Note: Only forward mode differentiation is supported currently.
+
+    Parameters
+    ----------
+    func : callable
+        Function to wrap. Should be a "pure" function, in that it has no side
```
YigitElma, line 759: The example you gave above (the one given in the docstring), doesn't it technically has a side effects? It saves the output to a file. It will probably work since the saving and loading steps are basically redundant, but for overall explanation, it could be better to give another example.

docs/dev_guide/notebooks/backend.ipynb
```diff
@@ -42,40 +42,60 @@
...
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [
     {
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.\n"
+      "WARNING:2025-01-31 16:53:19,421:jax._src.xla_bridge:969: An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.\n"
```
YigitElma, line 58: If you are on windows wsl, I found that our della instructions work. I haven't noticed any numerical issue yet.

YigitElma (changes requested):
Minor typos.

Maybe we should add a new notebook about external and user-defined objectives? I can add one while solving #1295, and we can add a proper example for the `ExternalObjective` there

ddudt:
> Minor typos.
> 
> Maybe we should add a new notebook about external and user-defined objectives? I can add one while solving #1295, and we can add a proper example for the `ExternalObjective` there

There is already a new notebook that will get added in #1046

desc/backend.py
```diff
@@ -94,6 +89,29 @@
     else:
         trapezoid = jnp.trapz  # for older versions of JAX, deprecated by jax 0.4.16
 
+    # TODO: update this when min JAX version >= 0.4.35
+    jax_vs = jaxlib.__version__.strip().split(".")
+    jax_version = float(jax_vs[1]) + float(jax_vs[2]) / 1e2
+    if jax_version >= 4.35:
```
f0uriest, line 95: This is generally not a great way to check version numbers since they're meant to be compared lexicographically not as decimal numbers. See plotting for how we do it for matplotlib

f0uriest: https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

desc/utils.py
ddudt: Technically you are correct. We should probably use `io_callback` instead of `pure_callback`, but `io_callback` doesn't work (I'm not certain why) and `pure_callback` does work even when writing/reading files. I like this example because most often this objective will involve I/O operations to interface with external codes. Maybe we could remove the part about no side effects from this description? Although it is supposed to be a pure function since we are using `pure_callback`, it just happens to work.

docs/dev_guide/notebooks/backend.ipynb
ddudt: JAX works on the GPU on my laptop, but I purposely uninstalled it because my GPU is too small to do any useful optimizations.

desc/utils.py
YigitElma: Hmm, I saw this description in the official Jax docs, too. But maybe they are refering to pure in the sense that except the outputs of the function it doesn't change anything in the memory that is used outside of that function. So, if a "pure" function takes in some input array A, the values of A shouldn't change?

desc/backend.py
ddudt: I changed it to be like how we do if for matplotlib, which is not using the Version package you linked to.

YigitElma: It helps for consistency, but Daniel's version seemed ok to me, why it is not recommended @f0uriest ?

f0uriest: bc in versioning, `v0.4.39` > `v0.4.4`, but if you compare them as decimal numbers, 4.4 > 4.39

desc/backend.py
```diff
@@ -94,6 +91,30 @@
     else:
         trapezoid = jnp.trapz  # for older versions of JAX, deprecated by jax 0.4.16
 
+    # TODO: update this when min JAX version >= 0.4.35
+    version = jax.__version__.split(".")
+    major = int(version[1])
+    minor = int(version[2])
+    if major >= 4 and minor >= 35:
```
f0uriest, line 98: I think this also won't work. jax 0.5.0 was just released, and that would evaluate to False here and use the wrong code.

desc/backend.py
YigitElma: Daniel's version was dividing the minor version by 100 to account for that, but anyway.

desc/utils.py
f0uriest: it is still pure as in "doesn't maintain external state". The file is written and read within the same function, so the global state is the same before and after the function call.

desc/backend.py
YigitElma: We can switch to something like (by assuming no package will ever have a version like `1.0.101`, so minor and major versions are always 2 digits max),
```python
version = jax.__version__.split(".")
main = int(version[0])
major = int(version[1])
minor = int(version[2])
# assume none of the indiviual ints is bigger than 99
d_version = main*100 + major + minor /100 
if d_version >= 4.35:
```

YigitElma: So, eventually when we get jax 1.0.0, the value of `d_version` will be 100, and the logic will still work.

f0uriest: I'd prefer to just use the method from the standard library I linked in the other comment

dpanici:
try without calling `eq.compute` as well

f0uriest (approved):

YigitElma (approved):

YigitElma:
@ddudt I guess this is ready to be merged?

PR #1030, opened 2024-05-22
ensure correct data types in getter methods

### LINKED ISSUE #1029: Attribute types are not always correct
Some class attributes are not always the intended data type. For example, `Equilibrium.NFP` is supposed to be an `int` but in the past it was sometimes a `float`. More recently, I discovered that `Equilibrium.sym` was of type `numpy.bool_` instead of the python `bool`, which is treated differently by the pytree utility functions we use in the coil objectives. I recommend we cast to the intended data type in class getter methods to ensure the returned type is always correct.

ddudt:
Resolves #1029


dpanici:
Post code where error occurs on master to better diagnose this?

kianorr:
Do we know why the bool ends up being in the pytree in the first place?

ddudt:
> Post code where error occurs on master to better diagnose this?

See latest comment in the Issue

desc/io/hdf5_io.py
```diff
@@ -129,6 +129,11 @@ def read_obj(self, obj, where=None):
                 continue
             if isinstance(loc[attr], h5py.Dataset):
                 s = self._decode_attr(loc, attr)
+                # cast NumPy data types to native Python types
```
ddudt, line 132: The main issue is that h5py converts everything to numpy data types when saving, so then when we load the data we need to cast it back to the native python types.

f0uriest: I think the best solution to this is declaring the attributes as static like so: https://github.com/PlasmaControl/DESC/blob/5b6bfdce331380867cc912651fc542bf600f0433/desc/magnetic_fields/_core.py#L1086-L1088

That way the actual data type shouldn't matter.

ddudt: @f0uriest Can you explain why this is better? Are you saying we should define all bool and int attributes for every class as static? 

I want to make sure our solution is backwards compatible with old files that were saved with numpy data types, and my hack covers that. 

Also, is the `"_period"` static attribute of the `SplineMagneticField` actually used? I don't see it referenced in the code.

desc/basis.py
```diff
@@ -238,7 +243,7 @@ def __init__(self, L, sym="even"):
         self._M = 0
         self._N = 0
         self._NFP = 1
-        self._sym = sym
+        self._sym = bool(sym) if not sym else str(sym)
```
kianorr, line 246: why make this a string?

desc/basis.py
```diff
@@ -351,7 +356,7 @@ def __init__(self, N, NFP=1, sym=False):
         self._M = 0
         self._N = check_nonnegint(N, "N", False)
         self._NFP = check_posint(NFP, "NFP", False)
-        self._sym = sym
+        self._sym = bool(sym) if not sym else str(sym)
```
kianorr, line 359: same with this?

desc/basis.py
ddudt: because `sym` is either False or a string specifying the type of symmetry, like "cos" or "sin"

desc/io/hdf5_io.py
f0uriest: so the real issue is actually coming from how the objects are flattened here: https://github.com/PlasmaControl/DESC/blob/5b6bfdce331380867cc912651fc542bf600f0433/desc/io/equilibrium_io.py#L109-L139

along with the default types that we assume are static here: https://github.com/PlasmaControl/DESC/blob/5b6bfdce331380867cc912651fc542bf600f0433/desc/io/equilibrium_io.py#L79-L87

my point was that in general we should probably avoid special casing that relies on particular data types, and instead declare particular attributes static or dynamic if the default behavior isn't correct. 

IE, our code shouldn't care whether something is a python `int` or `np.int64`, we should make sure it works for either.

f0uriest: FWIW I think pretty much all the changes in this PR could be replaced by just adding the following to `_unjitable`:

```python
if hasattr(x, "dtype") and np.isscalar(x):
    return np.issubdtype(x.dtype, np.bool_) or np.issubdtype(x.dtype, np.int_)
```

ddudt: Minimal example on `master`, where `coilset_sym` is from the `DummyCoilSet` in `conftest`: 
```
    (
        CoilLength(coils=coilset_sym),
        CoilCurvature(coil=coilset_sym),
    ),
)
constraints = (FixCoilCurrent(coilset_sym),)
[coilset_sym], _ = optimizer.optimize(
    things=coilset_sym,
    objective=objective,
    constraints=constraints,
    verbose=3,
)
```
Gives the error message: `AttributeError: 'numpy.bool_' object has no attribute 'N'`

With only your suggested change to `_unjitable`, it gives a different error message: 
`ValueError: Mismatch custom node data: [('_NFP', 4), ('_sym', ('ndarray', (), (True,))), ('_name', '')] != [('_NFP', 4), ('_name', '')]; value: CoilSet at 0x7fcfa9022690 (name=, with 3 submembers).`

I agree my solution in this PR is a little hacky, but it works and we need to fix this bug ASAP for coil optimizations to work. I doubt python and numpy are going to be adding new data types anytime soon that we need to add logic for. If we come up with a more elegant solution later we can always update it.

desc/objectives/_coils.py
ddudt, line 1: While not directly related to the underlying issue for this PR, a lot of the data type errors were coming from the `_CoilObjective.build` method. The pytree stuff we had in here was really clunky, like creating a `MixedCoilSet` that contained `_Grid`s instead of `_Coil`s. I tried to simplify the logic.

kianorr: I've looked at it a little bit but haven't gotten the chance to understand the changes yet

desc/io/hdf5_io.py
f0uriest: I think that's because of the stuff like this:
```python
is_leaf=lambda x: not hasattr(x, "__len__")
```

in the coil objectives. I'm not sure what thats actually supposed to be checking but it treats numpy arrays as leafs which im pretty sure we don't want.

f0uriest: Actually nevermind, I had a mistake above. the part added to unjitable should be 
```python
    if hasattr(x, "dtype") and np.ndim(x)==0:
        return np.issubdtype(x.dtype, np.bool_) or np.issubdtype(x.dtype, np.int_)
```

with that fix  on master your example above seems to run fine

desc/objectives/_coils.py
```diff
@@ -98,77 +98,48 @@ def build(self, use_jit=True, verbose=1):  # noqa:C901
...
-        else:
-            # this case covers an inputted list of grids that matches the size
-            # of the inputted coils. Can be a 1D list or nested list.
-            flattened_grid = tree_leaves(
-                self._grid, is_leaf=lambda x: isinstance(x, _Grid)
-            )
-            self._grid = tree_unflatten(coil_structure, flattened_grid)
+        assert isinstance(self._grid, list)
+
+        self._dim_f = np.sum(
```
f0uriest, line 119: can't this just be `sum(grid.num_nodes for grid in self._grid)`?

f0uriest: we do basically the same for `quad_weights` below

desc/objectives/_coils.py
```diff
@@ -98,77 +98,48 @@ def build(self, use_jit=True, verbose=1):  # noqa:C901
...
-                return [coilset]
-
-        # gives structure of coils, e.g. MixedCoilSet(coils, coils) would give a
-        # a structure of [[*, *], [*, *]] if n = 2 coils
-        coil_leaves, coil_structure = tree_flatten(
-            self.things[0], is_leaf=lambda x: not hasattr(x, "__len__")
-        )
-        self._num_coils = len(coil_leaves)
+        # get individual coils from coilset
+        coils = tree_leaves(coil, is_leaf=lambda x: hasattr(x, "_current"))
```
f0uriest, line 107: I'd prefer something like 

```python
def _is_single_coil(x):
    return isinstance(x, _Coil) and not isinstance(x, CoilSet)

...
is_leaf=_is_single_coil
```

rather than checking if it has `_current`, since other objects may have a `_current` attribute (like equilibrium, winding surface, etc).

f0uriest (changes requested):
The changes to the setting stuff is generally fine, ie stuff like 
```
self._sym = bool(sym)
```

But I'd prefer to get rid of the changes to the getting methods, ie
```
def sym(self):
    return bool(self._sym)
```

It's basically redundant, but in the off chance that they're inconsistent it can lead to some weird bugs (and major annoyance) if `self.sym` and `self._sym` have different types for some reason.

(also all of the IO and JAX stuff only cares about the private `_sym` attribute. the `sym` property is only accessed by the user.)


See also my comment above about changing the default `_unjitable` util to treat scalar bools and ints as constants.


My bigger point though is that we should try to avoid assigning static or dynamic attributes by checking data type whenever possible, since its error prone and may fail in weird edge cases like we've seen. The better way I think is to declare the attribute itself as static/dynamic, so it will be handled correctly regardless of the shape and dtype.

desc/objectives/_coils.py
ddudt: Actually I needed to update `quad_weights` to match the new logic like `dim_f`. Thanks for catching that. 

Before, the grids were mapped to the same tree structure as the input `coil`. That was nice because it always made sure the structures matched (there was a grid for every coil), but it was also awkward because we were creating a `MixedCoilSet` that had `Grid` elements instead of `Coils`. 

My changes now avoid that mapping, and the format for specifying the grids has changed a bit too. Say we have a `MixedCoilSet` that contains two `CoilSet`s, the first with 2 coils and the second with 3 coils. Before you would give the grid for each individual `Coil` like this: 
```
grid = [[grid1, grid2], [grid3, grid4, grid5]]
```
And now you only give the grids for each `CoilSet` like this: 
```
grid = [grid1, grid2]
```

This is better, because all of the coils in a `CoilSet` must have the same grid. Before, we were requiring the user to specify all of these individual grids and then throwing away all of them except for the first one. The downside of my new method is this extra logic to account for all of the coils within each coil set.

dpanici: Is this still necessary or can we go back to the more general pytree calls for this?

ddudt: I reverted some of my changes to `_CoilObjective` but I still needed to make some major revisions. It turns out that the `master` branch is not functional with general nested trees (we can't optimize a `MixedCoilSet` that contains other `MixedCoilSet`s). That is another bug technically different from the original intent of this PR, but I think fixed it here anyways. I also updated the tests to cover more cases like this.

desc/objectives/_coils.py
```diff
@@ -96,102 +96,74 @@ def build(self, use_jit=True, verbose=1):  # noqa:C901
...
+        # map grid to list of length coils
+        if grid is None:
+            grid = [LinearGrid(N=2 * c.N + 5, endpoint=False) for c in coils]
+        if isinstance(grid, numbers.Integral):
+            grid = LinearGrid(N=self._grid, endpoint=False)
+        if isinstance(grid, _Grid):
+            grid = [grid] * self._num_coils
+        if isinstance(grid, list):
+            grid = tree_leaves(grid, is_leaf=lambda g: isinstance(g, _Grid))
+        assert len(grid) == len(coils)
```
f0uriest, line 130: might want to make this a `ValueError` with a more descriptive error message?

Or maybe put a separate check after `if isinstance(grid, list):` since that's the only case where it could cause problems

desc/objectives/_coils.py
```diff
@@ -96,102 +96,74 @@ def build(self, use_jit=True, verbose=1):  # noqa:C901
...
-        )
-        transforms = tree_leaves(
-            to_list(transforms), is_leaf=lambda x: isinstance(x, dict)
-        )
-
-        errorif(
-            np.any([grid.num_rho > 1 or grid.num_theta > 1 for grid in self._grid]),
-            ValueError,
-            "Only use toroidal resolution for coil grids.",
+            lambda c, g: get_transforms(self._data_keys, obj=c, grid=g),
```
f0uriest, line 149: minor point: This will get more transforms than needed, since the ones for `CoilSet` are later pruned to just the unique ones. Would be nice if we can avoid that redundant calculation but not sure how feasible it is.

f0uriest (changes requested):
Mostly fine, just see comment about giving a more informative error message when grids/coils don't match.

desc/objectives/_coils.py
f0uriest: also what if the user passes in different grids for each coil in a regular `CoilSet`? I think in that case the code as is now will only use the first grid?

desc/objectives/_coils.py
ddudt: I think I've resolved this in the latest commit

desc/objectives/_coils.py
ddudt: Re your second question: yes the code will only use the first grid. That is how the existing code also behaves, so this PR is not changing that. My changes earlier this week corrected this, but then I reverted them because my solution didn't work for nested coilsets. We could try to fix this in the future.

f0uriest (approved):

dpanici (approved):

kianorr (approved):

PR #1037, opened 2024-05-26
Fix zernike_eval notebook

YigitElma:
There was also a small typo (when I was trying to fix the format I miss typed - as +).

I also changed the division inside `zernike_radial_coeffs` to be a float division. Probably the accuracy for big numbers was an issue before, but I used Decimal() to make division more accurate.


YigitElma:
And there is a really high chance that our implementations with JAX are not stable after L>50

YigitElma:
@f0uriest 3f544fd92d0b1dcd79d634550f2d4e53f2891fc5 is this the reason you wanted to use "//" instead of "/"?

dpanici (changes requested):
Undo decimal change

docs/notebooks/zernike_eval.ipynb
f0uriest, line 1: also bunch of new ``<>:32: SyntaxWarning: invalid escape sequence '\m'`` in here?

YigitElma: Can this be a matplotlib change? Because I didn't make any changes. If I change change the existing code from 
```python
ax[0].set_title(
    "$\max_{x \in (0,1)} |\\frac{d^3Z_{lm}(x)}{dx^3} - \\frac{d^3\\tilde{Z}_{lm}^{poly}(x)}{dx^3}| / |\\bar{Z}'''_{lm}|$",
    fontsize=14,
)
```
to this
```python
ax[0].set_title(
    "$\\max_{x \\in (0,1)} |\\frac{d^3Z_{lm}(x)}{dx^3} - \\frac{d^3\\tilde{Z}_{lm}^{poly}(x)}{dx^3}| / |\\bar{Z}'''_{lm}|$",
    fontsize=14,
)
```
It works.

f0uriest: yeah that's probably it

YigitElma:
I found a little bit of accuracy improvement on zernike_radial function. Instead of using gammaln and then taking the exponential of it to find coefficients for the derivatives, we can just use the multiplication, as I mentioned in the updated notebook.
```python
# coefficient for derivative
    coeffs = jnp.array(
        [
            1,
            (alpha + n + 1) / 2,
            (alpha + n + 2) * (alpha + n + 1) / 4,
            (alpha + n + 3) * (alpha + n + 2) * (alpha + n + 1) / 8,
            (alpha + n + 4) * (alpha + n + 3) * (alpha + n + 2) * (alpha + n + 1) / 16,
        ]
    )
    c = coeffs[dx]
```
All we need is use above code instead of
```python
# coefficient for derivative
    c = (
        gammaln(alpha + beta + n + 1 + dx)
        - dx * jnp.log(2)
        - gammaln(alpha + beta + n + 1)
    )
    c = jnp.exp(c)
```
This part. I also wrote and tested a version like this
```python
# coefficient for derivative
    def poch(x, dx):
        def body(k, val):
            return val * (x + k)

        return fori_loop(0, dx, body, 1)

    c = poch(alpha + n + 1, dx) / (2**dx)
```
GPU executing time doesn't change but I didn't want to risk fwd/reverse AD.

f0uriest:
Sure, I'd recommend this version:
```python
# coefficient for derivative
    coeffs = jnp.array(
        [
            1,
            (alpha + n + 1) / 2,
            (alpha + n + 2) * (alpha + n + 1) / 4,
            (alpha + n + 3) * (alpha + n + 2) * (alpha + n + 1) / 8,
            (alpha + n + 4) * (alpha + n + 3) * (alpha + n + 2) * (alpha + n + 1) / 16,
        ]
    )
    c = coeffs[dx]
```

to avoid loops and special function calls

f0uriest (approved):

dpanici (approved):
Out of curiosity, how different were the quantities that caused `test_compute_everything` to fail after you changed the part in the zernike algorithm?
EDIT: nevermind I see from the tests they are negligible changes in the higher order derived quantities

YigitElma:
> Out of curiosity, how different were the quantities that caused `test_compute_everything` to fail after you changed the part in the zernike algorithm? EDIT: nevermind I see from the tests they are negligible changes in the higher order derived quantities

```
Not equal to tolerance rtol=1e-10, atol=1e-10
Parameterization: desc.equilibrium.equilibrium.Equilibrium. Name: J^rho.
Mismatched elements: 23 / 660 (3.48%)
Max absolute difference: 2.29007128e-07
Max relative difference: 6.56482638e-09
 x: array([-4.121012e+02, -1.125881e+03, -1.537982e+03, -1.537982e+03,
       -1.125881e+03, -4.121012e+02, -6.978519e+02, -1.423246e+03,
       -1.440419e+03, -2.124563e+03, -2.423728e+03, -1.091549e+03,...
 y: array([-4.121012e+02, -1.125881e+03, -1.537982e+03, -1.537982e+03,
       -1.125881e+03, -4.121012e+02, -6.978519e+02, -1.423246e+03,
       -1.440419e+03, -2.124563e+03, -2.423728e+03, -1.091549e+03,...

Not equal to tolerance rtol=1e-10, atol=1e-10
Parameterization: desc.equilibrium.equilibrium.Equilibrium. Name: J^theta*sqrt(g).
Mismatched elements: 1 / 660 (0.152%)
Max absolute difference: 4.81682946e-07
Max relative difference: 1.12579674e-10
 x: array([-2.301279e+03, -1.684653e+03, -6.166258e+02,  6.166258e+02,
        1.684653e+03,  2.301279e+03,  1.783800e+05,  1.785994e+05,
        1.776159e+05,  1.768467e+05,  1.784834e+05,  1.805368e+05,...
 y: array([-2.301279e+03, -1.684653e+03, -6.166258e+02,  6.166258e+02,
        1.684653e+03,  2.301279e+03,  1.783800e+05,  1.785994e+05,
        1.776159e+05,  1.768467e+05,  1.784834e+05,  1.805368e+05,...

Not equal to tolerance rtol=1e-10, atol=1e-10
Parameterization: desc.equilibrium.equilibrium.Equilibrium. Name: J^theta.
Mismatched elements: 1 / 660 (0.152%)
Max absolute difference: 3.80525307e-07
Max relative difference: 1.1257934e-10
 x: array([         -inf,          -inf,          -inf,           inf,
                 inf,           inf,  1.135946e+06,  1.119998e+06,
        1.084518e+06,  1.049024e+06,  1.035086e+06,  1.034512e+06,...
 y: array([         -inf,          -inf,          -inf,           inf,
                 inf,           inf,  1.135946e+06,  1.119998e+06,
        1.084518e+06,  1.049024e+06,  1.035086e+06,  1.034512e+06,...
```

dpanici:
@f0uriest @ddudt can we just get rid of `test_1d_optimization_old`? it is testing the old `_optimize` function which I don't quite understand why we still have?

desc/utils.py
```diff
@@ -457,8 +457,8 @@ def combination_permutation(m, n, equals=True):
 def multinomial_coefficients(m, n):
     """Number of ways to place n objects into m bins."""
     k = combination_permutation(m, n)
-    num = factorial(n)
-    den = factorial(k).prod(axis=-1)
+    num = factorial(n, exact=True)
```
kianorr, line 460: what does `exact=True` do?

desc/basis.py
```diff
@@ -1424,14 +1434,15 @@ def zernike_radial_coeffs(l, m, exact=True):
...
-                    factorial((ll - s) // 2)
-                    * factorial((s + mm) // 2)
-                    * factorial((s - mm) // 2)
-                )
+            # Zernike polynomials can also be written in the form of [1] which
+            # states that the coefficients are given by the binomial coefficients
+            # hence they are all integers. So, we can use exact arithmetic with integer
+            # division instead of floating point division.
+            # [1]https://en.wikipedia.org/wiki/Zernike_polynomials#Other_representations
+            coeffs[ii, s] = ((-1) ** ((ll - s) // 2) * factorial((ll + s) // 2)) // (
```
kianorr, line 1442: there's no difference here right? Just the extra comments?

YigitElma: Yes, there is no change. Additional lines confused git, I guess.

desc/utils.py
YigitElma: Here, we are using scipy s factorial function which is by default an approximation with gamma function. To be more accurate, we should use integer operations instead. Exact=true converts it to integer operation. https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.factorial.html

f0uriest:
> @f0uriest @ddudt can we just get rid of `test_1d_optimization_old`? it is testing the old `_optimize` function which I don't quite understand why we still have?

I'm fine getting rid of the test, along with the `_optimize` method. It's still there in older versions for anyone who wants to reproduce the results from the paper.

YigitElma:
> > @f0uriest @ddudt can we just get rid of `test_1d_optimization_old`? it is testing the old `_optimize` function which I don't quite understand why we still have?
> 
> I'm fine getting rid of the test, along with the `_optimize` method. It's still there in older versions for anyone who wants to reproduce the results from the paper.

Since we have some other changes that affect `_optimize()`, maybe we should refer them to some version of DESC instead of this method.

If you are okay, I can remove the method and the test with this PR.

PR #1042, opened 2024-06-02
Gamma_c  (Nemov and Velasco et. al) Γ_c

### LINKED ISSUE #522: Add particle confinement proxies
such as Nemov's original $\Gamma_C$ [link](https://itp.tugraz.at/AG/APP/PLASMA/PUBLIC/OWN_PUBLICATIONS/PAPERS/Nemov2008-052501.pdf) or the newer $\Gamma_\alpha$ [link](https://iopscience.iop.org/article/10.1088/1741-4326/ac2994/pdf?casa_token=zeIpampigfkAAAAA:HrpLe7n6fwJAICFhvT5N9v8eV0lFEcyXXNnp8CzaiULodzT-AMlOqTnDFYPIMq25i4pXpgbUBxwKFZ4)

unalmis:
Resolves #522 . Automatically differentiable


desc/compute/_metric.py
```diff
@@ -1908,20 +1908,23 @@ def _cvdrift(params, transforms, profiles, data, **kwargs):
...
     profiles=[],
     coordinates="rtz",
-    data=["|B|^2", "b", "e^rho", "grad(|B|)"],
+    data=["|B|^2", "b", "e^rho", "grad(|B|)", "rho"],
 )
 def _cvdrift0(params, transforms, profiles, data, **kwargs):
     data["cvdrift0"] = (
-        1 / data["|B|^2"] * (dot(data["b"], cross(data["grad(|B|)"], data["e^rho"])))
+        2
+        * data["rho"]
```
unalmis, line 1926: @rahulgaur104 I'm assuming this conflicts with your ballooning branch?

desc/objectives/_neoclassical.py
```diff
@@ -454,7 +454,7 @@ def compute(self, params, constants=None):
             self._keys,
             params,
             get_transforms(self._keys, eq, grid, jitable=True),
-            get_profiles(self._keys, eq, grid, jitable=True),
+            get_profiles(self._keys, eq, grid),
```
unalmis, line 457: Would it be better to get profiles inside build? Or maybe these are changing throughout optimization. same thing on ripple branch

dpanici: I am pretty sure we can get them inside of build? if we are on the same flux surface, profiles are only functions of rho (for the most part). so that might at leats remove one possible source of error

rahulgaur104:
Two places where we can improve GammaC:

1. When we bounce average the radial and binormal drifts, we can eliminate the denominator integral(tau) since we only care about bounce-avg(normal)/bounce-avg(binormal).
2. The normal and binormal derivatives are the radial and binormal derivatives of the quantity J = int v_|| dl/tau, the second adiabatic invariant and it doesn't have a non-integrable singularity so it's more robust.

I'll add the analytical model from Hegna to the tests probably in the  bounce branch.

unalmis:
> Two places where we can improve GammaC:
> 
>     1. When we bounce average the radial and binormal drifts, we can eliminate the denominator integral(tau) since we only care about bounce-avg(normal)/bounce-avg(binormal).
> 
>     2. The normal and binormal averages  are the radial and binormal derivatives of the quantity J = int v_|| dl/tau, the second adiabatic invariant and it doesn't have a non-integrable singularity so it's more robust.
> 
> 
> I'll add the analytical model from Hegna to the tests probably in the bounce branch.

Documenting for others after discussion
1. Is already done
2. It's not possible to approximate a locally defined quantity such as J with a finite number of spectral coefficients

unalmis:
@ddudt I have added Nemov's Gamma_c in [921f4d5](https://github.com/PlasmaControl/DESC/pull/1042/commits/921f4d52b8acda70f901ce17e7832d9cb8805356)

desc/compute/_basis_vectors.py
```diff
@@ -2450,10 +2450,10 @@ def _e_sub_theta_over_sqrt_g(params, transforms, profiles, data, **kwargs):
...
 def _e_sub_vartheta_rp(params, transforms, profiles, data, **kwargs):
     # constant ρ and ϕ
-    e_vartheta = (
-        data["e_theta"].T * data["phi_z"] - data["e_zeta"].T * data["phi_t"]
-    ) / (data["theta_PEST_t"] * data["phi_z"] - data["theta_PEST_z"] * data["phi_t"])
-    data["e_theta_PEST"] = e_vartheta.T
+    data["e_theta_PEST"] = (
+        (data["e_theta"].T * data["phi_z"] - data["e_zeta"].T * data["phi_t"])
+        / (data["theta_PEST_t"] * data["phi_z"] - data["theta_PEST_z"] * data["phi_t"])
+    ).T
```
unalmis, line 2456: Made this cosmetic change so that it shows up in git diff and reviewer can confirm new quantities are correct by inspection.

desc/compute/_field.py
```diff
@@ -106,6 +106,220 @@ def _B_sup_zeta(params, transforms, profiles, data, **kwargs):
...
+    description="Contravariant cylindrical toroidal angle component of magnetic field",
+    dim=1,
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["B0", "phi_z", "lambda_t", "phi_t", "lambda_z"],
+)
+def _B_sup_phi(params, transforms, profiles, data, **kwargs):
+    data["B^phi"] = data["B0"] * (
```
unalmis, line 123: This is written in this form, independent of iota, so that it can be computed without performing any integrations.

desc/compute/_field.py
```diff
@@ -146,7 +360,7 @@ def _B_R(params, transforms, profiles, data, **kwargs):
 
 @register_compute_fun(
     name="B_phi",
-    label="B_{\\phi}",
+    label="B_{\\phi} = \\mathbf{B} \\cdot \\hat{\\phi}",
```
unalmis, line 363: Because this is not B dot e_phi

tests/test_compute_funs.py
```diff
@@ -1523,35 +1523,56 @@ def test_surface_equilibrium_geometry():
...
+            - data["|B|"]
+            * data["B^zeta_z|r,a"]
+            * np.sign(data["B^zeta"])
+            / data["B^zeta"] ** 2,
+        )
+        np.testing.assert_allclose(
+            data["B"], cross(data["grad(psi)"], data["grad(alpha)"])
+        )
+        np.testing.assert_allclose(
+            data["B^zeta"], data["psi_r"] / data["sqrt(g)_Clebsch"]
```
unalmis, line 1568: Linking for reference

desc/objectives/_neoclassical.py
```diff
@@ -248,3 +254,253 @@ def compute(self, params, constants=None):
...
+        Whether to use the Γ_c as defined by Nemov et al. or Velasco et al.
+        Default is Nemov. Set to ``False`` to use Velascos's.
+
+        Note that Nemov's Γ_c converges to a finite nonzero value in the
+        infinity limit of the number of toroidal transits.
+        Velasco's expression is defined to be zero on irrational surfaces;
+        and therefore, the numerical computation will converge to zero as the
+        number of toroidal transits increases. This is mentioned to remind
+        users that an optimization using Velasco's metric should be evaluated by
+        measuring decrease in Γ_c at a fixed number of toroidal transits.
```
unalmis, line 341: @PlasmaControl/desc-dev because grad(alpha) -> infinity and the bounce averages of radial/poloidal drift will therefore tend to 0. see first few pages of Nemov's gamma c to see how he handles this issue

f0uriest:
From looking through STELLOPTs version of Gamma_C [here](https://github.com/PrincetonUniversity/STELLOPT/blob/653cae047b08b7fbfe4ac0166e63247bbe2f6199/STELLOPTV2/Sources/Chisq/chisq_gamma_c.f90#L2) I see a few possible issues with their implementation that might lead to differences:
- Ntransits and Nzetapertransit hard coded to 400 each, and npitch hard coded to 80, so will need to recompile if we want to scan resolution (from a comment it looks like ROSE uses nzetapertransit=290?)
- They don't seem to do any root finding to get bounce points, or even interpolation. They take the bounce points as the first and last zeta points in the fixed zeta grid where B < 1/lambda. I think this means they probably miss a lot of the contribution to Gamma_C near the tops of wells
- The integration (over both lambda and zeta) seems to only be a first order Reimann sum so may need super high resolution. 

In contrast, the WISTELL implementation [here](https://gitlab.com/wistell/StellaratorOptimizationMetrics.jl/-/tree/master/src/GammaC?ref_type=heads) seems a lot better
- They spline qtys along a field line using cubic splines
- bounce points are found through rootfinding on these splines
- bounce integrals are done using adaptive tanh-sinh quadrature over splines
- integral over lambda also seems to be done with adaptive gauss-kronrod, by splining gamma_c(lambda)
- seems like they also can calculate eps_eff

Might be worth reaching out to Aaron and Ben to talk about benchmarking vs WISTELL

unalmis:
> From looking through STELLOPTs version of Gamma_C [here](https://github.com/PrincetonUniversity/STELLOPT/blob/653cae047b08b7fbfe4ac0166e63247bbe2f6199/STELLOPTV2/Sources/Chisq/chisq_gamma_c.f90#L2)

Thanks! I know that's not a leisure read. I agree that implementation is problematic. The integrals are strong functions of the bounce points; so it's necessary to obtain them accurately. A specialized quadrature is also recommended. Prior to writing all the DESC stuff, I tested this (because if we could just run along the field line using a reiman sum we could use less expensive algorithms. Assuming the bounce points are known with perfect accuracy, a non-adaptive trapezoidal scheme to evaluate these bounce integrals required orders of magnitude more quadrature points. Might become a race with the floating point plateau.

This STELLOPT implementation is better: https://github.com/PrincetonUniversity/STELLOPT/blob/feature/gamma2/STELLOPTV2/Sources/Chisq/chisq_gamma_c_v2_mod.f90. They mention use of tanh-sinh, but it's been a while since I looked at it, and at the moment I don't actually see it used.

f0uriest:
Yeah I looked at that as well but it seems the same? There are more comments explaining some of the coordinate transformations but the bounce points are still rounded to the nearest grid point (there is some stuff about root finding but it's inside a "if 0 > 1" block, and the following stuff doesn't seem to use it. Also looks like it's still just doing first order integration over both zeta and pitch, there's a lone comment about tanh sinh but I didn't see anything that actually looked like it was being used

unalmis:
> the WISTELL implementation

I'll look over this more thoroughly later, but from a quick glance, some points to consider

> integral over lambda also seems to be done with adaptive gauss-kronrod, by splining gamma_c(lambda)

I see that a cubic spline is fit to the integrand, i.e. onto $f$ in $\int f(\lambda) d\lambda$, and then an adaptive quadrature is used to integrate this spline. Simpson's rule integrates a cubic spline exactly, so I don't understand that strategy.

Likewise the integrand $f(\lambda)$ is not smooth. For $\Gamma_c$, each derivative of $f(\lambda)$ is less smooth than its primitive, and $f(\lambda) \to \infty$ if it intersects a local maximum of |B| (unless there is some magical cancellation between the diverging integrals among the different ripples), so any usual function approximation on $f$ will underestimate the integral.

> bounce integrals are done using adaptive tanh-sinh quadrature over splines

FYI, some reasons I chose not to use an adaptive quadrature for the bounce integrals
1. A fixed quadrature makes the computation homogenous, and this will improve performance on GPUs/numpy/jax etc.
2. The goal is computing a double integral (1. over pitch $\lambda$ and 2. some bounce integrals). For $\Gamma_c$, in theory, my claim is an adaptive quadrature on the bounce integrals will over-fit to 2. and require more pitch values to capture the true double integral

unalmis:
Unrelated, but since we're discussing quadrature, I'm not a fan of tanh-sinh because it packs nodes too densely near bounce points. This could make floating point errors in the bounce points punishing. Nemov's effective ripple paper mentions how there's additional phenomena to capture by considering particles trapped in wells with ripples, so intuitively that means when we do these bounce integrals we should try to capture more than just the features near the bounce points (hence reduce clustering of quadrature near bounce points as much as possible once the singularity is resolved).

unalmis:
In previous commits, I had an kwarg option to perform the integral over $\lambda$ with any adaptive quadrature from `quadax`  (without splining the integrand, so the functions behavior is truly captured). My preference for this being adaptive romberg. Do we want this option back?

gretahibbard:
Fix #1288

unalmis:
@gretahibbard Thanks for this. When you get a chance could you please

- Document in #1412 as well what the issue was.
- If it's a matter of not storing things in constants, I'd prefer just doing that. Right now it looks like objective.build api was changed. Is that necessary?
- In the [last commit](https://github.com/PlasmaControl/DESC/commit/27cfacf57d485dd63f5791f891c83fe6655657e6), Gamma_d was added. I'd prefer to keep that in #1229.

desc/compute/_neoclassical.py
```diff
@@ -272,3 +278,253 @@ def eps_32(data):
...
+    ),
+    units="~",
+    units_long="None",
+    description="Energetic ion confinement proxy",
+    dim=1,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="r",
+    data=["min_tz |B|", "max_tz |B|", "cvdrift0", "gbdrift", "fieldline length"]
```
f0uriest, line 298: min/max of B are just needed to compute the bounce points right?

f0uriest: but they're not included in `Bounce1D.required_names`?

desc/compute/_neoclassical.py
```diff
@@ -272,3 +278,253 @@ def eps_32(data):
 def _effective_ripple(params, transforms, profiles, data, **kwargs):
     data["effective ripple"] = data["effective ripple 3/2"] ** (2 / 3)
     return data
+
+
+@register_compute_fun(
+    name="Gamma_c Velasco",
+    label=(
+        # Γ_c = π/(8√2) ∫dλ 〈 ∑ⱼ [v τ γ_c²]ⱼ 〉
+        "\\Gamma_c = \\frac{\\pi}{8 \\sqrt{2}} "
```
f0uriest, line 287: is this correct? I thought this was for nemovs?

desc/compute/_neoclassical.py
```diff
@@ -272,3 +278,253 @@ def eps_32(data):
...
+        "\\int d\\lambda \\langle \\sum_j (v \\tau \\gamma_c^2)_j \\rangle"
+    ),
+    units="~",
+    units_long="None",
+    description="Energetic ion confinement proxy",
```
f0uriest, line 292: maybe add `as defined by Velasco (doi:...)`

desc/objectives/_neoclassical.py
```diff
@@ -260,3 +266,267 @@ def compute(self, params, constants=None):
...
+    A model for the fast evaluation of prompt losses of energetic ions in stellarators.
+    J.L. Velasco et al. 2021 Nucl. Fusion 61 116059.
+    https://doi.org/10.1088/1741-4326/ac2994.
+    Equation 16.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    target : {float, ndarray, callable}, optional
```
f0uriest, line 291: use the docstring util to simplify a lot of this

f0uriest: (should have been done for ripple too)

desc/objectives/_neoclassical.py
```diff
@@ -260,3 +266,267 @@ def compute(self, params, constants=None):
...
+        Returns
+        -------
+        result : ndarray
+            Γ_c as a function of the flux surface label.
+
+        """
+        if constants is None:
+            constants = self.constants
+        eq = self.things[0]
+        # TODO: compute all deps of gamma here
```
f0uriest, line 497: issue reference?

tests/test_objective_funs.py
f0uriest, line 1: can you add a test comparing `eq.compute("Gamma c")` to `GammaC().compute()` (to avoid anything like #1424)

Ideally also for EffectiveRipple

desc/objectives/_neoclassical.py
unalmis: My comment is opaque, but this is resolved by the fourier bounce stuff. https://github.com/PlasmaControl/DESC/pull/1003/files#r1774068985

desc/compute/_neoclassical.py
unalmis: They are the same except for the definition of γ_c

desc/compute/_neoclassical.py
unalmis: That's right min max B determine the pitch angle domain. They are not in the required names because it's not needed to make a `Bounce1D` object

desc/compute/_neoclassical.py
```diff
@@ -272,3 +278,253 @@ def eps_32(data):
...
+            * data["pitch_inv weight"]
+        ).sum(axis=-1)
+
+    # We rewrite equivalents of Nemov et al.'s expression's using single-valued
+    # maps of a physical coordinates. This avoids the computational issues of
+    # multivalued maps. It further enables use of more efficient methods, such as
+    # fast transforms and fixed computational grids throughout optimization, which
+    # are used in the ``Bounce2D`` operator on a developer branch. Also, Nemov
+    # assumes B^ϕ > 0 in some comments; this is not true in DESC, but the
+    # computations done here are invariant to the sign.
```
unalmis, line 508: I remove statements about other's assumptions in the fourier bounce branch to be more profssional

tests/test_objective_funs.py
unalmis: I'll add one marked xfail for ripple. https://github.com/PlasmaControl/DESC/pull/1003/files#r1777837505

tests/test_neoclassical.py
```diff
@@ -99,16 +104,65 @@ def test_Gamma_c():
...
+    grid = eq._get_rtz_grid(rho, alpha, zeta, coordinates="raz")
     data = eq.compute("Gamma_c", grid=grid)
     assert np.isfinite(data["Gamma_c"]).all()
     fig, ax = plt.subplots()
     ax.plot(rho, grid.compress(data["Gamma_c"]), marker="o")
     return fig
 
 
+@pytest.mark.unit
+def test_objective_compute():
```
f0uriest, line 120: can you move this to `test_objective_funs.py::TestObjectiveFunction`?

desc/objectives/_neoclassical.py
```diff
@@ -256,7 +250,242 @@ def compute(self, params, constants=None):
...
+        self,
+        eq,
+        target=None,
+        bounds=None,
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        deriv_mode="auto",
+        *,
```
f0uriest, line 347: one other thing: we had talked about eventually making all args keyword only for objectives so that we can add new ones and re-order as needed (and so we can make the ordering in the docs match the signature). We could start doing that here, but let's discuss at the meeting tomorrow how we want to proceed with that

dpanici: I second this

f0uriest: So from the dev meeting I think we should start doing this going forward. `eq` can be positional but everything else keyword only. We can start doing it for all new objectives, eventually we'll also do it for existing objectives.

unalmis: ok i made issue

f0uriest (approved):

rahulgaur104 (approved):

Issue #1034, opened 2024-05-22
Optimizers/objectives with auxiliary output


f0uriest:
For objectives that require some sort of iterative solution, it can be helpful to store the previous solution as a warm start. Right now this isn't possible/easy due to the way our objectives and optimizers are wrapped, but might be do-able. All of the jax AD stuff has a `has_aux` option where the first output of a function is the mathemtical function being differentiated, while the 2nd is internal info like solver state that is not differentiated. We could plumb this through and add some logic to the optimizer wrappers to handle this. Might also make stuff like proximal etc a lot easier?

PR #1043, opened 2024-06-04
Remove jit method of objective, directly compile methods

### LINKED ISSUE #957: Don't close over `self` when jitting objective functions
- This is a source of bugs, since attributes of the class get baked in, so don't get properly updated if they change
- This also likely makes compilation slower, since it's trying to bake in some big constants
- Also likely a source of memory leaks

### LINKED ISSUE #1191: Excessive recompilation in `ProximalProjection`
Running with `JAX_LOG_COMPILES="1"` I see stuff like this between every iteration of the outer loop:

```
Finished tracing + transforming jac_scaled for pjit in 2.1464920043945312 sec
Compiling jac_scaled with global shapes and types []. Argument mapping: [].
Finished jaxpr to MLIR module conversion jit(jac_scaled) in 0.13467741012573242 sec
Finished XLA compilation of jit(jac_scaled) in 0.7804932594299316 sec
Finished tracing + transforming compute_scaled_error for pjit in 0.4098942279815674 sec
Compiling compute_scaled_error with global shapes and types [ShapedArray(float64[298])]. Argument mapping: [UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(compute_scaled_error) in 0.09883785247802734 sec
Finished XLA compilation of jit(compute_scaled_error) in 0.16284894943237305 sec
Finished tracing + transforming project for pjit in 0.0073757171630859375 sec
Compiling project with global shapes and types [ShapedArray(float64[298])]. Argument mapping: [UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(project) in 0.014415740966796875 sec
Finished XLA compilation of jit(project) in 0.07303285598754883 sec
Finished tracing + transforming recover for pjit in 0.01389932632446289 sec
Compiling recover with global shapes and types [ShapedArray(float64[120])]. Argument mapping: [UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(recover) in 0.0190582275390625 sec
Finished XLA compilation of jit(recover) in 0.1190042495727539 sec
Finished tracing + transforming jac_scaled for pjit in 1.3376684188842773 sec
Compiling jac_scaled with global shapes and types []. Argument mapping: [].
Finished jaxpr to MLIR module conversion jit(jac_scaled) in 0.13845014572143555 sec
Finished XLA compilation of jit(jac_scaled) in 0.7922420501708984 sec
Finished tracing + transforming compute_scaled_error for pjit in 0.38459253311157227 sec
Compiling compute_scaled_error with global shapes and types [ShapedArray(float64[298])]. Argument mapping: [UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(compute_scaled_error) in 0.10509967803955078 sec
Finished XLA compilation of jit(compute_scaled_error) in 0.15252304077148438 sec
Finished tracing + transforming recover for pjit in 0.007174253463745117 sec
Compiling recover with global shapes and types [ShapedArray(float64[120])]. Argument mapping: [UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(recover) in 0.01448822021484375 sec
Finished XLA compilation of jit(recover) in 0.06177163124084473 sec
Finished tracing + transforming jac_scaled for pjit in 1.1975443363189697 sec
Compiling jac_scaled with global shapes and types []. Argument mapping: [].
Finished jaxpr to MLIR module conversion jit(jac_scaled) in 0.14225172996520996 sec
Finished XLA compilation of jit(jac_scaled) in 0.7997219562530518 sec
Finished tracing + transforming compute_scaled_error for pjit in 0.4294285774230957 sec
Compiling compute_scaled_error with global shapes and types [ShapedArray(float64[298])]. Argument mapping: [UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(compute_scaled_error) in 0.13470077514648438 sec
Finished XLA compilation of jit(compute_scaled_error) in 0.20583796501159668 sec
Finished tracing + transforming project for pjit in 0.007157325744628906 sec
Compiling project with global shapes and types [ShapedArray(float64[298])]. Argument mapping: [UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(project) in 0.012563467025756836 sec
Finished XLA compilation of jit(project) in 0.06066417694091797 sec
Finished tracing + transforming recover for pjit in 0.012331247329711914 sec
Compiling recover with global shapes and types [ShapedArray(float64[120])]. Argument mapping: [UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(recover) in 0.01821756362915039 sec
Finished XLA compilation of jit(recover) in 0.05025196075439453 sec
```

We expect things to compile on the first iteration and then be reused but here it seems that a bunch of stuff is getting recompiled on each step. Possibly related to #957 but not sure.

f0uriest:
Basically, right now we close over `self` when compiling the methods of `ObjectiveFunction`. This means that JAX may bake all the attributes of the objective (ie, transforms, profiles, fields, equilibrium etc) into the compiled function which likely both slows down compilation and may lead to extra memory usage.

This changes things so that instead we JIT the method directly, treating `self` as just another argument. Doing this requires refactoring how the derivatives get handled a bit (they are now only local to their respective functions rather than being created separately, shouldn't be any performance hit since creating the `Derivative` objects is basically free).

Resolves #957 
Resolves #1191


dpanici:
@YigitElma @dpanici @kianorr @rahulgaur104 we should profile this change memory-wise

f0uriest:
> ```diff
>  test_build_transform_fft_lowres         |     -1.56 +/- 9.77     | -8.12e-03 +/- 5.09e-02 |  5.13e-01 +/- 2.9e-02  |  5.21e-01 +/- 4.2e-02  |
>  test_build_transform_fft_midres         |     -0.57 +/- 2.07     | -3.38e-03 +/- 1.22e-02 |  5.89e-01 +/- 9.9e-03  |  5.92e-01 +/- 7.2e-03  |
>  test_build_transform_fft_highres        |     +3.10 +/- 3.37     | +3.05e-02 +/- 3.32e-02 |  1.01e+00 +/- 2.1e-02  |  9.84e-01 +/- 2.6e-02  |
>  test_equilibrium_init_lowres            |     +3.36 +/- 8.44     | +1.32e-01 +/- 3.32e-01 |  4.07e+00 +/- 1.7e-01  |  3.94e+00 +/- 2.9e-01  |
>  test_equilibrium_init_medres            |     -1.05 +/- 4.88     | -4.42e-02 +/- 2.04e-01 |  4.15e+00 +/- 1.0e-01  |  4.19e+00 +/- 1.8e-01  |
>  test_equilibrium_init_highres           |     +1.22 +/- 3.98     | +6.80e-02 +/- 2.22e-01 |  5.65e+00 +/- 1.4e-01  |  5.59e+00 +/- 1.7e-01  |
>  test_objective_compile_dshape_current   |     +2.97 +/- 4.68     | +1.17e-01 +/- 1.85e-01 |  4.06e+00 +/- 1.7e-01  |  3.95e+00 +/- 6.1e-02  |
>  test_objective_compile_atf              |     -5.39 +/- 3.01     | -4.55e-01 +/- 2.54e-01 |  7.98e+00 +/- 7.0e-02  |  8.44e+00 +/- 2.4e-01  |
> -test_objective_compute_dshape_current   |    +177.24 +/- 4.92    | +2.22e-03 +/- 6.17e-05 |  3.48e-03 +/- 5.1e-05  |  1.25e-03 +/- 3.4e-05  |
> -test_objective_compute_atf              |    +139.12 +/- 4.43    | +5.95e-03 +/- 1.89e-04 |  1.02e-02 +/- 1.1e-04  |  4.27e-03 +/- 1.5e-04  |
>  test_objective_jac_dshape_current       |     -0.65 +/- 8.29     | -2.51e-04 +/- 3.19e-03 |  3.82e-02 +/- 1.8e-03  |  3.85e-02 +/- 2.6e-03  |
>  test_objective_jac_atf                  |     +1.18 +/- 4.31     | +2.19e-02 +/- 8.03e-02 |  1.88e+00 +/- 3.7e-02  |  1.86e+00 +/- 7.1e-02  |
>  test_perturb_1                          |     -9.89 +/- 5.29     | -1.38e+00 +/- 7.40e-01 |  1.26e+01 +/- 3.2e-01  |  1.40e+01 +/- 6.7e-01  |
>  test_perturb_2                          |     -3.33 +/- 2.48     | -6.19e-01 +/- 4.61e-01 |  1.80e+01 +/- 4.5e-01  |  1.86e+01 +/- 1.1e-01  |
>  test_proximal_jac_atf                   |     +1.05 +/- 1.09     | +7.65e-02 +/- 7.96e-02 |  7.35e+00 +/- 6.6e-02  |  7.28e+00 +/- 4.4e-02  |
> -test_proximal_freeb_compute             |     +3.08 +/- 0.75     | +5.44e-03 +/- 1.33e-03 |  1.82e-01 +/- 9.9e-04  |  1.76e-01 +/- 8.9e-04  |
>  test_proximal_freeb_jac                 |     -0.08 +/- 0.95     | -5.65e-03 +/- 6.95e-02 |  7.35e+00 +/- 5.6e-02  |  7.35e+00 +/- 4.2e-02  |
> +test_solve_fixed_iter                   |    -60.47 +/- 16.46    | -1.08e+01 +/- 2.94e+00 |  7.06e+00 +/- 2.9e+00  |  1.78e+01 +/- 5.7e-01  |
> ```

I think the slowdown in the `compute` benchmark is because it has to flatten/unflatten the `ObjectiveFunction` pytree at each call. It looks like a big slowdown but in absolute time its only a few ms, so likely worth it to get a big speedup on actual solve/optimization (that said I think 60% faster solve is higher than what I'm seeing in practice, it's usually more like 10-20% faster, so might be unique to the benchmark case)

desc/objectives/objective_funs.py
```diff
@@ -63,62 +70,9 @@ def _set_derivatives(self):
                 self._deriv_mode = "batched"
```
YigitElma, line 70: With these deletions, do we still need this function? Maybe just move the code to init. If not, change the documentation since it doesn't set derivatives but sets the method to calculate derivatives.

desc/objectives/objective_funs.py
```diff
@@ -816,10 +848,6 @@ def __init__(
 
     def _set_derivatives(self):
```
YigitElma, line 849: Same comment as above, I would prefer having this in init function.

desc/objectives/objective_funs.py
```diff
@@ -390,54 +342,128 @@ def unpack_state(self, x, per_objective=True):
...
         """Compute Hessian matrix of self.compute_scalar wrt x."""
         if constants is None:
             constants = self.constants
-        return jnp.atleast_2d(self._hess(x, constants).squeeze())
+        return jnp.atleast_2d(
+            Derivative(self.compute_scalar, mode="hess")(x, constants).squeeze()
+        )
+
+    def _jac(self, op, x, constants=None):
+        # could also do something similar for grad and hess, but probably not
```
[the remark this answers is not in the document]
YigitElma: And I know this is the same code as before, but it was always been a bit complicated.

desc/objectives/utils.py
```diff
@@ -142,17 +143,8 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
...
-        """Project a full state vector into the reduced optimization vector."""
-        x_reduced = Z.T @ ((x - xp)[unfixed_idx])
-        return jnp.atleast_1d(jnp.squeeze(x_reduced))
-
-    @jit
-    def recover(x_reduced):
-        """Recover the full state vector from the reduced optimization vector."""
-        dx = put(jnp.zeros(objective.dim_x), unfixed_idx, Z @ x_reduced)
-        return jnp.atleast_1d(jnp.squeeze(xp + dx))
+    project = _Project(Z, xp, unfixed_idx)
```
YigitElma, line 146: This is just not to jit these functions until we use them, right? And maybe to save the null-space etc?

desc/optimize/_constraint_wrappers.py
```diff
@@ -1094,3 +1069,48 @@ def constants(self):
     def __getattr__(self, name):
         """For other attributes we defer to the base objective."""
         return getattr(self._objective, name)
+
+
+# in ProximalProjection we have an explicit state that we keep track of (and add
+# to as we go) meaning if we jit anything with self static it doesn't update
+# correctly, while if we leave self unstatic then it recompiles every time because
+# the pytree structure of ProximalProjection is changing. To get around that we
```
YigitElma, line 1077: So, for example ForceBalance constraint changes in terms of value but not the structure so, it doesn't recompile everytime. but do to appended state to the self, structure of ProximalProjection changes? Do I understand correctly?

YigitElma (changes requested):
- I woud prefer to get rid of `set_derivatives` function completely since it basically does nothing now.
- For blocked jacobian, extra explanation would be good

desc/objectives/utils.py
f0uriest: No it's more about avoiding recompilation. Right now every time you call `factorize_linear_constraints` you get a new `project/recover` function which gets compiled each time, since its a new function each time. By making them classes with jitted methods the compile cache is shared between class instances so it will only compile once (assuming size/shape doesnt change etc)

desc/optimize/_constraint_wrappers.py
f0uriest: yes exactly. In theory we might be able to get around with with #1034 but that's future work.

desc/objectives/objective_funs.py
f0uriest: I think we don't want to put this in init because it needs to be called after the sub-objectives get built, otherwise some of the logic for picking forward or reverse mode based on size of outputs etc won't work correctly.

desc/objectives/objective_funs.py
f0uriest: see comment above

desc/objectives/utils.py
YigitElma: ah ok makes sense

desc/objectives/objective_funs.py
YigitElma: Ok I am fine with new documentation

YigitElma (dismissed):

desc/objectives/objective_funs.py
```diff
@@ -136,17 +90,13 @@ def jit(self):  # noqa: C901
             "vjp_scaled_error",
             "vjp_unscaled",
         ]
-
         for method in methods:
             try:
-                delattr(self, method)
+                setattr(
+                    self, method, functools.partial(getattr(self, method)._fun, self)
```
dpanici, line 96: I was going to ask why grad jac etc had the `Derivative` call inside of them now, but this functools.partial call basically partially evaluates the method with`self`, which also creates the Derivative and sets up the correct AD method, so that subsequent calls are not re-creating the grad/jac function again, but rather just using it?

tests/test_objective_funs.py
```diff
@@ -1534,7 +1534,7 @@ def test_boundary_error_print(capsys):
     obj = VacuumBoundaryError(eq, coilset, field_grid=coil_grid)
     obj.build()
 
-    f = np.abs(obj.compute_unscaled(*obj.xs(eq)))
+    f = np.abs(obj.compute_unscaled(*obj.xs(eq, coilset)))
```
dpanici, line 1537: was the fact that the old way worked before somewhat unintentional?

dpanici (approved):
just had some questions but nothing on the content

desc/objectives/objective_funs.py
f0uriest: This code is only for **un**jitting something (ie if `use_jit=False`) in which case it will effectively create a new `Derivative` object on each call (`partial` doesn't actually evaluate anything when its created).

The default (when jitted) is that the creation of the `Derivative` object basically gets compiled away by jax so its not costing anything. And even if it's created each time instantiating the object is negligible compared to actually computing the derivative.

tests/test_objective_funs.py
f0uriest: yeah seems like it was an uncaught bug.

ddudt (approved):

YigitElma (approved):

PR #1046, opened 2024-06-07
TERPSICHORE

ddudt:
New objective function to wrap the TERPSICHORE stability code


dpanici:
put in separate experimental folder

docs/external_objectives.rst
```diff
@@ -0,0 +1,318 @@
...
+It was adapted from the test ``test_external_vs_generic_objectives`` in
+``tests/test_examples.py``.
+
+The function must take a single positional argument, which can be either a single
+Equilibrium or a list of Equilibria. Additional inputs can be passed as keyword
+arguments. In this example, the function returns the three scalar quatities
+:math:`\langle\beta\rangle_{vol}`, :math:`\langle\beta\rangle_{pol}`, and
+:math:`\langle\beta\rangle_{tor}`. It writes these quantities to a NetCDF file and then
+reads them back from the file. This mimics I/O from the common VMEC "wout" format, and
+also demonstrates an operation that is not automatically differentiable.
```
dpanici, line 30:
```suggestion
also demonstrates an operation that is not automatic-differentiable.
```

docs/external_objectives.rst
```diff
@@ -0,0 +1,318 @@
...
+* ``vectorized``, whether ``fun`` expects a single Equilibrium or a list of
+  Equilibria. Since the function ``beta_from_file`` takes a single Equilibrium as its
+  only positional argument, ``vectorized=False``.
+
+All other input parameters that the external function requires can be passed as keyword
+arguments to ``ExternalObjective``. In this example, ``path`` specifies the file name
+of the NetCDF file that the function writes/reads. Since ``vectorized=False``, the
+function ``beta_from_file`` will be evaluated sequentially when computing the finite
+differences (rather than in parallel), so the same file will be overwritten each time
+for the different Equilibria.
```
dpanici, line 69: the different equilibria part is bc for the finite diffs, a different equilibrium is solved with slightly perturbed bdry and passed to here right? might want to make that explicit

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+                    exec=exec,
+                    scalar=scalar,
+                    surfs=surfs,
+                    al0=al0,
+                    sleep_time=sleep_time,
+                    stop_time=stop_time,
+                ),
+                idxs,
+            )
+            pool.close()
```
dpanici, line 123: I am pretty sure the `with` statement will call `pool.terminate` which is `pool.close` + `pool.join` when it exits the block, so you could probably remove these last three lines in the block and instead just have the last `result=` line be outside of the `with` block

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+    N_max=4,
+    M_booz_max=19,
+    N_booz_max=18,
+    awall=2.0,
+    deltaJp=1e-2,
+    modelk=1,
+    nev=1,
+    al0=-5e-1,
+    sleep_time=1,
+    stop_time=60,
```
dpanici, line 45: what is sleep_time and stop_time in? seconds?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+def _write_wout(eq, path, surfs, data_transforms, fit_transform):  # noqa: C901
+    """Write the wout NetCDF file from the equilibrium."""
+    # this is a lightweight version of VMECIO.save
+    file = Dataset(path, mode="w", format="NETCDF3_64BIT_OFFSET")
+
+    Psi = eq.Psi
+    NFP = eq.NFP
+    M = eq.M
+    N = eq.N
+    M_nyq = M + 4
```
dpanici, line 142: these are the default values that VMEC has, but tbh it is probably better to increase these from their defaults or add an option to the objective to pass in the nyquist resolutions if desired. 

I doubt that M+4 is sufficient to describe sqrt(g) even, as that is a product of 3 Fourier series of M,N each. Might be worth a convergence test at differing nyquist resolutions to see how much of an effect this has

dpanici: and in VMEC, if the eq was solved with NTHETA and NZETA set, that is what sets the nyquist resolutions, so thuogh M+4 and N+2 are the defaults, people may be using higher values when they know they are gonna pass to Terpsichore, or if they know they need higher resolution for stuff like free bdry.

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+    )
+    x_mn = np.zeros((surfs - 1, m.size))
+    for i in range(surfs - 1):
+        x_mn[i, :] = fit(data[i, :])
+    xm, xn, s, c = ptolemy_identity_rev(m, n, x_mn)
+    gmnc[0, :] = 0
+    gmnc[1:, :] = -c  # negative sign for negative Jacobian
+    if not eq.sym:
+        gmns[0, :] = 0
+        gmns[1:, :] = -s
```
dpanici, line 344: I thought we only need to negate one of these (cos)? since we negate the sign modes bc the pol. angle changed direction and we had fit it with the DESC poloidal angle, not the VMEC one, and we also negate entire g because the jacobian now has a sign change due to the reverse of the poloidal angle

so it is like gmnc = - c and gmns = - (-s)?

dpanici: maybe I am misunderstanding since I see we have this in the VMECIO.save as well.

dpanici: nvm, the missing - is in fit()

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+    f.write(boz_str_title)
+
+    for n in range(-N_booz_max, N_booz_max + 1):
+        if n < 0:
+            f.write(boz_str_neg + "{:>3}\n".format(n))
+        else:
+            f.write(boz_str_pos + "{:>3}\n".format(n))
+
+    f.write("C\n")
+    f.write("      LLAMPR      LVMTPR      LMETPR      LFOUPR\n")
```
dpanici, line 449: if you konw what these are can you add comments to our code here explaining? or a link to any docs that TERPSICHORE may have to find these definitions?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+    f.write("C     TABLE OF FOURIER COEFFIENTS FOR STABILITY DISPLACEMENTS\n")
+    f.write("C\n")
+    f.write(
+        "C M=  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 "
+        + "0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5  N\n"
+    )
+
+    for n in range(N_min, N_max + 1):
+        mode_str = "     "
+        for m in range(56):
```
dpanici, line 502: 56 is what?

dpanici: should it be an input for this to change like a resolution?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+            if mode_family < 0 or n % 2 == mode_family:  # FIXME: modify for nfp != 2,3
```
dpanici, line 503: what is the modification needed? can you make it in this PR?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+
+    f.write("C\n")
+    f.write("C   NEV NITMAX         AL0     EPSPAM IGREEN MPINIT\n")
+    f.write("      {}   4500  {:10.3e}  1.000E-04      0      0\n".format(nev, al0))
+    f.write("C\n")
+
+    f.close()
+
+
+def _pool_fun(k, path, exec, scalar, surfs, al0, sleep_time, stop_time):
```
dpanici, line 528: what is scalar?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+        Total step size is ``abs_step + rel_step * mean(abs(x))``.
+    processes : int, optional
+        Maximum number of CPU threads to use for multiprocessing. Default = 1.
+    path : str, optional
+        Path to the directory where temporary files will be stored.
+    exec : str, optional
+        File name of the TERPSICHORE executable. Must be located in the directory
+        specified by ``path``.
+    scalar : bool, optional
+        If Ture, returns the growth rate of the fastest growing instability (default).
```
dpanici, line 671:
```suggestion
        If True, returns the growth rate of the fastest growing instability (default).
```

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+
+    return growth_rate[0] if scalar else delta_W
+
+
+class TERPSICHORE(ExternalObjective):
+    """Computes ideal MHD linear stability from calls to the code TERPSICHORE.
+
+    Returns the linear growth rate of the fastest growing instability, or the ΔW values
+    at each flux surface. A positive growth rate or negative ΔW denotes instability.
+
```
dpanici, line 652: might be good to cite it as well

https://link.springer.com/chapter/10.1007/978-1-4613-0659-7_8

docs/external_objectives.rst
```diff
@@ -0,0 +1,318 @@
...
+        # run TERPSICHORE on list of equilibria in parallel
+        if len(eq) == 1:  # no multiprocessing if only one equilibrium
+            result = jnp.atleast_1d(_pool_fun(0, path=tmp_path, exec=exec))
+        else:  # use multiprocessing if there are multiple equilibria
+            with mp.Pool(processes=min(processes, len(eq))) as pool:
+                results = pool.map(
+                    functools.partial(_pool_fun, path=tmp_path, exec=exec),
+                    idxs,
+                )
+                pool.close()
```
dpanici, line 173: if you change the with block in the source, change it here too

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+        (determined by ``M_booz_max`` and ``N_booz_max``). If TERPSICHORE fails to run,
+        try increasing this parameter. Default = 1000.
+    lssd : int, optional
+        Minimum number of possible permutations of stability mode combinations
+        (determined by ``M_max`` and ``N_max``). If TERPSICHORE fails to run,
+        try increasing this parameter. Default = 1000.
+    M_max : int, optional
+        Maximum poloidal mode number of stability modes to consider. Will include modes
+        with m ∈ [0,M_max] (if ``mode_family < 0``). Default = 8.
+    N_min : int, optional
```
dpanici, line 691: is this mode number the mode structure of the instability within one field period?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+        Minimum toroidal mode number of stability modes to consider. Will include modes
+        with n ∈ [N_min,N_max] (if ``mode_family < 0``). Default = -4.
+    N_max : int, optional
+        Maximum toroidal mode number of stability modes to consider. Will include modes
+        with n ∈ [N_min,N_max] (if ``mode_family < 0``). Default = 4.
+    M_booz_max : int, optional
+        Maximum poloidal mode number of Boozer spectrum. Will include modes with
+        m ∈ [0,M_booz_max]. Default = 19.
+    N_booz_max : int, optional
+        Maximum poloidal mode number of Boozer spectrum. Will include modes with
```
dpanici, line 701: is this the N within one field period?

docs/external_objectives.rst
```diff
@@ -0,0 +1,318 @@
...
+other codes. The code shown here is abbreviated and slightly modified for brevity, but
+the full source code can be found in ``desc/experimental/_terpsichore.py``.
+
+The external function in this case is named ``terpsichore``, and takes the following
+arguments:
+
+* ``eq``, a list of Equilibria.
+* ``processes``, the maximum number of processes to use for multiprocessing.
+* ``path``, the path to the directory where the optimization script is being executed.
+* ``exec``, the file name of the TERPSICHORE executable.
```
dpanici, line 116: this should be the file path to the executable right? not just filename

dpanici (changes requested):
some clarifying comments requested

docs/external_objectives.rst
```diff
@@ -0,0 +1,318 @@
...
+transformable, and wraps it with forward finite differences. This is most useful for
+cases when the function needs to call external codes that are written in another
+language besides Python, or cannot be rewritten with JAX for some other reason. If the
+user function can be made JAX transformable, it is recommended to use the
+``ObjectiveFromUser`` class instead.
+
+This guide begins with a simple example of how to use a custom function with the
+``ExternalObjective``. It then uses the ``TERPSICHORE`` objective as a more realistic
+example, which can be used as a template for wrapping other codes. Finally, an example
+script is shown for how to run the TERPSICHORE optimization with multiprocessing.
```
dpanici, line 15: add a note that one must obtain the TERPSICHORE source themselves and that DESC does not have the code.

dpanici: either add note here or later in this doc

desc/experimental/_terpsichore.py
```diff
@@ -707,7 +707,7 @@ class TERPSICHORE(ExternalObjective):
         plasma. If TERPSICHORE fails to run, try decreasing this parameter. Default = 2.
     deltaJp : float, optional
         Resonance detuning parameter to resolve parallel current density singularities.
-        Default = 0.5.
```
dpanici: is there any intuition to be gained from understanding this default? like why did you lower it?

ddudt: TERPSICHORE documentation recommends a value in the range 1E-4 to 4E-2, I just accidentally had the default way too high. 

The inputs `awall` and `deltaJp` both artificially stabilize the plasma but make the numerics better, so typically you need to do a convergence scan in both parameters and the "best" value is problem dependent.

desc/experimental/_terpsichore.py
ddudt: Update this to reflect the changes made in #1543

docs/external_objectives.rst
```diff
@@ -0,0 +1,318 @@
+Using external codes
+--------------------
+
+The ``ExternalObjective`` class allows external codes to be used within a DESC
+optimization. This takes a user-supplied function that does not need not be JAX
```
YigitElma, line 5:
```suggestion
optimization. This takes a user-supplied function that does not need to be JAX
```

docs/external_objectives.rst
```diff
@@ -0,0 +1,318 @@
...
+transformable, and wraps it with forward finite differences. This is most useful for
+cases when the function needs to call external codes that are written in another
+language besides Python, or cannot be rewritten with JAX for some other reason. If the
+user function can be made JAX transformable, it is recommended to use the
+``ObjectiveFromUser`` class instead.
+
```
YigitElma, line 11: Might be good to restate `desc/experimental/README` here

docs/external_objectives.rst
```diff
@@ -0,0 +1,318 @@
...
+
+TERPSICHORE objective
+---------------------
+
+This section walks through the implementation of wrapping the ideal MHD linear stability
+code TERPSICHORE, which is written in FORTRAN. It summarizes how the external function
+was written to call TERPSICHORE from DESC, and can be used as a template for wrapping
+other codes. The code shown here is abbreviated and slightly modified for brevity, but
+the full source code can be found in ``desc/experimental/_terpsichore.py``.
```
YigitElma, line 108: Agree with Dario, this sentence sounds like we have the source code of TERPSICHORE (or this could be my ignorance)

docs/index.rst
```diff
@@ -82,6 +80,7 @@
    dev_guide/notebooks/grid.ipynb
    dev_guide/adding_compute_funs.rst
    dev_guide/adding_objectives.rst
+   dev_guide/external_objectives.rst
```
YigitElma, line 83:
```suggestion
   external_objectives.rst
```
or move it to `dev_guide/`

docs/external_objectives.rst
ddudt: The `ExternalObjective` is tested and fully supported. The warning in `experimental/README` is only because the external codes are not supported by us.

desc/experimental/_terpsichore.py
ddudt: This is the max poloidal resolution that TERPSICHORE supports. It's hard-coded here to create the full table of all possible poloidal modes, but the input `M_max` determines which of those modes actually get used.

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+    ferr.close()
+
+
+def _is_terps_complete(path, run_time, stop_time, terps_subprocess):
+    """Check if TERPSICHORE is finished running."""
+    if not os.path.exists(path):
+        return False
+
+    errorif(
+        run_time > stop_time, RuntimeError, "TERPS was unable to find a growth rate!"
```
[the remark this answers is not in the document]
ddudt: TERPSICHORE does not give useful error messages -- if something goes wrong the code just stops. The expected run time here is really just a way to check if it ran successfully or not. So in most cases this is actually signaling that TERPSICHORE failed to converge, not that it was just taking too long to run. I don't know of another easy way to handle this.

desc/experimental/_terpsichore.py
ddudt: No, it is the N for the full torus

docs/external_objectives.rst
ddudt: It is only the filename. The real documentation for the `TERPSICHORE` objective explains that `exec` "must be located in the directory specified by `path`". But I am being succinct here since it is just a tutorial template.

dpanici (changes requested):
in the docstring of ExternalObjective, can you change the type of `fun_kwargs` to be a `dict` like it is expected to be?

docs/external_objectives.rst
ddudt: The `ExternalObjective`/`TERPSICHORE` objectives don't "re-solve" the equilibrium, they only evaluate the objective for the equilibrium parameters they are passed. Just like how `QuasisymmetryBoozer` or any other objectives don't solve the equilibrium themselves either. If you want to make sure it is a good equilibrium, you need to include the `ForceBalance` objective as a constraint like usual.

desc/experimental/_terpsichore.py
ddudt: I think I copied this from some example code and I know it works, so I'll keep it as is unless you feel strongly about changing it.

desc/experimental/_terpsichore.py
ddudt: Yes, seconds. It's documented in the `TERPSICHORE` objective function docstring (that is the part that is visible to the user, this is just an internal intermediate function).

desc/experimental/_terpsichore.py
ddudt: I've added a reference to this documentation page to the `TERPSICHORE` objective function: 
https://princetonuniversity.github.io/STELLOPT/TERPSICHORE.html

Most of these logical flags are either unused or simply control what gets printed during the calculation, so they don't have an actual effect on the result. Hence I hard-coded them.

desc/experimental/_terpsichore.py
ddudt: In other words, `M_max` cannot be greater than 56

desc/experimental/_terpsichore.py
ddudt: I have removed this, I was testing a method to optimize the dW values on each surface instead of the scalar growth rate.

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,872 @@
...
+                break  # NOTE: not including vacuum surfaces
+    file.close()
+
+    errorif(len(growth_rate) == 0, RuntimeError, "Growth rate not found!")
+    errorif(
+        len(growth_rate) > 1,
+        NotImplementedError,
+        "Not capable of handling multiple growth rates.",
+    )
+    errorif(len(delta_W) != surfs - 1, RuntimeError, "Incorrect number of dW values!")
```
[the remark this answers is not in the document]
ddudt: this is now obsolete

ddudt:
> Should there be a test tto check if the interface works? Maybe have a dummy function instead of the actual TERPSICHORE solver?

This is basically what the `ExternalObjective` and its tests already do.

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+                    stop_time=stop_time,
+                ),
+                idxs,
+            )
+            pool.close()
+            pool.join()
+            result = jnp.vstack(results, dtype=float)
+
+    # remove temporary directory
+    shutil.rmtree(tmp_path)
```
f0uriest: might be better to leave it and just overwrite? That way if something breaks in terpsichore the last input data will be available for inspection

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+    modelk=1,
+    nev=1,
+    al0=-5e-1,
+    sleep_time=1,
+    stop_time=60,
+    data_transforms=None,
+    fit_transform=None,
+    theta0_outboard=True,
+):
+    """TERPSICHORE driver function."""
```
f0uriest: I think its worth giving this a full docstring

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+    rel_step : float, optional
+        Relative finite difference step size. Default = 0.
+        Total step size is ``abs_step + rel_step * mean(abs(x))``.
+    processes : int, optional
+        Maximum number of CPU threads to use for multiprocessing. Default = 1.
+    path : str, optional
+        Path to the directory where temporary files will be stored.
+    exec : str, optional
+        File name of the TERPSICHORE executable. Must be located in the directory
+        specified by ``path``.
```
f0uriest: shouldn't these be required args?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+    # TERPSICHORE assumes theta=0 is on the outboard midplane
+    if not theta0_outboard:
+        eq = flip_theta(eq)
+
+    # create temporary directory to store I/O files
+    tmp_path = os.path.join(path, "tmp-TERPS")
+    os.mkdir(tmp_path)
+
+    # write input files for each equilibrium in serial
+    idxs = list(range(len(eq)))  # equilibrium indices
```
f0uriest, line 157: does this work for a singe eq? (not a list?)

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,320 @@
...
+
+    from desc.objectives import ExternalObjective, ObjectiveFunction
+
+    objective = ObjectiveFunction(
+        ExternalObjective(
+            eq=eq,
+            fun=beta_from_file,
+            dim_f=3,
+            vectorized=False,
+            path="path/to/file.nc",
```
f0uriest: this should be in `fun_kwargs={"path":...`

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,320 @@
...
+following other arguments:
+
+* ``fun``, the external function. In this example it is the function ``beta_from_file``.
+* ``dim_f``, the dimension of the output of the function ``fun``. In this example
+  ``dim_f=3`` because ``beta_from_file`` returns an array of size 3.
+* ``vectorized``, whether ``fun`` expects a single Equilibrium or a list of
+  Equilibria. Since the function ``beta_from_file`` takes a single Equilibrium as its
+  only positional argument, ``vectorized=False``.
+
+All other input parameters that the external function requires can be passed as keyword
```
f0uriest: should be in `fun_kwargs` and function is called as `fun(eq, **fun_kwargs)`

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+        lssd=1000,
+        M_max=8,
+        N_min=-4,
+        N_max=4,
+        M_booz_max=19,
+        N_booz_max=18,
+        awall=2.0,
+        deltaJp=1e-2,
+        modelk=0,
+        nev=1,
```
f0uriest: is dim_f=1 even if this is larger?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+    N_nyq=None,
+    lssl=1000,
+    lssd=1000,
+    M_max=8,
+    N_min=-4,
+    N_max=4,
+    M_booz_max=19,
+    N_booz_max=18,
+    awall=2.0,
+    deltaJp=1e-2,
```
mfmart: Does it make sense to not have a default for deltaJp? Considering how much it can affect the growth rate, forcing the user to specify it requires some extra thinking about what should be used.

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+    modelk=1,
+    nev=1,
```
mfmart: Minor point - I don't think nev > 1 is actually functional (at least with the modern version of the code according to Anthony). If there aren't plans to add this functionality, maybe this argument should be removed and nev=1 should be hard-coded?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+                    mode_str += " 0"
+            else:
+                mode_str += " 0"
+
+        mode_str += "{:>3}\n".format(n)
+        f.write(mode_str)
+
+    f.write("C\n")
+    f.write("C   NEV NITMAX         AL0     EPSPAM IGREEN MPINIT\n")
+    f.write("      {}   4500  {:10.3e}  1.000E-04      0      0\n".format(nev, al0))
```
mfmart: Hard-code in 'nev=1' here?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+    processes : int, optional
+        Maximum number of CPU threads to use for multiprocessing. Default = 1.
+    path : str, optional
+        Path to the directory where temporary files will be stored.
+    exec : str, optional
+        File name of the TERPSICHORE executable. Must be located in the directory
+        specified by ``path``.
+    mode_family : int, optional
+        The mode family of the instabilities to consider. Possible mode families are in
+        the range ``[0, eq.NFP // 2]``. If ``mode_family < 0``, all mode families are
```
mfmart: Maybe this isn't the place for it, but it would be nice to explicitly specify the toroidal modes considered in a mode family / NFP configuration.

n_i=i*NFP \pm k, where i=...,-1,0,+1,... and k∈[0,mode_family]

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+    surfs : int, optional
+        Number of surfaces to include in the equilibrium input. More surfaces provides
+        more accuracy at the cost of longer compute times. Default = 101.
+    M_nyq, N_nyq: int
+        The max poloidal and toroidal mode numbers to use in the Nyquist spectrum of the
+        equilibrium input. Defaults to ``eq.M + 4`` and ``eq.N + 2``.
+    lssl : int, optional
+        Minimum number of possible permutations of Boozer mode combinations
+        (determined by ``M_booz_max`` and ``N_booz_max``). If TERPSICHORE fails to run,
+        try increasing this parameter. Default = 1000.
```
mfmart: I thought LSSL and LSSD needed to be larger than 1000 in almost all cases of interest. Do you usually run with 1000 for both?

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+    deltaJp : float, optional
+        Resonance detuning parameter to resolve parallel current density singularities.
+        Default = 0.01.
+    modelk : int, optional
+        0 = Noninteracting anisotropic fast particle stability model with reduced
+        kintetic energy. 1 = Kruskal-Oberman anisotropic energy principle model with
+        reduced kinetic energy. 2 = Noninteracting anisotropic fast particle stability
+        model with physical kinetic energy. 3 = Kruskal-Oberman anisotropic energy
+        principle model with physical kinetic energy. Default = 0.
+    nev : int, optional
```
mfmart: See above comment

desc/experimental/_terpsichore.py
```diff
@@ -0,0 +1,887 @@
...
+        m ∈ [0,M_booz_max]. Default = 19.
+    N_booz_max : int, optional
+        Maximum poloidal mode number of Boozer spectrum. Will include modes with
+        n ∈ [-N_booz_max,N_booz_max]. Default = 18.
+    awall : float, optional
+        Ratio of the radius of the conformal conducting wall to the plasma minor radius.
+        The conducting wall is obtained by scaling the m ≠ 0 Fourier components of the
+        plasma boundary by ``awall``. A shorter wall offset will help stabilize the
+        plasma. If TERPSICHORE fails to run, try decreasing this parameter. Default = 2.
+    deltaJp : float, optional
```
mfmart: See above comment

desc/experimental/_terpsichore.py
dpanici: make these required args then package into kwargs internally once the user passes them in

keyword-only args without default

desc/experimental/_terpsichore.py
ddudt: Unfortunately TERPSICHORE will throw an error if old output files exist in the directory when it tries to run again. I suppose we could remove the files at the beginning instead of at the end? That would be messier logic though. It also doesn't give any useful error messages so they wouldn't be very helpful for debugging anyways.

desc/external/README.rst
```diff
@@ -0,0 +1,12 @@
...
+These objectives depend on other codes external to DESC. They are not routinely tested
+and may not be compatible with all versions of the codes. The DESC team is not
+responsible for maintaining or documenting those external codes, and we do not guarantee
+to regularly maintain these objectives. Use these at your own risk!
+
+The external codes are not included with DESC, and you must obtain access to them to use
+these objectives. Those codes may not be publicly available and may require a license.
+
+`_terpsichore.py` was last tested on August 27, 2025 with DESC v0.15.0 and a version of
+TERPSICHORE built for Thea Energy in July of 2025.
```
ddudt: Note to self: check this again before merging and update accordingly

desc/experimental/_terpsichore.py
ddudt: No, in this case it must be a list of eq (although it could be a list of length 1). This is already documented in `ExternalObjective` because of how we set it up to work with vectorization/parallelization.

desc/experimental/_terpsichore.py
ddudt: Really these shouldn't need to be inputs, since they are set by the other `M` and `N` resolution input variables. Do you know how the actual number of permutations are determined in TERPSICHORE? I've changed the default to a heuristic that should be high enough, but it would be nice to have the real formula if we can find it.

desc/experimental/_terpsichore.py
ddudt: I think I prefer having a default value because it at least gives the user a reference value with an approximate scale to use. I've changed the default to be `deltaJp=1e-4` so that it is more conservative, and added more detail to the doc string. Should I go even smaller? 

This same discussion applies to `awall` as well. The value the user chooses has a big effect on the growth rate, but I think it's best to give a conservative value (one that is not too stabilizing).

desc/experimental/_terpsichore.py
dpanici: make flag `store_data=True` determining if stores all files or deletes as it goes

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,320 @@
+====================
+Using external codes
```
YigitElma, line 2: May I suggest generalizing the title to something like `Using External Codes and AD-incompatible Objectives`? Something on that front. 

It would be helpful because getting used to the JAX mindset to write your objectives to use AD is hard for most users. `ExternalObjectives` class automates a lot of things for user to achieve finite-difference for an objective. This might be easily a reason for a user to stop using DESC (although we promote DESC as easy to use, it is hard to expand it!). So, with this class, one can use their own codes in combination with DESC objectives with DESC optimizers and other stuff we offer.

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,320 @@
...
+which scales with the number of optimization degrees of freedom. Evaluating the
+TERPSICHORE growth rates for each Equilibrium can be performed in parallel on different
+CPU threads using Python multiprocessing. Note that writing the equilibria data in step
+2 cannot be easily parallelized, since it involves computations using JAX that has
+issues with multiprocessing.
+
+::
+
+    # TERPSICHORE only runs on a CPU, but DESC is optimized to run on a GPU.
+    # This decorator will run this function on a CPU, even if other functions are being
```
YigitElma: I think this is only true because of `jaxify`? For example, you cannot put `@execute_on_cpu` to `ForceBalance.compute` when there are other objectives and the whole `ObjectiveFunction` is jitted. This is similar to https://github.com/jax-ml/jax/discussions/18789#discussioncomment-8487343

YigitElma: So, this decorator might be redundant since `jaxify` calls `pure_callback`, which specifically runs on CPU.

YigitElma (commented):
I will continue on reviewing later today.

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+from desc.compute.utils import get_transforms
+from desc.grid import LinearGrid
+from desc.objectives import ExternalObjective
+from desc.objectives.objective_funs import collect_docs
+from desc.transform import Transform
+from desc.utils import errorif, warnif
+from desc.vmec_utils import ptolemy_identity_rev, zernike_to_fourier
+
+
+@execute_on_cpu
```
YigitElma: As I mentioned below, I think this is redundant, and the `jaxify` wrapper should ensure the execution on the CPU. Probably doesn't cause any problems for performance etc but may cause misconception about use of `@execute_on_cpu`.

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+        Ratio of the radius of the conformal conducting wall to the plasma minor radius.
+        The conducting wall is obtained by scaling the m ≠ 0 Fourier components of the
+        plasma boundary by ``awall``. A shorter wall offset will help stabilize the
+        plasma. If TERPSICHORE fails to run, try decreasing this parameter. Default = 2.
+    deltaJp : float
+        Resonance detuning parameter to resolve parallel current density singularities.
+        A larger value can artificially improve the stability. Default = 1e-4.
+    modelk : int, optional
+        0 = Noninteracting anisotropic fast particle stability model with reduced
+        kintetic energy. 1 = Kruskal-Oberman anisotropic energy principle model with
```
YigitElma:
```suggestion
        kinetic energy. 1 = Kruskal-Oberman anisotropic energy principle model with
```

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+    ferr.close()
+
+
+def _is_terps_complete(path, run_time, stop_time, terps_subprocess):
+    """Check if TERPSICHORE is finished running."""
+    if not os.path.exists(path):
+        return False
+
+    errorif(
+        run_time > stop_time, RuntimeError, "TERPS was unable to find a growth rate!"
```
YigitElma: Maybe add `cannot find solution in supplied stop_time ({stop_time:.2f} seconds)`

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+        Ratio of the radius of the conformal conducting wall to the plasma minor radius.
+        The conducting wall is obtained by scaling the m ≠ 0 Fourier components of the
+        plasma boundary by ``awall``. A shorter wall offset will help stabilize the
+        plasma. If TERPSICHORE fails to run, try decreasing this parameter. Default = 2.
+    deltaJp : float
+        Resonance detuning parameter to resolve parallel current density singularities.
+        A larger value can artificially improve the stability. Default = 1e-4.
+    modelk : int, optional
+        0 = Noninteracting anisotropic fast particle stability model with reduced
+        kintetic energy. 1 = Kruskal-Oberman anisotropic energy principle model with
```
YigitElma:
```suggestion
        kinetic energy. 1 = Kruskal-Oberman anisotropic energy principle model with
```

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+            ValueError,
+            "Invalid mode family, must be <= eq.NFP // 2.",
+        )
+
+        # check if theta needs to be flipped so that theta=0 is on the outboard midplane
+        grid0 = LinearGrid(rho=0.0, M=0, N=0)
+        grid1 = LinearGrid(rho=1.0, M=0, N=0)
+        R0 = self._eq.compute("R", grid0)["R"][0]  # R(rho=0)
+        R1 = self._eq.compute("R", grid1)["R"][0]  # R(rho=1,theta=0,phi=0)
+        self._fun_kwargs["theta0_outboard"] = bool(R1 > R0)
```
YigitElma, line 896: Is this check enough for stellarator asymmetric cases?

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+                    _pool_fun,
+                    path=tmp_path,
+                    exec=exec,
+                    al0=al0,
+                    sleep_time=sleep_time,
+                    stop_time=stop_time,
+                ),
+                idxs,
+            )
+            pool.close()
```
YigitElma: These 2 lines might be redundant with `with` context manager.

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+    theta0_outboard=True,
+):
+    """TERPSICHORE driver function.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    processes : int, optional
+        Maximum number of CPU threads to use for multiprocessing. Default = 1.
```
YigitElma, line 58: Mention that if there is 1 equilibrium, this is not used. As it is, it sounds like you can allocate more threads (which is not the same as number of processes)

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    abs_step : float, optional
+        Absolute finite difference step size. Default = 1e-4.
+        Total step size is ``abs_step + rel_step * mean(abs(x))``.
+    rel_step : float, optional
+        Relative finite difference step size. Default = 0.
+        Total step size is ``abs_step + rel_step * mean(abs(x))``.
+    processes : int, optional
+        Maximum number of CPU threads to use for multiprocessing. Default = 1.
```
YigitElma, line 716: See above comment about len(eq)==1 case.

YigitElma: Also, since the objective expects 1 equilibrium, this argument seems unnecessary

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,967 @@
...
+                    stop_time=stop_time,
+                ),
+                idxs,
+            )
+            pool.close()
+            pool.join()
+            result = jnp.vstack(results, dtype=float)
+
+    # remove temporary directory
+    shutil.rmtree(tmp_path)
```
YigitElma: As Rory suggested, I can see that having an option to keep these files would be useful. Something like `store_output=["last", "all", None]`, if None delete everything. For `all` and `last` first move the files to a new directory `RESULTS`, if `last` first delete the contents of that directory. To distinguish different optimization iterations, you can use time stamp.

YigitElma:
```python
def handle_tmp_output(tmp_path, results_dir="RESULTS", store_output=None):
    """
    Handle intermediate files stored in tmp_path based on store_output mode.

    Parameters
    ----------
    tmp_path : str or Path
        Path to the temporary directory containing files and subfolders.
    results_dir : str or Path
        Base directory where results should be stored.
    store_output : {"last", "all", None}
        - None: delete tmp_path entirely
        - "last": delete previous RESULTS, keep only this run
        - "all": append results in timestamped subfolders
    """
    tmp_path = Path(tmp_path)
    results_dir = Path(results_dir)

    if store_output is None:
        # Just delete the temporary directory
        shutil.rmtree(tmp_path, ignore_errors=True)
        return

    # Ensure RESULTS exists
    results_dir.mkdir(parents=True, exist_ok=True)

    # Timestamp for this run
    timestamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
    run_dir = results_dir / timestamp

    if store_output == "last":
        # Clear RESULTS before saving this run
        shutil.rmtree(results_dir, ignore_errors=True)
        results_dir.mkdir(parents=True, exist_ok=True)

    # Move contents of tmp_path into RESULTS/timestamp/
    run_dir.mkdir(parents=True, exist_ok=True)
    for item in tmp_path.iterdir():
        target = run_dir / item.name
        shutil.move(str(item), str(target))

    # Remove now-empty tmp_path
    shutil.rmtree(tmp_path, ignore_errors=True)
```

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,320 @@
...
+
+        eq = get("W7-X")
+        optimizer = Optimizer("proximal-lsq-exact")
+        objective = ObjectiveFunction(
+            (
+                TERPSICHORE(
+                    eq=eq,
+                    abs_step=1e-2,
+                    rel_step=0,
+                    processes=os.cpu_count(),
```
YigitElma, line 301: Again, I think this is doing nothing.

desc/external/_terpsichore.py
YigitElma: As Daniel mentioned, `ExternalObjective` does some stuff under the hood to pass multiple eq to run function in parallel

desc/external/_terpsichore.py
YigitElma: May I ask you to test if this line errors or not during optimization by specifying a very small stop time? There was a really long discussion here https://github.com/jax-ml/jax/issues/26102#issuecomment-2619362685 for whether or not there should be errors in a pure callback.

YigitElma: https://docs.jax.dev/en/latest/external-callbacks.html#pure-callback-and-exceptions

desc/external/_terpsichore.py
ddudt: I did this, except I made it all or nothing (no "last" option).

desc/external/_terpsichore.py
ddudt: It should be. It is comparing R at (rho=0) and at (rho=1, theta=0, phi=0). That doesn't make any assumptions about stellarator symmetry. 

We only need to know if theta=0 is on the inboard or outboard side, since that choice is arbitrary in DESC but matters for TERPSICHORE.

YigitElma: Hmm, I don't know the expectations of TERPSICHORE, but what I meant was for the stellarator asym case, do $Z(\rho=1, \theta=0, \zeta=0)$ and $Z(\rho=1, \theta=\pi, \zeta=0)$ have to be 0? This line only checks if $\theta=0$ point is on the inner side or outer side but not exactly at Z=0. If that is not a problem then it is fine.

desc/external/_terpsichore.py
ddudt: Interesting. In my testing I can get these errors to raise. I have try-except logic to try and run TERPSICHORE, catch if it wasn't successful because of different errors, and use a default value in those cases. 

Maybe my code works because I am catching the errors? If an error is raised, it gets caught and prints a warning instead. The function I am calling `jax.pure_callback` on always returns a real value.

desc/external/_terpsichore.py
ddudt: Oh I see your point. I'm not sure exactly what TERPSICHORE expects, I don't even know if it works for asymmetric cases but I think it should. At least if it doesn't work, the user should now be able to tell that TERPSICHORE is failing to run and adjust their equilibrium accordingly.

docs/dev_guide/external_objectives.rst
YigitElma: As discussed in the dev meeting, our use of `pure_callback` is not the ideal one, so (although it can be) not advertising it as an alternative to wrap AD-incompatible code is better. Some pitfalls can be for example, if a user has `jax` computation inside the `pure_callback`, it might halt due to async dispatch (this was the usual-suspect for the block_until_ready call for the jacobian in `lsqtr`).

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,316 @@
...
+        # not shown: a delay to wait until TERPSICHORE finishes running
+        terps_subprocess.terminate()
+
+        fout.close()
+        ferr.close()
+
+Finally, the ``TERPSICHORE`` objective function simply inherits from
+``ExternalObjective`` and passes ``fun=terpsichore`` as the external function.
+``dim_f=1`` because TERPSICHORE returns a scalar growth rate, and ``vectorized=True``
+because the function ``terpsichore`` expects a list of Equilibria as its only positional
```
YigitElma: Can you explain to me how the vectorization actually comes into play again? When I look at the `ExternalObjective` class, the compute function is
```python
        f = self._fun_wrapped(params)
        return f
```
where `self._fun_wrapped` is the vectorized function (and the user needs to apply the multiprocessing manually), but how exactly do `params` become vectorized? Is it the `batched_vectorize` call (or any vmap flavor) that makes the params a set of params? I think the `vmap_method` argument of `jax.pure_callback` does the trick.

I think I understand it, but I will probably forget it in the future and come here to check it out :)

YigitElma (approved):

CHANGELOG.md
```diff
@@ -2,20 +2,20 @@ Changelog
 =========
 
 New Features
-------------
+
 - New basis vector and metric elements derivatives in PEST coordinates and quantities useful for a global MHD stability solver.
+- Adds ``desc.external.TERPSICHORE`` objective for computing linear ideal MHD stability growth rates. This objective subclasses from ``ExternalObjective`` and requires access to the TERPSICHORE code, which is not included with DESC or its dependencies.
```
dpanici, line 7: mention the new tutorial going through this as well showing an example use of external objetive with parallel processing

desc/external/README.rst
```diff
@@ -0,0 +1,12 @@
+**Warning:**
+
+These objectives depend on other codes external to DESC. They are not routinely tested
+and may not be compatible with all versions of the codes. The DESC team is not
+responsible for maintaining or documenting those external codes, and we do not guarantee
+to regularly maintain these objectives. Use these at your own risk!
+
+The external codes are not included with DESC, and you must obtain access to them to use
```
dpanici:
```suggestion
The external codes are not included with DESC, and you may need to obtain access to them to use or figure out how to install them yourself.
```

docs/dev_guide/external_objectives.rst
dpanici: so we will not advertise that it wraps AD-incompatible python things, just external codes?

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,316 @@
...
+included with DESC.)
+
+Simple example
+--------------
+
+The following is a simple example of a custom function that is not JAX transformable.
+It was adapted from the test ``test_external_vs_generic_objectives`` in
+``tests/test_examples.py``.
+
+The function must take a single positional argument, which can be either a single
```
dpanici: specify that this is `fun`, and that this requirement is a requirement of `ExternalObjective`, not just a requirement of the specific function you chose for this example

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,316 @@
...
+bottleneck of an optimization is computing the Jacobian matrix with finite differences,
+which scales with the number of optimization degrees of freedom. Evaluating the
+TERPSICHORE growth rates for each Equilibrium can be performed in parallel on different
+CPU threads using Python multiprocessing. Note that writing the equilibria data in step
+2 cannot be easily parallelized, since it involves computations using JAX that has
+issues with multiprocessing.
+
+::
+
+    def terpsichore(eq, *, processes=1, path, exec):
```
dpanici, line 146: We knwo what ,* , means but it is possible others don,t? maybe say this means all other args after must be passed as keyword args

docs/dev_guide/external_objectives.rst
```diff
@@ -0,0 +1,316 @@
...
+
+        # not shown: a delay to wait until TERPSICHORE finishes running
+        terps_subprocess.terminate()
+
+        fout.close()
+        ferr.close()
+
+Finally, the ``TERPSICHORE`` objective function simply inherits from
+``ExternalObjective`` and passes ``fun=terpsichore`` as the external function.
+``dim_f=1`` because TERPSICHORE returns a scalar growth rate, and ``vectorized=True``
```
dpanici:
```suggestion
``dim_f=1`` because the TERPSICHORE objective returns a single, scalar growth rate, and ``vectorized=True``
```

TERPSISCHORE gives other info right like eigfxns etc, so specify we are just using the single growth rate of the fastest growing mode as the objective value

desc/external/_terpsichore.py
```diff
@@ -0,0 +1,941 @@
...
+    if mode_family < 0:
+        n_modes = np.arange(N_min, N_max + 1)
+    else:
+        n_modes = np.arange(N_min, N_max + 1) * nfp
+        n_modes = np.concatenate((n_modes - mode_family, n_modes + mode_family))
+        n_modes = np.sort(np.unique(n_modes))
+
+    for n in range(N_min, N_max + 1):
+        mode_str = "     "
+        for m in range(56):
```
dpanici, line 597: what is this magic number?

docs/dev_guide/external_objectives.rst
ddudt: `ExternalObjective` does not do any vectorization/parallelization -- we decided that each external code wrapper should handle that on its own. This `vectorized` argument is only to tell if it should pass a single equilibrium or a list of equilibria to the wrapper function. 

So in this case, the `TERPSICHORE` objective expects a list of equilibria, and it uses multiprocessing to parallelize across that list.

docs/dev_guide/external_objectives.rst
ddudt: We don't need to document basic python syntax. The fact that all other args must be keyword args is already documented earlier in this tutorial and in `ExternalObjective`.

docs/dev_guide/external_objectives.rst
ddudt: correct

desc/external/_terpsichore.py
ddudt: 56 is the maximum poloidal mode number that TERPSICHORE can resolve. It's hard-coded because the input file needs that many columns in the mode number table.

YigitElma (approved):

PR #1047, opened 2024-06-10
Fixes for `jax>=0.4.29` and `interpax==0.3.2`

dpanici:
There are some tests failing due to changes in newest jax version (0.4.29) and apparently also due to the latest version of `interpax` (0.3.2 )

- Remove rcond kwarg from pinv to avoid warning in jax>= 0.4.29, the default value for rtol/rcond is None anyways so no need to explicitlt call it
- don't test `monotonic-0` in the `test_length` for `SplineXYZCurve` , as the test is for a circular curve and the zero-derivative endpoints don't make sense for this case, where the curvature is constant and nonzero for the curve.

### DIFF
--- a/desc/singularities.py
+++ b/desc/singularities.py
@@ -299,9 +299,7 @@ class DFTInterpolator(_BIESTInterpolator):
             M=source_grid.M, N=source_grid.N, NFP=source_grid.NFP
         )
         A = basis.evaluate(source_grid.nodes)
-        # TODO: eventually rcond will be deprecated in favor of rtol
-        # in jax, so must change it when that happens
-        Ainv = jnp.linalg.pinv(A, rcond=None)
+        Ainv = jnp.linalg.pinv(A)
 
         B = jnp.zeros((*theta_q.shape, basis.num_modes))
 
--- a/desc/singularities.py
+++ b/desc/singularities.py
@@ -299,6 +299,8 @@ class DFTInterpolator(_BIESTInterpolator):
             M=source_grid.M, N=source_grid.N, NFP=source_grid.NFP
         )
         A = basis.evaluate(source_grid.nodes)
+        # TODO: eventually rcond will be deprecated in favor of rtol
+        # in jax, so must change it when that happens
         Ainv = jnp.linalg.pinv(A, rcond=None)
 
         B = jnp.zeros((*theta_q.shape, basis.num_modes))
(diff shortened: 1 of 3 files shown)


dpanici:
keeping as draft for now to see if any other tests fail, I think a couple others may be failing due to the new. jax release as well

dpanici:
I think some spline tests fail with interpax 0.3.2 as well, based off them passing [here](https://github.com/PlasmaControl/DESC/actions/runs/9256931832/job/25464014832) and not [here](https://github.com/PlasmaControl/DESC/actions/runs/9452972383/job/26037310687?pr=1002). The tests for this PR are not finished yet but I bet they will fail here too, 0.3.2 interpax is only installed when jax 0.4.29 is also installed it seems so we have not ran into this issue before

dpanici:
This is the splinexyzCoil used in the test_compute_everything, it is weird... it is not periodic even though the interpolation should make it periodic
```
c = SplineXYZCurve(
             X=[5, 10, 2, 5], Y=[1, 2, 3, 1], Z=[-4, -5, -6, -4]
        )
data = c.compute(["X","Y","s","x_s","Z"],grid=LinearGrid(N=40))

plt.figure()
plt.plot(data["X"],data["Y"])

plt.figure()
plt.plot(data["s"],data["X"])
plt.scatter(c.knots,c.X)
# plt.xlim([6,6.3])
plt.figure()
plt.plot(data["s"],data["Y"])
plt.scatter(c.knots,c.Y)
plt.figure()
plt.plot(data["s"],data["Z"])
plt.scatter(c.knots,c.Z)

# plt.xlim([6,6.3])
plt.figure()
plt.plot(data["s"],data["x_s"])
# plt.xlim([6,6.3])```

X is periodic but Y and Z are not. 

<img width="476" alt="image" src="https://github.com/PlasmaControl/DESC/assets/37969854/cdfa0b72-556c-404a-a3d1-7b868c9651ba">


<img width="476" alt="image" src="https://github.com/PlasmaControl/DESC/assets/37969854/cc0ca504-2020-4e25-81f4-a4e948d85ee9">


<img width="476" alt="image" src="https://github.com/PlasmaControl/DESC/assets/37969854/d15899e9-5833-4d19-915a-0dcac053f25e">

dpanici:
Change in DESC to always pass in the closed curve to the interpolate calls

f0uriest:
> Change in DESC to always pass in the closed curve to the interpolate calls

should also require interpax >=0.3.2 to avoid mismatched versions

f0uriest:
Disregard previously comment, I yanked interpax v0.3.2 for now and rolled the changes kaya needed into v0.3.3

f0uriest (approved):

PR #1048, opened 2024-06-10
Option to use float32 

### LINKED ISSUE #1033: Using bfloat instead of regular float
I have heard about people using brain float instead of regular floating point precision for low-level code to save memory.
It doesn't usually work for non-linear, time-dependent solvers like GX because the error accumulates over time. But since DESC is solving a steady-state problem, the error accumulation problem may not be so bad.

It would be interesting to see if bfloat can help us save memory in DESC. (https://www.cerebras.net/machine-learning/to-bfloat-or-not-to-bfloat-that-is-the-question/)

rahulgaur104:
Resolves #1033 . We see that float32 is clearly desirable as it would allow us to fit more objectives on the same GPU.

But it causes half the tests to fail because the accuracy has been reduced. Let's find a way where we can use a lower precision without significantly altering the final equilibrium.

Perhaps, we could change the precision in the more sensitive parts of the code while still reducing overall memory consumption.

Suggestions?


f0uriest:
When this is working we should add an assert inside the `in32bit` function to ensure that the output is correctly 32 bit before upcasting.

f0uriest:
This might be just what we need: https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.disable_x64.html#jax.experimental.disable_x64

dpanici:
Mixed Precision Iterative refinement could be useful?

YigitElma:
Maybe a cleaner implementation could be to use `jax.enable_64` only for the linear system solves (recent changes made it more stable [see](https://github.com/jax-ml/jax/issues/27602#issuecomment-3383009141)). I bet we can compute the jacobian in float32 with no problem (for most of the objectives), but since almost all jacobians are ill-conditioned, the linear system has to be solved in full precision.

We can try
- use float32 globally
- wrap `svd`, `qr`, `solve`, `cho` etc with `enable_x64`
- also create a decorator that converts a given operation to full precision (cast input arrays to 64 then do computation)

If `jax.enable_x64` acts as identity for globally double precision case, with this decorator approach we can go back and forth easily. Something like `desc_config["precision"]="float32"` should be enough.

YigitElma:
Stuff like `elementary_charge**2` needs double precision, otherwise jax returns NaNs

YigitElma:
- try casting `x` and `tangents` (v) in float32, then internally before computing anything cast them back to float64

PR #1051, opened 2024-06-13
cast Ainv_full and Z to jnp to avoid host to device transfers when re…

dpanici:
…cover/project is called


simonbyrne:
Can confirm that this appears to remove the large HtoD transfers between iterations.

YigitElma (approved):
In future, we can also think of jiting project/recover functions

YigitElma (approved):

f0uriest (approved):

PR #1052, opened 2024-06-13
Add `jac_chunk_size` keyword argument to `ObjectiveFunction` to reduce memory usage of forward mode Jacobian calculation

### LINKED ISSUE #826: Use batched vmap to reduce memory usage
Something like this: https://github.com/google/jax/issues/11319

the built in `jax.jacfwd` and `jax.jacrev` are basically just wrappers that do `lambda x: vmap(lambda e: jax.jvp(f, x, e))(jnp.eye(x.size))` so we could use a chunked vmap to be more memory efficient.

Main thing would be coming up with some heuristic to determine how big the chunks can be etc.

dpanici:
- changes most `jnp.vectorize` calls to instead use `batched_vectorize` which performs the function vectorization in smaller chunks, which reduces the memory cost of the calculation, at the expense of taking longer the smaller the chunk size is.
- Add `jac_chunk_size` to `ObjectiveFunction` and `_Objective` to control the above chunk size for the `fwd` mode Jacobian calculation
  - if `None`, the chunk size is equal to `dim_x`, so no chunking is done
  - if an `int`, this is the chunk size to be used.
  - if `"auto"` for the `ObjectiveFunction`, will use a heuristic for the maximum `jac_chunk_size` needed to fit the jacobian calculation on the available device memory, according to the formula: `max_jac_chunk_size = (desc_config.get("avail_mem") / estimated_memory_usage - 0.22)  / 0.85  * self.dim_x`
- the `ObjectiveFunction` `jac_chunk_size` is used if `deriv_mode="batched"`, and the `_Objective` `jac_chunk_size` will be used if `deriv_mode="blocked"`

This works well, this is LMN18 equilibrium solve with 1.5 oversampled grid and `maxiter=10` memory trace vs time on GPU, where we get 4x memory decrease with negligible runtime increase:

image

Also, I can do up to an `LMN=20` eq `ForceBalance` objective with the default double grid oversampling, and with the `"auto"` chunk sizing, the jacobian compiles and computes without going OOM on an 80gb GPU (on master this would go OOM).

TODO
- [x] re-implement without relying on `netket`
- [x] change chunk_size to a better default value (something like 100 would be fine, maybe can dynamically choose based off of size of `dim_x`)
- [x] Add `chunk_size` argument to every Objective class
   - [x] I am choosing right now to not to add it as an arg to the `LinearObjective` classes, though technically you could 
- [x] Add `"chunked"` as a deriv_mode to `Derivative` (or, just as an argument to `Derivative` to be used when `"batched"` is used) - > I don't remember what this was exactly, I think we can keep just for Objectives
- [x] change `chunk_size` to `jacobian_chunk_size` for Objective kwarg
- [x] use in constraint wrappers

TODO Later
- [ ] add to singular integral calculation as well

Resolves #826


dpanici:
@dpanici jax batched vmap has been merged to master

dpanici:
@dpanici make separate branch with the implementation using JAX's version, and in this PR implement the one based off of `netket`

dpanici:
@kianorr @YigitElma

unalmis:
Can you also update adding objectives docs for chunk size option

dpanici:
> If you don't care about jax's native multi-GPU sharding support it should be easy to just vendor our implementation. In that case, you can just vendor our `netket/jax/_chunk_utils.py`, `netket/jax/_scanmap.py` and `netket/jax/_vmap_chunked.py` .
> 
> The former 2 files are on purpose standalone. Only `_vmap_chunked` depends on other things but only if you are using sharding.
> 
> Remove all branches hitting of `axis_0_is_sharded == True` and `config.netket_experimental_sharding`, which will allow you to remove `sharding_decorator`, which is a mess only needed to support efficient sharding of jax arrays.
> 
> Also replace `HashablePartial` with `functools.partial`

Thank you very much! This was extremely helpful.

How would you like us to credit your package in our code where we have used these functions? I currently mention the original filename and the `netket` package, but I am happy to credit it in whatever way you prefer

PhilipVinc:
Glad I could help.

Standard practice is to include a copyright notice along the lines of 

```python
# The following section of this code is a derived from the NetKet project 
# https://github.com/netket/netket/blob/9881c9fb217a2ac4dc9274a054bf6e6a2993c519/netket/jax/_vmap_chunked.py
# 
# The original copyright notice is as follows
# Copyright 2021 The NetKet Authors - All rights reserved.
# Licensed under the Apache License, Version 2.0 (the "License");
# 
```

which would be more than fine. 

I'm not terribly picky though. Do not worry.

desc/objectives/objective_funs.py
```diff
@@ -33,21 +34,38 @@ class ObjectiveFunction(IOAble):
...
+        to compute the Jacobian is roughly ``t ~1/chunk_size` with some baseline time,
+        so the larger the ``chunk_size``, the faster the calculation takes, at the cost
+        of requiring more memory.
+        If None, it will default to the largest possible size i.e. ``dim_x``
 
     """
 
     _io_attrs_ = ["_objectives"]
 
     def __init__(
```
dpanici, line 62: add warning to check for nonsensical input combos (chunk size on outer when blocked, or on inner obj when bacthec)

docs/adding_objectives.rst
```diff
@@ -234,3 +248,30 @@ available for the ``loss_function`` kwarg are ``[None,"mean","min","max"]``, wit
 ``None`` meaning using the usual default objective cost, while ``"mean"`` takes the
 average of the raw objective values (before subtracting the target/bounds or
 normalization), ``"min"`` takes the minimum, and ``"max"`` takes the maximum.
+
+Reducing Memory Size of Objective Jacobian Calculation
```
f0uriest, line 252: I'd recommend removing this here.

I think its useful but should
- probably be put somewhere more focused, like a performance tips page
- also be tested more to find out what combos of batched/blocked/chunksize work best

desc/utils.py
f0uriest, line 1: I think most of this stuff should go in backend, or maybe its own file

desc/utils.py
```diff
@@ -684,4 +697,515 @@ def broadcast_tree(tree_in, tree_out, dtype=int):
...
+
+# TODO put it somewhere
+def _multimap(f, *args):
+    try:
+        return tuple(map(lambda a: f(*a), zip(*args)))
+    except TypeError:
+        return f(*args)
+
+
+def scan_append_reduce(f, x, append_cond, op=_tree_add, zero_fun=_tree_zeros_like):
```
f0uriest, line 810: I think for the use case we care about (`append_cond=True`) this is equivalent to regular `jax.lax.scan`?

desc/utils.py
```diff
@@ -684,4 +697,515 @@ def broadcast_tree(tree_in, tree_out, dtype=int):
...
+    if not set(in_axes).issubset((0, None)):
+        raise NotImplementedError("Only in_axes 0/None are currently supported")
+
+    argnums = tuple(
+        map(lambda ix: ix[0], filter(lambda ix: ix[1] is not None, enumerate(in_axes)))
+    )
+    return in_axes, argnums
+
+
+def apply_chunked(
```
f0uriest, line 1009: this looks unused

f0uriest (changes requested):
Mostly fine, a few things to be cleaned up.

- Use `batched_vectorize` in constraint wrappers as well
- Don't need to worry about batched vs looped for singular integral stuff for now (I think batched + chunked jacobian is probably enough and we can remove the looped option from that but ill test it more)
- Make sure `chunk_size` is added to all objectives, there are some new ones I didn't see.

desc/objectives/objective_funs.py
```diff
@@ -33,21 +34,38 @@ class ObjectiveFunction(IOAble):
         Method for computing Jacobian matrices. "batched" uses forward mode, applied to
         the entire objective at once, and is generally the fastest for vector valued
         objectives, though most memory intensive. "blocked" builds the Jacobian for each
-        objective separately, using each objective's preferred AD mode. Generally the
-        most efficient option when mixing scalar and vector valued objectives.
-        "looped" uses forward mode jacobian vector products in a loop to build the
-        Jacobian column by column. Generally the slowest, but most memory efficient.
+        objective separately, using each objective's preferred AD mode (and
+        each objective's `chunk_size`). Generally the most efficient option when mixing
+        scalar and vector valued objectives. "looped" uses forward mode jacobian vector
```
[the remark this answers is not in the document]
ddudt: Question about looped: I've tried using looped mode in an optimization where I increase the number of optimization variables at each step, and at a later step it crashed with an out of memory error. That is expected for batched mode, but is it expected for looped? I thought looped would use about that same amount of memory regardless of the number of optimization variables

ddudt:
Slightly tangential comment:
`deriv_mode="batched"` appears to compute the Jacobian wrt only the optimization variables. `deriv_mode="blocked"` however, appears to compute each objective's Jacobian wrt all parameters related to that objective (even if they are fixed by constraints). This is problematic for the `ExternalObjective`, since it requires extra finite differences, which are generally expensive. Is there a way to remedy that?

desc/objectives/objective_funs.py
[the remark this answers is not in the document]
f0uriest: When was this? I know a while ago looped was actually broken, though I think it's fixed now. It could also be to some recompilation memory leaks that should have been fixed in #1043

f0uriest:
> Slightly tangential comment: `deriv_mode="batched"` appears to compute the Jacobian wrt only the optimization variables. `deriv_mode="blocked"` however, appears to compute each objective's Jacobian wrt all parameters related to that objective (even if they are fixed by constraints). This is problematic for the `ExternalObjective`, since it requires extra finite differences, which are generally expensive. Is there a way to remedy that?

In theory yes. I think what we'd need is a "blocked jvp" where it computes the total jvp as a sum of smaller jvps (then `LinearConstraintProjection` will only take jvps in the directions needed). The main reason I didn't do this is that the "blocked" logic will use forward or reverse mode as appropriate, while "jvp" usually implies forward mode only, so if we use reverse mode when doing a jvp that might cause some issues in places where we assume jvp/vjp mean forward/reverse only. This isn't a dealbreaker, just something we need to think about a bit more.

desc/utils.py
YigitElma: I vote for a new file, or put this in backend after we remove `numpy` backend.

docs/adding_objectives.rst
YigitElma: Yes, this deserves to have its own page.

tests/test_objective_funs.py
```diff
@@ -1292,8 +1292,8 @@ def test_derivative_modes():
     )
     obj2 = ObjectiveFunction(
         [
-            PlasmaVesselDistance(eq, surf),
```
YigitElma, line 1295: Can we add a test that check chunked and normal batched gives the same result?

YigitElma (commented):
Mostly fine, I agree with Rory's comments.

desc/objectives/getters.py
```diff
@@ -64,7 +64,10 @@ def get_equilibrium_objective(eq, mode="force", normalize=True):
         An objective function with default force balance objectives.
 
     """
-    kwargs = {"eq": eq, "normalize": normalize, "normalize_target": normalize}
+    kwargs = {
+        **{"eq": eq, "normalize": normalize, "normalize_target": normalize},
+        **kwargs,
```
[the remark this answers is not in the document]
dpanici: my bad, I will make it a separate kwarg

tests/test_objective_funs.py
dpanici: batched always will use chunked now, since `None` defaults to `dim_x/4`

desc/utils.py
dpanici: I put a new file but maybe a `utils` folder is warranted now

desc/objectives/objective_funs.py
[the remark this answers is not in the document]
dpanici: removed `looped` (with deprecation warning)

desc/objectives/objective_funs.py
```diff
@@ -130,10 +158,44 @@ def build(self, use_jit=None, verbose=1):
...
+            "different 'jac_chunk_size' for its Jacobian computation.",
+        )
+        warnif(
+            self._jac_chunk_size is not None and self._deriv_mode == "blocked",
+            UserWarning,
+            "'jac_chunk_size' was passed into ObjectiveFunction, but the"
+            "ObjectiveFunction is using 'blocked' deriv_mode, so sub-objective "
+            "'jac_chunk_size' are used to compute each sub-objective's Jacobian, "
+            "`ignoring the ObjectiveFunction's 'jac_chunk_size'.",
+        )
```
YigitElma, line 179: I have a question about this. If say the user wants to use `blocked` in a multiobjective optimization with one of the `jac_chunk_size` specified and the rest of the `jac_chunk_size` different than `dim_x/4`, wouldn't it be easier to set the `jac_chunk_size` of sub-objectives to the `jac_chunk_size` of the ObjectiveFunction?
So, like
```python
obj = ObjectiveFunction(
    [
    Objective_1(eq, jac_chunk_size=30),
    Objective_2(eq),
    Objective_3(eq),
    ],
    deriv_mode="blocked",
    jac_chunk_size=10,
)
```
The natural way I understand this, the user wants specifically `jac_chunk-size=30` but 10 for the rest, not the default value.

tests/test_objective_funs.py
```diff
@@ -1284,31 +1284,38 @@ def test_derivative_modes():
...
+            deriv_mode="looped",
+            use_jit=False,
+        )
+    with pytest.warns(UserWarning, match="jac_chunk_size"):
+        obj1.build()
+    with pytest.warns(UserWarning, match="jac_chunk_size"):
+        obj2.build()
+    # check that default size works for blocked
+    assert obj2.objectives[1]._jac_chunk_size == np.ceil(
+        sum(t.dim_x for t in obj2.objectives[1].things) / 4
```
YigitElma, line 1317: Check my above comment. A user may want to create `obj2` with `jac_chunk_size=10` for the not specified sub_objectives. It might be tedious for them to write `jac_chunk_size=10` for each one of them in case they have a lot of objectives. I guess for now we don't have high number of objectives for a single optimization but it might be good for future.

YigitElma (commented):
I am ok if the majority wants to use it like the current version, but I would prefer the other way

desc/objectives/objective_funs.py
dpanici: IF blocked is used, I want the default to be whatever is put for the sub objective. It is a choice, i'd rather be verbose than start having some not entirely clear logic for what sizes are used where

tests/test_objective_funs.py
dpanici: I am fine with the current way, I think it would be more confusing if we did not make it completely manual to specify

desc/objectives/_bootstrap.py
```diff
@@ -66,6 +66,19 @@ class BootstrapRedlConsistency(_Objective):
         or quasi-axisymmetry; set to +/-NFP for quasi-helical symmetry.
     name : str, optional
         Name of the objective function.
+    jac_chunk_size : int, optional
+        Will calculate the Jacobian for this objective ``jac_chunk_size``
+        columns at a time, instead of all at once.  The memory usage of the
+        Jacobian calculation is roughly ``memory usage = m0 + m1*jac_chunk_size``:
+        the higher the chunk size, the less memory the Jacobian calculation
```
ddudt, line 73: I think you mean **_more_** memory, not less. 

Also I think it's better to say "larger" chunk size rather than "higher"

desc/objectives/_bootstrap.py
```diff
@@ -66,6 +66,19 @@ class BootstrapRedlConsistency(_Objective):
...
+        will require (with some baseline memory usage). The time to compute the
+        Jacobian is roughly ``t=t0 +t1/jac_chunk_size``, so the larger the
+        ``jac_chunk_size``, the faster the calculation takes, at the cost of
+        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
+        memory intensive,  but slowest method of calculating the Jacobian.
```
ddudt, line 78: extra space on this line

desc/objectives/_bootstrap.py
```diff
@@ -66,6 +66,19 @@ class BootstrapRedlConsistency(_Objective):
...
+        If None, it will default to the largest possible
+        `jac_chunk_size` i.e. ``np.ceil(dim_x/4)``
```
ddudt, line 80: This isn't really the "largest possible" chuck size, wouldn't that be `jac_chunk_size = dim_x`? I don't mind this as the default, I just think this documentation needs to be clarified.

desc/objectives/getters.py
```diff
@@ -57,14 +59,19 @@ def get_equilibrium_objective(eq, mode="force", normalize=True):
...
 
     Returns
     -------
     objective, ObjectiveFunction
         An objective function with default force balance objectives.
 
     """
-    kwargs = {"eq": eq, "normalize": normalize, "normalize_target": normalize}
+    kwargs = {
+        **{"eq": eq, "normalize": normalize, "normalize_target": normalize},
```
ddudt, line 72: This feels like awkward notation. Isn't this equivalent to appending `eq`, `normalize`, and `normalize_target` to the existing `kwargs`?

desc/objectives/objective_funs.py
```diff
@@ -29,34 +31,60 @@ class ObjectiveFunction(IOAble):
...
         "auto" defaults to "batched" if all sub-objectives are set to "fwd",
         otherwise "blocked".
     name : str
         Name of the objective function.
+    jac_chunk_size : int, optional
+        If `"batched"` deriv_mode is used, will calculate the Jacobian
+        ``jac_chunk_size`` columns at a time, instead of all at once.
+        The memory usage of the Jacobian calculation is roughly
+        ``memory usage = m0 + m1*jac_chunk_size``: the higher the chunk size,
+        the less memory the Jacobian calculation will require (with some baseline
```
ddudt, line 51: See previous comment about changing to "the **_larger_** the chunk size, the **_more_** memory ..."

Also need to make these changes in all the objective docstrings.

desc/objectives/objective_funs.py
```diff
@@ -130,10 +158,44 @@ def build(self, use_jit=None, verbose=1):
...
+        )
+
         if not self.use_jit:
             self._unjit()
 
         self._set_things()
+        if self._jac_chunk_size is None and self._deriv_mode == "batched":
+            # set jac_chunk_size to 1/4 of number columns of Jacobian
+            # as the default for batched deriv_mode
+            self._jac_chunk_size = int(np.ceil(self.dim_x / 4))
```
ddudt, line 188: To make sure I understand: The existing master code when using `deriv_mode="batched"` is equivalent to `jac_chunk_size=dim.x`, correct? So why change the default to be one-fourth of that? Did you find that size is generally the best trade-off between speed and memory? Just want to check since I think this will change the behavior of old code.

desc/utils.py
ddudt: I agree a new utils directory would be good. Make an issue for it so we don't forget, if we aren't going to do it in this PR.

docs/adding_objectives.rst
```diff
@@ -70,6 +70,18 @@ A full example objective with comments describing the key points is given below:
             Collocation grid containing the nodes to evaluate at.
         name : str, optional
             Name of the objective function.
+        jac_chunk_size : int, optional
+            If `"blocked"` deriv_mode is used in the ObjectiveFunction, will
+            calculate the Jacobian for this objective ``jac_chunk_size`` columns at a time,
+            instead of all at once.  The memory usage of the Jacobian calculation is
```
ddudt, line 76: extra space between sentences

docs/adding_objectives.rst
```diff
@@ -70,6 +70,18 @@ A full example objective with comments describing the key points is given below:
...
+            linearly proportional to ``jac_chunk_size``: the smaller the ``jac_chunk_size``, the
+            less memory the Jacobian calculation will require (with some baseline memory
+            usage). The time to compute the Jacobian is roughly ``t ~1/jac_chunk_size``
+            with some baseline time, so the larger the ``jac_chunk_size``, the faster the
+            calculation takes, at the cost of requiring more memory. A ``jac_chunk_size``
+            of 1 corresponds to the least memory intensive,  but slowest method of
```
ddudt, line 82: extra space between phrases

docs/adding_objectives.rst
```diff
@@ -70,6 +70,18 @@ A full example objective with comments describing the key points is given below:
...
+            calculating the Jacobian.
+            If None, it will default to the largest possible `jac_chunk_size` i.e. ``dim_x``
```
ddudt, line 84: I think the actual default is `dim_x / 4`

tests/test_optimizer.py
```diff
@@ -1136,15 +1136,16 @@ def test_proximal_jacobian():
...
-    obj2 = ObjectiveFunction(
-        (
-            QuasisymmetryTripleProduct(eq2, deriv_mode="fwd"),
-            AspectRatio(eq2, deriv_mode="fwd"),
-            Volume(eq2, deriv_mode="fwd"),
-        ),
-        deriv_mode="looped",
-        use_jit=False,
-    )
+    with pytest.warns(DeprecationWarning, match="looped"):
```
ddudt, line 1139: The deprecated `deriv_mode="looped"` is tested both here and in `test_objective_funs.py`, and the test coverage for this PR is pretty low. I recommend changing one of these two tests to another case to increase coverage.

tests/test_examples.py
ddudt, line 1: An interesting new test could be to run the same optimization with the two extremes of `jac_chunk_size=1` and `jac_chunk_size=dim_x`, and check that the latter runs faster than the former. Could be a simple tokamak equilibrium solve that runs quickly in both cases.

desc/utils_batched_vectorize.py
```diff
@@ -0,0 +1,462 @@
...
+
+####
+
+# The following section of this code is derived from the NetKet project
+# https://github.com/netket/netket/blob/9881c9fb217a2ac4dc9274a054bf6e6a2993c519/
+# netket/jax/_scanmap.py
+#
+# The original copyright notice is as follows
+# Copyright 2021 The NetKet Authors - All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
```
ddudt, line 114: This looks like a duplicate of the notice at the top of this file. Can we remove this one?

desc/utils_batched_vectorize.py
```diff
@@ -0,0 +1,462 @@
...
+
+_tree_add = functools.partial(jax.tree_util.tree_map, jax.lax.add)
+_tree_zeros_like = functools.partial(
+    jax.tree_util.tree_map, lambda x: jnp.zeros(x.shape, dtype=x.dtype)
+)
+
+
+# TODO put it somewhere
```
ddudt, line 122: What does this comment mean?

desc/utils_batched_vectorize.py
```diff
@@ -0,0 +1,462 @@
...
+        _chunk, jac_chunk_size=_jac_chunk_size(x_chunked)
+    )
+
+
+def chunk(x, jac_chunk_size=None):
+    """Split an array (or a pytree of arrays) into chunks along the first axis.
+
+    Parameters
+    ----------
+        x: an array (or pytree of arrays)
```
ddudt, line 92: It looks like the parameters in the docstrings for most of the functions in this file are over-indented.

desc/utils_batched_vectorize.py
```diff
@@ -0,0 +1,462 @@
...
+    return f_
+
+
+# The following section of this code is derived from the NetKet project
+# https://github.com/netket/netket/blob/9881c9fb217a2ac4dc9274a054bf6e6a2993c519/
+# netket/jax/_vmap_chunked.py
+#
+# The original copyright notice is as follows
+# Copyright 2021 The NetKet Authors - All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
```
ddudt, line 242: see previous comment about removing this duplicate notice

ddudt (changes requested):
The substance looks good, just need to clean up the documentation.

tests/test_examples.py
YigitElma: I don't know how effective the time profiling on GitHub hosts is since they may sometimes suffer from temporary slow-downs. Also, GitHub actions use CPU, so the number of cores is less than the GPU. Does this new method change the speed on the CPU?

desc/utils.py
dpanici: Possibly? I pared down as much as I could but there is still some logic that is opaque to me preventing me from just making it a bare scan, so I kept `scan_append`

desc/objectives/_bootstrap.py
dpanici: leftover from a bad find+replace

desc/objectives/objective_funs.py
dpanici: yea, Rory/Yigit wanted something auto to be better. I should check dim_x and current avail. memory and do it from that, but have not done a study to get a heurstic memory use vs chunk_size and dim_x yet

tests/test_examples.py
dpanici: yes it should affect times on cpu

dpanici: but idk if this is a good test

desc/utils_batched_vectorize.py
dpanici: removed

desc/objectives/getters.py
f0uriest: dictionaries don't have an "append" option but you could do `dict.update`

desc/objectives/objective_funs.py
f0uriest: I'd prefer to keep `chunk_size=None` be no chunking (`chunk_size=dim_x`). We could add a `chunk_size="auto"` option that for now uses dim_x/4 until we come up with a better heuristic (or maybe "auto" means chunk_size=1000 or something which is probably a safer bet than dim_x/4)

desc/objectives/getters.py
dpanici: I think I will revert this change as for the equilibrium objective, we'd never do blocked anyways, and that is the only place where the underlying jac_chunk_size matters

desc/objectives/objective_funs.py
```diff
@@ -33,21 +34,38 @@ class ObjectiveFunction(IOAble):
...
-        objective separately, using each objective's preferred AD mode. Generally the
-        most efficient option when mixing scalar and vector valued objectives.
-        "looped" uses forward mode jacobian vector products in a loop to build the
-        Jacobian column by column. Generally the slowest, but most memory efficient.
+        objective separately, using each objective's preferred AD mode (and
+        each objective's `chunk_size`). Generally the most efficient option when mixing
+        scalar and vector valued objectives. "looped" uses forward mode jacobian vector
+        products in a loop to build the Jacobian column by column. Generally the
+        slowest, but most memory efficient.
         "auto" defaults to "batched" if all sub-objectives are set to "fwd",
```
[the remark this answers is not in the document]
dpanici: I think I will keep as is: if you give chunk_size to sub obj, it will warn that it wont be used unless you do blocked, and make them request it specifically

tests/test_examples.py
ddudt: Yeah we don't need to add this test, but I would at least like to see some time profiling to know how this PR actually affects speed performance.

dpanici: Fair point, I will do some

dpanici:
Some laptop benchmarking on the ATF jacobian benchmark example

image
image

The effect on compute time is more thane expected, I think I will implement something that only automatically chooses a smaller size if the expected memory is larger than the available device memory. I will re-do these tests on the gpu if I can as well, this was just on cpu on mac

YigitElma:
> I will re-do these tests on the gpu if I can as well, this was just on cpu on mac

Can you also add the benchmarks for `jnp.vectorize`? To be sure that calling `batched_vectorize` takes the same time as the original JAX one.

YigitElma:
> I think I will implement something that only automatically chooses a smaller size if the expected memory is larger than the available device memory.

We can do this based on predefined values for `dim_x`

dpanici:
> > I will re-do these tests on the gpu if I can as well, this was just on cpu on mac
> 
> Can you also add the benchmarks for `jnp.vectorize`? To be sure that calling `batched_vectorize` takes the same time as the original JAX one.
yea they are basically the same as having a chunk size equal to dim_x

image

image

dpanici:
based off these scalings for memory usage of jacobian on GPU and CPU for the ATF benchmark, I will have "auto" estimate mem usage based off of dim_x * dim_f and apply a chunk_size to bring the estimated usage down to below the device size

image

image

dpanici:
a conservative estimate for the actual memory usage over the peak memory usage according to the above formula
image

dpanici:
Putting it together, we have something like

chunk_size < (device_mem / estimated_mem - b) / a * dim_x

where a ~ 0.8 and b ~ 0.15, and device_mem is the current host device available memory (as given by `desc_config.get("avail_mem")` for example)

when the RHS of this inequality is <1, that means we estimate that we will get an OOM error as even with a chunk size of 1, we cannot fit the jacobian's fwd mode calculation in the device memory

desc/objectives/_bootstrap.py
```diff
@@ -85,6 +99,7 @@ def __init__(
         grid=None,
         helicity=(1, 0),
         name="Bootstrap current self-consistency (Redl)",
+        jac_chunk_size="auto",
```
f0uriest, line 102: for the `_Objective`s I think `None` would be a better default (consistent with current behavior), and remove the "auto" option since it's just hard-coded to 1000.

desc/objectives/objective_funs.py
```diff
@@ -130,12 +170,53 @@ def build(self, use_jit=None, verbose=1):
             self._scalar = False
 
         self._set_derivatives()
+        sub_obj_jac_chunk_sizes_are_ints = [
+            isposint(obj._jac_chunk_size) for obj in self.objectives
+        ]
+        warnif(
```
f0uriest, line 176: I could see possibly making this an error instead of a warning? While technically the code can run with this inconsistency, it may indicate that the user meant to decrease memory usage but didn't use the right args, meaning it's probably better to fail early rather than waiting until OOM errors

desc/objectives/objective_funs.py
```diff
@@ -130,12 +170,53 @@ def build(self, use_jit=None, verbose=1):
...
         self._set_things()
-
         self._built = True
+
+        if self._jac_chunk_size == "auto":
+            # Heuristic estimates of fwd mode Jacobian memory usage,
+            # slightly conservative, based on using ForceBalance as the objective
+            estimated_memory_usage = 2.4e-7 * self.dim_f * self.dim_x + 1  # in GB
+            max_chunk_size = round(
+                (desc_config.get("avail_mem") / estimated_memory_usage - 0.22)
```
f0uriest, line 207: some notes about `desc_config.get("avail_mem")`:
- In all cases, it's the available memory when desc was first imported, which may be different than the amount available for objective functions/optimization.
- For CPU on clusters, it returns the total memory on the node, not just the amount that was requested as part of a slurm job, so usually its a massive over-estimate which might cause issues. I'm not sure what the best option is, I tried a while back to make it give the actual allocated memory but it seems like that value depends on a few different slurm options so it's not straightforward to parse

desc/utils.py
```diff
@@ -2,14 +2,13 @@
...
 import warnings
-from functools import partial
 from itertools import combinations_with_replacement, permutations
 
 import numpy as np
 from scipy.special import factorial
 from termcolor import colored
 
-from desc.backend import flatnonzero, fori_loop, jit, jnp, take
+from desc.backend import flatnonzero, fori_loop, functools, jit, jnp, take
```
f0uriest, line 11: functools is its own module, why import from backend?

desc/utils_batched_vectorize.py
f0uriest, line 1: could name something like `desc.batching`?

f0uriest: Also looks like a lot of lines not hit by coverage, are these unused? or just codecov being wacky again?

docs/memory_usage.rst
```diff
@@ -0,0 +1,26 @@
...
+at the cost of requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least memory intensive,
+but slowest method of calculating the Jacobian. If ``jac_chunk_size="auto"``, it will default to ``obj.dim_x/4``.
+
+If ``deriv_mode="blocked"`` is specified when the ``ObjectiveFunction`` is created, then the Jacobian will
+be calculated individually for each of the sub-objectives inside of the ``ObjectiveFunction``, and in that case
+the ``jac_chunk_size`` of the individual ``_Objective`` objects inside of the ``ObjectiveFunction`` will be used.
+For example, if ``obj1 = QuasisymmetryTripleProduct(eq, jac_chunk_size=100)``, ``obj2 = MeanCurvature(eq, jac_chunk_size=2000)``
+and ``obj = ObjectiveFunction((obj1, obj2), deriv_mode="blocked")``, then the Jacobian will be calculated with a
+``jac_chunk_size=100`` for the quasisymmetry part and a ``jac_chunk_size=2000`` for the curvature part, then the full Jacobian
+will be formed as a block diagonal matrix with the individual Jacobians of these two objectives.
```
f0uriest, line 26: not block diagonal, just regular blocked matrix

desc/utils.py
dpanici: no reason, I can change

desc/utils_batched_vectorize.py
dpanici: some of them were unused yep I removed, others though are just like error checks that I've not tested myself

desc/objectives/objective_funs.py
dpanici: yea I think that is a good idea

desc/objectives/objective_funs.py
dpanici: Hm yea I knew about the second option, I feel like rarely on CPU do we do intensive stuff on the cluster.

I can add a disclaimer in the ObjectiveFunction docstring about this issue on the cluster if on a CPU

f0uriest (approved):

f0uriest (dismissed):

desc/objectives/objective_funs.py
```diff
@@ -474,8 +553,6 @@ def jac_scaled_error(self, x, constants=None):
 
         if self._deriv_mode == "batched":
             J = Derivative(self.compute_scaled_error, mode="fwd")(x, constants)
-        if self._deriv_mode == "looped":
```
YigitElma, line 477: This is a completely unrelated comment but I think we should explain this jacobian, jvp, etc better in some dev guide. For example, an individual objectives `jac_scaled_error` doesn't use `batched_vmap` it is usually `jax.jacfwd`. On the other hand, when we wrap the objective and constraints, the jacobian is calculated by corresponding `jvp_` method, not `jac_` method. Even now it confuses me. What is the case we use `jac_` method instead of `jvp_`? Good thing to clarify in long-waiting dev-guide

YigitElma (approved):

YigitElma (approved):

desc/batching.py
```diff
@@ -0,0 +1,322 @@
...
+        checked_func = _check_output_dims(
+            excluded_func, dim_sizes, output_core_dims, error_context
+        )
+
+        # Rather than broadcasting all arguments to full broadcast shapes, prefer
+        # expanding dimensions using vmap. By pushing broadcasting
+        # into vmap, we can make use of more efficient batching rules for
+        # primitives where only some arguments are batched (e.g., for
+        # lax_linalg.triangular_solve), and avoid instantiating large broadcasted
+        # arrays.
```
unalmis, line 284: commenting for reference

Qazalbash:
@dpanici Hi, It is irrelevant to this discussion, but what script you used to create the plot pasted on top of the PR?

YigitElma:
> @dpanici Hi, It is irrelevant to this discussion, but what script you used to create the plot pasted on top of the PR?

Hi @Qazalbash ,
Sorry for the delayed response. I will share the script in 1 or 2 days (this is a self-reminder😄)

In the mean time, either @rahulgaur104 or @dpanici if you have the script in your archieve, can you share it? It was the one that used pynvml I guess.

rahulgaur104:
Hi @Qazalbash 

This should do the job, more or less.
Good luck!

```python
#!/usr/bin/env python3
import subprocess
import time
import threading

def monitor_vram(duration, interval, vram_usage_list, timestamps):
    end_time = time.time() + duration
    while time.time() < end_time:
        result = subprocess.run(['nvidia-smi', '--query-gpu=memory.used', '--format=csv,noheader,nounits'], 
                                stdout=subprocess.PIPE)
        output = result.stdout.decode('utf-8').strip()
        vram_usage = int(output.split()[0])
        vram_usage_list.append(vram_usage)
        timestamps.append(time.time())
        time.sleep(interval)


if __name__ == "__main__":
    duration = 300  #duration to monitor in seconds
    interval = 0.01  #interval between checks in seconds
    vram_usage_list = []
    timestamps = []

    #create threads for monitoring VRAM and running GPU code
    vram_thread = threading.Thread(target=monitor_vram, args=(duration, interval, vram_usage_list, timestamps))

    #start the threads
    vram_thread.start()

    res = 12
    #run without blocking (Popen)
    subprocess.Popen(['python', "DESC_profiler_new_13-04.py", f"{res}"])
    #subprocess.Popen(['python', "DESC_profiler_new_06-04.py", f"{res}"])

    #wait for the thread to finish
    vram_thread.join()

    #write the VRAM usage to a file
    with open('vram_usage.txt', 'w') as file:
        for usage in vram_usage_list:
            file.write(f"{usage}\n")

    plt.figure(figsize=(15, 7))
    #plot the VRAM usage
    plt.plot([t - timestamps[0] for t in timestamps], vram_usage_list, '-or', ms=2)
    plt.xlabel('Time (s)', fontsize=20)
    plt.ylabel('VRAM Usage (MiB)', fontsize=20)
    plt.xticks(fontsize=20)
    plt.yticks(fontsize=20)
    plt.title('GPU VRAM Usage Over Time')
    plt.grid(True)
    plt.tight_layout()
    #plt.savefig(f"test0_QP_{res}.png", dpi=400)
    plt.savefig(f"test0_QP_{res}_lowp.png", dpi=400)
    #plt.show()
```

Qazalbash:
Thanks @YigitElma and @rahulgaur104

YigitElma:
You're welcome! Also, remember that you have to set the environment variable 
```
XLA_PYTHON_CLIENT_ALLOCATOR=platform
```
to prevent JAX from allocating 75% of your memory, otherwise, you will get a flat line.

Qazalbash:
I really appreciate that! I will mention these in the docs.

PR #1053, opened 2024-06-13
Fix indenting to prevent repeated calculation in check_positive_jacobian()

YigitElma:


unalmis (approved):

f0uriest (approved):

YigitElma:
I merged this branch, but the benchmarks caused a question in my mind about their validity. For equilibrium initialization benchmarks, it says that there is no change basically and even the highest equilibrium init takes 6 seconds. I think this is due to the benchmarks creating some very basic surface (the default one when nothing is given to FourierRZToroidalSurface) and this doesn't represent most of the actual cases. For comparison, I was doing GPU profiling and resolution 15 equilibrium takes consistently 25 seconds with this surface
```python
surface_2D = FourierRZToroidalSurface(
        R_lmn=np.array([10, -1]),  # boundary coefficients                                                                                                                          
        Z_lmn=np.array([1]),
        modes_R=np.array([[0, 0], [1, 0]]),  # [M, N] boundary Fourier modes                                                                                                        
        modes_Z=np.array([[-1, 0]]),
        NFP=5,  # number of (toroidal) field periods (does not matter for 2D, but will for 3D solution)   
```
This is also very basic actually. Is this again some CPU/GPU difference issue? I was measuring at least 20% improvement on LMN=15 from 25 seconds to 20 seconds. Any thoughts @f0uriest @dpanici  ?

f0uriest:
My hunch would be cpu/gpu difference, but im surprised its that much. You can try the regular line profiler on cpu and see where the differences are

YigitElma:
Yes, I can do that tomorrow. Also, I run the test script on my small laptop (the one we use for GPU), it takes only 13 seconds and 9-10 seconds when functions are jitted. This is probably again some for loop being too slow on GPU

f0uriest:
certainly a possibility. If that's the case we should see if there's some small fixes we can do to make it faster on gpu

PR #1054, opened 2024-06-13
Add benchmark for running fixed iteration solve

f0uriest:
Should be merged before #1050 so benchmark will run.

### DIFF
--- a/tests/benchmarks/benchmark_cpu_small.py
+++ b/tests/benchmarks/benchmark_cpu_small.py
@@ -407,3 +407,17 @@ def test_proximal_freeb_jac(benchmark):
         obj.jac_scaled(x, prox.constants).block_until_ready()
 
     benchmark.pedantic(run, args=(x,), rounds=15, iterations=1)
+
+
+@pytest.mark.slow
+@pytest.mark.benchmark
+def test_solve_fixed_iter(benchmark):
+    """Benchmark running eq.solve for fixed iteration count."""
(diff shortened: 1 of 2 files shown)


unalmis (approved):

YigitElma (approved):

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -407,3 +407,17 @@ def run(x):
         obj.jac_scaled(x, prox.constants).block_until_ready()
 
     benchmark.pedantic(run, args=(x,), rounds=15, iterations=1)
+
+
+@pytest.mark.slow
+@pytest.mark.benchmark
+def test_solve_fixed_iter(benchmark):
+    """Benchmark running eq.solve for fixed iteration count."""
+    jax.clear_caches()
```
YigitElma, line 416: Actually, if we don't put `jax.clear_caches()` inside a setup function it will only be called once when this test runs. Do we want this or to clear cache before each solve? I assume we don't need to clear it every time since after first iteration of solve everything will be already jitted, but it is worth keeping in mind.

f0uriest: yeah we include `clear_caches` in the `setup` or `run` functions for some of the other tests (like the ones measuring compile time, or ones that are usually only run once like init) but in this case we're mostly interested in the average behavior over many iterations so it's not necessary.

PR #1055, opened 2024-06-13
JIT compute functions

YigitElma:


unalmis:
(I did this for Gamma_c and effective ripple since those are expensive computations; some arguments needed to be marked static there.)

YigitElma:
Although this PR is for testing mainly, if I manage to merge in #1027, most of the `type str 'xyz' not a valid JAX type` errors can be resolved. Do we have any other kwarg that we send to compute functions?

YigitElma:
Maybe move jit to main compute function? Current verdion cause excessive amount of compilation time once any of the compute functions is called!

YigitElma:
Should we just close this PR without trying to make tests past?

PR #1056, opened 2024-06-14
Execute initializations on CPU much faster

YigitElma:
Some of the initialization function use a lot of for loop that are extremely slow on GPUs. If we move the execution  to CPU, we get at least x5 performance for equilibrium and geometry classes. Since the Github actions are run on CPU, we won't see any speed up, but on GPU interactive sessions, this should speed up plotting etc.

Add the 
```python
with jax.default_device(jax.devices("cpu")[0]):
```
for the parts you believe don't need to be run on GPUs @f0uriest @dpanici @rahulgaur104 @ddudt  @kianorr @unalmis  

This was actually noticed on previous PR #1053 when the benchmark didn't result in speed up, but the GPU eq init got faster by 5 seconds.

image (1)
image


desc/equilibrium/equilibrium.py
```diff
@@ -182,211 +182,217 @@ def __init__(
...
-                "fringe",
-            ],
-            ValueError,
-            "spectral_indexing should be one of 'ansi', 'fringe', None, got "
-            + f"{spectral_indexing}",
-        )
-        self._spectral_indexing = setdefault(
-            spectral_indexing, getattr(surface, "spectral_indexing", "ansi")
-        )
+        with jax.default_device(jax.devices("cpu")[0]):
```
f0uriest, line 185: I like the idea of having something in backend that we can just decorate functions with, without having to mess with jax everywhere.

like
```
@oncpu
def __init__(self, ...):
```

YigitElma: I actually tried that with
```python 
def func(fun, *args, **kwargs):
       with jax.default_device(jax.devices("cpu")[0]):
               fun                     
set_default_device = func
```
But it complaint about input arguments like `R_lmn=...`, and the functions we need to use this operator doesn't have same input structure.

YigitElma: Or maybe I had a bug somewhere else. I would also like to have an operator instead of this version.

YigitElma:
Building objectives and transforms might also benefit CPU
image
image
The first build is the same, since it is wrapped by eq.init but the others are the same function

desc/backend.py
```diff
@@ -73,6 +73,7 @@
     vmap = jax.vmap
     scan = jax.lax.scan
     bincount = jnp.bincount
+    set_default_cpu = jax.default_device(jax.devices("cpu")[0])
```
f0uriest, line 76: i think we want something like this:

```python
def set_default_cpu(func):
    
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        with jax.default_device(jax.devices("cpu")[0]):
            return func(*args, **kwargs)
        
    return wrapper
```

dpanici (approved):

f0uriest (changes requested):
Are you sure we want all the build methods on cpu? I would think that at high resolution building the transforms would be faster on gpu.

Also see my note about using the context manager, I think what you have now might be setting cpu as the default globally.

YigitElma:
> Are you sure we want all the build methods on cpu? I would think that at high resolution building the transforms would be faster on gpu.
> 
> Also see my note about using the context manager, I think what you have now might be setting cpu as the default globally.

I will make some final checks again tomorrow but from previous tests, I saw that building transform LMN=15 is faster on CPU, both pre jit after jit versions. The only build that took longer on CPU before the start of optimization loop was the LinearConstraintProjection. It was 3 times slower on CPU.

I don't think it default everythinh to CPU, because as I add this decorators to new function step by step, I noticed some time changes. So, I believe it only affects the decorated function. To be sure, I can switch to your method, but current version is also working as intended.

dpanici (approved):

ddudt (approved):
Could we use this decorator on an objective function's `compute` method. It might be useful for the external objectives, so that everything else can run on GPU except the external codes that need to run on CPU. Or would that break JAX stuff?

f0uriest:
> Could we use this decorator on an objective function's `compute` method? It might be useful for the external objectives, so that everything else can run on GPU except the external codes that need to run on CPU. Or would that break JAX stuff?

No, this will only make a difference outside of JIT. Inside JIT everything has to be on a single device (can't mix cpu and gpu, this was what #763 was aiming to fix but haven't been able to get it working).

desc/backend.py
```diff
@@ -112,6 +113,28 @@ def put(arr, inds, vals):
             return arr
         return jnp.asarray(arr).at[inds].set(vals)
 
+    def set_default_cpu(func):
```
f0uriest, line 116: I feel like we could come up with a better/more descriptive name, `default_device_cpu`? or something? but not a dealbreaker

desc/objectives/linear_objectives.py
```diff
@@ -275,6 +275,7 @@ def __init__(
             name=name,
         )
 
+    @set_default_cpu
```
f0uriest, line 278: Do we want to apply this to all the build methods? It seems kind of inconsistent with where it is used now.

f0uriest (changes requested):
We should probably also add this in a number of other places:
- to the curve/coil init and change resolution methods
- magnetic field classes init?
- profiles?
- all build methods of all objectives?
- init/build for `LinearConstraintProjection` and `ProximalProjection`
- `factorize_linear_constraints`

YigitElma:
> We should probably also add this in a number of other places:
> - to the curve/coil init and change resolution methods
> - magnetic field classes init?
> - profiles?
> - all build methods of all objectives?
> - init/build for `LinearConstraintProjection` and `ProximalProjection`
> - `factorize_linear_constraints`

I have tried this for LinearConstraintProjection, it was very slow. I don't know others. I have tried to put it for build methods that have for loops on modes etc.

I can try digging more into profiling when I return back to Princeton, I cannot use Princeton VPN properly on Linux. It causes problems to connect visualization node etc

YigitElma:
> Could we use this decorator on an objective function's `compute` method? It might be useful for the external objectives, so that everything else can run on GPU except the external codes that need to run on CPU. Or would that break JAX stuff? 

I don't know tbh. It wouldn't hurt to try. The compute functions that I know didn't have not vectorized for loops.

ddudt:
> > Could we use this decorator on an objective function's `compute` method? It might be useful for the external objectives, so that everything else can run on GPU except the external codes that need to run on CPU. Or would that break JAX stuff?
> 
> No, this will only make a difference outside of JIT. Inside JIT everything has to be on a single device (can't mix cpu and gpu, this was what #763 was aiming to fix but haven't been able to get it working).

Hmm OK. So if an external objective has to run on a CPU (like some STELLOPT code), then the whole optimization has to also be on a CPU? That's a bummer but not the end of the world.

f0uriest:
> > > Could we use this decorator on an objective function's `compute` method? It might be useful for the external objectives, so that everything else can run on GPU except the external codes that need to run on CPU. Or would that break JAX stuff?
> > 
> > No, this will only make a difference outside of JIT. Inside JIT everything has to be on a single device (can't mix cpu and gpu, this was what #763 was aiming to fix but haven't been able to get it working).
> 
> Hmm OK. So if an external objective has to run on a CPU (like some STELLOPT code), then the whole optimization has to also be on a CPU? That's a bummer but not the end of the world. 

No in that case the jaxify stuff will allow you to run just the stellopt part on CPU. But it does that by basically exiting Jax. Within Jax you can't mix CPU/GPU

desc/objectives/linear_objectives.py
YigitElma: I've added it to the parent objective class before and for my test case it didn't give better results, so I picked suitable build methods intuitively. 

Also, maybe one important thing to keep in mind is when we build smt on cpu, probably the arrays are on cpu memory, we may need to copy them to gpu memory as we did during hackathon.

ddudt:
> > > > Could we use this decorator on an objective function's `compute` method? It might be useful for the external objectives, so that everything else can run on GPU except the external codes that need to run on CPU. Or would that break JAX stuff?
> > > 
> > > 
> > > No, this will only make a difference outside of JIT. Inside JIT everything has to be on a single device (can't mix cpu and gpu, this was what #763 was aiming to fix but haven't been able to get it working).
> > 
> > 
> > Hmm OK. So if an external objective has to run on a CPU (like some STELLOPT code), then the whole optimization has to also be on a CPU? That's a bummer but not the end of the world.
> 
> No in that case the jaxify stuff will allow you to run just the stellopt part on CPU. But it does that by basically exiting Jax. Within Jax you can't mix CPU/GPU

OK maybe we can move this conversation to PR #1028 where I will test this out. I was having trouble running on GPU before, but I think that was related to multiprocessing with JAX. I will test if this decorator works for the external objective.

dpanici (approved):

ddudt (approved):

f0uriest (approved):

PR #1058, opened 2024-06-18
updating perlmutter instructions

rahulgaur104:
Perlmutter instructions have been updated in the current docs because the old instructions fail.


docs/installation.rst
```diff
@@ -145,7 +145,19 @@ Create a conda environment for DESC
 
     conda create -n desc-env python=3.9
     conda activate desc-env
-    pip install --no-cache-dir "jax[cuda11_cudnn82]==0.4.7" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
+    pip install --no-cache-dir "jax==0.4.23" "jax[cuda12_cudnn89]==0.4.23" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
```
ddudt, line 148: Do we have to include the `"jax==0.4.23"` part? We usually don't have to include that

rahulgaur104: Maybe not. But it doesn't change anything so let's just keep it.

dpanici: Add link to perlmutter instructions

dpanici (approved):

PR #1059, opened 2024-06-18
Hotfix for changes in `jax` version `0.4.30`

dpanici:
- add ignore for the specific deprecation warning from interpax

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,8 +51,7 @@ filterwarnings=
     ignore:numpy.ndarray size changed:RuntimeWarning
     # ignore benign Cython warnings on ndarray size
     ignore::DeprecationWarning:(?!desc)
-    # ignore DeprecationWarning from interpax dependency
-
+    # ignore DeprecationWarnings from dependency packages
 
 [flake8]
 # Primarily ignoring whitespace, indentation, and commenting etiquette that black does not catch
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,9 @@ filterwarnings=
     # ignore benign Cython warnings on ndarray size
     ignore::DeprecationWarning:ml_dtypes.*
     # ignore benign ml_dtypes DeprecationWarning
+    ignore:jax.core.pp_eqn_rules:DeprecationWarning
+    # ignore DeprecationWarning from interpax dependency
+
 
 [flake8]
 # Primarily ignoring whitespace, indentation, and commenting etiquette that black does not catch
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,10 +48,11 @@ markers=
 filterwarnings=
     error
     ignore::pytest.PytestUnraisableExceptionWarning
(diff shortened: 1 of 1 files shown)


rahulgaur104 (approved):

f0uriest:
I think we should actually ignore all deprecation warnings that aren't from DESC itself. If one of our dependencies is using deprecated code, that should be fixed on their end, we really only want to get warned if we're using deprecated code ourselves.

setup.cfg
```diff
@@ -50,8 +50,8 @@ filterwarnings=
     ignore::pytest.PytestUnraisableExceptionWarning
     ignore:numpy.ndarray size changed:RuntimeWarning
     # ignore benign Cython warnings on ndarray size
-    ignore::DeprecationWarning:ml_dtypes.*
-    # ignore benign ml_dtypes DeprecationWarning
+    ignore::DeprecationWarning:(?!desc)
+    # ignore DeprecationWarnings from dependency packages
```
YigitElma, line 54: On this occasion, can you fix the comment order here? The comment being on top the line it explains is better imo

YigitElma: Or wait tests are almost done. We merge this in, and then fix that

ddudt: I agree that as a standard we should always put comments on the line above, not below, the code it is related to. Let's just fix it now because opening a whole new PR will take longer.

ddudt (changes requested):

rahulgaur104 (approved):

YigitElma (approved):

rahulgaur104 (approved):
Merge once the tests have passed.

setup.cfg
```diff
@@ -48,10 +48,11 @@ markers=
 filterwarnings=
     error
     ignore::pytest.PytestUnraisableExceptionWarning
-    ignore:numpy.ndarray size changed:RuntimeWarning
     # ignore benign Cython warnings on ndarray size
-    ignore::DeprecationWarning:ml_dtypes.*
-    # ignore benign ml_dtypes DeprecationWarning
+    ignore:numpy.ndarray size changed:RuntimeWarning
+    # ignore DeprecationWarnings from dependency packages
+    ignore::DeprecationWarning:(?!desc)
```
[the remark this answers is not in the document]
dpanici: I checked it locally (by calling a deprecated function from numpy inside of `coils.py`) and this causes that warning to propagate thru and be fiiltered as an error, so I think it does what we want (i.e. if we call directly the deprecated function it will warn us)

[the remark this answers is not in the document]
dpanici: but correctly does not cause an error on the deprecation warning raised inside of interpax

ddudt (approved):

f0uriest (approved):

YigitElma:
I assume the last benchmark fails since it is running things on master and not using updating setup.cfg?

dpanici:
Yes, b/c it runs on the master branch and so encounters the error this PR fixes

PR #1061, opened 2024-06-18
Allow `GenericObjective` and `ObjectiveFromUser` to work with generic "things"

### LINKED ISSUE #841: Allow `GenericObjective` and `ObjectiveFromUser` to work with generic "things"
Presently they are hard-coded to only work for an `Equilibrium`, and therefore cannot be used for other compute quantities.

ddudt:
Resolves #841


ddudt:
check if `Optimizable` or `OptimizableCollection` to handle different logic

desc/objectives/_generic.py
```diff
@@ -399,11 +400,16 @@ def build(self, use_jit=True, verbose=1):
...
         if self._grid is None:
-            grid = QuadratureGrid(eq.L_grid, eq.M_grid, eq.N_grid, eq.NFP)
-        else:
-            grid = self._grid
+            errorif(
+                self._p != "desc.equilibrium.equilibrium.Equilibrium",
+                ValueError,
+                "grid must be supplied for things besides an Equilibrium.",
+            )
+            self._grid = QuadratureGrid(
```
f0uriest, line 410: does grid need to be an attribute of self?

desc/objectives/_generic.py
```diff
@@ -351,17 +351,17 @@ def myfun(grid, data):
...
 
     _units = "(Unknown)"
-    _print_value_fmt = "Custom Objective value: {:10.3e}"
+    _print_value_fmt = "Custom objective value: {:10.3e}"
 
     def __init__(
         self,
         fun,
-        eq,
+        thing,
```
f0uriest, line 364: im not crazy about changing the name of arguments without a deprecation warning etc, so that we don't suddenly break user code.

I'd recommend changing the name, but add `**kwargs` and check if `eq` is in kwargs and throw a warning in that case. See https://github.com/PlasmaControl/DESC/blob/f955d3de2be694618593057ea9c0585595255ef2/desc/utils.py#L492-L493

desc/objectives/_generic.py
```diff
@@ -51,19 +52,19 @@ class GenericObjective(_Objective):
...
     """
 
-    _print_value_fmt = "GenericObjective value: {:10.3e} "
+    _print_value_fmt = "Generic objective value: {:10.3e} "
 
     def __init__(
         self,
         f,
-        eq,
+        thing,
```
f0uriest, line 67: see other comment about adding a deprecation warning for this change

f0uriest (changes requested):

desc/objectives/_generic.py
ddudt: Sort of. The `_Objective` class does not have a grid, so we always set the user-input grid as an attribute of self to pass it from the `__init__` method to the `build` method. I will change it to `grid = self._grid` for local use in the `build` method because I think that's what we do in other objectives. But the only difference is do we want `self._grid` to return `None` or the actual default value?

ddudt:
> check if `Optimizable` or `OptimizableCollection` to handle different logic

I thought about how to handle general `OptimizableCollections` and I think it would be tricky. Because these two objectives are calling `_compute` on some parameterization, and collections (like a `CoilSet`) are not a "computable parameterization" (they have their own compute logic to handle how to compute quantities for each of their members. We could account for it if we knew the tree structure of the collection, but that logic is difficult to handle for general trees (collections within collections). 

I added a check to throw a `NotImplementedError` if the user tries to use this on an `OptimizableCollection`. I think that is best for now.

f0uriest (approved):

dpanici (approved):

PR #1068, opened 2024-06-23
Partial derivatives (of the same coordinate system) commute

unalmis:
A few compute functions do a lot of work to compute the same thing as another, sometimes with slightly different math. It's better to just alias them to the other so that there is less math to check as mentioned  [here](https://github.com/PlasmaControl/DESC/pull/648#issuecomment-1718219257).

### DIFF
--- a/desc/compute/_basis_vectors.py
+++ b/desc/compute/_basis_vectors.py
@@ -1728,6 +1728,7 @@ def _e_sub_rho_rrr(params, transforms, profiles, data, **kwargs):
         "desc.geometry.surface.ZernikeRZToroidalSection",
     ],
     basis="basis",
+    aliases=["e_theta_rrr"],
 )
 def _e_sub_rho_rrt(params, transforms, profiles, data, **kwargs):
     data["e_rho_rrt"] = jnp.array(
@@ -1807,6 +1808,7 @@ def _e_sub_rho_rrt(params, transforms, profiles, data, **kwargs):
         "phi",
     ],
     basis="basis",
+    aliases=["e_zeta_rrr"],
 )
 def _e_sub_rho_rrz(params, transforms, profiles, data, **kwargs):
     data["e_rho_rrz"] = jnp.array(
@@ -1885,7 +1887,7 @@ def _e_sub_rho_rrz(params, transforms, profiles, data, **kwargs):
         "omega_t",
         "phi",
     ],
-    aliases=["x_rrt", "x_rtr", "x_trr"],
+    aliases=["x_rrt", "x_rtr", "x_trr", "e_theta_rr"],
     parameterization=[
         "desc.equilibrium.equilibrium.Equilibrium",
         "desc.geometry.surface.ZernikeRZToroidalSection",
@@ -1953,6 +1955,7 @@ def _e_sub_rho_rt(params, transforms, profiles, data, **kwargs):
         "desc.geometry.surface.ZernikeRZToroidalSection",
     ],
     basis="basis",
+    aliases=["e_theta_rrt"],
 )
 def _e_sub_rho_rtt(params, transforms, profiles, data, **kwargs):
     data["e_rho_rtt"] = jnp.array(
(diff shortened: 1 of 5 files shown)


desc/compute/_basis_vectors.py
```diff
@@ -1811,59 +1811,43 @@ def _e_sub_rho_rrt(params, transforms, profiles, data, **kwargs):
 def _e_sub_rho_rrz(params, transforms, profiles, data, **kwargs):
     data["e_rho_rrz"] = jnp.array(
         [
-            -2 * data["omega_rz"] * data["R_r"] * data["omega_r"]
```
unalmis, line 1814: this is the same as the other existing mixed partial; there's nothing to check here just copy pasted

dpanici (changes requested):
Good catch, @kianorr has previously implemented an `alias` keyword to the decorator just for this purpose, can you remove the duplicate compute definitions and use the `alias` kwarg to accomplish this instead?

example is here: https://github.com/PlasmaControl/DESC/blob/f7383817d9aee40d54fec1d9a36e685d0cfe7c0c/desc/compute/_metric.py#L1373

dpanici:
Also maybe in `test_compute_everything` we also compute the aliases? or would that be too much duplicated computation? I am fine either way, as of now this PR works and everything is also tested bc of that test, but with the `alias` method, there won't be a test that the aliased things are computed correctly, idt.

unalmis:
> Also maybe in `test_compute_everything` we also compute the aliases? or would that be too much duplicated computation? I am fine either way, as of now this PR works and everything is also tested bc of that test, but with the `alias` method, there won't be a test that the aliased things are computed correctly, idt.

i don't think this is worth testing

desc/compute/_basis_vectors.py
```diff
@@ -1969,6 +1955,7 @@ def _e_sub_rho_rt(params, transforms, profiles, data, **kwargs):
         "desc.geometry.surface.ZernikeRZToroidalSection",
     ],
     basis="basis",
+    aliases=["e_theta_rrt"],
```
unalmis, line 1958: permutations like e_theta_rtr are connected automatically

unalmis:
If any of the quantities with an alias in the list of variables [here](https://desc-docs--1068.org.readthedocs.build/en/1068/variables.html) can be discontinuous, then we should undo commit d1863062f4ce83008e29a79e564a9e65eb00688e before merging. (The basis vectors are always continuous, so the jacobian derivatives etc. are too - and i don't think desc can get discontinuous derivatives of |B|)

kianorr:
> If any of the quantities with an alias in the list of variables [here](https://desc-docs--1068.org.readthedocs.build/en/1068/variables.html) can be discontinuous, then we should undo commit [d186306](https://github.com/PlasmaControl/DESC/commit/d1863062f4ce83008e29a79e564a9e65eb00688e) before merging. (The basis vectors are always continuous, so the jacobian derivatives etc. are too - and i don't think desc can get discontinuous derivatives of |B|)

what do you mean by discontinuous/continuous?

unalmis:
> > If any of the quantities with an alias in the list of variables [here](https://desc-docs--1068.org.readthedocs.build/en/1068/variables.html) can be discontinuous, then we should undo commit [d186306](https://github.com/PlasmaControl/DESC/commit/d1863062f4ce83008e29a79e564a9e65eb00688e) before merging. (The basis vectors are always continuous, so the jacobian derivatives etc. are too - and i don't think desc can get discontinuous derivatives of |B|)
> 
> what do you mean by discontinuous/continuous?

For example, the components of B tangential to a surface current will be discontinuous. It's not as obvious whether interchanging partial derivatives is valid near discontinuities, and one would have to do math or seek weaker conditions to justify it. I don't think we need to worry about this - all the derivatives of B are compositions of continuous functions as we have implemented. I was mainly just documenting the assumptions for commit [d186306](https://github.com/PlasmaControl/DESC/commit/d1863062f4ce83008e29a79e564a9e65eb00688e) in case anyone changes stuff in future.

dpanici (approved):

ddudt (approved):

PR #1069, opened 2024-06-23
Fix for JAX Errors in Coil Optimization

dpanici:
Skips the check to re-use transforms if `jitable=True` is passed to `get_transforms`

This is a hotfix to allow coil optimization to run without errors due to the checks in `get_transforms`

### DIFF
--- a/desc/compute/utils.py
+++ b/desc/compute/utils.py
@@ -361,7 +361,7 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
                         build=False,
                         method=method,
                     )
-            else:
+            else:  # don't perform checks if jitable=True as they are not jit-safe
                 c_transform = Transform(
                     grid,
                     basis,
--- a/desc/compute/utils.py
+++ b/desc/compute/utils.py
@@ -343,16 +343,25 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
         if hasattr(obj, c + "_basis"):  # regular stuff like R, Z, lambda etc.
             basis = getattr(obj, c + "_basis")
             # first check if we already have a transform with a compatible basis
-            for transform in transforms.values():
-                if basis.equiv(getattr(transform, "basis", None)):
-                    ders = np.unique(
-                        np.vstack([derivs[c], transform.derivatives]), axis=0
-                    ).astype(int)
-                    # don't build until we know all the derivs we need
-                    transform.change_derivatives(ders, build=False)
-                    c_transform = transform
-                    break
-            else:  # if we didn't exit the loop early
+            if not jitable:
+                for transform in transforms.values():
+                    if basis.equiv(getattr(transform, "basis", None)):
+                        ders = np.unique(
+                            np.vstack([derivs[c], transform.derivatives]), axis=0
+                        ).astype(int)
+                        # don't build until we know all the derivs we need
+                        transform.change_derivatives(ders, build=False)
+                        c_transform = transform
+                        break
+                else:  # if we didn't exit the loop early
+                    c_transform = Transform(
+                        grid,
+                        basis,
+                        derivs=derivs[c],
+                        build=False,
+                        method=method,
+                    )
+            else:
                 c_transform = Transform(
                     grid,
                     basis,
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -10,7 +10,13 @@ from qic import Qic
 from qsc import Qsc
 
 from desc.backend import jnp
-from desc.coils import FourierPlanarCoil, FourierRZCoil, MixedCoilSet
+from desc.coils import (
+    CoilSet,
+    FourierPlanarCoil,
+    FourierRZCoil,
+    FourierXYZCoil,
+    MixedCoilSet,
+)
 from desc.continuation import solve_continuation_automatic
 from desc.equilibrium import EquilibriaFamily, Equilibrium
 from desc.examples import get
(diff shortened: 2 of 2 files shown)


desc/compute/utils.py
```diff
@@ -343,16 +343,25 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
...
+                        break
+                else:  # if we didn't exit the loop early
+                    c_transform = Transform(
+                        grid,
+                        basis,
+                        derivs=derivs[c],
+                        build=False,
+                        method=method,
+                    )
+            else:  # don't perform checks if jitable=True as they are not jit-safe
```
kianorr, line 364: is it the `if` statement that isn't jit-safe? and do we know why they are not jit-safe?

desc/compute/utils.py
```diff
@@ -343,16 +343,25 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
...
-                        np.vstack([derivs[c], transform.derivatives]), axis=0
-                    ).astype(int)
-                    # don't build until we know all the derivs we need
-                    transform.change_derivatives(ders, build=False)
-                    c_transform = transform
-                    break
-            else:  # if we didn't exit the loop early
+            if not jitable:
+                for transform in transforms.values():
+                    if basis.equiv(getattr(transform, "basis", None)):
```
kianorr, line 348: What is this checking (if this is the check you were referencing below)? Is it important?

tests/test_examples.py
```diff
@@ -1276,6 +1282,53 @@ def test_second_stage_optimization():
     np.testing.assert_allclose(field[1].B0, 0, atol=1e-12)  # vertical field (vanishes)
 
 
+@pytest.mark.unit
+def test_second_stage_optimization_CoilSet():
```
kianorr, line 1286: Are you able to check `jitable=True` and `jitable=False` somehow? Or maybe it's somewhere already and I don't see it

desc/compute/utils.py
dpanici: No, it is the equivalence check of the basis, and the use of numpy in the block above. The JAX error is a bit esoteric but basically we are converting a traced quantity to a numpy array during the check and that is not ok to do,I think

desc/compute/utils.py
dpanici: R, Z and lambda, for example each need their own transform to compute the basis functions to calculate them. however usually these are all the same resolution (L, M, N) and Z and lambda are always the same symmetry (`"sin"` if stell symmetric) and in the case of a stellarator asymmetric equilibrium, R Z and lambda are all asymmetric bases. So, we dont need to have 2 (or 3) separate transforms for these quantiites, we can just reuse the same one (if stell sym, that means Z and lambda can share a transform, if asymmetric, means that all 3 can use the same transform)

the check here is checking if we can do that. It is nice to save some memory but is not completely necessary so I think it is fine to skip the check whenever `jitable=True` is passed

tests/test_examples.py
dpanici: It is implicitly inside of the compute function calls of the `QuadraticFlux` objective, sothe `jitable=True` part is checked, and we should not need to check the False case as that is never called inside of jit. The False cases are whenever we call `thing.compute` as `jitable=False` by default

ddudt: Are you sure this test doesn't already pass without these changes? Was the bug specific to `CoilSet`s and not individual `Coil`s? I'm curious why this bug is not a problem for `test_optimize_with_fourier_planar_coil` or any of the new tests Kian has added in PR #1025

dpanici: The error is because of `compute_magnetic_field`, where `curve.compute` is called. We had no tests with `QuadraticFlux` or any other objective which calls `coil.compute_magnetic_field` before now. If you don't believe it, try running this test on `master`, that should convince you

f0uriest (dismissed):

ddudt (approved):

desc/compute/utils.py
```diff
@@ -343,16 +343,25 @@ def get_transforms(keys, obj, grid, jitable=False, **kwargs):
...
-                if basis.equiv(getattr(transform, "basis", None)):
-                    ders = np.unique(
-                        np.vstack([derivs[c], transform.derivatives]), axis=0
-                    ).astype(int)
-                    # don't build until we know all the derivs we need
-                    transform.change_derivatives(ders, build=False)
-                    c_transform = transform
-                    break
-            else:  # if we didn't exit the loop early
+            if not jitable:
```
YigitElma, line 346: Are we sure this does what you want? It makes sense to assume that the compiler skips this if statement completely for jitable cases hence able to jit it without error, but I have never tried to put not jitable parts of the function inside an if statement. If it works this is cool

YigitElma: Can you check `jax.jit(get_transforms)(..., jitable=True)` to see if it works in addition to test you added below? Don't add a new test but I have some doubts that this is actually jitable.

f0uriest: This does work as intended, as long as jitable itself is static. We do a similar thing in the grid class

ddudt (changes requested):
Your solution works, but I have a request for an additional fix. `QuadraticFlux.compute` calls `_Coil.compute_magnetic_field`, which calls `Curve.compute`. Then the error is being thrown at the line: 
```
if transforms is None:
    transforms = get_transforms(
        names, obj=self, grid=grid, jitable=True, **kwargs
    )
```
You fixed the issue when creating these transforms inside a jitable function. But it is wasteful to keep re-building these same transforms each time the objective is computed. It would be safer and more efficient to pre-compute these transforms in `QuadraticFlux.build` and then pass them through to the relevant compute functions. That would also fix this bug and speed up the code.

dpanici:
I would like to do that change in another in #1002 or another PR @ddudt . Let this one be a hotfix and I can think of a better solution for the transform pass-through this week(We were not passing through the transforms before this PR anyways so it is not like this one makes things worse)

ddudt (approved):
I made a new issue about my point of pre-computing the transforms. I think that is the better solution to this issue, so if you don't add it to this PR we should still do it before the major release with the coil optimization.

f0uriest (approved):

PR #1073, opened 2024-06-25
Rescale optimizer termination criteria

### LINKED ISSUE #797: normalize stopping criteria
Everything we optimize is typically of order unity, except for the pressure (or kinetic) and current profiles. This isn't an issue for the optimization steps where things are already normalized relative to the Jacobian, but it could affect the stopping criteria. Most notably `gtol` can have weird units. We should think of a way to account for the norm of x in the optimizers.

f0uriest:
Most of our optimizers have 3 main exit criteria:
$$||\Delta f || / ||f|| < ftol$$
$$||\Delta x || / ||x|| < xtol$$
$$||\nabla f ||_\infty  < gtol$$

The normalizations in the objective functions generally ensure that the first one (`ftol`) is more or less independent of scaling/units etc, however `xtol` and `gtol` still have units of `x` or `1/x` respectively, meaning that rescaling variables can affect the stopping criteria. This is especially noticeable when doing coil optimization or optimizing a current profile, since current is ~1e6x larger than the other degrees of freedom, it can mean exiting early once the current reaches a good value but the other variables can still be far from optimal.

This PR adds an option `scaled_termination` (defaults to True) to all of the desc optimizers to measure the norms for `xtol` and `gtol` in the scaled norm provided by `x_scale` (which defaults to using an adaptive scaling based on the Jacobian or Hessian). This should make things a bit better when optimizing parameters with widely different magnitudes.

Resolves #797


YigitElma:
A suggestion for the intended functionality, can we just take the ratio `step_ratio=step/x` and check if any element of `step_ratio` is smaller then some new tolerance `rtol`?

tests/conftest.py
```diff
@@ -284,13 +284,13 @@ def DummyMixedCoilSet(tmpdir_factory):
         vf_coil, displacement=[0, 0, 2], n=3, endpoint=True
     )
     xyz_coil = FourierXYZCoil(current=2)
-    phi = 2 * np.pi * np.linspace(0, 1, 20, endpoint=True) ** 2
+    phi = 2 * np.pi * np.linspace(0, 1, 20, endpoint=True)
```
YigitElma: Not that it's important but why this change?

tests/test_examples.py
```diff
@@ -877,7 +877,7 @@ def mirrorRatio(params):
     )
     optimizer = Optimizer("lsq-auglag")
     (eq, field), _ = optimizer.optimize(
-        (eq, field), objective, constraints, maxiter=100, verbose=3
+        (eq, field), objective, constraints, maxiter=150, verbose=3
```
YigitElma, line 880: Question: Is this necessary for the test to pass or to see If it terminates properly? Why would changing the stopping criterion change the end result after 100 steps? I mean it doesn't change the step itself, right?

f0uriest: for the augmented lagrangian methods this also affects the internal criteria to determine when to update the multipliers and penalty param, so it can affect the results somewhat. The old test wasn't actually terminating on a tolerance, it was just running to maxiter, now it correctly terminates when it reaches an actual tolerance

tests/conftest.py
f0uriest: a few lines below we set phi = linspace, not linspace**2 so this just makes the knots match where we say they are

YigitElma (approved):

desc/optimize/least_squares.py
```diff
@@ -348,7 +353,7 @@ def lsqtr(  # noqa: C901
             success, message = check_termination(
```
YigitElma, line 353: @f0uriest can you also solve this TODO? I am pretty sure that this has to be taken outside. Because the while loop condition already checks the only important criteria for the trust region problem. Stopping the entire optimization based on a step that is not even the step we are taking might be causing unintended termination?

YigitElma (changes requested):
We can also resolve #1395 here

desc/optimize/least_squares.py
f0uriest: See response on #1395, we do need this check here to avoid infinite loops.

YigitElma: Can you delete it then?

YigitElma (dismissed):

dpanici (commented):
Do we need to re-run any notebooks? or do they all not change?

dpanici:
The advanced optimization notebook has some differing results now (probably due to differing termination events of the eq subproblem with this branch vs master, at least for the proximal case), the auglag result is slightly worse and the QS results change slightly too. I think it would be good to run all the notebooks again (or at least some of them, or make an issue/other branches that runs them and updates them) so that what we have up on the docs is what actually a user would see when running the notebook.

YigitElma (dismissed):

dpanici (approved):
Approving assuming #1406 is ran based off of this PR and merged in soon as well

YigitElma (approved):

dpanici (approved):

YigitElma:
Should we merge this before new release, or you want to wait until #1406 ?

f0uriest:
> Should we merge this before new release, or you want to wait until #1406 ?

I think we can wait till the next one.

PR #1076, opened 2024-06-26
Avoid NaN in reverse mode AD of Omnigenity

### LINKED ISSUE #943: `Omnigenity` objective is not reverse-mode differentiable
The issue seems to arise from the fact that `field_data` needs the value of iota from `eq_data`. If `eq_fixed=True` or the value of iota is otherwise hard-coded, then the objective is compatible with reverse-mode AD. 

If a solution is found, remember to:

- [ ] change the default to `deriv_mode="auto"`
- [ ] add the `Omnigenity` objective to `test_objective_no_nangrad`

f0uriest:
Resolves #943


dpanici (approved):

desc/compute/_omnigenity.py
```diff
@@ -443,15 +443,25 @@ def _omni_map(params, transforms, profiles, data, **kwargs):
...
+    # need a bunch of wheres to avoid division by zero causing NaN in backward pass
+    # this is fine since the incorrect values get ignored later, except in OT or OH
+    # where fieldlines are exactly parallel to |B| contours, but this is a degenerate
+    # case of measure 0 so this kludge shouldn't affect things too much.
+    mat_01 = jnp.array(
+        [[N, iota / jnp.where(N == 0, 1, N)], [0, 1 / jnp.where(N == 0, 1, N)]]
+    )  # OP
+    mat_10 = jnp.array([[0, -1], [M, -1 / jnp.where(iota == 0, 1.0, iota)]])  # OT
+    den = jnp.where((N - M * iota) == 0, 1.0, (N - M * iota))
+    mat_11 = jnp.array([[N, M * iota / den], [M, M / den]])  # OH
```
ddudt, line 455: What's the naming convention of `mat_01` vs `_10` vs `_11`? I suggest renaming to `mat_OP`, `_OT`, `_OH` for clarity.

desc/objectives/_omnigenity.py
```diff
@@ -891,17 +891,25 @@ def compute(self, params_1=None, params_2=None, constants=None):
             # update theta_B and zeta_B with new iota from the equilibrium
             M, N = constants["helicity"]
             iota = jnp.mean(eq_data["iota"])
+            # see comment in desc.compute._omnigenity for the explanation of these
+            # wheres
+            mat_01 = jnp.array(
```
ddudt, line 896: see previous comment about these variable names

ddudt (commented):
This must have broken something because the main omnigenity optimization test now fails

desc/compute/_omnigenity.py
f0uriest: this was meant to be like `mat_MN` where `MN in [01, 10, 11]`, but I like your idea better.

f0uriest:
> This must have broken something because the main omnigenity optimization test now fails

It seems to be some jax issue about escaped tracers and functions with side effects, which I can't seem to figure out...

desc/objectives/_omnigenity.py
```diff
@@ -891,20 +891,26 @@ def compute(self, params_1=None, params_2=None, constants=None):
...
-                    jnp.array(
-                        [N, M * iota / (N - M * iota), M, M / (N - M * iota)]  # OH
-                    ),
+                    mat_OT,
+                    mat_OH,
                 ),
-            ).reshape((2, 2))
+            )
             booz = matrix @ jnp.vstack((field_data["alpha"], field_data["h"]))
-            field_data["theta_B"] = booz[0, :]
```
f0uriest, line 906: in this if block, `field_data` is a constant, but `booz` is a traced value, and we can't add traced values to constant dicts. This should have caused errors before but for some reason only shows up on the backward pass.

rahulgaur104 (dismissed):

tests/test_objective_funs.py
```diff
@@ -2418,6 +2418,35 @@ def test_objective_no_nangrad_coils(self, objective):
...
+            NFP=1,
+            sym=True,
+        )
+        eq = Equilibrium(Psi=6e-3, M=4, N=4, surface=surf)
+        field = OmnigenousField(
+            L_B=0,
+            M_B=2,
+            L_x=0,
+            M_x=0,
+            N_x=0,
```
ddudt, line 2440: I'm not certain it will make a difference, but we should increase these resolutions to be non-zero. If the coordinate mapping is constant then there aren't actually any free parameters involved in the part of the code that was causing the NaNs. 

Also you are using a QP equilibrium and testing with OH/OT but I think that should be fine for only testing nan grad

ddudt (changes requested):

ddudt (approved):

dpanici (approved):

unalmis:
https://github.com/PlasmaControl/DESC/blob/5d3af143a37ed83009a26f63f67fe8fee765868a/tests/test_objective_funs.py#L2275 ?

PR #1080, opened 2024-06-26
Have current computations use the current profile if assigned to an `Equilibrium`?

dpanici:
- `"current"` compute quantity and its derivatives were previously not using the current profile assigned to an equilibrium

Is this something intended? On second thought I assume yes as current never gets used anywhere directly and where it matters is iota, which uses the profiles correctly, just wanted to bring this up and ask


unalmis:
a year ago you mentioned:
I think if we did it for I that could mess up other QS parts that rely on calculating I properly

dpanici:
Closing as old me answered the question apparently, thanks @unalmis

PR #1082, opened 2024-06-26
Fix free boundary tutorial notebook

### LINKED ISSUE #1081: SOLOVEV Free boundary tutorial not giving expected solution

f0uriest:
Splits SOLOVEV free boundary solve into an initial phase without the sheet current, then adds the sheet current in later. This seems to be much more robust (and is what I've been doing in most of the benchmarks for the paper).

Resolves #1081


rahulgaur104 (approved):

kianorr:
What's the difference between adding in the sheet current earlier or later? And earlier or later than what?

dpanici (approved):

f0uriest:
> What's the difference between adding in the sheet current earlier or later? And earlier or later than what?

Including the sheet current makes the problem more ill conditioned, such that if you include it from the start where the boundary shape is far from correct, it can lead to incorrect solutions (it seems like it finds a solution with a large sheet current but wrong boundary shape). We expect the sheet current to be small in most cases, so it's usually safer to leave it out at the start and just vary the boundary shape until its close, then allow the sheet current to change as well for a small correction.

PR #1085, opened 2024-06-27
signed curvature

### LINKED ISSUE #1039: signed curvature
Our curvature compute quantities are always positive. It would be useful to define a signed curvature to distinguish between convex and concave shapes. This should be particularly easy to implement for the `FourierPlanarCurve` class. We could either: 

- Update the `FourierPlanarCurve` compute quantity to be the signed version, leaving all of the other curves with the positive version. Or
- Add a new compute quantity for "signed curvature" or something, which is not implemented for the other curve types (return `NaN`?)

ddudt:
Resolves #1039 

Also fixes a bug in `FourierRZCurve.compute(["x_s", "x_ss", "x_sss"])`. 

Need to add tests for the following: 

- [x] new compute quantity `"center"`
- [x] `FourierPlanarCoil` with both `"xyz"` and `"rpz"` basis
- [x] update `CoilCurvature` objective tests


desc/compute/_curve.py
```diff
@@ -146,6 +146,50 @@ def _Z_Curve(params, transforms, profiles, data, **kwargs):
...
+    dim=3,
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="s",
+    data=["x"],
+    parameterization="desc.geometry.core.Curve",
+)
+def _center_Curve(params, transforms, profiles, data, **kwargs):
+    data["center"] = jnp.mean(data["x"], axis=0) * jnp.ones_like(data["x"])
```
f0uriest, line 164: this will be weird if the grid only has 1 point. It might be better to make specialized versions for the different classes that get the n=0 fourier mode or average of the spline knots etc.

dpanici: must be in xyz for this to work too

desc/compute/_curve.py
```diff
@@ -325,6 +350,32 @@ def _x_sss_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
...
+    coordinates="s",
+    data=["x"],
+    parameterization="desc.geometry.curve.FourierRZCurve",
+)
+def _center_FourierRZCurve(params, transforms, profiles, data, **kwargs):
+    idx_Rc = transforms["R"].basis.get_idx(N=1, error=False)
+    idx_Rs = transforms["R"].basis.get_idx(N=-1, error=False)
+    idx_Z = transforms["Z"].basis.get_idx(error=False)
+    X0 = params["R_n"][idx_Rc] / 2 if isinstance(idx_Rc, int) else 0
+    Y0 = params["R_n"][idx_Rs] / 2 if isinstance(idx_Rs, int) else 0
```
ddudt, line 372: I'm not 100% on this math, someone please double check it for me

f0uriest: checked it, looks correct to me

desc/compute/_curve.py
```diff
@@ -325,6 +353,35 @@ def _x_sss_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
...
+    transforms={"R": [[0, 0, 0]], "Z": [[0, 0, 0]]},
+    profiles=[],
+    coordinates="s",
+    data=["x"],
+    parameterization="desc.geometry.curve.FourierRZCurve",
+)
+def _center_FourierRZCurve(params, transforms, profiles, data, **kwargs):
+    idx_Rc = transforms["R"].basis.get_idx(N=1, error=False)
+    idx_Rs = transforms["R"].basis.get_idx(N=-1, error=False)
+    idx_Z = transforms["Z"].basis.get_idx(error=False)
```
f0uriest, line 373: missing an N=0 here?

desc/compute/_curve.py
```diff
@@ -432,28 +484,54 @@ def _x_ss_FourierRZCurve(params, transforms, profiles, data, **kwargs):
...
+    dim=3,
+    params=["X_n", "Y_n", "Z_n", "rotmat", "shift"],
+    transforms={"X": [[0, 0, 0]], "Y": [[0, 0, 0]], "Z": [[0, 0, 0]]},
+    profiles=[],
+    coordinates="s",
+    data=["x"],
+    parameterization="desc.geometry.curve.FourierXYZCurve",
+)
+def _center_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
+    idx_X = transforms["X"].basis.get_idx(error=False)
```
f0uriest, line 521: better to be explicit about what index you're getting, if only for readability

desc/compute/_curve.py
```diff
@@ -568,6 +646,31 @@ def _x_sss_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
 
 
+@register_compute_fun(
+    name="center",
+    label="\\langle\\mathbf{x}\\rangle",
+    units="m",
+    units_long="meters",
+    description="Centroid of the curve",
+    dim=3,
+    params=["X", "Y", "Z", "shift"],
```
f0uriest, line 656: missing rotmat here

desc/compute/_curve.py
```diff
@@ -794,13 +897,12 @@ def _frenet_tangent(params, transforms, profiles, data, **kwargs):
     transforms={},
     profiles=[],
     coordinates="s",
-    data=["x_ss"],
+    data=["x_s", "x_ss", "frenet_tangent"],
```
f0uriest, line 900: frenet_tangent looks unused here?

desc/compute/_curve.py
```diff
@@ -830,20 +932,22 @@ def _frenet_binormal(params, transforms, profiles, data, **kwargs):
     label="\\kappa",
     units="m^{-1}",
     units_long="Inverse meters",
-    description="Scalar curvature of the curve",
+    description="Scalar curvature of the curve, positive/negative = concave/convex",
```
f0uriest, line 935: what is concave/convex in relation to here? Might be clearer to say whether the curve is curving towards the center or away?

f0uriest: or maybe pictorially:
`*center   )+curvature    (-curvature`

desc/objectives/_coils.py
```diff
@@ -395,7 +396,7 @@ def __init__(
         name="coil curvature",
     ):
         if target is None and bounds is None:
-            bounds = (0, 1)
+            bounds = (-1, 0)
```
f0uriest, line 399: might want to throw a warning if the user passes in bounds like (0, max) since that wouldn't be possible i think (if a circle has curvature -1)? For people used to unsigned curvature this might be confusing. Or maybe we just flip our sign convention so that a circle has positive curvature? though that slighlty conflicts with our convention for surface curvature

f0uriest (changes requested):
few typos, main comment is about the sign convention and possible confusion with people used to unsigned curvature.

Also, suppose we flipped our parameterization s -> -s, then the frenet frame would flip, and wouldn't that make the curvature change sign? even though the curve is the same?

desc/objectives/_coils.py
f0uriest: could also have a flag for whether to use signed or unsigned curvature?

ddudt: I wish our sign convention was flipped (positive = convex), but I went with the sign convention we already use for the surface curvature. I would be fine with changing them both. 

A closed curve has to be convex at some points, but in theory you could target a curve to be locally concave in certain regions. And I think there is a conceivable use case for targeting positive curvature everywhere, even though it is not possible, to try and force the coils to be straight (low curvature everywhere). 

Instead of a flag for signed/unsigned, I think the better solution would be to add a `loss_function="abs"` option. But that can be a separate PR.

ddudt:
> few typos, main comment is about the sign convention and possible confusion with people used to unsigned curvature.
> 
> Also, suppose we flipped our parameterization s -> -s, then the frenet frame would flip, and wouldn't that make the curvature change sign? even though the curve is the same?

If we flip the parameterization, the tangent vector will flip but the normal vector will still point inwards towards the center of local curvature. The normal vector is what gets used to determine the sign of the curvature, so I think this should be fine.

tests/inputs/master_compute_data_rpz.pkl
ddudt, line 1: "frenet_normal" and "frenet_binormal" changed for all curves, because the old values were wrong. "curvature" flipped sign for all curves now that it is a signed quantity.

desc/objectives/_coils.py
f0uriest: I agree the sign convention is annoying, I based it on several textbooks which define positive curvature of a surface as curving *towards* the outward normal vector to the surface. 

I don't think we necessarily have to keep the same convention for curves, since signed curvature for non-planar curves isn't even necessarily well defined, and to me it makes more sense to define it such that a circle has positive curvature.

At the very least, I think we should include a note in the docstring for the curvature objective reminding users of the sign convention.

desc/compute/_curve.py
```diff
@@ -830,20 +932,23 @@ def _frenet_binormal(params, transforms, profiles, data, **kwargs):
...
     parameterization="desc.geometry.core.Curve",
 )
 def _curvature(params, transforms, profiles, data, **kwargs):
+    # magnitude of curvature
     dxn = jnp.linalg.norm(data["x_s"], axis=-1)[:, jnp.newaxis]
-    data["curvature"] = jnp.linalg.norm(
-        cross(data["x_s"], data["x_ss"]) / dxn**3, axis=-1
-    )
+    curvature = jnp.linalg.norm(cross(data["x_s"], data["x_ss"]) / dxn**3, axis=-1)
+    # sign of curvature (positive = "concave", negative = "convex")
```
dpanici, line 949: I think I vote for positive for convex but I am like barely off the fence on it

desc/objectives/_coils.py
```diff
@@ -982,15 +987,15 @@ def build(self, use_jit=True, verbose=1):
             "Plasma/Surface grid includes interior points, should be rho=1.",
         )
 
-        self._dim_f = coils.num_coils
+        self._dim_f = coil.num_coils
```
dpanici, line 990: either change name `coils` or add a check in this init to say it must have a coilset passed in

desc/objectives/_coils.py
```diff
@@ -982,15 +981,15 @@ def build(self, use_jit=True, verbose=1):
...
```
ddudt, line 990: `_Coil.num_coils = 1`

desc/objectives/_generic.py
```diff
@@ -257,13 +257,14 @@ def build(self, use_jit=False, verbose=1):
...
+            param += np.random.rand(param.size) * 10
+        params = tree_unflatten(params_tree, params)
         J1 = jax.jacrev(self._fun)(thing.params_dict)
-        params = thing.params_dict.copy()
-        for key, value in params.items():
-            params[key] = value + np.random.rand(value.size) * 10
         J2 = jax.jacrev(self._fun)(params)
-        for key in J1.keys():
-            assert np.all(J1[key] == J2[key]), "Function must be linear!"
+        for j1, j2 in zip(tree_leaves(J1), tree_leaves(J2)):
```
dpanici, line 266: why was this change needed?

desc/objectives/_coils.py
```diff
@@ -746,7 +745,7 @@ class CoilSetMinDistance(_Objective):
 
     def __init__(
         self,
-        coils,
+        coil,
```
ddudt, line 754: add error check that input `coil` is a `CoilSet` and not a `_Coil`

tests/test_curves.py
```diff
@@ -17,6 +17,20 @@
 class TestRZCurve:
     """Tests for FourierRZCurve class."""
 
+    @pytest.mark.unit
+    def test_center(self):
+        """Test center of curve."""
+        c = FourierRZCurve(R_n=[-2, 10, 4], Z_n=[1, 3, 2])
```
dpanici, line 23: also add rpz to this test to increase coverage

tests/test_curves.py
```diff
@@ -461,6 +493,24 @@ def test_asserts(self):
 class TestPlanarCurve:
     """Tests for FourierPlanarCurve class."""
 
+    @pytest.mark.unit
+    def test_center(self):
+        """Test center of curve."""
+        c = FourierPlanarCurve(center=[5, 4, 3], r_n=[1, 2, 0.5], basis="xyz")
```
dpanici, line 499: also add one with basis_in as rpz to increase test coverage

dpanici (changes requested):
address changes and it will be good to go

dpanici:
And fix the docs

desc/compute/_curve.py
```diff
@@ -146,6 +146,34 @@ def _Z_Curve(params, transforms, profiles, data, **kwargs):
...
+    coordinates="s",
+    data=["x"],
+    parameterization="desc.geometry.curve.FourierPlanarCurve",
+    basis_in="{'rpz', 'xyz'}: Basis for input params vectors, Default 'xyz'",
+)
+def _center_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
+    # convert to xyz
+    if kwargs.get("basis_in", "xyz").lower() == "rpz":
+        center = rpz2xyz(params["center"])
+    else:
```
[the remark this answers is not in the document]
ddudt: This is already done [here](https://github.com/PlasmaControl/DESC/blob/dd/curvature/tests/test_curves.py#L509)

desc/compute/_curve.py
```diff
@@ -325,6 +353,35 @@ def _x_sss_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
...
+    parameterization="desc.geometry.curve.FourierRZCurve",
+)
+def _center_FourierRZCurve(params, transforms, profiles, data, **kwargs):
+    idx_Rc = transforms["R"].basis.get_idx(N=1, error=False)
+    idx_Rs = transforms["R"].basis.get_idx(N=-1, error=False)
+    idx_Z = transforms["Z"].basis.get_idx(N=0, error=False)
+    X0 = params["R_n"][idx_Rc] / 2 if isinstance(idx_Rc, int) else 0
+    Y0 = params["R_n"][idx_Rs] / 2 if isinstance(idx_Rs, int) else 0
+    Z0 = params["Z_n"][idx_Z] if isinstance(idx_Z, int) else 0
+    center = jnp.array([X0, Y0, Z0])
```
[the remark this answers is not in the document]
ddudt: This is already done [here](https://github.com/PlasmaControl/DESC/blob/dd/curvature/tests/test_curves.py#L21)

desc/compute/_curve.py
```diff
@@ -366,20 +423,19 @@ def _x_FourierRZCurve(params, transforms, profiles, data, **kwargs):
...
-    dphi = R0
-    coords = jnp.stack([dR, dphi, dZ], axis=1)
-    # convert to xyz for displacement and rotation
+    coords = jnp.stack([dR, R0, dZ], axis=1)
+    # convert to xyz for rotation using phi=s
     coords = rpz2xyz_vec(coords, phi=transforms["grid"].nodes[:, 2])
     coords = coords @ params["rotmat"].reshape((3, 3)).T
-    # convert back to rpz
-    coords = xyz2rpz_vec(coords, phi=transforms["grid"].nodes[:, 2])
+    # convert back to rpz using real phi to account for displacement
```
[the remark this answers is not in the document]
ddudt: We use the curve parameter `phi=s` to convert from rpz -> xyz because the derivatives `coords = [dR, R0, dZ]` assumes that the curve is in the standard orientation centered on the origin. When we convert back xyz -> rpz we need to account for possible rotation/translation from the `shift` and `rotmat` params. So here we use `phi = data["phi"]` which is the true toroidal angle phi.

desc/compute/_curve.py
```diff
@@ -395,24 +451,20 @@ def _x_s_FourierRZCurve(params, transforms, profiles, data, **kwargs):
...
     profiles=[],
     coordinates="s",
-    data=[],
+    data=["phi"],
     parameterization="desc.geometry.curve.FourierRZCurve",
 )
 def _x_ss_FourierRZCurve(params, transforms, profiles, data, **kwargs):
     R0 = transforms["R"].transform(params["R_n"], dz=0)
-    dR = transforms["R"].transform(params["R_n"], dz=1)
+    d1R = transforms["R"].transform(params["R_n"], dz=1)
```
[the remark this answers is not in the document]
ddudt: This line was just a formatting change to make the variable names consistent with the other functions. But this compute quantity `"x_ss"` is implicitly covered by [test_curvature](https://github.com/PlasmaControl/DESC/blob/dd/curvature/tests/test_curves.py#L49) and `test_compute_everything`

desc/objectives/_generic.py
ddudt: This is technically unrelated to the signed curvature, but I threw it in this PR (sorry). This change is necessary to allow `LinearObjectiveFromUser` to work with an `OptimizableCollection` like a `CoilSet`.

tests/test_curves.py
ddudt: That won't increase coverage because everything is already computed in rpz basis now by default, so computing in xyz effectively covers both cases. In this test it was easier for me to specify the "correct" center in xyz coordinates. FYI patch coverage is already at 100%.

tests/test_curves.py
ddudt: That is already done in this test [a few lines below](https://github.com/PlasmaControl/DESC/blob/dd/curvature/tests/test_curves.py#L509). 

You may be confused because the kwarg is called `basis` in the `FourierPlanarCurve` class, but we call it `basis_in` in the compute functions to differentiate it from the other "basis" kwarg (the latter is really "basis_out").

tests/test_curves.py
dpanici: didn't realize, I thought a few spots said it was not covered

desc/objectives/_generic.py
dpanici: is this tested anywhere in this PR?

desc/coils.py
```diff
@@ -1665,7 +1670,7 @@ def __init__(self, *coils, name="", check_intersection=True):
     @property
     def num_coils(self):
         """int: Number of coils."""
-        return sum([c.num_coils if hasattr(c, "num_coils") else 1 for c in self])
+        return sum([c.num_coils for c in self])
```
kianorr, line 1673: Does this work for a nested coilset?

dpanici: I think it should, because num_coils will just get called recursively until the self is a single coilset

desc/objectives/_generic.py
ddudt: The new functionality is not tested in this PR, no. But there are existing tests for `LinearObjectiveFromUser` that still pass.

dpanici (changes requested):
If there was a change for linear objective from user to have it work with optimizable collections, I'd like to see it tested if possible. If not in this PR then make an issue to test it in the future

ddudt:
> If there was a change for linear objective from user to have it work with optimizable collections, I'd like to see it tested if possible. If not in this PR then make an issue to test it in the future

I just added a test for it in #1130, which will now fail until this PR is merged.

dpanici (approved):

desc/objectives/_generic.py
ddudt: see #1130 for a test of this new functionality

kianorr (approved):

f0uriest (approved):

PR #1086, opened 2024-06-28
Use max rho of surface to compute A(z)

### LINKED ISSUE #1083: A(z) computed wrong on grids without lcfs
Take the surface in `test_compute_everything` and add rho=0.8 to the constructor. `FourierRZToroidalSurface(**elliptic_cross_section_with_torsion, rho=0.8)`. Then the grid only has rho =0.8 nodes and A(z) is computed as zero. I think the fix is to compute this like we do `perimeter(z)`; fix_surface to the rho value of the surface and then divide by that rho value to get an approximation, but I'll take a closer look tomorrow.

_Originally posted by @unalmis in https://github.com/PlasmaControl/DESC/pull/1024#discussion_r1656635137_

unalmis:
Resolves #1083

### DIFF
--- a/desc/compute/_basis_vectors.py
+++ b/desc/compute/_basis_vectors.py
@@ -3523,3 +3523,25 @@ def _n_zeta(params, transforms, profiles, data, **kwargs):
         ).T,
     )
     return data
+
+
+@register_compute_fun(
+    name="e_theta|r,p",
+    label="\\mathbf{e}_{\\theta} |_{\\rho, \\phi}",
+    units="m",
+    units_long="meters",
+    description="Tangent vector along boundary of constant phi toroidal surface",
+    dim=3,
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["e_theta", "e_phi", "phi_t"],
+    parameterization=[
+        "desc.equilibrium.equilibrium.Equilibrium",
+        "desc.geometry.surface.FourierRZToroidalSurface",
+    ],
+)
+def _e_sub_theta_rp(params, transforms, profiles, data, **kwargs):
+    data["e_theta|r,p"] = data["e_theta"] - (data["e_phi"].T * data["phi_t"]).T
+    return data
--- a/desc/compute/_basis_vectors.py
+++ b/desc/compute/_basis_vectors.py
@@ -3532,7 +3532,7 @@ def _n_zeta(params, transforms, profiles, data, **kwargs):
     units="m",
     units_long="meters",
     description=(
-        "Covariant poloidal basis vector in (R,ϕ,Z) coordinates. "
+        "Covariant poloidal basis vector in (ρ,θ,ϕ) coordinates. "
         "ϕ increases counterclockwise when viewed from above "
         "(cylindrical R,ϕ plane with Z out of page)."
     ),
--- a/desc/compute/_basis_vectors.py
+++ b/desc/compute/_basis_vectors.py
@@ -1514,6 +1514,7 @@ def _e_sup_zeta_zz(params, transforms, profiles, data, **kwargs):
     parameterization=[
         "desc.equilibrium.equilibrium.Equilibrium",
         "desc.geometry.surface.FourierRZToroidalSurface",
+        "desc.geometry.core.Surface",
     ],
     basis="basis",
 )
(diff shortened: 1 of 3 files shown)


desc/compute/_geometry.py
```diff
@@ -196,7 +196,7 @@ def _A_of_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwarg
             transforms["grid"],
             data["Z"] * n[:, 2] * jnp.sqrt(data["g_tt"]),
             line_label="theta",
-            fix_surface=("rho", 1.0),
```
f0uriest, line 199: do we also want to scale the area up if the largest rho isnt 1?

unalmis: I don't think that makes sense. It's proportional sure but the integrand will be different, so it's not exact.

tests/inputs/master_compute_data.pkl
unalmis, line 1: Changed from adding e_theta|r,p. A(z) is same for rho=1 surface.

desc/compute/_geometry.py
unalmis: Was our justification to do that for `perimeter(z)` because we assume integrand doesn't change much near edge of quadrature node and the lcfs?

desc/compute/_geometry.py
```diff
@@ -176,27 +176,38 @@ def _A_of_z(params, transforms, profiles, data, **kwargs):
...
+    data=["Z", "n_rho", "e_theta|r,p"],
     parameterization=["desc.geometry.surface.FourierRZToroidalSurface"],
+    # FIXME: Add source grid requirement once omega is nonzero.
 )
 def _A_of_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwargs):
-    # divergence theorem: integral(dA div [0, 0, Z]) = integral(ds n dot [0, 0, Z])
-    # but we need only the part of n in the R,Z plane
+    # Denote any vector v = [vᴿ, v^ϕ, vᶻ] with a tuple of its contravariant components.
+    # We use a 2D divergence theorem over constant ϕ toroidal surface (i.e. R, Z plane).
+    # Denote the divergence operator div by [∂_R, ∂_ϕ, ∂_Z] ⊗ [1, 0, 1].
```
unalmis, line 193: in the RZ plane divergence operator is cartesian

desc/compute/_geometry.py
```diff
@@ -176,27 +176,39 @@ def _A_of_z(params, transforms, profiles, data, **kwargs):
...
-    description="Cross-sectional area as function of zeta",
+    description="Enclosed cross-sectional (constant phi surface) area, "
+    "as function of zeta",
     dim=1,
     params=[],
     transforms={"grid": []},
     profiles=[],
     coordinates="z",
-    data=["Z", "n_rho", "g_tt"],
+    data=["Z", "n_rho", "e_theta|r,p", "rho"],
```
dpanici, line 186: why this change? is this for when omega is nonzero?

desc/compute/_basis_vectors.py
```diff
@@ -3523,3 +3523,25 @@ def _n_zeta(params, transforms, profiles, data, **kwargs):
...
+    profiles=[],
+    coordinates="rtz",
+    data=["e_theta", "e_phi", "phi_t"],
+    parameterization=[
+        "desc.equilibrium.equilibrium.Equilibrium",
+        "desc.geometry.surface.FourierRZToroidalSurface",
+    ],
+)
+def _e_sub_theta_rp(params, transforms, profiles, data, **kwargs):
+    data["e_theta|r,p"] = data["e_theta"] - (data["e_phi"].T * data["phi_t"]).T
```
dpanici, line 3546: `e_theta` by definition a partial derivative at constant rho and phi right? I do not quite get this expression

unalmis: No, it's constant rho and zeta

desc/compute/_geometry.py
unalmis: Yes, I was checking the math for this to be sure, and i was gonna need to do this anyway so I did it here

dpanici (approved):

desc/compute/_basis_vectors.py
```diff
@@ -3523,3 +3523,25 @@ def _n_zeta(params, transforms, profiles, data, **kwargs):
...
     )
     return data
+
+
+@register_compute_fun(
+    name="e_theta|r,p",
+    label="\\mathbf{e}_{\\theta} |_{\\rho, \\phi}",
+    units="m",
+    units_long="meters",
+    description="Tangent vector along boundary of constant phi toroidal surface",
```
f0uriest, line 3533: Is this the most accurate description?

desc/compute/_geometry.py
f0uriest: Yeah pretty much. Obviously if you're using the rho=0.5 surface it's a bad guess, but for rho near 1 it should be reasonably accurate (and exact for elliptical cross section)

desc/compute/_basis_vectors.py
unalmis: It is precise. i'll add the standard "covariant vector ..."

f0uriest (approved):

dpanici (approved):

PR #1089, opened 2024-07-01
Fix bug where basis vectors are converted to xyz erroneously assuming input is in rpz

### LINKED ISSUE #1088: Quantities are computed wrong unless returned data is in cylindrical basis vectors

unalmis:
Resolves #1088 .

We only want to do a basis conversion on quantities that directly compute a vector's contravariant components in the cylindrical coordinate system. Otherwise we'll be converting basis vectors from cylindrical to Cartesian multiple times because the dependencies are already in cartesian. The issue is that after the first conversion, we're in cartesian coordinates, yet the following conversions still assume the input is in polar basis.

### DIFF
--- a/desc/compute/_curve.py
+++ b/desc/compute/_curve.py
@@ -2,6 +2,7 @@ from interpax import interp1d
 
 from desc.backend import jnp
 
+from ..utils import errorif
 from .data_index import register_compute_fun
 from .geom_utils import rotation_matrix, rpz2xyz, rpz2xyz_vec, xyz2rpz, xyz2rpz_vec
 from .utils import cross, dot, safenormalize
@@ -156,6 +157,7 @@ def _phi_Curve(params, transforms, profiles, data, **kwargs):
     parameterization="desc.geometry.core.Curve",
 )
 def _Z_Curve(params, transforms, profiles, data, **kwargs):
+    errorif(kwargs.get("basis", "rpz").lower()[-1] != "z", NotImplementedError)
     data["Z"] = data["x"][:, 2]
     return data
 
--- a/desc/compute/_curve.py
+++ b/desc/compute/_curve.py
@@ -157,7 +157,9 @@ def _phi_Curve(params, transforms, profiles, data, **kwargs):
     parameterization="desc.geometry.core.Curve",
 )
 def _Z_Curve(params, transforms, profiles, data, **kwargs):
-    errorif(kwargs.get("basis", "rpz").lower()[-1] != "z", NotImplementedError)
+    errorif(
+        kwargs.get("basis", "rpz").lower() not in {"rpz", "xyz"}, NotImplementedError
+    )
     data["Z"] = data["x"][:, 2]
     return data
 
--- a/desc/compute/_equil.py
+++ b/desc/compute/_equil.py
@@ -270,7 +270,9 @@ def _J_phi(params, transforms, profiles, data, **kwargs):
     data=["J"],
 )
 def _J_Z(params, transforms, profiles, data, **kwargs):
-    errorif(kwargs.get("basis", "rpz").lower()[-1] != "z", NotImplementedError)
+    errorif(
+        kwargs.get("basis", "rpz").lower() not in {"rpz", "xyz"}, NotImplementedError
+    )
     data["J_Z"] = data["J"][:, 2]
     return data
 
--- a/desc/compute/_equil.py
+++ b/desc/compute/_equil.py
@@ -13,6 +13,7 @@ from scipy.constants import mu_0
 
 from desc.backend import jnp
 
+from ..utils import errorif
 from .data_index import register_compute_fun
 from .utils import cross, dot, safediv, safenorm, surface_averages
 
@@ -231,6 +232,7 @@ def _J_sqrt_g_r(params, transforms, profiles, data, **kwargs):
     data=["J"],
 )
 def _J_R(params, transforms, profiles, data, **kwargs):
+    errorif(kwargs.get("basis", "rpz").lower() != "rpz", NotImplementedError)
     data["J_R"] = data["J"][:, 0]
     return data
 
@@ -249,6 +251,7 @@ def _J_R(params, transforms, profiles, data, **kwargs):
     data=["J"],
 )
 def _J_phi(params, transforms, profiles, data, **kwargs):
+    errorif(kwargs.get("basis", "rpz").lower() != "rpz", NotImplementedError)
     data["J_phi"] = data["J"][:, 1]
     return data
 
@@ -267,6 +270,7 @@ def _J_phi(params, transforms, profiles, data, **kwargs):
     data=["J"],
 )
 def _J_Z(params, transforms, profiles, data, **kwargs):
+    errorif(kwargs.get("basis", "rpz").lower()[-1] != "z", NotImplementedError)
     data["J_Z"] = data["J"][:, 2]
     return data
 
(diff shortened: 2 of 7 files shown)


unalmis:
So I fixed the bugs here, and added guards to some compute funs that pull request #1027 should add logic to not trip.

desc/compute/_equil.py
```diff
@@ -677,6 +681,7 @@ def _W_B(params, transforms, profiles, data, **kwargs):
     data=["B", "sqrt(g)"],
 )
 def _W_Bpol(params, transforms, profiles, data, **kwargs):
+    errorif(kwargs.get("basis", "rpz").lower() != "rpz", NotImplementedError)
```
dpanici, line 684: why does this matter here? is it because some of the intermediate quantities get messed up when basis is xyz, so B itself is wrong at this point?

dpanici (approved):

ddudt:
Is this PR redundant with #1027? If so I would prefer to close this and finish that one instead (it should be nearly ready to merge).

desc/compute/_equil.py
unalmis: This compute function assumes the basis of its dependencies are in cylindrical coordinates.

desc/compute/_curve.py
```diff
@@ -156,6 +157,7 @@ def _phi_Curve(params, transforms, profiles, data, **kwargs):
     parameterization="desc.geometry.core.Curve",
 )
 def _Z_Curve(params, transforms, profiles, data, **kwargs):
+    errorif(kwargs.get("basis", "rpz").lower()[-1] != "z", NotImplementedError)
```
ddudt, line 160: It would be safer to check for `"rpz"` instead of just `"z"`, in case we add more bases in the future that might use z to mean something different (e.g., `"rtz"` the z means zeta instead of Z). Same for other files where you have this check for only z

tests/test_compute_funs.py
```diff
@@ -1683,3 +1683,17 @@ def test_surface_equilibrium_geometry():
...
                 atol=1e-13,
             )
+
+
+@pytest.mark.unit
+def test_basis_kwarg():
+    """Test that we don't change basis to xyz erroneously assuming input is in rpz."""
+    eq = get("W7-X")
+    names = ["b", "e^rho", "e_theta_PEST", "phi"]
+    data_rpz = eq.compute(names)
```
ddudt, line 1693: Just to be extra safe and future-proof, let's explicitly input `basis="rpz"` here instead of assuming the default.

ddudt (changes requested):

ddudt:
> > Is this PR redundant with #1027? If so I would prefer to close this and finish that one instead (it should be nearly ready to merge).
> 
> The bug is currently "worse" on that PR. Just having the basis as a keyword argument when it's not needed will give the wrong result there. There's benefit to fixing this now in a pr that doesn't add new features

OK I'm fine with this fix since it won't hurt. But I still think the other PR also resolves this issue because it removes the `basis` kwarg from being used in the compute functions and only does the coordinate transform once at the end. But maybe I'm not understanding the issue properly?

unalmis:
If we don't want to merge into master, it may be worthwile to merge into 1027. I'll leave up to y'all

desc/compute/_equil.py
```diff
@@ -231,6 +232,7 @@ def _J_sqrt_g_r(params, transforms, profiles, data, **kwargs):
     data=["J"],
 )
 def _J_R(params, transforms, profiles, data, **kwargs):
+    errorif(kwargs.get("basis", "rpz").lower() != "rpz", NotImplementedError)
```
f0uriest, line 235: Instead of erroring in all these cases, couldn't we convert J (or whatever else) to rpz locally then get the component?

f0uriest (changes requested):

desc/compute/_surface.py
```diff
@@ -217,6 +221,9 @@ def _phi_z_Surface(params, transforms, profiles, data, **kwargs):
     parameterization="desc.geometry.core.Surface",
 )
 def _Z_Surface(params, transforms, profiles, data, **kwargs):
+    errorif(
+        kwargs.get("basis", "rpz").lower() not in {"rpz", "xyz"}, NotImplementedError
```
YigitElma, line 225: We can instead just add this line with PR #1027 to the main `compute` function. This shouldn't be added to every single compute function one by one.

desc/compute/_field.py
```diff
@@ -140,6 +141,7 @@ def _B(params, transforms, profiles, data, **kwargs):
     data=["B"],
 )
 def _B_R(params, transforms, profiles, data, **kwargs):
+    errorif(kwargs.get("basis", "rpz").lower() != "rpz", NotImplementedError)
```
YigitElma, line 144: With #1027, `B` will be in `rpz` by default during compute functions, so as long as every compute function assumes `rpz` it is consistent. Then, 3D quantities will converted "after" actual calculation. What you did would be relevant if we had `B_X` and it was defined as `data["B"][:, 0]` without any prior conversion.

desc/compute/_equil.py
YigitElma: Again, #1027 solves this.

YigitElma (commented):
I couldn't see any fix that is not covered by #1027. Maybe we can add the `"xyz,"rpz"` check for `basis` (but that is trivial). If people can review #1027, we can cover the missing points there and close this one.

unalmis:
Yes from the dev meeting I remembered that #1027 did the basis change recursively, which would cause the issue to be worse and so it seemed that PR was unaware of issue #1088. Since the logic there was changed to do the basis change at the end as Daniel mentioned, I'll leave up to y'all whether to close or merge there etc. Maybe link #1088 there as well then and
> Just make sure things like cartesian X which should be in xyz natively, even without passing in the additonal basis kwarg, aren't mistakenly converted to xyz again

YigitElma:
> Yes from the dev meeting I remembered that #1027 did the basis change recursively, which would cause the issue to be worse and so it seemed that PR was unaware of issue #1088. Since the logic there was changed to do the basis change at the end as Daniel mentioned, I'll leave up to y'all whether to close or merge there etc. Maybe link #1088 there as well then and
> 
> > Just make sure things like cartesian X which should be in xyz natively, even without passing in the additonal basis kwarg, aren't mistakenly converted to xyz again

Yes, the first version of #1027  was like that but I changed the basis conversion to be at the end (just as you mentioned, it was really hard to keep track of which quantity was in which basis). And yes, if the compute quantity has the basis included in its name like `X`, `Y`, etc. then that quantity is returned in that basis (only in those cases basis conversion is done locally)

desc/compute/_equil.py
unalmis: No because the basis kwarg is not sufficient to specify that "J" is in xyz or rpz.
If user does `eq.compte(["J", "J_R"], basis="xyz")`,
then after the changes in this PR, (master would compute wrong), we would know that "J" is in "xyz" and we could do what you say.

If user passes in precomputed data  `eq.compte(["J_R"], basis="xyz", data={"J": J})` then we don't know whether the dependencies are in xyz or rpz.

ddudt: If the user is passing in data we are always going to have basis consistency issues, and I don't think there is an easy way to avoid that. We just have to trust the users know what they are doing.

dpanici:
Close one #1027 is closed and confirm that it fixes everything

f0uriest:
Closing in favor of #1027

tests/test_compute_funs.py
```diff
@@ -1683,3 +1683,17 @@ def test_surface_equilibrium_geometry():
...
+    """Test that we don't change basis to xyz erroneously assuming input is in rpz."""
+    eq = get("W7-X")
+    names = ["b", "e^rho", "e_theta_PEST", "phi"]
+    data_rpz = eq.compute(names, basis="rpz")
+    data_xyz = eq.compute(names, basis="xyz")
+    for name in names:
+        if name != "phi":
+            np.testing.assert_allclose(
+                rpz2xyz_vec(data_rpz[name], phi=data_rpz["phi"]), data_xyz[name]
+            )
```
unalmis, line 1699: comment to link

PR #1090, opened 2024-07-01
PEST coordinate system basis vectors

unalmis:
Generalizes theta_PEST basis vectors for toroidal stream function and changes e_phi docstring since it is somewhat misleading.

 While debugging a test that only assumes nested flux surfaces, I checked the computation of the pest sfl Jacobian which looked wrong. We were computing
```python
# e_rho fixed theta and zeta, e_vartheta fixed rho and zeta, e_phi at fixed rho and theta
dot(data["e_rho|t,z"], cross(data["e_vartheta|r,z"], data["e_phi|r,t"]))
```
when we should have been computing
```python
# e_rho fixed vartheta and phi, e_vartheta at fixed rho and phi, e_phi at fixed rho and vartheta
dot(data["e_rho|v,p"], cross(data["e_vartheta|r,p"], data["e_phi|r,v"]))
```
Assuming $\zeta = \phi$, only the second of these vectors are the same between the two compuations; the first and third are of course not. However, the triple scalar product suppresses the differences so that the final result is equal when $\zeta = \phi$.


desc/compute/_basis_vectors.py
```diff
@@ -1385,11 +1385,11 @@ def _e_sup_zeta_zz(params, transforms, profiles, data, **kwargs):
...
 
 @register_compute_fun(
-    name="e_phi",
-    label="\\mathbf{e}_{\\phi}",
+    name="e_phi|r,t",
+    label="\\mathbf{e}_{\\phi} |_{\\rho, \\theta}",
     units="m",
     units_long="meters",
-    description="Covariant cylindrical toroidal basis vector",
+    description="Covariant toroidal basis vector in (ρ,θ,Φ) coordinates",
```
dpanici, line 1392: phi should not be capitalized here

dpanici, line 1392: can you maybe make it clear in the description that it is the partial derivative of the position vector evaluated at constant rho, theta?

desc/compute/_basis_vectors.py
```diff
@@ -1400,10 +1400,39 @@ def _e_sup_zeta_zz(params, transforms, profiles, data, **kwargs):
...
+    data["e_phi|r,t"] = (data["e_zeta"].T / data["phi_z"]).T
+    return data
+
+
+@register_compute_fun(
+    name="e_phi|R,Z",
+    label="\\mathbf{e}_{\\phi} |_{R, Z}",
+    units="m",
+    units_long="meters",
+    description="Tangent vector along increasing cylindrical toroidal angle phi",
```
dpanici, line 1419: can you maybe make it clear in the description that it is the partial derivative of the position vector evaluated at constant R, Z?

desc/compute/_basis_vectors.py
```diff
@@ -1400,10 +1400,39 @@ def _e_sup_zeta_zz(params, transforms, profiles, data, **kwargs):
         "desc.equilibrium.equilibrium.Equilibrium",
         "desc.geometry.surface.FourierRZToroidalSurface",
     ],
+    aliases=["e_phi"],
+    # Our usual notation implies e_phi = (∂X/∂Φ)|R,Z, but we need to alias e_phi to
```
f0uriest, line 1404: I thought our usual assumption would be rho, theta held constant? Where do we need derivatives with R,Z constant?

tests/test_data_index.py
```diff
@@ -39,6 +40,11 @@ def get_parameterization(fun, default="desc.equilibrium.equilibrium.Equilibrium"
         matches.discard("")
         return matches if matches else {default}
 
+    @staticmethod
+    def _is_function(func):
+        # JITed functions are not functions according to inspect.
+        return inspect.isfunction(func) or callable(func)
```
f0uriest, line 46: All functions are callable so I don't think you need the first part?

desc/compute/_basis_vectors.py
unalmis: It's more like we hold the variables in the natural coordinate system constant.

For example, `e_theta` is at fixed rho and zeta because the coordinate system is (rho, theta, zeta).
Likewise, `e_theta_PEST = e_vartheta` is at fixed rho and phi because the coordinate system is (rho, vartheta, phi).
I have followed this convention in the PRs that add quantities derivatives along field lines such as
`e_alpha` which is at fixed rho and zeta because the coordinate system is (rho, alpha, zeta).

Same for `B_z` is at fixed rho, theta, but `B_a` is at fixed rho, zeta.

unalmis: This is moreso to clarify the inconsistency that `e_phi` is not `e_phi|R,Z`. One could have also used this when computing `A(z)`.

dpanici (dismissed):
remove e_phi|R,Z

desc/compute/_basis_vectors.py
```diff
@@ -1400,10 +1400,13 @@ def _e_sup_zeta_zz(params, transforms, profiles, data, **kwargs):
         "desc.equilibrium.equilibrium.Equilibrium",
         "desc.geometry.surface.FourierRZToroidalSurface",
     ],
+    aliases=["e_phi"],
+    # Our usual notation implies e_phi = (∂X/∂ϕ)|R,Z = R ϕ̂, but we need to alias e_phi
```
f0uriest, line 1404: nit: usual notation would be at const rho, theta

f0uriest: Wait NVM, I get it now

f0uriest (dismissed):

unalmis:
Just fixed merge conflicts since last review

dpanici:
Add notation on basis to list of variables

f0uriest (approved):

desc/compute/_basis_vectors.py
```diff
@@ -1407,11 +1400,13 @@ def _e_sup_zeta_zz(params, transforms, profiles, data, **kwargs):
         "desc.geometry.surface.FourierRZToroidalSurface",
         "desc.geometry.core.Surface",
     ],
+    aliases=["e_phi"],
+    # Our usual notation implies e_phi = (∂X/∂ϕ)|R,Z = R ϕ̂, but we need to alias e_phi
+    # to e_phi|r,t = (∂X/∂ϕ)|ρ,θ for compatibility with older versions of the code.
```
ddudt, line 1405: Do we really care about backwards compatibility here? I doubt a lot of users have code that depends on computing `"e_phi"`

ddudt (approved):

desc/compute/_basis_vectors.py
unalmis: We could deprecate it later in that case

PR #1092, opened 2024-07-02
Add jitable to surface compute to fix JAX error

dpanici:
- Adds `jitable` as a kwarg to `get_transforms` in `Surface.compute` to avoid errors when jitting the `.compute_magnetic_field` method of surface current objects


dpanici:
Putting change in #1002 instead

PR #807, opened 2023-12-13
plot elongation

ddudt:
Resolves #804.  

Does the same grid resampling that we do for functions of $\rho$, but now also for $\zeta$. 

The `grid_type` data index tag is no longer used, but I keep it because we should start using this in the future. This PR resolves an issue about plotting functions of $\zeta$ vs $\rho$, but that tag will still be useful for determining the proper type of grid to use (e.g. `"quad"` vs `"linear"`). 

I added a new test and also changed the other 1D plotting tests to get examples so they don't depend on doing an equilibrium solve.

PR #1094, opened 2024-07-02
Integrate on boundary for more quantities. Enforce use of full grid when required.

### LINKED ISSUE #1101: Incorrect computations with stellarator symmetry
We have an inconsistent use of stellarator symmetry (the discrete one) that leads to computing quantities incorrectly.
Here is a toroid surface with an elliptic cross section and torsion.
```python
surf = FourierRZToroidalSurface(
    R_lmn=[10, 1, 0.2],
    Z_lmn=[-2, -0.2],
    modes_R=[[0, 0], [1, 0], [0, 1]],
    modes_Z=[[-1, 0], [0, -1]],
)
```
DESC determines this is stellarator symmetric, and will also, by default, label equilibrium with such a boundary as a symmetric equilibrium. We use the same characteristic to construct computational grids for stellarator symmetry that truncate the poloidal domain from $\theta \in [0, 2 \pi]$ to $\theta \in [0, \pi]$. By doing so, we are assuming that, under stellarator symmetry, the following relation holds $\int_0^{2\pi} f(\theta) d \theta = 2 \int_0^{\pi} f(\theta) d \theta$, which is not true because f is not symmetric.

For example, this [quantity](https://github.com/PlasmaControl/DESC/blob/0a6b9957f652637142a66de922396d7312e55a16/desc/compute/_geometry.py#L190)  is not computed correctly on symmetric objects such as the surface above when integrated along a boundary, so the `Elongation` and `Aspect ratio` objectives that target toroid surfaces introduced in #884 have never been correct on `master` for symmetric surfaces (the [equilibrium ones](https://github.com/PlasmaControl/DESC/blob/0a6b9957f652637142a66de922396d7312e55a16/tests/test_compute_funs.py#L176) are fine edit: fine for this surface).

One may expect that the same issue exists in the volume computation, and hence the magnetic well, but those also integrate over zeta, and even though their integrands, assuming symmetry, do not satisfy $\int_0^{2\pi} f(\theta) d \theta = 2 \int_0^{\pi} f(\theta) d \theta$, I think they do satisfy $\int_0^{2\pi} \int_0^{2\pi} f(\theta) d \theta d\zeta = 2 \int_0^{2\pi} \int_0^{\pi} f(\theta) d \theta d\zeta$ (one can show analytically they do for above surface), which is why all the other tests for correctness of this work. Still we should review code and review the symmetry condition.

(fyi these results were confirmed/tested with analytic tests using `sympy`).

unalmis:
- [x] Some quantities like the plasma volume `V`, major radius `R0`, minor radius `a` etc. can be computed via stoke's theorem on the last closed flux surface. This pull request modifies the compute functions to default to such an integration on the boundary rather than throughout the volume so that the length scale quantities can be computed accurately on grids other than `QuadratureGrid`. This lets us avoid building transforms for `QuadratureGrid` in objectives, saving memory. Also, has some improvements in general

> 1. Reducing dimensionality of integral makes integration more tractable with faster convergence.
> 2. Transforming the integration to one over periodic domains makes it more accurate due to the exponential convergence of the quadrature on periodic domains.

- [x] Merging this PR completes my list of reviewing tasks to resolve #1101


dpanici:
Add error to line_integrals if grid is sym and coord is theta

dpanici:
change defualt grid in AspectRatio and Elongation to sym=False if thing=FourierRZToroidalSurface

desc/compute/_geometry.py
```diff
@@ -232,23 +211,46 @@ def _A_of_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwarg
...
+    resolution_requirement="tz",
 )
 def _A(params, transforms, profiles, data, **kwargs):
-    data["A"] = jnp.mean(
-        transforms["grid"].compress(data["A(z)"], surface_label="zeta")
+    # Denote any vector v = [vᴿ, v^ϕ, vᶻ] with a tuple of its contravariant components.
+    # We use a 2D divergence theorem over constant ϕ toroidal surface (i.e. R, Z plane).
+    # In this geometry, the divergence operator on a polar basis vector is
+    # div = ([∂_R, ∂_ϕ, ∂_Z] ⊗ [1, 0, 1]) dot .
+    # ∫ dA div v = ∫ dℓ n dot v
```
unalmis, line 232: On master we just take a uniform mean of `A(z)` to compute `A`. That seems wrong if the flux surface area between zeta points is not uniform. I think to compute `A` like this we would want a vector v such that `div v = ||e_ϕ|ρ,θ||` instead of `div v = 1` and then replace `jnp.ptp(phi)` with `jnp.dot(data["e_phi|r,t"], transforms["grid"].weights)`. Outside scope of this PR.

desc/compute/_geometry.py
```diff
@@ -155,45 +158,20 @@ def _V_rrr_of_r(params, transforms, profiles, data, **kwargs):
...
+    data=["Z", "n_rho", "e_theta|r,p", "rho"],
     parameterization=[
         "desc.equilibrium.equilibrium.Equilibrium",
         "desc.geometry.surface.ZernikeRZToroidalSection",
+        "desc.geometry.surface.FourierRZToroidalSurface",
     ],
-    resolution_requirement="rt",
-)
-def _A_of_z(params, transforms, profiles, data, **kwargs):
-    data["A(z)"] = surface_integrals(
```
unalmis, line 172: When the grid is truncated to θ ∈ [0, π], this surface integral give more accurate results than the line integral method, sometimes significantly. It will also give the correct result for a broader range of surfaces, for example `test_elongation` on master passes using the surface integral but not with a line integral. However, I think it's still wrong in general when truncated θ ∈ [0, π] just like the line integral method, so I removed it in favor of the line integral for the purpose of this PR's title.

unalmis: So, as documented with the numerical differences in commit [be11015](https://github.com/PlasmaControl/DESC/pull/1094/commits/be1101524a11f49d293b1661c44b2b67a5efd747), the new master compute data for 'A(z)' in this PR is less correct than on `master`.

desc/compute/_geometry.py
```diff
@@ -232,23 +212,48 @@ def _A_of_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwarg
     label="A",
     units="m^{2}",
     units_long="square meters",
-    description="Average cross-sectional area",
+    description="Average enclosed cross-sectional area, scaled by max(ρ)⁻²",
```
unalmis, line 215: This is a simple toroidal average in the sense of $A_0 = \int_{\zeta} A(\zeta) d\zeta / (2 \pi)$ matching the convention that `R0` is the average in the sense  $R_0 = \int_{\zeta} R(\rho=0) d\zeta / (2 \pi)$. This is different than the average value over the device. That would be $\langle R \rangle =  \int_{\zeta} R \Vert e_{\zeta}\Vert(\rho=0) d \zeta / \int_{\zeta}  \Vert e_{\zeta}\Vert(\rho=0) d \zeta$.

desc/plotting.py
unalmis, line 1: These changes were made to pass the elongation test after relaxing the dependencies of a_major/a_minor to `resolution_requirement=t` from `rt` since we only need rho near the surface.

desc/plotting.py
```diff
@@ -629,8 +628,7 @@ def plot_2d(
         grid_kwargs = {"M": 33, "N": 33, "NFP": eq.NFP, "axis": False}
         grid = _get_grid(**grid_kwargs)
     plot_axes = _get_plot_axes(grid)
-    if len(plot_axes) != 2:
-        return ValueError(colored("Grid must be 2D", "red"))
```
unalmis, line 633: (should be raising not returning error)

tests/baseline/test_1d_elongation.png
unalmis, line 1: There are small numerical differences from integrating on boundary. I concluded that's why this needed regeneration.

tests/test_plotting.py
```diff
@@ -90,7 +90,7 @@ def test_1d_p(self):
     def test_1d_elongation(self):
         """Test plotting 1d elongation as a function of toroidal angle."""
         eq = get("precise_QA")
-        grid = LinearGrid(N=20, NFP=eq.NFP)
+        grid = LinearGrid(M=eq.M_grid, N=20, NFP=eq.NFP)
```
unalmis, line 93: Until #1138 is resolved, users will not be warned that poloidal resolution is required to plot this, giving junk like
result

unalmis: This gave correct plot prior to this PR because the `resolution_requirement=rt` instead of  `resolution_requirement=t` and `eq.compute` would automatically construct a lineargrid with radial and poloidal resolution to compute and then copy data back to 1d grid to plot. Now that it's just  `resolution_requirement=t` we don't make a new grid automatically. I don't think we need to add logic to that automatically since `eq.compute` will raise a warning .

desc/equilibrium/equilibrium.py
```diff
@@ -973,7 +973,6 @@ def need_src(name):
             # Warn if best way to compute accurately is increasing resolution.
```
unalmis, line 973: Adjusted logic to be more self-consistent. Will raise all warnings raised previously and covers some additional cases.

desc/compute/utils.py
```diff
@@ -685,7 +685,10 @@ def _has_transforms(qty, transforms, parameterization):
 
 
 def dot(a, b, axis=-1):
-    """Batched vector dot product.
+    """Batched coordinate dot product.
+
+    This returns the dot product between elements of a vector space only
+    if the basis vectors associated with these coordinates are orthonormal.
```
unalmis, line 691: people asked in other pr, so i documented

desc/compute/_geometry.py
```diff
@@ -440,9 +450,10 @@ def _R0_over_a(params, transforms, profiles, data, **kwargs):
         "desc.equilibrium.equilibrium.Equilibrium",
         "desc.geometry.core.Surface",
     ],
-    resolution_requirement="rt",  # just need max(rho) near 1
+    resolution_requirement="t",
```
unalmis, line 453: otherwise, we get unnecessary warnings when using a grid with rho=1 surface

desc/compute/_geometry.py
```diff
@@ -39,27 +46,19 @@ def _V(params, transforms, profiles, data, **kwargs):
     label="V",
     units="m^{3}",
     units_long="cubic meters",
-    description="Volume",
+    description="Volume enclosed by surface, scaled by max(ρ)⁻²",
```
unalmis, line 49: Do we want this?

desc/compute/_geometry.py
```diff
@@ -275,25 +282,22 @@ def _A_of_r(params, transforms, profiles, data, **kwargs):
     label="S",
     units="m^{2}",
     units_long="square meters",
-    description="Surface area of outermost flux surface",
+    description="Surface area of outermost flux surface, scaled by max(ρ)⁻¹",
```
unalmis, line 285: Do we want this?

tests/test_plotting.py
f0uriest: I would prefer to add logic to `eq.compute` for that case so that it "just works", especially for plotting stuff. It especially feels weird if you need to create a 2d grid to make a 1d plot.

tests/test_surfaces.py
```diff
@@ -405,6 +405,7 @@ class TestZernikeRZToroidalSection:
     """Tests for ZernikeRZToroidalSection class."""
 
     @pytest.mark.unit
+    @pytest.mark.xfail(reason="GitHub issue 1127.")
```
f0uriest, line 408: why is this failing now? Shouldn't a cross section be able to compute area?

tests/test_configuration.py
```diff
@@ -462,9 +462,12 @@ def test_get_rho_surface(self):
         rho = 0.5
         surf = eq.get_surface_at(rho=rho)
         assert surf.rho == rho
-        np.testing.assert_allclose(surf.compute("S")["S"], 4 * np.pi**2 * R0 * rho)
+        np.testing.assert_allclose(
+            surf.compute("S")["S"] * rho, 4 * np.pi**2 * R0 * rho
+        )
 
     @pytest.mark.unit
+    @pytest.mark.xfail(reason="GitHub issue 1127.")
```
f0uriest, line 470: see other comment, I think this should still work?

tests/test_vmec.py
```diff
@@ -483,12 +483,16 @@ def test_vmec_save_1(VMEC_save):
...
     np.testing.assert_allclose(vmec.variables["chipf"][:], desc.variables["chipf"][:])
     np.testing.assert_allclose(
-        vmec.variables["Rmajor_p"][:], desc.variables["Rmajor_p"][:]
+        vmec.variables["Rmajor_p"][:], desc.variables["Rmajor_p"][:], rtol=3e-4
     )
     np.testing.assert_allclose(
-        vmec.variables["Aminor_p"][:], desc.variables["Aminor_p"][:]
+        vmec.variables["Aminor_p"][:],
+        desc.variables["Aminor_p"][:],
+        rtol=2e-4,
```
f0uriest, line 491: shouldn't the new method of computing things be even more accurate?

desc/compute/_geometry.py
```diff
@@ -26,11 +28,16 @@
     transforms={"grid": []},
     profiles=[],
     coordinates="",
-    data=["sqrt(g)"],
+    data=["sqrt(g)", "V(r)", "rho"],
     resolution_requirement="rtz",
 )
 def _V(params, transforms, profiles, data, **kwargs):
-    data["V"] = jnp.sum(data["sqrt(g)"] * transforms["grid"].weights)
+    if isinstance(transforms["grid"], QuadratureGrid):
```
f0uriest, line 35: can we include logic like this for some of the other cases? My main worry is that for higher order quantities that depend on V,A,a etc we're making it so that you need multiple different grids just to compute everything correctly, but i think in theory a quadrature grid should work for everything (though may be overkill in some cases)

desc/objectives/_geometry.py
```diff
@@ -238,6 +240,8 @@ def __init__(
     ):
         if target is None and bounds is None:
             target = 1
+        if grid is not None:
+            errorif_sym(grid, name)
```
f0uriest, line 244: do we need to check this for other objectives?

f0uriest (changes requested):

tests/test_vmec.py
unalmis: In the limit toward high resolution yes. Looks like at the resolution used in this [test](https://github.com/PlasmaControl/DESC/blob/4b21abcc54f898a07d478edfc263a5f9cb0d6cd5/tests/conftest.py#L328), the new method agrees less with vmec.

I could update my sympy script to compute the analytic result of these finite quadrature sampled on this grid for the SOLOVEV's boundary. This would prove whether the new method is less accurate at this resolution for these sample points.

I haven't done that to confirm. However, I had done similar for other surfaces [e.g. here](https://github.com/PlasmaControl/DESC/pull/807#issuecomment-1881858032), and the numerical quadrature agreed exactly with the analytic quadrature (both of which converge to the true integral).

tests/test_surfaces.py
unalmis: I was pushing this off to the PR that resolved issue 1127, but you are right that I should address this here.

desc/objectives/_geometry.py
unalmis: aspect ratio is fine as is. I checked other objectives were also fine. (haven't looked at new coil objectives)

tests/test_plotting.py
dpanici: In this PR now, will `plot_1d("a_major/a_minor")` work by default (i.e. will the default grid be sufficient)? If this is a quantity that needs the poloidal resolution, I think the .compute logic should by default compute it on the right grids

dpanici (changes requested):
A few questions in the comments, main one is, will `"a_major/a_minor"` be computed on the correct grid by default when called in `eq.compute`? how about when passed into `plot_1d`? I think it should if not

desc/compute/_geometry.py
```diff
@@ -232,23 +212,50 @@ def _A_of_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwarg
...
+    resolution_requirement="tz",
 )
 def _A(params, transforms, profiles, data, **kwargs):
-    data["A"] = jnp.mean(
-        transforms["grid"].compress(data["A(z)"], surface_label="zeta")
+    # Denote any vector v = v¹ R̂ + v² ϕ̂ + v³ Ẑ by v = [v¹, v², v³] where R̂, ϕ̂, Ẑ
+    # are the normalized basis vectors of the cylindrical coordinates R, ϕ, Z.
+    # We use a 2D divergence theorem over constant ϕ toroidal surface (i.e. R, Z plane).
+    # In this geometry, the divergence operator in this coordinate system is
+    # div = ([∂_R, ∂_ϕ, ∂_Z] ⊗ [1, 0, 1]) dot .
```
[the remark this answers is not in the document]
unalmis: . is the English period. `dot` is vector inner product

desc/compute/_geometry.py
```diff
@@ -39,27 +46,19 @@ def _V(params, transforms, profiles, data, **kwargs):
...
-            surface_integrals(
-                transforms["grid"],
-                cross(data["e_theta"], data["e_zeta"])[:, 2] * data["x"][:, 2],
-                expand_out=False,
-            )
-        )
-    )
+    # To approximate volume at ρ ~ 1, we scale by ρ⁻², assuming the integrand
+    # varies little from ρ = max_rho to ρ = 1 and a roughly circular cross-section.
+    data["V"] = jnp.max(data["V(r)"]) / jnp.max(data["rho"]) ** 2
```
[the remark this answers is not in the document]
dpanici: this is a good point, if I want an interior volume this would still extrapolate. Maybe there should be a "V_full" key that does extrapolation and a "V" that just gives the volume? I guess that is just an alias for V(r) though

desc/compute/_geometry.py
unalmis: Yes, but I also changed the compute function so that we scale by max rho. Do we want to do that?

unalmis:
would be less annoying to resolve #1127 before this

desc/compute/_geometry.py
```diff
@@ -233,23 +213,50 @@ def _A_of_z_FourierRZToroidalSurface(params, transforms, profiles, data, **kwarg
...
+    # Denote any vector v = v¹ R̂ + v² ϕ̂ + v³ Ẑ by v = [v¹, v², v³] where R̂, ϕ̂, Ẑ
+    # are the normalized basis vectors of the cylindrical coordinates R, ϕ, Z.
+    # We use a 2D divergence theorem over constant ϕ toroidal surface (i.e. R, Z plane).
+    # In this geometry, the divergence operator in this coordinate system is
+    # div = ([∂_R, ∂_ϕ, ∂_Z] ⊗ [1, 0, 1]) dot .
+    # ∫ dA div v = ∫ dℓ n dot v
+    # where n is the unit normal such that n dot e_θ|ρ,ϕ = 0 and n dot e_ϕ|R,Z = 0,
+    # and the labels following | denote those coordinates are fixed.
+    # Now choose v = [0, 0, Z], and n in the direction (e_θ|ρ,ζ × e_ζ|ρ,θ) ⊗ [1, 0, 1].
+    n = data["n_rho"]
```
unalmis, line 241: iirc the task is to rewrite this to obtain normal vector purely from e_theta which is defined on zeta cross section alone. basically like https://github.com/PlasmaControl/DESC/issues/597#issuecomment-1663194440.

desc/compute/_geometry.py
unalmis: done

tests/test_plotting.py
unalmis: There is not any logic in `master` to automatically override to a grid with sufficient resolution when the resolution requirement is a single coordinate. We have only done this for volume quantities and surface quantities.

This PR corrects the resolution requirement string of dependencies of `a_major/a_minor` to be only poloidal resolution rather than poloidal and toroidal. So, we are now in an uncovered case where the grid overriding doesn't happen automatically. Instead, a warning is raised to user to increase XX resolution. 

I don't think it's worth adding the three more cases to `eq.compute` to cover this. One could just use one of the existing cases, computing a quantity that only needs poloidal resolution on a flux surface grid. However, perhaps the user's quantity requires significant memory and they can't afford computing it on that higher resolution grid when it doesn't change the output.

Feel free to make the change if you prefer.

desc/compute/_geometry.py
unalmis, line 1: https://github.com/PlasmaControl/DESC/pull/1094#discussion_r1688802949

tests/test_vmec.py
unalmis: i don't think my previous comment is relevant

desc/compute/_geometry.py
unalmis: Now, we extrapolate to LCFS on equilibrium and Poincare sections, but not on flux surfaces

desc/compute/_geometry.py
```diff
@@ -358,10 +475,12 @@ def _S_rr_of_r(params, transforms, profiles, data, **kwargs):
 
 @register_compute_fun(
     name="R0",
-    label="R_{0}",
+    label="R_{0} = V / (2\\pi A) = \\int R(\\rho=0) d\\zeta / (2\\pi)",
     units="m",
     units_long="meters",
     description="Average major radius",
+    # This differs from the average value of R on the magnetic axis.
+    # R₀ ≠ 〈 R(ρ=0) 〉 =  ∫ (R ‖e_ζ‖)(ρ=0) dζ / ∫ ‖e_ζ‖(ρ=0) dζ.
```
unalmis, line 483: The following comment refers to the compute_function of `R0`: for flux surfaces, we don't extrapolate for `V` or `A`, so this is still consistent.

tests/test_vmec.py
unalmis: resolved

desc/compute/_geometry.py
unalmis: doable, but fine as is right now.

unalmis:
I do not know why the docs raise a warning and hence fail.

desc/compute/_geometry.py
```diff
@@ -153,74 +153,173 @@ def _V_rrr_of_r(params, transforms, profiles, data, **kwargs):
     return data
 
 
+def _compute_A_of_z(grid, data, extrap=False, mean=False, expand_out=False):
+    max_rho = jnp.max(data["rho"])
+    # TODO: Find new vector field v such that surface divergence
```
unalmis: add marker

desc/compute/_geometry.py
```diff
@@ -153,74 +153,173 @@ def _V_rrr_of_r(params, transforms, profiles, data, **kwargs):
...
+        # Denote any vector v = v¹ R̂ + v² ϕ̂ + v³ Ẑ by v = [v¹, v², v³] where
+        # R̂, ϕ̂, Ẑ are the normalized basis vectors of the cylindrical coordinates
+        # R, ϕ, Z. We use a 2D divergence theorem over constant ϕ toroidal surface
+        # (i.e. R, Z plane). In this geometry, the divergence operator in this
+        # coordinate system is div = ([∂_R, ∂_ϕ, ∂_Z] ⊗ [1, 0, 1]) dot .
+        # ∫ dA div v = ∫ dℓ n dot v where n is the unit normal such that
+        # n dot e_θ|ρ,ϕ = 0 and n dot e_ϕ|R,Z = 0, and the labels following
+        # | denote those coordinates are fixed. Now choose v = [0, 0, Z], and
+        # n in the direction (e_θ|ρ,ζ × e_ζ|ρ,θ) ⊗ [1, 0, 1].
+        n = data["n_rho"]
```
unalmis, line 182: use e_theta_t option

unalmis:
Ok this PR is done now.

desc/compute/_geometry.py
```diff
@@ -153,74 +153,170 @@ def _V_rrr_of_r(params, transforms, profiles, data, **kwargs):
     return data
 
 
+def _compute_A_of_z(grid, data, extrap=False, mean=False, expand_out=False):
+    max_rho = jnp.max(data["rho"])
+    if isinstance(grid, QuadratureGrid) or "n_rho" not in data:  # TODO(#1761)
+        assert extrap
```
f0uriest, line 162: why does extrap need to be true here if its not used?

desc/compute/_geometry.py
```diff
@@ -153,74 +153,170 @@ def _V_rrr_of_r(params, transforms, profiles, data, **kwargs):
...
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="",
+    data=["Z", "n_rho", "e_theta", "rho", "|e_rho x e_theta|"],
+    parameterization=["desc.equilibrium.equilibrium.Equilibrium"],
+    resolution_requirement="tz",
+)
+def _A(params, transforms, profiles, data, **kwargs):
+    # noqa: unused dependency
+    data["A"] = _compute_A_of_z(transforms["grid"], data, extrap=True, mean=True)
```
f0uriest, line 250: Why not just take the mean of A(z) here? Won't the way it's structured now lead to double calculations if the user requests both `A` and `A(z)`?

desc/compute/_geometry.py
```diff
@@ -494,7 +620,87 @@ def _a_major_over_a_minor(params, transforms, profiles, data, **kwargs):
...
+    "extrapolated to last closed flux surface.",
+    dim=1,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="z",
+    data=["A(z)", "perimeter(z)", "rho"],
+    parameterization=["desc.geometry.surface.FourierRZToroidalSurface"],
+)
+def _a_major_over_a_minor_flux_surface_lcfs(
```
f0uriest: what's the need for this? It seems like all the other stuff for RZSurface doesn't do any extrapolation?

desc/grid.py
```diff
@@ -992,7 +998,11 @@ class LinearGrid(_Grid):
         Change this only if your nodes are placed within one field period
         or should be interpreted as spanning one field period.
     sym : bool
-        True for stellarator symmetry, False otherwise (Default = False).
+        Whether to truncate the poloidal domain to [0, π] ⊂ [0, 2π).
```
f0uriest: For users I think it would be better to make it clearer that this is "our version of stellarator symmetry"

eg something like
```
True for stellarator symmetry, False otherwise (Default = False).
Note that this is implemented as poloidal up/down symmetry, which is 
distinct from stellarator symmetry, but when stellarator symmetry exists,
flux surface integrals and volume integrals are invariant to this 
truncation.
```

desc/objectives/_stability.py
```diff
@@ -473,6 +473,9 @@ def build(self, eq=None, use_jit=True, verbose=1):
         iota_profiles = get_profiles(self._iota_keys, obj=eq, grid=iota_grid)
         iota_transforms = get_transforms(self._iota_keys, obj=eq, grid=iota_grid)
 
+        # TODO: Generalize balloning stabilty funs to multiple flux surfaces,
```
f0uriest: Tag the PR/issue number here for reference

tests/test_plotting.py
f0uriest: Is this still the case?

desc/compute/_geometry.py
unalmis: When we integrate on a quadrature grid, the weights are always set to integrate from rho = [0, 1]. ``assert extrap`` ensures that this branch only executes when the intention is to compute an integral from rho = [0, 1] (or equivalently integrate on rho=1 flux surface via stokes).

desc/compute/_geometry.py
unalmis: `A(z)` requires `grid.sym == False` to compute. `A` does not have that limitation. If I make `A(z)` a dependency of `A` then `A` is also constrained by this limitation. I would like to compute length scale quantities such as `R0`, `A`, `a` on grids with symmetry, so that I can compute them in objectives without making additional transforms.

desc/compute/_geometry.py
unalmis: All other objects  besides `FourierRZToroidalSurface` have the extrapolation built-in so that the computation is the LCFS of rho=1 computation.

For `FourierRZToroidalSurface`, the natural expectation is that the computations are done on that surface and hence are not extrapolated to the rho=1 surface.

This is natural because when a user requests to compute the surface area of a flux surface, the intention need not be to compute the area of the last closed flux surface. When user requests to compute the surface area of a Poincare cross-section, the expectation, I believe, is always to compute surface area from rho = [0, 1] x theta = [0, 2pi] rather than some subset of rho.

Now elongation is a quantity that is associated to a Poincare cross-section, not a flux surface. So when a user computes elongation on a `FourierRZToroidalSurface` that isn't on `rho=1` for whatever reason (maybe they are computing something else with the transforms built on this grid that would give nan on boundary etc.), it is no longer clear whether the expectation is to not extrapolate, since they requested the computation to be done on rho!=1, or to extrapolate to rho=1, since they requested to compute a quantity that is associated with a Poincare section.

By including this quantity, this dilema is resolved. Unlike the extrapolation for a simple quantity like volume which is a simple scaling, extrapolation for elongation is nonlinear, so it is not something that is simple for the user to do themselves.

tests/test_plotting.py
unalmis: Yes. `eq.compute`, both here and on `master` does not override to the "correct" grid for quantities with `reolution_requirement=` a single coordinate.

unalmis: However, it will always warn the user, assuming that warning is not suppressed explicitly elsehwere in the code, when the given grid is not "correct" for any quantity, so there has not been a need to automate this process for single coordinate resolution_requirement things.

desc/grid.py
unalmis: Good idea

desc/compute/_geometry.py
f0uriest: I think we could still get the best of both worlds if we did something like:
```python
if "A(z)" in data:
    data['A'] = jnp.mean(data['A(z)'])
else:
    data['A'] = _compute_A_of_z(...)
```

desc/compute/_geometry.py
f0uriest: I think if they ask for the elongation of an RZToroidalSurface the assumption is they want it on that particular surface, not extrapolated, hence I don't think we really need this function.

desc/compute/_geometry.py
unalmis: i added this per your request, but the utility of it may be narrower than it appear because there is no guarentee A(z) is computed before A.

f0uriest (approved):

rahulgaur104 (approved):

unalmis:
I need assistance overriding codecov for the squash merge.

PR #1095, opened 2024-07-03
Improve warning for missing IOable attributes

unalmis:


rahulgaur104 (approved):

desc/io/hdf5_io.py
```diff
@@ -124,7 +124,23 @@ def read_obj(self, obj, where=None):
         for attr in obj._io_attrs_:
             if attr not in loc.keys():
                 warnings.warn(
-                    "Save attribute '{}' was not loaded.".format(attr), RuntimeWarning
+                    f"Save attribute '{attr}' was not loaded from the input file.\n"
+                    "This is likely because the file from which you are loading the "
+                    f"Python object '{obj}' was created prior to the time '{attr}' "
+                    f"became an attribute of objects of the class '{type(obj)}'.\n"
+                    "\n"
+                    "Note to developers: Add 'def _set_up(self)' as a method to class "
```
f0uriest, line 132: Not sure we want to put this as part of the public warning here. Maybe just explain to the user that a default value will (hopefully) be set automatically, so they can try saving and reloading, and if the warning persists then it might be a bug?

YigitElma: I am actually ok with more descriptive warnings. Maybe for those people who repetitively load old equilibrium might get sick of getting this warning but overall I am fine.

YigitElma:
@unalmis Since it is a bit late for the USA, I fixed the problem that caused the tests to fail. I believe everything should pass now.

desc/io/hdf5_io.py
rahulgaur104: I prefer more descriptive warnings as well.

unalmis: Developers have lost time with this in #586 and #819, so I would like a warning that says how to fix it

YigitElma (approved):

rahulgaur104 (dismissed):

f0uriest (dismissed):

YigitElma (dismissed):

desc/io/hdf5_io.py
ddudt: This warning will still be thrown even if the missing attribute is added to the `_set_up` method, right? In that case the user needs to check the default value and then re-save the object to avoid this warning again. So I don't think this is the best place for that note to developers. 

Also why blue? I think all of our other warnings are colored yellow (and errors are red).

unalmis: Yes. There's no other place to put this warning for developers to see. For it to be useful it would need to be a warning that is raised prior to this one. I chose blue because yellow and red imply something is wrong to the user, and that should not be the goal of this warning. 

A complete solution would be adding a verbosity plumbing to all warnings. By default users would get verbose=`user_warnings` and only see the first half of the message. The default when running pytest should be user_warning + developer_warning and they can see both.

dpanici (changes requested):
It is not an "input file", it is really an "output file" since it is an .h5, maybe instead say "file being loaded".

f0uriest (approved):

dpanici (approved):

PR #1099, opened 2024-07-04
Ko/test coil opt

### LINKED ISSUE #1021: Add test to Check that we can optimize with every coil type

kianorr:
Test every coil type. Resolves #1021 

Also contains bug fixes:
- fixes bug when computing magnetic field with `FourierRZcoil` that has `NFP>1`, previously the grid used had the same `NFP` but that meant that only part of the coil was being included in the biot savart integral. Now grids for magnetic field computation from `FourierRZCoil` have `NFP=1` and have `NFP` multiplying the usual default grid resolution to avoid this issue

Replaces PR #1025


kianorr:
Passes when `QuadraticFlux` is not an objective, but having it there causes a jax error. Maybe will be resolved by #1002 ?

kianorr:
And also, should we remove `test_single_coil_optimization()` with the addition of this test?

tests/conftest.py
```diff
@@ -276,12 +277,22 @@ def DummyMixedCoilSet(tmpdir_factory):
 
     tf_coil = FourierPlanarCoil(current=3, center=[2, 0, 0], normal=[0, 1, 0], r_n=[1])
     tf_coilset = CoilSet.linspaced_angular(tf_coil, n=4)
+    tf_coilset = CoilSet(tf_coilset.coils, NFP=tf_coilset.NFP, sym=True)
```
dpanici, line 280: Why do this? if you set `sym=True` but pass in these 4 coils, it essentially doubles the magnetic field by placing the symmetry coils on top of the other coils, and the min dist btwn coils in this coilset is now 0. I think we don't want this

tests/test_examples.py
```diff
@@ -1336,30 +1337,49 @@ def test_second_stage_optimization_CoilSet():
...
+        load_from=str(DummyMixedCoilSet["output_path"]), file_format="hdf5"
+    )
+    nested_coils = MixedCoilSet(sym_coils, mixed_coils)
+
+    def test(c, method):
+        target = 11
+        obj = ObjectiveFunction(
+            (
+                CoilLength(c, target=target),
+                QuadraticFlux(eq=Equilibrium(), field=c, vacuum=True),
```
dpanici, line 1357: Instead of instantiating an `Equilibrium` each time test is called, just pass it in to `test` as an argument or something like that, to avoid the cost of making a new eq each time.

dpanici (changes requested):
couple small changes

dpanici:
> Okay, I checked off all the boxes in the corresponding issue.
> 
> Should we remove `test_single_coil_optimization()` with the addition of this test? Or do we want to keep it because it tests curvature and torsion optimization?

I vote for removing it, maybe you can add the torsion and curvature as an additional test for the mixed coilset though (like just for the mixed coilset or the nested coils, when you call test you also include the torsion and curvtsure)? with some loose tols on it

tests/conftest.py
kianorr: Hmm I remember @ddudt  requesting to pass in `sym=True` for this coilset to test symmetry but maybe I misunderstood what he said

dpanici: I think you can only pass in the first two of the coils, then set `sym=True` or something like that,so it uses symmetry to get the other two coils

dpanici: hm or maybe not, I will think but you should be able to make the same coilset using `sym=True` and `NFP` I think

dpanici:
```python
from desc.coils import CoilSet, FourierPlanarCoil
tf_coil = FourierPlanarCoil(current=3, center=[2, 0, 0], normal=[0, 1, 0], r_n=[1])
tf_coilset = CoilSet.linspaced_angular(tf_coil, n=4)
tf_coilset1 = CoilSet(tf_coilset.coils[0], NFP=2, sym=True)```

@ddudt why does this not work? when I do compute position I find two of the coils are on top of eachother, when I would expect there to b3 4 distinct coils

Comment (dpanici) [reply tests/conftest.py:280]:
```python
tf_coil = FourierPlanarCoil(current=3, center=[2, 0, 0], normal=[0, 1, 0], r_n=[1])
tf_coil.rotate(angle=np.pi/4)
tf_coilset1 = CoilSet(tf_coil, NFP=2, sym=True)
```
this works @kianorr

dpanici:
I made the test run a bit faster @kianorr

dpanici (dismissed):

desc/geometry/curve.py
```diff
@@ -925,6 +925,45 @@ def method(self, new):
...
+        grid : Grid or int, optional
+            Grid of coordinates to evaluate at. Defaults to a Linear grid.
+            If an integer, uses that many equally spaced points.
+        params : dict of ndarray
+            Parameters from the equilibrium. Defaults to attributes of self.
+        transforms : dict of Transform
+            Transforms for R, Z, lambda, etc. Default is to build from grid
+        data : dict of ndarray
+            Data computed so far, generally output from other compute functions
+        Returns
```
f0uriest, line 952: need a blank line here

tests/test_examples.py
```diff
@@ -1336,30 +1299,84 @@ def test_second_stage_optimization_CoilSet():
...
-    (c,), _ = optimizer.optimize(c, objective=objective, maxiter=200, ftol=0, xtol=0)
-    np.testing.assert_allclose(c.compute("length")[1]["length"], 11, atol=1e-3)
-
-    # in CoilSet
-    c = CoilSet(FourierPlanarCoil(), sym=True, NFP=2)
-    objective = ObjectiveFunction(CoilLength(c, target=11))
-    optimizer = Optimizer("fmintr")
-    (c,), _ = optimizer.optimize(c, objective=objective, maxiter=200, ftol=0, xtol=0)
-    np.testing.assert_allclose(c.compute("length")[0]["length"], 11, atol=1e-3)
+    test(FourierPlanarCoil(), "fmintr")
```
f0uriest, line 1368: might be able to refactor this to use `pytest.mark.parametrize` which treats it as different tests rather than 1 big one so easier to debug and run specific ones in case of failure.

dpanici (approved):

f0uriest (dismissed):

tests/test_examples.py
dpanici: #1104

tests/test_examples.py
```diff
@@ -1336,30 +1299,84 @@ def test_second_stage_optimization_CoilSet():
...
+def test_optimize_with_all_coil_types(DummyCoilSet, DummyMixedCoilSet):
+    """Test optimizing for every type of coil and dummy coil sets."""
+    sym_coils = load(load_from=str(DummyCoilSet["output_path_sym"]), file_format="hdf5")
+    asym_coils = load(
+        load_from=str(DummyCoilSet["output_path_asym"]), file_format="hdf5"
+    )
+    mixed_coils = load(
+        load_from=str(DummyMixedCoilSet["output_path"]), file_format="hdf5"
+    )
+    nested_coils = MixedCoilSet(sym_coils, asym_coils)
```
ddudt, line 1313: `nested_coils` here actually has a similar structure as `mixed_coils`. What I meant by "nested" coilsets was `MixedCoilSet`s within other `MixedCoilSet`s. So I think we want something like this instead: 
```
nested_coils = MixedCoilSet(mixed_coils, sym_coils, asym_coils, spline_coil)
```
or anything else that contains `mixed_coils` inside another `MixedCoilSet`

tests/test_examples.py
```diff
@@ -1336,30 +1299,84 @@ def test_second_stage_optimization_CoilSet():
...
+        )
+        optimizer = Optimizer(method)
+        (c,), _ = optimizer.optimize(c, obj, maxiter=2, ftol=0, xtol=1e-15)
+
+        # now check with optimizing geometry and actually check result
+        objs = [
+            CoilLength(c, target=target),
+        ]
+        extra_msg = ""
+        if isinstance(c, MixedCoilSet):
```
ddudt, line 1343: Why do we only test curvature and torsion for `MixedCoilSet` and not other types of coils?

tests/test_linear_objectives.py
```diff
@@ -998,13 +996,11 @@ def test_fix_coil_current(DummyMixedCoilSet):
...
-    assert obj.dim_f == 8
-    np.testing.assert_allclose(obj.target, [3, 3, 3, 3, -1, -1, -1, 2])
+    assert obj.dim_f == 6
+    np.testing.assert_allclose(obj.target, [3, -1, -1, -1, 2, 1])
 
     # only fix currents of some coils in the coil set
-    obj = FixCoilCurrent(
-        coil=coilset, indices=[[True, False, True, False], False, True]
-    )
+    obj = FixCoilCurrent(coil=coilset, indices=[[True], False, True, True])
```
ddudt, line 1003: It would be good to still include a case where only some coils within a `CoilSet` have fixed currents. Now only the vertical field coilset has multiple coils, so I think something like this should work: 
```
obj = FixCoilCurrent(coil=coilset, indices=[[True], [True, False, True], True, False])
```
and then I think the test in the line below would be: 
```
np.testing.assert_allclose(obj.target, [3, -1, -1, 2])
```

ddudt (changes requested):

tests/test_examples.py
kianorr: Not sure. My thought was that it just wasn't needed. @dpanici any reason for this?

dpanici: MixedCoilSet contains the other types of coils, I did not want to reduce test coverage by removing the other test for it so I added it here.

ddudt: @dpanici but why not test them for all the coils? Removing this if statement should increase the test coverage, if anything

dpanici: They are expensive and this test already takes a while to run so I did not think it was necessary.

f0uriest (approved):

tests/test_examples.py
dpanici: As it is now, it takes 100s to run locally for me the test. If I change it around so that torsion and curvature are included with the quad flux for just that part of the test to ensure it runs without error, it takes way longer (5 minutes locally, checking again now to be sure it was not a fluke). That is really long, I'd rather not test it and trust that the logic works the same as the length. Or maybe just test ONE of them to cover the category of coil geometry objective that is not dim 1 like length is

dpanici: @ddudt torsion just adds too much time to the optimization, so I am fine with keeping the test as is.

dpanici (dismissed):

dpanici:
@kianorr wait for daniel and rory to approve before merging, my approval is a bit of a rubber stamp as I made commits too

desc/objectives/_coils.py
```diff
@@ -128,7 +128,10 @@ def _prune_coilset_tree(coilset):
 
         # map grid to list of length coils
         if grid is None:
-            grid = [LinearGrid(N=2 * c.N + 5, endpoint=False) for c in coils]
+            grid = []
+            for c in coils:
+                NFP = c.NFP if hasattr(c, "NFP") else 1
+                grid.append(LinearGrid(N=2 * c.N + 5, NFP=NFP, endpoint=False))
```
ddudt, line 134: @dpanici I don't think we want this change. Only `CoilSet` has an `NFP` attribute, and `coils` here is always a list of `Coil` objects, so this will never get called. And it would be bad if it did, because NFP>1 would mean that the grid wouldn't span the entire length of the coils (remember we use the zeta coordinates of the grid to represent the curve parameter s).

ddudt (changes requested):

desc/objectives/_coils.py
dpanici: `FourierRZCoil` also has `NFP`

dpanici: I have not thought about that though, for a `FourierRZCurve` biot savart law... yea I will revert this as it probably has unintended consequences for that calculation

dpanici (dismissed):

desc/objectives/_coils.py
ddudt: OK I forgot about `FourierRZCoil`. But I still think for these objectives that inherit from `_CoilObjective` like `CoilLength` etc. we always want the grid to span the entire length of the coil.

ddudt (dismissed):

desc/objectives/_coils.py
dpanici: Food for thought: should this actually be
```python
 for c in coils:
      NFP = c.NFP if hasattr(c, "NFP") else 1
      grid.append(LinearGrid(N=2 * c.N * NFP + 5, NFP=1, endpoint=False))
```

Since say a `FourierRZCoil` has `N=5` and `NFP=19`... without this change the coil would have only `LinearGrid(N=15)` which is `31` points for an effective toroidal resolution of `NFP*5 = 95`

dpanici: @ddudt

dpanici: I think it should be

f0uriest (dismissed):

desc/objectives/_coils.py
ddudt: I think you are right, good catch. But then we also need to make this same change [here](https://github.com/PlasmaControl/DESC/blob/master/desc/geometry/core.py#L116) (and maybe other places).

dpanici: maybe just in the .compute_magnetic_field method of `FourierRZCoil` would be enough, then we basically cover everywhere that would require this (since none of the data index compute quantities care about having `NFP>1` on the grid,they work fine)

dpanici (dismissed):
approved pending tests passing

f0uriest (dismissed):

kianorr:
Can I merge this or do you want to give this another review @ddudt ?

ddudt:
> Can I merge this or do you want to give this another review @ddudt ?

Give me a minute to review again, thanks

desc/geometry/curve.py
ddudt, line 1: I still think we should edit the default grid in `Curve.compute` to account for NFP. It might not matter for any of the existing compute quantities, but this could introduce bugs in the future.

desc/coils.py
```diff
@@ -226,6 +226,12 @@ def compute_magnetic_field(
             current = self.current
         else:
             current = params.pop("current", self.current)
+        if source_grid is None and hasattr(self, "NFP"):
+            # NFP=1 to ensure we have points along whole grid
+            # multiply by NFP in case the coil has nonzero NFP
```
ddudt, line 231: typo: "... has NFP>1 ..." (not nonzero)

desc/objectives/_coils.py
```diff
@@ -128,7 +128,13 @@ def _prune_coilset_tree(coilset):
 
         # map grid to list of length coils
         if grid is None:
-            grid = [LinearGrid(N=2 * c.N + 5, endpoint=False) for c in coils]
+            grid = []
+            for c in coils:
+                NFP = c.NFP if hasattr(c, "NFP") else 1
+                # NFP=1 to ensure we have points along whole grid
+                # multiply by NFP in case the coil has nonzero NFP
+                # to make ensure sufficient resolution across periods
```
ddudt, line 136: typos:
"... has NFP>1 ..." (not nonzero)
"to ensure" (remove "make")

ddudt (changes requested):

desc/geometry/curve.py
dpanici: Can we not do this in this PR though? I don't want to have to add it then see if any tests fail. It works fine and as expected now and as far as we checked, the only things that matter are the compute magnetic field and we've addressed that now

ddudt: It's literally a one line change, and it's related to the other changes already in this PR. We shouldn't get in the habit of not making fixes just because we are too impatient to wait for the tests to pass

ddudt: For the record, I think this issue only matters for `Coil.compute_magnetic_field` and not the curve geometry stuff. But since we already changed it for `_CoilObjective` (which only deals with curve geometry for now), I also changed it for `Curve.compute` for consistency.

dpanici (approved):

desc/objectives/_coils.py
ddudt: My main point was that since you already changed it in `_CoilObjective.build` we also had to change it in `Curve.compute` for consistency, since `_CoilObjective` basically just calls `Curve.compute`.

desc/geometry/curve.py
ddudt: My main point was that since you already changed it in `_CoilObjective.build` we also had to change it in `Curve.compute` for consistency, since `_CoilObjective` basically just calls `Curve.compute`.

ddudt:
OK now `test_compute_everything` is failing because of the new default grid. Options: 

1. Update the `master_compute_data.pkl` to reflect the new grid. 
2. Revert the changes to the default grid in `Curve.compute` and `_CoilObjective.build`, but keep the `NFP` factor in `Coil.compute_magnetic_field`. 

I am fine with either option. @dpanici

dpanici:
Let's revert

desc/geometry/curve.py
dpanici: It does not though, it has the logic for the grids because it pre-computes the transforms, and if we are computing things in coil objectives it is mjuch more likely that we need to know the entire coil curve (for example, if we have a fOUrierRZcoil and some other coils that lack the same discrete symmetry in the same coilset, we may need to know the pts on the whole length for coilset distance to be computed accurately)

dpanici (approved):

ddudt (approved):

f0uriest (approved):

PR #1102, opened 2024-07-05
Clebsch field line coordinate functions

unalmis:
New quantities used to compute some neoclassical quantities are moved here as requested at developer meeting.


tests/test_compute_funs.py
```diff
@@ -1683,3 +1684,37 @@ def test_surface_equilibrium_geometry():
...
+            "B^zeta_z|r,a",
+            "|B|",
+        ],
+    )
+    np.testing.assert_allclose(data["e_zeta|r,a"], (data["B"].T / data["B^zeta"]).T)
+    np.testing.assert_allclose(
+        data["|B|_z|r,a"], dot(data["grad(|B|)"], data["e_zeta|r,a"])
+    )
+    # FIXME: Don't know why below test fails.
+    np.testing.assert_allclose(
```
unalmis, line 1713: See discussion [here](https://github.com/PlasmaControl/DESC/pull/1003/files#r1655660318). I can't add the finite difference test from that branch here because it relies on logic introduced in #1024 .

dpanici:
add our notation in List of Vairables and that by default is rho theta zeta

f0uriest (approved):

unalmis:
There is a resolution requirement for `<|grad(rho)|>`  that was on `ripple` branch, but I forgot to add here. Corrected now.

f0uriest (approved):

dpanici (approved):

PR #1106, opened 2024-07-07
Allow MixedCoilSet.from_symmetry to work when the given coils are not CoilSet-compatible

dpanici:
- Fixes bug that can occur when trying to call `MixedCoilSet.from_symmetry()` with a group of coils which are not `CoilSet` compatible (that is, they are not all the same type, or they are the same type but have differing parameterizations like different number of spline points or different Fourier resolutions)
  - `MixedCoilSet.from_symmetry` should work in these cases, as it is just doing rotation and reflection operations, it is perfectly valid for a diverse set of unique coils.

### DIFF
--- a/desc/coils.py
+++ b/desc/coils.py
@@ -1153,9 +1153,15 @@ class CoilSet(OptimizableCollection, _Coil, MutableSequence):
 
         """
         if not isinstance(coils, CoilSet):
-            coils = CoilSet(coils)
-
-        [_check_type(coil, coils[0]) for coil in coils]
+            try:
+                coils = CoilSet(coils)
+            except (TypeError, ValueError):
+                # likely there are multiple coil types,
+                # so make a MixedCoilSet
+                coils = MixedCoilSet(coils)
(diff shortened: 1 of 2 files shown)


desc/coils.py
```diff
@@ -1153,9 +1153,15 @@ def from_symmetry(cls, coils, NFP=1, sym=False):
 
         """
         if not isinstance(coils, CoilSet):
-            coils = CoilSet(coils)
-
-        [_check_type(coil, coils[0]) for coil in coils]
+            try:
+                coils = CoilSet(coils)
+            except (TypeError, ValueError):
```
dpanici, line 1158: the code cov here is outdated, these lines are covered if you check the code cov report

f0uriest (approved):

ddudt (approved):
This is somewhat unrelated, but would it make more sense to have `MixedCoilSet` be the parent class and `CoilSet` as the child? Since `MixedCoilSet` is the most general and `CoilSet` is the special case where all of the coils have to be the same type.

PR #1109, opened 2024-07-08
Change error to warning when coils lie in symmetry plane when calling `from_symmetry` method, and add Coil-Coil distance method to `CoilSet`

### LINKED ISSUE #1107: `CoilSet` `NFP` symmetry plane check is too strict for `sym=False` case
When `sym=False` and `CoilSet.from_symmetry` is called with `NFP>1`, currently an error is thrown if any coil has a `phi` value greater than `2pi/NFP`. However, it is perfectly fine for that to happen, as long as the coil on the other side of the field period is  far enough away from the symmetry plane. This should instead be a warning to check the coilset for possibleintersections

example of a coilset I've found (on `dp/pyREGCOIL`) which has a coil across the symmetry line (though technically when the `phi` is computed, it goes negative at the symmetry plane so the check does not actually catch this anyways, but I found it when I was trying to do the checking as the coils were cut from the surface current), but the actual coilset is fine and has no intersections:

(Note that you will need to reinstall the requirements.txt to run this example as I added a new dependency for the contour finding on that PR, since the matplotlib contour finding is deprecated so I use `skimage` now)

```python
import numpy as np

from desc.backend import jnp
from desc.examples import get
from desc.io import load
from desc.magnetic_fields import (
    FourierCurrentPotentialField,
    run_regcoil,
)
from desc.objectives import (
   CoilsetMinDistance,
)

"""Run regcoil for precise QA eq and surface with modular coils."""
eq = get("precise_QA")
surf_winding = eq.surface.constant_offset_surface(
    offset=0.2,  # desired offset
    M=16,  # Poloidal resolution of desired offset surface
    N=12,  # Toroidal resolution of desired offset surface
    grid=LinearGrid(M=32, N=16, NFP=eq.NFP),
)
M_Phi = 8
N_Phi = 8
M_egrid = 30
N_egrid = 30
M_sgrid = 50
N_sgrid = 50
alpha = 1e-18

surface_current_field = FourierCurrentPotentialField.from_surface(
    surf_winding, M_Phi=M_Phi, N_Phi=N_Phi
)
surface_current_field, data = run_regcoil(
    surface_current_field,
    eq,
    eval_grid=LinearGrid(M=M_egrid, N=N_egrid, NFP=eq.NFP, sym=True),
    source_grid=LinearGrid(M=M_sgrid, N=N_sgrid, NFP=eq.NFP),
    alpha=alpha,
    current_helicity=0,
    vacuum=True,
)

"""Test precise QA modular regcoil solution coils."""

initial_surface_current_field = surface_current_field
eq,
surface_current_field = initial_surface_current_field.copy()

# test finding coils

numCoils = 20


coilset2 = surface_current_field.to_CoilSet(
    desirednumcoils=numCoils, stell_sym=False,use_FourierXYZ=True, N=150 
)
coords = eq.compute(["R", "phi", "Z", "B"])
# coilset magnetic field is totally fine
B = coords["B"]
coords = np.vstack([coords["R"], coords["phi"], coords["Z"]]).T
B_from_coils = coilset2.compute_magnetic_field(coords, basis="rpz")
np.testing.assert_allclose(B, B_from_coils, atol=4e-3)
```
```python
# the last coil crosses the symmetry plane of phi=pi
plt.figure()
plt.plot(coilset2[-1].compute("phi")["phi"])
plt.ylabel("phi")
```
image

```python
# coilset is not self intersecting
from desc.objectives import CoilsetMinDistance
obj = CoilsetMinDistance(coilset2,grid=LinearGrid(N=200))
obj.build()
ds=obj.compute(coilset2.params_dict)
print(ds)
print(np.any(np.isclose(ds,0)))
```
yields
```
[0.07708691 0.07709347 0.07879145 0.08251444 0.08812162 0.09420819
 0.09942298 0.10626569 0.11640509 0.129338   0.13302544 0.12933886
 0.11639498 0.10625844 0.09943329 0.09420785 0.08811909 0.08248767
 0.07877026 0.07708691 0.07708691 0.07709347 0.07879145 0.08251444
 0.08812162 0.09420819 0.09942298 0.10626569 0.11640509 0.129338
 0.13302544 0.12933886 0.11639498 0.10625844 0.09943329 0.09420785
 0.08811909 0.08248767 0.07877026 0.07708691]
False
```

```python
# confirm visually that coilset is fine
from desc.plotting import *
plot_coils(coilset2)```
```
image

dpanici:
- Changes the error to a warning when when coils lie in symmetry plane when calling `from_symmetry` method of `CoilSet`, as there can be valid coilsets which have coils that lie over the symmetry plane.
- Add method ``is_self_intersecting`` to ``CoilSet``, which checks if any coils intersect eachother in the coilset.
- Removes error in ``from_symmetry`` method of ``CoilSet`` when a coil crosses the symmetry plane,
and instead adds a check for intersection, to allow for valid coilsets which may cross the
symmetry plane but not be self-intersecting after rotation/reflection.
- Also fixes small bug in `FourierPlanarCurve` that can occur during the error checking of setting `center` or `normal`, where `_basis` is attempted to be accessed before it is assigned.
- Also changes the default grids for `PlasmaCoilSetMinDistance` and `CoilSetMinDistance` to be the defaults for each coil type in the grid instead of an arbitrary resolution LinearGrid.
- [x] put check into init of `CoilSet`
- [x] make issue for compute position default grid for mixed coilset: do efficient way for CoilSet or if grid is passed in, if girdis none and MixedCoilSet, then use logic from coilobjective to compute based on default grid and then rotate/reflect individually
- [x] put default grid for init check (any auto check) and just use the N=100 grid for default inside the self intersecting check.

Resolves #1107


desc/coils.py
```diff
@@ -1166,16 +1166,32 @@ def from_symmetry(cls, coils, NFP=1, sym=False):
...
         for i, cdata in enumerate(data):
-            errorif(
+            warnif(
                 np.any(cdata["phi"] > maxphi),
-                ValueError,
-                f"coil {i} exceeds the toroidal extent for NFP={NFP} and sym={sym}",
+                UserWarning,
+                f"coil {i} crosses the phi=2pi/{NFP*(sym+1)} symmetry plane"
+                + f" for NFP={NFP} and sym={sym},"
+                + "it is recommended to check the coil-coil separation with method"
```
f0uriest, line 1176: given that we check this automatically below i don't think we need these warnings here (maybe we just store the indices of offending coils, and include those in the warning below?)

desc/coils.py
```diff
@@ -1198,8 +1214,29 @@ def from_symmetry(cls, coils, NFP=1, sym=False):
             rotated_coils = coils.copy()
             rotated_coils.rotate(axis=[0, 0, 1], angle=2 * jnp.pi * k / NFP)
             coilset += rotated_coils
+        coilset_to_return = cls(*coilset)
+        if check_coil_coil_dists:
+            grid = LinearGrid(N=100)
+            min_dists = coilset_to_return.compute_minimum_intercoil_distances(grid=grid)
+            intersection = np.any(min_dists < np.finfo(min_dists.dtype).eps)
+            near_intersection = np.any(min_dists < 1e-2)
```
f0uriest, line 1222: For a discretized coil I think the only way they'd be actually touching is if they're almost literally the same coil in the same position, so it might be better to just lump everything into near intersections

Also, for near intersection it's probably good to scale things by some characteristic size and account for the finite grid spacing. A simple heuristic could be "if the closest point to zeta=x on coil 0 is on a different coil"?

desc/coils.py
```diff
@@ -1492,6 +1529,27 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
         coils = [coil.to_SplineXYZ(knots, grid, method) for coil in self]
         return self.__class__(*coils, NFP=self.NFP, sym=self.sym, name=name)
 
+    def compute_minimum_intercoil_distances(self, grid=None):
```
f0uriest, line 1532: we had talked about this before and while I agree something like this is necessary, I don't know if we want to commit to this API at the moment.

For the current purpose, it might be simpler to make it something more like `eq.is_nested` where we just check if there are any likely intersections, returning a bool and raising a warning.

desc/coils.py
dpanici: True, I will add tmrw but if anuone wants to beat me to it go ahead

desc/coils.py
dpanici: I will make the change tmrw, if anyone wants to beat me to it go ahead

I will just keep the same exact fxn but add a threshold argument and have it return a bool if any of the distances are below the threshold (maybe also returning the indices of the offending coils)

desc/coils.py
dpanici: compute dist matrix and check if argmin along each axis is different than the the inds corresponding to that coil

tests/test_objective_funs.py
```diff
@@ -867,12 +867,13 @@ def test(coil, grid=None):
     def test_coil_min_distance(self):
         """Tests minimum distance between coils in a coilset."""
 
-        def test(coils, mindist, grid=None):
+        def test(coils, mindist, grid=None, test_method=False, expect_intersect=False):
```
f0uriest, line 870: what is `test_method`?

desc/coils.py
f0uriest: I think what you have now also works fine. Assuming the objective is normalized, a sensible default tol would be 1/number of grid poiints, I think that would be pretty similar to what we talked about but probably easier to implement

tests/test_objective_funs.py
dpanici: it was before when I was testing the method's warning when things intersect. I removed now

desc/coils.py
dpanici: I implemented your suggestion, not really sure which is better tbh. My original implementation was a bit cleaner but did rely on a tol.

It is not clear to me which way is more robust either

dpanici: For now my test for this passes fine, so I think I am good with keeping this implementation for now. We could also do it where if you pass in `tol` it will use my original implementation, and if you dont it defaults to `"auto"` which uses this way of checking based off of which coil pt is closest

f0uriest (approved):

desc/coils.py
```diff
@@ -1172,17 +1173,17 @@ def from_symmetry(cls, coils, NFP=1, sym=False):
...
-                f"coil {i} is on the symmetry plane phi=0",
+            # check if any coils lie on symmetry planes, these may cause
+            # self-intersection after reflection/rotation
+            maybe_bad_bool = np.any(cdata["phi"] > maxphi) or (
+                sym and np.any(cdata["phi"] < np.finfo(cdata["phi"].dtype).eps)
             )
+            if maybe_bad_bool:
+                # check the coilset for self-intersection before returning coilset
+                # if any lie on symmetry planes
+                check_for_intersection = True
```
ddudt, line 1186: You are using one bool to set another bool. This could be simplified as: 
```
check_for_intersection = np.any(cdata["phi"] > maxphi) or (
    sym and np.any(cdata["phi"] < np.finfo(cdata["phi"].dtype).eps)
)
```

desc/coils.py
```diff
@@ -1204,8 +1205,14 @@ def from_symmetry(cls, coils, NFP=1, sym=False):
             rotated_coils = coils.copy()
             rotated_coils.rotate(axis=[0, 0, 1], angle=2 * jnp.pi * k / NFP)
             coilset += rotated_coils
+        coilset_to_return = cls(*coilset)
+        if check_for_intersection:
```
ddudt, line 1209: So this will always return the coilset regardless, this is just to possibly throw the warning?

desc/coils.py
```diff
@@ -1204,8 +1205,14 @@ def from_symmetry(cls, coils, NFP=1, sym=False):
...
+            grid = LinearGrid(N=100)
```
ddudt, line 1210: Is this a good hard-coded value or would it be better to make this based on the coil resolution?

desc/coils.py
```diff
@@ -1498,6 +1505,99 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
+            two coils are closer than this tolerance, then the function will return
+            True and a warning will be raised. If not passed, then the method used
+            to determine coilset intersection will be based off of checking that
+            each point on a coil is closest to a point on the same coil, which does
+            not rely on a ``tol`` parameter.
+
+        Returns
+        -------
+        is_self_intersecting : bool
+            Whether or not any coils in the CoilSet come close enough to eachother to
```
ddudt, line 1541: typo: "each other"
This returned bool doesn't actually get used anywhere? That's fine, just want to make sure I understand what's going on.

desc/coils.py
```diff
@@ -1498,6 +1505,99 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
         coils = [coil.to_SplineXYZ(knots, grid, method) for coil in self]
         return self.__class__(*coils, NFP=self.NFP, sym=self.sym, name=name)
 
+    def is_self_intersecting(self, grid=None, tol=None):
```
kianorr, line 1508: where is this function used? I see it's in the test, but is it used anywhere in the coils code?

desc/coils.py
```diff
@@ -1498,6 +1505,99 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
+            not rely on a ``tol`` parameter.
+
+        Returns
+        -------
+        is_self_intersecting : bool
+            Whether or not any coils in the CoilSet come close enough to eachother to
+            possibly be intersecting.
+
+        """
+        from desc.objectives._coils import CoilsetMinDistance
```
kianorr, line 1545: is there a circular import if you don't import locally?

desc/coils.py
dpanici: `maybe_bad_bool` is defined at each iteration inside of a for loop, so I need a second bool to check if any of the intermediate bools tripped as True

desc/coils.py
dpanici: Yes

desc/coils.py
dpanici: It is hard to do basedon coil resolution if the coiilset has a diverse set of coils

desc/coils.py
dpanici: the returned bool is up to the user, it is like is_nested()

It is called in from_symmetry simply to check if the coils are intersecting

desc/coils.py
dpanici: `from_symmetry` of `CoilSet`

desc/coils.py
dpanici: I assume so as `desc.objectives._coils` might import from `desc.coils`. I put it here to be safe

tests/test_coils.py
```diff
@@ -321,6 +321,30 @@ def test_from_symmetry(self):
...
         np.testing.assert_allclose(B_true, B_approx, rtol=1e-3, atol=1e-10)
 
+    @pytest.mark.unit
+    def test_from_symmetry_warnings(self):
+        """Test warning in from_symmetry for symmetry plane and self-intersection."""
+        N = 40
+        # test warning after crossing symmetry plane
+        # tilt coils so they cross the symmetry plane
+        coil = FourierPlanarCoil(normal=[1e-4, 1, 3])
+        coils_list = [coil] + [coil.copy() for i in range(N // 4 - 1)]
```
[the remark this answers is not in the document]
dpanici: good catch yes, I removed it

desc/coils.py
dpanici: I refactored so that I only use one bool now, it should look cleaner

dpanici:
Just realized this wont work with  multiple coilsets in a MixedCoilSet

```python
from desc.coils import *
c1 = CoilSet.from_symmetry(FourierPlanarCoil(),NFP=3)
c2 = CoilSet.from_symmetry(FourierPlanarCoil(),NFP=3)
c = MixedCoilSet([c1,c2])
# this wont work
c.is_self_intersecting()
```

f0uriest:
> Just realized this wont work with a MixedCoilSet, unless you pass in a `grid`
> 
> ```python
> from desc.coils import *
> from desc.grid import LinearGrid
> c1 = CoilSet.from_symmetry(FourierPlanarCoil(),NFP=3)
> c2 = CoilSet.from_symmetry(FourierPlanarCoil(),NFP=3)
> c = MixedCoilSet([c1,c2])
> # this wont work
> c.is_self_intersecting()
> # this does
> c.is_self_intersecting(grid=LinearGrid(N=30))
> ```

why doesn't this work? Is it something wrong with the default grid?
Could probably just use a single default grid with high resolution, assuming coils are usually N~10, with grid N~100 should be fine for most cases.

dpanici:
> > Just realized this wont work with a MixedCoilSet, unless you pass in a `grid`
> > ```python
> > from desc.coils import *
> > from desc.grid import LinearGrid
> > c1 = CoilSet.from_symmetry(FourierPlanarCoil(),NFP=3)
> > c2 = CoilSet.from_symmetry(FourierPlanarCoil(),NFP=3)
> > c = MixedCoilSet([c1,c2])
> > # this wont work
> > c.is_self_intersecting()
> > # this does
> > c.is_self_intersecting(grid=LinearGrid(N=30))
> > ```
> 
> why doesn't this work? Is it something wrong with the default grid? Could probably just use a single default grid with high resolution, assuming coils are usually N~10, with grid N~100 should be fine for most cases.

```python
Cell In[5], line 1
----> 1 c.is_self_intersecting()

File ~/Research/DESC/desc/coils.py:1559, in CoilSet.is_self_intersecting(self, grid, tol)
   1556     return is_nearly_intersecting
   1557 else:
-> 1559     pts = obj._constants["coilset"]._compute_position(
   1560         params=self.params_dict, grid=obj._constants["grid"]
   1561     )
   1562     pts = np.array(pts)
   1563     num_nodes = pts.shape[1]

File ~/Research/DESC/desc/coils.py:1733, in MixedCoilSet._compute_position(self, params, grid, **kwargs)
   1715 def _compute_position(self, params=None, grid=None, **kwargs):
   1716     """Compute coil positions accounting for stellarator symmetry.
   1717 
   1718     Parameters
   (...)
   1731 
   1732     """
-> 1733     errorif(
   1734         grid is None,
   1735         ValueError,
   1736         "grid must be supplied to MixedCoilSet._compute_position, since the "
   1737         + "default grid for each coil could have a different number of nodes.",
   1738     )
   1739     params = self._make_arraylike(params)
   1740     grid = self._make_arraylike(grid)

File ~/Research/DESC/desc/utils.py:535, in errorif(cond, err, msg)
    529 """Raise an error if condition is met.
    530 
    531 Similar to assert but allows wider range of Error types, rather than
    532 just AssertionError.
    533 """
    534 if cond:
--> 535     raise err(msg)

ValueError: grid must be supplied to MixedCoilSet._compute_position, since the default grid for each coil could have a different number of nodes.
```

dpanici:
need to add catches in the tests to all the spots we add coilsets and the new coiilset is not nested

desc/coils.py
```diff
@@ -798,13 +798,27 @@ class CoilSet(OptimizableCollection, _Coil, MutableSequence):
         assuming 'virtual' coils from the other half field period. Default = False.
     name : str
         Name of this CoilSet.
-
+    check_intersection: bool
+        Whether or not to check the coils in the coilset for intersections.
+    check_grid: Grid, optional
```
f0uriest, line 803: Do we need this? I feel like we can use the default grid for the automatic checking. For comparison, eq.is_nested allows custom grids, but the automatic check that happens as part of init always uses the default

desc/coils.py
```diff
@@ -1540,6 +1553,7 @@ def is_self_intersecting(self, grid=None, tol=None):
         """
         from desc.objectives._coils import CoilsetMinDistance
 
+        errorif(grid is None and isinstance(self, MixedCoilSet))
```
f0uriest, line 1556: any reason we can't just always default to the N=100 grid?

desc/coils.py
```diff
@@ -1498,6 +1514,101 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
+                UserWarning,
+                "Found coils which are nearly intersecting according to the given tol "
+                + "(min coil-coil distance = "
+                + f"{np.min(min_dists):1.3e} m < {tol:1.3e} m)"
+                + " in the coilset, it is recommended to check coils closely.",
+            )
+            return is_nearly_intersecting
+        else:
+
+            pts = obj._constants["coilset"]._compute_position(
```
f0uriest, line 1573: I think you need to pass `basis="xyz"` here, and also needs to be fixed in `CoilsetMinDistance.compute`

dpanici: Good catch

dpanici: by default it is xyz but I will pass it in anyways for clarity's sake

desc/coils.py
```diff
@@ -1172,18 +1185,14 @@ def from_symmetry(cls, coils, NFP=1, sym=False):
...
-            )
-
+        check_for_intersection = False
+        for cdata in data:
+            # check if any coils lie on symmetry planes, these may cause
+            # self-intersection after reflection/rotation
+            check_for_intersection = (
+                np.any(cdata["phi"] > maxphi)
+                or (sym and np.any(cdata["phi"] < np.finfo(cdata["phi"].dtype).eps))
+            ) or check_for_intersection
```
ddudt, line 1195: If you want to get fancy (and I think is more readable code): 
```
check_intersection = any(
    [
        np.any(cdata["phi"] > maxphi)
        or (sym and np.any(cdata["phi"] < np.finfo(cdata["phi"].dtype).eps))
        for cdata in data
    ]
)
```

desc/coils.py
```diff
@@ -1205,7 +1214,7 @@ def from_symmetry(cls, coils, NFP=1, sym=False):
             rotated_coils.rotate(axis=[0, 0, 1], angle=2 * jnp.pi * k / NFP)
             coilset += rotated_coils
 
-        return cls(*coilset)
+        return cls(*coilset, check_intersection=True)
```
ddudt, line 1217: This should be `check_intersection=check_intersection`, otherwise there is no point in checking if any coils lie on the symmetry plane. But as of now, `check_intersection=True` is the default in `CoilSet.__init__` so this isn't actually necessary anyways...

desc/coils.py
ddudt: See comment below. We probably don't actually need this check at all if we always check for intersections in `CoilSet.__init__` by default anyways.

desc/coils.py
ddudt: It's actually used in `CoilSet.__init__`

tests/test_coils.py
```diff
@@ -321,11 +321,30 @@ def test_from_symmetry(self):
...
+
+        for i, c in enumerate(coils_list_sym[1:]):
+            c.rotate(angle=2 * np.pi / N * (i + 1))
+
+        # test the warning for self-intersecting coils, as two
+        #  of the coils in each field period lie nearly
+        # in the same physical space (intersecting at 2 points) after reflection
+        with pytest.warns(UserWarning) as warninfo:
+            _ = CoilSet.from_symmetry(coils_list_sym, NFP=4, sym=True)
+        assert "nearly intersecting" in str(warninfo[0].message)
```
ddudt, line 341: Just a note on the documentation:

This test should really be called `test_is_self_intersecting_warning` since that is where the warning comes from, not the `from_symmetry` method. And you can remove the first comment that says "test warning after crossing symmetry plane" since that doesn't actually raise a warning anymore.

tests/test_coils.py
```diff
@@ -982,8 +1002,8 @@ def test_load_makegrid_coils_header_asserts(tmpdir_factory):
 @pytest.mark.unit
 def test_repr():
     """Test string representation of Coil objects."""
-    coil = FourierRZCoil()
-    assert "FourierRZCoil" in str(coil)
+    coil = FourierPlanarCoil()
+    assert "FourierPlanarCoil" in str(coil)
```
ddudt, line 1006: Any particular reason for this change? It's fine, I'm just curious.

desc/coils.py
dpanici: True

tests/test_coils.py
dpanici: It avoids a self-intersection later in the test when it forms a coilset

f0uriest (commented):
I'm a little worried that we have to add `check_intersection=False` for so many tests. Are those coilsets actually self intersecting? or are we getting false positives?

dpanici:
> I'm a little worried that we have to add `check_intersection=False` for so many tests. Are those coilsets actually self intersecting? or are we getting false positives?

The ones I checked actually were self intersecting but I didn't check every one. Our default linear linspace moves things in the x direction and if we use it with a Fourier planar or Fourier RZ coil, that'll almost always result in self intersections

ddudt:
> > I'm a little worried that we have to add `check_intersection=False` for so many tests. Are those coilsets actually self intersecting? or are we getting false positives?
> 
> The ones I checked actually were self intersecting but I didn't check every one. Our default linear linspace moves things in the x direction and if we use it with a Fourier planar or Fourier RZ coil, that'll almost always result in self intersections

Also the `DummyMixedCoilSet` in conftest.py is self-intersecting, but that was intentional when I made it.

f0uriest:
OK, I checked them all and they are actually self-intersecting (in many cases totally overlapping). In the future we may want to fix that but it's fine for now.

f0uriest (approved):

ddudt (approved):

PR #1114, opened 2024-07-08
Add Stage-Two Coil Optimization Tutorial

dpanici:
- Add tutorial notebook for stage-two filamentary coil optimization


dpanici:
Sorry this was pretty WIP, I can add FourierPlanar etc to it. tomorrow. Just wanted to put up what I had in case anyone else wanted to work on it

dpanici:
maybe use 
```python
import plotly.io as pio
pio.renderers.default = "jupyterlab+png"
```
to fix 3d plot render issues

f0uriest:
Updated the examples, they now seem to run well and give pretty decent results. I switched the first one to use regular least squares instead of augmented lagrangian because auglag seems kind of finicky for these problems where the coil currents are way larger in magnitude than the geometric dofs. 

Could maybe use some additional test/math description but I think the code is OK.

kianorr:
I think I'm still confused on how $\left< B_n \right>$ is defined. I see in [Wiedman 2024](https://arxiv.org/abs/2311.16386) (eq. 2.10), that they define it as

$$
\frac{\left<|\mathbf{B}\_{\text{coils}} \cdot \mathbf{\hat{n}\_{\text{target}}}|\right>}{\left<|\mathbf{B}\_{\text{target}}|\right>}.
$$

But I think in this tutorial we're using

$$
\frac{\left<|\mathbf{B}\_{\text{coils}} \cdot \mathbf{\hat{n}\_{\text{target}}}|\right>}{\left<|\mathbf{B}\_{\text{coils}}|\right>}.
$$

Is this situation dependent or am I misunderstanding something?

f0uriest:
"B normal" is pretty much always defined in terms of the total field B_total = B_coil + B_plasma + B_whatever. We need the normal component of that to vanish. If it's a vacuum equilibrium, than B_total == B_coil, which I think is the case considered in Wiedman's paper.

The other question is how its normalized. Again this is usually normalized by the magnitude of the total field |B_total|, but it's a bit trickier in vacuum.

For a vacuum field, assuming you have B*n = 0, then you can rescale the coil currents such that B_coil -> 100 B_coil and B*n will still be zero, even if now the tangent part of B_coil doesn't match with the tangent field from the fixed boundary equilibrium (effectively you need to scale eq.Psi along with the coil currents). 

What this means in practice is that if you're doing vacuum coil design and fixing the null space using the `ToroidalFlux` objective then everything should match, and you can normalize by either B_coil or B_equilibrium and they should be the same. However, if you fix the null space by fixing a single coil current, then you should normalize by the total coil field, not the equilibrium field which may be different.

f0uriest:
A few more fixes I think:

- [ ] Rename to just "Filamentary Coil Optimization"? (at least remove the "in DESC" part)
- [ ] Fix definition of normalized normal field error metric, and calculation thereof (see previous comment, should be B_total*n / |B_total|)
- [ ] Figure out why 3d plots aren't showing - they seem to work fine in the "Using DESC outputs" notebook
- [ ] For sub-headings, might be better to focus on "vacuum" vs "finite beta" rather than precise QA vs w7x

f0uriest (dismissed):

ddudt (changes requested):
We should change this PR to point to `dd/curvature` instead of `master`, because the bug fixes and changes on that branch are going to break this tutorial. And so we should merge #1085 before this.

kianorr (approved):

f0uriest (approved):

ddudt (changes requested):
See comments on ReviewNB

f0uriest:
<p>I think this is just an artifact of reviewnb, it shows correctly on RTD: <a href="https://desc-docs--1114.org.readthedocs.build/en/1114/notebooks/tutorials/coil_stage_two_optimization.html" target="_blank">https://desc-docs--1114.org.readthedocs.build/en/1114/notebooks/tutorials/coil_stage_two_optimization.html</a></p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233879118'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

f0uriest:
<p>Pretty sure they can both be non-sym with NFP&gt;1. The logic in coils.compute_magnetic_field and coils._compute_position should account correctly for multiple field periods. We can safely do <code>all coils -&gt; 1 plasma period</code> or <code>1 period coils -&gt; all plasma periods</code> but don't need both. </p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233879179'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

f0uriest:
<p>This is a vacuum problem so the actual scale of the currents/field doesn't matter, fixing the current in a single coil is a pretty common way to avoid the trivial solution (i believe the default in simsopt and focus)</p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233879482'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

f0uriest:
<p>I think a lot of the linting ignores things in the docs folder by default. We can change that if needed, I remember it causing a major headache though because it tried to reformat some of the sphinx generated stuff as well so we'd need to figure it out. For now I'll manually run black on this.</p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233879059'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ddudt:
<p>Ah ok. It just feels confusing to pass in a surface as a hack to avoid computing the plasma field, I think a vacuum flag or something would make the code easier to understand. </p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233878653'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

f0uriest:
<p>See previous response, I don't think that actually simplifies anything. I can add a  <code>vacuum=True</code> thing instead. </p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233879533'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ddudt:
<p>I still think we need <code>sym=False</code> for the <code>PlasmaCoilSetMinDistance</code> grid. If the closest point between the plasma and coils was on the underside of the torus (theta &gt; pi) then a symmetric grid won't capture that. Whatever we decide here, we should update the default grids for those two objectives. </p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233879179'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ddudt:
<p>Fixed where? I don't see a change to the code and if the bound is given in meters this needs to be set to True. Maybe for coil length we can use a normalized target, but for other coil objectives like plasma-coil distance I think it's much more common to use real engineering units, since you need to fit a blanket of a certain size, etc. </p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233879444'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

dpanici:
<p>vacuum flag I support</p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233878653'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

dpanici:
<p>set default to eq.sym for grid of plasmacoilset etc</p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1114/discussion/#comment-2233879179'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ddudt (commented):
See two minor comments on ReviewNB. Main thing is I think we should add a brief explanation of negative/positive curvature for concave/convex, since you include negative curvature in the optimization.

dpanici:
> See two minor comments on ReviewNB. Main thing is I think we should add a brief explanation of negative/positive curvature for concave/convex, since you include negative curvature in the optimization. 

I won't be free for another couple hours so whoever is can make the change

f0uriest (approved):

ddudt (approved):

PR #1115, opened 2024-07-09
Add poincare plotting function

f0uriest:
I've noticed several notebooks etc where we make poincare plots from external fields, this PR adds a specific plotting function to do that, since some of the logic to make all the points line up correctly is non-trivial.


dpanici (approved):

desc/plotting.py
```diff
@@ -1724,6 +1726,156 @@ def plot_surfaces(eq, rho=8, theta=8, phi=None, ax=None, return_data=False, **kw
     return fig, ax
 
 
+def poincare_plot(
```
kianorr, line 1729: should we add a warning if there is a coil going through the plasma? Because I think there's no reason to field line trace if that's happening and it's taken me a while to figure out before (but maybe a separate issue)

tests/baseline/test_plot_poincare.png
kianorr, line 1: Will the function plot the field trace as well as the target field?

f0uriest: no, you'd need to plot that separately to the same axes, eg
```python
fig, ax = plot_surfaces(eq)
fig, ax = poincare_plot(coils, ..., ax=ax)

kianorr (approved):

PR #1116, opened 2024-07-09
`from_values` method for `FourierRZCoil` and `FourierPlanarCoil`

### LINKED ISSUE #1097: Add `FourierPlanarCurve.from_values` Method

dpanici:
- adds `from_values` method that was present in `FourierRZCurve` but missing in `FourierRZCoil`
- also adds new `from_values` method for `FourierPlanarCurve` and `FourierPlanarCoil`

Resolves #1097


f0uriest (dismissed):

desc/geometry/curve.py
```diff
@@ -740,10 +771,82 @@ def compute(
...
+
+        # normal
+        idx = coords.shape[0] // 2
+        normal = np.cross(coords[0, :], coords[idx, :])  # 2 arbitrary points on curve
+        normal = normal / np.linalg.norm(normal)
+
+        # axis and angle of rotation
+        Z_axis = np.array([0, 0, 1])
+        axis = np.cross(Z_axis, normal)
+        axis = axis / np.linalg.norm(axis)
```
f0uriest, line 817: what if the normal is already along the Z axis? (eg, converting a FourierRZCurve to FourierPlanarCurve)

desc/geometry/curve.py
```diff
@@ -740,10 +771,82 @@ def compute(
...
+        angle = np.arccos(np.dot(Z_axis, normal))
+        rotmat = rotation_matrix(axis, angle)
+        coords = coords @ rotmat  # rotate to X-Y plane
+        warnif(
+            np.max(np.abs(coords[:, 2])) > 1e-14,  # check that Z=0 for all points
+            UserWarning,
+            "Curve values are not planar!",
```
f0uriest, line 824: Might want to give more info here, ie it still tries a least squares fit? or it takes the projection of the curve into a plane?

desc/geometry/curve.py
```diff
@@ -740,10 +771,82 @@ def compute(
...
+            coords = rpz2xyz(coords)
+        coords = np.atleast_2d(coords)
+
+        # center
+        center = np.mean(coords, axis=0)
+        coords = coords - center  # shift to origin
+
+        # normal
+        idx = coords.shape[0] // 2
+        normal = np.cross(coords[0, :], coords[idx, :])  # 2 arbitrary points on curve
```
f0uriest, line 811: Could compute all the normals and take the average which might be better if the curve isn't perfectly planar

desc/geometry/curve.py
ddudt: In that case `axis = [0, 0, 0]` and then the logic in `rotation_matrix` already makes sure that there will be no rotation (`rotmat` will be the identity matrix).

desc/geometry/curve.py
ddudt: It takes the projection into a plane. And now that I made your other requested change about using the average normal vector, that plane is the "best fit" in some sense.

desc/geometry/curve.py
```diff
@@ -740,10 +771,86 @@ def compute(
...
+        coords = np.atleast_2d(coords)
+
+        # center
+        center = np.mean(coords, axis=0)
+        coords = coords - center  # shift to origin
+
+        # normal
+        normal = np.empty((0, 3))  # avg of normals btw all pts in case it is non-planar
+        for idx in range(1, coords.shape[0]):
+            norm = np.cross(coords[0, :], coords[idx, :])
```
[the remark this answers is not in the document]
ddudt: Thanks for finding this, it's definitely simpler and better.

desc/geometry/curve.py
```diff
@@ -740,10 +771,86 @@ def compute(
...
+            norm = norm / np.linalg.norm(norm)
+            sgn = sign(-int(all(sign(norm) != sign(normal[0, :])))) if idx > 1 else 1
```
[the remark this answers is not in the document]
ddudt: This logic was to handle the fact that there are two normal vectors that differ by a sign convention but represent the same physical plane. Irrelevant now that I changed it to use the SVD instead.

tests/test_coils.py
```diff
@@ -204,13 +204,27 @@ def test_adding_MagneticField_with_Coil_or_CoilSet(self):
     def test_converting_coil_types(self):
         """Test conversions between coil representations."""
         s = np.linspace(0, 2 * np.pi, 100, endpoint=False)
-        coil1 = FourierRZCoil(1e6, [0, 10, 2], [-2, 0, 0])
+        coil1 = FourierRZCoil(1e6, [0, 10, 0], [0, 0, 0])
```
dpanici, line 207: The `FourierPlanarCoil` fails to fit the `FourierRZCoil` well if I make the n=1 component of it anything but zero... this seems like something it should be able to do with sufficient resolution? Maybe I messed something up in my latest commit on the fit?

dpanici:
Just leaving here for a record of the sensitivity, maybe due to the SVD?

```python
s = np.linspace(0, 2 * np.pi, 100, endpoint=False)
grid = LinearGrid(zeta=s)
coil1 = FourierRZCoil(1e6, [0, 10, 1], [0, 0, 0])
diffs = []
Ns = np.arange(1,33,3)
for N in Ns:
    coil5 = coil1.to_FourierPlanar(N=N)

    x1 = coil1.compute("x", grid=grid, basis="xyz")["x"]
    # find the arctan angle corresponding to the above pts for the
    # planar coil
    arctan_angle=np.arctan2(x1[:, 1] - coil5.center[1], x1[:, 0] - coil5.center[0])
    x5 = coil5.compute(
        "x",
        grid=LinearGrid(zeta=arctan_angle),
        basis="xyz",
    )["x"]
    diffs.append(np.mean(np.linalg.norm(x1-x5,axis=1)))
plt.figure()
plt.scatter(Ns,diffs)
plt.xlabel("Fourier Resolution of Fit (Default Grids)")
plt.ylabel("Mean error in position of coil vs original")
```
image

```python
s = np.linspace(0, 2 * np.pi, 100, endpoint=False)
grid = LinearGrid(zeta=s)
coil1 = FourierRZCoil(1e6, [0, 10, 1], [0, 0, 0])
diffs = []
Ns = np.arange(1,33,3)
for N in Ns:
    coil5 = coil1.to_FourierPlanar(N=N,grid=LinearGrid(2*N+1, endpoint=True))

    x1 = coil1.compute("x", grid=grid, basis="xyz")["x"]
    # find the arctan angle corresponding to the above pts for the
    # planar coil
    arctan_angle=np.arctan2(x1[:, 1] - coil5.center[1], x1[:, 0] - coil5.center[0])
    x5 = coil5.compute(
        "x",
        grid=LinearGrid(zeta=arctan_angle),
        basis="xyz",
    )["x"]
    diffs.append(np.mean(np.linalg.norm(x1-x5,axis=1)))
plt.figure()
plt.scatter(Ns,diffs)
plt.xlabel("Fourier Resolution of Fit (Default Grids, endpoint=True)")
plt.ylabel("Mean error in position of coil vs original")
```
image


```python
s = np.linspace(0, 2 * np.pi, 100, endpoint=False)
grid = LinearGrid(zeta=s)
coil1 = FourierRZCoil(1e6, [0, 10, 1], [0, 0, 0])
diffs = []
Ns = np.arange(1,33,3)
for N in Ns:
    coil5 = coil1.to_FourierPlanar(N=N,grid=grid)

    x1 = coil1.compute("x", grid=grid, basis="xyz")["x"]
    # find the arctan angle corresponding to the above pts for the
    # planar coil
    arctan_angle=np.arctan2(x1[:, 1] - coil5.center[1], x1[:, 0] - coil5.center[0])
    x5 = coil5.compute(
        "x",
        grid=LinearGrid(zeta=arctan_angle),
        basis="xyz",
    )["x"]
    diffs.append(np.mean(np.linalg.norm(x1-x5,axis=1)))
plt.figure()
plt.scatter(Ns,diffs)
plt.xlabel("Fourier Resolution of Fit (Fixed Grid)")
plt.ylabel("Mean error in position of coil vs original")
```
image

```python
s = np.linspace(0, 2 * np.pi, 100, endpoint=True)
grid = LinearGrid(zeta=s)
coil1 = FourierRZCoil(1e6, [0, 10, 1], [0, 0, 0])
diffs = []
Ns = np.arange(1,33,3)
for N in Ns:
    coil5 = coil1.to_FourierPlanar(N=N,grid=grid)

    x1 = coil1.compute("x", grid=grid, basis="xyz")["x"]
    # find the arctan angle corresponding to the above pts for the
    # planar coil
    arctan_angle=np.arctan2(x1[:, 1] - coil5.center[1], x1[:, 0] - coil5.center[0])
    x5 = coil5.compute(
        "x",
        grid=LinearGrid(zeta=arctan_angle),
        basis="xyz",
    )["x"]
    diffs.append(np.mean(np.linalg.norm(x1-x5,axis=1)))
plt.figure()
plt.scatter(Ns,diffs)
plt.xlabel("Fourier Resolution of Fit (Fixed Grid, endpoint=True)")
plt.ylabel("Mean error in position of coil vs original")
```
image

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+New Features
+
+- adds from_values method that was present in FourierRZCurve but missing in FourierRZCoil
```
f0uriest, line 6: use backticks for code literals.

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
+- also adds new from_values method for FourierPlanarCurve and FourierPlanarCoil
+
+Minor changes
+
+-  modifies FourierPlanarCurve compute implementation to check if normal is already parallel to the Zaxis, in which case no rotation is needed.
```
f0uriest, line 11: don't think we need to include this, it shouldn't affect any user stuff

desc/coils.py
```diff
@@ -345,14 +345,81 @@ def to_SplineXYZ(self, knots=None, grid=None, method="cubic", name=""):
...
+
+        Returns
+        -------
+        curve : FourierRZCoil
+            New representation of the coil parameterized by Fourier series for R,Z.
+
+        """
+        NFP = 1 or NFP
+        if grid is None:
+            grid = LinearGrid(N=2 * N + 1, NFP=NFP)
```
f0uriest, line 377: I think we want NFP=1 here, since usually the starting curve doesn't have discrete symmetry, so we need to sample the full thing

f0uriest (changes requested):

f0uriest:
@dpanici I think it's fine, it's just different curve parameter definitions. Plotting them all they're on top of each other:

image

desc/coils.py
dpanici: I think it is possible for the coil to have an NFP symmetry though even if it does not have the NFP attribute (like from a MAKEGRID file or something), I will remove NFP from the grid but allow it to still be passed in to the from_values call

f0uriest (dismissed):

ddudt (dismissed):

f0uriest (approved):

ddudt (approved):

PR #1124, opened 2024-07-11
Document basis for input data

unalmis:


unalmis:
feel free to merge after approval

desc/compute/utils.py
```diff
@@ -121,6 +123,9 @@ def _compute(
 ):
     """Same as above but without checking inputs for faster recursion.
 
+    All vectors v = [vᴿ, v^ϕ, vᶻ] in ``data`` should be given in contravariant
```
ddudt, line 126: Can we get the phi superscript to work?

ddudt: Also in all the other new docstrings

f0uriest: I don't think unicode has a superscript phi.

Another small nit: cylindrical coordinates are orthogonal, so the co- and contra-variant bases are the same

unalmis: I couldn't find superscript phi either.

I think that contravariant = covariant only if the coordinate system is orthonormal. Orthogonal is not sufficient as that only implies direction of basis vectors are same and not their magnitudes. (I think for cylindrical the phi component will differ by factors of R).

f0uriest: hmm i think you're right, nvm

f0uriest (dismissed):

desc/compute/utils.py
ddudt: On the topic of our cylindrical coordinates not being orthonormal, does that cause any problems when we take cross-products in those coordinates with our `cross` util function? I'm trying to track down some more bugs with the Frenet-Serret frame and was wondering about this

unalmis: Our cross and dot product utilities only work in orthonormal (and right-handed for cross) coordinate systems, so they would not work in cylindrical coordinates which are orthogonal but not orthonormal.

However, I now recall that the tuple of values we return when we request a vector quantity  is neither the contravariant components $[v^R, v^\phi, v^Z]$ nor the covariant components $[v_R, v_\phi, v_Z]$ of the cylindrical coordinate system $R, \phi, Z$.

Instead, we return the coordinates $[v^1, v^2, v^3]$ of a different, orthonormal, coordinate system.

For a vector $v$ we have
$v = v^R e_R + v^\phi e_\phi + v^Z e_z = v_R e^R + v_\phi e^\phi + v_Z e^z = v^1 \hat{R} + v^2 \hat{\phi} + v^3 \hat{Z}$.

All three sets of basis vectors point in the same direction but the magntiudes obey $\Vert e_\phi \Vert = \Vert v^1 \hat{\phi} \Vert$ and  $\Vert e^\phi \Vert = \Vert (1/v^1) \hat{\phi} \Vert $. Hence, the $\phi$ components will differ by the inverse of these factors.

Books like dhaeseleer's and frankel's will refer to components of a vector in the cylindrical coordinate system $R, \phi, Z$ as $[v^R, v^\phi, v^Z]$, while most of wikipedia will use $[v^1, v^2, v^3]$.

unalmis: As an example of how this could cause issues, note that here https://en.wikipedia.org/wiki/Gradient, the quantity $\partial f^i / \partial x_j$ in  $\nabla f$ is differentiating the contravariant components $f^R, f^\phi, f^Z$ with respect to $R, \phi, Z$ and not the components $f^1, f^2, f^3$.

f0uriest (approved):

PR #1130, opened 2024-07-16
`FixSumCoilCurrent` objective

### LINKED ISSUE #1111: Make Linear Objective to Constrain Sum of Coil Currents
Instead of using `ToroidalFlux` or fixing a single coil current to avoid the trivial solution of zero current in a stage two vacuum coil optimization, you can also calculate the line integral of the covariant zeta component of the field, which is related to the sum of coil currents through an amperean loop as

$$
\begin{align}
     \oint_{\partial A} \mathbf{B} \cdot d\mathbf{l}&= I_{pol}= \mu_0 \sum_{i=1}^{N_{coils}} I_{i}^{coil} N_i^{poloidal~ transits}\\
     &\text{$\mathbf{B} \cdot d\mathbf{l} = \mathbf{B}\cdot \frac{\partial \mathbf{r}}{\partial \zeta} d\zeta = B_{\zeta} d\zeta$, LHS becomes}\\
     \Aboxed{N_{FP}\int_{0}^{2\pi/N_{FP}} B_{\zeta}d\zeta &= \mu_0 \sum_{i=1}^{N_{coils}} I_{i}^{coil} N_i^{poloidal~transits}}
\end{align}
$$

Where $N_i^{poloidal~transits}$ is the number of poloidal transits a coil makes (this is one for modular coils and equal to the winding number for helical coils). This constraint and a good quadratic flux optimization ensures the coils exactly create the correct magnetic field magnitude. Of course this is not strictly needed for vacuum optimization but is a 3rd way of avoiding the trivial solution

### LINKED ISSUE #1131: Test new functionality for `LinearObjectiveFromUser`
The new functionality is not tested in this PR. But there are existing tests for `LinearObjectiveFromUser` that still pass.

_Originally posted by @ddudt in https://github.com/PlasmaControl/DESC/pull/1085#discussion_r1680104716_

ddudt:
Resolves #1111 
Resolves #1131 

@dpanici I could use your help adding an example to the docs of how to set the target based on an equilibrium. There is the complication that coils with all positive/negative currents does not mean all of those currents are in the same physical direction, since the orientation of the coil geometries can differ. That is difficult to check in general, so we might need to simply warn the users to check their coil orientations themselves.


dpanici:
> Resolves #1111
> 
> @dpanici I could use your help adding an example to the docs of how to set the target based on an equilibrium. There is the complication that coils with all positive/negative currents does not mean all of those currents are in the same physical direction, since the orientation of the coil geometries can differ. That is difficult to check in general, so we might need to simply warn the users to check their coil orientations themselves.

coil orientations is definitely not something I had thought of before, good catch... maybe we can just make a note in the docstring of what convention we use (positive being CCW or CW if facing the phi direction). I will use the orientation from my regcoil stuff

desc/objectives/linear_objectives.py
```diff
@@ -2726,6 +2726,141 @@ def build(self, use_jit=False, verbose=1):
...
+class FixSumCoilCurrent(FixCoilCurrent):
+    """Fixes the sum of coil current(s) in a Coil or CoilSet.
+
+    Parameters
+    ----------
+    coil : Coil
+        Coil(s) that will be optimized to satisfy the Objective.
+    target : {float, ndarray}, optional
+        Target value(s) of the objective. Only used if bounds is None.
+        Must be broadcastable to Objective.dim_f. Default is FIXME.
```
dpanici, line 2738: Can default it to whatever the initial sum is

desc/objectives/linear_objectives.py
```diff
@@ -2726,6 +2726,141 @@ def build(self, use_jit=False, verbose=1):
...
+    bounds : tuple of {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
+        Both bounds must be broadcastable to to Objective.dim_f.
+        Default is FIXME.
```
dpanici, line 2742: default would be to use the target of whatever the current sum is

desc/objectives/linear_objectives.py
```diff
@@ -2726,6 +2726,141 @@ def build(self, use_jit=False, verbose=1):
...
+        vf_coil = FourierRZCoil(current=-1, R_n=3, Z_n=-1)
+        vf_coilset = CoilSet.linspaced_linear(
+            vf_coil, displacement=[0, 0, 2], n=3, endpoint=True
+        )
+        # another single coil
+        xyz_coil = FourierXYZCoil(current=2)
+        # full coil set with TF coils, VF coils, and other single coil
+        full_coilset = MixedCoilSet((tf_coilset, vf_coilset, xyz_coil))
+
+        # fix the current of the 1st & 3rd TF coil
```
dpanici, line 2784: fix the *sum of* currents

ddudt:
> > Resolves #1111
> > @dpanici I could use your help adding an example to the docs of how to set the target based on an equilibrium. There is the complication that coils with all positive/negative currents does not mean all of those currents are in the same physical direction, since the orientation of the coil geometries can differ. That is difficult to check in general, so we might need to simply warn the users to check their coil orientations themselves.
> 
> coil orientations is definitely not something I had thought of before, good catch... maybe we can just make a note in the docstring of what convention we use (positive being CCW or CW if facing the phi direction). I will use the orientation from my regcoil stuff

I don't even think we have a set convention for the sign of the current. Positive current means it is flowing in the direction of increasing curve parameter `s`, but that could be different depending on the curve type and orientation. 

I think we should give an example showing how to set the target assuming all of the coil currents are flowing up/down through the center of the torus (or something like that). And then say it is the responsibility of the user to check the orientation of their coils to match that.

dpanici:
> > > Resolves #1111
> > > @dpanici I could use your help adding an example to the docs of how to set the target based on an equilibrium. There is the complication that coils with all positive/negative currents does not mean all of those currents are in the same physical direction, since the orientation of the coil geometries can differ. That is difficult to check in general, so we might need to simply warn the users to check their coil orientations themselves.
> > 
> > 
> > coil orientations is definitely not something I had thought of before, good catch... maybe we can just make a note in the docstring of what convention we use (positive being CCW or CW if facing the phi direction). I will use the orientation from my regcoil stuff
> 
> I don't even think we have a set convention for the sign of the current. Positive current means it is flowing in the direction of increasing curve parameter `s`, but that could be different depending on the curve type and orientation.
> 
> I think we should give an example showing how to set the target assuming all of the coil currents are flowing up/down through the center of the torus (or something like that). And then say it is the responsibility of the user to check the orientation of their coils to match that.

There is an implicit orientation from the equilibrium `Psi`, so if `Psi` is positive then I think that means that `B_zeta` is positive and then `G` (which is a line integral of `B_zeta` around `zeta` which is equal to the net enclosed net poloidal current inside of the torus hole) is positive. That implies that coil currents flowing  upwards through the torus hole are positive (I believe)

dpanici:
I think in general we can leave this onus onto the user, our objective here just fixes the coil currents, if you want to use it to do something physcially meaningful, you have to know what that means (i.e. that the coils must link poloidally the torus, that there is a sign convention). This objective could also be used to just constrain the coil current sum for some other engineering reason (like if they are in parallel or something and the input feed has a limit of how much current it can take), so we don't need to go overboard on warnings inside this objective about orientations.

We could update the tutorial about this but I'd rather not before the release, we already changed so much in it.

desc/objectives/linear_objectives.py
```diff
@@ -2726,6 +2726,147 @@ def build(self, use_jit=False, verbose=1):
...
+        )
+        # another single coil
+        xyz_coil = FourierXYZCoil(current=2)
+        # full coil set with TF coils, VF coils, and other single coil
+        full_coilset = MixedCoilSet((tf_coilset, vf_coilset, xyz_coil))
+
+        # equilibrium G(rho=1) determines the necessary poloidal current through
+        # the coils (as dictated by Ampere's law)
+        grid_at_surf = LinearGrid(rho=1.0, M=eq.M_grid, N=eq.N_grid)
+        G_tot = eq.compute("G", grid=grid_at_surf)["G"][0] / (mu_0 * 2 * jnp.pi)
```
ddudt, line 2793: @dpanici I added these parentheses during a merge conflict with you. Please confirm if that is what you meant, and if not I think it would be better to put the `2 * np.pi` factor at the beginning so it does not get confused as being part of the denominator.

desc/objectives/linear_objectives.py
```diff
@@ -2726,6 +2726,147 @@ def build(self, use_jit=False, verbose=1):
...
+
+        # only coils that link the equilibrium poloidally should be included in the sum,
+        # which is the TF coil set and the FourierXYZ coil, but not the VF coil set
+        obj = FixSumCoilCurrent(full_coilset, indices=[True, False, True], target=G_tot)
```
ddudt, line 2799: Should we add a note here to say this assumes the coils have positive currents when the currents are flowing up through the torus hole?

docs/api.rst
```diff
@@ -187,6 +187,8 @@ Objective Functions
     desc.objectives.FixParameters
     desc.objectives.FixPressure
     desc.objectives.FixPsi
+    desc.objectives.FixSumCoilCurrent
```
f0uriest, line 190: Should also add to `api_objectives.rst`

desc/objectives/linear_objectives.py
dpanici: it is times 2pi, I can add to the front

dpanici (approved):

f0uriest (approved):

PR #1132, opened 2024-07-17
`Curve` default grid `NFP=1`

### LINKED ISSUE #1118: Make default grid behavior for `FourierRZCurve` consistent with rest of objects that have `NFP` attibute
>Right but I mean now it is inconsistent behavior for NFP objects, where any other object with NFP as an attribute will compute on a grid from 0 -> 2pi/NFP except `FourierRZCoil/Curve`
>This is fine but I will make an issue for this as I don't like the inconsistency

_Originally posted by @dpanici in https://github.com/PlasmaControl/DESC/pull/1099#discussion_r1670879252_

ddudt:
Resolves #1118


desc/geometry/core.py
```diff
@@ -117,18 +118,14 @@ def compute(
...
-            NFP = self.NFP if hasattr(self, "NFP") else 1
-            grid = LinearGrid(N=grid, NFP=NFP, endpoint=False)
-        elif hasattr(grid, "NFP"):
-            NFP = grid.NFP
-        else:
-            raise TypeError(
-                "must pass in a Grid object or an integer for argument grid!"
-                f" instead got type {type(grid)}"
-            )
+            grid = LinearGrid(N=grid * getattr(self, "NFP", 1))
```
f0uriest, line 123: I think if a number is passed in we should use it as is, assuming the user knows what they're doing.

f0uriest (approved):

dpanici (approved):

PR #1134, opened 2024-07-18
read VMEC input ZAXIS as ZAXIS_CS

### LINKED ISSUE #1120: Issue reading axis from vmec input
VMEC input files specify the axis geometry coeffs usually as 
`raxis_cc = ...`
`zaxis_cs = ...`

But some only use
`raxis = ...`
`zaxis = ...`

If it's the latter, we read it incorrectly and assign `zaxis` to `zaxis_cc`. It looks like the VMEC parser was designed to handle the latter case, but we need to swap CC and CS somewhere around https://github.com/PlasmaControl/DESC/blob/986f529ac12b67ec2e3e8029af41d5e5eb363b7d/desc/input_reader.py#L1337

ddudt:
Resolves #1120


f0uriest (commented):
This looks like its swapping CC and CS, doesn't that break the expected parity and stuff? (the minus sign is in a different place?)
I think we just want to swap where the parens are?

dpanici (dismissed):

ddudt:
> This looks like its swapping CC and CS, doesn't that break the expected parity and stuff? (the minus sign is in a different place?) I think we just want to swap where the parens are?

Woops, my bad. Should be fixed now.

f0uriest (approved):

dpanici (approved):

PR #1137, opened 2024-07-20
Remove basis kwarg from _compute function calls in magnetic field classes

dpanici:
- Fixes some leftover basis args inside of `_compute` calls, which now throw an error since that function only returns `rpz` quantities

### DIFF
--- a/desc/coils.py
+++ b/desc/coils.py
@@ -248,13 +248,13 @@ class _Coil(_MagneticField, Optimizable, ABC):
         else:
             data = compute_fun(
                 self,
-                name=["x", "x_s", "ds"],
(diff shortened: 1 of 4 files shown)


f0uriest (approved):

ddudt (changes requested):
If this is wrong on master but wasn't being covered, we should add tests.

dpanici:
> If this is wrong on master but wasn't being covered, we should add tests.

Tests caught a bug, thanks for making me write them

f0uriest (approved):

ddudt (approved):

YigitElma (approved):

PR #1139, opened 2024-07-23
Remove `ds` compute quantity

### LINKED ISSUE #1093: `ds` compute quantity for `Curve` is not correct when `NFP` is not equal to `1`
`ds` is supposed to be the spacing between the points in `s` for a given grid, however `grid.spacing` for `zeta` is always `2pi/num_zeta`, so when `NFP` is not equal to 1, `ds` is not equal to the spacing between the grid points in `s`. This only matters for `FourierRZCurve`, but we should probably make a sub parameterization for `ds`(and maybe also `length`) for `FourierRZCurve` that accounts for `NFP`

dpanici:
Resolves #1093


dpanici:
based off wrong branch on accident

PR #1140, opened 2024-07-23
Change description of `ds` to be clear what it actually is

### LINKED ISSUE #1093: `ds` compute quantity for `Curve` is not correct when `NFP` is not equal to `1`

dpanici:
Resolves #1093


dpanici:
just change description of `ds`

desc/compute/_curve.py
```diff
@@ -31,7 +31,10 @@ def _s(params, transforms, profiles, data, **kwargs):
     label="ds",
     units="~",
     units_long="None",
-    description="Spacing of curve parameter",
+    description=(
+        "Quadrature weights for integration along the curve, "
```
dpanici, line 35: double space here

f0uriest (approved):

ddudt (approved):

PR #1141, opened 2024-07-23
Throw error for when `Curve` is used for coil objectives

### LINKED ISSUE #1020: Allow `CoilXXX` objectives also target curves, or add error if a curve is passed in
It would be nice to allow our coil objectives (like CoilLength, CoilCurvature) also allow for optimization of curve objects, or if that is not easily doable, to at least throw an error if a curve is passed in (like `errorif(not isinstance(things,Coil)` or something), as one could naively assume (like I) that they could optimize geometry of curves as well.

this is a test that shows the issue
```python
    c = FourierPlanarCurve()
    objective = ObjectiveFunction(CoilLength(c, target=11))
    np.testing.assert_allclose(c.compute("length")["length"], 11, atol=1e-3)
```

has an error when the objective is built
```python
desc/optimize/optimizer.py:217: in optimize
desc/objectives/objective_funs.py:173: in build
desc/objectives/_coils.py:337: in build
desc/objectives/_coils.py:153: in build
    self._grid = tree_map(
../../miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/tree_util.py:312: in tree_map
    return treedef.unflatten(f(*xs) for xs in zip(*all_leaves))
../../miniconda3/envs/desc-env/lib/python3.12/site-packages/jax/_src/tree_util.py:312: in <genexpr>
    return treedef.unflatten(f(*xs) for xs in zip(*all_leaves))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = array([0., 1., 0.])

    lambda x: LinearGrid(
>       N=2 * x.N + 5, NFP=getattr(x, "NFP", 1), endpoint=False
    ),
    self.things[0],
    is_leaf=lambda x: is_single_coil(x),
E   AttributeError: 'numpy.ndarray' object has no attribute 'N'

desc/objectives/_coils.py:155: AttributeError
```

kianorr:
Resolves #1020 

The given example in #1020 uses a `FourierPlanarCurve` which actually results in a list of numpy arrays being created after going through `tree_flatten`. This PR gives a better error message, but I'm not sure if we can explicitly say that a `Curve` was inputted.

If there is a `Curve` object in a `CoilSet` then there are already checks in the `CoilSet` class that throw an error.


f0uriest:
could do something like:
```python
for coil in coils:
    errorif(not isinstance(coil, _Coil),
            TypeError,
            f"Expected object of type Coil, got {type(coil)}",
)
```

that way it actually says what the offending member is.

dpanici (changes requested):
add a test

f0uriest (approved):

dpanici (approved):

PR #1142, opened 2024-07-23
add knots to attrs

### LINKED ISSUE #1136: `CoilSet` of `SplineXYZCoil` with different knots can give wrong values
I think there might be an issue with `CoilSet`s if there are spline curves that have different breakpoints. We only vmap over params, not transforms, since its assumed that every coil in a `CoilSet` can use the same transforms. I think this is also an issue if they have different knots after #1099 

Possible fixes:
- move `knots` and `intervals` back to `params` instead of `transforms` (not sure if this will work, what was the reason they were moved in the first place?)
- Add logic to `desc.coils._check_type` to ensure that `knots` and `intervals` are the same for all `SplineXYZCoil` in a `CoilSet`

_Originally posted by @f0uriest in https://github.com/PlasmaControl/DESC/pull/970#pullrequestreview-2189083297_
            

Confirmed this is an issue on master:

```python
coil = FourierPlanarCoil(1e6)
coil1 = coil.to_SplineXYZ(knots=2*np.pi*np.linspace(0,1,5))
coil2 = coil.to_SplineXYZ(knots=2*np.pi*np.linspace(0,1,5)**4)
coilset = CoilSet(coil1, coil2)
B = coilset.compute_magnetic_field(np.array([10, 1, 1]))
B12 = coil1.compute_magnetic_field(np.array([10, 1, 1])) + coil2.compute_magnetic_field(np.array([10, 1, 1]))
np.testing.assert_allclose(B,B12)

kianorr:
Resolves #1136 

Adds `knots` to `attrs` in `_check_type` such that there is an error if the knots are different between different coils. This way, the user has to use a `MixedCoilSet`

### DIFF
--- a/desc/coils.py
+++ b/desc/coils.py
@@ -769,7 +769,7 @@ def _check_type(coil0, coil):
         FourierRZCoil: ["R_basis", "Z_basis", "NFP", "sym"],
         FourierXYZCoil: ["X_basis", "Y_basis", "Z_basis"],
         FourierPlanarCoil: ["r_basis"],
-        SplineXYZCoil: ["method", "N"],
+        SplineXYZCoil: ["method", "N", "knots"],
     }
 
     for attr in attrs[coil0.__class__]:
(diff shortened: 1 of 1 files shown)


f0uriest (approved):
Remember to do the same for breakpoints in #970

dpanici (approved):

ddudt (approved):

PR #1147, opened 2024-07-24
Add "tiers" to `data_index`

f0uriest:
This is the first step in moving our `compute` infrastructure from a recursive model to a looped one, which is potentially faster to compile and requires less memory.

For now all this does is automatically assign a "tier" to each computeable quantity that indicates how "deep" in the dependency tree that quantity lies. 

tier of 0 means no dependencies on other data, but may depend on transforms, profiles etc.
tier of 1 means it depends on only tier 0 stuff,
tier of 2 means it depends on tier 0 and tier 1, etc etc.

The goal is that if you were to compute things in a loop with the order determined by the tiers, then it would always compute dependencies in the correct order.


YigitElma (commented):
I think the ordering you have in mind will only work for tier 0 and 1 computable quantities. If we only plan to have 3 tiers, then tier 2 quantities may depend on other tier 2 quantities and this needs some other logic for ordering.

f0uriest:
> I think the ordering you have in mind will only work for tier 0 and 1 computable quantities. If we only plan to have 3 tiers, then tier 2 quantities may depend on other tier 2 quantities and this needs some other logic for ordering.

the number of tiers isnt fixed, its determined automatically. The only rule is that to compute tier N you first need to compute tiers 0:N-1

dpanici (approved):

desc/compute/__init__.py
```diff
@@ -94,3 +94,30 @@ def _build_data_index():
...
+
+    Designed such that if you compute things in the order determined by tiers,
+    all dependencies will always be computed in the correct order.
+    """
+    if "tier" in data_index[p][name]:
+        return
+    if len(data_index[p][name]["full_with_axis_dependencies"]["data"]) == 0:
+        data_index[p][name]["tier"] = 0
+    else:
+        thistier = 0
```
ddudt, line 114: I don't think it will matter, but technically this should be `thistier = 1` since being in this else statement means that it has a finite amount of dependencies and therefore must be at least tier 1 or higher.

ddudt: Also something like `dep_tier` for "dependency tier" might be a better variable name

unalmis: This is correct as written; setting `thistier` to anything greater than 0 would be incorrect. We add +1 in line 118 to max dependency tier.

desc/compute/__init__.py
```diff
@@ -94,3 +94,30 @@ def _build_data_index():
 
 
 _build_data_index()
+
+
+def set_tier(name, p):
```
unalmis, line 99: kind of like (p, name) to match `data_index[p][name]`

unalmis (approved):

PR #1150, opened 2024-07-26
Remove numpy backend

### LINKED ISSUE #1149: Remove `numpy` backend
DESC requires JAX now to run (due to `interpax` dependency), so the numpy backend is not really used or relevant anymore. We should remove it then to reduce amount of code to maintain (and it is not used at all)

YigitElma:
Resolves #1149


unalmis:
This will cause merge conflicts with all my open PRs, so could we wait to merge this?

desc/backend.py
```diff
@@ -11,14 +11,10 @@
...
-if os.environ.get("DESC_BACKEND") == "numpy":
-    jnp = np
+if os.environ.get("DESC_BACKEND") == "numpy":  # pragma: no cover
     use_jax = False
-    set_device(kind="cpu")
-    print(
-        "DESC version {}, using numpy backend, version={}, dtype={}".format(
-            desc.__version__, np.__version__, np.linspace(0, 1).dtype
-        )
+    warnings.warn(
```
ddudt, line 16: should raise a `DeprecationWarning`

desc/backend.py
```diff
@@ -47,16 +43,13 @@
...
-        x = jnp.linspace(0, 5)
-        y = jnp.exp(x)
+    except ModuleNotFoundError:  # pragma: no cover
         use_jax = False
-        set_device(kind="cpu")
-        warnings.warn(colored("Failed to load JAX", "red"))
-        print(
-            "DESC version {}, using NumPy backend, version={}, dtype={}".format(
-                desc.__version__, np.__version__, y.dtype
+        warnings.warn(
```
ddudt, line 48: should raise a `DeprecationWarning`

desc/backend.py
```diff
@@ -394,420 +387,11 @@ def tangent_solve(g, y):
...
-        -------
-        xk : float
-            Root, or best approximation
-        info : tuple of (float, int)
-            Residual of fun at xk and number of iterations of outer loop
-
-        """
-        out = scipy.optimize.root_scalar(
-            fun, args, x0=x0, fprime=jac, xtol=tol, rtol=tol
+    warnings.warn(
```
ddudt, line 391: should raise a `DeprecationWarning`

setup.cfg
```diff
@@ -73,6 +73,7 @@ ignore =
 per-file-ignores =
     # need to import things to top level even if they aren't used there
     desc/*/__init__.py: F401
+    desc/backend.py: F401
```
ddudt, line 76: Why is this needed?

ddudt (changes requested):
Does any part of the code actually work without JAX? If not, we should just throw an error instead of a warning.

setup.cfg
YigitElma: For some reason, flake was complaining about imports not being used.

unalmis:
it would cause one that looks simple to resolve

YigitElma:
I assume we won't do this in the near future. So, until we consider removing numpy backend again, I am closing this.

PR #1151, opened 2024-07-26
`CoilSet.to_FourierRZ` and `CoilSet.to_FourierPlanar`

ddudt:
These are methods that we forgot to add to #1116. 

It might also be worth adding an option to `CoilSet.from_makegrid_coilfile` so that the coil set can be loaded as any type, instead of always using `SplineXYZCoil`. But at least now we can load the spline coils and then convert to any other coil type.

### DIFF
--- a/desc/coils.py
+++ b/desc/coils.py
@@ -362,10 +362,9 @@ class _Coil(_MagneticField, Optimizable, ABC):
             Number of field periods, the coil will have a discrete toroidal symmetry
             according to NFP.
         sym : bool, optional
-            whether the curve is stellarator-symmetric or not. Default
-            is False.
+            Whether the curve is stellarator-symmetric or not. Default is False.
         name : str
-            name for this coil
+            Name for this coil.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


desc/coils.py
```diff
@@ -1599,6 +1596,64 @@ def flatten_coils(coilset):
...
+        name : str
+            Name for this coilset.
+
+        Returns
+        -------
+        coilset : CoilSet
+            New representation of the coilset parameterized by a Fourier series for R,Z.
+
+        """
+        coils = [coil.to_FourierRZ(N=N, grid=grid, NFP=NFP, sym=sym) for coil in self]
```
f0uriest, line 1654: do the coils themselves need to have the same nfp and sym as the coilset? I feel like there are cases where that isn't the case

ddudt: This does not restrict them to be the same. This function converts an existing `CoilSet` from one parameterization to another, but the NFP and sym for that coil set is always preserved. If the user wants to convert to the `FourierRZ` type, then they have to specify NFP and sym as inputs to this method. And those can be independent of the `CoilSet` attributes.

f0uriest (approved):

dpanici (approved):

PR #1153, opened 2024-07-26
Add map_clebsch_coords

unalmis:


tests/test_equilibrium.py
```diff
@@ -61,6 +64,9 @@ def test_map_coordinates():
     )
     assert not np.any(np.isnan(out))
 
+    iota = np.broadcast_to(iota, shape=(n,))
+    np.testing.assert_allclose(eq.map_clebsch_coords(coords, iota), out)
```
unalmis, line 68: The new method doesn't match the old one, so I must be doing something wrong.

unalmis: Removing `eq.change_resolution` fixes this. I think this is because changing radial resolution affects `map_coordinates` but it does not effect `map_clebsch_coords`, or some inconsistency like that.

dpanici:
When it is ready may I request a benchmark (just running with the map coords vs the map clebsch and seeing what the timing difference is) so we know how much it improves

unalmis:
> When it is ready may I request a benchmark (just running with the map coords vs the map clebsch and seeing what the timing difference is) so we know how much it improves

Sure, the bigger improvement will come from #1154, but the changes here should also help. Note that this is not just helpful for speed. The general map coords does 3d root finding, and will converge to incorrect results if the radial resolution of the equilibrium is not sufficient, while the 1d root finding here is unaffected https://github.com/PlasmaControl/DESC/pull/1153#discussion_r1693731921.

unalmis:
> > When it is ready may I request a benchmark (just running with the map coords vs the map clebsch and seeing what the timing difference is) so we know how much it improves
> 
> Sure, the bigger improvement will come from #1154, but the changes here should also help, and are precursor to that pr. Note that this is not just helpful for speed. The general map coords does 3d root finding, and will converge to different results if the radial resolution of the equilibrium is not sufficient or if the spline approximates the radial derivative of iota poorly, while the 1d root finding here is unaffected [#1153 (comment)](https://github.com/PlasmaControl/DESC/pull/1153#discussion_r1693731921).

```python
rho = np.linspace(0, 1, 3)
alpha = np.array([0, 1])
zeta = np.linspace(0, 20 * np.pi, 300)
grid = Grid.create_meshgrid(
)
inbasis = {
    "r": "rho",
    "t": "theta",
    "v": "theta_PEST",
    "a": "alpha",
    "z": "zeta",
}
eq = get("W7-X")
iota_grid = LinearGrid(rho=rho, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=eq.sym)
iota = grid.copy_data_from_other(eq.compute("iota", grid=iota_grid)["iota"], iota_grid)
```
```python
%%timeit

rtz_nodes = eq.map_clebsch_coords(
    grid.nodes,
    iota,
    eq.L_lmn,
    eq.L_basis,
    maxiter=20,
)
# 4.64 s ± 109 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
%%timeit

rtz_nodes = eq.map_coordinates(
    grid.nodes,
    inbasis=[inbasis[char] for char in grid.coordinates],
    outbasis=("rho", "theta", "zeta"),
    period=grid.period,
    maxiter=20,
)
# 16.9 s ± 787 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

desc/equilibrium/coords.py
```diff
@@ -312,6 +322,103 @@ def fixup(x, *args):
...
+    info : tuple
+        2 element tuple containing residuals and number of iterations for each point.
+        Only returned if ``full_output`` is True.
+
+    """
+    kwargs.setdefault("maxiter", maxiter)
+    kwargs.setdefault("tol", tol)
+
+    rho, alpha, zeta = clebsch_coords.T
+    alpha = alpha % (2 * np.pi)
```
dpanici, line 377: do we want to modulo alpha by 2pi? I would think for most cases yes but would there ever be a case when we don't want this?

I dont have an example off the top of my head, just want to pose the question

unalmis: We can either do it once here or move the modulo on line 385 to act on `r`. I chose the former

desc/equilibrium/equilibrium.py
```diff
@@ -1243,18 +1248,80 @@ def compute_theta_coords(
...
             self,
             flux_coords,
             L_lmn=L_lmn,
+            tol=tol,
             maxiter=maxiter,
+            full_output=full_output,
+            **kwargs,
+        )
+
+    def map_clebsch_coords(
```
dpanici, line 1269: Can this also be `compute_clebsch_coords` to align with `compute_theta_coords`?

dpanici (changes requested):
change name of function so it is obvious it is the clebsch analog to `compute_theta_coords`

desc/equilibrium/equilibrium.py
unalmis: Counter: this is closer to `map_coordinates` name and

the theta in `compute_theta_coords` is DESC theta. In both functions we compute DESC theta, so `compute_clebsch_coords` is a bit weird

f0uriest (commented):
My main comment is about API surface. This adds another specific mapping function which only works in limited circumstances. That's fine, but I think it would be cleaner if we kept the main API as `map_coordinates` but then include something like

```python
if inbasis == ["rho", "alpha", "zeta"] and outbasis == ["rho", "theta", "zeta"]:
    return _map_clebsch_coordinates(...)
```

(and maybe do the same for `compute_theta_coordinates`). 

I think this would simplify things and avoid having to think "when should I use method A vs method B". It would also generalize more easily if we get #1154 to work.

desc/equilibrium/equilibrium.py
unalmis: I'm never going to use `eq.map_clebsch`, so I can remove that. I would like to use `map_clebsch`. We can redirecto to that through `map_coordinates`

desc/equilibrium/coords.py
```diff
@@ -67,44 +71,66 @@ def map_coordinates(  # noqa: C901
...
-    basis_derivs = tuple([f"{X}_{d}" for X in inbasis for d in ("r", "t", "z")])
+    if outbasis == ("rho", "theta", "zeta"):
+        # TODO: get iota if not supplied using below logic
+        if inbasis == ("rho", "alpha", "zeta") and "iota" in kwargs:
+            return _map_clebsch_coordinates(
+                coords,
+                kwargs.pop("iota"),
+                params["L_lmn"],
+                eq.L_basis,
+                guess,
```
f0uriest, line 102: guess would be an nx3 array but i think we only want to pass in `guess[:,1]` here (and below)

desc/equilibrium/coords.py
```diff
@@ -67,44 +71,66 @@ def map_coordinates(  # noqa: C901
...
+    params = setdefault(params, eq.params_dict)
     inbasis = tuple(inbasis)
     outbasis = tuple(outbasis)
-    assert (
-        np.isfinite(maxiter) and maxiter > 0
-    ), f"maxiter must be a positive integer, got {maxiter}"
-    assert np.isfinite(tol) and tol > 0, f"tol must be a positive float, got {tol}"
 
-    basis_derivs = tuple([f"{X}_{d}" for X in inbasis for d in ("r", "t", "z")])
+    if outbasis == ("rho", "theta", "zeta"):
```
f0uriest, line 94: add a TODO to make this also work if the coordinate labels are permuted for some reason (ie it should work for [zeta, rho, theta] etc.)

desc/equilibrium/coords.py
```diff
@@ -307,51 +346,58 @@ def fixup(x, *args):
...
         Each row is a different point in space.
     iota : ndarray
         Shape (k, )
         Rotational transform on each node.
-    L_lmn : ndarray
+    L_lmn : jnp.ndarray
         Spectral coefficients for lambda.
     L_basis : Basis
         Spectral basis for lambda.
+    guess : ndarray, shape(k,3)
```
f0uriest, line 396: i think guess should just be a 1d array for theta here

f0uriest (approved):

desc/vmec.py
```diff
@@ -1687,7 +1687,7 @@ def root_fun(theta):
                 axis=-1,
             )
             theta_star_k = theta + lmbda  # theta* = theta + lambda
-            err = theta_star - theta_star_k
+            err = theta_star - theta_star_k  # FIXME: mod by 2pi
```
dpanici, line 1690: should we make an issue for this fixme?

dpanici (approved):
Approved but make an issue for that FIXME in vmec.py

PR #1158, opened 2024-07-31
Scaled Linear Constraints

### LINKED ISSUE #1155: coil currents don't change much when using `FixSumCoilCurrent`
@dpanici has mentioned this issue offline, but I want to document it: 

The linear constraint `FixSumCoilCurrent` does what it is intended to do, but when used in an optimization does not allow the coil currents to change much relative to their nominal values on the order of 1e6 Amps.  I have noticed that using the nonlinear constraint `ToroidalFlux` allows the coil currents to change much more and gives better results in practice.  

When the linear constraint is included, `factorize_linear_constraints` normalizes the optimization variables `x_reduced` to order unity, while the particular solution `xp` is order 1e6 so that the full state vector `x_full = xp + Z @ x_reduced` has the proper magnitude in Amperes.  The null space matrix `Z` is also order unity, so even large changes in the optimization space result in very small changes to the true coil currents.  

A hacky solution would be to redefine the units of the coil currents to be in MA instead of Amps. Alternatively, it would be nice if the rows of `Z` were scaled proportionally to `xp`, such that relative changes in the optimization variables `x_reduced` yield similar relative changes in the full state vector `x_full`.

ddudt:
Resolves #1155 

~~After discussion with @dpanici, we decided the best solution was to add an attribute to the `_Coil` class to normalize the magnitude of the current. This effectively allows the user to specify the currents in units other than Amps, e.g. MA, so that the raw values are closer to order unity. This makes the optimization work well with linear constraints on the coil currents.~~

New solution: Adds a diagonal scaling matrix `D` based on the magnitude of the values of the particular solution. This scales the variables projected/recovered from the linear constraints such that relative changes in the optimization variables `x_reduced` result in proportional relative changes in the full state vector `x_full`. 

TODO:

- [x] Multiple `Z` by `D` where necessary in perturbations, constraint wrappers, etc. 
- [x] Add tests.


f0uriest:
with the usual automatic scaling the variable the optimizer sees is `x / |f'(x)|`
this is basically making a new function `g(y=ax) = f(ax)` so the optimizer sees `y / |g'(y)| = ax / |a f'(x)| = x \ |f'(x)|`, so it seems like it wouldn't actually affect the optimization?

ddudt:
> with the usual automatic scaling the variable the optimizer sees is `x / |f'(x)|` this is basically making a new function `g(y=ax) = f(ax)` so the optimizer sees `y / |g'(y)| = ax / |a f'(x)| = x \ |f'(x)|`, so it seems like it wouldn't actually affect the optimization?

Where in the code do you think this automatic scaling happens? Because if there are not linear constraints

> with the usual automatic scaling the variable the optimizer sees is `x / |f'(x)|` this is basically making a new function `g(y=ax) = f(ax)` so the optimizer sees `y / |g'(y)| = ax / |a f'(x)| = x \ |f'(x)|`, so it seems like it wouldn't actually affect the optimization?

See my reply to your comment on the issue.  The underlying problem is with how the linear constraints are projected/recovered, not with the optimization itself.  So the automatic scaling within the optimizer does not resolve the issue.

dpanici:
Is it better to just open a new PR rather than having a "revert all commits" commit?

ddudt:
> Is it better to just open a new PR rather than having a "revert all commits" commit? too late now I suppose

It was only 3 commits so not too big of a deal. Technically I think this is the more "correct" git approach so we don't have loose ends in the commit history from unused branches

dpanici:
> > Is it better to just open a new PR rather than having a "revert all commits" commit? too late now I suppose
> 
> It was only 3 commits so not too big of a deal. Technically I think this is the more "correct" git approach so we don't have loose ends in the commit history from unused branches

Can always delete branches, but this is fine

desc/objectives/utils.py
```diff
@@ -33,6 +33,8 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
         Combined RHS vector.
     Z : ndarray
         Null space operator for full combined A such that A @ Z == 0.
+    D : ndarray
+        Diagonal scaling matrix, based on the magnitude of the particular solution xp.
```
ddudt, line 37: Reminder to myself to update this if we change the scaling heuristic.

desc/objectives/utils.py
```diff
@@ -130,32 +132,49 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
             )
         A = A[unfixed_rows][:, unfixed_idx]
         b = b[unfixed_rows]
+
+    # unscaled particular solution to get scale of x
```
ddudt, line 136: This seems a little wasteful because we are now performing the SVD twice: once for the original system to get the scale of the variables, then again for the scaled system. 

The issue is we need a way to know the scale of the full state vector, and the particular solution `xp` gives us that. Otherwise we would have add the optimizable `things` as an input to this function so that we can compute the initial state vector `x0` and use that as the scale. 

In practice I don't think this SVD takes much computation time so I'm not too worried about it. But I'm open to suggestions for improvement here.

desc/objectives/utils.py
```diff
@@ -130,32 +132,49 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
...
     else:
-        Ainv_full = A.T
+        A_inv = A.T
         Z = np.eye(A.shape[1])
-    Ainv_full = jnp.asarray(Ainv_full)
-    Z = jnp.asarray(Z)
-    b = jnp.asarray(b)
-    xp = put(xp, unfixed_idx, Ainv_full @ b)
+    xp = put(xp, unfixed_idx, A_inv @ b)
+    x_scale = np.where(np.abs(xp) < 1, 1, np.abs(xp))  # TODO: adjust threshold?
```
ddudt, line 144: Right now, this helps rescale variables with very large amplitudes. We could change this threshold to something like 
```
x_scale = np.where(np.abs(xp) < 1e-6, 1, np.abs(xp))
```
to also help rescale variables with very small amplitudes. Is that worth doing? I don't think it can hurt, but also I don't think we have any variables (yet) with small normalizations.

desc/optimize/_constraint_wrappers.py
```diff
@@ -1018,7 +1035,11 @@ def jvp_unscaled(self, v, x, constants=None):
     @functools.partial(jit, static_argnames=("self", "op"))
     def _jvp_f(self, xf, dc, constants, op):
         Fx = getattr(self._constraint, "jac_" + op)(xf, constants)
-        Fx_reduced = Fx[:, self._unfixed_idx] @ self._Z
+        Fx_reduced = (
+            Fx[:, self._unfixed_idx]
+            @ self._D[self._unfixed_idx][:, self._unfixed_idx]
+            @ self._Z  # TODO: replace with self._unfixed_idx_mat
+        )
```
ddudt: I think this should be equivalent to
```
Fx @ self._unfixed_idx_mat[self._eq_idx]
```
but I'm not certain and I want another opinion

desc/optimize/_constraint_wrappers.py
```diff
@@ -1018,7 +1027,7 @@ def jvp_unscaled(self, v, x, constants=None):
     @functools.partial(jit, static_argnames=("self", "op"))
     def _jvp_f(self, xf, dc, constants, op):
         Fx = getattr(self._constraint, "jac_" + op)(xf, constants)
-        Fx_reduced = Fx[:, self._unfixed_idx] @ self._Z
+        Fx_reduced = Fx @ self._D[:, self._unfixed_idx] @ self._Z
```
ddudt, line 1030: I think this can be replaced with `self._unfixed_idx_mat` but I'm not certain and want another opinion

desc/objectives/utils.py
f0uriest: To get `xp` we don't need the null space, so in theory we don't need to use an SVD at all, we could do a regular QR least norm which would in theory be quite a bit faster.

That said, I think basing the scale of x0 might have some other benefits, such as properly rescaling variables that aren't necessarily being constrained. IE, for bootstrap current optimization the current profile is usually free so it wouldn't show up in any particular solution and hence wouldn't get scaled, though it may help the optimizer if it was.

desc/objectives/utils.py
```diff
@@ -130,32 +132,49 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
...
+    D = np.diag(x_scale)
```
f0uriest, line 145: it's probably more efficient to only store this as a 1d vector and do all the operations elementwise with broadcasting.

ie
```python
D = np.diag(x_scale)
D @ A == A * x_scale[:,None]
A @ D == A * x_scale[None,:]
```

dpanici:
Use `x_scale` as optional arg to `facotrize linear constraint` , if auto use x0 way (if things is not None) otherwise use x_scale as auto

desc/objectives/utils.py
```diff
@@ -130,32 +141,51 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
...
+    fixed_idx = np.delete(np.arange(xp.size), unfixed_idx)
+
+    # get x_scale if not provided
+    if x_scale == "auto":
+        if things is None:
+            things = objective.things
+        elif not isinstance(things, list):
+            things = [things]
+        x_scale = objective.x(*[things[things.index(t)] for t in objective.things])
+    D = np.where(np.abs(x_scale) < 1e1, 1, np.abs(x_scale))  # TODO: adjust threshold?
```
ddudt, line 155: I increased the threshold to `10` because otherwise some tests were failing. This basically sets it so that the scaling is not applied to the usual R, Z, Lambda, etc. coefficients and will only affect parameters like current, density, temperature, etc.

desc/objectives/utils.py
```diff
@@ -130,32 +149,57 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
...
     unfixed_idx = indices_idx
+    fixed_idx = np.delete(np.arange(xp.size), unfixed_idx)
+
+    # compute x_scale if not provided
+    if x_scale == "auto":
+        if things is None:
+            things = objective.things
+        else:
+            things = [things] if not isinstance(things, list) else things
+            things = [get_instance(things, type(t)) for t in objective.things]
```
ddudt, line 162: This logic will not work as intended if you are optimizing multiple things that are of the same type (because `get_instance` always grabs the first instance from the list). However: 

1. I don't think we have any use cases for that yet. 
2. This will only affect the scaling, so as long as both things have similar scale parameters it will be fine.

tests/test_bootstrap.py
ddudt, line 1: This test shouldn't have needed to change, since the scales are all 1. I think the bootstrap current optimization is just really sensitive and the extra identity matrices changed the numerics slightly. Is that plausible?

desc/objectives/utils.py
f0uriest: couldn't this be an issue if we're optimizing eg modular coils with large currents + trim coils with small currents?

I was thinking about the use case for having things different from objective.things and I thought we haven't officially supported that for a while? Also after #1043 there won't be any reason to re-use objectives since the compile cache will be shared between instances.

desc/objectives/utils.py
```diff
@@ -130,32 +149,57 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
...
+        x_scale = objective.x(*things)
+    errorif(
+        x_scale.shape != xp.shape,
+        ValueError,
+        "x_scale must be the same size as the full state vector.",
```
f0uriest, line 167: would he helpful to print the expected size and the size you got here

desc/objectives/utils.py
```diff
@@ -130,32 +149,57 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
...
+    )
+    D = np.where(np.abs(x_scale) < 2e1, 1, np.abs(x_scale))
```
f0uriest, line 169: something more like 100-1000 might be better to clearly separate stuff like R, Z (which could be ~20 for reactor size stuff) from pressure, current, etc which would usually be > 1e3~1e4

desc/optimize/_constraint_wrappers.py
```diff
@@ -281,13 +280,17 @@ def hess(self, x_reduced, constants=None):
...
+            @ df[self._unfixed_idx, :][:, self._unfixed_idx]
+            @ (self._Z * self._D[self._unfixed_idx, None])
+        )
 
     def _jac(self, x_reduced, constants=None, op="scaled"):
         x = self.recover(x_reduced)
         if self._objective._deriv_mode == "blocked":
             fun = getattr(self._objective, "jac_" + op)
-            return fun(x, constants)[:, self._unfixed_idx] @ self._Z
+            return fun(x, constants) @ self._unfixed_idx_mat
```
f0uriest, line 293: if `len(unfixed_idx) << fun.shape[1]` then this will be slower doing it this way

desc/objectives/utils.py
ddudt: In the case of modular + trim coils, they should already be combined into a `MixedCoilSet` in order to compute the total magnetic field, so that's not an issue here. An example of what I mean is if you were optimizing multiple `Equilibrium` or multiple `_Coil` for some reason, which I don't think you should ever be doing (yet). 

It might be uncommon that `things` will be different from `objective.things`, but I still think it's a use case we need to support. For example, you could save an `ObjectiveFunction` that you used in a previous optimization (as an HDF5 output) and then load it later to use on a different initial guess.

f0uriest: > It might be uncommon that things will be different from objective.things, but I still think it's a use case we need to support. For example, you could save an ObjectiveFunction that you used in a previous optimization (as an HDF5 output) and then load it later to use on a different initial guess.

im pretty sure that won't work, and i don't know if it ever really did? At the very least we don't test it or document it, so I don't have any expectations that it would or should work.

desc/objectives/utils.py
```diff
@@ -22,6 +32,13 @@ def factorize_linear_constraints(objective, constraint):  # noqa: C901
         Objective function to optimize.
     constraint : ObjectiveFunction
         Objective function of linear constraints to enforce.
+    things : Optimizable or tuple/list of Optimizable
```
VascoSch92, line 35: perhaps use `optimisables` as name for this parameters or something similar? `things` is not so descriptive :-)

ddudt: `things` is the argument name we use consistently throughout the rest of the code. It's not perfect, but `Optimizable` is already a class name so we couldn't think of anything better

desc/objectives/utils.py
ddudt: @f0uriest What are you requesting me to do here? This code works for all of our expected use cases. Or are you saying we should *always* use `objective.things` and not allow the user to pass in separate `things`?

f0uriest: that would be my vote but its up to you

ddudt: It would be cleaner if we always used `objective.things`, I just want to make sure that won't cause any problems. I still think it is possible to re-use objectives if the user wants to. But if we aren't worried about that, then I'll make that change.

f0uriest: I think reusing objectives currently falls under "undefined behavior" in that its not promised to work but may in certain circumstances. And the "correct" way would be to re-assign `objective.things`, not to pass in a separate set of things anyways.

ddudt: OK, I've changed it to always use `objective.things`. And since this only affects the scaling, it's not that big of a deal if someone tries to re-use an objective with different things anyways.

f0uriest (approved):

dpanici (approved):

PR #1159, opened 2024-08-01
Update Toroidal Flux Objective for Finite Beta Equilibria

dpanici:
Removes the warnings when a finite beta equilibrium is passed into `ToroidalFlux` objective, and accounts for the plasma currents as well.

In relation to [comment](https://github.com/PlasmaControl/DESC/issues/1155#issuecomment-2257163466) in #1158 

With finite beta stage 2 optimization, no objective on toroidal flux or coil curents is needed for the coils to make the correct toroidal flux and field magnitude, assuming that normal field error is zero. With finite normal field error, the coils may not create the necessary fluxes or have the requisite net poloidal linking current for the given equilbrium.


f0uriest (approved):

f0uriest (changes requested):
Actually, at finite plasma current I think we need to change the compute function to account for the contribution to the flux from the plasma field?

dpanici:
> Actually, at finite plasma current I think we need to change the compute function to account for the contribution to the flux from the plasma field?

Ah I think you are right, I can add that.

f0uriest:
actually im now doubting myself. Can plasma current produce toroidal flux? i thought that wasn't possible (or maybe it just cant produce "enough" flux?)

dpanici:
> actually im now doubting myself. Can plasma current produce toroidal flux? i thought that wasn't possible (or maybe it just cant produce "enough" flux?)

Any poloidally circulating current in the plasma would produce a toroidal flux right? so I think you're right in that for finite beta this is something that should be taken into account for this objective.

I am imagining simplest case of a theta-pinch, the currents flow in the theta direction and the toroidal (Z-directed at least) flux enclosed by the bdry certainly has a contribution from those poloidal currents

dpanici:
Actually the poloidal currents wouuld be the diamagnetic currents right? so if anything they would reduce the toroidal flux... this is something I have not thought about the subtleties of, nor have I seen it in any papers on finite beta coil stuff

dpanici:
@dpanici check Helander 2014 for this

YigitElma:
@dpanici @f0uriest, what was the final verdict on this?

PR #1165, opened 2024-08-06
Take `p_newton` out of inner while loop

### LINKED ISSUE #1078: Take `p_newton` calculation out of the inner while loop when solving trust region subproblem
> I still think that we can take `p_newton` calculation out of the inner while loop. I assume there will be a follow-up PR about QR to make `falsefun` faster (at least I plan to work on that), so I can try some additional optimizations there. For now, this version is fine and works better than SVD on GPU.

_Originally posted by @YigitElma in https://github.com/PlasmaControl/DESC/pull/1050#pullrequestreview-2137189979_

YigitElma:
Resolves #1078 

Some performance improvements for QR decomposition used in optimization which was first introduced in #1050.

- Take the `p_newton` calculation out of inner while loop, since it is basically calculating the same QR over and over again
- ~Use proper QR update procedure for the `falsefun` in `trust_region_step_exact_qr`. That is we already now QR decomposition of `J=QR`, if we stack a diagonal matrix `aI` to `J` then instead of taking the whole QR decomposition again, there is a more clever way of updating the QR.There are methods for updating a QR factorization when you add rows. Suppose we have~

$$
QR = J
$$

what we want is 

$$
\tilde{Q} \tilde{R} = \begin{pmatrix} J \\ 
\alpha I \end{pmatrix}
$$

The QR update procedure can be implemented on a later PR with Householder matrices, but for now, it seems a bit inefficient to implement using JAX since QR is calculated by Fortran package LAPACK on Scipy and Jax, our custom QR'ish thing will be slow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+New Features
+
+- Add ``use_signed_distance`` flag to ``PlasmaVesselDistance`` which will use a signed distance as the target, which is positive when the plasma is inside of the vessel surface and negative if the plasma is outside of the vessel surface, to allow optimizer to distinguish if the equilbrium surface exits the vessel surface and guard against it by targeting a positive signed distance.
+
 v0.12.1
 -------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,21 @@
 Changelog
 =========
 
-New Features
+v0.12.1
+-------
 
-- adds ``from_values`` method that was present in ``FourierRZCurve`` but missing in ``FourierRZCoil``
-- also adds new ``from_values`` method for ``FourierPlanarCurve`` and ``FourierPlanarCoil``
+[Github Commits](https://github.com/PlasmaControl/DESC/compare/v0.12.0...v0.12.1)
+
+- Optimizers now default to use QR factorization for least squares which is much faster
+especially on GPU.
+- Fix bug when reading VMEC input ZAXIS as ZAXIS_CS
+- Some fixes/improvements for computing quantities along a fieldline.
+- Adds compute quantities for PEST coordinate system basis vectors
+- Many init methods now default to running on CPU, even when GPU is enabled, as CPU was found
+to be much faster for these cases.
+- New objectives `desc.objectives.FixNearAxis{R,Z,Lambda}` for fixing near axis behavior.
+- Adds ``from_values`` method that was present in ``FourierRZCurve`` but missing in ``FourierRZCoil``
+- Adds new ``from_values`` method for ``FourierPlanarCurve`` and ``FourierPlanarCoil``
 
 
 v0.12.0
--- a/desc/backend.py
+++ b/desc/backend.py
@@ -76,6 +76,7 @@ if use_jax:  # noqa: C901 - FIXME: simplify this, define globally and then assig
     repeat = jnp.repeat
     take = jnp.take
     scan = jax.lax.scan
+    rsqrt = jax.lax.rsqrt
     from jax import custom_jvp
     from jax.experimental.ode import odeint
     from jax.scipy.linalg import block_diag, cho_factor, cho_solve, qr, solve_triangular
@@ -811,3 +812,7 @@ else:  # pragma: no cover
         else:
             out = np.take(a, indices, axis, out, mode)
         return out
+
+    def rsqrt(x):
+        """Reciprocal square root."""
+        return 1 / np.sqrt(x)
--- a/desc/backend.py
+++ b/desc/backend.py
@@ -76,7 +76,6 @@ if use_jax:  # noqa: C901 - FIXME: simplify this, define globally and then assig
     repeat = jnp.repeat
     take = jnp.take
     scan = jax.lax.scan
-    rsqrt = jax.lax.rsqrt
     from jax import custom_jvp
     from jax.experimental.ode import odeint
     from jax.scipy.linalg import block_diag, cho_factor, cho_solve, qr, solve_triangular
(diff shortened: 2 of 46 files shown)


YigitElma:
I have used Givens rotations for zeroing the elements but maybe Householder reflections are better? Maybe try implementing that.

YigitElma:
> ```diff
>  test_build_transform_fft_lowres         |     -2.08 +/- 9.64     | -1.10e-02 +/- 5.10e-02 |  5.18e-01 +/- 4.1e-02  |  5.29e-01 +/- 3.1e-02  |
>  test_build_transform_fft_midres         |     -1.19 +/- 5.78     | -7.22e-03 +/- 3.52e-02 |  6.02e-01 +/- 2.9e-02  |  6.09e-01 +/- 2.0e-02  |
>  test_build_transform_fft_highres        |     -2.08 +/- 3.18     | -2.10e-02 +/- 3.21e-02 |  9.89e-01 +/- 1.3e-02  |  1.01e+00 +/- 3.0e-02  |
>  test_equilibrium_init_lowres            |     -2.47 +/- 5.04     | -9.46e-02 +/- 1.93e-01 |  3.73e+00 +/- 1.3e-01  |  3.82e+00 +/- 1.4e-01  |
>  test_equilibrium_init_medres            |     -1.82 +/- 4.48     | -7.74e-02 +/- 1.91e-01 |  4.18e+00 +/- 9.5e-02  |  4.26e+00 +/- 1.7e-01  |
>  test_equilibrium_init_highres           |     -1.33 +/- 2.01     | -7.53e-02 +/- 1.14e-01 |  5.59e+00 +/- 5.3e-02  |  5.66e+00 +/- 1.0e-01  |
>  test_objective_compile_dshape_current   |     -1.55 +/- 3.14     | -6.12e-02 +/- 1.24e-01 |  3.89e+00 +/- 2.5e-02  |  3.95e+00 +/- 1.2e-01  |
>  test_objective_compile_atf              |     -1.07 +/- 2.79     | -9.00e-02 +/- 2.36e-01 |  8.35e+00 +/- 1.0e-01  |  8.44e+00 +/- 2.1e-01  |
>  test_objective_compute_dshape_current   |     -1.31 +/- 3.37     | -1.65e-05 +/- 4.25e-05 |  1.25e-03 +/- 2.5e-05  |  1.26e-03 +/- 3.5e-05  |
>  test_objective_compute_atf              |     -0.65 +/- 4.70     | -2.77e-05 +/- 1.99e-04 |  4.21e-03 +/- 1.1e-04  |  4.24e-03 +/- 1.6e-04  |
>  test_objective_jac_dshape_current       |     -0.23 +/- 7.42     | -8.59e-05 +/- 2.71e-03 |  3.65e-02 +/- 1.6e-03  |  3.66e-02 +/- 2.2e-03  |
>  test_objective_jac_atf                  |     -0.22 +/- 2.64     | -4.22e-03 +/- 4.95e-02 |  1.87e+00 +/- 2.8e-02  |  1.88e+00 +/- 4.1e-02  |
>  test_perturb_1                          |     -1.39 +/- 1.07     | -1.96e-01 +/- 1.51e-01 |  1.39e+01 +/- 8.6e-02  |  1.41e+01 +/- 1.2e-01  |
>  test_perturb_2                          |     -1.95 +/- 1.07     | -3.73e-01 +/- 2.05e-01 |  1.88e+01 +/- 1.3e-01  |  1.92e+01 +/- 1.6e-01  |
>  test_proximal_jac_atf                   |     +0.76 +/- 0.84     | +5.57e-02 +/- 6.13e-02 |  7.38e+00 +/- 4.8e-02  |  7.33e+00 +/- 3.9e-02  |
>  test_proximal_freeb_compute             |     -0.43 +/- 1.32     | -7.79e-04 +/- 2.40e-03 |  1.81e-01 +/- 1.4e-03  |  1.82e-01 +/- 1.9e-03  |
>  test_proximal_freeb_jac                 |     -0.32 +/- 1.32     | -2.36e-02 +/- 9.73e-02 |  7.36e+00 +/- 8.7e-02  |  7.38e+00 +/- 4.3e-02  |
> -test_solve_fixed_iter                   |   +5639.94 +/- 7.00    | +1.04e+03 +/- 1.29e+00 |  1.06e+03 +/- 1.1e+00  |  1.85e+01 +/- 6.4e-01  |
> ```
> 
> 

😶🫣🤔

YigitElma:
I think the method is correct. The new Q and R matrices are almost the same as the ones found by `q,r = jax.scipy.linalg.qr(A_t)`. The only difference is that Some rows of the our method differs in sign, so instead of `Q-Q_our==0` we have `|Q|-|Q_our|==0`. The double for loop (even with jax) is very slow (actually it was known but I thought it only applies to GPU). I will try to implement Househoulder reflections since they don't require nested for loops but a single one over columns.

YigitElma:
Just take `p_newton` out for this PR. Maybe try Householder later

f0uriest (approved):

dpanici (approved):

unalmis:
Can you add a test?

YigitElma:
> Can you add a test?

Technically I didn't change any logic. The code coverage is lower because previously the `qr` part was only in the `trust_region_step_exact_subproblem` and it was tested (or not, we don't have a test for that I guess) once. Now, the same `qr` part appears in 2 files (`aug_lagrangian_ls.py` and `least_squares.py`). I guess the only way to test these is to construct an optimization problem with tall/wide Jacobian for augmented lagrangian and least squares optimizers.

ddudt (approved):
What is the speed up with this change?

YigitElma:
> What is the speed up with this change?

It is hard to quantify. This basically saves us from taking the QR of the same thing multiple times in the inner while loop. For most of the problems inner while loop is iterated once, so no speed up there, but for those problems with multiple iterations there is some speed up.

YigitElma:
@ddudt Ok I think the total time saved can be found by  (Total nfev - total iteration)*(time a single QR takes) and this is 0 for our benchmark case. But usually when I run optimizations, there are 10 15 more function evaluations than total iterations(which is equivalent to total jacobian iterations)

Issue #1163, opened 2024-08-05
zeta_B computation is sometimes wrong


unalmis:
Sometimes when I rerun the test that compares everything is computed the same as master up to floating point errors, I get significant errors on `zeta_B`.

```sh
Not equal to tolerance rtol=1e-10, atol=1e-10
Parameterization: desc.magnetic_fields._core.OmnigenousField. Name: zeta_B.
Mismatched elements: 270 / 297 (90.9%)
Max absolute difference: 11.42397329
Max relative difference: 34.0466585
 x: array([ 0.      ,  0.550532,  1.076765,  1.557329,  1.976363,  2.325429,
        2.604526,  2.822094,  2.993993,  0.      ,  0.550532,  1.076765,
        1.557329,  1.976363,  2.325429,  2.604526,  2.822094,  2.993993,...
 y: array([0.      , 0.550532, 1.076765, 1.557329, 1.976363, 2.325429,
       2.604526, 2.822094, 2.993993, 0.      , 0.550532, 1.076765,
       1.557329, 1.976363, 2.325429, 2.604526, 2.822094, 2.993993,...
```

This error is not always reproducible. However,  it has even occurred on new branches, even when the file that the results are compared to  was regenerated immediately prior to running the test.

Therefore, either

1. `zeta_B` is not deterministic or subject to a numerical instability (silly)
2. Pickle is messing something up when the file is dumped
3. Some issue from changes to this test in #1027 ?

f0uriest:
Could they be the same modulo 2pi/NFP?

YigitElma:
Is this `test_compute_everthing`? #1027 only added `xyz` computation to the test, it didn't change logic or etc.

unalmis:
> Could they be the same modulo 2pi/NFP?

I checked: no. This is reproducible on my machine. Repeated calls to the test give 40/60 pass,fail split respectively, sample size 8. It is consistent in that when the failure occurs it is always the same result. I'm now leaning towards option 1.

> Is this `test_compute_everthing`? #1027 only added `xyz` computation to the test, it didn't change logic or etc.

Yes it looked fine when we merged. I think it was around time failure starts occurring.


```python
@register_compute_fun(
    name="zeta_B",
    label="\\zeta_{B}",
    units="rad",
    units_long="radians",
    description="Boozer toroidal angular coordinate",
    dim=1,
    params=[],
    transforms={},
    profiles=[],
    coordinates="rtz",
    data=["zeta", "nu"],
[...]

PR #1166, opened 2024-08-06
Fix bug in boozer coordinate computation for Omnigenous magnetic field

unalmis:
- [x] Move test compute everything to own file and exclude source grid quantities. from commit 8a8d9de4a993cf8565cba959fc1ff9e39f6d0152.
- [x] Remove xyz basis file in favor of correctness check to catch bug #1088 
- [x] Debug zeta_B computation. #1163 has been reproduced here (and on master by Dario). This issue was resolved after fixing the computation of zeta_B and theta_B for omnigenous fields.


tests/test_compute_everything.py
```diff
@@ -0,0 +1,238 @@
...
+            assert this_branch_data_xyz[p].keys() == names_xyz, (
+                f"Parameterization: {p}. Can't compute "
+                + f"{names_xyz - this_branch_data_xyz[p].keys()}."
+            )
+            error_xyz, update_master_data_xyz = _compare_against_master(
+                p,
+                this_branch_data_xyz,
+                master_data_xyz,
+                error_xyz,
+                update_master_data_xyz,
```
unalmis, line 226: my suggestion would be to just compute in `rpz` and `xyz` and check that `convert(xyz_data to rpz) == rpz_data`. Then just check if `rpz_data == master_rpz_data` .

dpanici: I like this

unalmis:
> I can't seem to reproduce #1163 when trying on a CPU compute node on della, though it actually fails with very slim margins for some of the current density and force compoents. You are running on linux locally I assume as well?

Yes. Fedora linux. CPU, jax 0.4.30, jaxlib 0.4.28.

dpanici:
> > I can't seem to reproduce #1163 when trying on a CPU compute node on della, though it actually fails with very slim margins for some of the current density and force compoents. You are running on linux locally I assume as well?
> 
> Yes. Fedora linux. CPU, jax 0.4.30, jaxlib 0.4.28.

wait I cannot read, yes I also get that `zeta_B` is off by large margins, AND it is not consistent (running it twice, one it failed and the other it did not). Using older versions of jax/jaxlib (0.4.14 for both)

dpanici:
take a look at the omni field in test compute everything, see if it looks weird, try changing params and see if it is more stable

dpanici:
@ddudt

f0uriest:
I do see some small "blobs" near theta=0 and theta=2pi
image

Not sure if this is a plotting artifact, or could be something weird with the coordinate mapping there?

ddudt:
> I do see some small "blobs" near theta=0 and theta=2pi 
> 
> Not sure if this is a plotting artifact, or could be something weird with the coordinate mapping there?

I think that is just an artifact of plotting with `tricontour`, although there was a small bug I just fixed that shouldn't change much.

ddudt:
I think I figured out the issue: it has to do with how `"zeta_B"` is an alias of `"theeta_B"`. These two quantities are always computed simultaneously -- you can't compute one without also getting the other. So we have the compute function with `name="theta_B"` and `aliases=["zeta_B"]`. This works fine if you always ask to compute both quantities or even only ask for `"theta_B"`. But for some reason, it looks like when you ask to only compute `"zeta_B"` that quantity in the compute data is being overwritten with the values for `"theta_B"`. 

I'm not sure what the solution is, but I'll let you figure that out.

unalmis:
> I think I figured out the issue: it has to do with how `"zeta_B"` is an alias of `"theeta_B"`. These two quantities are always computed simultaneously -- you can't compute one without also getting the other. So we have the compute function with `name="theta_B"` and `aliases=["zeta_B"]`. This works fine if you always ask to compute both quantities or even only ask for `"theta_B"`. But for some reason, it looks like when you ask to only compute `"zeta_B"` that quantity in the compute data is being overwritten with the values for `"theta_B"`.
> 
> I'm not sure what the solution is, but I'll let you figure that out.

Why are these aliases?
We compute
```python
@register_compute_fun(
    name="theta_B",
    label="(\\theta_{B},\\zeta_{B})",
    units="rad",
    units_long="radians",
    description="Boozer angular coordinates",
    dim=1,
    params=[],
    transforms={},
    profiles=[],
    coordinates="rtz",
    data=["alpha", "h"],
    aliases=["zeta_B"], # should remove this
    parameterization="desc.magnetic_fields._core.OmnigenousField",
)
....
    # solve for (theta_B,zeta_B) corresponding to (eta,alpha)
    booz = matrix @ jnp.vstack((data["alpha"], data["h"]))
    data["theta_B"] = booz[0, :]
    data["zeta_B"] = booz[1, :]
```

Unless `booz[1, :]` is the same as `booz[0, :]`, we shouldnot mark them as aliases.

rahulgaur104:
> I think that is just an artifact of plotting with tricontour, although there was a small bug I just fixed that shouldn't change much.

@ddudt That doesn't make sense. I plotted the same think with regular contour and it still persists. These is something else going on with those blobs at the ends.

f0uriest (dismissed):

ddudt:
> > I think I figured out the issue: it has to do with how `"zeta_B"` is an alias of `"theeta_B"`. These two quantities are always computed simultaneously -- you can't compute one without also getting the other. So we have the compute function with `name="theta_B"` and `aliases=["zeta_B"]`. This works fine if you always ask to compute both quantities or even only ask for `"theta_B"`. But for some reason, it looks like when you ask to only compute `"zeta_B"` that quantity in the compute data is being overwritten with the values for `"theta_B"`.
> > I'm not sure what the solution is, but I'll let you figure that out.
> 
> Why are these aliases? We compute
> 
> ```python
> @register_compute_fun(
>     name="theta_B",
>     label="(\\theta_{B},\\zeta_{B})",
>     units="rad",
>     units_long="radians",
>     description="Boozer angular coordinates",
>     dim=1,
>     params=[],
>     transforms={},
>     profiles=[],
>     coordinates="rtz",
>     data=["alpha", "h"],
>     aliases=["zeta_B"], # should remove this
>     parameterization="desc.magnetic_fields._core.OmnigenousField",
> )
> ....
>     # solve for (theta_B,zeta_B) corresponding to (eta,alpha)
>     booz = matrix @ jnp.vstack((data["alpha"], data["h"]))
>     data["theta_B"] = booz[0, :]
>     data["zeta_B"] = booz[1, :]
> ```
> 
> Unless `booz[1, :]` is the same as `booz[0, :]`, we shouldnot mark them as aliases. Otherwise they will get overridden by one another

They are aliases in the sense that if you ask for one of the two quantities, you get both. The problem is with the alias logic outside of the compute function, not in the compute function itself. 

I don't like this solution -- it creates redundant code and computations. Maybe we shouldn't call them aliases, but I'm sure we can think of another way to have a single compute function return multiple variables.

desc/compute/_omnigenity.py
```diff
@@ -439,12 +439,59 @@ def _omni_angle(params, transforms, profiles, data, **kwargs):
     profiles=[],
     coordinates="rtz",
     data=["alpha", "h"],
-    aliases=["zeta_B"],
     parameterization="desc.magnetic_fields._core.OmnigenousField",
     helicity="tuple: Type of quasisymmetry, (M,N). Default (1,0)",
     iota="float: Value of rotational transform on the Omnigenous surface. Default 1.0",
 )
-def _omni_map(params, transforms, profiles, data, **kwargs):
+def _omni_map_theta_B(params, transforms, profiles, data, **kwargs):
```
ddudt, line 446: This is duplicate code and computations.

ddudt (changes requested):

unalmis:
> They are aliases in the sense that if you ask for one of the two quantities, you get both. The problem is with the alias logic outside of the compute function, not in the compute function itself.

Marking some quantity `A` as an alias of `B` means that `A = B`.

> I don't like this solution -- it creates redundant code and computations.

The redundancy is only cosmetic. [The computation is only done once](https://github.com/PlasmaControl/DESC/blob/44bac68a99782d9f0e89391ea489bb50c645c974/desc/compute/utils.py#L167).

> Maybe we shouldn't call them aliases, but I'm sure we can think of another way to have a single compute function return multiple variables.

This would require plumbing that is outside the scope of this PR. In particular, the `register_compute_fun` decorator would need to register multiple quantities with multiple descriptions, dimensions, etc.

There is a failing regression test now. My guess is it's due to the plotting change and not the compute function change? Could you look into this sometime

Before:
test_plot_omnigenous_field

After:
test_plot_omnigenous_field

ddudt:
I think the following will work. It's similar to your change, but avoids needing the duplicate code: 

```
@register_compute_fun(
    name="theta_B",
    label="\\theta_{B}",
    units="rad",
    units_long="radians",
    description="Boozer poloidal coordinate",
    dim=1,
    params=[],
    transforms={},
    profiles=[],
    coordinates="rtz",
    data=["alpha", "h"],
    parameterization="desc.magnetic_fields._core.OmnigenousField",
    helicity="tuple: Type of quasisymmetry, (M,N). Default (1,0)",
    iota="float: Value of rotational transform on the Omnigenous surface. Default 1.0",
)
def _omni_map_theta(params, transforms, profiles, data, **kwargs):
    M, N = kwargs.get("helicity", (1, 0))
    iota = kwargs.get("iota", 1)

    # coordinate mapping matrix from (alpha,h) to (theta_B,zeta_B)
    # need a bunch of wheres to avoid division by zero causing NaN in backward pass
    # this is fine since the incorrect values get ignored later, except in OT or OH
    # where fieldlines are exactly parallel to |B| contours, but this is a degenerate
    # case of measure 0 so this kludge shouldn't affect things too much.
    mat_OP = jnp.array(
        [[N, iota / jnp.where(N == 0, 1, N)], [0, 1 / jnp.where(N == 0, 1, N)]]
    )
    mat_OT = jnp.array([[0, -1], [M, -1 / jnp.where(iota == 0, 1.0, iota)]])
    den = jnp.where((N - M * iota) == 0, 1.0, (N - M * iota))
    mat_OH = jnp.array([[N, M * iota / den], [M, M / den]])
    matrix = jnp.where(
        M == 0,
        mat_OP,
        jnp.where(
            N == 0,
            mat_OT,
            mat_OH,
        ),
    )

    # solve for (theta_B,zeta_B) corresponding to (eta,alpha)
    booz = matrix @ jnp.vstack((data["alpha"], data["h"]))
    data["theta_B"] = booz[0, :]
    data["zeta_B"] = booz[1, :]
    return data


@register_compute_fun(
    name="zeta_B",
    label="\\zeta_{B}",
    units="rad",
    units_long="radians",
    description="Boozer toroidal coordinate",
    dim=1,
    params=[],
    transforms={},
    profiles=[],
    coordinates="rtz",
    data=["theta_B"],
    parameterization="desc.magnetic_fields._core.OmnigenousField",
)
def _omni_map_zeta(params, transforms, profiles, data, **kwargs):
    # some dummy line of unused code that references the dependency theta_B
    _ = data["theta_B"]
    return data
```

rahulgaur104 (approved):
Omnigenity gives me the same results as before. Approving!

ddudt (approved):

f0uriest (approved):

desc/plotting.py
```diff
@@ -2756,30 +2756,22 @@ def plot_boozer_surface(
...
+        }
         grid_compute = _get_grid(**grid_kwargs)
     if grid_plot is None:
         grid_kwargs = {
             "rho": rho,
             "theta": 91,
             "zeta": 91,
             "NFP": thing.NFP,
-            "endpoint": True,
+            "endpoint": eq_switch,
```
dpanici, line 2774: why can't there be an endpoint in the plot grid for the Omnigenous field?

dpanici (approved):

dpanici:
If this resolves any of the issues on boozer stuff, make sure to link them to this

desc/plotting.py
ddudt: The old code used `grid_compute` for the Omnigenous field, which also had `endpoint=False` so this doesn't change that. The only thing this changes is to correctly plot on `grid_plot` instead of the `grid_compute`. 

In theory I think it should be able to use `endpoint=True`, but it was giving a worse looking plot for the test. I suspect it's an issue with `tricontour`.

PR #1168, opened 2024-08-06
`HeatingPower` objective

ddudt:
New objective to minimize the required heating power, determine by the ISS04 energy confinement time scaling.

### DIFF
--- a/desc/compute/_field.py
+++ b/desc/compute/_field.py
@@ -2221,6 +2221,28 @@ def _B_sub_zeta_rz(params, transforms, profiles, data, **kwargs):
 
 
 @register_compute_fun(
+    name="B0",
+    label="B_0",
+    units="T",
+    units_long="Tesla",
+    description="Average magnitude of magnetic field on the magnetic axis",
+    dim=0,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="",
+    data=["|B|", "sqrt(g)"],
+)
+def _B0(params, transforms, profiles, data, **kwargs):
+    # indices of the innermost flux surface
+    idx = jnp.where(transforms["grid"].inverse_rho_idx == 0)[0]
+    data["B0"] = jnp.sum(
+        data["|B|"][idx] * data["sqrt(g)"][idx] * transforms["grid"].weights[idx]
+    ) / jnp.sum(data["sqrt(g)"][idx] * transforms["grid"].weights[idx])
+    return data
+
+
+@register_compute_fun(
     name="|B|^2",
     label="|\\mathbf{B}|^{2}",
     units="T^2",
--- a/desc/compute/_field.py
+++ b/desc/compute/_field.py
@@ -27,7 +27,7 @@ from .utils import (
 
 
 @register_compute_fun(
-    name="B0",
+    name="psi_r/sqrt(g)",
     label="\\psi' / \\sqrt{g}",
     units="T \\cdot m^{-1}",
     units_long="Tesla / meter",
@@ -40,8 +40,8 @@ from .utils import (
     data=["psi_r", "sqrt(g)"],
     axis_limit_data=["psi_rr", "sqrt(g)_r"],
 )
-def _B0(params, transforms, profiles, data, **kwargs):
-    data["B0"] = transforms["grid"].replace_at_axis(
+def _psi_r_over_sqrtg(params, transforms, profiles, data, **kwargs):
+    data["psi_r/sqrt(g)"] = transforms["grid"].replace_at_axis(
         safediv(data["psi_r"], data["sqrt(g)"]),
         lambda: safediv(data["psi_rr"], data["sqrt(g)_r"]),
     )
@@ -77,10 +77,10 @@ def _B_sup_rho(params, transforms, profiles, data, **kwargs):
     transforms={},
     profiles=[],
     coordinates="rtz",
-    data=["B0", "iota", "lambda_z", "omega_z"],
+    data=["psi_r/sqrt(g)", "iota", "lambda_z", "omega_z"],
 )
 def _B_sup_theta(params, transforms, profiles, data, **kwargs):
-    data["B^theta"] = data["B0"] * (
+    data["B^theta"] = data["psi_r/sqrt(g)"] * (
         data["iota"] * data["omega_z"] + data["iota"] - data["lambda_z"]
     )
     return data
@@ -97,10 +97,10 @@ def _B_sup_theta(params, transforms, profiles, data, **kwargs):
     transforms={},
     profiles=[],
     coordinates="rtz",
-    data=["B0", "iota", "lambda_t", "omega_t"],
+    data=["psi_r/sqrt(g)", "iota", "lambda_t", "omega_t"],
 )
 def _B_sup_zeta(params, transforms, profiles, data, **kwargs):
-    data["B^zeta"] = data["B0"] * (
+    data["B^zeta"] = data["psi_r/sqrt(g)"] * (
         -data["iota"] * data["omega_t"] + data["lambda_t"] + 1
     )
     return data
@@ -181,7 +181,7 @@ def _B_Z(params, transforms, profiles, data, **kwargs):
 
 
 @register_compute_fun(
-    name="B0_r",
+    name="(psi_r/sqrt(g))_r",
     label="\\partial_{\\rho} (\\psi' / \\sqrt{g})",
     units="T \\cdot m^{-1}",
     units_long="Tesla / meter",
@@ -194,8 +194,8 @@ def _B_Z(params, transforms, profiles, data, **kwargs):
     data=["psi_r", "psi_rr", "sqrt(g)", "sqrt(g)_r"],
     axis_limit_data=["psi_rrr", "sqrt(g)_rr"],
 )
-def _B0_r(params, transforms, profiles, data, **kwargs):
-    data["B0_r"] = transforms["grid"].replace_at_axis(
+def _psi_r_over_sqrtg_r(params, transforms, profiles, data, **kwargs):
+    data["(psi_r/sqrt(g))_r"] = transforms["grid"].replace_at_axis(
         safediv(
             data["psi_rr"] * data["sqrt(g)"] - data["psi_r"] * data["sqrt(g)_r"],
             data["sqrt(g)"] ** 2,
@@ -223,8 +223,8 @@ def _B0_r(params, transforms, profiles, data, **kwargs):
     profiles=[],
     coordinates="rtz",
     data=[
-        "B0",
-        "B0_r",
+        "psi_r/sqrt(g)",
+        "(psi_r/sqrt(g))_r",
         "iota",
         "iota_r",
         "lambda_rz",
@@ -234,12 +234,12 @@ def _B0_r(params, transforms, profiles, data, **kwargs):
     ],
 )
 def _B_sup_theta_r(params, transforms, profiles, data, **kwargs):
-    data["B^theta_r"] = data["B0"] * (
+    data["B^theta_r"] = data["psi_r/sqrt(g)"] * (
         data["iota"] * data["omega_rz"]
         + data["iota_r"] * data["omega_z"]
         + data["iota_r"]
         - data["lambda_rz"]
-    ) + data["B0_r"] * (
+    ) + data["(psi_r/sqrt(g))_r"] * (
         data["iota"] * data["omega_z"] + data["iota"] - data["lambda_z"]
     )
     return data
@@ -260,8 +260,8 @@ def _B_sup_theta_r(params, transforms, profiles, data, **kwargs):
     profiles=[],
     coordinates="rtz",
     data=[
-        "B0",
-        "B0_r",
+        "psi_r/sqrt(g)",
+        "(psi_r/sqrt(g))_r",
         "iota",
         "iota_r",
         "lambda_rt",
@@ -271,11 +271,13 @@ def _B_sup_theta_r(params, transforms, profiles, data, **kwargs):
     ],
 )
 def _B_sup_zeta_r(params, transforms, profiles, data, **kwargs):
-    data["B^zeta_r"] = data["B0"] * (
+    data["B^zeta_r"] = data["psi_r/sqrt(g)"] * (
         -data["iota"] * data["omega_rt"]
         - data["iota_r"] * data["omega_t"]
         + data["lambda_rt"]
-    ) + data["B0_r"] * (-data["iota"] * data["omega_t"] + data["lambda_t"] + 1)
+    ) + data["(psi_r/sqrt(g))_r"] * (
+        -data["iota"] * data["omega_t"] + data["lambda_t"] + 1
+    )
     return data
 
 
@@ -312,7 +314,7 @@ def _B_r(params, transforms, profiles, data, **kwargs):
 
 
 @register_compute_fun(
-    name="B0_t",
+    name="(psi_r/sqrt(g))_t",
     label="\\partial_{\\theta} (\\psi' / \\sqrt{g})",
     units="T \\cdot m^{-1}",
     units_long="Tesla / meter",
@@ -325,8 +327,8 @@ def _B_r(params, transforms, profiles, data, **kwargs):
     data=["psi_r", "sqrt(g)", "sqrt(g)_t"],
     axis_limit_data=["psi_rr", "sqrt(g)_r", "sqrt(g)_rt"],
 )
-def _B0_t(params, transforms, profiles, data, **kwargs):
-    data["B0_t"] = transforms["grid"].replace_at_axis(
+def _psi_r_over_sqrtg_t(params, transforms, profiles, data, **kwargs):
+    data["(psi_r/sqrt(g))_t"] = transforms["grid"].replace_at_axis(
         safediv(-data["psi_r"] * data["sqrt(g)_t"], data["sqrt(g)"] ** 2),
         lambda: safediv(-data["psi_rr"] * data["sqrt(g)_rt"], data["sqrt(g)_r"] ** 2),
     )
@@ -347,12 +349,20 @@ def _B0_t(params, transforms, profiles, data, **kwargs):
     transforms={},
     profiles=[],
     coordinates="rtz",
-    data=["B0", "B0_t", "iota", "lambda_tz", "lambda_z", "omega_tz", "omega_z"],
+    data=[
+        "psi_r/sqrt(g)",
+        "(psi_r/sqrt(g))_t",
+        "iota",
+        "lambda_tz",
+        "lambda_z",
+        "omega_tz",
+        "omega_z",
+    ],
 )
 def _B_sup_theta_t(params, transforms, profiles, data, **kwargs):
-    data["B^theta_t"] = data["B0"] * (
+    data["B^theta_t"] = data["psi_r/sqrt(g)"] * (
         data["iota"] * data["omega_tz"] - data["lambda_tz"]
-    ) + data["B0_t"] * (
+    ) + data["(psi_r/sqrt(g))_t"] * (
         data["iota"] * data["omega_z"] + data["iota"] - data["lambda_z"]
     )
     return data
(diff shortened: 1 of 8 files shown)


desc/compute/_field.py
```diff
@@ -27,7 +27,7 @@
 
 
 @register_compute_fun(
-    name="B0",
+    name="psi_r/sqrt(g)",
```
ddudt, line 30: I know we don't normally like to rename existing variables, but I will argue that in this case it is desirable. The old quantity `"B0"` was only ever used internally for computing $B^\zeta$ and is very unlikely to ever be computed by the user directly. `"B0"` is now the magnetic field strength on axis, which is what this variable is commonly used for in the literature.

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,211 @@
...
+        normalization. Note: Has no effect for this objective.
+    deriv_mode : {"auto", "fwd", "rev"}
+        Specify how to compute jacobian matrix, either forward mode or reverse mode AD.
+        "auto" selects forward or reverse mode based on the size of the input and output
+        of the objective. Has no effect on self.grad or self.hess which always use
+        reverse mode and forward over reverse mode respectively.
+    f_ren : float, optional
+        ISS04 renormalization factor. Default = 1.
+    H_ISS04 : float, optional
+        ISS04 confinement enhancement factor. Default = 1.
```
ddudt, line 56: We could probably combine these into a single term, although I think they are technically separate fudge factors in the literature.

desc/compute/_field.py
```diff
@@ -2160,6 +2220,28 @@ def _B_sub_zeta_rz(params, transforms, profiles, data, **kwargs):
...
+    dim=0,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="",
+    data=["|B|", "sqrt(g)"],
+)
+def _B0(params, transforms, profiles, data, **kwargs):
+    # indices of the innermost flux surface
+    idx = jnp.where(transforms["grid"].inverse_rho_idx == 0)[0]
```
dpanici, line 2238: check if this works with JAX and see surface integral code if it does not

unalmis: `sum((transforms["grid"].inverse_rho_idx == 0) * integrand)`

desc/compute/_field.py
```diff
@@ -2203,6 +2263,28 @@ def _B_sub_zeta_rz(params, transforms, profiles, data, **kwargs):
...
+    coordinates="",
+    data=["|B|", "sqrt(g)"],
+)
+def _B_mag_axis(params, transforms, profiles, data, **kwargs):
+    # mask of indices of the innermost flux surface
+    mask = transforms["grid"].inverse_rho_idx == 0
+    data["<|B|>_axis"] = jnp.sum(
+        mask * data["|B|"] * data["sqrt(g)"] * transforms["grid"].weights
+    ) / jnp.sum(mask * data["sqrt(g)"] * transforms["grid"].weights)
+    return data
```
unalmis, line 2285: maybe also
```python
sqrt_g = transforms["grid"].replace_at_axis(
    data["sqrt(g)"], lambda: data["sqrt(g)_r"], copy=True
)
```

desc/compute/_field.py
```diff
@@ -2203,6 +2263,32 @@ def _B_sub_zeta_rz(params, transforms, profiles, data, **kwargs):
...
+    description="Average magnitude of magnetic field on the magnetic axis",
+    dim=0,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="",
+    data=["|B|", "sqrt(g)"],
+    axis_limit_data=["sqrt(g)_r"],
+)
+def _B_mag_axis(params, transforms, profiles, data, **kwargs):
```
f0uriest, line 2280: does this need a `resolution_requirement`?

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,216 @@
...
+            "desc.equilibrium.equilibrium.Equilibrium",
+            self._data_keys,
+            params=params,
+            transforms=constants["transforms"],
+            profiles=constants["profiles"],
+            gamma=constants["gamma"],
+        )
+
+        P = (
+            (data["W_p"] / -1e6)  # MJ
```
f0uriest, line 186: dividing by a negative here might cause issues if gamma is nonzero. Might be better to just take `abs(W_p/1e6)`

desc/compute/_profiles.py
```diff
@@ -686,6 +744,27 @@ def _gradbeta_a(params, transforms, profiles, data, **kwargs):
...
+    description="Rotational transform (normalized by 2pi) at rho=2/3",
+    dim=0,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="",
+    data=["rho", "iota"],
+    method="str: Interpolation method. Default 'cubic'.",
+)
+def _iota_23(params, transforms, profiles, data, **kwargs):
```
f0uriest, line 761: I'm not crazy about this as its own compute function. I feel like it would make more sense to have the power from iss04 as a compute function that uses iota and interpolates locally to get the 2/3 value

tests/test_profiles.py
```diff
@@ -507,3 +505,28 @@ def test_kinetic_pressure(self):
...
+
+    @pytest.mark.unit
+    def test_line_average(self):
+        """Test that line-averaged profiles are correct."""
+        ne0 = 3e19
+        Te0 = 2e3
+        ne = PowerSeriesProfile(2 * ne0 * np.array([1, -1]), modes=[0, 1])
+        Te = PowerSeriesProfile(2 * Te0 * np.array([1, -1]), modes=[0, 1])
+
+        with pytest.raises(UserWarning):
```
f0uriest, line 517: what warning does this raise?

desc/compute/_field.py
ddudt: It's a scalar quantity (`dim=0`) so it will always be computed with a `QuadratureGrid`, but I'll add `"z"` as a resolution requirement

tests/test_profiles.py
ddudt: The profiles are not even power series. I made them linear so the average was simpler

desc/objectives/_confinement.py
ddudt: I removed `f_ren`

desc/compute/_field.py
unalmis: We only automatically compute quantities with associated best grids when `resolution_requirement` is set and `coordinates` is such that it can be interpolated back in a lossless manner.

You'll need poloidal resolution on the innermost flux surface too. (FYI the derivative of the 3d volume jacobian $\sqrt(g)_r$ will be single valued in theta on axis, but the derivatives of the 2d ones will not).

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,198 @@
...
+from desc.compute.utils import _compute as compute_fun
+from desc.compute.utils import get_profiles, get_transforms
+from desc.grid import QuadratureGrid
+from desc.utils import Timer, errorif
+
+from .normalization import compute_scaling_factors
+from .objective_funs import _Objective
+
+
+class HeatingPower(_Objective):
```
f0uriest, line 12: I'd prefer something like `HeatingPowerISS04` to distinguish from other possible objectives for heating power we may eventually add

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,198 @@
...
+    """Heating power required by the ISS04 energy confinement time scaling.
+
+    tau_E^ISS04 = 0.134 a^2.28 R^0.64 P^-0.61 n_e^0.54 B^0.84 iota^0.41 (s)
```
f0uriest, line 15: Maybe include that `tau_E = W_p/P`

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,198 @@
...
+    bounds : tuple of {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
+        Both bounds must be broadcastable to to Objective.dim_f.
+        Defaults to ``target=0``.
+    weight : {float, ndarray}, optional
+        Weighting to apply to the Objective, relative to other Objectives.
+        Must be broadcastable to to Objective.dim_f
+    normalize : bool, optional
+        Whether to compute the error in physical units or non-dimensionalize.
+        Has no effect for this objective.
```
f0uriest, line 40: I think it does

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,198 @@
...
+    normalize_target : bool, optional
+        Whether target and bounds should be normalized before comparing to computed
+        values. If `normalize` is `True` and the target is in physical units,
+        this should also be set to True. Note: Has no effect for this objective.
```
f0uriest, line 44: see above

desc/compute/_equil.py
```diff
@@ -802,3 +803,37 @@ def _beta_volpol(params, transforms, profiles, data, **kwargs):
...
+    profiles=[],
+    coordinates="",
+    data=["a", "iota", "rho", "R0", "W_p", "<ne>_rho", "<|B|>_axis"],
+    method="str: Interpolation method. Default 'cubic'.",
+    H_ISS04="float: ISS04 confinement enhancement factor. Default 1.",
+)
+def _P_ISS04(params, transforms, profiles, data, **kwargs):
+    rho = transforms["grid"].compress(data["rho"], surface_label="rho")
+    iota = transforms["grid"].compress(data["iota"], surface_label="rho")
+    iota_23 = interp1d(2 / 3, rho, iota, method=kwargs.get("method", "cubic"))
```
unalmis, line 826: If your objective already computes `iota_r`, you can do this to give more accurate result
```python
fx = {}
if "iota_r" in data:
  fx["fx"] = transforms["grid"].compress(data["iota_r"])  # noqa: unused dependency
iota_23 = interp1d(2 / 3, rho, iota, method=kwargs.get("method", "cubic"), **fx)
```

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,198 @@
...
+            "gamma": self._gamma,
+        }
+
+        timer.stop("Precomputing transforms")
+        if verbose > 1:
+            timer.disp("Precomputing transforms")
+
+        if self._normalize:
+            scales = compute_scaling_factors(eq)
+            self._normalization = scales["W"]
```
f0uriest, line 147: Do you want to add a scaling factor for `W_p` in this PR?

desc/compute/_profiles.py
```diff
@@ -216,6 +216,26 @@ def _Te_rr(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="<ne>_rho",
+    label="\\bar{n}_e",
+    units="m^{-3}",
+    units_long="1 / cubic meters",
+    description="Line-averaged electron density",
```
unalmis, line 224: Can you change to 
```python
label="\\bar{n}_e = \\int n_e d\\rho"  # Simple radial average〈nₑ〉ᵨ = ∫ nₑ dρ 
```
to emphasize your average does not include radial integration weight like $\Vert e_{\rho} \Vert$ and then average.

desc/compute/_profiles.py
```diff
@@ -216,6 +216,26 @@ def _Te_rr(params, transforms, profiles, data, **kwargs):
...
+    dim=0,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="",
+    data=["ne"],
```
unalmis, line 230: radial resolution requirement

desc/compute/_profiles.py
```diff
@@ -216,6 +216,26 @@ def _Te_rr(params, transforms, profiles, data, **kwargs):
...
+)
+def _bar_ne(params, transforms, profiles, data, **kwargs):
+    data["<ne>_rho"] = jnp.sum(data["ne"] * transforms["grid"].spacing[:, 0]) / jnp.sum(
+        transforms["grid"].spacing[:, 0]
+    )
```
unalmis, line 235: Can you add comment explaining your assumptions on the grid?
If you're just doing a simple average over $\rho$, don't you want to require a tensor product grid and do
```python
# or reshape in fortran order for LinearGrid etc.
data["<ne>_rho"] = (data["ne"] * grid.spacing[:, 0]).reshape(num_rho, num_theta, num_zeta).sum(axis=0).mean() 
```

unalmis: or if $n_e$ is assume to be radial profile then compress and then integrate

desc/objectives/_confinement.py
ddudt: I already added it in #1192 so let's merge that first and then I update this

desc/compute/_profiles.py
ddudt: Usually it will only be a 1D radial profile, but I suppose we should allow for 3D profiles. Is there any way we can handle general grids like `ConcentricGrid`? If not, how do we specify in the data index that it has to be a tensor product grid?

unalmis: If you have a 3d profile, I think what you want to do is like

```python
data["<ne>_rho"] = np.dot(
  surface_averages(grid, data["ne"], data["|e_theta x e_zeta|"], expand_out=False),
  grid.compress(grid.spacing[:, 0])
)
```
my choice of `|e_theta x e_zeta|` as the Jacobian as opposed to `sqrt(g)` is based off how we compute `data["A"]` as 
```python
# Simple toroidal average A₀ = ∫ A(ζ) dζ / (2π) matches the convention for the
# average major radius R₀ = ∫ R(ρ=0) dζ / (2π)
```
There, when we compute $A(z)$, we use the 2d jacobian `|e_rho x e_theta|` and not `sqrt(g)`.

ddudt:
> I think you'll also need to update master_compute_data with the new variables and name changes

I didn't update this yet but the tests are passing. Is that a bug with our test?

desc/compute/_profiles.py
ddudt: Thanks, I updated to your suggestion

desc/compute/_equil.py
```diff
@@ -802,3 +803,42 @@ def _beta_volpol(params, transforms, profiles, data, **kwargs):
...
+    profiles=[],
+    coordinates="",
+    data=["a", "iota", "rho", "R0", "W_p", "<ne>_rho", "<|B|>_axis"],
+    method="str: Interpolation method. Default 'cubic'.",
+    H_ISS04="float: ISS04 confinement enhancement factor. Default 1.",
+)
+def _P_ISS04(params, transforms, profiles, data, **kwargs):
+    rho = transforms["grid"].compress(data["rho"], surface_label="rho")
+    iota = transforms["grid"].compress(data["iota"], surface_label="rho")
+    kwargs = {}
```
unalmis, line 826: **kwargs is emptied out here i think

ddudt: Good catch, thanks

desc/compute/_profiles.py
f0uriest: including `|e_rho x e_theta|` means this has units of 1/m not 1/m^3.

Usually the line average is something like $\int f(x) dx / \int dx$ where $x$ is a real space coordinate like R or Z. I think the least bad approximation of that in flux coordinates would be something like `dx ~ <sqrt(g^rr)>_fsa dr`

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,197 @@
...
+    loss_function : {None, 'mean', 'min', 'max'}, optional
+        Loss function to apply to the objective values once computed. This loss function
+        is called on the raw compute value, before any shifting, scaling, or
+        normalization. Note: Has no effect for this objective.
+    deriv_mode : {"auto", "fwd", "rev"}
+        Specify how to compute jacobian matrix, either forward mode or reverse mode AD.
+        "auto" selects forward or reverse mode based on the size of the input and output
+        of the objective. Has no effect on self.grad or self.hess which always use
+        reverse mode and forward over reverse mode respectively.
+    H_ISS04 : float, optional
```
f0uriest, line 53: can you include this in the formula above

desc/objectives/_confinement.py
```diff
@@ -0,0 +1,197 @@
...
+from desc.utils import Timer, errorif
+
+from .normalization import compute_scaling_factors
+from .objective_funs import _Objective
+
+
+class HeatingPowerISS04(_Objective):
+    """Heating power required by the ISS04 energy confinement time scaling.
+
+    tau_E = W_p / P = 0.134 a^2.28 R^0.64 P^-0.61 n_e^0.54 B^0.84 iota^0.41 (s)
```
f0uriest, line 15: using unicode for this might be cleaner

desc/compute/_profiles.py
unalmis: It's a surface average not an integral, so the units do not change

unalmis:
> > I think you'll also need to update master_compute_data with the new variables and name changes
> 
> I didn't update this yet but the tests are passing. Is that a bug with our test?

The test will fail if stuff computed on this branch differ from the stuff computable on master.

If you add new stuff on this branch, and the test doesn't fail, it will automatically update the file with the new stuff. What it can't do is automatically git commit the new file. There's probably a way to get the CI to automatically generate a new commit in that case...

desc/compute/_profiles.py
dpanici: just do vol avg of ne

ddudt:
Changed the density to use the volume-average instead of a line-average and updated the master compute data, so this PR should be good to go now

desc/compute/_profiles.py
```diff
@@ -216,6 +216,27 @@ def _Te_rr(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="<ne>_vol",
+    label="\\lange n_e \\rangle_{vol}",
```
unalmis, line 221:
```suggestion
    label="\\langle n_e \\rangle_{vol}",

File: tests/test_profiles.py [hunk 22]
```diff
@@ -510,6 +508,19 @@ def test_kinetic_pressure(self):
         np.testing.assert_allclose(data1["p"], data2["p"])
         np.testing.assert_allclose(data1["p_r"], data2["p_r"])
 
+    @pytest.mark.unit
+    def test_line_average(self):
```
f0uriest, line 512: I think this needs to be removed

f0uriest (approved):

desc/compute/_equil.py
```diff
@@ -803,3 +804,42 @@ def _beta_volpol(params, transforms, profiles, data, **kwargs):
...
+        / (
+            0.134
+            * data["a"] ** 2.28  # m
+            * data["R0"] ** 0.64  # m
+            * (data["<ne>_vol"] / 1e19) ** 0.54  # 1/m^3 -> 1e19/m^3
+            * data["<|B|>_axis"] ** 0.84  # T
+            * iota_23**0.41
+            * kwargs.get("H_ISS04", 1)
+        )
+    ) ** (1 / 0.39)
```
unalmis, line 844: Doesn't τₑ ∼ P⁻⁰ᐧ⁶¹ nₑ⁻ᐧ⁵⁴ ⟹ P ∼ ( nₑ⁰ᐧ⁵⁴ / τₑ)^(1/0.61)?

unalmis (changes requested):

desc/compute/_equil.py
unalmis: Or actually τₑ ∼ P⁻⁰ᐧ⁶¹ nₑ⁻ᐧ⁵⁴ ⟹ P ∼ (1 / τₑ  nₑ⁰ᐧ⁵⁴ )^(1/0.61)?

desc/compute/_equil.py
```diff
@@ -803,3 +804,42 @@ def _beta_volpol(params, transforms, profiles, data, **kwargs):
...
+            data["iota_r"]
+        )  # noqa: unused dependency
+    iota_23 = interp1d(2 / 3, rho, iota, method=kwargs.get("method", "cubic"), **fx)
+    data["P_ISS04"] = 1e6 * (  # MW -> W
+        jnp.abs(data["W_p"] / 1e6)  # J -> MJ
+        / (
+            0.134
+            * data["a"] ** 2.28  # m
+            * data["R0"] ** 0.64  # m
+            * (data["<ne>_vol"] / 1e19) ** 0.54  # 1/m^3 -> 1e19/m^3
```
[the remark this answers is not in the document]
unalmis: actually this is fine nevermind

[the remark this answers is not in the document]
unalmis: actually i have no issue with this. sorry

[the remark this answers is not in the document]
ddudt: image

n_e has a positive power. Maybe you thought the bar above the n (to denote line-averaged) was a negative sign?

unalmis (approved):

Issue #1174, opened 2024-08-12
Bug in math for surface computations


unalmis:
Let $R$ be a position vector and $x^i$ be the coordinates of a curvilinear system and $y^j$ those of an orthonormal system.
Then $\hat{y}^j$ can be defined unambiguously

$$\hat{y}^j = \frac{\partial R}{\partial y^j} / \Vert \frac{\partial R}{\partial y^j} \Vert $$

Now, the relation below does not follow (hence the $\neq$)

$$\frac{\partial R}{\partial x^i}  \cdot \hat{y}^j \neq \frac{\partial y^j}{\partial x^i}$$

However, DESC assumed it does for all surface computations. (Note that a sufficient condition for the relation to hold is if the basis vectors associated with $y^j$ for all $j$ are orthonormal and constant, which is not true in a cylindrical system.) 

After fixing this, here are some of the quantities that significantly differ once corrected.

```
Not equal to tolerance rtol=1e-10, atol=1e-10
Parameterization: desc.geometry.surface.FourierRZToroidalSurface. Name: phi_z.
Mismatched elements: 225 / 225 (100%)
Max absolute difference: 10.2
Max relative difference: 0.91071429
 x: array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,...
 y: array([11.2     , 11.113545, 10.869131, 10.509017, 10.095472,  9.7     ,
        9.390983,  9.221852,  9.221852,  9.390983,  9.7     , 10.095472,
       10.509017, 10.869131, 11.113545, 11.182709, 11.096255, 10.85184 ,...

Not equal to tolerance rtol=1e-10, atol=1e-10
Parameterization: desc.geometry.surface.FourierRZToroidalSurface. Name: e_phi|r,t.
Mismatched elements: 660 / 675 (97.8%)
Max absolute difference: 10.2
Max relative difference: 10.2
 x: array([[ 0.      , 11.2     , -0.2     ],
       [ 0.      , 11.113545, -0.2     ],
       [ 0.      , 10.869131, -0.2     ],...
 y: array([[ 0.      ,  1.      , -0.017857],
[...]

Issue #1127, opened 2024-07-12
Surfaces can compute more than they should


unalmis:
We have things in the list of variables that we say we can compute on certain objects, but shouldn't be able. One could be misled that a quantity not typically associated with their surface was computed somehow by integrating on the boundary, or just by sampling the given surface.

If a quantity cannot be computed on an object, we should just not have a compute function for that object instead of setting dummy zero values.

Stetting dummy values causes bugs. For example `n_rho` defined as $\nabla \rho / \Vert \nabla \rho \Vert$ should not be defined on a ZernikeRZToroidalSection. But we say it is, and give it a dummy value of 0. Now, the area of that cross-section is well-defined, however we compute that area using `n_rho`, and since that dependency is a dummy for this surface, it shouldn't give the correct value.

unalmis:
Remove `@pytest.mark.xfail(reason="GitHub issue 1127.")` from `tests/test_configuration.py::TestGetSurfaces::test_get_zeta_surface` once resolved.

dpanici:
Fix for this is to remove the `Surface` level implementations of the compute quantities in question, and implement them only for the subclasses that actually make sense

dpanici:
Would resolution of this issue make sense to happen on #1094 ? @unalmis

unalmis:
Either there or separate is fine

dpanici:
I think I will do it separate for now. That branch still has some things that have to be fixed it seems and it requires this to be done anyways

dpanici:
I have a PR I think addressing this now. Basically,

- For "desc.geometry.surface.FourierRZToroidalSurface" , the radial derivative is undefined (is a constant rho surface so has no notion of how things change off-surface), so any $\partial_{\rho}$ quantity is now undefined
  - also has its own implementation of `n_rho` which has no axis limit defined (avoiding errors from the data index when it tried to build `n_rho` dependencies for the surface, and saw that the surface has no radial deriv defined which are needed for the axis limit of `n_rho`

"desc.geometry.surface.ZernikeRZToroidalSection" , the toroidal derivative is undefined (is a constant zeta surface so has no notion of how things change off-surface), so any $\partial_{\zeta}$ quantity is now undefined
[...]

PR #1175, opened 2024-08-12
Resolve Github issue #1174, bug in surface computations

unalmis:


desc/compute/_surface.py
unalmis, line 1: These weren't correct

desc/compute/_core.py
```diff
@@ -3037,151 +2837,6 @@ def _theta_PEST_t(params, transforms, profiles, data, **kwargs):
     return data
 
 
-@register_compute_fun(
-    name="theta_PEST_tt",
```
f0uriest, line 3041: I would prefer to keep stuff like this as independent quantities. For stuff like `g_rt` and `g_tr`, those are the same both physically and numerically, independent of how we parameterize or discretize things.

Stuff like `theta_PEST_tt` and `lambda_tt` are actually different things physically, its just that they happen to the same numerically because we define `theta_PEST = theta + lambda`, but in theory other relations are possible and equally valid physically.

desc/compute/_core.py
```diff
@@ -3205,66 +2860,40 @@ def _zeta(params, transforms, profiles, data, **kwargs):
...
-        "desc.equilibrium.equilibrium.Equilibrium",
-        "desc.geometry.core.Surface",
-    ],
+    coordinates="rtz",
+    data=["zeta", "omega"],
 )
-def _zeta_t(params, transforms, profiles, data, **kwargs):
-    data["zeta_t"] = data["0"]
+def _phi(params, transforms, profiles, data, **kwargs):
+    data["phi"] = (data["zeta"] + data["omega"]) % (2 * jnp.pi)
```
unalmis, line 2876: also added mod by 2pi in case lost in diff

YigitElma:
Ideally, we don't need repeated code for same quantity if it is calculated exactly the same as some other parametrization (as you did for some of them here). Maybe we can add a test or write a local one to see if some quantity gives different result while it should give the same. What I mean is, this bug can be spotted with the following code,
```python
grid = LinearGrid(rho=1., M=5, N=0)
surface = eq.get_surface_at(rho=1.)
section = eq.get_surface_at(zeta=0)
keys = ["phi_z", "e_zeta"]
data = eq.compute(keys, grid=grid);
data_surf = surface.compute(keys, grid=grid)
data_sect = section.compute(keys, grid=grid)
key = "phi_z"
print(data[key])
print(data_surf[key])
print(data_sect[key])
```
``` bash
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
[ 8.7         8.90637041  9.45996048 10.18500929 10.85131895 11.24734087 11.24734087 10.85131895 10.18500929  9.45996048  8.90637041]
[0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
```
Here, all should have given the same result (maybe some parameterizations have different definitions due to surface shape  i.e ZernikeSurface is axisymmetric etc). It is a bit laborious to find which quantities shouldn't give the same result but I guess they are limited in number.
```python
eq_keys = data_index['desc.equilibrium.equilibrium.Equilibrium'].keys()
surf_keys = data_index['desc.geometry.surface.FourierRZToroidalSurface'].keys()
common_keys = set(eq_keys) & set(surf_keys)
len(common_keys)
```
For example, these 2 parameterizations have 194 common keys. Most of them should give same result when evaluated on the same linear grid.

I think something like,
```python
grid = LinearGrid(rho=1., M=5, N=0)
surface = eq.get_surface_at(rho=1.)
keys = list(common_keys)
data = eq.compute(keys, grid=grid);
data_surf = surface.compute(keys, grid=grid)

for key in keys:
    if not np.allclose(data[key], data_surf[key]):
        print(f"{key} is not equal for surface and equilibrium")  
```
```bash
g_tr is not equal for surface and equilibrium
|e_zeta x e_rho| is not equal for surface and equilibrium
|e_rho x e_theta|_r is not equal for surface and equilibrium
|e_theta x e_zeta|_r is not equal for surface and equilibrium
g_rz is not equal for surface and equilibrium
e_rho is not equal for surface and equilibrium
g_zr is not equal for surface and equilibrium
|e_rho x e_theta| is not equal for surface and equilibrium
g_tt_r is not equal for surface and equilibrium
e_theta_r is not equal for surface and equilibrium
g_rr is not equal for surface and equilibrium
e_rho_z is not equal for surface and equilibrium
n_theta is not equal for surface and equilibrium
|e_theta x e_zeta|_rr is not equal for surface and equilibrium
e_zeta_r is not equal for surface and equilibrium
e_rho_t is not equal for surface and equilibrium
S is not equal for surface and equilibrium
e^theta*sqrt(g) is not equal for surface and equilibrium
g_rt is not equal for surface and equilibrium
g_tz_r is not equal for surface and equilibrium
n_zeta is not equal for surface and equilibrium
```

unalmis:
> Ideally, we don't need repeated code for same quantity if it is calculated exactly the same as some other parametrization (as you did for some of them here).

What do you mean? This PR removes an incorrectly computed quantity and moves computation to rely on the general method.

Regarding the rest of your comment, those bugs are due to #1127 . We can add another consistency check between objects like we do between xyz and rpz in `test_compute_everything` in the PR that solves #1127.

YigitElma:
> What do you mean? This PR removes an incorrectly computed quantity and moves computation to rely on the general method.

Yes, exactly. I suggested to do what you did here to solve more possible bugs.
Oh, I forgot about #1127. But in addition to removing the quantities we shouldn't be able to compute, we can also remove the duplicated code. The above code states that 173 of the 194 common computable quantities give exactly the same result.

f0uriest (approved):

YigitElma (approved):

PR #1177, opened 2024-08-12
Update equations to match physics and boozer fit for general toroidal angle

### LINKED ISSUE #1176: Change equations to match physics
I would prefer to keep stuff like this as independent quantities. For stuff like `g_rt` and `g_tr`, those are the same both physically and numerically, independent of how we parameterize or discretize things.

Stuff like `theta_PEST_tt` and `lambda_tt` are actually different things physically, its just that they happen to the same numerically because we define `theta_PEST = theta + lambda`, but in theory other relations are possible and equally valid physically.

_Originally posted by @f0uriest in https://github.com/PlasmaControl/DESC/pull/1175#discussion_r1713107215_

unalmis:
Resolves #1176.

- [x] Update equations to use poloidal and toroidal angles rather than a particular choice of stream function. No changes were made in places where a particular choice of stream function is necessary to simplify the computation.
- [x] Genearlizes boozer fit for arbitrary toroidal angle


desc/compute/_omnigenity.py
```diff
@@ -263,18 +263,20 @@ def _zeta_B(params, transforms, profiles, data, **kwargs):
...
 )
 def _sqrtg_B(params, transforms, profiles, data, **kwargs):
-    data["sqrt(g)_B"] = (1 + data["lambda_t"]) * (1 + data["nu_z"]) + (
-        data["iota"] - data["lambda_z"]
-    ) * data["nu_t"]
+    data["sqrt(g)_B"] = (
+        data["theta_PEST_t"] * (data["phi_z"] + data["nu_z"])
+        - data["theta_PEST_z"] * (data["phi_t"] + data["nu_t"])
+        + data["iota"] * (data["nu_t"] * data["phi_z"] - data["nu_z"] * data["phi_t"])
+    )
```
unalmis, line 279: generalized for phi != zeta

desc/compute/_core.py
```diff
@@ -2764,6 +2764,25 @@ def _phi_rr(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="phi_rrz",
+    label="\\partial_{\\rho \\rho \\zeta} \\phi",
+    units="rad",
+    units_long="radians",
+    description="Toroidal angle in lab frame, second derivative wrt radial coordinate "
```
dpanici, line 2772: these are 3rd derivs?

desc/compute/_core.py
```diff
@@ -2783,6 +2802,25 @@ def _phi_rt(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="phi_rtz",
+    label="\\partial_{\\rho \\theta \\zeta} \\phi",
+    units="rad",
+    units_long="radians",
+    description="Toroidal angle in lab frame, second derivative wrt radial and "
```
dpanici, line 2810: third deriv

desc/compute/_core.py
```diff
@@ -2802,6 +2840,25 @@ def _phi_rz(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="phi_rzz",
+    label="\\partial_{\\rho \\zeta \\zeta} \\phi",
+    units="rad",
+    units_long="radians",
+    description="Toroidal angle in lab frame, second derivative wrt radial and "
```
dpanici, line 2848: third deriv

desc/compute/_core.py
```diff
@@ -2843,12 +2900,31 @@ def _phi_tt(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="phi_ttz",
+    label="\\partial_{\\theta \\theta \\zeta} \\phi",
+    units="rad",
+    units_long="radians",
+    description="Toroidal angle in lab frame, second derivative wrt poloidal "
```
dpanici, line 2908: third

dpanici: nvm I see what you are doing

dpanici (dismissed):

f0uriest (approved):

desc/compute/_core.py
```diff
@@ -2986,6 +3100,83 @@ def _theta_PEST_r(params, transforms, profiles, data, **kwargs):
...
+    "radial and DESC poloidal coordinate",
+    dim=1,
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["lambda_rt"],
+)
+def _theta_PEST_rt(params, transforms, profiles, data, **kwargs):
+    data["theta_PEST_rt"] = data["lambda_rt"]
```
rahulgaur104, line 3118: Should we think about aliases ? What if someone adds theta_PEST_tr later?

desc/compute/_core.py
```diff
@@ -2986,6 +3100,83 @@ def _theta_PEST_r(params, transforms, profiles, data, **kwargs):
...
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["lambda_rtt"],
+)
+def _theta_PEST_rtt(params, transforms, profiles, data, **kwargs):
+    data["theta_PEST_rtt"] = data["lambda_rtt"]
+    return data
+
```
rahulgaur104, line 3178: Same as above.

rahulgaur104 (approved):

desc/compute/_core.py
unalmis: When you add one of those, all the aliases are added automatically. So if someone adds `theta_PEST_tr` it would just be duplicate code, but I don't think DESC will errror

f0uriest (approved):

PR #1181, opened 2024-08-13
Fix the sign error in wide QR factorization

YigitElma:
When Jacobian is wide, instead of decomposing $\mathbf{J=QR}$, we do $\mathbf{J^T=QR}$. This later requires solving 
$$\mathbf{J}x= -f $$
$$\mathbf{R}^T \mathbf{Q}^Tx = - f $$
$$\mathbf{R}^Tv = - f,  \hspace{1cm}    \text{where } v = \mathbf{Q}^Tx$$
$$x = \mathbf{Q}v$$
where we call `p_newton` to `x`.

### DIFF
--- a/desc/optimize/tr_subproblems.py
+++ b/desc/optimize/tr_subproblems.py
@@ -421,7 +421,7 @@ def trust_region_step_exact_qr(
         p_newton = solve_triangular_regularized(R, -Q.T @ f)
     else:
         Q, R = qr(J.T, mode="economic")
-        p_newton = Q @ solve_triangular_regularized(R.T, f, lower=True)
+        p_newton = Q @ solve_triangular_regularized(R.T, -f, lower=True)
 
     def truefun(*_):
         return p_newton, False, 0.0


unalmis (approved):
Approving, but we should add unit tests which check that these deterministic subproblems are solved correctly. It should catch stuff like this.

Right now, I only see optimization test for strictly convex functions in `test_optimizer`. A basic gradient descent would work for that; a lot could be wrong in more specialized methods that would not be detected by minimizing a convex function.

Edit: maybe useful non-unit tests could be a simple tests like functions with two minima and making sure it converges to the one near initial guess when trust region is small etc.

YigitElma:
> Approving, but we should add unit tests which check that these deterministic subproblems are solved correctly. It should catch stuff like this.
> 
> Right now, I only see optimization test for strictly convex functions in `test_optimizer`. A basic gradient descent would work for that; a lot could be wrong in more specialized methods that would not be detected by minimizing a convex function.
> 
> Edit: maybe useful non-unit tests could be a simple tests like functions with two minima and making sure it converges to the one near initial guess when trust region is small etc.

Yes, it can be good. I think we usually solve problems where output array size is bigger than input array size, so the jacobian is tall (primarily due to collocation nodes number being double the number of modes), that is why we didn't catch it with the current tests. Maybe some scalar objective can catch it.

dpanici (approved):

rahulgaur104 (approved):
Should've marked this PR EZ!

YigitElma:
> Should've marked this PR EZ!

It was really hard! Don't underestimate my effort :rofl:

f0uriest (approved):

PR #1188, opened 2024-08-13
Reduce`QuadratureGrid` number of radial points to match intended functionality

### LINKED ISSUE #983: Quadrature grid has too many nodes for given L, M
```python
for n in range(3,13):

    basis = ZernikePolynomial(n,n)
    grid1 = QuadratureGrid(n,n,0)
    grid2 = ConcentricGrid(n,n,0)

    print(f"{n} {grid1.num_nodes/basis.num_modes:.2f} {grid2.num_nodes/basis.num_modes:.2f}")
```

```
3  2.80 1.00
4  3.00 1.00
5  3.14 1.00
6  3.25 1.00
7  3.33 1.00
8  3.40 1.00
9  3.45 1.00
10 3.50 1.00
11 3.54 1.00
12 3.57 1.00
```

We expect the concentric grid to have as many nodes as the basis has modes (which is True), and the quadrature grid to have double that. However in practice the quadrature grid seems to have 3x-3.5x as many nodes.

dpanici:
- internally to `QuadratureGrid`, reduces the number of radial points fro `L+1` to `(L+2)//2`, which is the minimum amount of points to integrate a polynomial of radial degree `L` with jacobi weight function exactly in general

Resolves #983


tests/test_bootstrap.py
```diff
@@ -1196,7 +1196,7 @@ def test_BootstrapRedlConsistency_normalization(self):
         )
 
         # Results are not perfectly identical because ln(Lambda) is not quite invariant.
-        np.testing.assert_allclose(results, expected, rtol=2e-3)
+        np.testing.assert_allclose(results, expected, rtol=3e-3)
```
rahulgaur104, line 1199: Is this an acceptable loss of accuracy when reducing the number of L points?

tests/test_grid.py
```diff
@@ -482,14 +482,14 @@ def test_quadrature_grid(self):
         N = 0
         NFP = 1
         grid_quad = QuadratureGrid(L, M, N, NFP)
-        roots, weights = special.js_roots(3, 2, 2)
+        roots, weights = special.js_roots(2, 2, 2)
```
rahulgaur104, line 485: Not relevant to this PR but a comment above 485 explaining the purpose of js_roots would be super helpful!

tests/test_grid.py
```diff
@@ -482,14 +482,14 @@ def test_quadrature_grid(self):
...
         quadrature_nodes = np.stack(
             [
-                np.array([roots[0]] * 5 + [roots[1]] * 5 + [roots[2]] * 5),
+                np.array([roots[0]] * 5 + [roots[1]] * 5),
```
rahulgaur104, line 488: I guess I don't understand why roots[2] is replaced with roots[1] .

tests/test_objective_funs.py
```diff
@@ -206,7 +206,7 @@ def test(eq):
             f = obj.compute_unscaled(*obj.xs(eq))
             f_scaled = obj.compute_scaled_error(*obj.xs(eq))
             np.testing.assert_allclose(f, 1.3 / 0.7, rtol=5e-3)
-            np.testing.assert_allclose(f_scaled, 2 * (1.3 / 0.7), rtol=5e-3)
+            np.testing.assert_allclose(f_scaled, 2 * (1.3 / 0.7), rtol=8e-3)
```
rahulgaur104, line 209: Is this loss of accuracy acceptable?

unalmis: yes because the convergence should be exponential. if half nodes stays at same order of accuracy we're good

tests/test_grid.py
unalmis: roots[2] is the third node, which no longer exists in ``roots``

tests/test_grid.py
dpanici: A comment is probably overkill when the doc string of special says they are shifted jacobi roots imo, but I can add

tests/test_bootstrap.py
dpanici: I think yes, we halve the nodes and only lose one part in 1000 relative accuracy in this test

f0uriest (approved):

dpanici:
Merge #1175 first

dpanici:
For the `plot_b_mag` test, I find a difference like this when using this PR (where the default quad grid has lower resolution by a factor of 2), which seems to be from a difference in the `iota` used to make the `zeta` nodes for the desired field line coordinates passed to `map_coordinates`. The difference in the plot then came bc the interpolation in this PR is from further apart nodes in `rho` then on master.

This `iota` in that test is for some reason interpolated from `eq.compute("iota")` to the desired `rho` value instead of being just computed at the desired rho value, so I made that change and now this issue won't arise again.

image

unalmis:
@dpanici The reason the map_coordinates test is failing is because of the weakness of 3d root finding discussed here https://github.com/PlasmaControl/DESC/pull/1153#issue-2432807739. Reducing the number of quadrature points decreases the number of radial knots used to construct a cubic spline of `iota`, which deteriorates the 3d root finding used by default in `map_coordinates`.

Given that `iota` is the main ingredient in all these coordinate mappings, the solution is to update `map_coordinates` to also compute `iota_r` and feed that into `interpax.interp1d` with the undocumented kwarg `fx=iota_r` to compute Hermite spline. This is what I have been doing for the neoclassical stuff like effective ripple and Gamma. All the plumbing for this is done in #1003 . I should probably pull that out into a new PR that we merge before this.

unalmis:
You'll still need to slightly increase tolerance on that failing test after #1199. Probably should mark `Todo: Reduce tolerance after GitHub issue #1207 is resolved.`

tests/test_grid.py
dpanici: @dpanici add a comment here or replace 5 with grid.num_theta

unalmis (approved):

tests/test_plotting.py
```diff
@@ -826,14 +825,12 @@ def test_plot_b_mag():
...
     alpha = 0
-    # compute and fit iota profile
+    # compute iota
     eq = get("W7-X")
-    data = eq.compute("iota")
-    fi = interp1d(data["rho"], data["iota"])
+    iota = eq.compute("iota", grid=LinearGrid(rho=np.sqrt(psi), NFP=eq.NFP))["iota"][0]
 
     # get flux tube coordinate system
     rho = np.sqrt(psi)
```
ddudt, line 833: Could move this up a few lines so you don't have to recompute when computing iota

ddudt (dismissed):

desc/grid.py
```diff
@@ -1243,7 +1243,9 @@ def _create_nodes(self, L=1, M=1, N=1, NFP=1):
         self._N = check_nonnegint(N, "N", False)
         self._NFP = check_posint(NFP, "NFP", False)
         self._period = (np.inf, 2 * np.pi, 2 * np.pi / self._NFP)
-        L = L + 1
+        # divide (L+1) by 2 bc only need (L+1)/2  points to
+        # integrate L-th order jacobi polynomial exactly
+        L = round((L + 1) / 2)
```
unalmis, line 1248: To integrate an algebraic polynomial of degree of L exactly requires `L // 2` points (if you have complete freedom to choose quadrature points, which we do.

A lobatto grid would require `(L+2)//2`, which is what you have written if the computer performs exact arithmetic and round ups every time).

unalmis (changes requested):
i have a small request to reduce number of nodes by one or change formula to not use round

desc/grid.py
unalmis: So far the linked issue and tests show that the accuracy loss is acceptable, but are we sure that's not a characteristic unique to those tests? For example, if I multiply two polynomials together the degree doubles and you in theory need double the quadrature points to integrate it now. So when we multiply two quantities whose radial parts are given by a Jacobi polynomial of degree L in the compute functions, we would need a quadrature grid with `L` points to integrate it

dpanici: change to just `L//2`

unalmis:
Any chance we could also resolve this here:
https://github.com/PlasmaControl/DESC/blob/425fb024fee993a7f190b62669b6bb5592db0e5c/desc/equilibrium/coords.py#L96
so that we always default to the more accurate method for this common coordinate mapping?

desc/grid.py
dpanici: `(L+2)//2` is what I think we want based on this, or equivalently `L//2 + 1` I think.

we want to exactly integrate a `L` degree polynomial, want to know how many points `n` we need

Gaussian quadrature tells us `n` points lets us exactly integrate a `2n-1` order poly.

$$
\begin{align}
&L= 2n-1\\
(L+1)/2 &= n
\end{align}
$$

If `L` is odd, then this works fine and we get integer `n`.

If `L` is even, we actually get `n` being a fraction. If we floor divide, we will ignore the extra part and actually not be able to correctly integrate the `L` order poly, as we will be one short. (example, `L=2` means `(L+1)//2=1=n` which can only integrate a `2*1-1 = 1` degre polynomial.

unalmis: Yes right (L+2)//2 is interior nodes minimum. You can do L//2 if you do lobatto nodes but already know the endpoints. Got them mixed up.

dpanici: I realize this is the same as `np.ceil((L+1)/2)` but whatever I will leave as is, this will work fine

unalmis (dismissed):

dpanici:
> Any chance we could also resolve this here:
> 
> https://github.com/PlasmaControl/DESC/blob/425fb024fee993a7f190b62669b6bb5592db0e5c/desc/equilibrium/coords.py#L96
> 
> 
> so that we always default to the more accurate method for this common coordinate mapping?

just made the change, can you check and if I did something dumb in how I re-ordered, feel free to fix

unalmis (approved):

unalmis (dismissed):

ddudt (dismissed):

unalmis (approved):

unalmis:
My bad for making multiple commits that should have been one. I'm done committing now.

unalmis (dismissed):

desc/equilibrium/coords.py
```diff
@@ -118,29 +133,20 @@ def map_coordinates(  # noqa: C901
...
-    basis_derivs = tuple(f"{X}_{d}" for X in inbasis for d in ("r", "t", "z"))
-    for key in basis_derivs:
-        errorif(
-            key not in data_index["desc.equilibrium.equilibrium.Equilibrium"],
-            NotImplementedError,
-            f"don't have recipe to compute partial derivative {key}",
-        )
+    # do surface average to get iota once
+    if "iota" in profiles and profiles["iota"] is None:
+        profiles["iota"] = eq.get_profile(["iota", "iota_r"], params=params)
```
rahulgaur104, line 138: Is iota_r being used anywhere or is it just there so you can use Hermite?

rahulgaur104 (dismissed):

desc/equilibrium/coords.py
unalmis: `iota` and `iota_r` get computed on `QuadratureGrid` then we use those values to construct Hermite spline for $\iota(\rho)$. $\iota(\rho)$ is then evaluated at whatever sequence of $\rho$ values the Newton iteration proceeds with. For $\rho, \alpha, \zeta$ coordinate mappings, the $\rho$ values don't need to change throughout Newton iteration since we can just do 1D root finding, which this PR now ensures is done.

unalmis (dismissed):

f0uriest (approved):

unalmis (approved):

f0uriest (approved):

ddudt (approved):

PR #1190, opened 2024-08-14
Reduce test stringency for NAE

dpanici:
It is hard to put a good tolerance on how closely we expect the near-axis surfaces to match the original NAE when we use these constraints, so I changed to just evaluating |B|, lambda, and then iota on the axis (if iota matches closely, since it relies on derivatives of geometry on-axis, the surfaces and other quantiites should be good near-axis compared to the NAE)


f0uriest (approved):

YigitElma (approved):
Can't wait to test this in Poincare branch :smile:

PR #1192, opened 2024-08-14
Simplify normalization scales

YigitElma:
Simplify normalizations in `compute_scaling_factors` to speed up the building of objectives.

### DIFF
--- a/desc/objectives/normalization.py
+++ b/desc/objectives/normalization.py
@@ -41,7 +41,7 @@ def compute_scaling_factors(thing):
         scales["Psi"] = abs(thing.Psi)
         scales["A"] = np.pi * scales["a"] ** 2
         scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
-        scales["B"] = scales["Psi"] / scales["A"]
+        scales["B"] = scales["Psi"] / scales["A"] * 1.25
         B_pressure = scales["B"] ** 2 / (2 * mu_0)
         scales["I"] = scales["B"] * 2 * np.pi / mu_0
         scales["W"] = B_pressure * scales["V"]
--- a/desc/objectives/normalization.py
+++ b/desc/objectives/normalization.py
@@ -42,7 +42,7 @@ def compute_scaling_factors(thing):
         scales["n"] = 1e19
         scales["A"] = np.pi * scales["a"] ** 2
         scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
-        scales["B"] = scales["Psi"] / scales["A"]
+        scales["B"] = scales["Psi"] / scales["A"] * np.sqrt(2)
         scales["I"] = scales["B"] * 2 * np.pi / mu_0
         scales["p"] = scales["B"] ** 2 / (2 * mu_0)
         scales["W"] = scales["p"] * scales["V"]
--- a/desc/objectives/normalization.py
+++ b/desc/objectives/normalization.py
@@ -39,17 +39,31 @@ def compute_scaling_factors(thing):
         scales["R0"] = R00
         scales["a"] = np.sqrt(np.abs(R10 * Z10))
         scales["Psi"] = abs(thing.Psi)
-        scales["n"] = 1e19
         scales["A"] = np.pi * scales["a"] ** 2
         scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
-        scales["B"] = scales["Psi"] / scales["A"] * np.sqrt(2)
+        scales["B"] = scales["Psi"] / scales["A"]
+        B_pressure = scales["B"] ** 2 / (2 * mu_0)
         scales["I"] = scales["B"] * 2 * np.pi / mu_0
-        scales["p"] = scales["B"] ** 2 / (2 * mu_0)
-        scales["W"] = scales["p"] * scales["V"]
+        scales["W"] = B_pressure * scales["V"]
         scales["J"] = scales["B"] / scales["a"] / mu_0
-        scales["F"] = scales["p"] / scales["a"]
+        scales["F"] = B_pressure / scales["a"]
         scales["f"] = scales["F"] * scales["V"]
-        scales["T"] = scales["p"] / (scales["n"] * elementary_charge)
+
+        if thing.pressure is not None:
+            p0 = float(thing.pressure(0)[0])
+        else:
+            scales["n"] = float(
+                ((thing.atomic_number(0) + 1) / 2 * thing.electron_density(0))[0]
+            )
+            scales["T"] = np.mean(
+                [thing.electron_temperature(0), thing.ion_temperature(0)]
+            )
+            p0 = elementary_charge * 2 * scales["n"] * scales["T"]
+        if p0 < 1:  # vacuum
+            scales["p"] = B_pressure
+        else:
+            scales["p"] = p0
+        scales["W_p"] = scales["p"] * scales["V"] / 2
 
     elif isinstance(thing, FourierRZToroidalSurface):
         R00 = thing.R_lmn[thing.R_basis.get_idx(M=0, N=0)]
--- a/desc/objectives/normalization.py
+++ b/desc/objectives/normalization.py
@@ -41,14 +41,7 @@ def compute_scaling_factors(thing):
         scales["A"] = np.pi * scales["a"] ** 2
         scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
         scales["B_T"] = abs(thing.Psi) / scales["A"]
-        if thing.iota is not None:
-            iota_avg = np.mean(np.abs(thing.iota(np.linspace(0, 1, 20))))
-        else:
-            iota_avg = np.mean(np.abs(thing.get_profile("iota")(np.linspace(0, 1, 20))))
-        if np.isclose(iota_avg, 0):
-            scales["B_P"] = scales["B_T"]
-        else:
-            scales["B_P"] = scales["B_T"] * iota_avg
+        scales["B_P"] = scales["B_T"]
         scales["B"] = np.sqrt(scales["B_T"] ** 2 + scales["B_P"] ** 2)
         scales["I"] = scales["B_P"] * 2 * np.pi / mu_0
         scales["p"] = scales["B"] ** 2 / (2 * mu_0)
--- a/desc/objectives/normalization.py
+++ b/desc/objectives/normalization.py
@@ -41,7 +41,10 @@ def compute_scaling_factors(thing):
         scales["A"] = np.pi * scales["a"] ** 2
         scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
         scales["B_T"] = abs(thing.Psi) / scales["A"]
-        iota_avg = np.mean(np.abs(thing.get_profile("iota")(np.linspace(0, 1, 20))))
+        if thing.iota is not None:
+            iota_avg = np.mean(np.abs(thing.iota(np.linspace(0, 1, 20))))
+        else:
+            iota_avg = np.mean(np.abs(thing.get_profile("iota")(np.linspace(0, 1, 20))))
         if np.isclose(iota_avg, 0):
             scales["B_P"] = scales["B_T"]
         else:
--- a/desc/objectives/normalization.py
+++ b/desc/objectives/normalization.py
@@ -38,19 +38,17 @@ def compute_scaling_factors(thing):
 
         scales["R0"] = R00
         scales["a"] = np.sqrt(np.abs(R10 * Z10))
+        scales["Psi"] = abs(thing.Psi)
+        scales["n"] = 1e19
         scales["A"] = np.pi * scales["a"] ** 2
         scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
-        scales["B_T"] = abs(thing.Psi) / scales["A"]
-        scales["B_P"] = scales["B_T"]
-        scales["B"] = np.sqrt(scales["B_T"] ** 2 + scales["B_P"] ** 2)
-        scales["I"] = scales["B_P"] * 2 * np.pi / mu_0
+        scales["B"] = scales["Psi"] / scales["A"]
+        scales["I"] = scales["B"] * 2 * np.pi / mu_0
         scales["p"] = scales["B"] ** 2 / (2 * mu_0)
         scales["W"] = scales["p"] * scales["V"]
         scales["J"] = scales["B"] / scales["a"] / mu_0
         scales["F"] = scales["p"] / scales["a"]
         scales["f"] = scales["F"] * scales["V"]
-        scales["Psi"] = abs(thing.Psi)
-        scales["n"] = 1e19
         scales["T"] = scales["p"] / (scales["n"] * elementary_charge)
 
     elif isinstance(thing, FourierRZToroidalSurface):


YigitElma:
```python
eq = get("ATF")

from scipy.constants import elementary_charge, mu_0
def get_lowest_mode(basis, coeffs):
    """Return lowest order coefficient (excluding m=0 modes)."""
    # lowest order modes: [0, +1, -1, +2, -2, ...]
    m_modes = np.arange(1, eq.M + 1)
    m_modes = np.vstack((m_modes, -m_modes)).flatten(order="F")
    n_modes = np.arange(eq.N + 1)
    n_modes = np.vstack((n_modes, -n_modes)).flatten(order="F")
    for n in n_modes:
        for m in m_modes:
            try:
                x = coeffs[basis.get_idx(M=m, N=n)]
                if not np.isclose(x, 0):  # mode exists and coefficient is non-zero
                    return x
            except ValueError:
                pass
    raise ValueError("No modes found, geometry is unphysical.")

def fun():
    scales = {}
    R00 = eq.Rb_lmn[eq.surface.R_basis.get_idx(M=0, N=0)]
    R10 = get_lowest_mode(eq.surface.R_basis, eq.Rb_lmn)
    Z10 = get_lowest_mode(eq.surface.Z_basis, eq.Zb_lmn)

    scales["R0"] = R00
    scales["a"] = np.sqrt(np.abs(R10 * Z10))
    scales["A"] = np.pi * scales["a"] ** 2
    scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
    scales["B_T"] = abs(eq.Psi) / scales["A"]
    if eq.iota is not None:
        iota_avg = np.mean(np.abs(eq.iota(np.linspace(0, 1, 20))))
    else:
        iota_avg = np.mean(np.abs(eq.get_profile("iota")(np.linspace(0, 1, 20))))
    if np.isclose(iota_avg, 0):
        scales["B_P"] = scales["B_T"]
    else:
        scales["B_P"] = scales["B_T"] * iota_avg
    scales["B"] = np.sqrt(scales["B_T"] ** 2 + scales["B_P"] ** 2)
    scales["I"] = scales["B_P"] * 2 * np.pi / mu_0
    scales["p"] = scales["B"] ** 2 / (2 * mu_0)
    scales["W"] = scales["p"] * scales["V"]
    scales["J"] = scales["B"] / scales["a"] / mu_0
    scales["F"] = scales["p"] / scales["a"]
    scales["f"] = scales["F"] * scales["V"]
    scales["Psi"] = abs(eq.Psi)
    scales["n"] = 1e19
    scales["T"] = scales["p"] / (scales["n"] * elementary_charge)

def fun2():
    scales = {}
    R00 = eq.Rb_lmn[eq.surface.R_basis.get_idx(M=0, N=0)]
    R10 = get_lowest_mode(eq.surface.R_basis, eq.Rb_lmn)
    Z10 = get_lowest_mode(eq.surface.Z_basis, eq.Zb_lmn)

    scales["R0"] = R00
    scales["a"] = np.sqrt(np.abs(R10 * Z10))
    scales["A"] = np.pi * scales["a"] ** 2
    scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
    scales["B_T"] = abs(eq.Psi) / scales["A"]
    iota_avg = np.mean(np.abs(eq.get_profile("iota")(np.linspace(0, 1, 20))))
    if np.isclose(iota_avg, 0):
        scales["B_P"] = scales["B_T"]
    else:
        scales["B_P"] = scales["B_T"] * iota_avg
    scales["B"] = np.sqrt(scales["B_T"] ** 2 + scales["B_P"] ** 2)
    scales["I"] = scales["B_P"] * 2 * np.pi / mu_0
    scales["p"] = scales["B"] ** 2 / (2 * mu_0)
    scales["W"] = scales["p"] * scales["V"]
    scales["J"] = scales["B"] / scales["a"] / mu_0
    scales["F"] = scales["p"] / scales["a"]
    scales["f"] = scales["F"] * scales["V"]
    scales["Psi"] = abs(eq.Psi)
    scales["n"] = 1e19
    scales["T"] = scales["p"] / (scales["n"] * elementary_charge)

def fun3():
    con = FixBoundaryR(eq)
    con.build()

%timeit -n 10 fun()   # this PR scaling factor
%timeit -n 10 fun2()  # master scaling factor (without iota check)
%timeit -n 10 fun3()  # this PR constraint build
```
```
2.27 ms ± 46 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
176 ms ± 3.56 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
13.7 ms ± 866 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
And constraint build (fun3()) on master takes,
```
190 ms ± 7.81 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

dpanici:
Under the hood, `get_profiles` computes `iota`, so if you run your benchmark code you posted on an equilibrum that has a current profile assigned instead of an iota profile (ATF has an iota profile), master and this PR will be the same amt of time.

So I guess this PR would just speed it up for equilibria with iota profiles assigned

dpanici:
just use |B| scale for both Bp and Bt stuff, or make scale |B| some multiple of Bt scale

desc/objectives/normalization.py
```diff
@@ -41,11 +41,7 @@ def get_lowest_mode(basis, coeffs):
         scales["A"] = np.pi * scales["a"] ** 2
         scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
         scales["B_T"] = abs(thing.Psi) / scales["A"]
-        iota_avg = np.mean(np.abs(thing.get_profile("iota")(np.linspace(0, 1, 20))))
-        if np.isclose(iota_avg, 0):
-            scales["B_P"] = scales["B_T"]
-        else:
-            scales["B_P"] = scales["B_T"] * iota_avg
+        scales["B_P"] = scales["B_T"]
```
dpanici, line 44: I think we should make it even simpler, idt anyone will ever use B_T and B_P as the scales for a problem, so we can just do `scales["B"] = abs(thing.Psi) / scales["A"]` and get rid of the B_T and B_P keys entirely

dpanici (changes requested):
simplify further

f0uriest (approved):

dpanici:
The test fails on CI, but seems to pass locally for me, how about you @YigitElma

YigitElma:
The same one fails locally for me

YigitElma:
The same test on master:
```
Number of parameters: 10
Number of objectives: 1
Starting optimization
Using method: scipy-trust-krylov
       0              1          1.334e-01                                    3.795e-03   
       1              2          1.334e-01      1.224e-05      7.180e-03      6.181e-05   
       2              3          1.334e-01      1.224e-07      1.038e-02      2.029e-06   
       3              4          1.334e-01      1.497e-11      1.680e-05      3.672e-09   
Warning: `callback` raised `StopIteration`.
         Current function value: 1.334e-01
         Total delta_x: 3.034e+00
         Iterations: 3
         Function evaluations: 4
         Gradient evaluations: 4
         Hessian evaluations: 4
```
On this PR:
```
Number of parameters: 10
Number of objectives: 1
Starting optimization
Using method: scipy-trust-krylov
       0              1          5.336e-01                                    1.518e-02   
       1              2          5.336e-01      4.897e-05      7.180e-03      2.472e-04   
       2              3          5.336e-01      4.895e-07      1.038e-02      8.117e-06   
       3              4          5.336e-01      5.990e-11      1.680e-05      1.469e-08   

../../miniconda3/envs/desc-env/lib/python3.12/site-packages/scipy/optimize/_trustregion.py:225: in _minimize_trust_region
    p, hits_boundary = m.solve(trust_radius)


FAILED tests/test_optimizer.py::TestAllOptimizers::test_all_optimizers_scalar[scipy-trust-krylov] - RuntimeWarning: invalid value encountered in multiply
```

YigitElma:
@f0uriest Idk why but free boundary notebook throws an error for plot_2d. I ran it locally (making sure that my matplotlib version is the same with CI) and don't see a problem. Can you check?

ddudt (changes requested):
I think some of our scales are off. Testing with approximate W7-X values of `Psi = 2` and `a = 0.5`, this results in: 

- B = 5.6 T
- p = 12.7 MPa
- T = 7.9 MeV

These values are way too high! I suggest the following changes: 

- Remove the `np.sqrt(2)` factor from B
- Hard-code `scales["T"] = 1e3` like we do for density
- Change the pressure scale to `scales["p"] = 2 * elementary_charge * scales["n"] * scales["T"]` (which will be 3.2 kPa)

It would also be nice if we had a way to change these hard-coded temperatures and densities. These values are fine for experiment scales but low for reactor scales. Maybe something like this could work: 
```
scales["n"] = np.max(eq.ne_l)
```
(and similar for temperature and pressure). I think that would give a good estimate independent of the Profile type

f0uriest:
> I think some of our scales are off. Testing with approximate W7-X values of `Psi = 2` and `a = 0.5`, this results in:
> 
>     * B = 5.6 T
> 
>     * p = 12.7 MPa
> 
>     * T = 7.9 MeV
> 
> 
> These values are way too high! I suggest the following changes:
> 
>     * Remove the `np.sqrt(2)` factor from B
> 
>     * Hard-code `scales["T"] = 1e3` like we do for density
> 
>     * Change the pressure scale to `scales["p"] = 2 * elementary_charge * scales["n"] * scales["T"]` (which will be 3.2 kPa)
> 
> 
> It would also be nice if we had a way to change these hard-coded temperatures and densities. These values are fine for experiment scales but low for reactor scales. Maybe something like this could work:
> 
> ```
> scales["n"] = np.max(eq.ne_l)
> ```
> 
> (and similar for temperature). I think that would give a decent estimate independent of the Profile type

They seem ok to me. Running on the actual w7x equilibrium I get the scale for B is 3.11 T which is pretty close to the actual volume average B=2.78 T

I'd prefer to keep pressure scaling based on B, since for vacuum configs thats the only thing that makes sense. 

I do like the idea of making temperature and density scales relative to the profiles if they exist (though i'd actually evaluate the profile rather than relying on the magnitude of the coefficients which may not mean the same thing for different types)

ddudt:
> They seem ok to me. Running on the actual w7x equilibrium I get the scale for B is 3.11 T which is pretty close to the actual volume average B=2.78 T

I still don't think we need the `sqrt(2)` factor, since I think it is trying to account for an equal contribution from the poloidal field but the majority of the field is toroidal. So this correction factor is generally too high, although we only care about order of magnitude so I'm splitting hairs here. 

> I'd prefer to keep pressure scaling based on B, since for vacuum configs thats the only thing that makes sense.

We need to at least add a correction factor to the existing pressure scale on the order of `1e-2`, since the thermal energy is typically only a few percent of the magnetic energy. 

> I do like the idea of making temperature and density scales relative to the profiles if they exist (though i'd actually evaluate the profile rather than relying on the magnitude of the coefficients which may not mean the same thing for different types)

I agree that would be better, but the whole point of this PR is to avoid calling any compute functions to reduce compile times. So I was trying to think of a heuristic based only on the coefficients.

f0uriest:
> We need to at least add a correction factor to the existing pressure scale on the order of 1e-2, since the thermal energy is typically only a few percent of the magnetic energy.

That's fine, as long as we leave the scale for force as it is now, since the force error should be proportional to B^2, independent of p

> I agree that would be better, but the whole point of this PR is to avoid calling any compute functions to reduce compile times. So I was trying to think of a heuristic based only on the coefficients.

Calling `profile.compute` is super cheap so that's probably fine (that was the initial idea of this PR anyways, and we'd only do it if the profile isn't None so in most cases its a no-op)

desc/objectives/normalization.py
```diff
@@ -38,23 +38,17 @@ def get_lowest_mode(basis, coeffs):
 
         scales["R0"] = R00
         scales["a"] = np.sqrt(np.abs(R10 * Z10))
+        scales["Psi"] = abs(thing.Psi)
+        scales["n"] = 1e19
```
ddudt, line 42: Instead of hard-coding this, compute `ne` on axis (and only do this if the equilibrium has a density profile assigned). 

Same for temperature profile.

desc/objectives/normalization.py
```diff
@@ -38,23 +38,17 @@ def get_lowest_mode(basis, coeffs):
...
-        iota_avg = np.mean(np.abs(thing.get_profile("iota")(np.linspace(0, 1, 20))))
-        if np.isclose(iota_avg, 0):
-            scales["B_P"] = scales["B_T"]
-        else:
-            scales["B_P"] = scales["B_T"] * iota_avg
-        scales["B"] = np.sqrt(scales["B_T"] ** 2 + scales["B_P"] ** 2)
-        scales["I"] = scales["B_P"] * 2 * np.pi / mu_0
+        scales["B"] = scales["Psi"] / scales["A"] * np.sqrt(2)
+        scales["I"] = scales["B"] * 2 * np.pi / mu_0
         scales["p"] = scales["B"] ** 2 / (2 * mu_0)
```
ddudt, line 47: Multiply this by 1e-2 since thermal energy is typically a small fraction of the magnetic energy.

desc/objectives/normalization.py
```diff
@@ -38,23 +38,17 @@ def get_lowest_mode(basis, coeffs):
...
         scales["W"] = scales["p"] * scales["V"]
         scales["J"] = scales["B"] / scales["a"] / mu_0
         scales["F"] = scales["p"] / scales["a"]
```
ddudt, line 50: Multiply this by 1e+2 to counter-act the requested change to pressure above.

desc/objectives/normalization.py
```diff
@@ -38,23 +38,17 @@ def get_lowest_mode(basis, coeffs):
...
         scales["f"] = scales["F"] * scales["V"]
-        scales["Psi"] = abs(thing.Psi)
-        scales["n"] = 1e19
         scales["T"] = scales["p"] / (scales["n"] * elementary_charge)
```
ddudt, line 52: See comment above about computing `Te` on axis.

ddudt (dismissed):

desc/objectives/normalization.py
```diff
@@ -38,23 +38,17 @@ def get_lowest_mode(basis, coeffs):
...
-        if np.isclose(iota_avg, 0):
-            scales["B_P"] = scales["B_T"]
-        else:
-            scales["B_P"] = scales["B_T"] * iota_avg
-        scales["B"] = np.sqrt(scales["B_T"] ** 2 + scales["B_P"] ** 2)
-        scales["I"] = scales["B_P"] * 2 * np.pi / mu_0
+        scales["B"] = scales["Psi"] / scales["A"] * np.sqrt(2)
+        scales["I"] = scales["B"] * 2 * np.pi / mu_0
         scales["p"] = scales["B"] ** 2 / (2 * mu_0)
         scales["W"] = scales["p"] * scales["V"]
```
ddudt, line 48: If `p` is the peak pressure on-axis, this will be a huge over-estimate of the total plasma thermal energy. We will need to divide by some correction factor.

f0uriest: It's mean to be the total energy (thermal + magnetic)

f0uriest: (using p ~ B^2)

desc/objectives/normalization.py
f0uriest: or maybe max of Te and Ti

desc/objectives/normalization.py
ddudt: Well p is almost never proportional to B^2, it is usually 2 orders of magnitude less...

It would be nice to have a normalization for thermal energy, because that's what I'm using in #1168

f0uriest: Feel free to add it, this one here is meant to be the scaling factor for the `Energy` objective so its proportional to `B^2`, not the thermal pressure

f0uriest (approved):

f0uriest:
Update title and description before merging

YigitElma:
Thanks for taking this over @ddudt

f0uriest (approved):

YigitElma:
I cannot approve but I would 😅

PR #1194, opened 2024-08-15
Increase grid resolution of signed distance test to make more robust

dpanici:


dpanici:
@ddudt try this and see if locally it passes for you now

ddudt (approved):
Works for me locally now!

f0uriest (approved):

PR #1197, opened 2024-08-16
Vectorize boozer transform over multiple surfaces

### LINKED ISSUE #754: Improving DESC Surface Boozer Transform
While we wait on progress on #369 , there are a few improvements that can be made when using the surface transform in DESC
- [ ] Adding utility function to perform the transform in the most efficient way possible (reusing matrices etc), similar to what was done in #680 
- [x] Resolving #720 to make zernike evaluations faster - this issue was actually reverted though...
- [ ] changing the utility functions used in calculating the transform at multiple surfaces (like in `make_boozmn_output`) to use jax.numpy, which can allow for more parts of the calculation to be performed on the GPU and potentially allow jitting more parts of the calculation (like the `zernike_to_fourier` function, for example) #1197

f0uriest:
- [x] Allows the calculation of `B_mn` on tensor product grids with multiple flux surfaces.
- [x] Allows `QuasisymmetryBoozer` objective to target multiple surfaces at once
~- [ ] Allows `Omnigenity` objective to target multiple surfaces at once~ moved to #1225 
- [x] Simplify boozer/QS plotting functions

Resolves #754


desc/compute/_omnigenity.py
```diff
@@ -26,16 +26,24 @@
     + "magnetic field",
     dim=1,
     params=[],
-    transforms={"B": [[0, 0, 0]]},
+    transforms={"B": [[0, 0, 0]], "grid": []},
     profiles=[],
     coordinates="rtz",
     data=["B_theta"],
+    resolution_requirement="tz",
+    source_grid_requirement={"is_meshgrid": True},
```
unalmis, line 34: When you add source grid requirement, it's going to throw an error unless the grid has ``grid.source_grid`` attribute. What you want is just ``grid_requirement``. This can be done via
1. Adding a new kwarg ``grid_requirement`` and doing what you have here. In that case, add the following code to `check_fun` [here](https://github.com/PlasmaControl/DESC/blob/d528b3313b1716bcf1441e9161f9605f90833564/desc/compute/utils.py#L92.).
```python
reqs = data_index[p][name]["grid_requirement"]
for req in reqs:
    errorif(
        not hasattr(transforms["grid"], req)
        or reqs[req] != getattr(transforms["grid"], req),
        AttributeError,
        f"Expected grid with '{req}:{reqs[req]}' to compute {name}.",
    )
```

2. Replacing ``source_grid_requirement`` with ``grid_requirement``. In that case we'll just push ``source_grid_requirement`` into the ``grid_requirement``  dictionary as a key. The above code will work equally well since having a source grid is just an attribute.
```

Comment (f0uriest) [reply desc/compute/_omnigenity.py:34]: Thanks, option 1 sounds the cleanest.

I'm running into another problem, with the definition of `is_meshgrid`:

```python
    @property
    def is_meshgrid(self):
        """bool: Whether this grid is a tensor-product grid.

        Let the tuple (r, p, t) ∈ R³ denote a radial, poloidal, and toroidal
        coordinate value. The is_meshgrid flag denotes whether any coordinate
        can be iterated over along the relevant axis of the reshaped grid:
        nodes.reshape(num_radial, num_poloidal, num_toroidal, 3).
        """
        return self.__dict__.setdefault("_is_meshgrid", False)
```

This would exclude most of our regular grids (`LineaGrid`, `QuadratureGrid` etc) because they sort the nodes differently, as `nodes.reshape((num_poloidal, num_radial, num_toroidal, 3), order="F")` so that it works with `Transform.transform`
Can we change it so that `is_meshgrid` means the latter? Will that overly mess with any of your stuff?

unalmis: If the test below passes, then my stuff could be changed to work with that convention. However, I'm slightly against it, and would prefer changing that boolean flag to a string to specify which ordering you want.

My issues with that ordering are that it
1. Seems error prone if we need to remember to specify Fortran ordering `order=F` everywhere we call `reshape`.
2. Slightly awkward in use for algorithms that do coordinate mappings. The current `is_meshgrid` convention is nice because even when you lose the tensor-product characteristic of the nodes when you do a coordinate mapping, you can still group the data by flux surfaces like `f.reshape(num_radial, -1)`. With your suggested change I think you'd need to do `_reshape_mesh_data(f).reshape(num_radial, -1)`.

```python
def reshape_mesh_data(f):
  """Reshape data between meshgrid conventions.

  Given ``f`` on meshgrid like
  nodes.reshape(num_poloidal, num_radial, num_toroidal, 3, order="F").
  rearange ``f`` to form so that any coordinate
  can be iterated over along the relevant axis of a meshgrid like
  nodes.reshape(num_radial, num_poloidal, num_toroidal, 3). 
  """
  shape = [num_poloidal, num_radial, num_toroidal]
  if f.ndim > 1:
    shape.append(-1)
  f = np.swapaxes(f.reshape(shape, order="F"), 0, 1)
  return f

rho = np.linspace(0, 1, 10)
theta = np.linspace(0, np.pi, 13)
zeta = np.linspace(0, 2*np.pi, 7)
grid1 = Grid.create_meshgrid([rho, theta, zeta]) 
grid2 = LinearGrid(rho=rho, theta=theta, zeta=zeta)
np.testing.assert_allclose(grid2.nodes, grid1.nodes.reshape(rho.size, theta.size, zeta.size, 3))
```

unalmis: Also, for my work in fourier bounce thing... right now I use the optimal reshaping/broadcasting order. Previously, I had an additional call to `moveaxis` that caused jax to warn that the code was hard to compile. I'm not sure if ``reshape_mesh_data`` will cause the same warning.

unalmis: Actually I'm less opposed to this now because this is just a preprocessing step on the input data, so I don't need to change all calls to reshape.

f0uriest: OK, if you can make your stuff work with our current regular ordering that would be ideal. I think we should try to be consistent in how we flatten tensor product arrays, so the other option would be modifying all the transform and plotting stuff but that'd be a lot more work I think.

f0uriest:
@unalmis can you look at adding the boozer stuff to `test_axis_limits.py`? It looks like we'll still need some special logic when integrating over theta/zeta for the `_mn` qtys.

unalmis:
when you merge master please fix units label here:
https://github.com/PlasmaControl/DESC/blob/04e6e564066aab0f128bdc782fa8bcbd3d526fd5/desc/compute/_omnigenity.py#L46

ddudt (approved):

dpanici (approved):
what changed in the master compute data, out of curiosity?

f0uriest:
> what changed in the master compute data, out of curiosity?

The size of the boozer data. I think I mentioned in the meeting last week, master is currently wrong if you try to compute the boozer stuff on a grid with more than 1 surface (as is done in `test_compute_everything`. It would fit the whole volume to a single double fourier series, effectively doing some sort of radial average of the boozer qtys. With this PR it now correctly returns different data for each surface.

Issue #1195, opened 2024-08-15
RecursionError when attempting to plot coilset within DESC


jmathewson1869:
After loading a makegrid coilset from a VMEC input I am able to solve for an equilibrium(that is incorrect), however when I attempt to plot the coilset, it gives me the error shown. Any help would be much appreciated. 

[DESC_GABRIEL_ROT_ELLIPSE.md](https://github.com/user-attachments/files/16630018/DESC_GABRIEL_ROT_ELLIPSE.md)

f0uriest:
```python
coil_sym = r"coils.rotatingellipse"

field = MixedCoilSet.from_makegrid_coilfile(coil_sym,method='linear')

plot_coils(coil_sym)
```

it looks like you're trying to plot the string `coil_sym` not the actual loaded coilset `field` (though we could probably make a better error message)

jmathewson1869:
Sorry to just be seeing this. Have been busy with my other job. But yes
this works! Thank you for the help.

PR #1198, opened 2024-08-16
Improve error msg for plot coils when non-coil object is passed in

dpanici:
Inspired by #1195

### DIFF
--- a/desc/plotting.py
+++ b/desc/plotting.py
@@ -2384,6 +2384,8 @@ def plot_coils(coils, grid=None, fig=None, return_data=False, **kwargs):
         dictionary of the data plotted, only returned if ``return_data=True``
 
     """
+    from desc.coils import _Coil
+
     lw = kwargs.pop("lw", 5)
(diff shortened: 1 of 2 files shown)


f0uriest (approved):

desc/plotting.py
```diff
@@ -2394,6 +2396,12 @@ def plot_coils(coils, grid=None, fig=None, return_data=False, **kwargs):
         ValueError,
         f"plot_coils got unexpected keyword argument: {kwargs.keys()}",
     )
+    errorif(
+        not isinstance(coils, _Coil),
+        ValueError,
+        "Expected `coils` to be object of type `_Coil`, instead got type"
```
VascoSch92, line 2402:
```suggestion
        "Expected `coils` to be of type `_Coil`, instead got type"
```
just a suggestion as everything in python is an object :-)

YigitElma (approved):

ddudt (approved):

PR #1199, opened 2024-08-16
HermiteSplineProfile to improve accuracy of coordinate mapping

### LINKED ISSUE #1200: Improve root finding accuracy on low res grids
After reducing resolution of Quadrature grid in draft pull request, the map_coordinates test is failing because this decreases the number of radial knots used to construct a cubic spline of `iota`, which deteriorates the 3d root finding used by default in `map_coordinates`.

Given that `iota` is the main ingredient in all these coordinate mappings, the solution is to update `map_coordinates` to also compute `iota_r` and feed that into `interpax.interp1d` with the undocumented kwarg `fx=iota_r` to compute Hermite spline.

_Originally posted by @unalmis in https://github.com/PlasmaControl/DESC/issues/1188#issuecomment-2291717947_

### LINKED ISSUE #1203: Issue when map coordinates is used under JIT
The automatic initial guess heuristic for `map_coordinates` breaks under JIT, for example when called inside an objective, because a grid is not constructed in a jit compatible manner.

_Originally posted by @unalmis in https://github.com/PlasmaControl/DESC/pull/1199#discussion_r1720481919_

unalmis:
- [x] Adds `HermiteSplineProfile`.
  - Relevant for #1201.
- [x] Resolves #1200.
  - Another solution would be to use the radial basis polynomials of the quadrature grid for spectral spline of iota.
- [x]  Resolves #1203

### DIFF
--- a/desc/equilibrium/coords.py
+++ b/desc/equilibrium/coords.py
@@ -138,7 +138,7 @@ def map_coordinates(  # noqa: C901
 
     # do surface average to get iota once
     if "iota" in profiles and profiles["iota"] is None:
-        profiles["iota"] = eq.get_profile("iota", params=params)
+        profiles["iota"] = eq.get_profile(["iota", "iota_r"], params=params)
         params["i_l"] = profiles["iota"].params
 
     @functools.partial(jit, static_argnums=1)
--- a/desc/equilibrium/coords.py
+++ b/desc/equilibrium/coords.py
@@ -146,11 +146,11 @@ def map_coordinates(  # noqa: C901
         grid = Grid(y, sort=False, jitable=True)
         data = {}
         if "iota" in deps:
-            data["iota"] = profiles["iota"](grid, params=params["i_l"])
+            data["iota"] = profiles["iota"].compute(grid, params=params["i_l"])
         if "iota_r" in deps:
-            data["iota_r"] = profiles["iota"](grid, dr=1, params=params["i_l"])
+            data["iota_r"] = profiles["iota"].compute(grid, dr=1, params=params["i_l"])
         if "iota_rr" in deps:
-            data["iota_rr"] = profiles["iota"](grid, dr=2, params=params["i_l"])
+            data["iota_rr"] = profiles["iota"].compute(grid, dr=2, params=params["i_l"])
         transforms = get_transforms(basis, eq, grid, jitable=True)
         data = compute_fun(eq, basis, params, transforms, profiles, data)
         x = jnp.array([data[k] for k in basis]).T
@@ -243,7 +243,10 @@ def _initial_guess_heuristic(yk, coords, inbasis, eq, profiles):
         theta = coords[:, inbasis.index(poloidal)]
     elif poloidal == "alpha":
         alpha = coords[:, inbasis.index("alpha")]
-        iota = profiles["iota"](rho)
+        rho = jnp.atleast_1d(rho)
+        zero = jnp.zeros_like(rho)
+        grid = Grid(nodes=jnp.column_stack([rho, zero, zero]), sort=False, jitable=True)
+        iota = profiles["iota"].compute(grid)
         theta = (alpha + iota * zeta) % (2 * jnp.pi)
 
     yk = jnp.column_stack([rho, theta, zeta])
@@ -677,7 +680,7 @@ def get_rtz_grid(
         rtz : rho, theta, zeta
     period : tuple of float
         Assumed periodicity for each quantity in inbasis.
-        Use np.inf to denote no periodicity.
+        Use ``np.inf`` to denote no periodicity.
     jitable : bool, optional
         If false the returned grid has additional attributes.
         Required to be false to retain nodes at magnetic axis.
@@ -691,6 +694,8 @@ def get_rtz_grid(
     grid = Grid.create_meshgrid(
         [radial, poloidal, toroidal], coordinates=coordinates, period=period
     )
+    if "iota" in kwargs:
+        kwargs["iota"] = grid.expand(kwargs["iota"])
     inbasis = {
         "r": "rho",
         "t": "theta",
--- a/desc/profiles.py
+++ b/desc/profiles.py
@@ -614,7 +614,7 @@ class PowerSeriesProfile(_Profile):
 
     def set_params(self, l, a=None):
         """Set specific power series coefficients."""
-        l, a = np.atleast_1d(l), np.atleast_1d(a)
+        l, a = np.atleast_1d(l, a)
         a = np.broadcast_to(a, l.shape)
         for ll, aa in zip(l, a):
             idx = self.basis.get_idx(ll, 0, 0)
@@ -800,10 +800,10 @@ class SplineProfile(_Profile):
     ----------
     values: array-like
         Array containing values of the dependent variable.
-    knots : int or ndarray
+    knots : array-like
         1-D array containing values of the independent variable.
         Values must be real, finite, and in strictly increasing order in [0, 1].
-        If an integer, uses that many uniformly spaced points.
+        If not given, assumes values is uniformly spaced in [0, 1].
     method : str
         Method of interpolation. Default is cubic2.
         - `'nearest'`: nearest neighbor interpolation
--- a/desc/profiles.py
+++ b/desc/profiles.py
@@ -799,11 +799,11 @@ class SplineProfile(_Profile):
     Parameters
     ----------
     values: array-like
-        1-D Array containing values of the dependent variable.
+        1-D array containing values of the dependent variable.
     knots : array-like
         1-D array containing values of the independent variable.
         Must be real, finite, and in strictly increasing order in [0, 1].
-        If not given, assumes ``values`` is uniformly spaced in [0, 1].
+        If ``None``, assumes ``values`` is given on knots uniformly spaced in [0, 1].
     method : str
         Method of interpolation. Default is cubic2.
         - `'nearest'`: nearest neighbor interpolation
@@ -865,10 +865,10 @@ class SplineProfile(_Profile):
         Parameters
         ----------
         grid : Grid
-            locations to compute values at.
+            Locations to compute values at.
         params : array-like
-            spline values to use. If not given, uses the
-            values given by the params attribute
+            Values of the function at ``self.knots``.
+            If not given, uses ``self.params``.
         dr, dt, dz : int
             derivative order in rho, theta, zeta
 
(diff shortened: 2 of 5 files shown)


desc/profiles.py
```diff
@@ -811,12 +811,14 @@ class SplineProfile(_Profile):
...
         name of the profile
+    df : array-like
+        Optional. Values of the function derivative at knot locations.
 
     """
 
     _io_attrs_ = _Profile._io_attrs_ + ["_knots", "_method"]
 
-    def __init__(self, values=None, knots=None, method="cubic2", name=""):
+    def __init__(self, values=None, knots=None, method="cubic2", name="", df=None):
```
unalmis, line 821: > not crazy about adding this option, since it basically overrides method, and if we wanted to do something like this df would need to be part of params so it can be optimized as well. - Rory

desc/profiles.py
```diff
@@ -851,13 +854,14 @@ def params(self):
...
             self._params = jnp.asarray(new)
+            self._params_derivative = None
         else:
             raise ValueError(
                 "params should have the same size as the knots, "
                 + f"got {len(new)} values for {len(self._knots)} knots"
             )
 
-    def compute(self, grid, params=None, dr=0, dt=0, dz=0):
+    def compute(self, grid, params=None, dr=0, dt=0, dz=0, params_derivative=None):
```
unalmis, line 864: > If I load in a vmec wout file, we assign `SplineProfile` for pressure and iota/current, this then will cause issues as `df` is not assigned and so defaults to `None`, but then there are errors when anything is attempted to be computed from that profile - Dario

> What is the error? `df` is not used unless it is not `None`. - Kaya

desc/equilibrium/coords.py
```diff
@@ -138,19 +138,19 @@ def map_coordinates(  # noqa: C901
...
+        profiles["iota"] = eq.get_profile(["iota", "iota_r"], params=params)
         params["i_l"] = profiles["iota"].params
 
     @functools.partial(jit, static_argnums=1)
     def compute(y, basis):
         grid = Grid(y, sort=False, jitable=True)
         data = {}
         if "iota" in deps:
-            data["iota"] = profiles["iota"](grid, params=params["i_l"])
+            data["iota"] = profiles["iota"].compute(grid, params=params["i_l"])
```
unalmis, line 149: in the ripple branch, i noticed you get issues when map_coordinates is used inside jitted objective.compute. The switch to `.compute` here and other changes are necessary to fix the issue

f0uriest: I'm surprised this made any difference, if you supply a grid then the call method is literally the same as compute just with the order swapped?

unalmis: Ah right, it doesn't matter on this line, but it does on line 246 below

unalmis:
After merging #1188 , the error in the map_coordinates_test that fails on that branch is reduced by a factor of 20. If that error is not acceptable, we'll either have to use more than `eq.L_grid` when making the quadrature grid in `eq.compute_profile` or change the `jacfun` in `map_coordinates` to replace the radial derivatives with 1 when the radial coordinate of in basis and outbasis matches, so that the Jacobian is effectively  2d.

desc/profiles.py
f0uriest: Another possible solution would be to make a new profile class specifically for hermite splines where the derivative is required as part of params but method is always cubic. Reasoning as follows:

- For the existing spline class, all the different `method` options are just different ways of estimating the derivative. If the derivatives are supplied then this is redundant.
- Adding `fx` as an optional parameter means that sometimes the number of parameters is `n`, other times `2*n` which I feel like will cause problems if it gets used in optimization.

dpanici:
Can you add a simple check trying to solve with this? I get an error when I try, I can try to debug today/tmrw, I think tho it is due to `eq.params_dict["p_l"]` here being a 2D array instead of a 1D array like it is for the usual SplineProfile, and I think we expect everything inside of `params_dict` to be a 1D array. To fix, either 
- making the params of `HermiteSplineProfile` be twice the length and have the 2nd half be the deriv values?
- or we could have a p_r_l (and etc for i and c) that is for parameters describing the derivs of the profiles. This could lead nicely into allowing for the derivaitves of profiles to be given as an input (like the I'(r) profile which is often given in other codes). This is more work to do though and could be made as an issue for future work

```python
import numpy as np
from desc.examples import get
from desc.profiles import HermiteSplineProfile

eq = get("DSHAPE")
rho = np.linspace(0,1.0,20,endpoint=True)
eq.pressure = HermiteSplineProfile(r=rho, f = eq.pressure(rho), dfdr = eq.pressure(rho,dr=1))
eq.solve()
```

yields 

```python
Cell In[4], line 8
      6 rho = np.linspace(0,1.0,20,endpoint=True)
      7 eq.pressure = HermiteSplineProfile(r=rho, f = eq.pressure(rho), dfdr = eq.pressure(rho,dr=1))
----> 8 eq.solve()

File [~/Research/DESC/desc/equilibrium/equilibrium.py:2058](http://localhost:8888/lab/tree/~/Research/DESC/desc/equilibrium/equilibrium.py#line=2057), in Equilibrium.solve(self, objective, constraints, optimizer, ftol, xtol, gtol, maxiter, x_scale, options, verbose, copy)
   2043 warnif(
   2044     self.N > self.N_grid or self.M > self.M_grid or self.L > self.L_grid,
   2045     msg="Equilibrium has one or more spectral resolutions "
   (...)
   2049     + "to avoid this warning.",
   2050 )
   2051 errorif(
   2052     self.bdry_mode == "poincare",
   2053     NotImplementedError,
   2054     "Solving equilibrium with poincare XS as BC is not supported yet "
   2055     + "on master branch.",
   2056 )
-> 2058 things, result = optimizer.optimize(
   2059     self,
   2060     objective,
   2061     constraints,
   2062     ftol=ftol,
   2063     xtol=xtol,
   2064     gtol=gtol,
   2065     x_scale=x_scale,
   2066     verbose=verbose,
   2067     maxiter=maxiter,
   2068     options=options,
   2069     copy=copy,
   2070 )
   2072 return things[0], result

File [~/Research/DESC/desc/optimize/optimizer.py:220](http://localhost:8888/lab/tree/~/Research/DESC/desc/optimize/optimizer.py#line=219), in Optimizer.optimize(self, things, objective, constraints, ftol, xtol, gtol, ctol, x_scale, verbose, maxiter, options, copy)
    218     objective.build(verbose=verbose)
    219 if linear_constraint is not None and not linear_constraint.built:
--> 220     linear_constraint.build(verbose=verbose)
    221 if nonlinear_constraint is not None and not nonlinear_constraint.built:
    222     nonlinear_constraint.build(verbose=verbose)

File [~/Research/DESC/desc/backend.py:136](http://localhost:8888/lab/tree/~/Research/DESC/desc/backend.py#line=135), in execute_on_cpu.<locals>.wrapper(*args, **kwargs)
    133 @functools.wraps(func)
    134 def wrapper(*args, **kwargs):
    135     with jax.default_device(jax.devices("cpu")[0]):
--> 136         return func(*args, **kwargs)

File [~/Research/DESC/desc/objectives/objective_funs.py:174](http://localhost:8888/lab/tree/~/Research/DESC/desc/objectives/objective_funs.py#line=173), in ObjectiveFunction.build(self, use_jit, verbose)
    172         if verbose > 0:
    173             print("Building objective: " + objective.name)
--> 174         objective.build(use_jit=self.use_jit, verbose=verbose)
    175     self._dim_f += objective.dim_f
    176 if self._dim_f == 1:

File [~/Research/DESC/desc/objectives/linear_objectives.py:691](http://localhost:8888/lab/tree/~/Research/DESC/desc/objectives/linear_objectives.py#line=690), in FixBoundaryR.build(self, use_jit, verbose)
    689     scales = compute_scaling_factors(eq)
    690     self._normalization = scales["a"]
--> 691 super().build(use_jit=use_jit, verbose=verbose)

File [~/Research/DESC/desc/objectives/linear_objectives.py:192](http://localhost:8888/lab/tree/~/Research/DESC/desc/objectives/linear_objectives.py#line=191), in FixParameters.build(self, use_jit, verbose)
    190 # default target
    191 if self.target is None and self.bounds is None:
--> 192     self.target = np.concatenate(
    193         [
    194             np.atleast_1d(param[idx])
    195             for param, idx in zip(tree_leaves(thing.params_dict), self._indices)
    196         ]
    197     )
    199 super().build(use_jit=use_jit, verbose=verbose)

ValueError: all the input arrays must have same number of dimensions, but the array at index 0 has 1 dimension(s) and the array at index 18 has 2 dimension(s)
```

desc/profiles.py
```diff
@@ -876,15 +876,104 @@ def compute(self, grid, params=None, dr=0, dt=0, dz=0):
...
+
+    """
+
+    _io_attrs_ = _Profile._io_attrs_ + ["_knots", "_params"]
+
+    def __init__(self, r, f, dfdr, name=""):
+        super().__init__(name)
+        r, f, dfdr = jnp.atleast_1d(r, f, dfdr)
+        self._knots = r
+        self._params = jnp.stack([f, dfdr])
```
f0uriest, line 915: Everywhere else we assume `params` is a 1d array so I think this should be `concatenate` not `stack`

desc/profiles.py
```diff
@@ -890,35 +892,42 @@ def compute(self, grid, params=None, dr=0, dt=0, dz=0):
...
-        Array containing values of the dependent variable.
-    dfdr: array-like
-        Array containing derivatives of the dependent variable.
+        1-D Array containing values of the dependent variable.
+    df: array-like
+        1-D Array containing derivatives of the dependent variable.
+    knots : array-like
+        1-D array containing values of the independent variable.
+        Must be real, finite, and in strictly increasing order in [0, 1].
+        If not given, assumes ``values`` is uniformly spaced in [0, 1].
```
unalmis, line 906:
```suggestion
        If not given, assumes ``f`` and ``df`` are uniformly spaced in [0, 1].

f0uriest (dismissed):

f0uriest:
> or change the `jacfun` in `map_coordinates` to replace the radial derivatives with zero when the radial coordinate of in basis and outbasis matches, so that the Jacobian is effectively  2d.

This might be a good idea regardless

desc/profiles.py
dpanici:
```suggestion
        If not given, assumes ``f`` and ``df`` correspond to knots uniformly spaced in [0, 1].

dpanici (changes requested):
one tiny comment on the docstring

dpanici (approved):

Issue #1180, opened 2024-08-13
don't mod angles; they are not periodic


unalmis:
Recall  single valued functions $f$ that are periodic in $(\theta, \zeta)$ obey $f(\theta, \zeta) = f(\theta + j  2 \pi, \zeta + k 2 \pi)$. Of course $\zeta \neq \zeta + k 2 \pi$.

1. I think we should not mod any of the angles we compute $\theta, \zeta, \phi, \alpha, \vartheta$ inside our compute functions. Otherwise, we can run into issues because there are functions of interest which are not periodic. [^1]

We don't mod $\phi$ right now, but if we did, attempts to map coordinates $\alpha \in \mathbf{R}, \phi \subset \mathbf{R}$ using $\alpha = \vartheta  \text{mod} (2 \pi) - \iota (\phi \text{mod} (2 \pi))$ will give incorrect results because the map $\alpha \colon \vartheta, \phi \mapsto \alpha(\vartheta, \phi)$ is not periodic. Likewise since the "branch cut" (for my lack of better description) we choose to define $\alpha$ on is centered at $\phi =\vartheta = 0$, we want that

$$\lim_{\vert\phi\vert \to \infty} \vert \frac{\partial \alpha}{\partial \rho}\vert (\rho, \vartheta, \phi)  = \infty \text{ when } \frac{\partial \iota}{\partial \rho} \neq 0$$

but this will not be satisfied if we mod $\phi$ in the compute function.

Currently on `master` we mod $\vartheta$ and $\alpha$. ~This doesn't cause issues for my work,~ but FYI it would impede computing $\phi$ through the relation $(\vartheta - \alpha) / \iota = \phi$.

2. Inside the coordinate mapping routines, we mod the coordinates by the period common to all periodic functions of those coordinates. This is fine if we only care about feeding the output to periodic functions, and should help accelerate convergence of the root finding. If we care about feeding these coordinates to non periodic functions, such as a radial derivative of $\alpha$,  dependency, or chain multiple coordinate mappings together, then we could get wrong results.

I think 1. can be done without drawbacks. ~Not sure if 2 affects anyone.~ 2 affects me

[...]

PR #1204, opened 2024-08-19
Fix some plot issues when NFP differs from one for objects, or when passed-in phi exceeds 2pi/nfp

### LINKED ISSUE #1202: TypeError in 'plot_comparison'
Hello,

When I run the **plot_comparison** function, I get some errors.
It seems that it is necessary for the user to define "phi" parameters to avoid this TypeError. 

**TypeError: reshape total size must be unchanged, got new_sizes (5, 1, 180) for shape (1080,).**

```
import numpy as np
import desc.io

from desc.plotting import plot_comparison

# Load Equilibrium and Make Surface

eq0 = desc.io.load("QA.h5")
eq1 = desc.io.load("QA3_1_iota_0_42_AR5.h5")

surf = FourierRZToroidalSurface(
    R_lmn =   [0.97, 0.33],
    modes_R = [(0, 0), (-1, 0)],
    Z_lmn =   [0.4],
    modes_Z = [(1, 0)],
)

# No error occurs for some Equilibria and Surfaces
plot_comparison([eq0, eq1], rho=np.array(1.0), theta=0);
plot_comparison([surf, eq0], rho=np.array(1.0), theta=0);

# However, error occurs for some Equilibrium and Surface.
plot_comparison([surf, eq1], rho=np.array(1.0), theta=0);

# This error no longer occurs, when I specify the phi parameter.
plot_comparison([surf, eq1], rho=np.array(1.0), theta=0, phi=np.linspace(0, 2*np.pi/eq0.NFP * 5/6, 6));
```

Attached is the Jupyter notebook and *.h5 file I used.
[Bug_report_plot_comparison.zip](https://github.com/user-attachments/files/16647577/Bug_report_plot_comparison.zip)

dpanici:
- in `plot_surfaces` (and `plot_section` and `plot_boundary`) when `phi` is supplied which exceeds `NFP`, the grid used for the map coordinates (a lineargrid with `NFP=eq.NFP` would truncate the `phi`. This is fine unless the new truncated phi has some duplicates. The grid is still a meshgrid of the supplied rho, theta, zeta, however, then the `grid.num_zeta` value, which gives the unique zeta values, might not correspond to the original length of `phi`. This causes reshaping errors like those seen in #1202. These grids though dont need to have the NFP arg at all as they always supply `phi` as an array (instead of supplying `N`), so we can set them to `1` to avoid this issie
- in `plot_comparison` and `plot_boundaries`, there is some ambiguity on what should happen by default when things of differing NFP are passed in. This changes to by default, throw an error if multiple nonaxisymmetric objects with difering NFP are passed in (as it is not clear what phis to do, and even then the plot title "phi*NFP/2pi" is ambiguous (which NFP does it refer to?)). If there are differing NFPs but the only ones with nonzero NFPs have the same NFP, and the remaining are axisymmetric objects, it will change the axisymmetric object NFP to match the nonaxisymetric ones (we can do that and it does not affect the plot since they are the same at every zeta).

The change in `plot_boundaries` behavior is something I am not 100% sure on, what do people think it should be? we could leave as is and just plot a bunch of XS, maybe with the phi being determined inside of the for loop instead of outside, so that if the default is used you get (0,2np.pi,4,endpoint=False). The only weird thing is now you have a bunch of XS where they are at differing angles and it is not clear how they should be compared to eachother.

Resolves #1202


unalmis (commented):
For your first point, isn't the root issue  #1180?

Coordinates are not periodic functions, so we shouldn't treat them as such in the compute functions or constructing grids.

Here is my proposed solution:
1. Change `LinearGrid` to not mod $\zeta$ by $2 \pi / \text{NFP}$ and also not mod $\theta$ by $2 \pi$.
2. In `LinearGrid`, if the coordinates span multiple periods, that is if `np.ptp`($\zeta$) $> 2 \pi / \text{NFP}$ or `np.ptp`($\theta$) $> 2 \pi$, then set the attributes `_spacing` and `_weights` to `None`.

Reasoning:
1. Periodic functions should not be affected by this logic. This would be a bug fix for non-periodic functions.
2. Those attributes are quadrature weights for an integration over a geometry of NFP periods. When coordinates span a different length than that, one is effectively defining multiple copies of a volume or surface. We always want to compute an integral over NFP periods, so we modify the weights accordingly.
In theory one could always modify the quadrature weights so that the integral is over the domain we want regardless of the coordinate span,  but the API of `LinearGrid` is too broad to make that simple.

dpanici:
Throw an error when `NFP` differs and say `phi` must be given explicitly

desc/plotting.py
```diff
@@ -2082,6 +2093,28 @@ def plot_boundaries(
...
+    if not np.allclose(NFPs, NFPs[0]) and np.any(Ns == 0):
+        # if all NFPs are not equal, maybe there are some axisymmetric
+        # objects. We can try to change those to match the NFP of the first
+        # of the nonaxisymmetric objects
+        eqs = [thing.copy() for thing in eqs]  # make copy so we dont modify originals
+        NFP_nonax = int(NFPs[NFPs > 1][0])
+        [
+            thing.change_resolution(NFP=NFP_nonax if thing.N == 0 else thing.NFP)
+            for thing in eqs
+        ]
```
f0uriest, line 2107: can simplify all this stuff by just requiring that `phi` be given explicitly whenever `NFP` differs.

f0uriest (changes requested):

f0uriest (approved):

desc/plotting.py
```diff
@@ -2083,7 +2090,17 @@ def plot_boundaries(
...
+    # if NFPs are not all equal, means there are
+    # objects with differing NFPs, which it is not clear
+    # how to choose the phis for by default, so we will throw an error
+    # unless phi was given.
     phi = parse_argname_change(phi, kwargs, "zeta", "phi")
+    errorif(
+        not np.allclose([thing.NFP for thing in eqs], eqs[0].NFP) and phi is None,
+        ValueError,
+        "supplied objects must have the same number of field periods, "
+        "or if there are differing field periods, `phi` must be given explicitly.",
```
YigitElma, line 2102: Can we add the supplied NFPs in the error message as we usually do i.e. "same nfp expected but got {[thing.NFP for thing in eqs]}"? Also, maybe also mentioning about axisymmetric case is better since that is a more common case (but with this pr we dont let it)

YigitElma (changes requested):

YigitElma (approved):

f0uriest (approved):

desc/plotting.py
```diff
@@ -2085,7 +2092,21 @@ def plot_boundaries(
         fig, ax = plot_boundaries((eq1, eq2, eq3))
 
     """
+    # if NFPs are not all equal, means there are
+    # objects with differing NFPs, which it is not clear
+    # how to choose the phis for by default, so we will throw an error
+    # unless phi was given.
     phi = parse_argname_change(phi, kwargs, "zeta", "phi")
+    errorif(
+        not np.allclose([thing.NFP for thing in eqs], eqs[0].NFP) and phi is None,
```
YigitElma, line 2101: @dpanici Maybe we should change this to 
```python
not np.allclose([thing.NFP for thing in eqs], eqs[0].NFP) and (phi is None or isinstance(phi, numbers.Integral)),
```
Because integer `phi` is still ambigious.

dpanici: ah, what does it do rn if phi integral is passed in? btw 0->2pi or 0->NFP of the first thing?

YigitElma: It uses 0 to 2pi/nfp of the last thing.

YigitElma: Also, doesn't throw any error or warning.

PR #1205, opened 2024-08-19
Allow `BoundaryError` etc. to take multiple `MagneticField` objects

### LINKED ISSUE #1164: Add `external_field` argument to `QuadraticFlux` and other normal field objectives
If I want to say optimize a modular coilset in the presence of a background `ToroidalField`, currently we would have to do so with a `SumMagneticField` object, passing in the coils and the TF. however, our coil objectives cannot handle a `SumMagneticField` object so we would be unable to penalize coil coil distance or curvature or anything like that.

If we add in a `external_field` argument to these objectives that can take in a static external field, we would avoid these issues.

A longer term fix would be to get the coil objectives to work with objects where the field is not given solely by coilsets.

ddudt:
Resolves #1164 

This allows for optimization with multiple different types of magnetic fields in combination; for example, a `MixedCoilSet`, a `FourierCurrentPotentialField`, and a `VerticalMagneticField`. Objectives like `QuadraticFlux` need to know about the total magnetic field from all of these sources, and you could combine them all in a `SumMagneticField` to make that work. However, that then breaks the logic of objectives that expect a particular type of magnetic field, such as the coil objectives. Rather than editing the logic of all other objectives to work with a `SumMagneticField`, this PR edits the logic of `QuadraticFlux` to take a list of fields and internally combines them into a `SumMagneticField` to get the total field. 

TODO:

- [x] BoundaryError
- [x] BoundaryErrorNESTOR
- [x] QuadraticFlux
- [x] VacuumBoundaryError
- [x] tests


rahulgaur104:
Is it a good idea to do this for CoilLength, CoilCurvature and other coil shape metrics?
The idea is give the user freedom to have the same objective applied with different bounds/weights to different coils.

ddudt:
> Is it a good idea to do this for CoilLength, CoilCurvature and other coil shape metrics? The idea is give the user freedom to have the same objective applied with different bounds/weights to different coils, right? Currently, one can only do that with QuadraticFlux and BoundaryError.

This isn't needed for the Coil geometry objectives. This PR only affects objectives related to the total magnetic field, where you would normally have to combine multiple coils into a single `CoilSet` object. You can already use a separate `CoilLength` objective for each individual `Coil` object.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Optionally can also contract the profiles of the original ``Equilibrium`` so tha
 - Adds error for incorrect grids in ``desc.objectives.BootstrapRedlConsistency`` and when computing ``current Redl`` and ``<J*B> Redl`` compute quantities
 - Allows Redl compute quantities to use SplineProfile
 - Updates Redl bootstrap current consistency tutorial to include a ``SplineProfile`` optimization
+- Allows the following objectives to take multiple optimizable "things" and internally sums their magnetic fields together: ``desc.objectives.BoundaryError``, ``desc.objectives.BoundaryErrorNESTOR``, ``desc.objectives.VacuumBoundaryError``, ``desc.objectives.QuadraticFlux``
```
dpanici, line 12: I thought we were gonna do this "quietly" ?

dpanici (dismissed):
If people are fine with the change being in changelog then I approve

CHANGELOG.md
YigitElma: Why exactly? It looks to me worth mentioning.

YigitElma (dismissed):

dpanici:
Let's wait for dev meeting before merging

YigitElma (approved):

dpanici (approved):

PR #1210, opened 2024-08-20
add fixtures to parameterization

### LINKED ISSUE #1104: Refactor coil tests to use `pytest.mark.parametrize` which treats it as different tests rather than 1 big one so easier to debug and run specific ones in case of failure.
> might be able to refactor this to use `pytest.mark.parametrize` which treats it as different tests rather than 1 big one so easier to debug and run specific ones in case of failure.

_Originally posted by @f0uriest in https://github.com/PlasmaControl/DESC/pull/1099#discussion_r1667216952_

kianorr:
resolves #1104


kianorr:
Not sure if it's worth it to load each h5 file every time.

Also I left out `nested_coils = MixedCoilSet(sym_coils, mixed_coils, check_intersection=False)` and `spline_coil = mixed_coils.coils[-1].copy()` because I couldn't think of a clean way to incorporate them yet.

tests/test_examples.py
```diff
@@ -1314,16 +1314,22 @@ def test_second_stage_optimization_CoilSet():
...
+    "coil,optimizer,path",
+    [
+        (FourierPlanarCoil(), "fmintr", ""),
+        (FourierRZCoil(), "fmintr", ""),
+        (FourierXYZCoil(), "fmintr", ""),
+        ("DummyCoilSet", "lsq-exact", "output_path_sym"),
+        ("DummyCoilSet", "lsq-exact", "output_path_asym"),
+        ("DummyMixedCoilSet", "lsq-exact", "output_path"),
+    ],
+)
```
dpanici, line 1327: make the coilsets as conftest fixtures

kianorr:
I added a nested coil set fixture so it's working now. But, I still need to decide on whether we should save the h5 file in the fixture or if we should just return the coil set.

dpanici:
close when #1362 is merged in, I think

YigitElma:
closing since #1362 solved this

PR #1211, opened 2024-08-21
Update `VMECIO` to specify Nyquist spectrum and fix issues with asymmetric wouts

### LINKED ISSUE #1062: Saving asymmetric eq as a `wout` has some errors
@SebereX found a probable bug, when saving a DESC eq as a wout and running booz xform for an AR scan of a NAE solution, which should converge to the NAE values as AR is increased, DESC does not converge (which might separately be a problem with high aspect ratio solves) and seems to have a flipped sign for some of the saved B coefficients (the sin theta coefficients specifically)

DESC looks like
WhatsApp Image 2024-06-05 at 9 26 50 AM

While the VMEC ones look like
WhatsApp Image 2024-06-05 at 9 27 22 AM

We dont have a test for wout asym saving so we should add this and check results (load a asym vmec in DESC, save as wout, check quantities agains the original wout within some tolerance)

dpanici:
- `plot_vmec_comparison` sneakily only worked for `sym=True` `wout` files, this changes it to work correctly
- allow `M_nyq` and `N_nyq` to be passed as keywords to `VMECIO.save`
- fix typo in axis loading from an asymmetric VMEC equilibrium
- fix sneaky bug in asymmetric wout saving where some of the derived quantities were being fit with the transform directly instead of using the `fullfit` function which did the fit AND negated the correct modes to account for poloidal angle reversal... this will help in #680 and also resolve #1062 I believe, as the magnetic field coeffs being saved before were wrt the wrong poloidal angle (and the opposite one as the jacobian fourier coefficients saved).
- [x] make test with an asymmetric stellarator 

hopefully resolves #1062 

Notes
------
For an `NFP=4` stellarator, VMEC with `NZETA=200` yields a max nyquist N of `400` while `NZETA=100` yields a max nyquist N of `200`, which leads me to believe the max nyquist toroidal modenumber is set as `NFP * NZETA/2` or something along those lines (I support with a line from educational VMEC : https://github.com/jonathanschilling/educational_VMEC/blob/ad4db0a56b23c3324df3d6101781b69b2e3b7e13/src/fixaray.f90#L28)

default values of ntheta, nzeta are [here](https://github.com/jonathanschilling/educational_VMEC/blob/ad4db0a56b23c3324df3d6101781b69b2e3b7e13/src/read_indata.f90#L91)

defaults:
NTHETA = 2M+6
NZETA = 2N+4

then nyquist freq determined by
mnyq = MAX(0, NTHETA, 2*(M-1))
nnyq = MAX(0, NZETA, 2*ntor)

translates to default of M+3 (in practice it seems to be M+4... don't understand completely but this is what the `wout` has)) and N+2 for the nyquist max spectrum defaults

When VMEC is given a right-handed bdry (such that sqrt(g)>0 ), it seems to change the modes such that the new bdry is left-handed (but the way of changing the bdry handedness changes where the theta=0 point is on the bdry, and is NOT how we enforce the change in handedness in DESC) (this is bc they change [theta -> pi - theta] (https://github.com/jonathanschilling/educational_VMEC/blob/ad4db0a56b23c3324df3d6101781b69b2e3b7e13/src/flip_theta.f90#L29).)
For example, the asymmetric rotating ellipse bdry  given by

image

will be, in the resulting wout file, replaced by

image

This is a pitfall if trying to compare the wout from a VMEC run to a DESC equilibrium saved as a wout, as the output bdry modes in VMEC are not what you used as the input modes, AND the theta definition is changed not only in its direction but also with a pi shift.


dpanici:
check `NZETA` does what I expect

desc/vmec.py
```diff
@@ -1672,19 +1691,30 @@ def compute_theta_coords(cls, lmns, xm, xn, s, theta_star, zeta, si=None):
         if si is None:
             si = np.linspace(0, 1, lmns.shape[0])
             si[1:] = si[0:-1] + 0.5 / (lmns.shape[0] - 1)
-        lmbda_mn = interpolate.CubicSpline(si, lmns)
+        lmbda_mns = interpolate.CubicSpline(si, lmns)
+        if lmnc is None:
+            lmbda_mnc = lambda s: 0
+        else:
+            lmbda_mnc = interpolate.CubicSpline(si, lmnc)
```
rahulgaur104, line 1698: Is cubic spline the best way to interpolate fourier coefficients in the radial direction?
A fourier coefficient may not always be smooth.

tests/test_vmec.py
```diff
@@ -874,6 +866,369 @@ def test_vmec_save_2(VMEC_save):
...
+    np.testing.assert_allclose(
+        vmec.variables["DShear"][20:100], desc.variables["DShear"][20:100], rtol=3e-2
+    )
+    np.testing.assert_allclose(
+        vmec.variables["DCurr"][20:100],
+        desc.variables["DCurr"][20:100],
+        atol=1e-4,  # nearly zero bc vacuum
+    )
+    np.testing.assert_allclose(
+        vmec.variables["DWell"][20:100], desc.variables["DWell"][20:100], rtol=1e-2
```
rahulgaur104, line 1033: Why exclude points close to the axis? Because of strong variations?

dpanici: b/c VMEC is notorious for being inaccurate near-axis. But in this case the two solutions (the DESC and VMEC) one are not exactly similar (even visually) near-axis, so we also want to exclude it for that reason.

I dont need this test to be a validation of DESC v VMEC, I just want it to tell me if I am way off in the asymmetric quantities basically

desc/vmec.py
dpanici: It might not be, esp. at edge, but I don't feel the need to change as this is not really used anywhere important in the code

f0uriest: I think vmec itself implicitly assumes the fourier coeffs at at least c1 smooth

desc/vmec.py
```diff
@@ -242,8 +248,9 @@ def save(cls, eq, path, surfs=128, verbose=1):  # noqa: C901 - FIXME - simplify
         NFP = eq.NFP
         M = eq.M
         N = eq.N
-        M_nyq = M + 4
-        N_nyq = N + 2 if N > 0 else 0
+        M_nyq = M + 4 if M_nyq is None else M_nyq
```
f0uriest, line 251: are these still the correct defaults? It sounded like `N_nyq = NZETA*2`?

f0uriest: also you can use `setdefault` for a cleaner version of this

dpanici: These are still correct, I think. At least VMEC by default gives M nyq of Mpol-1+4, and Mpol-1 = our Mpol

f0uriest (approved):

desc/vmec.py
```diff
@@ -208,7 +208,9 @@ def load(
         return eq
 
     @classmethod
-    def save(cls, eq, path, surfs=128, verbose=1):  # noqa: C901 - FIXME - simplify
+    def save(  # noqa: C901 - FIXME - simplify
+        cls, eq, path, surfs=128, verbose=1, M_nyq=None, N_nyq=None
```
ddudt, line 212: I won't force you to change this, but I would prefer to consistently have `verbose` always be the last kwarg

desc/vmec.py
```diff
@@ -242,8 +248,9 @@ def save(cls, eq, path, surfs=128, verbose=1):  # noqa: C901 - FIXME - simplify
         NFP = eq.NFP
         M = eq.M
         N = eq.N
-        M_nyq = M + 4
-        N_nyq = N + 2 if N > 0 else 0
+        M_nyq = M + 4 if M_nyq is None else M_nyq
+        N_nyq = N + 2 if N_nyq is None else N_nyq
+        N_nyq = 0 if int(N) == 0 else N_nyq
```
ddudt, line 253: Do we want to force this even if the user purposely inputs a finite value for `N_nyq`? If that shouldn't be allowed, it would be better to throw an error or warning rather than secretly overwrite the user input.

desc/vmec.py
```diff
@@ -1782,36 +1812,80 @@ def compute_coord_surfaces(cls, equil, vmec_data, Nr=10, Nt=8, Nz=None, **kwargs
...
             vmec_data["xm"],
             vmec_data["xn"],
             t_nodes[:, 0],
             t_nodes[:, 1],
             t_nodes[:, 2],
+            lmnc=vmec_data["lmnc"] if not sym else None,
         )
 
         t_nodes[:, 1] = v_nodes
+        if sym:
```
ddudt, line 1828: I don't think you actually need this if statement? It looks like the only difference between the two cases is the `sym` argument, so you can just call it with `sym=sym`

ddudt (changes requested):

desc/vmec.py
dpanici: I won't just because changing it means changing positional args that others might be using and that would break code. I know we have not committed to an API but for this I will keep as is

desc/vmec.py
dpanici: hm, I can throw a warning, if N is zero it makes no sense to have N_nyq, so I think a warning is ok. I will make that change

desc/vmec.py
dpanici: The way `vmec.interpolate` is made, `sym` lets you return the cos and sin sym modes independently, `sym=False` will add the two. I did not want to change the logic of it, so I think this if statement is needed

f0uriest (approved):

rahulgaur104 (approved):

desc/vmec.py
unalmis: maybe relevant: https://docs.python.org/3.8/whatsnew/3.8.html#positional-only-parameters

ddudt: This is cool! In a future PR I would be in favor of requiring stuff like `verbose` to be keyword arguments. That would help with backward-compatibility of the code

ddudt (approved):

PR #1212, opened 2024-08-21
Make integration stuff visible to users and update dev guide

### LINKED ISSUE #723: Add Example Usage of Compute Utils to docs
Would be useful to show how to use surface_averages etc in docs for users, as it is easier than them looking at the source and otherwise hey wouldn't even know these utilities exist

unalmis:
This PR moves the integration utilities to their own subfolder so that they are visible to users. We have decided not to include in public API, because that's for optimization stuff.

- [x] Increase visibility of integration utilities.
- [x] Resolve #723 by adding link to tutorial in docstring of `surface_integrals.py`.
- [x] Update `grid.py` developer guide.
  - Cherry picked from #353 with the new section from #1101 that emphasizes poloidal midplane symmetry is not stellarator symmetry. Removed section on custom grids.


desc/integrals/surface_integral.py
unalmis, line 1: We'll have a `bounce_integral` file in this folder as well.

desc/integrals/surface_integral.py
```diff
@@ -0,0 +1,723 @@
...
+"""
+
+from desc.backend import cond, fori_loop, jnp, put
+from desc.grid import ConcentricGrid, LinearGrid
+from desc.utils import errorif, warnif
+
+# TODO: Make the surface integral stuff objects with a callable method instead of
+#       returning functions. Would make simpler, allow users to actually see the
+#       docstrings of the methods, and less bookkeeping to default to more
+#       efficient methods on tensor product grids.
```
unalmis, line 13: I know JAX style is functional programming, but i think OOP is better

unalmis:
This tutorial is thorough and existed when the issue was created. Does this resolve the issue?

rahulgaur104 (approved):

desc/integrals/surface_integral.py
f0uriest: yeah I've been seeing similar things on #1043. Probably less of an issue here since the local functions are usually created under jit themselves so they won't cause a lot of recompilation but classes with jitted methods also work well.

f0uriest (commented):
We could also add an `__init__.py` to `integrals` and import stuff there to make it more uniform.

IE,
```python
from desc.integrals import compute_B_plasma, surface_integrals
```
rather than
```python
from desc.integrals.surface_integrals import surface_integral
from desc.integrals.singularities import compute_B_plasma
```

f0uriest (approved):

f0uriest (approved):

ddudt (approved):
The updated grid notebook looks really nice!

YigitElma:
I will have a couple of suggestions for the grid notebook, but there is no easy way to put comments on the code. Instead, I will try to write them here. Can you wait for tonight?

unalmis:
> I will have a couple of suggestions for the grid notebook, but there is no easy way to put comments on the code. Instead, I will try to write them here. Can you wait for tonight? @unalmis
> 
> Also, I will just look at the notebook, so some of the things might be from the the previous version :(
Might be easier to write them here: https://app.reviewnb.com/PlasmaControl/DESC/blob/integrals/docs/notebooks/dev_guide/grid.ipynb/

YigitElma:
> Might be easier to write them here: https://app.reviewnb.com/PlasmaControl/DESC/blob/integrals/docs/notebooks/dev_guide/grid.ipynb/

Oh thanks, it a lot better :D

YigitElma:
Ok, I added some comments. I guess none of them is regarding your changes but the old version itself. You can either make changes or I can do them on another PR, it is up to you @unalmis

unalmis:
@YigitElma Feel free to make any changes, thanks. It's fine to do it here, or we can merge this and do it in the dev guide pull request later.

PR #1213, opened 2024-08-21
Higher splitting for Github Actions and Cache venv

rahulgaur104:
Here is the description for this PR by @YigitElma 

 - I created a new workflow that runs every 2 days and updates the requirements and stores them in a venv that is cached
 - Black, unit, regression, notebook and benchmark workflows use the cached venv (which takes around 4-8seconds whereas installing dependencies usually more than a minute)
 - Increase the spitting of `unit`, `notebook`, and `benchmark` workflows.
 - For some reason I couldn't make the linting test faster using venv (so set it back to old version which is fine), actually for some reason venv made it way slower around 8 minutes?!!!?

### Other Fixes
- We stopped using Python 3.8 in our tests with #923 but we forgot to remove them from weekly tests. Since then (I guess the dates check) the weekly tests fail (for the last 3 months). I removed Python 3.8 from weekly tests.


dpanici:
Does this work with pytest-split?

rahulgaur104:
> Does this work with pytest-split?
I am using pytest-xdist. It definitely works on della and my laptop. Is pytest-split better or already being used?

rahulgaur104:
Nah, everything slows down. But I think using this on della and your laptop is still a good idea.

f0uriest:
Yeah I tried this a while back. IIRC another issue is it would often crash on CI causing random failures, so even if it was faster it's also a lot more annoying to have to restart jobs constantly

YigitElma:
@rahulgaur104 Can I use this PR to test if increasing the splitting helps?

rahulgaur104:
@YigitElma  Go ahead!

YigitElma:
So, I counted the jobs running at the same time, and I think we dont have team account, so only 20 jobs at a time. With this change, I increased the number jobs per pr to 21 but 2 of them (black and linting) takes a minute to complete, hence 19 actual jobs. This means CI will test a single pr at a time but since each job is shorter, we might still be better off.

YigitElma:
This is not a solution to @f0uriest s point of running tests locally under 30mins, but still using github resources more efficiently.

YigitElma:
Also, we can try to add a dependency cache which gets updated every week. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows

YigitElma:
@rahulgaur104 since this is your PR initially, I cannot ask for your review, but feel free to give feedback.

YigitElma:
@ddudt @f0uriest @dpanici I don't want to change @rahulgaur104 s message, so here is the description for this PR.
1. I created a new workflow that runs every 2 days and updates the requirements and stores them in a venv that is cached
2. Black, unit, regression, notebook and benchmark workflows use the cached venv (which takes around 4-8seconds whereas installing dependencies usually more than a minute)
3. For some reason I couldn't make the linting test faster using venv (so set it back to old version which is fine), actually for some reason venv made it way slower around 8 minutes?!!!?


- Maybe instead of dividing regression tests more we should use it for notebook test?
- Benchmarks would be a bit to much effort to split, so kept it unchanged (except for the venv)

f0uriest (dismissed):

f0uriest:
Recommend updating description before merging since that's what gets used at the commit message

rahulgaur104:
Looks good to me!

f0uriest:
How does caching work if we update requirements as part of a PR? I think in that case it should use the new requirements file, otherwise we're not really testing the parts that changed.

YigitElma:
> How does caching work if we update requirements as part of a PR? I think in that case it should use the new requirements file, otherwise we're not really testing the parts that changed.

That's a good point! One of the reasons I didn't want to merge this in yet is that I want to see the cache workflow works fine, and rest of the workflows don't throw an error while it is updating the cache files.

I can try to add an if statement to check if requirements file is changed

.github/workflows/pythonpip.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Cache Python environment
+        id: cache-env
+        uses: actions/cache@v3
+        with:
+          path: ./.venv-${{ matrix.python-version }}
+          key: ${{ runner.os }}-venv-${{ matrix.python-version }}-${{ hashFiles('devtools/dev-requirements.txt') }}
```
YigitElma, line 30: @f0uriest Ok so this line will hash the contents of the file, so any change to the file will create a new cache. But the problem is if you update `requirements.txt` I don't think it will be reflected to the hashing since `devtools/dev-requirements.txt` only has is as `-r requirements.txt`. Any suggestions?

YigitElma: I can change it like `-${{ hashFiles('devtools/dev-requirements.txt') }}-${{ hashFiles('requirements.txt') }}`

YigitElma:
Figure out why different checkout methods create different hash for the dependency files? Even though the contents are the same... Make thr hash just by the content related

.github/workflows/benchmark.yml
```diff
@@ -22,6 +22,12 @@ concurrency:
 jobs:
   benchmark:
     runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ github.token }}
+    strategy:
+      matrix:
+        python-version: ['3.9']
+        group: [1, 2]
```
f0uriest, line 30: im not sure if we want to split the benchmarks, since they need to run on the same hardware for it to be a meaningful comparison

.github/workflows/pythonpip.yml
f0uriest, line 1: can we rename this file to something more descriptive

.github/workflows/pythonpip.yml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/cache@v4
+        with:
+          path: .venv-${{ matrix.python-version }}
+          key: ${{ runner.os }}-venv-${{ matrix.python-version }}-${{ hashFiles('devtools/dev-requirements.txt', 'requirements.txt') }}
+
+      - name: Verify virtual environment activation
+        run: |
+          source .venv-${{ matrix.python-version }}/bin/activate
+          python --version
+          pip --version
```
f0uriest, line 54: could also run `pip list` to print a list of all the packages installed, might be useful for debugging workflow stuff at some point

.github/workflows/regression_test.yml
```diff
@@ -29,20 +31,34 @@ jobs:
...
+
+      - name: Set up virtual environment if not restored from cache
+        if: steps.restore-env.outputs.cache-hit != 'true'
         run: |
+          gh cache list
+          python -m venv .venv-${{ matrix.python-version }}
+          source .venv-${{ matrix.python-version }}/bin/activate
           python -m pip install --upgrade pip
           pip install -r devtools/dev-requirements.txt
-          pip install matplotlib==3.7.2
```
f0uriest, line 36: I think we want to keep this line so that plotting regression tests don't randomly fail due to version differences

.github/workflows/unittest.yml
```diff
@@ -18,30 +18,50 @@ jobs:
...
+
+      - name: Set up virtual environment if not restored from cache
+        if: steps.restore-env.outputs.cache-hit != 'true'
         run: |
+          gh cache list
+          python -m venv .venv-${{ matrix.combos.python_version }}
+          source .venv-${{ matrix.combos.python_version }}/bin/activate
           python -m pip install --upgrade pip
           pip install -r devtools/dev-requirements.txt
-          pip install matplotlib==3.7.2
```
f0uriest, line 38: see above

f0uriest (changes requested):

.github/workflows/benchmark.yml
YigitElma: When we split master and latest PR still runs on the same hardware. The only difference is some portion of the benchmarks runs on the second job. I am assuming pytest-split always splits the same tests. SO, compile and jac tests may run on different hardware but the same test for aster and PR runs on the same job and thus hardware

f0uriest (approved):

dpanici:
mark new actions as "required to pass" before merging in repo settings once this is in

dpanici (approved):

YigitElma:
> mark new actions as "required to pass" before merging in repo settings once this is in

I tried to do that but I don't have access to those settings. For me, the settings page is very limited. @f0uriest or @dpanici can you do that?

YigitElma:
And also you may squash merge this PR, because I committed relentlessly to test the github actions

PR #1214, opened 2024-08-21
Reorder nodes from ``create_meshgrid`` to match other sorting

f0uriest:
- Ensure all tensor product grids are flattened in a consistent order
- adds a utility method for reshaping flattened data into ndarrays such that each coordinate varies along a single axis

### DIFF
--- a/desc/grid.py
+++ b/desc/grid.py
@@ -598,6 +598,52 @@ class _Grid(IOAble, ABC):
             )
         return x
 
+    def meshgrid_reshape(self, x, order):
+        """Reshape data to match grid coordinates.
+
+        Given flattened data on a tensor product grid, reshape the data such that
+        the axes of the array correspond to coordinate values on the grid.
+
+        Parameters
+        ----------
+        x : ndarray, shape(N,) or shape(N,3)
+            Data to reshape.
+        order : str
+            Desired order of axes for returned data. Should be a permutation of
+            ``grid.coordinates``, eg ``order="rtz"`` has the first axis of the returned
+            data correspond to different rho coordinates, the second axis to different
+            theta, etc.  ``order="trz"`` would have the first axis correspond to theta,
+            and so on.
+
+        Returns
+        -------
+        x : ndarray
+            Data reshaped to align with grid nodes.
+        """
+        errorif(
+            not self.is_meshgrid,
+            ValueError,
+            "grid is not a tensor product grid, so meshgrid_reshape doesn't "
+            "make any sense",
+        )
+        errorif(
+            sorted(order) != sorted(self.coordinates),
+            ValueError,
+            f"order should be a permutation of {self.coordinates}, got {order}",
+        )
+        shape = (self.num_poloidal, self.num_rho, self.num_zeta)
+        vec = False
+        if x.ndim > 1:
+            vec = True
+            shape += (-1,)
+        x = x.reshape(shape, order="F")
+        x = jnp.moveaxis(x, 1, 0)  # now shape rtz/raz etc
+        newax = tuple(list(self.coordinates).index(c) for c in order)
+        if vec:
+            newax += (3,)
+        x = jnp.transpose(x, newax)
+        return x
+
 
 class Grid(_Grid):
     """Collocation grid with custom node placement.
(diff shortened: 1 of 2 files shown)


tests/test_grid.py
```diff
@@ -738,20 +738,61 @@ def test_meshgrid(self):
...
+        np.testing.assert_allclose(
+            grid.meshgrid_reshape(grid.nodes[:, 1], "rzt")[:, :, 3], t[3]
+        )
+        # reshaping tzr should have zeta along 2nd axis
+        np.testing.assert_allclose(
+            grid.meshgrid_reshape(grid.nodes, "tzr")[:, 0, :, 2], z[0]
+        )
+        np.testing.assert_allclose(
+            grid.meshgrid_reshape(grid.nodes, "tzr")[:, 3, :, 2], z[3]
+        )
```
ddudt, line 786: Why is the array 4D in this case? And is there a reason why we can't check all the nodes instead of only indices 0 and 3?

f0uriest: we're reshaping the full nodes array from `(K,3)` to `(Nt, Nz, Nr, 3)`, in the other cases we're only reshaping a single column from `(K,)` to `(Nr, Nt, Nz)` etc.

unalmis (commented):
Looks good but I have a request to move a test here.

If I remember correctly, in that test the reason I'm using a function for reshaping rather than coordinates themselves is because the functions are not generated by a tensor-product while the coordinates are. This might be better because the values of the coordinates may get mixed up but not be detected because they have the same value.

unalmis (dismissed):

desc/grid.py
```diff
@@ -598,6 +598,52 @@ def replace_at_axis(self, x, y, copy=False, **kwargs):
...
+            f"order should be a permutation of {self.coordinates}, got {order}",
+        )
+        shape = (self.num_poloidal, self.num_rho, self.num_zeta)
+        vec = False
+        if x.ndim > 1:
+            vec = True
+            shape += (-1,)
+        x = x.reshape(shape, order="F")
+        x = jnp.moveaxis(x, 1, 0)  # now shape rtz/raz etc
+        newax = tuple(list(self.coordinates).index(c) for c in order)
```
unalmis, line 641: don't need call to list

desc/grid.py
```diff
@@ -598,6 +598,52 @@ def replace_at_axis(self, x, y, copy=False, **kwargs):
...
+        if vec:
+            newax += (3,)
```
unalmis, line 643: -1 would be better here because I sometimes I am reshaping a subset of the 3 components of a vector.
For example to perform this operation with `meshgrid_reshape`, we need to use `-1` instead of `3`
``grid.nodes[:, 1:].reshape(num rho, -1, 2)```

unalmis: Actually  never mind, this is just transpose order

unalmis (approved):

dpanici (approved):

desc/grid.py
```diff
@@ -598,6 +598,52 @@ def replace_at_axis(self, x, y, copy=False, **kwargs):
...
+            sorted(order) != sorted(self.coordinates),
+            ValueError,
+            f"order should be a permutation of {self.coordinates}, got {order}",
+        )
+        shape = (self.num_poloidal, self.num_rho, self.num_zeta)
+        vec = False
+        if x.ndim > 1:
+            vec = True
+            shape += (-1,)
+        x = x.reshape(shape, order="F")
```
unalmis, line 639: `f = grid.meshgrid_reshape(f)` seems simpler than `f = f.reshape(grid.num_poloidal, grid.num_radial, grid.num_toroidal, order="F").`

Is it worth to default `order=None` and early exit?

```python
inorder = self.coordinates[1] + self.coordinates[0] + self.coordinates[2]
order = setdefault(order, inorder)
...
x = x.reshape(shape, order="F")
if order == inorder:
  return x
x = jnp.swapaxes(x, 1, 0)
```

f0uriest: I considered that but thought it would be clearer and less error prone to make the ordering explicit.

rahulgaur104:
Created and merged in 8 hours is wild!

PR #1220, opened 2024-08-22
`FusionPower` objective

ddudt:
Very similar to #1168, but this one is for output power.


f0uriest (changes requested):

desc/compute/_equil.py
```diff
@@ -843,3 +843,39 @@ def _P_ISS04(params, transforms, profiles, data, **kwargs):
...
+)
+def _P_fusion(params, transforms, profiles, data, **kwargs):
+    reaction = kwargs.get("fuel", "T(d,n)4He")
+    match reaction:
+        case "T(d,n)4He":
+            energy = 3.52e6 + 14.06e6  # eV
+        case "D(d,p)T":
+            energy = 1.01e6 + 3.02e6  # eV
+        case "D(d,n)3He":
+            energy = 0.82e6 + 2.45e6  # eV
```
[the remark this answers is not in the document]
ddudt: OK I switched this to `fuel` instead of `reaction`. For now let's only implement DT, and we can worry about adding more complicated fuels later if people want them. 

One idea I though of is adding an additional kwarg to allow the user to specify the energy themselves if they want to include extra reactions between products (and if not supplied it would default to the basic case for that fuel). But I don't love adding lots of extra compute kwargs

dpanici (approved):

desc/objectives/_power_balance.py
```diff
@@ -9,6 +9,183 @@
...
+        deriv_mode="auto",
+        fuel="DT",
+        grid=None,
+        name="fusion power",
+    ):
+        errorif(
+            fuel not in ["DT"], ValueError, f"fuel must be one of ['DT'], got {fuel}."
+        )
+        if target is None and bounds is None:
+            target = 0
```
f0uriest, line 84: Is this the most sensible default?

desc/objectives/_power_balance.py
```diff
@@ -9,6 +9,183 @@
...
+        Parameters
+        ----------
+        use_jit : bool, optional
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        eq = self.things[0]
+        errorif(
```
f0uriest, line 111: should also add a check for ion temp since that's needed as well?

desc/objectives/_power_balance.py
ddudt: A good target will depend on the scale of the machine, but I changed it to 1,000 MW to be realistic for a power plant

desc/objectives/_power_balance.py
```diff
@@ -9,6 +9,188 @@
...
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    target : {float, ndarray}, optional
+        Target value(s) of the objective. Only used if bounds is None.
+        Must be broadcastable to Objective.dim_f. Defaults to ``target=0``.
+    bounds : tuple of {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
+        Both bounds must be broadcastable to to Objective.dim_f.
+        Defaults to ``target=0``.
```
f0uriest, line 33: also need to update this

f0uriest (dismissed):

dpanici (approved):
Approving, but is Fusion power density something that could be of interest/be optimized? would it make sense to have one compute quantity be local fusion power density and then another that does the integration to get total power?

f0uriest (approved):

dpanici (approved):

PR #1222, opened 2024-08-23
`LinkingCurrent` objective

ddudt:
This is a nonlinear form of the existing linear objective `FixSumCoilCurrent`. It ensures the total current through the coils matches the poloidal current of the equilibrium. This is necessary for single-stage optimization where the equilibrium and coils are changing simultaneously.


dpanici:
do #1231

dpanici:
Any progress on this @daniel-dudt ?

dpanici:
Use Linking number calc for linear objective for sum coil currents as well, to use as weights for sum

dpanici:
I think an equivalent sort of objective to ensure that the coils are creating the necessary current, without needing the linking number calculation (though I'd still like that in sometime), is to directly calculate $\int_{o}^{2\pi/NFP} \mathbf{B}_{coil} \cdot \mathbf{e}_{\zeta} d\zeta - \int_{o}^{2\pi/NFP} \mathbf{B}_{DESC} \cdot \mathbf{e}_{\zeta} d\zeta$ which is the relation that Ampere's law tells us (the second term here is $G(\rho=1)$, possibly sans a $2\pi$ factor).

This being zero means the coils are consistent with Ampere's law and avoids needing to worry about the linking currents (though this is a bit more expensive, as it relies on biot-savart integration over the coils to get $\mathbf{B}_{coil}$

tests/test_objective_funs.py
```diff
@@ -2971,6 +3018,7 @@ def test_objective_print_widths():
                 )
 
 
+@pytest.mark.unit
```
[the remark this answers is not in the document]
YigitElma: Ok, I fixed it, it should be fine now.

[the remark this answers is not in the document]
YigitElma: Only `test_objective_docstring` was failing. We fixed some typos in `collect_docs` but forgot to fix them in original `_Objective`

f0uriest:
We had talked about adding the linking number stuff to `FixSumCoilCurrents`, but it might be easier to just add an `eq_fixed` option to this, in which case we precompute `G` and set `self._linear = True`

desc/objectives/_coils.py
```diff
@@ -1364,6 +1365,193 @@ def compute(self, field_params=None, constants=None):
...
+            )
+            eq_linking_current = 2 * jnp.pi * data["G"][0] / mu_0
+
+        coil_currents = jnp.concatenate(
+            [
+                jnp.atleast_1d(param[idx])
+                for param, idx in zip(tree_leaves(coil_params), self._indices)
+            ]
+        )
+        coil_linking_current = jnp.sum(constants["link"] * coil_currents)
```
f0uriest, line 1551: This actually won't work for coilsets with symmetry or NFP, because `link` contains the linking numbers for the full coilset, but `coil_currents` are only the unique coils. Will need to think about this more.

ddudt (commented):
I can't officially approve, but it looks good.

desc/objectives/_coils.py
```diff
@@ -1762,6 +1763,194 @@ def compute(self, params_1, params_2=None, constants=None):
...
+            eq.axis.Z_basis.modes[:, 2],
+            eq.axis.NFP,
+        )
+        dummy_coilset = MixedCoilSet(axis_coil, coil, check_intersection=False)
+        # linking number for coils with axis
+        link = np.round(dummy_coilset._compute_linking_number())[0, 1:]
+
+        self._constants = {
+            "quad_weights": 1.0,
+            "link": link,
```
dpanici, line 1884: by including `link` here we are assuming the coils and the plasma never intersect and thus never change their linking number right? That is fine but might be good to note somewhere, since if a coil does somehow jump through/across the plasma this would be wrong. Though I guess that is a rare occurence

desc/objectives/_coils.py
```diff
@@ -1762,6 +1763,194 @@ def compute(self, params_1, params_2=None, constants=None):
...
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    coil : CoilSet
+        Coil(s) that are to be optimized.
+    grid : Grid, optional
+        Collocation grid containing the nodes to evaluate plasma current at.
+        Defaults to ``LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=eq.sym)``.
```
dpanici, line 1780: no option for the coil grid discretization for the linking number calculation? Is that calculation pretty robust?

desc/objectives/_coils.py
f0uriest: There's already a note about that in the docstring

desc/objectives/_coils.py
f0uriest: Yeah since it's an integer pretty much any reasonable grid will do fine

desc/objectives/_coils.py
```diff
@@ -1762,6 +1763,194 @@ def compute(self, params_1, params_2=None, constants=None):
...
+            Dictionary of equilibrium degrees of freedom, eg ``Equilibrium.params_dict``
+            Only required if eq_fixed=False.
+        constants : dict
+            Dictionary of constant data, eg transforms, profiles etc.
+            Defaults to self._constants.
+
+        Returns
+        -------
+        f : array of floats
+            Linking current error.
```
dpanici, line 1926: techincally just a single float but not a big deal

desc/objectives/_coils.py
```diff
@@ -1762,6 +1763,194 @@ def compute(self, params_1, params_2=None, constants=None):
         return Psi
 
 
+class LinkingCurrent(_Objective):
+    """Target the self-consistent poloidal linking current between the plasma and coils.
```
dpanici, line 1767: should we say somewhere that target should be zero?

dpanici: as I could see target being passed in as the linking current if a user did not know any better (rather, if the user did know the source code)

dpanici (changes requested):
I think the name could be changed to `LinkingCurrentConsistency` or something, or the docstring should make it clearer that target should be zero, and NOT the targeted linking current G

tests/test_optimizer.py
```diff
@@ -1405,15 +1405,17 @@ def test_optimize_coil_currents(DummyCoilSet):
...
+    constraints = LinkingCurrentConsistency(eq, coils, eq_fixed=True)
     optimizer = Optimizer("lsq-exact")
-    [coils_opt], _ = optimizer.optimize(
-        things=coils,
-        objective=objective,
-        constraints=constraints,
-        verbose=2,
-        copy=True,
-    )
+    with warnings.catch_warnings():
```
ddudt, line 1410: Do we understand why this constraint is being flagged as "incompatible"? This is like the simplest possible linear constraint, so I'm worried if this doesn't work without warnings.

f0uriest: The error is like 1e-9 which is about what we'd expect given the currents are ~MA

desc/objectives/_coils.py
```diff
@@ -1762,6 +1763,201 @@ def compute(self, params_1, params_2=None, constants=None):
         return Psi
 
 
+class LinkingCurrentConsistency(_Objective):
+    """Target the self-consistent poloidal linking current between the plasma and coils.
+
+    A self-consistent coil + plasma configuration must have the sum of the signed
+    currents in the coils that poloidally link the plasma equal to the total poloidal
+    current in the plasma G. This objective computes the difference between these two
```
dpanici, line 1771:
```suggestion
    current required to be linked by the plasma according to the loop integral of its toroidal magnetic field, given by `G`. This objective computes the difference between these two
```

`G` is technically not the current in the plasma but rather the poloidal linking current enclosed (i.e. thru the torus hole) by the plasma, as dictated by Ampere's law, and is a function just of the toroidal magnetic field of the plasma and the boundary geometry for the integral, not the plasma's poloidal current.

dpanici (approved):

PR #1229, opened 2024-08-28
Proxies for neoclassical transport of energetic ions

gretahibbard:
Adding proxies for neoclassical transport from Velasco's "A model for the fast evaluation of prompt losses of
energetic ions in stellarators"  https://doi.org/10.1088/1741-4326/ac2994

- Gamma_c [ eq. 16 ]  
- Gamma_delta [  eq. 19 ] 
- and Gamma_alpha [ eq. 25 ]


YigitElma:
I know this is still a draft and you will probably make many changes but I want to make some kind suggestions,

- You can install pre-commit for automatic checks when you commit such that you don't get comments from github-actions https://github.com/PlasmaControl/DESC/blob/master/CONTRIBUTING.rst#python-styleguide
- On VS Code, there is black formatted plugin which will automatically format a file when you save it, I find this very usefull
- to exclude the files (you use for testing and debugging purposes) from git version control, you can update your `.git/info/exclude` file which is a local .gitignore file (because we don't want to add a bunch of files to the main .gitignore)
- if you run a jupyter notebook on tutorials folder, even if you don't make actual code changes, some device info part of the file will change which cause git to track those changes. I recommend don't commit those changes (in VS Code you can revert those files to the original format 'before committing') because in future it may cause merge conflicts which is very cumbersome to resolve for notebooks. If you already committed the changes, you can copy/paste the files from master branch

These are general suggestions for devs

dpanici:
benchmark with KNOSOS

desc/compute/_neoclassical.py
```diff
@@ -529,3 +534,110 @@ def compute(data):
         / data["<L|r,a>"]
     )
     return data
+
+
+@register_compute_fun(
+    name="Gamma_d Velasco",
+    label=(
+        # Γ_c = π/(8√2) ∫dλ 〈 ∑ⱼ [v τ γ_c²]ⱼ 〉
+        "\\Gamma_c = \\frac{\\pi}{8 \\sqrt{2}} "
```
dpanici, line 543: update math label

dpanici:
There are some changes in the files which are confusing to me, should go through and clean this up as well (for example, the basic equilibrium notebook is deleted, that should not be deleted)

Issue #1221, opened 2024-08-23
Speed up testing by automatically passing if certain files don't change


f0uriest:
We talked about this before and in general it's tricky, but I think we can add a few simple cases:
- if nothing in `desc` or `tests` changes then unit and regression tests aren't needed
- if nothing in `docs/notebooks` or `desc` or `tests` changes then notebook tests aren't needed
- if nothing in `desc` or `tests/benchmarks` changes then benchmarks aren't needed

unalmis:
Might be nice to also have a `run_all_tests` label that we can use to override the heuristics given above

YigitElma:
What if we change dependencies? That requires the whole test to run. I thought about this kind of check before but I think the best we can do is skipping notebook tests if nothing in `docs/notebooks`, `desc` or `requirements` changed. And also we can skip all tests except notebook ones, if the only change is in `docs/notebooks`.

YigitElma:
So, I think the number of PR's that would benefit this kind of change is really low. But I can look into it

YigitElma:
https://github.com/PlasmaControl/DESC/pull/1235#issuecomment-2318810723

PR #1235, opened 2024-08-29
Fix readme badges and expected number of tests

f0uriest:
Some stuff that was missing from #1213 

- Updates readme badges to point to the renamed gh actions
- Tells codecov to expect 14 batches of coverage data, not 10 (we used to have 6 regression + 4 unit, now 6 regression + 8 unit)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -111,12 +111,12 @@ Contribute
     :target: https://desc-docs.readthedocs.io/en/latest/?badge=latest
     :alt: Documentation
 
-.. |UnitTests| image:: https://github.com/PlasmaControl/DESC/actions/workflows/unittest.yml/badge.svg
-    :target: https://github.com/PlasmaControl/DESC/actions/workflows/unittest.yml
+.. |UnitTests| image:: https://github.com/PlasmaControl/DESC/actions/workflows/unit_tests.yml/badge.svg
+    :target: https://github.com/PlasmaControl/DESC/actions/workflows/unit_tests.yml
     :alt: UnitTests
 
-.. |RegressionTests| image:: https://github.com/PlasmaControl/DESC/actions/workflows/regression_test.yml/badge.svg
-    :target: https://github.com/PlasmaControl/DESC/actions/workflows/regression_test.yml
+.. |RegressionTests| image:: https://github.com/PlasmaControl/DESC/actions/workflows/regression_tests.yml/badge.svg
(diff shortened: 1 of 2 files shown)


YigitElma:
Oh sorry, I didn't know these exist.
Can we try to resolve #1221 here?

YigitElma:
I assume it will not affect a lot of PR's but some small ones (like this one or just docs update) can be faster. The conditions can be

- if no change in `desc` or `requirements` don't run `notebook` tests
- if no change in `desc`, `tests (without benchmarks)` or `requirements` don't run `unit` and `regression` tests (also disable code-cov)
- if no change in `desc`, `tests/benchmarks` or `requirements` don't run `benchmarks`

Also, I asked ChatGPT if there is a way to check changes that are made to comments (not to actual code). Here is a basic example,
```
name: Check Code Changes

on: [push, pull_request]

jobs:
  check_code_changes:
    runs-on: ubuntu-latest
    steps:
    - name: Check out the repository
      uses: actions/checkout@v2

    - name: Set up Python
      uses: actions/setup-python@v2
      with:
        python-version: '3.x'

    - name: Install git diff tool (if needed)
      run: sudo apt-get install diffutils

    - name: Check if there are code changes
      run: |
        git diff HEAD^1 -- '*.py' > changes.diff
        if grep -q -v -e '^\s*#' -e '^\s*$' changes.diff; then
          echo "There are code changes!"
          exit 1
        else
          echo "No code changes detected."
          exit 0
        fi
```

f0uriest:
I'd rather fix this for now since it makes the main landing page look bad right now, we can do the testing improvements in another one.

YigitElma (approved):

YigitElma:
Btw, it is not related to #1213 because even before that this was the case, but why do we have extremely high standard deviation in `test_fixed_iter`?

f0uriest:
> Btw, it is not related to #1213 because even before that this was the case, but why do we have extremely high standard deviation in `test_fixed_iter`?

Could be due to the first step being compiled? We used to properly account for that but since #1043 it recompiles a lot less so the later ones are faster?

YigitElma:
> > Btw, it is not related to #1213 because even before that this was the case, but why do we have extremely high standard deviation in `test_fixed_iter`?
> 
> Could be due to the first step being compiled? We used to properly account for that but since #1043 it recompiles a lot less so the later ones are faster?

Ah yes, it looks like the std dev increased after that. Nice

PR #1236, opened 2024-08-30
Single-stage optimization example notebook

IssraAli:
Created a single-stage optimization example notebook.


ddudt:
Noting that this will have to wait for #1222 and possibly also #1205

dpanici:
> Noting that this will have to wait for #1222 and possibly also #1205

IDT it needs to wait for #1205 if it is vacuum, though yea we should wait for both of these if we want a finite beta case also in this

IssraAli:
I wanted to mention, certain combinations of hyperparameters gave very bad outputs, see below for an example. I didn't really systematically document this, but I can do that if you guys are interested (sorry for the bad pic):

d4a6cf41-9f66-479b-a60f-9d74d820680a

The optimizer also converged to designs where the coils were all together in the centre of the torus all tangled up. Perhaps it would be worth it to create a constraint function that ensures that the coils are within a certain "winding volume" padding the torus to prevent these kinds of outputs. 

I generally got much more normal looking coils with the coil-related losses weighted on the order of 100x higher than the QS loss. However, there seems to be quite a lot of field error, despite the fact that i have the ftol for the equilibrium problem all the way down at 5e-9: 

image

dpanici:
@ddudt use APS example

PR #1237, opened 2024-08-31
Fix connected theta edges for plot3d

YigitElma:
Currently, if you don't give a full theta grid to `plot_3d`, it will connect the theta edges and end up this kind of plots,
image

I don't know about you, but I want it to behave like this,
image


YigitElma:
Here is the code to generate this plot,
```python
import plotly.graph_objects as go
from desc.plotting import plot_3d
import numpy as np

fig = go.Figure()
for r in [0.3, 0.6, 1]:
    grid = LinearGrid(
        rho=r,
        theta=np.linspace(0, 2 * np.pi, 200),
        zeta=np.linspace(0, 2 * np.pi/eq.NFP, 200),
        sym=True,
    )
    grid_points = LinearGrid(
            rho=r,
            theta=np.linspace(0, 2 * np.pi, 20),
            zeta=np.linspace(0, 2 * np.pi/eq.NFP, 10),
            NFP=eq.NFP,
            sym=True,
        )
    data = eq.compute(["X", "Y", "Z"], grid=grid_points)
    x = data["X"]
    y = data["Y"]
    z = data["Z"]
    plot_3d(eq, "|B|", grid=grid, fig=fig)
fig.show()

**Review by dpanici — APPROVED** [review]

**Comment by f0uriest** [comment]
Does this still work correctly if the surface is closed?

**Comment by YigitElma** [comment]
> Does this still work correctly if the surface is closed?

Those were also behaving well with this change. I can test again, it was really late when I tested it yesterday.

**Comment by YigitElma** [comment]
> Does this still work correctly if the surface is closed?

Yes, it works, unless you specifically give a grid like this,
```python
eq = get("ATF")
grid = LinearGrid(
        rho=1.,
        theta=np.linspace(0, 2 * np.pi, 200, endpoint=False),
        zeta=np.linspace(0, 2 * np.pi/eq.NFP, 200),
    )
plot_3d(eq, '|B|', grid=grid)
```
image


Because if you don't give a grid, the default grid created by `_get_grid()` function used `endpoint=True` which inherently closes the surface.

f0uriest (approved):

PR #1238, opened 2024-08-31
Fix UTC - ET confusion in cache dependency action

YigitElma:
I originally wanted to have the cache dependency action run at 4.30 am Princeton time, but cron syntax uses UTC which I didn't account for. This is not that important but at 4.30 am there will be fewer test runs and hence fewer workflows to install pip.


dpanici (approved):

f0uriest (approved):

YigitElma:
We can use this PR for validating #1239

YigitElma:
#1239 is working expectedly. Can someone override codecov since unit/regression tests didn't run🙏

ddudt (approved):

PR #1239, opened 2024-09-01
Conditional Workflows

### LINKED ISSUE #1221: Speed up testing by automatically passing if certain files don't change

YigitElma:
Resolves #1221 

I assume it will not affect a lot of PR's but some small ones (like this one or just docs update) can be faster. The conditions are,

- if no change in `desc` or `requirements` don't run `notebook` tests
- if no change in `desc`, `tests (without benchmarks)` or `requirements` don't run `unit` and `regression` tests 
- if no change in `desc`, `tests/benchmarks` or `requirements` don't run `benchmarks`

If there is a change in the workflow file itself, it will run. If unit and regression tests are skipped no coverage report will be uploaded and someone has to override these PRs to merge to master. I have tried uploading 100% fake coverage report but it could have unintentional effects as Rory mentioned.


YigitElma:
https://github.com/PlasmaControl/DESC/pull/1235#issuecomment-2318810723

f0uriest:
If the tests don't run, does github still consider them "passed"? IE, we require the tests to pass before merging, will this cause any issues if they don't run at all?

YigitElma:
> If the tests don't run, does github still consider them "passed"? IE, we require the tests to pass before merging, will this cause any issues if they don't run at all?

This is good question that I don't have the answer of. As a check I can delete the lines that says if the workflow files are changed run the test, for testing that and re add them after. Because basically this pr doesn't need any unit or regression tests.

f0uriest:
hmmm thats kind of annoying. Can you see if you can modify the action in some way to make them pass in that case?

YigitElma:
> hmmm thats kind of annoying. Can you see if you can modify the action in some way to make them pass in that case?

@f0uriest now, it works. The code coverage was failing since nothing was uploaded, I upload a fake 100% coverage data to make that pass (only if actual upload is skipped)

.github/workflows/unit_tests.yml
```diff
@@ -102,3 +124,40 @@ jobs:
...
+                    <lines>
+                      <line number="3" hits="1"/>
+                    </lines>
+                  </class>
+                </classes>
+              </package>
+            </packages>
+          </coverage>' > cov.xml
+
+      - name: Upload 100% Coverage Report if skipped
```
f0uriest, line 150: Any chance this could have some weird unintended consequences?
IE, suppose we merge a PR that only changes some minor thing such that these tests don't run. This then tells codecov that the coverage is 100%. On the next PR where tests do run, we get coverage ~95%, does that mean the 2nd PR fails, even though the actual coverage is fine?

YigitElma: Hmm, I don't know how exactly the codecov works, does it compare to the last PR? 

Tbh, this was also not my first preference but even tho the unit/regression tests pass the codecov was a problem. Can people with override authorization merge PRs before codecov tests start? Like maybe I can change it back to the version without the fake codecov upload and for these PRs we can override and merge?

f0uriest:
possibly relevant: 
- https://community.codecov.com/t/i-would-like-to-skip-the-codecov-when-source-code-when-files-unrelated-to-the-tests-are-modified/4433/8
- https://community.codecov.com/t/how-to-make-codecov-branch-status-checks-conditional/3549

another option might be removing codecov from the list of required checks? And we just manually verify that it's ok when we expect it to be?

YigitElma:
> possibly relevant:
> 
> * https://community.codecov.com/t/i-would-like-to-skip-the-codecov-when-source-code-when-files-unrelated-to-the-tests-are-modified/4433/8
> * https://community.codecov.com/t/how-to-make-codecov-branch-status-checks-conditional/3549
> 
> another option might be removing codecov from the list of required checks? And we just manually verify that it's ok when we expect it to be?

I think the way it is now is fine. It won't upload anything and we will just override it for these kind of PRs which are small in number. The links you sent say this is still under development. I couldn't find a good way of doing this

f0uriest (approved):

ddudt (approved):

PR #1241, opened 2024-09-02
Resolves #1240

### LINKED ISSUE #1240: circular imports with compute utils
Due to how we register things in `data_index`:
```python
# just need to import all the submodules here to register everything in the
     27 # data_index
---> 29 from . import (
     30     _basis_vectors,
     31     _bootstrap,
     32     _core,
     33     _curve,
     34     _equil,
     35     _field,
     36     _geometry,
     37     _metric,
     38     _neoclassical,
     39     _omnigenity,
     40     _profiles,
     41     _stability,
     42     _surface,
     43 )
     44 from .data_index import all_kwargs, allowed_kwargs, data_index
     45 from .geom_utils import rpz2xyz, rpz2xyz_vec, xyz2rpz, xyz2rpz_vec
```
any function defined inside `A = desc/compute`, called inside some directory $B \notin A$, will generate a circular import error whenever something $b \in B$ is called in $A$. This is problematic because we have general utilities inside  `desc/compute` such as all the `safediv`s. We should move anything from desc/compute/utils  that isn't directly related to `data_index` stuff to `desc/utils`.

To reproduce, create a compute function that uses desc/integrals/singularities or bounce integrals which both call `safediv`.

unalmis:


f0uriest (approved):

desc/compute/_basis_vectors.py
```diff
@@ -11,8 +11,8 @@
 
 from desc.backend import jnp
 
+from ..utils import cross, dot, safediv
```
dpanici, line 14: why do we not do `desc.utils` here? or is this equivalent?

unalmis: it's equivalent. this is what Pycharm's automated refactoring chose.

dpanici (approved):

PR #1247, opened 2024-09-11
Add hidden files to upload artifacts step

YigitElma:
`upload-artifact` action made a change 2 weeks ago that excludes hidden files (the ones start with . like /.benchmarks) for upload step. This broke our way of uploading the artifact for benchmarks workflow. As a result, benchmark stats shown as comment to PR had only half of the results.

Maybe another fix with this PR could be waiting for artifact-1 to be uploaded before running `download artifacts` step of group 2. For now it shouldn't cause an issue because group 1 of benchmarks is ~10 minutes shorter than group 2. But rarely, both may not find an artifact to download and the final comment can only have the half of the results

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -136,17 +136,6 @@ jobs:
           pattern: benchmark_artifact_*
           path: tests/benchmarks
 
-      - name: Unzip artifacts if downloaded
-        if: always() && env.has_changes == 'true'
-        run: |
-          cd tests/benchmarks
-          ls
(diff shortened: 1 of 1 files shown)


ddudt (dismissed):

f0uriest (dismissed):

f0uriest (approved):

dpanici (approved):

YigitElma:
@ddudt Can you merge this when you approve? An then #1238 too

ddudt (approved):

PR #1248, opened 2024-09-11
`flip_theta` compatibility function

ddudt:
Redefines $\theta_{new} = \theta_{old} + \pi$ by flipping the sign of all $R$, $Z$, $\lambda$ coefficients with odd $m$. 

This is useful for stuff like omnigenity and TERPSICHORE that assume $\theta=0$ is on the inboard/outboard side.

### DIFF
--- a/desc/compat.py
+++ b/desc/compat.py
@@ -111,6 +111,46 @@ def flip_helicity(eq):
     return eq
 
 
+def flip_theta(eq):
+    """Change the gauge freedom of the poloidal angle of an Equilibrium.
+
+    Equivalent to redefining theta_new = theta_old + π
+
+    Parameters
+    ----------
+    eq : Equilibrium or iterable of Equilibrium
+        Equilibria to flip the helicity of.
+
+    Returns
+    -------
+    eq : Equilibrium or iterable of Equilibrium
+        Same as input, but with the poloidal angle redefined.
+
+    """
+    # maybe it's iterable:
+    if hasattr(eq, "__len__"):
+        for e in eq:
+            flip_theta(e)
+        return eq
+
+    rone = np.ones_like(eq.R_lmn)
+    rone[eq.R_basis.modes[:, 1] % 2 == 1] *= -1
+    eq.R_lmn *= rone
+
+    zone = np.ones_like(eq.Z_lmn)
+    zone[eq.Z_basis.modes[:, 1] % 2 == 1] *= -1
+    eq.Z_lmn *= zone
+
+    lone = np.ones_like(eq.L_lmn)
+    lone[eq.L_basis.modes[:, 1] % 2 == 1] *= -1
+    eq.L_lmn *= lone
+
+    eq.axis = eq.get_axis()
+    eq.surface = eq.get_surface_at(rho=1)
+
+    return eq
+
+
 def rescale(
     eq, L=("R0", None), B=("B0", None), scale_pressure=True, copy=False, verbose=0
 ):
--- a/desc/compat.py
+++ b/desc/compat.py
@@ -119,7 +119,7 @@ def flip_theta(eq):
     Parameters
     ----------
     eq : Equilibrium or iterable of Equilibrium
-        Equilibria to flip the helicity of.
+        Equilibria to redefine the poloidal angle of.
 
     Returns
     -------
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -3,7 +3,7 @@
 import numpy as np
 import pytest
 
-from desc.compat import flip_helicity, rescale
+from desc.compat import flip_helicity, flip_theta, rescale
 from desc.examples import get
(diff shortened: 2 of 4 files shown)


desc/compat.py
```diff
@@ -111,6 +111,46 @@ def flip_helicity(eq):
     return eq
 
 
+def flip_theta(eq):
+    """Change the gauge freedom of the poloidal angle of an Equilibrium.
```
f0uriest, line 115: If we flip this and then re-solve or optimize will it mess with the existing gauge freedom constraint on lambda?

rahulgaur104:
The tests are insufficient for checking is theta flip is correct. We should at least test the field aligned quantities used for GX, ripple, Gamma_c, etc.

YigitElma:
I feel dumb to ask this, but when I plot the flipped eq and original eq, they have the same flux surfaces. Is this because of the `map_coordinates` call inside the plotting routine?
```python
from desc.examples import get
from desc.plotting import plot_comparison
from desc.compat import flip_theta
eq = get("ATF")
eq0 = eq.copy()
eq_flip = flip_theta(eq0)
plot_comparison(eqs=[eq, eq_flip]);
```
image

The compute function works expectedly,
```python
grid = LinearGrid(theta=6)
R_eq = eq.compute("R",grid=grid)["R"]
R_eq_flip = eq_flip.compute("R",grid=grid)["R"]
print(R_eq)
print(R_eq_flip)
```
```
[2.25 2.15 1.95 1.85 1.95 2.15]
[1.85 1.95 2.15 2.25 2.15 1.95]
```

desc/compat.py
ddudt: No it should be fine. The `FixLambdaGauge` constraint enforces $\lambda(\theta=0,\zeta=0)=0$ and this will still be true. All this function does is change $\theta=0$ from the inboard to outboard side (or vice-versa).

ddudt:
> I feel dumb to ask this, but when I plot the flipped eq and original eq, they have the same flux surfaces. Is this because of the `map_coordinates` call inside the plotting routine?

This is the expected behavior, the equilibrium is physically the same and we are only changing the definition of $\theta$ (which the user has the freedom to choose when defining the boundary input). If you plot with an odd number of $\theta$ contours you can see that $\theta=0$ flips from the outboard to inboard side: 

image

ddudt:
> The tests are insufficient for checking is theta flip is correct. We should at least test the field aligned quantities used for GX, ripple, Gamma_c, etc.

This has nothing to do with field-aligned quantities. $\theta$ is the arbitrary computation angle, which the user defines through the boundary surface input. 

For stellarator symmetry, there are 2 unique boundary inputs that give the same physical equilibrium: $\theta=0$ on the outboard or inboard side. This function converts between these two cases. 

This new function is very similar to the existing `flip_helicity` function, so I copied and modified those tests.

desc/compat.py
ddudt: Unless that is what you meant? The two different boundary inputs are physically equivalent, but it might be possible that numerically they converge slightly differently during an equilibrium solve. Conceptually it's similar to positive vs negative Jacobian for the same physical equilibrium.

desc/compat.py
```diff
@@ -111,6 +111,46 @@ def flip_helicity(eq):
...
+
+    Equivalent to redefining theta_new = theta_old + π
+
+    Parameters
+    ----------
+    eq : Equilibrium or iterable of Equilibrium
+        Equilibria to flip the helicity of.
```
dpanici, line 122: docstring should be updated, still says helicity

desc/compat.py
f0uriest: the sign should be the same, so I think it's fine.

tests/test_compat.py
```diff
@@ -135,6 +133,88 @@ def test_flip_helicity_current():
...
+
+    grid = QuadratureGrid(L=eq.L_grid, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP)
+    nodes = grid.nodes.copy()
+    nodes[:, 1] = np.mod(nodes[:, 1] + np.pi, 2 * np.pi)
+    grid_flip = Grid(nodes)  # grid with flipped theta values
+    data_keys = ["current", "|F|", "D_Mercier", "f_C"]
+
+    data_old = eq.compute(data_keys, grid=grid, helicity=(1, eq.NFP))
+    eq = flip_theta(eq)
+    data_new = eq.compute(data_keys, grid=grid, helicity=(1, eq.NFP))
```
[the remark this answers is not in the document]
ddudt: Thanks for catching this, I can get rid of the second computation. I had copied this test from `flip_helicity` and in that case we compute it twice with different signs for the helicities.

f0uriest (approved):

dpanici (approved):

PR #1119, opened 2024-07-10
fourier bounce

unalmis:
Resolves #1045. Uses pseudospectral technique.

## Summary
This solves the dominant cost of optimization objectives relying on the methods used in https://github.com/PlasmaControl/DESC/pull/854: interpolating DESC's 3D transforms to an optimization-step dependent grid that is dense enough for function approximation with local splines. This is in some sense referred to as off-grid interpolation in literature; it is often a bottleneck.

- The function approximation done here requires DESC transforms on a fixed grid with typical resolution, using FFTs to compute the map $\alpha, \zeta \mapsto \theta(\alpha, \zeta)$ between coordinate systems. This enables evaluating functions along field lines without root-finding. 
- The faster convergence of spectral interpolation requires a less dense grid to interpolate onto from DESC's 3D transforms.
- Spectral function approximation is more accurate than cubic splines.
- 2D interpolation enables tracing the field line for more toroidal transits.
- With these improvements, optimization with objectives on #1290 now take 6gb of memory whereas before they may not have fit onto today's gpu.

## Future performance improvements
- #1154 
- #1294 
- #1303 
- #1206
- #1034

PR #1254, opened 2024-09-18
Stop modding of non-periodic angles

### LINKED ISSUE #1173: Mod SFL theta by 2pi inside of `vmec.VMECIO.compute_theta_coords`
_Originally posted by @dpanici in https://github.com/PlasmaControl/DESC/pull/1153#discussion_r1712110839_

### LINKED ISSUE #1180: don't mod angles; they are not periodic

unalmis:
Resolves #1180 and resolves #1173 .

Here is what $\theta$ looks like now on #1119 :
theta

Likewise, all the other quantities have become smooth bumps, as desired.

### DIFF
--- a/desc/compute/_core.py
+++ b/desc/compute/_core.py
@@ -1494,7 +1494,7 @@ def _Z_zzz(params, transforms, profiles, data, **kwargs):
     label="\\alpha",
     units="~",
     units_long="None",
-    description="Field line label, defined on [0, 2pi)",
+    description="Field line label",
     dim=1,
     params=[],
     transforms={},
@@ -1503,7 +1503,7 @@ def _Z_zzz(params, transforms, profiles, data, **kwargs):
     data=["theta_PEST", "phi", "iota"],
 )
 def _alpha(params, transforms, profiles, data, **kwargs):
-    data["alpha"] = (data["theta_PEST"] - data["iota"] * data["phi"]) % (2 * jnp.pi)
+    data["alpha"] = data["theta_PEST"] - data["iota"] * data["phi"]
     return data
 
 
@@ -3077,7 +3077,7 @@ def _theta(params, transforms, profiles, data, **kwargs):
     data=["theta", "lambda"],
 )
 def _theta_PEST(params, transforms, profiles, data, **kwargs):
-    data["theta_PEST"] = (data["theta"] + data["lambda"]) % (2 * jnp.pi)
+    data["theta_PEST"] = data["theta"] + data["lambda"]
     return data
 
 
--- a/desc/equilibrium/coords.py
+++ b/desc/equilibrium/coords.py
@@ -327,8 +327,6 @@ def _map_PEST_coordinates(
 
     """
     rho, theta_PEST, zeta = coords.T
-    # Assume λ=0 for initial guess.
-    guess = setdefault(guess, theta_PEST)
 
     # Root finding for θₖ such that r(θₖ) = ϑₖ(ρ, θₖ, ζ) − ϑ = 0.
     def rootfun(theta_DESC, theta_PEST, rho, zeta):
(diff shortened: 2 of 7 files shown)


f0uriest (approved):

rahulgaur104:
>Here is what θ looks like now on https://github.com/PlasmaControl/DESC/pull/1119 

Does it work for negative zeta values? What if I give zeta in [-4pi, 4pi] to map_coordinates. Will the theta still be smooth?

unalmis:
> > Here is what θ looks like now on #1119
> 
> Does it work for negative zeta values? What if I give zeta in [-4pi, 4pi] to map_coordinates. Will the theta still be smooth?

Where we start $\zeta$ seems to not matter. However there is still an issue, I believe completely separate to this PR. I have found that $\theta$ is still discontinuous every two toroidal transits, which is causing issues: https://github.com/PlasmaControl/DESC/pull/1119#discussion_r1766062255. 

Here is a plot of $\theta(\alpha=0, \zeta)$. 

tt

rahulgaur104 (approved):
Approving, but I am slightly concerned that it might break the ballooning PR. 
Merge at will.

unalmis:
For the equilibrium I computed this plot on, two toroidal transits corresponds to one poloidal transit. So it seems the information of which poloidal transit $\theta$ is at is lost somewhere

unalmis:
So the issue is probably that I'm paramterizing my FourierChebyshev series for $\theta$ using $\alpha \in [-\pi, \pi]$ and I should be using $\alpha \in [0, 2 \pi]$. I'll look into this on the weekend.

unalmis:
I resolved the remaining discontinuous issue fyi.

PR #1258, opened 2024-09-21
Add docstring inheriter

### LINKED ISSUE #879: Simplify `Objective` class args and docstrings
We have an ever growing number of objectives, with pretty much all of the init methods being something like this
```python
def __init__(self, thing, 
target, 
bounds, 
weight, 
normalize, 
... (other args generic to all objectives), 
arg1_specific_to_this_objective, 
arg2_specific_to_this_objective
):
```

And each of the args is documented separately in each class. This means that if we want to add a new property generic to all objectives (like we did with `loss_function`, `normalize`, `deriv_mode`, etc) we have to add it to ALL of the init methods and docstrings

A cleaner method that requires less duplication could be something like 

```python
def __init__(self, 
thing, 
arg1_specific_to_this_objective, 
arg2_specific_to_this_objective, 
**kwargs_generic_to_all_objectives
):
```

Doing this in combination with something like https://github.com/AntoineD/docstring-inheritance would allow us to reduce the amount of redundant docstrings in the code while still having all the args documented to the user in an IDE or docs page. We could also get fancy and override the subclass `__signature__` attribute to include the extra args from the parent class.

YigitElma:
Most of the documentation of our objectives has the same parameters that are inherited from the main `_Objective` class. This PR removes the repeated docstring from each objective and updates the docstring by inheriting which reduces the lines of code and also facilitates the maintenance. For example, when we add `jac_chunk_size` in #1052, we have to copy-paste the docs to every single objective which is tedious.

Introduces `collect_docs` function that creates docstring for common parameters and with option to overwrite user can give a custom definition for a parameter without changing the order of the docs 

Resolves #879


f0uriest:
As part of this would also be good to make most of the "generic" arguments keyword only. Like

```python
def __init__(self, eq, specific_arg1, *, target=None, bounds=None, other_args_shared_by_all=default, ...):
```

That way we can add more and reorder them in the future without breaking things

f0uriest:
Here's a sketch of what I was talking about at the meeting:

```python
target_doc = """
    target : {float, ndarray}, optional
        Target value(s) of the objective. Only used if bounds is None.
        Must be broadcastable to Objective.dim_f.
"""
bounds_doc = """
    bounds : tuple of {float, ndarray}, optional
        Lower and upper bounds on the objective. Overrides target.
        Both bounds must be broadcastable to to Objective.dim_f
"""
weight_doc = """
    weight : {float, ndarray}, optional
        Weighting to apply to the Objective, relative to other Objectives.
        Must be broadcastable to to Objective.dim_f
"""

callable_target_doc = """
    target : {float, ndarray, callable}, optional
        Target value(s) of the objective. Only used if bounds is None.
        Must be broadcastable to Objective.dim_f. If a callable, should take a
        single argument `rho` and return the desired value of the profile at those
        locations. Defaults to ``target=0``.
"""
callable_bounds_doc = """
    bounds : tuple of {float, ndarray, callable}, optional
        Lower and upper bounds on the objective. Overrides target.
        Both bounds must be broadcastable to to Objective.dim_f
        If a callable, each should take a single argument `rho` and return the
        desired bound (lower or upper) of the profile at those locations.
        Defaults to ``target=0``.
"""

def _unused(doc):
    return doc.rstrip() + " Unused by this objective.\n"


default_doc = target_doc + bounds_doc + weight_doc
profile_doc = callable_target_doc + callable_bounds_doc + weight_doc
dimensionless_doc = target_doc + bounds_doc + _unused(normalize_doc) + _unused(normalize_target_doc)


class MyObjective(_Objective):
    """Do some stuff.

    Parameters
    ----------
    arg1 : type
        arg specific to this objective
    """ + default_doc


class MyProfileObjective(_Objective):
    """Do some stuff.

    Parameters
    ----------
    arg1 : type
        arg specific to this objective
    """ + profile_doc

class DimensionlessObjective(_Objective)
    """Do some stuff.

    Parameters
    ----------
    arg1 : type
        arg specific to this objective
    """ + dimensionless_doc

```

This could also maybe done automatically via inspection or something with a util function?
One other thing to be careful of is the defaults for target/bounds which are currently in the docstring but may be unique to each objective.

YigitElma:
@f0uriest I implemented something similar to what you suggested. If you are fine with it I will make the remaining changes

YigitElma:
Do we want to use something similar for optimizers and methods of `ObjectiveFunction`? Those are also the same for parent class and repeated multiple times.

dpanici:
- have `collect_docs` take in `target_str` etc to replace the target string in the same spot in the docstring
- specific stuff can be at the top, so docstring order might not match init order

desc/objectives/_bootstrap.py
```diff
@@ -31,31 +31,6 @@ class BootstrapRedlConsistency(_Objective):
     ----------
     eq : Equilibrium
         Equilibrium that will be optimized to satisfy the Objective.
-    target : {float, ndarray}, optional
-        Target value(s) of the objective. Only used if bounds is None.
-        Must be broadcastable to Objective.dim_f. Defaults to ``target=0``.
```
f0uriest, line 36: need to override this and bounds to include default target. (it might be better to add default target/bounds as args to collect_docs since each objective will have to override this)

desc/objectives/_coils.py
```diff
@@ -29,50 +29,23 @@ class _CoilObjective(_Objective):
...
-        will require (with some baseline memory usage). The time to compute the
-        Jacobian is roughly ``t=t0 +t1/jac_chunk_size``, so the larger the
-        ``jac_chunk_size``, the faster the calculation takes, at the cost of
-        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
-        memory intensive, but slowest method of calculating the Jacobian.
-        If None, it will use the largest size i.e ``obj.dim_x``.
 
     """
 
+    overwrite = {
```
f0uriest, line 38: make this a `_private_attribute`

desc/objectives/_geometry.py
```diff
@@ -21,52 +21,15 @@ class AspectRatio(_Objective):
...
-    bounds : tuple of {float, ndarray}, optional
-        Lower and upper bounds on the objective. Overrides target.
-        Both bounds must be broadcastable to to Objective.dim_f.
-        Defaults to ``target=2``.
-    weight : {float, ndarray}, optional
-        Weighting to apply to the Objective, relative to other Objectives.
-        Must be broadcastable to to Objective.dim_f
-    normalize : bool, optional
-        Whether to compute the error in physical units or non-dimensionalize.
-        Has no effect for this objective.
```
f0uriest, line 36: need to override `normalize` and `normalize_target` to include `Has no effect for this objective` (this is another common one that might be easiest with a flag passed to `collect_docs`)

f0uriest (changes requested):
Main thing is we need overrides in a few more places to include the default target/bounds and noting when normalize doesn't do anything.

Also it looks like some of the newer objectives don't include docs on the default target/bounds, this would be a good chance to fix that. The convention we've been using is to state the default under both target and bounds and make clear which it is.

ie

```
target : {float, array}
    ...
    Default is target=0.
bounds : tuple(float, array)
    ...
    Default is target=0
```

or
```
target : {float, array}
    ...
    Default is bounds=(0,1).
bounds : tuple(float, array)
    ...
    Default is bounds=(0, 1).
```

desc/objectives/_geometry.py
f0uriest: need this in a few other places too

desc/objectives/_bootstrap.py
f0uriest: need this in several other places too

desc/objectives/_coils.py
```diff
@@ -780,52 +610,15 @@ class CoilSetMinDistance(_Objective):
...
-        will require (with some baseline memory usage). The time to compute the
-        Jacobian is roughly ``t=t0 +t1/jac_chunk_size``, so the larger the
-        ``jac_chunk_size``, the faster the calculation takes, at the cost of
-        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
-        memory intensive, but slowest method of calculating the Jacobian.
-        If None, it will use the largest size i.e ``obj.dim_x``.
 
     """
 
+    __doc__ = __doc__.rstrip() + collect_docs(coil=True)
```
f0uriest, line 620: default bounds for this are `(1, np.inf)`

desc/objectives/_coils.py
```diff
@@ -997,22 +764,11 @@ class PlasmaCoilSetMinDistance(_Objective):
...
```
f0uriest, line 770: default bounds = `(1, np.inf)`

desc/objectives/_coils.py
```diff
@@ -1240,22 +980,15 @@ class QuadraticFlux(_Objective):
...
-        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
-        memory intensive, but slowest method of calculating the Jacobian.
-        If None, it will use the largest size i.e ``obj.dim_x``.
 
     """
 
+    __doc__ = __doc__.rstrip() + collect_docs(
+        target_default="``target=0``.",
+        bounds_default="``target=0``.",
+        loss_detail=" Note: has no effect for this objective.",
```
f0uriest, line 989: @dpanici why does this not allow for a loss function?

desc/objectives/_coils.py
```diff
@@ -1463,23 +1172,13 @@ class ToroidalFlux(_Objective):
...
-        ``jac_chunk_size``, the faster the calculation takes, at the cost of
-        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
-        memory intensive, but slowest method of calculating the Jacobian.
-        If None, it will use the largest size i.e ``obj.dim_x``.
-
 
     """
 
+    __doc__ = __doc__.rstrip() + collect_docs(
+        loss_detail=" Note: has no effect for this objective."
```
f0uriest, line 1179: default target is `eq.Psi`

desc/objectives/_geometry.py
```diff
@@ -196,52 +165,21 @@ class Elongation(_Objective):
...
-        If None, it will use the largest size i.e ``obj.dim_x``.
 
     """
 
+    __doc__ = __doc__.rstrip() + collect_docs(
+        target_default="``target=1``.",
+        bounds_default="``target=1``.",
+        normalize_detail=" Note: Has no effect for this objective.",
+        normalize_target_detail=" Note: Has no effect for this objective.",
+        loss_detail=" Note: Has no effect for this objective.",
```
f0uriest, line 180: I think this was incorrect before, the loss function should be allowed? (it defaults to max?)

desc/objectives/_power_balance.py
```diff
@@ -24,52 +24,18 @@ class FusionPower(_Objective):
...
-        Jacobian is roughly ``t=t0 +t1/jac_chunk_size``, so the larger the
-        ``jac_chunk_size``, the faster the calculation takes, at the cost of
-        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
-        memory intensive, but slowest method of calculating the Jacobian.
-        If None, it will use the largest size i.e ``obj.dim_x``.
 
     """
 
+    __doc__ = __doc__.rstrip() + collect_docs(
+        target_default="``target=1e9``.", bounds_default="``target=1e9``."
```
f0uriest, line 36: also needs loss function unused

desc/objectives/_power_balance.py
```diff
@@ -220,54 +186,20 @@ class HeatingPowerISS04(_Objective):
...
-        Jacobian is roughly ``t=t0 +t1/jac_chunk_size``, so the larger the
-        ``jac_chunk_size``, the faster the calculation takes, at the cost of
-        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
-        memory intensive, but slowest method of calculating the Jacobian.
-        If None, it will use the largest size i.e ``obj.dim_x``.
 
     """
 
+    __doc__ = __doc__.rstrip() + collect_docs(
+        target_default="``target=0``.", bounds_default="``target=0``."
```
f0uriest, line 200: also needs loss function unused

desc/objectives/_profiles.py
```diff
@@ -188,56 +166,16 @@ class RotationalTransform(_Objective):
...
-        will require (with some baseline memory usage). The time to compute the
-        Jacobian is roughly ``t=t0 +t1/jac_chunk_size``, so the larger the
-        ``jac_chunk_size``, the faster the calculation takes, at the cost of
-        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
-        memory intensive, but slowest method of calculating the Jacobian.
-        If None, it will use the largest size i.e ``obj.dim_x``.
 
     """
 
+    __doc__ = __doc__.rstrip() + collect_docs(overwrite=profile_overwrite)
```
f0uriest, line 177: normalize and normalize target dont do anything for this

desc/objectives/_profiles.py
```diff
@@ -372,56 +310,16 @@ class Shear(_Objective):
...
```
f0uriest, line 321: normalize and normalize target dont do anything

desc/objectives/_stability.py
```diff
@@ -222,56 +200,16 @@ class MagneticWell(_Objective):
...
-        will require (with some baseline memory usage). The time to compute the
-        Jacobian is roughly ``t=t0 +t1/jac_chunk_size``, so the larger the
-        ``jac_chunk_size``, the faster the calculation takes, at the cost of
-        requiring more memory. A ``jac_chunk_size`` of 1 corresponds to the least
-        memory intensive, but slowest method of calculating the Jacobian.
-        If None, it will use the largest size i.e ``obj.dim_x``.
 
     """
 
+    __doc__ = __doc__.rstrip() + collect_docs(overwrite=overwrite_stability)
```
f0uriest, line 211: normalize and normalize target dont do anything

desc/objectives/objective_funs.py
```diff
@@ -31,6 +31,153 @@
...
 
+doc_target = """
+    target : {float, ndarray}, optional
+        Target value(s) of the objective. Only used if bounds is None.
+        Must be broadcastable to Objective.dim_f.
+"""
+doc_bounds = """
+    bounds : tuple of {float, ndarray}, optional
+        Lower and upper bounds on the objective. Overrides target.
+        Both bounds must be broadcastable to to Objective.dim_f
```
unalmis, line 42:
```suggestion
        Both bounds must be broadcastable to Objective.dim_f

File: desc/objectives/objective_funs.py [hunk 15]
```diff
@@ -31,6 +31,153 @@
...
+"""
+doc_weight = """
+    weight : {float, ndarray}, optional
+        Weighting to apply to the Objective, relative to other Objectives.
+        Must be broadcastable to to Objective.dim_f
```
unalmis, line 47:
```suggestion
        Must be broadcastable to Objective.dim_f

File: desc/objectives/objective_funs.py [hunk 16]
```diff
@@ -31,6 +31,153 @@
...
+"""
+doc_loss_function = """
+    loss_function : {None, 'mean', 'min', 'max'}, optional
+        Loss function to apply to the objective values once computed. This loss function
+        is called on the raw compute value, before any shifting, scaling, or
+        normalization.
+"""
+doc_deriv_mode = """
+    deriv_mode : {"auto", "fwd", "rev"}
+        Specify how to compute jacobian matrix, either forward mode or reverse mode AD.
```
unalmis, line 67:
```suggestion
        Specify how to compute Jacobian matrix, either forward mode or reverse mode AD.

f0uriest (approved):
Approving, but check with @dpanici about the one that says loss function doesn't do anything

desc/objectives/_coils.py
dpanici: Remove the loss detail

dpanici: @YigitElma

dpanici (changes requested):
should make relevant changes and mention what we are doing with docstrings in the [Adding objectives rst doc file](https://desc-docs.readthedocs.io/en/latest/adding_objectives.html)

f0uriest (approved):

dpanici (approved):

PR #1259, opened 2024-09-22
Ku/fourier bounce part1

### LINKED ISSUE #1260: test_compute_everything fails
due to recent changes in jax as discussed in slack

unalmis:
- [x] Moves some stuff added in `ripple` and `ku/fourier_bounce` branches here to prevent those branches from diverging.
- [x] Resolves #1260


dpanici:
Clever of you to put the fix to #1260 here so I would have to review the rest of the PR as well

desc/integrals/bounce_integral.py
```diff
@@ -289,11 +292,11 @@ def check_points(self, z1, z2, pitch_inv, *, plot=True, **kwargs):
     def integrate(
         self,
         integrand,
+        points,
```
f0uriest, line 295: should this be an optional argument, with the default to calculate from `pitch_inv`?

desc/integrals/bounce_integral.py
```diff
@@ -311,33 +314,28 @@ def integrate(
             accept the arrays in ``f`` as arguments as well as the additional keyword
             arguments: ``B`` and ``pitch``. A quadrature will be performed to
             approximate the bounce integral of ``integrand(*f,B=B,pitch=pitch)``.
+        points : tuple[jnp.ndarray]
```
f0uriest, line 317: should reference function to compute these

desc/integrals/bounce_utils.py
```diff
@@ -27,36 +27,34 @@
...
+        1/λ values and weights.
 
     """
-    # Floating point error impedes consistent detection of bounce points riding
-    # extrema. Shift values slightly to resolve this issue.
-    min_B = (1 + relative_shift) * min_B
-    max_B = (1 - relative_shift) * max_B
+    errorif(
+        num_pitch > 1e5,
+        msg=f"{num_pitch} > 1e5 not possible due to floating point errors.",
```
f0uriest, line 51: can you explain this a bit? Where do the errors pop up?

desc/integrals/bounce_utils.py
```diff
@@ -27,36 +27,34 @@
...
+    )
     # Samples should be uniformly spaced in |B| and not λ (GitHub issue #1228).
-    pitch_inv = jnp.moveaxis(composite_linspace(jnp.stack([min_B, max_B]), num), 0, -1)
-    assert pitch_inv.shape == (*min_B.shape, num + 2)
-    return pitch_inv
+    x, weight = chebgauss_uniform(num_pitch)
```
f0uriest, line 54: isnt the end result the same as before, with linear spacing between 1/bmin 1/bmax?

desc/integrals/bounce_utils.py
unalmis: This replaces the relative shift logic above. if a given pitch_inv value is too close to global min and max B, then floating point errors will make it impossible to detect whether a given pitch horizontal line intersects B.

unalmis: i’ll add back the comment

desc/integrals/bounce_utils.py
unalmis: Yes; The difference is that it’s an open quadrature, which is better because gammac is infinite at max|B| so you don’t want a closed quadrature

unalmis: linear spacing between bmin and bmax, not pitch

desc/integrals/quad_utils.py
```diff
@@ -190,14 +184,67 @@ def leggauss_lob(deg, interior_only=False):
     return x, w
 
 
+def chebgauss_uniform(deg):
+    """Gauss-Chebyshev quadrature with uniformly spaced nodes.
```
f0uriest, line 188: Not sure I totally understand this. The points are linearly spaced, and the weights are uniform, so isn't this just 1st order reimann sum?

desc/integrals/quad_utils.py
```diff
@@ -190,14 +184,67 @@ def leggauss_lob(deg, interior_only=False):
...
+    x, w : (jnp.ndarray, jnp.ndarray)
+        Shape (deg, ).
+        Quadrature points and weights.
+
+    """
+    # Adapted from
+    # github.com/scipy/scipy/blob/v1.14.1/scipy/special/_orthogonal.py#L1803-L1851.
+    t = jnp.arange(deg, 0, -1) * jnp.pi / (deg + 1)
+    x = jnp.cos(t)
+    w = jnp.pi * jnp.abs(jnp.sin(t)) / (deg + 1)
```
f0uriest, line 236: shouldn't this be sin^2(t)

unalmis: abs(sin(t)) = sqrt(1-x^2). This is useful for quadrature where that weight factor is not separable from the function to integrate. in our case f = g sqrt(1-lambda B)

desc/integrals/bounce_integral.py
unalmis: I'd rather user just call bounce.points()

f0uriest: The reason for this is just to avoid duplicate calculations right?

I figure there are 2 main use cases for this:
1. in some compute or objective function where the user doesn't interact directly, and we as developers will know to precompute the bounce points and pass them in
2. Used directly by a user to just compute some arbitary bounce integral. In that case I'm not a fan of having multiple arguments to a function that have to be self-consistent (and where checking self-consistency is non-trivial), so in that case I think it makes more sense to allow the user to just specify the pitch values

tests/test_compute_everything.py
```diff
@@ -35,8 +35,8 @@ def _compare_against_master(
                 np.testing.assert_allclose(
                     actual=data[p][name],
                     desired=master_data[p][name],
-                    atol=1e-10,
-                    rtol=1e-10,
+                    atol=1e-9,
```
dpanici, line 38: maybe try making atol to be `1e-12 * np.mean(np.abs(data[p][name]))` and then we can keep rtol tighter

desc/integrals/quad_utils.py
unalmis: yes assuming the riemann sum starts and ends at the correct endpoints, it is the same as gaussian quadrature in a transformed variable y

desc/integrals/bounce_integral.py
```diff
@@ -47,27 +47,25 @@ class Bounce1D(IOAble):
...
     The function approximation in ``Bounce1D`` is ignorant that the objects to
     approximate are defined on a bounded subset of ℝ². Instead, the domain is
     projected to ℝ, where information sampled about the function at infinity
     cannot support reconstruction of the function near the origin. As the
     functions of interest do not vanish at infinity, pseudo-spectral techniques
     are not used. Instead, function approximation is done with local splines.
-    This is useful if one can efficiently obtain data along field lines and
-    most efficient if the number of toroidal transits to follow a field line is
-    not too large.
+    This is useful if one can efficiently obtain data along field lines the
```
dpanici, line 67: *And the number

desc/integrals/bounce_integral.py
```diff
@@ -142,9 +146,8 @@ def __init__(
             Whether the arrays in ``data`` are already reshaped to the expected form of
             shape (..., N) or (..., L, N) or (M, L, N). This option can be used to
             iteratively compute bounce integrals one field line or one flux surface
-            at a time, respectively, potentially reducing memory usage. To do so,
-            set to true and provide only those axes of the reshaped data.
-            Default is false.
+            at a time, respectively, reducing memory usage. To do so, set to true and
+            provide only those axes of the reshaped data. Default is false.
```
dpanici, line 150: *False

desc/integrals/bounce_integral.py
```diff
@@ -250,18 +253,18 @@ def points(self, pitch_inv, *, num_well=None):
...
 
-    def check_points(self, z1, z2, pitch_inv, *, plot=True, **kwargs):
+    def check_points(self, points, pitch_inv, *, plot=True, **kwargs):
         """Check that bounce points are computed correctly.
 
         Parameters
         ----------
-        z1, z2 : (jnp.ndarray, jnp.ndarray)
-            Shape (M, L, P, num_well).
+        points : tuple[jnp.ndarray]
```
dpanici, line 261: should you specify that the tuple is length 2? since that is expected?

tests/test_compute_everything.py
unalmis: Ok I changed, but if this fails on CI due to floating point errors, it's your job to reduce to whatever makes is pass on your mac

unalmis: this also probably won't work on a different machine because if the latter is small, then atol gets smaller. small implies close to zero, so in that regiem rtol is not helpful

dpanici (approved):

tests/test_compute_everything.py
dpanici: fair point, I made a change that I think fixes this

dpanici (dismissed):

desc/integrals/quad_utils.py
unalmis: i am saying that the extra sin factor is buried inside the function to integrate

desc/integrals/bounce_integral.py
unalmis: I have moved `points` to the final allowable nameless argument so that users can use the function as if the parameter wasn't there

dpanici (dismissed):

desc/integrals/quad_utils.py
unalmis: There's a longer comment on the ripple branch, but put another way simply, all quadrature tricks can be stated in the most general sense to make your function as smooth as possible before integrating. This ensures that a sample of the integrand at one point is the best approximation to stuff around it. Hence equivalently you can increase observed smoothness by putting more nodes in non-smooth areas of the function.

Secondary to the smoothness objective is the objective to minimize interpolation errors that result from weaknesses in the choice of basis functions. A Gaussian quadrature in x = 1/pitch sacrifices the first goal, for the secondary goal: minimizing runge. If the quantity to integrate is sufficiently smooth to begin with, then this choice is fine. Neither Gamma_c nor epsilon effective were sufficiently smooth for this. For the former it's easy to see from the expression, for the latter I can only claim that's what i see empirically

desc/integrals/quad_utils.py
YigitElma: To be honest, I would prefer the proper definition of the quadrature weights instead of yours which depends on the function to integrate. Or at least state the specific use case of this utility function in docstring such that in the future we don't forget why it is how it is.
What I mean is someone may misuse this function without knowing your implementation case.

desc/integrals/quad_utils.py
```diff
@@ -190,14 +184,67 @@ def leggauss_lob(deg, interior_only=False):
...
+    x, w : (jnp.ndarray, jnp.ndarray)
+        Shape (deg, ).
+        Quadrature points and weights.
+
+    """
+    # Define x = 2/π arcsin y and g : y ↦ f(x(y)).
+    #   ∫₋₁¹ f(x) dx = 2/π ∫₋₁¹ (1−y²)⁻⁰ᐧ⁵ g(y) dy
+    # ∑ₖ wₖ f(x(yₖ)) = 2/π ∑ₖ ωₖ g(yₖ)
+    # Given roots yₖ of Chebyshev polynomial, x(yₖ) is uniform in (-1, 1).
+    x = jnp.arange(-deg + 1, deg + 1, 2) / deg
```
YigitElma, line 209: I am also confused as Rory, in the docstring, you mention Gauss-Chabyshev, but the function itself has nothing to do with that. Probably the way you use it in your implementation is related to that but if this is going to be a general integral utility function, I would prefer function docstrings to explain what actually this function is doing.

unalmis: Ok thanks, would ""Gauss-Chebyshev quadrature in transformed variable?" be better

These are the x(y_k) where y_k are Chebyshev roots. It performs a Gauss-Chebyshev quadrature in the transformed variable y. I want to mention this somewhere because uniform spacing is still ambigous as to how the endpoints are chosen (i.e. midpoints spacing, include endpoints) and the weights associated to points (i.e. in some cases points near boundaries are weighted less). Specifying it is a Gauss-Chebyshev quadrature makes that explicit.

unalmis: I am also fine with making this method private

unalmis:
> To be honest, I would prefer the proper definition of the quadrature weights instead of yours which depends on the function to integrate. Or at least state the specific use case of this utility function in docstring such that in the future we don't forget why it is how it is.
What I mean is someone may misuse this function without knowing your implementation case.

The docstring mentions
> Returns quadrature points xₖ and weights wₖ for the approximate evaluation
    of the integral ∫₋₁¹ f(x) dx ≈ ∑ₖ wₖ f(xₖ) where f(x) = g(x) √(1−x²).

These are the proper definition of the quadrature weights. Note that this quadrature is different from  ∫₋₁¹ √(1−x²) f(x) dx ≈ ∑ₖ wₖ f(xₖ) where f(x) = g(x). The latter would fail for bounce integrals. I can change the name of the function to your preference, but I cannot change the definition of the weights

desc/integrals/quad_utils.py
unalmis: The docstring mentions
> Returns quadrature points xₖ and weights wₖ for the approximate evaluation
    of the integral ∫₋₁¹ f(x) dx ≈ ∑ₖ wₖ f(xₖ) where f(x) = g(x) √(1−x²).

Note that this quadrature is different from  ∫₋₁¹ √(1−x²) f(x) dx ≈ ∑ₖ wₖ f(xₖ) where f(x) = g(x). The latter would fail for bounce integrals. I can change the name of the function to your preference, but I cannot change the definition of the weights

desc/integrals/quad_utils.py
YigitElma: Yeah, "Gauss-Chebyshev quadrature in transformed variable" is better to distinguish it from the actual Gauss-Chebyshev quadrature. Making it private could also work, since this is not a stand-alone quadrature function but helps to create one

desc/integrals/quad_utils.py
YigitElma: Ah ok, I see now.

> Note that this quadrature is different from ∫₋₁¹ √(1−x²) f(x) dx ≈ ∑ₖ wₖ f(xₖ) where f(x) = g(x).

I confused this with the original definition. My bad.

YigitElma (approved):

dpanici (approved):

unalmis:
feel free to merge whenever; codecov might block me

Issue #19, opened 2020-09-27
allow for ansi indexing


f0uriest:
I added an option in the get_zern_basis_idx_dense function for either fringe or ansi indexing, but the symmetry constraint matrix still assumes fringe indexing. Can we change that to just use the zern_idx array instead of M,N? Also we'll need to add some sort of "indexing" option to the input file and parser

PR #1264, opened 2024-09-24
Add Quadratic flux minimizing surface functionality

### LINKED ISSUE #1162: Quadratic flux minimizing surfaces
A few people also asked if this was possible. I think it could be but we'd need to modify the QuadraticFlux objective to allow the surface to vary? Or make a new objective for it?

dpanici:
- Add ``SurfaceQuadraticFlux`` objective which minimizes the quadratic magnetic flux through a ``FourierRZToroidalSurface`` object, allowing for optimizing for Quadratic flux minimizing (QFM) surfaces.
- Allow ``ToroidalFlux`` objective to accept ``FourierRZToroidalSurface`` so it can be used to specify the toroidal flux through a QFM surface.
- Add ``eq_fixed`` flag to ``ToroidalFlux`` to allow for the equilibrium/QFM surface to vary during optimization, useful for single-stage optimizations.
- Add tutorial notebook showcasing QFM surface capability.

Resolve #1162


desc/objectives/_coils.py
```diff
@@ -1110,11 +1110,17 @@ class QuadraticFlux(_Objective):
...
+    case a `FourierRZToroidalSurface` should be passed instead of an Equilibrium,
+    and `qfm_surface=True` should be passed into the objective.
+
     Parameters
     ----------
-    eq : Equilibrium
-        Equilibrium upon whose surface the normal field error will be minimized.
-        The equilibrium is kept fixed during the optimization with this objective.
+    eq : Equilibrium or FourierRZToroidalSurface
+        Equilibrium (or QFM surface) upon whose surface the normal field error
```
f0uriest, line 1120: should maybe add a note to use BoundaryError or VacuumBoundaryError for free boundary / single stage optimization

desc/objectives/_coils.py
```diff
@@ -1182,8 +1198,23 @@ def __init__(
         self._eq = eq
         self._field = field
         self._field_grid = field_grid
-        self._vacuum = vacuum
-        things = [field]
+        self._vacuum = vacuum or qfm_surface
+        self._qfm_surface = qfm_surface
+        errorif(
+            qfm_surface and hasattr(eq, "L_lmn"),
```
f0uriest, line 1204: would prefer an actual instance check here instead of relying on attributes to be/not be there

desc/objectives/_coils.py
```diff
@@ -1182,8 +1198,23 @@ def __init__(
         self._eq = eq
         self._field = field
         self._field_grid = field_grid
-        self._vacuum = vacuum
-        things = [field]
+        self._vacuum = vacuum or qfm_surface
```
f0uriest, line 1201: might be better to throw an error here if they're incompatible?

desc/objectives/_coils.py
```diff
@@ -1282,14 +1323,25 @@ def compute(self, field_params, constants=None):
         """
         if constants is None:
             constants = self.constants
-
+        field_params = params_1 if not self._qfm_surface else params_2
```
f0uriest, line 1326: don't you also need to check if the field is fixed? in which case the only DOFs will be the surface?

desc/objectives/_coils.py
```diff
@@ -1282,14 +1323,25 @@ def compute(self, field_params, constants=None):
...
+        surf_params = params_2 if not self._qfm_surface else params_1
         # B_plasma from equilibrium precomputed
-        eval_data = constants["eval_data"]
         B_plasma = constants["B_plasma"]
 
-        x = jnp.array([eval_data["R"], eval_data["phi"], eval_data["Z"]]).T
-
-        # B_ext is not pre-computed because field is not fixed
+        if not self._qfm_surface:
```
f0uriest, line 1331: minor point but I think its clearer if this is re-ordered to be 
```python
if qfm_surface:
    ...
else:
    ...
```

desc/objectives/_coils.py
```diff
@@ -1505,9 +1602,22 @@ def compute(self, field_params=None, constants=None):
         """
         if constants is None:
             constants = self.constants
+        field_params = params_1 if not self._qfm_surface else params_2
+        surf_params = params_2 if not self._qfm_surface else params_1
+
+        if not self._qfm_surface:
```
f0uriest, line 1608: see above

desc/objectives/_coils.py
dpanici: The thing is, by default `vacuum=False`, so an error would be thrown if `qfm_surface=True` is passed in.

I guess that is fine if we want users to have to be verbose about asking for qfm surface and specifying they are for a vacuum field

dpanici:
make a new objective for this specifically

dpanici:
update if #1205 is merged first

desc/objectives/_coils.py
```diff
@@ -1285,10 +1492,10 @@ class ToroidalFlux(_Objective):
 
     Parameters
     ----------
-    eq : Equilibrium
-        Equilibrium for which the toroidal flux will be calculated.
-        The Equilibrium is assumed to be held fixed when using this
-        objective.
+    eq : Equilibrium or FourierRZToroidalSurface
```
YigitElma, line 1495: General comment, why don't we change the argname to something like `thing`? For objectives that can take multiple different class in addition to `Equilibrium`.

desc/objectives/_coils.py
```diff
@@ -1267,6 +1283,197 @@ def compute(self, field_params, constants=None):
...
+    Used to find a quadratic-flux-minimizing (QFM) surface, so a
+    `FourierRZToroidalSurface` should be passed to the objective.
+    Should always be used along with a ``ToroidalFlux`` objective to
+    ensure that the resulting QFM surface has the desired amount of
+    flux enclosed and avoid trivial solutions.
+
+    Parameters
+    ----------
+    surface :  FourierRZToroidalSurface
+        QFM surface upon which the normal field error will be minimized.
```
YigitElma, line 1298: This is a dummy question but I just ask it to learn these kind of stuff. So, `QuadraticFlux` objective takes an equilibrium and computes the `B dot n` error on only the LCFS (then this keeps the equilibrium fixed and changes the coils to minimize the error). This new one though, takes a surface and minimizes the error on that. This could be either LCFS or not. Am I correct?

The part I am confused though is if the surface is also kept fixed or not? Docs of `ToroidalFlux` says that surface may vary.

YigitElma: And is this anyway related to free boundary objective?

desc/objectives/_coils.py
dpanici: we hesitate to change existing objective argnames in order to not break existing code using it

desc/objectives/_coils.py
dpanici: Ah yes I need to change ToroidalFlux to allow the thing to vary...

QFM the surface can vary.

It is sort of related, in that for vacuum free bdry techincally if you match the toroidal flux and the B dot n error is zero, that surface is the LCFS surface for a free bdry equilibrium (assuming the true field inside has nested surfaces).

Some people use QFM surfaces instead of doing full free bdry when they do single stage optimization for this reason, I believe, it is cheaper than doing actual free bdry

desc/objectives/_coils.py
```diff
@@ -1267,6 +1283,197 @@ def compute(self, field_params, constants=None):
         return f
 
 
+class QuadraticFluxMinimizingSurface(_Objective):
```
f0uriest, line 1286: this name makes it sound like this itself is a surface, not an objective. 

Maybe `SurfaceQuadraticFlux` or something?

desc/objectives/_coils.py
```diff
@@ -1148,6 +1151,13 @@ def __init__(
         self._field = field
         self._field_grid = field_grid
         self._vacuum = vacuum
+        errorif(
+            isinstance(eq, FourierRZToroidalSurface),
+            TypeError,
+            "Detected FourierRZToroidalSurface object "
+            "if attempting to find qfm_surface=True, please use "
```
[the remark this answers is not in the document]
dpanici: ah oops, yea holdover from when it was a flag

desc/objectives/_coils.py
```diff
@@ -1267,6 +1283,197 @@ def compute(self, field_params, constants=None):
         return f
 
 
+class QuadraticFluxMinimizingSurface(_Objective):
+    """Target B*n = 0 on a surface.
+
+    Used to find a quadratic-flux-minimizing (QFM) surface, so a
+    `FourierRZToroidalSurface` should be passed to the objective.
+    Should always be used along with a ``ToroidalFlux`` objective to
```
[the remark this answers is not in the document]
dpanici: I think that is true yes (volume at least, aspect ratio probably?)

desc/objectives/_coils.py
```diff
@@ -1285,10 +1492,10 @@ class ToroidalFlux(_Objective):
...
     Parameters
     ----------
-    eq : Equilibrium
-        Equilibrium for which the toroidal flux will be calculated.
-        The Equilibrium is assumed to be held fixed when using this
-        objective.
+    eq : Equilibrium or FourierRZToroidalSurface
+        Equilibrium (or QFM surface) for which the toroidal flux will be calculated.
+        The equilibrium is kept fixed during the optimization with this objective,
+        but if qfm_surface=True is passed, then the surface will be allowed to vary.
```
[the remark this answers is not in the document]
dpanici: I can just make the objective more general and give a eq fixed/free flag, as well as a coils fixed/free flag

docs/notebooks/tutorials/QFM_surface.ipynb
```diff
@@ -0,0 +1,783 @@
+{
```
YigitElma, line 1: <ul><li>I think \argmin is not a proper command. Maybe change it to \arg\min_S ?</li><li>I would prefer double Fourier series instead of "two Fourier series"</li><li>Can you also mention here that the surface has to enclose some given toroidal flux too</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1264/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

YigitElma, line 1: <pre class="ql-syntax" spellcheck="false">Line #19.    )
</pre><p>Can you put semicolon to the end, I get triggered by the text of matplotlib :D</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1264/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

YigitElma, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    )
</pre><p>Again I would appreciate a semicolon here :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1264/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

YigitElma, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    poincare_plot(field, R0=data["R"], Z0=data["Z"], ax=ax, NFP=eq.NFP, size=10)
</pre><p>;</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1264/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

desc/objectives/_coils.py
```diff
@@ -1267,6 +1283,197 @@ def compute(self, field_params, constants=None):
...
+    Should always be used along with a ``ToroidalFlux`` objective to
+    ensure that the resulting QFM surface has the desired amount of
+    flux enclosed and avoid trivial solutions.
+
+    Parameters
+    ----------
+    surface :  FourierRZToroidalSurface
+        QFM surface upon which the normal field error will be minimized.
+    field : MagneticField
+        External field produced by coils or other source, which will be optimized to
```
YigitElma, line 1300: Maybe mention `field_fixed` argument here to clarify that this doesn't always change? And also for `surface`, mention it is always allowed to change

desc/objectives/_coils.py
```diff
@@ -1267,6 +1283,197 @@ def compute(self, field_params, constants=None):
...
+
+        if self._normalize:
+            scales = compute_scaling_factors(surface)
+            Bscale = 1.0  # surface has no inherent B scale
+            self._normalization = Bscale * scales["R0"] * scales["a"]
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params_1, params_2=None, constants=None):
+        """Compute boundary force error.
```
YigitElma, line 1432:
```suggestion
        """Compute Bnorm on the QFM surface.
```

desc/objectives/_coils.py
```diff
@@ -1267,6 +1283,197 @@ def compute(self, field_params, constants=None):
...
+            self._surface,
+            self._data_keys,
+            surf_params,
+            constants["eval_transforms"],
+            constants["eval_profiles"],
+        )
+        x = jnp.array([eval_data["R"], eval_data["phi"], eval_data["Z"]]).T
+        if field_params is None:
+            field_params = constants["field"].params_dict
+        B_ext = constants["field"].compute_magnetic_field(
```
YigitElma, line 1466: A Python question. Is `constants["field"]` a copy of the field object at the time of build (so actually a constant) or a pointer to the field object such that when the object changes the `constants["field"]` also changes? If it is the former, then the compute function kind of assumes the field is always fixed?

desc/objectives/_coils.py
```diff
@@ -1301,12 +1508,26 @@ class ToroidalFlux(_Objective):
         Collocation grid containing the nodes to evaluate the normal magnetic field at
         plasma geometry at. Defaults to a LinearGrid(L=eq.L_grid, M=eq.M_grid,
         zeta=jnp.array(0.0), NFP=eq.NFP).
+    qfm_surface : bool
+        Whether to look for quadratic-flux-minimizing surfaces or not.
```
YigitElma, line 1512:
```suggestion
        Whether to look for quadratic-flux-minimizing surface or not.
```

desc/objectives/_coils.py
YigitElma: Probably, this implementation is correct but still a bit confused by something inside `constants` to be changing...

YigitElma: Okok, I check it is a pointer not a copy

dpanici: It is a pointer BUT also it would not matter if it was a copy, here we only are using its compute_magnetic_field method to use the internal algorithms, we pass in explicitly `field_params` with the current DOFs of the field, which then get used in computing the magnetic field.

YigitElma: Hmm ok, thanks

YigitElma (dismissed):

YigitElma (dismissed):

YigitElma (dismissed):

docs/notebooks/tutorials/QFM_surface.ipynb
```diff
@@ -0,0 +1,825 @@
...
```
f0uriest, line 1: <p>missing semicolon.</p><p> </p><p>also this might be better on a log scale?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1264/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

YigitElma (approved):

f0uriest (approved):

PR #1269, opened 2024-09-26
Adjusted logic for dealing with negative arguments in most_rational

lisamessier:
I'm a very new contributor and tried following along with the logic you gave. Let me know if this looks good.


rahulgaur104:
@cfe316 does this look correct?

cfe316:
Welcome @lisamessier and nice work!

1) Did you test each relevant case?
2) It might be preferable to keep the `# handle empty range` if statement and `# return 0 if in range` if statements early, just after the two lines normalizing `a` and `b` via the `float(_round(...))` functions. Keeping these return statements early means you don't need to think about as many different cases while reading through the rest of the logic. 

Also, this is totally separate, but I think in python the standard way to swap two variables is just `a, b = b, a`. Unless there's some reason to keep the three-line version this should be favored over the a/b/c shuffle. (I thought maybe it was a jax thing, but it works in jax as well.)

lisamessier:
1. No, I am trying to figure out how to do that right now. I'm looking in tests to see if there is an existing test that uses this function so I can add those specific cases. Is there any easier way to do this? I would appreciate any help!
2. I've made those changes (including the swap) but am going to wait to test it before I commit anything.

dpanici:
There is a test [here](https://github.com/PlasmaControl/DESC/blob/7d9ce27eb560da8ed6d2f426ad75742e49532eba/tests/test_grid.py#L818) however I don't think it ends up calling `most_rational` with two negative arguments. It would be worth putting another test in `tests/test_utils.py` that specifically tests calling `most_rational` with negative arguments

lisamessier:
Thank you! The test was able to run to completion without freezing, although I am not 100% sure the function is returning the correct value. I assume (-1, -2) will be interpreted as (1, 2)? And with this logic, (-2, -1) -> (2, 1) -> (1, 2)? Thanks for the assistance.

cfe316:
@dpanici is correct, we could add a new set of tests.

In the meantime (during development), one can check manually
```
from desc.grid import most_rational

all_same = [
most_rational(1, 2),
most_rational(2, 1),
most_rational(-1, -2),
most_rational(-2, -1)]
```

and check that those are all the same. Also can check that these return zeros:

```
has_zero = [most_rational(0, 1),
most_rational(0, -1),
most_rational(0,0),
most_rational(-1, 0),
most_rational(1, 0)]
```

(FWIW I'm pretty sure the logic is now correct.)

cfe316 (dismissed):
Looks good! I tested all the cases I could think of manually:

`normalizes_to_1_2 = [(1,2), (2, 1), (-1, -2), (-2, -1)]`
`includes_zero = [(0, 1), (0,0), (0, -1), (1,0), (-1, 0)]`

rahulgaur104:
Approving workflow run. @lisamessier please add the new tests that ensure this feature works for negative values.

lisamessier:
I will add new tests to test_utils.py if this is the desired location for them

rahulgaur104:
Perhaps, it would be better to just append your test to test_find_most_rational_surfaces() in test_grid.
Just add iota = PowerSeriesProfile([<negative coefficients>])

cfe316:
Hi @lisamessier ,

I'm sorry, I think I was being a bit too casual in my writing. The function must return a value which is between `a` and `b`, so yes, `a=-2, b=-1` will return a negative value. 

 > Because if we are going to return 0 in every case where one argument is negative and the other is positive, then a must be negative in another case (so b will also be negative) and they will both become positive numbers. Then, if a > b, then they will be swapped and evaluated as (1,2) in this example. Am I missing an important component of the logic? Removing s and returning dec allows the tests to pass without any issues.
 
I'm afraid I don't understand this part. The code looks good as it is now. The test `if sign(a) < 0` only needs to check `a` because at that point, either `a` and `b` must have the same sign, either `+,+` or `-,-`.

lisamessier:
Perfect, thank you for the clarification. In that case, it is working just as intended 😅 

I ran tests on the specific cases you gave earlier: `normalizes_to_1_2 = [(1,2), (2, 1), (-1, -2), (-2, -1)]` and `includes_zero = [(0, 1), (0,0), (0, -1), (1,0), (-1, 0)]` and those passed. I was a bit confused about the use of `iota = PowerSeriesProfile([-1, -2])`. Is this the correct way to format the test for a negative range? 
`# linear iota with negative range
    iota = PowerSeriesProfile([-1, -2])
    rho, io = find_most_rational_surfaces(iota, 5)
    np.testing.assert_allclose(rho, np.linspace(0, 1, 5), atol=1e-14, rtol=0)
    np.testing.assert_allclose(io, np.linspace(-1, -3, 5), atol=1e-14, rtol=0)`

lisamessier:
So sorry, I had some issues with Git. grid.py should be unmodified in the latest commit and test_grid.py should have the visible changes.

rahulgaur104 (approved):
I don't know why benchmarks are failing but all tests pass so we can merge it.

Issue #1271, opened 2024-09-26
Legends should appear on the ax object when passed to plotting functions


missing-user:
Some plotting functions support passing an Axis object to plot to, e.g. allowing to use the DESC plotting functions on subplots. At the moment the legend is created on the figure, instead of the axis, making it unclear which subplot the legend corresponds to.  

## Minimal Example
```python
import desc.io
import matplotlib.pyplot as plt

eq_fam = desc.io.load("tests/inputs/NCSX_older.h5")
eq = eq_fam[-1]
fig, ax = plt.subplots(2, 1)

desc.plotting.plot_1d(eq, "iota", ax=ax[0])
ax[0].legend(["Iota Profile"])
desc.plotting.plot_boundary(eq, ax=ax[1])
fig.show()

plt.show()
```

## Current Behavior
![current_behavior](https://github.com/user-attachments/assets/3b7173ca-2c29-4349-8727-d674d7f51246)

## Expected Behavior
![expected_behavior](https://github.com/user-attachments/assets/19c59e44-559e-4b49-aa49-fc2bb30a9e94)

PR #1272, opened 2024-09-26
Updated Legends to appear on ax instead of fig

missing-user:
Resolves https://github.com/PlasmaControl/DESC/issues/1271 "Legends should appear on the ax object when passed to plotting functions"
Resolves https://github.com/PlasmaControl/DESC/issues/1280 "Plotting Unit Test Regression"
- Changed default legend positioning from fig to ax, to better support using DESC functions in subplots
- Passing an array of phi values to plot_boundaries plots cross-sections for each phi value, instead of skipping the last one.
- Reduced tolerances for plotting unit tests
- Updated plotting baseline


tests/test_plotting.py
```diff
@@ -507,7 +507,7 @@ def test_plot_boundary_surface(self):
         return fig
 
     @pytest.mark.unit
-    @pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_2d)
+    @pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_1d)
```
YigitElma, line 510: I assume this change was for debugging. Can you change it back to `tol_2d`?

tests/baseline/test_plot_boundaries.png
YigitElma, line 1: Maybe we should resolve #1280 too in this PR!

YigitElma (commented):
Thanks you for your help @missing-user !

tests/test_plotting.py
missing-user: The change was intentional, since the boundaries plot has much less lines than i.e. a countour plot, image plot, heatmap... so the RMS is less sensitive to change and the threshold for the test should be lowered accordingly. 

If you think this will cause confusion, I'll revert the change.

tests/baseline/test_plot_boundaries.png
missing-user: I just wasn't sure how strictly the DESC team enforces PR atomicity :)

f0uriest (approved):
Tests pass locally for me with reduced tolerances. Would be good if a few other people can check as well, we've had issues in the past where if the tolerances are too tight we get a lot of false positives just due to hardware differences.

YigitElma:
Tests also pass for me but I think I have the same hardware as @f0uriest . Can @dpanici or @kianorr tests this on Mac?

YigitElma (approved):

YigitElma:
And please don't mind the `benchmark` action, our current CI doesn't allow PRs from forks to add automatic comments (which `benchmark` does at the end)

YigitElma:
> Resolve issue #1271 "Legends should appear on the ax object when passed to plotting functions"

Can you change the description to include "Resolves #1280" ?

missing-user:
Sorry, thought you meant the title for a second... I think everything is ready to merge now, thanks for the help!

dpanici:
works fine for me as well

dpanici (approved):

YigitElma:
@missing-user Thanks for the contribution!

PR #1275, opened 2024-09-26
Modify Coil Cutting to work with umbilic coils

dpanici:


dpanici:
Once #1511 is in, we won't need this and can close

PR #1277, opened 2024-09-27
Let check_intersection be passed to more coilset methods

dpanici:
- can now pass `check_intersection` flag for a `MixedCoilSet`'s conversion methods to avoid the intersection check, and also to `from_makegrid` for a `CoilSet` and `MixedCoilSet` to avoid the intersection check (some makegrids have a ton of coils...)


desc/coils.py
```diff
@@ -2487,7 +2489,12 @@ def to_FourierPlanar(
             minor radius r in a plane specified by a center position and normal vector.
 
         """
-        coils = [coil.to_FourierPlanar(N=N, grid=grid, basis=basis) for coil in self]
+        coils = [
+            coil.to_FourierPlanar(
+                N=N, grid=grid, basis=basis, check_intersection=check_intersection
```
f0uriest, line 2494: the base coils don't have a check_intersection option, just coilsets

dpanici: yes but a MixedCoilSet might contain CoilSets or other MixedCoilSets as its submembers, so we have to pass it here to this call, otherwise the submembers will check themselves for intersection when they convert

f0uriest (approved):

Issue #1276, opened 2024-09-26
Objective `deriv_mode` using `thing.dim_x` results in different Jacobian size


YigitElma:
The `deriv_mode` of each objective is decided by comparing the objectives `dim_f` to the half of the `dim_x` of the thing it is optimizing. After merging #1052 to the Poincare branch, I realized that the extra Poincare variables switch some objectives from `fwd` to `rev` mode auto-diff which is probably ok, but I had to remove `jac_chunk_size` argument from `test_ATF_results` since it was triggering one of the errors related to `jac_chunk_size` and `deriv_mode` consistencies.

Any thoughts on this?

YigitElma:
And probably another related issue is that if we have a single objective, the deriv mode being blocked or batched shouldn't cause a difference, right? If so, we may need an additional logic for set_derivatives to use batched since it is more suited for jac_chunk_size's checks.

dpanici:
hm, what is the error with the ATF one?

YigitElma:
This one,
```python
errorif(
            self._jac_chunk_size not in ["auto", None]
            and self._deriv_mode == "blocked",
            ValueError,
            "'jac_chunk_size' was passed into ObjectiveFunction, but the "
            "ObjectiveFunction is using 'blocked' deriv_mode, so sub-objective "
            "'jac_chunk_size' are used to compute each sub-objective's Jacobian, "
            "`ignoring the ObjectiveFunction's 'jac_chunk_size'.",
        )
```

dpanici:
How does adding poincare parameters cause the mode to change to rev? the dim_x of the objective is only the Rlmn Zlmn lambda LMN, and the dimf is set by the grid size?

Maybe these shuld be warnings though instead of error, I dont like that it can pop up like this

YigitElma:
No, `dim_x` is the sum of all optimizable parameters, so R_lmn+Z_lmn+L_lmn+Rb_lmn+Zb_lmn+Rp_lmn + ... axis stuff etc. And the reduced optimization variables are calculated during build of `LinearConstraintProjection` which doesn't change the `dim_x` or doesn't call `set_derivatives`.

f0uriest:
I think this is because it uses `jac_chunk_size=N` and `deriv_mode="auto"` whereas it should be specifying `deriv_mode="batched"`

f0uriest:
I see a few different cases:

|                   | deriv_mode="auto" | deriv_mode="batched" | deriv_mode="blocked" |
[...]

PR #1279, opened 2024-09-28
Allow "blocked" method for computing jvps

### LINKED ISSUE #1249: "blocked jvp" to allow finite differencing with `deriv_mode="blocked"`
> Slightly tangential comment: `deriv_mode="batched"` appears to compute the Jacobian wrt only the optimization variables. `deriv_mode="blocked"` however, appears to compute each objective's Jacobian wrt all parameters related to that objective (even if they are fixed by constraints). This is problematic for the `ExternalObjective`, since it requires extra finite differences, which are generally expensive. Is there a way to remedy that?

In theory yes. I think what we'd need is a "blocked jvp" where it computes the total jvp as a sum of smaller jvps (then `LinearConstraintProjection` will only take jvps in the directions needed). The main reason I didn't do this is that the "blocked" logic will use forward or reverse mode as appropriate, while "jvp" usually implies forward mode only, so if we use reverse mode when doing a jvp that might cause some issues in places where we assume jvp/vjp mean forward/reverse only. This isn't a dealbreaker, just something we need to think about a bit more.

_Originally posted by @f0uriest in https://github.com/PlasmaControl/DESC/issues/1052#issuecomment-2335033767_

f0uriest:
- Refactors some of the objective derivative stuff to use more efficient chunking/blocking etc.
- Adds chunked versions of jacfwd and jacrev

Resolves #1249

### DIFF
--- a/desc/batching.py
+++ b/desc/batching.py
@@ -1,7 +1,30 @@
 """Utility functions for the ``batched_vectorize`` function."""
 
 import functools
-from typing import Callable, Optional
+from functools import partial
+from typing import Any, Callable, Optional, Sequence
+
+import numpy as np
+from jax._src import core, dispatch, dtypes
+from jax._src.api_util import (
+    _ensure_index,
+    argnums_partial,
+    check_callable,
+    flatten_fun_nokwargs,
+    flatten_fun_nokwargs2,
+    shaped_abstractify,
+)
+from jax._src.interpreters import ad
+from jax._src.lax import lax as lax_internal
+from jax._src.tree_util import (
+    Partial,
+    tree_flatten,
+    tree_map,
+    tree_structure,
+    tree_transpose,
+    tree_unflatten,
+)
+from jax._src.util import safe_map, wraps
 
 from desc.backend import jax, jnp
 
@@ -16,6 +39,8 @@ from jax._src.numpy.vectorize import (
     _parse_gufunc_signature,
     _parse_input_dimensions,
 )
+
+_dtype = partial(dtypes.dtype, canonicalize=True)
 
 # The following section of this code is derived from the NetKet project
 # https://github.com/netket/netket/blob/9881c9fb217a2ac4dc9274a054bf6e6a2993c519/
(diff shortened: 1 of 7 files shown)


tests/test_derivatives.py
```diff
@@ -77,13 +77,17 @@ def test_fun(x, y, a):
...
 
-        jac_AD = AutoDiffDerivative(test_fun, argnum=0)
-        J_AD = jac_AD.compute(x, y, a)
+        jacf_AD = AutoDiffDerivative(test_fun, argnum=0, mode="fwd")
+        Jf_AD = jacf_AD.compute(x, y, a)
+        jacr_AD = AutoDiffDerivative(test_fun, argnum=0, mode="rev")
+        Jr_AD = jacr_AD.compute(x, y, a)
 
-        jac_FD = AutoDiffDerivative(test_fun, argnum=0)
+        jac_FD = FiniteDiffDerivative(test_fun, argnum=0)
```
f0uriest, line 85: this was wrong on master, it should be comparing FD to AD but was comparing AD to itself

desc/batching.py
```diff
@@ -320,3 +336,164 @@ def wrapped(*args, **kwargs):
...
+        )
+        tree_map(partial(_check_input_dtype_jacfwd, holomorphic), dyn_args)
+        if not has_aux:
+            pushfwd: Callable = partial(_jvp, f_partial, dyn_args)
+            y, jac = vmap_chunked(pushfwd, chunk_size=chunk_size)(_std_basis(dyn_args))
+            y = tree_map(lambda x: x[0], y)
+            jac = tree_map(lambda x: jnp.moveaxis(x, 0, -1), jac)
+        else:
+            pushfwd: Callable = partial(_jvp, f_partial, dyn_args, has_aux=True)
+            y, jac, aux = vmap_chunked(pushfwd, chunk_size=chunk_size)(
```
f0uriest, line 409: these bits are unused atm but will be necessary for #1034 and kept for now since its just copy/pasted from jax

unalmis:
Can you add a comment in code to https://github.com/PlasmaControl/DESC/blob/86376c43b1c01761c288c5180941d1f8a19d97d4/desc/batching.py#L108
explaining why we don't default to `desc.backend.imap` (alias for `jax.lax.map` with the new chunking option) when newer versions of JAX are installed?

dpanici (approved):

desc/objectives/objective_funs.py
```diff
@@ -978,11 +968,13 @@ def __init__(
...
-            # choose based on shape of jacobian. fwd mode is more memory efficient
-            # so we prefer that unless the jacobian is really wide
+            # choose based on shape of jacobian. dim_x is usually an overestimate of
+            # the true number of DOFs because linear constraints remove some. Also
+            # fwd mode is more memory efficient so we prefer that unless the jacobian
+            # is really wide
             self._deriv_mode = (
                 "fwd"
-                if self.dim_f >= 0.5 * sum(t.dim_x for t in self.things)
+                if self.dim_f >= 0.2 * sum(t.dim_x for t in self.things)
```
YigitElma, line 977: Probably resolves #1276

desc/derivatives.py
```diff
@@ -323,9 +326,13 @@ def _set_mode(self, mode) -> None:
 
         self._mode = mode
         if self._mode == "fwd":
-            self._compute = jax.jacfwd(self._fun, self._argnum)
+            self._compute = jacfwd_chunked(
```
YigitElma, line 329: A question about this. So, basically, now we never use `jax.jacfwd` or `jax.jacrev`, instead calculate the jacobian from jvp or vjp's by `batched_vectorize` manor. Then, why do we have new functions `jacfwd_chunked` and `jacrev_chunked`? As far as I understand, they basically do `batched_vectorize` in another way. For `blocked` can't we just call each sub-objective `jvp_` method and get the whole thing? Also, I think the `jvp_` and `jac_` methods in our code kind of used in an interchangeable way, right? If we give `v` for each parameter like `jnp.eye(dim_x)` to `jvp_` then it is basically `jac_`. I point out this because it gets complicated once we start wrapping stuff. Like, `LinearConstraintProjection` uses `jvp_` of objective for `jac_` but not for its own `jvp_`. I see this PR gets rid of the difference between `blocked` and `batched` but I still think we need a proper dev guide for this. How do we calculate jacobians? For each type of `_Objective` and `ObjectiveFunction`...

YigitElma (approved):
Apart from my general confusion, it looks ok. I may try to make a dev guide for this (AD in general) or since @f0uriest knows it well give some brief explanation and I can put it in an organized form.

desc/objectives/objective_funs.py
YigitElma: Yes, it resolves.

f0uriest:
> Can you add a comment in code to
> 
> https://github.com/PlasmaControl/DESC/blob/86376c43b1c01761c288c5180941d1f8a19d97d4/desc/batching.py#L108
> 
> explaining why we don't default to `desc.backend.imap` (alias for `jax.lax.map` with the new chunking option) when newer versions of JAX are installed?

@unalmis I thought the plan was to start using `lax.map` at some point once we increase the minimum jax version?

desc/objectives/objective_funs.py
f0uriest: I think we should keep that issue open, since there's still some confusion over which "auto" takes precedence that we should sort out.

desc/derivatives.py
f0uriest: `jacfwd_chunked` and `jacrev_chunked` are still used for `_Objective.jac`. It's a bit confusing since under the hood, `jacfwd` is literally just doing `vmap(jvp(fun))`. It's also messy because in JAX parlance, jvps (aka A@x) are always computed with forward mode AD, while vjp (x@A = A.T@x) are always done with reverse mode.

This PR updates things so that the `jvp` methods of objectives will now sometimes use reverse mode when efficient or required (by getting the full jacobian with jacrev and then doing a matmul). 

Here's a rough outline of how everything works now:
- `_Objective`:
  - `_Objective.jac` method uses `jacfwd` or `jacrev` (chunked) so materializes the full matrix using either forward or reverse mode as appropriate
  - `_Objective.jvp` computes `A@x`. If the objective uses forward mode, then this basically just calls `jax.jvp` which uses forward mode. However if the objective uses reverse mode, we use `_Objective.jac` to get the full matrix and then matmul. This avoids trying to do forward mode on objectives for which that is innefficient or undefined, and vice versa.
- `ObjectiveFunction`
  - `ObjectiveFunction.jac` always just calls batched `ObjectiveFunction.jvp`, see below
  - `ObjectiveFunction.jvp` depends on whether `deriv_mode` is "batched" or "blocked". If "batched", then this basically just calls `jax.jvp` on `ObjectiveFunction.compute`, using forward mode for everything. If "blocked" then it calls `_Objective.jvp` for each objective, which as noted above will use forward or reverse mode as appropriate.
- `LinearConstraintProjection`, `ProximalProjection` etc.
   - These objectives are basically wrappers, so their jacobian looks like `A@B` where `A` is the jacobian of the underlying `ObjectiveFunction` and `B` is the transformation (projection etc). 
   - For `*Projection.jac`, we do this as a batched `ObjectiveFunction.jvp`, so `A@x` where `x=B`, and it uses whatever deriv mode the underlying `ObjectiveFunction` and `_Objective` wants (ie batched/blocked/fwd/rev)
   - `*Projection.jvp` also just passes through to `ObjectiveFunction.jvp` but modifies the tangent, like `A@(B@x) = A@y,    y=B@x`

YigitElma: Thanks for the explanation. So, if the derivative we want is forward mode, then jvp or jac calls will result in a vectorized jvp (not always batched_vectorize though?). If it is reverse mode, than full jacobian is calculated in reverse mode ad and if vjp, the matmul is taken.

In a nutshell, can we say `jac_` method is a wrapper for jvp and vjps, such that if everthing was forward mode, we wouldn't need extra jac method?

f0uriest: i mean we never really need the `jac` methods, since they can always be written as a vectorized jvp/vjp, but its annoying to have to create an identity matrix of the right type each time.

The vectorization/vmapping is now always batched/chunked

PR #1283, opened 2024-09-30
Update rules for chunking jacobian

### LINKED ISSUE #1276: Objective `deriv_mode` using `thing.dim_x` results in different Jacobian size

f0uriest:
Resolves #1276


desc/objectives/objective_funs.py
```diff
@@ -178,12 +170,20 @@ def build(self, use_jit=None, verbose=1):
...
+        )
         sub_obj_jac_chunk_sizes_are_ints = [
             isposint(obj._jac_chunk_size) for obj in self.objectives
         ]
         errorif(
-            any(sub_obj_jac_chunk_sizes_are_ints) and self._deriv_mode != "blocked",
+            any(sub_obj_jac_chunk_sizes_are_ints) and self._deriv_mode == "batched",
             ValueError,
             "'jac_chunk_size' was passed into one or more sub-objectives, but the"
             " ObjectiveFunction is  using 'batched' deriv_mode, so sub-objective "
```
YigitElma, line 189: can we delete the extra space here, since we are at it :D

f0uriest: what extra space?

dpanici (changes requested):
Should put a note in the docstring of _Objective of what `"auto"` for `jac_chunk_size` does now (uses a conservative heuristic estimate based on problem size)

desc/objectives/objective_funs.py
YigitElma:
```suggestion
            " ObjectiveFunction is using 'batched' deriv_mode, so sub-objective "
```

f0uriest:
> Should put a note in the docstring of _Objective of what `"auto"` for `jac_chunk_size` does now (uses a conservative heuristic estimate based on problem size)

"auto" is still only an option for `ObjectiveFunction`, not `_Objective` so I don't think it makes sense to describe it there.

dpanici (approved):

desc/objectives/objective_funs.py
```diff
@@ -325,35 +317,34 @@ def build(self, use_jit=None, verbose=1):
...
+            ValueError,
+            "'jac_chunk_size' was passed into ObjectiveFunction, but the "
+            "ObjectiveFunction is not using 'batched' deriv_mode",
+        )
         sub_obj_jac_chunk_sizes_are_ints = [
             isposint(obj._jac_chunk_size) for obj in self.objectives
         ]
         errorif(
-            any(sub_obj_jac_chunk_sizes_are_ints) and self._deriv_mode != "blocked",
+            any(sub_obj_jac_chunk_sizes_are_ints) and self._deriv_mode == "batched",
```
YigitElma, line 333: This is again a design choice but let's say I have a single objective,
```python
objective = ObjectiveFunction(ForceBalance(eq, jac_chunk_size))
```
This may complicate the logic but could we add something like `errorif( ..... and not len(self._objectives) == 1)` ? 

Some other edge case could be giving same `jac_chunk_size` to each objective, which should be allowed but I don't think anyone wanna use it that way, it is more laborious...

YigitElma (approved):

PR #1284, opened 2024-09-30
Make naming of grids among objectives more uniform 

### LINKED ISSUE #997: Make naming of grids among objectives more uniform
`VacuumBoundaryError` and `BoundaryError`both have need of an `eval_grid` and `source_grid` yet the former uses `grid` as the kwarg for this while the latter uses `eval_grid`, making these uniform would be nice

felicia22:
Resolves #997


dpanici:
Thanks, it looks good! But can you add a catch to check if `grid` is passed in, to catch if so and have it still work, but emit a warning? something like is done [here](https://github.com/PlasmaControl/DESC/blob/056fcc4e6c42ec5cc6f2e7909aed053b3508d5b8/desc/objectives/_geometry.py#L682) using the `parse_argname_change` utility function.

Also, some tests (and possibly the [free_boundary tutorial](https://github.com/PlasmaControl/DESC/blob/master/docs/notebooks/tutorials/free_boundary_equilibrium.ipynb) notebook) might pass in `grid` directly as a kwarg to `VacuumBoundaryError`, in which case there will be errors now that we changed the name. You should try checking these for any that do so and making the change from `grid` to `eval_grid` there

dpanici (changes requested):
see above comment

felicia22:
Hi, I made the edits requested (I think that I used the function correctly). I didn't see any tests directly pass in grid as a kwarg. I ran "pytest tests -k test_compute_everything" and it seems to be working for me but let me know if I should make any other edits!.

desc/objectives/_free_boundary.py
```diff
@@ -104,15 +104,17 @@ def __init__(
...
         deriv_mode="auto",
-        grid=None,
+        eval_grid=None,
         field_grid=None,
         field_fixed=False,
         name="Vacuum boundary error",
         jac_chunk_size=None,
+        **kwargs,
     ):
+        self._eval_grid = parse_argname_change(eval_grid, kwargs, "grid", "eval_grid")
```
f0uriest, line 114: Currently, if the user passed in `grid=foo`, then this will correctly assign grid to `self._eval_grid` however, the line below
`self._eval_grid = eval_grid` 
would then overwrite it with `None`

I think the fix is just make this `eval_grid = parse_argname_change(...` and then the line below will work as intended.

f0uriest (approved):

YigitElma (approved):

PR #1286, opened 2024-10-02
C0 fourier

### LINKED ISSUE #1122: C0 `FourierPlanarCurve` Implementation
I think `FourierPlanarCurve` should be possibly straightforward to make into a C0 curve. Say we have 2 C0 points, and we have two curves defined by 2 sets of Fourier coefficients describing the radii from a given center (with the same normal vector as well, so these two curves are coplanar). The constriant they satisfy is, say

$$
\begin{align}
r_1(\theta=0) &= r_2(\theta=0)\\
r_1(\theta=\pi) &= r_2(\theta=\pi)
\end{align}
$$
which are simply linear constraints on the coeffiicents of $r_1$ and $r_2$

Then the C0 curve is just 

$$
\begin{align}
r(\theta) = \begin{cases}
r_1(\theta) ~~~ \text{if} ~ 0\leq \theta \le \pi\\
r_2(\theta) ~~~ \text{if} ~ \pi\leq \theta \le 2\pi\\
\end{cases}
\end{align}
$$

we could pretty easily make a wrapper class called like `C0FourierPlanarCurve` that basically contains `N_{C0~pts}` `FourierPlanarCurve` objects and its DOFs are all the Fourier coeffs of its underlying curves, and all that would be needed is a linear objective for this class that applies the above constraint. Also would have to add a compute sub-parameterization for the C0FourierPlanarCurve with the logic for which params to use as well which might be tricky, or that logic could go into the C0FourierPlanarCurve compute definition, maybe (any difficulty lies in this part I think)

kianorr:
Resolves #1122


kianorr:
I get this after optimizing with a length target of 2pi:

image

dpanici:
This is actually not useful/ill-posed, I think we can close, we are able to do C0 better with #970

Issue #1288, opened 2024-10-02
JIT Error encountered due to jax, hashing, and `constants`


dpanici:
Error seems to occur when optimizing `GammaC` objective on `gh/Gamma_c` branch, happens on the second optimization step and seems related to the JIT cache? The error also only occurs if attempting an optimization at a resolution that you have previously optimized at, changing the eq resolution between steps seems to avoid this issue, so I assume it is related to the caching

MWE:
```python
from desc import set_device

set_device("gpu")

import jax
import numpy as np

import desc.examples
from desc.continuation import solve_continuation_automatic
from desc.equilibrium import EquilibriaFamily, Equilibrium
from desc.geometry import FourierRZToroidalSurface
from desc.grid import ConcentricGrid, LinearGrid
from desc.io import load
from desc.objectives import (  # FixIota,
    AspectRatio,
    Elongation,
    FixBoundaryR,
    FixBoundaryZ,
    FixCurrent,
    FixPressure,
    FixPsi,
    ForceBalance,
    GammaC,
    GenericObjective,
    ObjectiveFunction,
    QuasisymmetryTwoTerm,
)
from desc.optimize import Optimizer
from desc.plotting import plot_boozer_surface
import pdb
from desc.backend import jnp
from desc.examples import get
def run_opt_step(k, eq):
    """Run a step of the optimization example."""
    # this step will only optimize boundary modes with |m|,|n| <= k
    # we create an ObjectiveFunction, in this case made up of multiple objectives
    # which will be combined in a least squares sense

    shape_grid = LinearGrid(
        M=int(eq.M), N=int(eq.N), rho=np.array([1.0]), NFP=eq.NFP, sym=True, axis=False
    )

    ntransits = 8

    zeta_field_line = np.linspace(0, 2 * np.pi * ntransits, 64 * ntransits)
    alpha = jnp.array([0.0])
    rho = jnp.linspace(0.85, 1.0, 2)
    # rho = np.linspace(0.85, 1.0, 2)
    flux_surface_grid = LinearGrid(
        rho=rho, M=eq.M_grid, N=eq.N_grid, sym=eq.sym, NFP=eq.NFP
    )

    objective = ObjectiveFunction(
        (
            GammaC(
                eq=eq,
                rho=rho,
                alpha=alpha,
                deriv_mode="fwd",
                batch=False,
                weight=1e3,
                Nemov = False,
            ),
[...]

PR #1290, opened 2024-10-03
ε and Γ_c with Fourier Bounce

unalmis:
This PR updates the
- `Gamma_c`, `Gamma_c Velasco` #1042 
- `effective ripple` #1003

objectives to use the numerical methods from #1119 

Also includes some incremental performance improvements


unalmis:
The objectives have been replaced with the newer methods, but my vote is to keep both the old and new compute functions for effective ripple and Gamma_c. The old ones have a `*` appended to the `data_index` name.

tests/baseline/test_effective_ripple_1D.png
unalmis, line 1: This is converged

rahulgaur104:
Here's what I am trying to run based on your example script
```
#!/usr/bin/env python3
from desc import set_device
set_device("gpu")

import numpy as np

from desc.integrals.bounce_integral import Bounce1D, Bounce2D


eq = Equilibrium.load("/scratch/gpfs/rg6256/eq_omni_NFP3_10.h5")
rho_arr = np.linspace(0, 1, 12)
data_arr = np.zeros((len(rho_arr),))


grid = LinearGrid(rho=rho_arr, theta=eq.M_grid, zeta=eq.N_grid, NFP=eq.NFP, sym=False)
num_transit = 1
data = eq.compute(
    "effective ripple 3/2",
    grid=grid,
    theta=Bounce2D.compute_theta(eq, X=32, Y=64, rho=rho_arr),
    num_transit=num_transit,
    num_quad=64,
    num_pitch=num_pitch,
    num_well=20 * num_transit,
)

data_arr = data["effective ripple 3/2"]


print(data["effective ripple"])

```


and it throws the following error
```
DESC version 0.12.2+824.gd20ee8702,using JAX backend, jax version=0.4.31, jaxlib version=0.4.31, dtype=float64
Using device: NVIDIA A100-SXM4-80GB (id=0), with 80.00 GB available memory
  File "/scratch/gpfs/rg6256/DESC_tests/ripple_calculator_2D.py", line 25, in <module>
    theta=Bounce2D.compute_theta(eq, X=32, Y=64, rho=rho_arr),
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/gpfs/rg6256/DESC/desc/integrals/bounce_integral.py", line 543, in compute_theta
    return eq.map_coordinates(
           ^^^^^^^^^^^^^^^^^^^
  File "/scratch/gpfs/rg6256/DESC/desc/equilibrium/equilibrium.py", line 1210, in map_coordinates
    return map_coordinates(
           ^^^^^^^^^^^^^^^^
  File "/scratch/gpfs/rg6256/DESC/desc/equilibrium/coords.py", line 119, in map_coordinates
    return _map_clebsch_coordinates(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/gpfs/rg6256/DESC/desc/equilibrium/coords.py", line 477, in _map_clebsch_coordinates
    theta, (res, niter) = vecroot(guess, alpha, rho, zeta, iota)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/gpfs/rg6256/DESC/desc/equilibrium/coords.py", line 461, in <lambda>
    lambda x0, *p: root_scalar(
                   ^^^^^^^^^^^^
TypeError: root_scalar() got an unexpected keyword argument 'X'
--------------------
```

unalmis:
Maybe should share, I've been using this to help me guess resolution requirements
```python
def well_plotter(
    eq, grid, theta, Y_B=None, num_transit=10, num_pitch=10, num_well=None
):
    """Use this to analyse what resolution may be needed."""
    data = eq.compute(Bounce2D.required_names + ["min_tz |B|", "max_tz |B|"], grid=grid)
    bounce = Bounce2D(grid, data, theta, Y_B, num_transit=num_transit)
    pitch_inv, _ = Bounce2D.get_pitch_inv_quad(
        grid.compress(data["min_tz |B|"]),
        grid.compress(data["max_tz |B|"]),
        num_pitch,
    )
    points = bounce.points(pitch_inv, num_well=num_well)
    # calling this automatically shows plots on my local machine.
    plots = bounce.check_points(points, pitch_inv)
    return plots
```

desc/compute/_neoclassical.py
```diff
@@ -0,0 +1,509 @@
...
+        fieldline_quad = kwargs["fieldline_quad"]
+    else:
+        fieldline_quad = leggauss(Y_B // 2)
+    if "quad" in kwargs:
+        quad = kwargs["quad"]
+    else:
+        quad = get_quadrature(
+            leggauss(kwargs.get("num_quad", 32)),
+            (automorphism_sin, grad_automorphism_sin),
+        )
```
unalmis, line 410: @ddudt last week the default quadrature for the 2D stuff here was using the quadratures best for weak singularities i.e.
```python
chebgauss2(num_quad)
```
and I have since corrected it.

The 1D stuff is unchanged since it was using the correct quadrature.

unalmis:
> Here's what I am trying to run based on your example script after checking out this branch
>  JAX errors in root stuff

I can't reproduce. maybe let's wait until things like #1293 get merged before debugging

tests/baseline/test_effective_ripple_2D.png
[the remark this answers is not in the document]
unalmis: Due to how we make sure enough resolution is used when we call `eq.compute`, the code I gave above actually computes this quantity by computing all dependencies of the effective ripple on this higher resolution grid:
```python
all_dependencies_grid = LinearGrid(rho=rho, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False)
```
Because #1313 was not resolved yet, I had marked `effective ripple` as a quantity that should not be overriden when we call `eq.compute`. That means that once all the input data to the `effective ripple` was computed accurately on the above grid. The computation of the effective ripple itself used this grid
```python
eps_eff_grid = LinearGrid(rho=rho, theta=eq.M_grid, zeta=eq.N_grid, NFP=eq.NFP, sym=False)
```
This grid has half as many nodes in poloidal and toroidal directions on which we compute FFTs to do all the integrals. The plot above is the result of this computation against Neo. To generate the Neo plot we converted the equilibrium to boozer coordinates at three times the DESC coordinate resolution.

So the conclusion is that when all the bounce integrals are done using FFTs that are generated from half typical desc resolution, then effective ripple  converges to Neo produces from an input with boozer resolution three times the DESC resolution.



Now that #1313 is resolved, I have merged that PR into this branch, and marked effective ripple as a flux surface quantity (by saying resolution_requirement="tz"). That means unless `override_grid=False` is given then everything, including all the bounce integrals will be computed on a grid with resolution at least
```python
all_dependencies_grid = LinearGrid(rho=rho, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False)
```

The old behavior can be recovered by marking `effective ripple 3/2` compute as something other than requiring theta and zeta resolution. i.e. changing `resolution_requirement="tz"` to `resolution_requirement="z"`. Then the code I gave above (repeated here), will reproduce that plot.
```python
eq = get("W7-X")
rho = np.linspace(1e-12, 1, 60)  # the axis limit IS implemented by the way
grid = LinearGrid(rho=rho, theta=eq.M_grid, zeta=eq.N_grid, NFP=eq.NFP, sym=False)
num_transit = 64
data = eq.compute(
    "effective ripple 3/2",
    grid=grid,
    theta=Bounce2D.compute_theta(eq, X=32, Y=64, rho=rho),
    num_transit=num_transit,
    num_quad=64,
    num_pitch=200,
    num_well=20 * num_transit,
)
```

dpanici:
If this PR depends on a function from a newer version of JAX @ddudt @unalmis , update the requirements to reflect this

unalmis:
> > Here's what I am trying to run based on your example script after checking out this branch
> > JAX errors in root stuff
> 
> I can't reproduce. maybe let's wait until things like #1293 get merged before debugging

Still can't reproduce.

unalmis:
Another optimization script. Currently queued on cluster.
[multigrid_multimode_opt.zip](https://github.com/user-attachments/files/17926861/multigrid_multimode_opt.zip)

desc/integrals/bounce_integral.py
```diff
@@ -290,31 +293,20 @@ def __init__(
             ``FourierChebyshevSeries.nodes(X,Y,rho,domain=(0,2*jnp.pi))``.
             Use the ``Bounce2D.compute_theta`` method to obtain this.
         Y_B : int
-            Desired Chebyshev spectral resolution for |B|.
-            Default is to double the resolution of ``theta``.
+            Desired resolution for |B| along field lines to compute bounce points.
```
dpanici, line 296: resolution is in what though? should say (Chebyshev)

dpanici (changes requested):
Lot of small things, mainly relating to improving documentation.
May also be good to add a test which uses these in optimization (esp. to test error from #1288 which I think might occur here too?)

desc/compute/_field.py
```diff
@@ -105,6 +105,221 @@ def _B_sup_zeta(params, transforms, profiles, data, **kwargs):
...
+    description="Contravariant cylindrical toroidal angle component of magnetic field",
+    dim=1,
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["psi_r/sqrt(g)", "phi_z", "theta_PEST_t", "phi_t", "theta_PEST_z"],
+)
+def _B_sup_phi(params, transforms, profiles, data, **kwargs):
+    # Written like this, independent of iota, to enable computing without integration.
```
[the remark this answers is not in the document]
unalmis: Yes. This reduces the computational graph, and therefore the memory penalty of AD.

desc/equilibrium/coords.py
```diff
@@ -105,7 +105,11 @@ def map_coordinates(  # noqa: C901
             f"don't have recipe to compute partial derivative {key}",
         )
 
-    profiles = get_profiles(inbasis + basis_derivs, eq)
+    profiles = (
+        kwargs["profiles"]
```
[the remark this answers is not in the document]
unalmis: This is the start of the fix that Rahul avoided by using the same technique I use for the coordinate mapping in #1335. It is not documented because it's not fully fixed yet, and the api may change etc.

devtools/dev-requirements_conda.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - pip:
     # Conda only parses a single list of pip requirements.
     # If two pip lists are given, all but the last list is skipped.
-    - jax >= 0.4.24, < 0.5.0
+    - jax >= 0.4.31, < 0.5.0
```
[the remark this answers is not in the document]
unalmis: That's when JAX added it

requirements.txt
```diff
@@ -1,4 +1,4 @@
-jax >= 0.4.24, < 0.5.0
+jax >= 0.4.31, < 0.5.0
```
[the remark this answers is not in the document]
unalmis: see above

requirements_conda.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - pip:
     # Conda only parses a single list of pip requirements.
     # If two pip lists are given, all but the last list is skipped.
-    - jax >= 0.4.24, < 0.5.0
+    - jax >= 0.4.31, < 0.5.0
```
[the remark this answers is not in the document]
unalmis: see above

tests/test_integrals.py
```diff
@@ -938,93 +936,66 @@ class TestBounceQuadrature:
...
     def _fixed_elliptic(integrand, k, deg):
         k = np.atleast_1d(k)
-        a = np.zeros_like(k)
+        a = -2 * np.arcsin(k)
         b = 2 * np.arcsin(k)
         x, w = get_quadrature(leggauss(deg), (automorphism_sin, grad_automorphism_sin))
         Z = bijection_from_disc(x, a[..., np.newaxis], b[..., np.newaxis])
         k = k[..., np.newaxis]
-        quad = integrand(Z, k).dot(w) * grad_bijection_from_disc(a, b)
+        quad = integrand(Z, k).dot(w) * grad_bijection_from_disc(a, b) / 2
```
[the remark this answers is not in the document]
unalmis: because I changed the integration from [0, 2 arcsin] to [-2 arcsin, 2 arcsin]. This is because the modified chebyshev quadrature I use only works with singularity at both ends, while the modified gass legendre handles anything (no singularity at 0)

tests/test_neoclassical_1D.py
```diff
@@ -0,0 +1,131 @@
...
+        data["V_r(r)"] / (4 * np.pi**2),
+        rtol=1e-3,
+    )
+    assert np.all(data["fieldline length"] > 0)
+    assert np.all(data["fieldline length/volume"] > 0)
+
+
+@pytest.mark.unit
+@pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_1d)
+def test_effective_ripple_1D():
```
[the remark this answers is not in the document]
unalmis: Like a minute. 2 minute for Gamma. Most of the time is spent on the coordinate  mapping which should be solved by #1154 . The resolution can be reduced.

desc/compute/_neoclassical_1D.py
```diff
@@ -0,0 +1,461 @@
...
+)
+def _effective_ripple_1D(params, transforms, profiles, data, **kwargs):
+    data["deprecated(effective ripple)"] = data["deprecated(effective ripple 3/2)"] ** (
+        2 / 3
+    )
+    return data
+
+
+@register_compute_fun(
+    name="deprecated(Gamma_c)",
```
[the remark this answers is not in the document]
unalmis: They are not tested in `test_compute_everything`.

I don't want them to show up in the list of variables either.

desc/objectives/_neoclassical.py
```diff
@@ -0,0 +1,528 @@
...
+
+_bounce_overwrite = {
+    "deriv_mode": """
+    deriv_mode : {"auto", "fwd", "rev"}
+        Specify how to compute Jacobian matrix, either forward mode or reverse mode AD.
+        ``auto`` selects forward or reverse mode based on the size of the input and
+        output of the objective. Has no effect on ``self.grad`` or ``self.hess`` which
+        always use reverse mode and forward over reverse mode respectively.
+
+        Default is ``fwd``. If ``rev`` is chosen, then ``jac_chunk_size=1`` is chosen
```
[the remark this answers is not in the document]
unalmis: Yes, but for this scalar quantity, jac_chunk_size is basically how many flux surface to differentiate at a time. For 5 flux surface, the max jac chunk size would be 5. In the 2D methods, I detach from the 3D fourier zernike to a 2D spectral represenation early with the 3D fourier zernike transform as a precomputed constant, so the compute graph of the flux surfaces is mostly separable into flux surface. 

The coordinate mapping doesn't do the partial summation right now, so I think the compute graph is still linked across flux surfaces due to the Zernike stuff there. I think once I do that then the jac chunk size will have less effect. This is  informal analysis.

desc/equilibrium/coords.py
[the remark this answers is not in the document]
dpanici: Is there an issue to link for this?

tests/test_neoclassical_1D.py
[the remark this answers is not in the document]
dpanici: I would throw a slow marker on them as well then just to document they take a little while

desc/compute/_neoclassical_1D.py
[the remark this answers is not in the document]
dpanici: Oh I see, you remove them from the name to compute in that test.

If we can also remove these from docs, that would be great, possibly by adding something to avoid things with `deprecated` in their names to https://github.com/PlasmaControl/DESC/blob/master/docs/write_variables.py

desc/objectives/_neoclassical.py
```diff
@@ -0,0 +1,528 @@
...
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        eq = self.things[0]
+        if self._grid is None:
+            self._grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False)
+        assert self._grid.can_fft
+        # Constants can only hold 1D arrays for some reason.
```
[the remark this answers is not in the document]
YigitElma: Does the error still persist if you make a copy instead of flattening it? Maybe it has to do something with pointers? It could be that somewhere in the code `FourierChebyshevSeries.nodes` changes size and it causes an inconsistency... I didn't look at it in detail but `.ravel()` creates a copy, just a thought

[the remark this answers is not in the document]
unalmis: Thanks. I should emphasize the issue has been "resolved" here, and I have made the GitHub issue so that we document that `constants` must be a 1D array only, or fix that limitation.

> In my case, I obtain the same error if I replace the objective.compute here with a dummy return jnp.ones(self._dim_f), so this error has nothing to do with bounce integrals.

With that mentioned modification, the fourierchebyshev.nodes array isn't referenced after it is made in `build`, so I don't think it's related.

desc/compute/_neoclassical.py
```diff
@@ -0,0 +1,626 @@
...
+        Used to compute bounce integrals.
+        Quadrature points xₖ and weights wₖ for the
+        approximate evaluation of the integral ∫₋₁¹ f(x) dx ≈ ∑ₖ wₖ f(xₖ).
+        """,
+    "spline": "bool : Whether to use cubic splines to compute bounce points.",
+}
+
+
+def _compute(fun, fun_data, data, theta, grid, num_pitch, simp=False):
+    """Compute ``fun`` for each ρ value iteratively to reduce memory usage.
```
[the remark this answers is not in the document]
unalmis: Probably linear in number of flux surface in compute. For reverse mode you could remove the call to imap to vectorize and it would probably make less difference on memory since the jacobian computation still dominates.

desc/objectives/_neoclassical.py
```diff
@@ -0,0 +1,528 @@
...
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        deriv_mode="fwd",
+        grid=None,
+        name="Effective ripple",
+        jac_chunk_size=None,
+        *,
+        X=16,  # X is cheap to increase.
```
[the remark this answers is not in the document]
unalmis: Whether it is cheap or not depends on the resolution, and so I do not want to add such statments publically until #1154 is resolved, as that may change the default used resolution.

desc/objectives/_neoclassical.py
```diff
@@ -0,0 +1,528 @@
...
+        Y=32,
+        # Y_B is expensive to increase if one does not fix num well per transit.
```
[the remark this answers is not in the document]
unalmis: see above

desc/objectives/_neoclassical.py
```diff
@@ -0,0 +1,528 @@
...
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        deriv_mode="fwd",
+        grid=None,
+        name="Gamma_c",
+        jac_chunk_size=None,
+        *,
+        X=16,  # X is cheap to increase.
```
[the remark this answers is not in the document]
unalmis: see above

desc/objectives/_neoclassical.py
```diff
@@ -0,0 +1,528 @@
...
+        Y=32,
+        # Y_B is expensive to increase if one does not fix num well per transit.
```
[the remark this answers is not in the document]
unalmis: see above

devtools/dev-requirements_conda.yml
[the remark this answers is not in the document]
f0uriest: I'd prefer to not require such a recent version of jax if possible. If the only new feature is `imap` can we not just use our existing batched vmap?

desc/compute/_neoclassical.py
```diff
@@ -0,0 +1,626 @@
...
+def _gbdrift(data, B, pitch):
+    return safediv(
+        (data["gbdrift (periodic)"] + data["gbdrift (secular)/phi"] * data["zeta"])
+        * (1 - 0.5 * pitch * B),
+        jnp.sqrt(jnp.abs(1 - pitch * B)),
+    )
+
+
+@register_compute_fun(
+    name="Gamma_c Velasco",
```
f0uriest, line 513: did we sort out whether this goes to zero for an infinite fieldline?

desc/integrals/bounce_integral.py
```diff
@@ -49,15 +48,19 @@ def points(self, pitch_inv, *, num_well=None):
...
 
     @abstractmethod
-    def check_points(self, points, pitch_inv, *, plot=True, **kwargs):
+    def check_points(self, points, pitch_inv, *, plot=True):
         """Check that bounce points are computed correctly."""
 
     @abstractmethod
     def integrate(
-        self, integrand, pitch_inv, f=None, weight=None, points=None, *, quad=None
+        self, integrand, pitch_inv, data=None, names=None, points=None, *, quad=None
```
f0uriest, line 56: should make these API changes (and others in this file) in #1119

desc/objectives/_neoclassical.py
```diff
@@ -0,0 +1,532 @@
...
+        Grid resolution in poloidal direction for Clebsch coordinate grid.
+        Preferably power of 2.
+    Y : int
+        Grid resolution in toroidal direction for Clebsch coordinate grid.
+        Preferably power of 2.
+    Y_B : int
+        Desired resolution for algorithm to compute bounce points.
+        Default is double ``Y``. Something like 100 is usually sufficient.
+        Currently, this is the number of knots per toroidal transit over
+        to approximate |B| with cubic splines.
```
f0uriest, line 82: is this still the case? I thought this one wasnt using splines?

desc/objectives/_neoclassical.py
```diff
@@ -0,0 +1,532 @@
...
+    num_transit : int
+        Number of toroidal transits to follow field line.
+        For axisymmetric devices, one poloidal transit is sufficient. Otherwise,
+        assuming the surface is not near rational, more transits will
+        approximate surface averages better, with diminishing returns.
+    num_well : int
```
f0uriest, line 88: A thought:

Suppose we didn't have num_well high enough so that we were ignoring some wells. However, neoclassical optimization tends to smooth out small wells, so that after a few iterations, the actual number of wells will likely decrease, so that now num_well may be large enough? IE, could we set num_well relatively small and rely on the optimization to smooth them out successively?

desc/objectives/objective_funs.py
```diff
@@ -203,18 +206,17 @@ class ObjectiveFunction(IOAble):
     name : str
         Name of the objective function.
     jac_chunk_size : int or "auto", optional
-        If `"batched"` deriv_mode is used, will calculate the Jacobian
+        Will calculate the Jacobian
```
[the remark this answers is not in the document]
unalmis: I thought this was outdated.
- We call it  `chunked` right now?
- We can chunk even if `fwd` and `rev` is used instead of `batched` or `chunked`

desc/objectives/_neoclassical.py
unalmis: The splines are only used to compute the bounce points, and they are generated from a fourier series in (theta, zeta) (output of `rfft2`), which is much less expensive that to do it from FourierZernike.

Except for high NFP cases (e.g. heliotron), chebyshev approximation is significantly more efficient than splines. However, if I use chebyshev series for root finding with bounce points, we need to reverse mode differentiate through an eigenvalue solve. A naive reliance on a jnp.vectorize(orthax.chebroot)` would not be efficient for that. To use chebyshev approximation with AD, the algorithm I wrote out to avoid all but one of the eigenvalue solves per transit should be implemented. I stopped development work on algorithm; if I were to continue development on bounce integrals, I would be solving #1303 instead, i think probably with jax ffi stuff to hook in c code.

desc/objectives/_neoclassical.py
unalmis: 1. If you're targeting effective ripple = 0 and the optimizer succeeds yes. If you are targeting a finite ripple, you should keep in mind that in these metrics we compute some `function ~ (sum over all wells) / (length of field line)`. If num well is less than the actual number of wells over the sample of the field line, then the sum in the numerator terminates early, so the metric would be underestimated. Then the optimization may terminate early.
2. Is there a reason you suggest that over ignoring narrow wells when they are computed in `bounce.points`? That could be done fine, so that the integrals is done over `num_well` wells with `min_well_length`.
3. For ripple its clear 2 is reasonable. For Gamma_c, the pitch angle determines the contribution of the integral moreso that the width, so it's not immediately obvious to me that 2 works.
4. As you know for high NFP configurations smoothing isn't really relevant

desc/objectives/_neoclassical.py
unalmis: https://github.com/PlasmaControl/DESC/pull/1119#discussion_r1868381236

desc/compute/_neoclassical.py
unalmis: There is no change in the status on this.

The secular term drives  `Gamma_c Velasco` to zero for an infinite fieldline. Dr. Velasco informed me they had proven the secular term is supposed to average out of the final result on paper. This may provide more motivation for #1428 . That being said the results don't change even at high resolution for fixed quadrature, and we know the quadratures are supposed to converge on paper and empirically.

Once #1406 udpates the equilibrium from #1369 I think it's worth checking again.

unalmis: Indeed, if I remove the secular term from the bounce integrals, the result converges to a nonzero profile quickly.

unalmis:
The benchmarks for 3.9 are supposed to fail since I removed python 3.9.

unalmis:
other tests pass locally but a few fail on CI for reasons unknown to me

desc/compute/_neoclassical_1D.py
unalmis, line 1: https://www.diffchecker.com/ER2rgXBa/

desc/compute/_neoclassical.py
unalmis, line 1: Best to view side by side and ignore old version here. The relevant diff to old stuff is https://github.com/PlasmaControl/DESC/pull/1290/files#r1889184785.

tests/baseline/test_Gamma_c_Velasco.png
unalmis, line 1: The reviewer should compare the deleted plot to the `Gamma_c_Velasco_1D` plot

unalmis, line 1: The reviewer should compare the deleted plot to the `Gamma_c_Velasco_1D` plot. Keep https://github.com/PlasmaControl/DESC/pull/1290#discussion_r1870623560 in mind.

tests/baseline/test_Gamma_c.png
unalmis, line 1: The reviewer should compare the deleted plot to the `Gamma_c_1D` plot

tests/inputs/master_compute_data_rpz.pkl
unalmis, line 1: Please see the message I sent in the slack.

f0uriest (changes requested):
Mostly small documentation and API stuff.

Main thing is I think we shouldn't drop support for python 3.9 until its actually at end of life (~10 months out).

desc/compute/_neoclassical_1D.py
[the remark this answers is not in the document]
unalmis: If the registered name includes "deprecated", they are excluded from the list of variables. I'm fine with placing them wherever

CHANGELOG.md
```diff
@@ -1,7 +1,18 @@
 Changelog
 =========
 
-New Feature
+New Features
+
+- Bounce integral methods with ``desc.integrals.Bounce1D`` and ``desc.integrals.Bounce2D``.
```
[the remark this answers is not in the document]
unalmis: Ok in that case, can you also take a look at the docstring for `Bounce1D` and 2d and let me know if you prefer to make anything a regular comment instead of a public doc

docs/notebooks/tutorials/EffectiveRipple.ipynb
```diff
@@ -0,0 +1,480 @@
+{
```
f0uriest, line 1: <p>should include a bit of intro/explanation before getting into the code. What is effective ripple? why do we care about it? How do we optimize it?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1290/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

f0uriest, line 1: <p>looks like the plot here is missing?</p><p> </p><p>Should also add some text after explaining what info we get from this plot and how its used later (ie, is it how you decided what to use for num_pitch etc below?)</p><p> </p><p>might also be useful to do it at deliberately too low resolution to show what a bad example looks like</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1290/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

f0uriest, line 1: <p>may want to reduce the resolution of the equilibrium itself to make this a bit faster. for the ballooning one I did (L,M,N)=(12,6,2), grids=(18,12,4)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1290/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

tests/test_compute_everything.py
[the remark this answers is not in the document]
unalmis: I would like to bunt this to #1094 if possible.

desc/objectives/_neoclassical.py
```diff
@@ -54,12 +66,15 @@ class EffectiveRipple(_Objective):
...
-        Unique coordinate values specifying field line labels to compute on.
-    batch : bool
-        Whether to vectorize part of the computation. Default is true.
+    grid : Grid
+        Optional, tensor-product grid in (ρ, θ, ζ) with uniformly spaced nodes
+        (θ, ζ) ∈ [0, 2π) × [0, 2π/NFP). Powers of two are preferable.
+    X : int
+        Grid resolution in poloidal direction for Clebsch coordinate grid.
+        Preferably power of 2.
+    Y : int
```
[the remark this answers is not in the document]
unalmis: I don't want ``num_zeta`` because people would ask me how that's different than ``grid.num_zeta``, whether it uses the latter as a default, and would associate the resolution with the fourier resolution rather than chebyshev or spline

desc/objectives/_neoclassical.py
```diff
@@ -54,12 +66,15 @@ class EffectiveRipple(_Objective):
...
-    rho : ndarray
-        Unique coordinate values specifying flux surfaces to compute on.
-    alpha : ndarray
-        Unique coordinate values specifying field line labels to compute on.
-    batch : bool
-        Whether to vectorize part of the computation. Default is true.
+    grid : Grid
+        Optional, tensor-product grid in (ρ, θ, ζ) with uniformly spaced nodes
+        (θ, ζ) ∈ [0, 2π) × [0, 2π/NFP). Powers of two are preferable.
+    X : int
```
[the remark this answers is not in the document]
unalmis: `num_alpha` would be confused for how many field lines the integration is started on

desc/compute/_neoclassical.py
```diff
@@ -68,87 +90,88 @@ def _compute(fun, fun_data, data, grid, num_pitch, simp=False, reduce=True):
...
+        fun_data[name] = Bounce2D.fourier(Bounce2D.reshape_data(grid, fun_data[name]))
+    fun_data["iota"] = grid.compress(data["iota"])
+    fun_data["theta"] = theta
+    fun_data["pitch_inv"], fun_data["pitch_inv weight"] = Bounce2D.get_pitch_inv_quad(
         grid.compress(data["min_tz |B|"]),
         grid.compress(data["max_tz |B|"]),
         num_pitch,
         simp=simp,
     )
+    return grid.expand(imap(fun, fun_data))
```
unalmis, line 114: Does calling `grid.expand` accordingly increase reverse mode AD memory?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-jax >= 0.4.24, != 0.4.36, <= 0.4.37
+jax >= 0.4.31, != 0.4.36, <= 0.4.37
```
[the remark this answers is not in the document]
unalmis: Yes, https://github.com/PlasmaControl/DESC/pull/1290#discussion_r1870384380

devtools/dev-requirements_conda.yml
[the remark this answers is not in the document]
dpanici: which issue is this?

desc/compute/_neoclassical.py
unalmis: Well you guys requested changes to move things around, so above not helpful anymore.

devtools/dev-requirements_conda.yml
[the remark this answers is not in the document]
unalmis: the one with "O(n) algorithm produces O(n^2)" in the title

docs/notebooks/tutorials/EffectiveRipple.ipynb
unalmis: better done on #1425

unalmis:
done.

desc/integrals/bounce_integral.py
```diff
@@ -208,25 +123,101 @@ class Bounce2D(Bounce):
...
       F decreases whereas C increases. Also, Q >> F and Q >> C.
 
-    Attributes
-    ----------
-    required_names : list
-        Names in ``data_index`` required to compute bounce integrals.
-
     """
 
+    # Notes
```
unalmis, line 139: image
if we want any of this in docstring please make a commit. also how to not show the interior of the quadrature array?

desc/compute/_neoclassical.py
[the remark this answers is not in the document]
unalmis: i added batching there too.

desc/compute/_neoclassical.py
unalmis: it is fine

desc/compute/_deprecated.py
```diff
@@ -0,0 +1,382 @@
...
+
+_bounce1D_doc = {
+    "num_well": _bounce_doc["num_well"],
+    "num_quad": _bounce_doc["num_quad"],
+    "num_pitch": _bounce_doc["num_pitch"],
+    "quad": _bounce_doc["quad"],
+}
+
+
+def _compute(fun, fun_data, data, grid, num_pitch, simp=False, reduce=True):
```
rahulgaur104, line 33: Does the AD chain go through each iteration of the loop when you do multiple alpha or rho values?

desc/compute/_deprecated.py
```diff
@@ -0,0 +1,382 @@
...
+    out = imap(foreach_rho, fun_data)
+    # Simple mean over α rather than integrating over α and dividing by 2π
+    # (i.e. f.T.dot(dα) / dα.sum()), because when the toroidal angle extends
+    # beyond one transit we need to weight all field lines uniformly, regardless
+    # of their spacing wrt α.
+    return grid.expand(out.mean(axis=0)) if reduce else out
+
+
+@register_compute_fun(
+    name="deprecated(effective ripple 3/2)",
```
rahulgaur104, line 77: Is that the best name? Seems pretty long.
How about old effective ripple 3/2?

desc/compute/_deprecated.py
```diff
@@ -0,0 +1,382 @@
...
+        },
+        data=data,
+        grid=grid,
+        num_pitch=num_pitch,
+        simp=False,
+    ) / (data["fieldline length"] * 2**1.5 * jnp.pi)
+    return data
+
+
+def _gbdrift(data, B, pitch):
```
rahulgaur104, line 306: We are calculating the numerator of the bounce integral.
Can we try a different name? Not important.

desc/compute/_fast_ion.py
```diff
@@ -0,0 +1,355 @@
...
+            Y_B,
+            num_transit,
+            quad=quad,
+            is_fourier=True,
+            spline=spline,
+        )
+
+        def fun(pitch_inv):
+            points = bounce.points(pitch_inv, num_well=num_well)
+            v_tau, drift1, drift2 = bounce.integrate(
```
rahulgaur104, line 165: I want to know how difficult it would be to calculate these quantities as a function of alpha, pitch_inv if my grid contains multiple alpha values. I need to vectorize these calculations in alpha for Gamma_alpha, Gamma_delta.

desc/compute/_fast_ion.py
```diff
@@ -0,0 +1,355 @@
...
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="r",
+    data=[
+        "min_tz |B|",
+        "max_tz |B|",
+        "cvdrift0",
+        "gbdrift (periodic)",
+        "gbdrift (secular)/phi",
```
rahulgaur104, line 257: Should we add a test to ensure that the secular gbdrift gives the same answer as periodic gbdrift as field line length is increased? Otherwise, it can be added later.

desc/compute/_neoclassical.py
```diff
@@ -68,85 +86,46 @@ def _compute(fun, fun_data, data, grid, num_pitch, simp=False, reduce=True):
...
-    resolution_requirement="z",
-    source_grid_requirement={"coordinates": "raz", "is_meshgrid": True},
-)
-def _fieldline_length(data, transforms, profiles, **kwargs):
-    grid = transforms["grid"].source_grid
-    L_ra = simpson(
-        y=grid.meshgrid_reshape(1 / data["B^zeta"], "arz"),
-        x=grid.compress(grid.nodes[:, 2], surface_label="zeta"),
-        axis=-1,
+def _dH(data, B, pitch):
```
rahulgaur104, line 116: I know it's called H in Nemov et al but can we add more letters/words to this. Otherwise, it may become a problem later.
Perhaps  _dH_ripple?
Not urgent.

desc/integrals/_bounce_utils.py
```diff
@@ -126,7 +126,7 @@ def _check_spline_shape(knots, g, dg_dz, pitch_inv=None):
 def bounce_points(
     pitch_inv, knots, B, dB_dz, num_well=None, check=False, plot=True, **kwargs
 ):
-    """Compute the bounce points given spline of |B| and pitch λ.
+    """Compute the bounce points given spline of B and pitch λ.
```
rahulgaur104, line 129: I am assuming the bounce points are calculated per field line so it's a 1D calculation, right?
Please add that to the description because I might play around with a 2D intersection idea later.

desc/integrals/_bounce_utils.py
```diff
@@ -710,46 +711,16 @@ def _get_extrema(knots, g, dg_dz, sentinel=jnp.nan):
...
-        & (ext[..., jnp.newaxis, :] < z2[..., jnp.newaxis]),
-        g_ext[..., jnp.newaxis, :],
-        upper_sentinel,
-    )
-
-
-def interp_to_argmin(
-    h, points, knots, g, dg_dz, method="cubic", beta=-100, upper_sentinel=1e2
-):
+def interp_to_argmin(h, points, knots, g, dg_dz, method="cubic"):
```
rahulgaur104, line 720: No suggestion here but can you explain what this function is doing?

rahulgaur104 (approved):
A bunch of minor points. I'll ask some more questions over the next hour.

tests/test_neoclassical.py
```diff
@@ -10,11 +10,73 @@
...
 
 
 @pytest.mark.unit
+@pytest.mark.slow
+@pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_1d)
+def test_effective_ripple_2D():
+    """Test effective ripple with W7-X against NEO."""
+    eq = get("W7-X")
+    rho = np.linspace(0, 1, 10)
+    grid = LinearGrid(rho=rho, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False)
```
rahulgaur104, line 25: How sensitive is the ripple to the grid resolution now?

rahulgaur104 (approved):

desc/integrals/_bounce_utils.py
unalmis: We interpolate the function `h` to the global minimum of `g` that lies between each bounce point. In our application $g = \vert B \vert$ and $h$ is either $\nabla \psi$ or $e_{\alpha}$ at fixed $\rho, \phi$.

desc/compute/_fast_ion.py
unalmis: It will not because the secular term drives the computation here to zero. We warn this in the objective docstring for now.

desc/compute/_neoclassical.py
unalmis: done in different PR

desc/integrals/_bounce_utils.py
unalmis: done in new PR

desc/compute/_fast_ion.py
unalmis: doing this now in different PR

desc/compute/_deprecated.py
unalmis: done in different PR

desc/compute/_deprecated.py
unalmis: I'll do it in a different PR

desc/compute/_deprecated.py
unalmis: How would I test this

desc/compute/_deprecated.py
dpanici: issue for this?

unalmis: PR will be done soon so no issue

Issue #1154, opened 2024-07-29
Improve coordinate mapping performance


unalmis:
Neoclassical stuff is bottlenecked by root finding. The PR that closes this issue will resolve this, and the benefits should assist any other objective that requires a coordinate mapping.

The performance of the coordinate mapping routine `map_clebsch_coordinates` can be further improved by using partial summation techniques and precomputing basis functions.

Notice $\rho$ and $\zeta$ are fixed throughout the root finding for $\vartheta \to \theta$ or $\alpha \to \theta$. Therefore,  one can condense the 3d map $\lambda \colon \rho, \theta, \zeta \mapsto \lambda(\rho, \theta, \zeta)$ to a set of 1d maps $\lambda \colon \rho, \theta, \zeta \mapsto \lambda_{\rho, \zeta}(\theta)$ by computing those parts of each Fourier Zernike basis function exterior to the rootfinding. This significantly reduces the computation to

$$\lambda_{\rho, \zeta}(\theta) = \sum_{\ell, m, n} a_{\ell,m,n} \phi_{\ell}(\theta)$$

Moreover, for rootfinding on a tensor-product grid in $\rho, \zeta$, one can further reduce evaluation to the inner product

$$\lambda_{\rho, \zeta}(\theta) = \sum_{\ell} b_{\ell} \phi_{\ell}(\theta)$$

or a single matrix product for all the $\rho, \zeta$.

Assuming $L = M = N$, this reduces the computation from $\mathcal{O}(N^6) \to \mathcal{O}(N^4)$ with the same constant of proportionality in the front if implemented correctly.

The literature mentions such partial summation techniques have achieved factors of 10000 improvement in runtime for $N = 100$. Given that those were actually compiled codes, I doubt that JIT is smart enough to find an algorithmic optimization like this.

Only benchmarks will show if we obtain the same benefit, but it should be wortwhile for us as this computation is in a Newton iteration. That means for us $N$ is proportional to the grid resolution times the number of iterations of the root finding.

For bounce integrals a typical resolution may be $X = 32$ and $Y = 64$.
So in that case, we have something like $N^2  \sim Y \times F \times \text{num iter}$ where $F$ is resolution of $\lambda$. Less computation means less memory consumed by AD to keep track of all the gradients, so this should also decrease memory in optimization.

f0uriest:
[...]

PR #1293, opened 2024-10-04
Make `map_coordinates` differentiable for JAX `0.4.34`

### LINKED ISSUE #1291: Error in `map_coordinates` with `jax==0.4.34`
Calling, for example plot_comparison in our basic equilibriim notebook yields a JAX error on 0.4.34

related to `_jacobi_jvp` which[ seems like we did something slightly](https://github.com/PlasmaControl/DESC/blob/e35436dbf4f49d3a4c438ed085ed2eca4f503e91/desc/basis.py#L1813) hacky to avoid escaped trace values, maybe not JAX does not like that anymore

```python
{
	"name": "TypeError",
	"message": "Called multiply with a float0 array. float0s do not support any operations by design because they are not compatible with non-trivial vector spaces. No implicit dtype conversion is done. You can use np.zeros_like(arr, dtype=np.float) to cast a float0 array to a regular zeros array. 
If you didn't expect to get a float0 you might have accidentally taken a gradient with respect to an integer argument.",
	"stack": "---------------------------------------------------------------------------
JaxStackTraceBeforeTransformation         Traceback (most recent call last)
File <frozen runpy>:198, in _run_module_as_main()

File <frozen runpy>:88, in _run_code()

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel_launcher.py:18
     16 from ipykernel import kernelapp as app
---> 18 app.launch_new_instance()

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/traitlets/config/application.py:1075, in launch_instance()
   1074 app.initialize(argv)
-> 1075 app.start()

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel/kernelapp.py:739, in start()
    738 try:
--> 739     self.io_loop.start()
    740 except KeyboardInterrupt:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/tornado/platform/asyncio.py:205, in start()
    204 def start(self) -> None:
--> 205     self.asyncio_loop.run_forever()

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/asyncio/base_events.py:618, in run_forever()
    617 while True:
--> 618     self._run_once()
    619     if self._stopping:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/asyncio/base_events.py:1951, in _run_once()
   1950     else:
-> 1951         handle._run()
   1952 handle = None

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/asyncio/events.py:84, in _run()
     83 try:
---> 84     self._context.run(self._callback, *self._args)
     85 except (SystemExit, KeyboardInterrupt):

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel/kernelbase.py:545, in dispatch_queue()
    544 try:
--> 545     await self.process_one()
    546 except Exception:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel/kernelbase.py:534, in process_one()
    533         return
--> 534 await dispatch(*args)

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel/kernelbase.py:437, in dispatch_shell()
    436     if inspect.isawaitable(result):
--> 437         await result
    438 except Exception:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel/ipkernel.py:362, in execute_request()
    361 self._associate_new_top_level_threads_with(parent_header)
--> 362 await super().execute_request(stream, ident, parent)

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel/kernelbase.py:778, in execute_request()
    777 if inspect.isawaitable(reply_content):
--> 778     reply_content = await reply_content
    780 # Flush output before sending the reply.

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel/ipkernel.py:449, in do_execute()
    448 if accepts_params[\"cell_id\"]:
--> 449     res = shell.run_cell(
    450         code,
    451         store_history=store_history,
    452         silent=silent,
    453         cell_id=cell_id,
    454     )
    455 else:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/ipykernel/zmqshell.py:549, in run_cell()
    548 self._last_traceback = None
--> 549 return super().run_cell(*args, **kwargs)

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/IPython/core/interactiveshell.py:3075, in run_cell()
   3074 try:
-> 3075     result = self._run_cell(
   3076         raw_cell, store_history, silent, shell_futures, cell_id
   3077     )
   3078 finally:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/IPython/core/interactiveshell.py:3130, in _run_cell()
   3129 try:
-> 3130     result = runner(coro)
   3131 except BaseException as e:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/IPython/core/async_helpers.py:128, in _pseudo_sync_runner()
    127 try:
--> 128     coro.send(None)
    129 except StopIteration as exc:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/IPython/core/interactiveshell.py:3334, in run_cell_async()
   3331 interactivity = \"none\" if silent else self.ast_node_interactivity
-> 3334 has_raised = await self.run_ast_nodes(code_ast.body, cell_name,
   3335        interactivity=interactivity, compiler=compiler, result=result)
   3337 self.last_execution_succeeded = not has_raised

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/IPython/core/interactiveshell.py:3517, in run_ast_nodes()
   3516     asy = compare(code)
-> 3517 if await self.run_code(code, result, async_=asy):
   3518     return True

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/IPython/core/interactiveshell.py:3577, in run_code()
   3576     else:
-> 3577         exec(code_obj, self.user_global_ns, self.user_ns)
   3578 finally:
   3579     # Reset our crash handler in place

Cell In[6], line 3
      2 print(\"Number of equilibria in the EquilibriaFamily:\", len(eq_fam))
----> 3 fig, ax = plot_comparison(
      4     eqs=[eq_fam[1], eq_fam[3], eq_fam[-1]],
      5     labels=[
      6         \"Axisymmetric w/o pressure\",
      7         \"Axisymmetric w/ pressure\",
      8         \"Nonaxisymmetric w/ pressure\",
      9     ],
     10 )

File ~/Research/DESC/desc/plotting.py:2388, in plot_comparison()
   2380 for i, eq in enumerate(eqs):
   2381     fig, ax, _plot_data = plot_surfaces(
   2382         eq,
   2383         rho,
   2384         theta,
   2385         phi,
   2386         ax,
   2387         theta_color=color[i % len(color)],
-> 2388         theta_ls=ls[i % len(ls)],
   2389         theta_lw=lw[i % len(lw)],
   2390         rho_color=color[i % len(color)],
   2391         rho_ls=ls[i % len(ls)],
   2392         rho_lw=lw[i % len(lw)],
   2393         lcfs_color=color[i % len(color)],
   2394         lcfs_ls=ls[i % len(ls)],
   2395         lcfs_lw=lw[i % len(lw)],
   2396         axis_color=color[i % len(color)],
   2397         axis_alpha=0,
   2398         axis_marker=\"o\",
   2399         axis_size=0,
   2400         label=labels[i % len(labels)],
   2401         title_fontsize=title_fontsize,
   2402         xlabel_fontsize=xlabel_fontsize,
   2403         ylabel_fontsize=ylabel_fontsize,
   2404         return_data=True,
   2405     )
   2406     for key in _plot_data.keys():

File ~/Research/DESC/desc/plotting.py:1685, in plot_surfaces()
   1676     tnr, tnt, tnz = t_grid.num_rho, t_grid.num_theta, t_grid.num_zeta
   1677     v_grid = Grid(
   1678         map_coordinates(
   1679             eq,
   1680             t_grid.nodes,
   1681             [\"rho\", \"theta_PEST\", \"phi\"],
   1682             [\"rho\", \"theta\", \"zeta\"],
   1683             period=(np.inf, 2 * np.pi, 2 * np.pi),
   1684             guess=t_grid.nodes,
-> 1685         ),
   1686         sort=False,
   1687     )
   1688 rows = np.floor(np.sqrt(nphi)).astype(int)

File ~/Research/DESC/desc/equilibrium/coords.py:218, in map_coordinates()
    215 # See description here
    216 # https://github.com/PlasmaControl/DESC/pull/504#discussion_r1194172532
    217 # except we make sure properly handle periodic coordinates.
--> 218 yk, (res, niter) = vecroot(yk, coords)
    220 out = compute(yk, outbasis)

File ~/Research/DESC/desc/equilibrium/coords.py:203, in map_coordinates.<locals>.<lambda>()
    199 yk = fixup(yk)
    201 vecroot = jit(
    202     vmap(
--> 203         lambda x0, *p: root(
    204             residual,
    205             x0,
    206             jac=jac,
    207             args=p,
    208             fixup=fixup,
    209             tol=tol,
    210             maxiter=maxiter,
    211             **kwargs,
    212         )
    213     )
    214 )
    215 # See description here
    216 # https://github.com/PlasmaControl/DESC/pull/504#discussion_r1194172532
    217 # except we make sure properly handle periodic coordinates.

File ~/Research/DESC/desc/backend.py:398, in root()
    396     return _lstsq(A, jnp.atleast_1d(y))
--> 398 x, (res, niter) = jax.lax.custom_root(
    399     res, x0, solve, tangent_solve, has_aux=True
    400 )
    401 return x, (safenorm(res), niter)

File ~/Research/DESC/desc/backend.py:344, in root.<locals>.<lambda>()
    342     jac2 = lambda x: jnp.atleast_2d(jac(x, *args))
--> 344 res = lambda x: jnp.atleast_1d(fun(x, *args)).flatten()
    346 # want to use least squares for rank-defficient systems, but
    347 # jnp.linalg.lstsq doesn't have JVP defined and is slower than needed
    348 # so we use the normal equations with regularized cholesky

File ~/Research/DESC/desc/equilibrium/coords.py:174, in residual()
    172 @jit
    173 def residual(y, coords):
--> 174     xk = compute(y, inbasis)
    175     return _fixup_residual(xk - coords, period)

File ~/Research/DESC/desc/equilibrium/coords.py:167, in compute()
    166     data[\"iota_rr\"] = profiles[\"iota\"].compute(grid, dr=2, params=params[\"i_l\"])
--> 167 transforms = get_transforms(basis, eq, grid, jitable=True)
    168 data = compute_fun(eq, basis, params, transforms, profiles, data)

File ~/Research/DESC/desc/backend.py:112, in wrapper()
    111 with jax.default_device(jax.devices(\"cpu\")[0]):
--> 112     return func(*args, **kwargs)

File ~/Research/DESC/desc/compute/utils.py:631, in get_transforms()
    630     if hasattr(t, \"build\"):
--> 631         t.build()
    633 return transforms

File ~/Research/DESC/desc/transform.py:385, in build()
    384     for d in self.derivatives:
--> 385         self.matrices[\"direct1\"][d[0]][d[1]][d[2]] = self.basis.evaluate(
    386             self.grid.nodes, d, unique=False
    387         )
    389 if self.method in [\"fft\", \"direct2\"]:

File ~/Research/DESC/desc/basis.py:1134, in evaluate()
   1132     n = n[nidx]
-> 1134 radial = zernike_radial(r[:, np.newaxis], lm[:, 0], lm[:, 1], dr=derivatives[0])
   1135 poloidal = fourier(t[:, np.newaxis], m, dt=derivatives[1])

File ~/Research/DESC/desc/basis.py:1533, in zernike_radial()
   1532 if dr == 0:
-> 1533     out = r**m * _jacobi(n, alpha, beta, jacobi_arg, 0)
   1534 elif dr == 1:

JaxStackTraceBeforeTransformation: TypeError: Called multiply with a float0 array. float0s do not support any operations by design because they are not compatible with non-trivial vector spaces. No implicit dtype conversion is done. You can use np.zeros_like(arr, dtype=np.float) to cast a float0 array to a regular zeros array. 
If you didn't expect to get a float0 you might have accidentally taken a gradient with respect to an integer argument.

The preceding stack trace is the source of the JAX operation that, once transformed by JAX, triggered the following exception.

--------------------

The above exception was the direct cause of the following exception:

Cell In[6], line 3
      1 eq_fam = desc.io.load(\"input.HELIOTRON_output.h5\")
      2 print(\"Number of equilibria in the EquilibriaFamily:\", len(eq_fam))
----> 3 fig, ax = plot_comparison(
      4     eqs=[eq_fam[1], eq_fam[3], eq_fam[-1]],
      5     labels=[
      6         \"Axisymmetric w/o pressure\",
      7         \"Axisymmetric w/ pressure\",
      8         \"Nonaxisymmetric w/ pressure\",
      9     ],
     10 )

File ~/Research/DESC/desc/plotting.py:2388, in plot_comparison(eqs, rho, theta, phi, ax, cmap, color, lw, ls, labels, return_data, **kwargs)
   2386     plot_data[string] = []
   2387 for i, eq in enumerate(eqs):
-> 2388     fig, ax, _plot_data = plot_surfaces(
   2389         eq,
   2390         rho,
   2391         theta,
   2392         phi,
   2393         ax,
   2394         theta_color=color[i % len(color)],
   2395         theta_ls=ls[i % len(ls)],
   2396         theta_lw=lw[i % len(lw)],
   2397         rho_color=color[i % len(color)],
   2398         rho_ls=ls[i % len(ls)],
   2399         rho_lw=lw[i % len(lw)],
   2400         lcfs_color=color[i % len(color)],
   2401         lcfs_ls=ls[i % len(ls)],
   2402         lcfs_lw=lw[i % len(lw)],
   2403         axis_color=color[i % len(color)],
   2404         axis_alpha=0,
   2405         axis_marker=\"o\",
   2406         axis_size=0,
   2407         label=labels[i % len(labels)],
   2408         title_fontsize=title_fontsize,
   2409         xlabel_fontsize=xlabel_fontsize,
   2410         ylabel_fontsize=ylabel_fontsize,
   2411         return_data=True,
   2412     )
   2413     for key in _plot_data.keys():
   2414         plot_data[key].append(_plot_data[key])

File ~/Research/DESC/desc/plotting.py:1685, in plot_surfaces(eq, rho, theta, phi, ax, return_data, **kwargs)
   1682     t_grid = _get_grid(**grid_kwargs)
   1683     tnr, tnt, tnz = t_grid.num_rho, t_grid.num_theta, t_grid.num_zeta
   1684     v_grid = Grid(
-> 1685         map_coordinates(
   1686             eq,
   1687             t_grid.nodes,
   1688             [\"rho\", \"theta_PEST\", \"phi\"],
   1689             [\"rho\", \"theta\", \"zeta\"],
   1690             period=(np.inf, 2 * np.pi, 2 * np.pi),
   1691             guess=t_grid.nodes,
   1692         ),
   1693         sort=False,
   1694     )
   1695 rows = np.floor(np.sqrt(nphi)).astype(int)
   1696 cols = np.ceil(nphi / rows).astype(int)

File ~/Research/DESC/desc/equilibrium/coords.py:218, in map_coordinates(eq, coords, inbasis, outbasis, guess, params, period, tol, maxiter, full_output, **kwargs)
    201 vecroot = jit(
    202     vmap(
    203         lambda x0, *p: root(
   (...)
    213     )
    214 )
    215 # See description here
    216 # https://github.com/PlasmaControl/DESC/pull/504#discussion_r1194172532
    217 # except we make sure properly handle periodic coordinates.
--> 218 yk, (res, niter) = vecroot(yk, coords)
    220 out = compute(yk, outbasis)
    221 if full_output:

    [... skipping hidden 14 frame]

File ~/Research/DESC/desc/equilibrium/coords.py:203, in map_coordinates.<locals>.<lambda>(x0, *p)
    197     yk = _initial_guess_nn_search(coords, inbasis, eq, period, compute)
    199 yk = fixup(yk)
    201 vecroot = jit(
    202     vmap(
--> 203         lambda x0, *p: root(
    204             residual,
    205             x0,
    206             jac=jac,
    207             args=p,
    208             fixup=fixup,
    209             tol=tol,
    210             maxiter=maxiter,
    211             **kwargs,
    212         )
    213     )
    214 )
    215 # See description here
    216 # https://github.com/PlasmaControl/DESC/pull/504#discussion_r1194172532
    217 # except we make sure properly handle periodic coordinates.
    218 yk, (res, niter) = vecroot(yk, coords)

File ~/Research/DESC/desc/backend.py:398, in root(fun, x0, jac, args, tol, maxiter, maxiter_ls, alpha, fixup)
    395     A = jnp.atleast_2d(jax.jacfwd(g)(y))
    396     return _lstsq(A, jnp.atleast_1d(y))
--> 398 x, (res, niter) = jax.lax.custom_root(
    399     res, x0, solve, tangent_solve, has_aux=True
    400 )
    401 return x, (safenorm(res), niter)

    [... skipping hidden 14 frame]

File ~/Research/DESC/desc/backend.py:344, in root.<locals>.<lambda>(x)
    341 else:
    342     jac2 = lambda x: jnp.atleast_2d(jac(x, *args))
--> 344 res = lambda x: jnp.atleast_1d(fun(x, *args)).flatten()
    346 # want to use least squares for rank-defficient systems, but
    347 # jnp.linalg.lstsq doesn't have JVP defined and is slower than needed
    348 # so we use the normal equations with regularized cholesky
    349 def _lstsq(a, b):

    [... skipping hidden 47 frame]

File ~/Research/DESC/desc/basis.py:1816, in _jacobi_jvp(x, xdot)
   1811 df = _jacobi(n, alpha, beta, x, dx + 1)
   1812 # in theory n, alpha, beta, dx aren't differentiable (they're integers)
   1813 # but marking them as non-diff argnums seems to cause escaped tracer values.
   1814 # probably a more elegant fix, but just setting those derivatives to zero seems
   1815 # to work fine.
-> 1816 return f, df * xdot + 0 * ndot + 0 * alphadot + 0 * betadot + 0 * dxdot

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/jax/_src/numpy/array_methods.py:1036, in _forward_operator_to_aval.<locals>.op(self, *args)
   1035 def op(self, *args):
-> 1036   return getattr(self.aval, f\"_{name}\")(self, *args)

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/jax/_src/numpy/array_methods.py:573, in _defer_to_unrecognized_arg.<locals>.deferring_binary_op(self, other)
    571 args = (other, self) if swap else (self, other)
    572 if isinstance(other, _accepted_binop_types):
--> 573   return binary_op(*args)
    574 # Note: don't use isinstance here, because we don't want to raise for
    575 # subclasses, e.g. NamedTuple objects that may override operators.
    576 if type(other) in _rejected_binop_types:

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/jax/_src/numpy/ufunc_api.py:177, in ufunc.__call__(self, out, where, *args)
    175   raise NotImplementedError(f\"where argument of {self}\")
    176 call = self.__static_props['call'] or self._call_vectorized
--> 177 return call(*args)

    [... skipping hidden 11 frame]

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/jax/_src/numpy/ufuncs.py:1142, in _multiply(x, y)
   1115 @partial(jit, inline=True)
   1116 def _multiply(x: ArrayLike, y: ArrayLike, /) -> Array:
   1117   \"\"\"Multiply two arrays element-wise.
   1118 
   1119   JAX implementation of :obj:`numpy.multiply`. This is a universal function,
   (...)
   1140     Array([ 0, 10, 20, 30], dtype=int32)
   1141   \"\"\"
-> 1142   x, y = promote_args(\"multiply\", x, y)
   1143   return lax.mul(x, y) if x.dtype != bool else lax.bitwise_and(x, y)

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/jax/_src/numpy/util.py:355, in promote_args(fun_name, *args)
    353 \"\"\"Convenience function to apply Numpy argument shape and dtype promotion.\"\"\"
    354 check_arraylike(fun_name, *args)
--> 355 _check_no_float0s(fun_name, *args)
    356 check_for_prngkeys(fun_name, *args)
    357 return promote_shapes(fun_name, *promote_dtypes(*args))

File ~/miniconda3/envs/desc-env-10-24/lib/python3.12/site-packages/jax/_src/numpy/util.py:326, in check_no_float0s(fun_name, *args)
    324 \"\"\"Check if none of the args have dtype float0.\"\"\"
    325 if any(dtypes.dtype(arg) == dtypes.float0 for arg in args):
--> 326   raise TypeError(
    327       f\"Called {fun_name} with a float0 array. \"
    328       \"float0s do not support any operations by design because they \"
    329       \"are not compatible with non-trivial vector spaces. No implicit dtype \"
    330       \"conversion is done. You can use np.zeros_like(arr, dtype=np.float) \"
    331       \"to cast a float0 array to a regular zeros array. \
\"
    332       \"If you didn't expect to get a float0 you might have accidentally \"
    333       \"taken a gradient with respect to an integer argument.\")

TypeError: Called multiply with a float0 array. float0s do not support any operations by design because they are not compatible with non-trivial vector spaces. No implicit dtype conversion is done. You can use np.zeros_like(arr, dtype=np.float) to cast a float0 array to a regular zeros array. 
If you didn't expect to get a float0 you might have accidentally taken a gradient with respect to an integer argument."
}
```

YigitElma:
- Adds `full_output` flags to `root` and `root_scalar` to make them differentiable.
- Adds tests for differentiability of `root` and `root_scalar`  in addition to `map_coordinates_derivative`

- While working on jax problems, I used this PR to update our `test_jax` workflow with new jax versions and better dependency installation routine (i.e. previously since jax was uploaded later, rest of the packages were latest and only jax was old, this was causing incompatibilities and false-errors)

Resolves #1291


YigitElma:
Check https://github.com/patrick-kidger/equinox/pull/871 for possible fix!

YigitElma:
@f0uriest Can you check if `desc.backend.root` is still working fine? I think it might be the cause of the problem,
```python
import jax.numpy as jnp
from desc.backend import root

def fun(x):
    return ((x-1)*(x**2+x+1))**2

def find_root_fun(x0):
    xk, _ = root(fun, x0)
    return xk

x0 = jnp.array([0., 0., 0.])
xk = find_root_fun(x0)

print("xk = ", xk)
print("fun(xk) = ", fun(xk))
jax.jacfwd(find_root_fun)(x0)
```
This gives exactly the same error message!

YigitElma:
https://github.com/jax-ml/jax/issues/24295

YigitElma:
I added flags `full_output` to `root` and `root_scalar` functions which solved the problem for now. If for debugging reasons, a user wants to get auxiliary outputs, it can be called by `full_output=True`, but this is not differentiable. Maybe a better way to implement is to add `custom_jvp` to those functions, and set the derivative of the auxiliary part to `float0` or something. 

Since the auxiliary outputs are not used in the objective calculations, it is fine for now.

Btw, this issue causes a problem only for the `BallooningStability` objective. I recommend adding where the specific version of `map_coordinates` (i.e. in_basis=(rho, alpha, zeta)) is used in this test as a comment.

YigitElma:
> Maybe a better way to implement is to add `custom_jvp` to those functions, and set the derivative of the auxiliary part to `float0` or something.

I mean the output with auxiliary output is `x, (res, niter)` and only the `niter` part causes trouble. We can write `custom_jvp` for multiple output like this https://stackoverflow.com/questions/77952302/jax-custom-vjp-for-functions-with-multiple-outputs. Or just basically change the type of niter to float.

YigitElma:
Check older versions of JAX, they seem to fail for `full_output=True` case.

desc/basis.py
f0uriest, line 1: are these changes necessary? I thought we found it was only an issue with root finding?

Also note that the explicit casts to float64 may make #1048 harder

requirements.txt
```diff
@@ -1,8 +1,8 @@
+jax[cpu] >= 0.3.2, < 0.5.0
```
f0uriest, line 1: minor point but can we keep this in alphabetical order?

Also don't we need to bump the minimum jax version?

requirements_conda.yml
```diff
@@ -15,8 +15,8 @@ dependencies:
   - pip:
     # Conda only parses a single list of pip requirements.
     # If two pip lists are given, all but the last list is skipped.
-    - interpax >= 0.3.3
     - jax[cpu] >= 0.3.2, < 0.5.0
```
f0uriest, line 18: see above

desc/backend.py
```diff
@@ -269,18 +273,25 @@ def bodyfun(state):
                 xk1, fk1 = backtrack(xk1, fk1, d)
                 return xk1, fk1, k1 + 1
 
-            state = guess, res(guess), 0
+            state = guess, res(guess), 0.0
```
f0uriest, line 276: is making this a float also required for it to be differentiable?

f0uriest (dismissed):

requirements.txt
YigitElma: Yes, we need to bump it up. Are we ok with `0.4.24`?

YigitElma: I changed this to make `test_jax` work better. In the alphabetical order (when jax comes later) pip was having trouble finding proper versions. 

I made some other changes after that, maybe this is not an issue anymore but basically this prioritizes `jax` version over others in a way

desc/basis.py
YigitElma: I will only leave the `nondiff_argnums` for `_jacobi` since I think might be usefull, I will revert the rest.

desc/basis.py
```diff
@@ -1800,17 +1800,17 @@ def _jacobi_body_fun(kk, d_p_a_b_x):
...
+def _jacobi_jvp(dx, x, xdot):
+    (n, alpha, beta, x) = x
+    (*_, xdot) = xdot
     f = _jacobi(n, alpha, beta, x, dx)
     df = _jacobi(n, alpha, beta, x, dx + 1)
     # in theory n, alpha, beta, dx aren't differentiable (they're integers)
     # but marking them as non-diff argnums seems to cause escaped tracer values.
     # probably a more elegant fix, but just setting those derivatives to zero seems
     # to work fine.
-    return f, df * xdot + 0 * ndot + 0 * alphadot + 0 * betadot + 0 * dxdot
```
YigitElma, line 1816: I deleted redundant 0 multiplications because in some cases, this gives an error saying `float0 cannot be used in math operations...`

desc/backend.py
YigitElma: So the changes I made prevent the problem in 2 different way. We don't need both of them implemented but kept them.
1. make `niter` float such that it wouldn't cause the issue
2. add `full_output` flag and call `map_coordinates` with `full_output=False`
Actually, 1 is enough to solve the problem fully. I implemented 2 first, idk maybe it helps with something else in future?!

requirements.txt
f0uriest: What are the issues with older versions of jax?

YigitElma: - `v0.4.23`-`v0.4.13` fails due to some scipy function
image
I don't know why pip doesn't choose proper scipy version, but it is probably due to `jax` requirements which I cannot fix.

- `v0.4.12` and older won't even past the pip install phase.

f0uriest: for the scipy one, if you manually then install the correct version of scipy does everything work? If so we can probably leave our jax requirement at 0.4.13, assuming that if people have an old version of jax installed then they also have an old version of scipy which should be compatible.

YigitElma: I didn't try that. Let me check

f0uriest: also I think we can remove the `[cpu]` flag for jax, by default it installs the cpu version and this would make it less annoying for installing gpu stuff

YigitElma: if we remove `[cpu]` than we need to add `jaxlib` to the requirements

f0uriest: I think all the recent versions install jaxlib automatically

YigitElma: btw, even if I install older `scipy` version, I get new errors
image
I just checked 0.4.13 locally because I need to add some logic to CI to test it. If you wanna test, feel free to do it here

dpanici:
0.4.24 as minimum

dpanici:
Update install docs with new jax cpu/gpu instruction

ddudt:
> Update install docs with new jax cpu/gpu instruction

When you update the Perlmutter installation instructions, we also need to correct the python version. The current instructions say to do `module load python` and then create the conda environment with `python=3.9`. But the default python version on Perlmutter is now 3.11, not 3.9, so we need to update the instructions to make this consistent. We should probably change it to:
```
module load python/3.11
conda create -n desc-env python=3.11
```

Also why do we have a different requirement on `scipy` specifically for the Perlmutter installation? Is that limitation still relevant with newer versions of python like 3.11? We should either remove those lines from the instructions if they are no longer necessary, or explain why Perlmutter is unique.

desc/basis.py
dpanici: I thought that was an issue before? but I guess if you ran it with the jax versions and this passes then this is fine

dpanici (changes requested):
looks good, just add to the install instruction that we dont need to remove the JAX from the requirements anymore, and update hte version info for the clusters (or at least for any that you've tested)

YigitElma:
> > Update install docs with new jax cpu/gpu instruction
> 
> When you update the Perlmutter installation instructions, we also need to correct the python version. The current instructions say to do `module load python` and then create the conda environment with `python=3.9`. But the default python version on Perlmutter is now 3.11, not 3.9, so we need to update the instructions to make this consistent. We should probably change it to:
> 
> ```
> module load python/3.11
> conda create -n desc-env python=3.11
> ```
> 
> Also why do we have a different requirement on `scipy` specifically for the Perlmutter installation? Is that limitation still relevant with newer versions of python like 3.11? We should either remove those lines from the instructions if they are no longer necessary, or explain why Perlmutter is unique.

I don't use `perlmutter`, can someone else change it and test it?

docs/installation.rst
```diff
@@ -184,22 +184,19 @@ check the link to install JAX with the most up-to-date recommendations from the
...
-    # then install as usual
     pip install --editable .
     # optionally install developer requirements (if you want to run tests)
     pip install -r devtools/dev-requirements.txt
 
-Tested and confirmed to work on the Della and Stellar clusters at Princeton as of June 20, 2024.
+Tested and confirmed to work on the Della and Stellar clusters at Princeton as of October 23, 2024.
 
 
 RAVEN (IPP, Germany)
```
YigitElma, line 202: RAVEN instructions use the jax version that has the `scipy` problem. If it works for them, we can bump up the JAX version and delete the messy scipy instructions

desc/basis.py
YigitElma: Ah no, the problem there was different. Back then we thought the problem was for `zernike_radial`, but it was `root` and `root_scalar`

devtools/dev-requirements.txt
```diff
@@ -1,6 +1,10 @@
 # standard install requirements
 -r ../requirements.txt
 
+# use notebooks on cluster, default env won't show-up on
```
dpanici, line 4: what do you mean by this line? that you need to call `ipykernel`  with the conda env if you want to use an environment with the jupyter?

this is not cluster specific btw, even on a local laptop, jupyter (and VS code jupyter) (in my experience) do not automatically detect my conda env and so I must install and use ipykernel to register it as an available kernel

docs/installation.rst
dpanici: I think maybe someone else (@Mandelbr0t or @IssraAli possibly?) had done this check of installation for us in the past? I don't believe I have access to RAVEN anymore

devtools/dev-requirements.txt
YigitElma: Yeah it is for jupyter. I was testing on 2 different  laptops simultenously and wrote some dummy comment. I will rephrase it. 

But yeah the idea to include this is, whenever you want to use your new environment, jupyter complains, for vs code it is straightforward but for clusters it is more of an issue to understand why it is not showing up. Either you need to select it before you create interactive session (show all environment even thought they dknt have ipykernel) or install it.

desc/equilibrium/coords.py
```diff
@@ -107,9 +107,8 @@ def map_coordinates(  # noqa: C901
 
     profiles = get_profiles(inbasis + basis_derivs, eq)
 
-    # TODO: make this work for permutations of in/out basis
-    if outbasis == ("rho", "theta", "zeta"):
-        if inbasis == ("rho", "alpha", "zeta"):
+    if set(outbasis) == {"rho", "theta", "zeta"}:
+        if set(inbasis) == {"rho", "alpha", "zeta"}:
```
unalmis, line 111: The if guards here shouldn't be changed unless the coordinate mapping routines are also generalized to account for this ordering.

Right now this will give wrong results unless the ordering on the left also matches the ordering on the right because the arrays are assumed to be [radial, poloidal, toroidal] coordinates.

desc/equilibrium/coords.py
```diff
@@ -128,7 +127,7 @@ def map_coordinates(  # noqa: C901
                 full_output=full_output,
                 **kwargs,
             )
-        if inbasis == ("rho", "theta_PEST", "zeta"):
+        if set(inbasis) == {"rho", "theta_PEST", "zeta"}:
```
unalmis, line 130: same issue as above

unalmis (changes requested):

docs/installation.rst
IssraAli: Rahul and I made those instructions for Raven in August and they are working for me. The jax version is 0.4.31. I could try today with a newer jax version if you'd like

YigitElma: @IssraAli I guess the instructions are outdated then. They install JAX `0.4.23` specifically. If you can use `0.4.31` then the instructions for `scipy` are redundant.

desc/equilibrium/coords.py
YigitElma: Hmm, yeah that makes sense. I will revert it then. Should we at least show a message to the user then? Like, `Given inbasis {inbasis} and outbasis {outbasis} combination has a better implementation of map_coordinates. But the order of elements must be given in a specific way. Until this issue is resolved, consider rearranging your inputs.`. 
The old version of the `test_map_coordinates_derivative` was using a different ordering, was that deliberate then?

YigitElma:
@rahulgaur104 will update the instructions for Perlmutter and RAVEN in a later PR.

desc/equilibrium/coords.py
unalmis: the 3d root finding routine was general, it's the 1d ones in those if statements that are not

YigitElma: Yes, the message I wrote is not an error but a suggestion to the user to use the better implementation. Anyway, someone can fix that `todo` when working on #1154 or some other related issue.

dpanici (approved):
Approving, make an issue to update installation instructions though if there is not already one and tag @IssraAli for the Raven one please (if he is willing to update that one)

devtools/dev-requirements.txt
```diff
@@ -1,6 +1,9 @@
 # standard install requirements
 -r ../requirements.txt
 
+# to use jupyter notebooks
+ipykernel
```
unalmis, line 5: Are you sure you want to add this? `ipykernel` may be installed in different ways depending on the user's development setup. Sometimes its managed separately through an older package `nb_conda_kernels` or the package manager `sudo dnf install jupyterlab`.

unalmis (approved):

devtools/dev-requirements.txt
YigitElma: Tbh, I don't know why this happens, but sometimes `ipykernel` is not installed with `pip install -r devtools/dev-requirements.txt`. However, most of the time it is installed without needing to add this line. When it is not installed VS-Code complain for notebooks using `desc-env`, and clusters don't even show the `desc-env` for interactive sessions (unless when you start you change this option 
image
)
I don't think this will cause any issue for other users. I added this just to prevent this kind of rare instances.

dpanici (changes requested):
don't include jupyter/ipykernel as a dependency.

Maybe add this note on ipykernel to the installation docs if you want

unalmis (approved):

f0uriest (approved):

devtools/dev-requirements_conda.yml
```diff
@@ -15,9 +15,9 @@ dependencies:
   - pip:
     # Conda only parses a single list of pip requirements.
     # If two pip lists are given, all but the last list is skipped.
+    - jax >= 0.4.24, < 0.5.0
```
ddudt, line 18: I prefer to keep these files in alphabetical order

requirements.txt
```diff
@@ -1,8 +1,8 @@
+jax >= 0.4.24, < 0.5.0
```
ddudt, line 1: see previous comment about alphabetical order

requirements_conda.yml
```diff
@@ -15,8 +15,8 @@ dependencies:
   - pip:
     # Conda only parses a single list of pip requirements.
     # If two pip lists are given, all but the last list is skipped.
+    - jax >= 0.4.24, < 0.5.0
```
ddudt, line 18: see previous comment about alphabetical order

tests/test_equilibrium.py
```diff
@@ -104,12 +103,39 @@ def test_map_coordinates_derivative():
     def foo(params, in_coords):
         out = eq.map_coordinates(
             in_coords,
-            inbasis,
-            outbasis,
+            ("rho", "alpha", "phi"),
```
ddudt, line 106: It would be better to reference `inbasis` here

ddudt (changes requested):

devtools/dev-requirements_conda.yml
YigitElma: See discussion here https://github.com/PlasmaControl/DESC/pull/1293#discussion_r1808063333. For general usage, you are right alphabetical order is better, but for `test_jax_versions` workflow having `jax` at the top and then installing the rest according to the `jax` version is helping `pip` a lot. Since many of our dependencies have `jax` dependency themselves, `pip` struggles with having the reverse order of installation.

YigitElma:
I added this to the change log, because after @dpanici's changes in #1327, we need to say this bug is fixed.

unalmis (commented):
#1333

f0uriest (approved):

unalmis (approved):

ddudt (approved):

PR #1296, opened 2024-10-09
Add objective for coil linking number

### LINKED ISSUE #1231: Add utility for calculating Linking Number of two given curves
useful for #1222 and for `FixSumCoilCurrents`

f0uriest:
- Adds a new objective for preventing interlinking coils by penalizing the linking number between each pair of coils.
- Adds a private method to calculate all linking numbers in a coilset (until we figure out a better API for stuff like that)

Note that while in theory the linking number is an integer and so isn't really differentiable, due to discretization of the integrals it actually can be fairly smooth:
image

The transition gets sharper as you increase the resolution as expected, though even at super high res where the gradient is always ~zero its still useful a useful objective within the line search or trust region subproblem to force the optimizer to reject steps that create linked coils.

Resolves #1231


desc/objectives/_coils.py
```diff
@@ -1362,3 +1362,116 @@ def compute(self, field_params=None, constants=None):
...
+        coilset = self.things[0]
+        grid = self._grid or LinearGrid(N=50)
+
+        self._dim_f = coilset.num_coils
+        self._constants = {"coilset": coilset, "grid": grid, "quad_weights": 1.0}
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params, constants=None):
+        """Compute minimum distances between coils.
```
dpanici, line 1455: docstring needs to be updated

dpanici (dismissed):

desc/coils.py
```diff
@@ -1369,25 +1383,37 @@ def _compute_position(self, params=None, grid=None, **kwargs):
...
             Coil positions, in [R,phi,Z] or [X,Y,Z] coordinates.
+        x_s : ndarray, shape(len(self),source_grid.num_nodes,3)
+            Coil position derivatives, in [R,phi,Z] or [X,Y,Z] coordinates.
+            Only returned if dx1=True.
 
         """
         basis = kwargs.pop("basis", "xyz")
+        keys = ["x"]
+        if dx1:
+            keys += ["x_s"]
```
ddudt, line 1401: Could consolidate this into a single line like you did above:
`keys = ["x", "x_s"] if dx1 else ["x"]`

desc/coils.py
```diff
@@ -1396,27 +1422,72 @@ def _compute_position(self, params=None, grid=None, **kwargs):
             )
             xyz_sym = xyz @ reflection_matrix(normal).T @ reflection_matrix([0, 0, 1]).T
             xyz = jnp.vstack((xyz, jnp.flipud(xyz_sym)))
+            if dx1:
```
ddudt, line 1425: You test it on a `MixedCoilSet`, which is never symmetric by construction, but it contains a symmetric `CoilSet` so I think these lines are supposed to be covered by the tests. This makes me think that the `dx1` argument is not being passed in correctly to the sub-coils of the MixedCoilSet

desc/coils.py
```diff
@@ -2331,15 +2407,17 @@ def _compute_position(self, params=None, grid=None, **kwargs):
...
-        x = jnp.vstack(
-            [
-                coil._compute_position(par, grd, **kwargs)
-                for coil, par, grd in zip(self.coils, params, grid)
-            ]
-        )
-        return x
+        out = []
+        for coil, par, grd in zip(self.coils, params, grid):
+            out.append(coil._compute_position(par, grd, dx1, **kwargs))
```
ddudt, line 2415: Check that this is correct, because I think the existing test should cover the logic of `sym=True` and `dx1=True` for `CoilSet._compute_position` (see previous comment).

desc/objectives/_coils.py
```diff
@@ -1362,3 +1362,118 @@ def compute(self, field_params=None, constants=None):
             )
 
         return Psi
+
+
+class CoilSetLinkingNumber(_Objective):
+    """Prevents coils from becoming interlinked.
+
+    The linking number of 2 curves is  (approximately) 0 if they are not linked, and
```
ddudt, line 1370: extra space between `is` and `(approximately)`

ddudt (changes requested):

ddudt:
Thanks for doing this Rory. My only real comment is that I think the test coverage should be 100%, and the fact that it is not makes me think something might be amiss.

desc/coils.py
f0uriest: Should be fixed now. The weirdness is because `CoilSet.from_symmetry` returns something with `sym=False` and `NFP=1`, because it doesn't use "virtual" coils, it manually duplicates and rotates the input coils. which makes me wonder if we should get rid of that? Since you get the same thing by just doing `CoilSet(coils, NFP=5, sym=True)` but its more efficient? Or maybe we just rename it to something like `CoilSet.materialize_virtual_coils` or something...

ddudt (approved):

desc/coils.py
ddudt: I think we can leave it as it. I believe the original design of `CoilSet.from_symmetry` was to be an easy way to initialize coils, but yes in general you shouldn't use it for symmetry. It could still be useful if you override the default to only rotate through a half field period, then create a symmetric `CoilSet` from those unique coils.

dpanici (approved):

PR #1300, opened 2024-10-14
Add degenerate row check to factorize linear constraints

### LINKED ISSUE #1297: Constraint degeneracy
When I have this in my optimisation:
```python
R_modes = np.vstack(
        (
            [0, 0, 0],
            eq.surface.R_basis.modes[
                np.max(np.abs(eq.surface.R_basis.modes), 1) > k, :
            ],
        )
    )
    Z_modes = eq.surface.Z_basis.modes[
        np.max(np.abs(eq.surface.Z_basis.modes), 1) > k, :
    ]
    # next we create the constraints, using the mode number arrays just created
    # if we didn't pass those in, it would fix all the modes (like for the profiles)
    constraints = (
        ForceBalance(eq=eq),
        # try to keep the aspect ratio between 7 and 9
        AspectRatio(eq=eq, bounds=(0, 10)),
        # similarly, try to keep it from getting too elongated
        Elongation(eq=eq, bounds=(0, 6)),
        # Keep volume the same as the initial volume
        #Volume(eq=eq0, target=eq0.compute("V")["V"]),
        # target for average iota
        RotationalTransform(eq=eq, target=1.1, loss_function="mean"),
        # fix major radius
        FixBoundaryR(eq=eq, modes=[0, 0, 0]),
        # fix vacuum profiles
        FixPressure(eq=eq),
        FixCurrent(eq=eq),
        FixPsi(eq=eq),
    )
    optimizer = Optimizer("lsq-auglag")
```

...there is a bug, because of FixBoundaryR(0,0,0) and the FixBoundaryR(modes), which both fix the (000) mode. This should presumably not bug out, I assume?

dpanici:
Resolves #1297 

Though what I found is that having degenerate constraints (like fixing [0,0,0] of Rb_mn with FixBoundaryR while also fixing that mode + others with another FixBoundaryR constraint) is only an issue when doing `lsq-auglag` and `lsq-exact` (when ForceBalance is not a constraint), not for `proximal-lsq-exact`... basically for when we aren't removing the eq DOFs from the system to solve. I did not dig into the real root but just found that removing the degenerate rows of the constraint matrix `A` worked to fix it, probably something to do with how the particular solution is assigned when there are degenerate rows

 @YigitElma there is likely a better fix than the one I chose which was simply to remove degenerate rows before we get into the while loop which finds the simple constraints and adjusts A,b accordingly, if you want feel free to make changes here

### DIFF
--- a/desc/objectives/utils.py
+++ b/desc/objectives/utils.py
@@ -92,6 +92,28 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
         A = A[:, cols]
     assert A.shape[1] == xp.size
 
+    # check for degenerate rows and delete if necessary
+    row_idx_to_delete = np.array([], dtype=int)
+    for row_idx in range(A.shape[0]):
+        # find all rows equal to this row
+        rows_equal_to_this_row = np.where(np.all(A[row_idx, :] == A, axis=1))[0]
+        # find the rows equal to this row that are not this row
+        rows_equal_to_this_row_but_not_this_row = rows_equal_to_this_row[
+            rows_equal_to_this_row != row_idx
+        ]
+        # if there are rows equal to this row that aren't this row, AND this particular
+        # row has not already been detected as a duplicate of an earlier one and slated
+        # for deletion, add the duplicate row indices to the array of
+        # rows to be deleted
+        if (
+            rows_equal_to_this_row_but_not_this_row.size
+            and row_idx not in row_idx_to_delete
+        ):
+            row_idx_to_delete = np.append(row_idx_to_delete, rows_equal_to_this_row[1:])
+    # delete the affected rows, and also the corresponding rows of b
+    A = np.delete(A, row_idx_to_delete, axis=0)
+    b = np.delete(b, row_idx_to_delete)
+
     # will store the global index of the unfixed rows, idx
     indices_row = np.arange(A.shape[0])
     indices_idx = np.arange(A.shape[1])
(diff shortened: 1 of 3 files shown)


desc/objectives/utils.py
```diff
@@ -92,6 +92,28 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
         A = A[:, cols]
     assert A.shape[1] == xp.size
 
+    # check for degenerate rows and delete if necessary
+    row_idx_to_delete = np.array([], dtype=int)
+    for row_idx in range(A.shape[0]):
+        # find all rows equal to this row
+        rows_equal_to_this_row = np.where(np.all(A[row_idx, :] == A, axis=1))[0]
```
f0uriest, line 99: I think you also need to check that b is the same for those rows, otherwise the constraints are incompatible. Could just add b as another column before doing this loop

dpanici: ah right, good catch

f0uriest (approved):

desc/objectives/utils.py
```diff
@@ -92,6 +92,36 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
...
+        )[0]
+        # find the rows equal to this row that are not this row
+        rows_equal_to_this_row_but_not_this_row = rows_equal_to_this_row[
+            rows_equal_to_this_row != row_idx
+        ]
+        # if there are rows equal to this row that aren't this row, AND this particular
+        # row has not already been detected as a duplicate of an earlier one and slated
+        # for deletion, add the duplicate row indices to the array of
+        # rows to be deleted
+        if (
```
YigitElma, line 115: With this logic do we prevent deleting both of the constraints? Let's say 2nd row and 6th row of augmented A are the same, then when we run the for loop at second row, we will add 6 and for 6th row, we will add 2? So, both of them are deleted? This would probably cause a failing test but I want to be sure.

YigitElma: Oh I see, it takes this into account

PR #1305, opened 2024-10-16
Add missing unit test markers

dpanici:
- Adds missing unit test labels for some tests.
- Adds a precommit hook that checks tests folder and warns the user if a test without mark is added

### DIFF
--- a/devtools/check_unmarked_tests.sh
+++ b/devtools/check_unmarked_tests.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+unmarked=$(pytest tests/ --collect-only -m "not unit and not regression" -q | head -n -2) 2> /dev/null
+num_unmarked=$(echo "$unmarked"  | sed '/^\s*$/d' | wc -l)
+
+if [ $num_unmarked -gt 0 ]
+then
+    echo "----found unmarked tests----"
+    echo "$unmarked"
+    exit 1
+fi
--- a/devtools/check_unmarked_tests.sh
+++ b/devtools/check_unmarked_tests.sh
@@ -1,11 +1,22 @@
 #!/bin/sh
 
-unmarked=$(pytest tests/ --collect-only -m "not unit and not regression" -q | head -n -2) 2> /dev/null
-num_unmarked=$(echo "$unmarked"  | sed '/^\s*$/d' | wc -l)
+# Start the timer using date (in seconds since epoch)
+start_time=$(date +%s)
 
-if [ $num_unmarked -gt 0 ]
-then
+echo "Files to check: $@"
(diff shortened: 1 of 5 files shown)


unalmis (dismissed):

rahulgaur104 (dismissed):

f0uriest:
We should add a pre commit check to make sure this doesn't keep happening

YigitElma:
> We should add a pre commit check to make sure this doesn't keep happening

I can try to add this in this PR or some other one

devtools/check_unmarked_tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+unmarked=$(pytest tests/ --collect-only -m "not unit and not regression" -q | head -n -2) 2> /dev/null
```
YigitElma, line 3: This will check every file in tests folder every time we add a commit, right? I guess there is a way to just check the changed files. 

```bash
#!/bin/bash

# Start the timer
start_time=$SECONDS

# Initialize an empty variable to accumulate unmarked tests
unmarked=""

# Loop through each provided file
for file in "$@"; do
  # Check if the file is a Python file in the tests directory
  if [[ "$file" == tests/*.py ]]; then
    echo "checking tests in $file"
    # Collect unmarked tests for the specific file
    file_unmarked=$(pytest "$file" --collect-only -m "not unit and not regression" -q | head -n -2) 2> /dev/null
    if [[ -n "$file_unmarked" ]]; then
      unmarked+="$file_unmarked"$'\n'
    fi
  fi
done

# Count the number of unmarked tests found
num_unmarked=$(echo "$unmarked" | sed '/^\s*$/d' | wc -l)

# Calculate the elapsed time
elapsed_time=$(($SECONDS - $start_time))
# If there are any unmarked tests, print them and exit with status 1
if [ "$num_unmarked" -gt 0 ]; then
    echo "Time taken: ${elapsed_time} seconds"
    echo "----found unmarked tests----"
    echo "$unmarked"
    exit 1
fi
```

Also, need to set `pass_filenames` to `true` in config yaml

f0uriest: yeah but most of the time is spent on pytest init, regardless of how many files are changed, so checking everything isn't any more expensive

YigitElma: Yes, but this change reduced it from 8 seconds to 4 seconds on my laptop.

f0uriest: feel free to change it if you want, i don't have any strong preference.

f0uriest (approved):

unalmis (approved):

Issue #1169, opened 2024-08-06
finite-build coils


ddudt:
We need to decide on a class structure for finite-build coils. There are several categories of coil classes that we might want to support: 

- filamentary vs finite-build coils
- circular vs rectangular cross-sections
- different curve parameterizations (Fourier, Spline, Planar, etc.)

Possible ideas we discussed:

1. Add new width/height attributes to the existing `_Coil` class, which could default to 0/None. Non-planar coils will also have to add a winding angle attribute. Then each coil class could have it's own implementation for finite-build stuff. This prevents needing to add any new classes, but is maybe not the cleanest object-oriented design. 
2. Add a new `_FiniteBuildCoil` class. Then there would be new classes that inherit from this and the existing coil classes. This keeps the filamentary coils separate, but adds a lot of new classes to keep track of. 
3. Add a new `_CrossSection` ABC with circular and rectangular options. Similar to the previous option, new coil classes will inherit from these new classes and the existing coil classes. This gives a lot of flexibility, but might be excessive. 

Other ideas?

ryanwu4:
The finite build coil computation would be be an implementation of these papers by Hurtwitz et. al and Landreman et. al: 

- [Circular cross section finite build paper](https://arxiv.org/pdf/2310.09313)
- [Rectangular cross section finite build paper](https://arxiv.org/pdf/2310.12087)

An existing Julia implementation of the finite build rectangular cross section method is [here](https://github.com/landreman/CoilForces.jl).

dpanici:
- this will need a new optimizable param for `geom` which is `r` for circular and `L` `W`
 - i.e. FourierPlanarCoil -> FourierPlanarFiniteBuildCoil
   -  maybe try doing `FourierPlanarFiniteBuildCoil( _FiniteBuildCoil, FourierPlanarCoil)` and see if it works
- basically just copy the way we have compute_magnetic_field setup but rename `compute_self_force` etc and define new formulas in `coils.py` as needed 
  - need a `cross_section="circular"` or "rectangular" argument

dpanici:
@ryanwu4 @ddudt We should have a solution for the twist angle as well, so we don't implement the single thing for planar coils

ryanwu4:
[...]

PR #1311, opened 2024-10-19
Small Miscellaneous coil fixes

dpanici:
- adds setup method for coils, so when loading attributes will be set to None if they do not exist.
- add missing name to docstring of `FourierXYZCoil`


dpanici:
This might not be needed once we decide on #1169 actually, if we have some sort of nfilaments thing there...

dpanici:
Better to do a subclass of curve instead of coil for stuff like Rogowski etc

desc/coils.py
```diff
@@ -238,6 +252,7 @@ def current(self):
 
     @current.setter
     def current(self, new):
+        new = setdefault(new, 1, new)
```
ddudt, line 255: What is the purpose of this? Wouldn't this prevent someone from setting the current to 0, since 0 = False and so it would get replaced with the default value of 1?

desc/coils.py
```diff
@@ -689,8 +704,6 @@ def from_values(cls, current, coords, N=10, NFP=1, basis="rpz", sym=False, name=
         sym : bool
             Whether to enforce stellarator symmetry.
         name : str
-            name for this coil
```
ddudt, line 692: Why delete this? I think it should still say ` Name for this coil.` like we have for all the similar methods of other coil classes.

desc/coils.py
```diff
@@ -834,7 +849,7 @@ class FourierPlanarCoil(_Coil, FourierPlanarCurve):
     basis : {'xyz', 'rpz'}
         Coordinate system for center and normal vectors. Default = 'xyz'.
     name : str
-        Name for this coil.
+        name for this coil
```
ddudt, line 852: I prefer proper punctuation in our doc strings, I think we do that in most cases

ddudt (changes requested):

desc/coils.py
ddudt: I think you might have meant to do `new = setdefault(new, 1)`

dpanici: check if can set current to 0

desc/coils.py
dpanici: we are pretty inconsistent even in this file

YigitElma (approved):

f0uriest (approved):

ddudt (approved):

PR #1315, opened 2024-10-22
Ensure numpy arrays are not modified in place

unalmis:


f0uriest (approved):

YigitElma (approved):
Is this related to any issue? Or just to make numpy arrays behave the same as jax arrays?

unalmis:
yea it came up in a different pull request review. for people who don't have jax installed, there's likely somewhere in the code where these changes due qualify as a bugfix, but our tests will never catch it because they use jax

YigitElma (approved):

f0uriest (approved):

PR #1320, opened 2024-10-24
Add initial implementation of `ShareParameters`

dpanici:


tests/test_linear_objectives.py
```diff
@@ -1085,3 +1087,72 @@ def test_linear_objective_from_user_on_collection(DummyCoilSet):
...
+    np.testing.assert_allclose(subobj.target, 0)
+
+    # check compute
+    np.testing.assert_allclose(obj.compute_unscaled(obj.x(eq1, eq2, eq3, eq4)), 0)
+
+    # check the jacobian
+    J = obj.jac_unscaled(obj.x(eq1, eq2, eq3, eq4))
+    # make sure Jacobian is not trivial
+    assert not np.allclose(J, 0)
+    # now, check that each row sums to zero, and abs(J) rows sum to 1,
```
f0uriest, line 1117: Sum to 2

f0uriest (changes requested):

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,170 @@ def update_target(self, thing):
...
+        thing = self.things[0]
+
+        # default params
+        default_params = tree_map(lambda dim: np.arange(dim), thing.dimensions)
+        self._params = setdefault(self._params, default_params)
+        self._params = broadcast_tree(self._params, default_params)
+        self._indices = tree_leaves(self._params)
+        assert tree_structure(self._params) == tree_structure(default_params)
+
+        self._dim_f = sum(idx.size for idx in self._indices) * (len(self.things) - 1)
```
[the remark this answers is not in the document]
dpanici: I think so as all of my tests just use `True` for the param. 
The `default_params` and `setdefault` in this build method are what make sure `self._indices` becomes the full arange or not if it is True or False

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,170 @@ def update_target(self, thing):
...
 
+class ShareParameters(_Objective):
+    """Fix specific degrees of freedom to be the same between Optimizable things.
+
+    Parameters
+    ----------
+    things : list of Optimizable
+        list of objects whose degrees of freedom are being fixed to eachother's values.
+        Must be at least length 2, but may be of arbitrary length.
+        Every object must be of the same type, and have the same size array for the
```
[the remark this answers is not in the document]
dpanici: yea but then I thought of all the ways this could mess up, like `Equilibrium.R_lmn` and `FourierRZToroidalSurface.R_lmn` are the same name, yet even if they are the same shape it makes no sense to have them be shared.

I can change it if you think that is not a good enough counterpoint not to, but is there an example where it makes sense to allow this?

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,170 @@ def update_target(self, thing):
...
+        default_params = tree_map(lambda dim: np.arange(dim), thing.dimensions)
+        self._params = setdefault(self._params, default_params)
+        self._params = broadcast_tree(self._params, default_params)
+        self._indices = tree_leaves(self._params)
+        assert tree_structure(self._params) == tree_structure(default_params)
+
+        self._dim_f = sum(idx.size for idx in self._indices) * (len(self.things) - 1)
+
+        # default target
+        self.target = np.zeros(self._dim_f)
```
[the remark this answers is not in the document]
dpanici: should I make sure that target does not get passed in ever? or is not including it in docstring enough to say that one should never pass it in? I worry someone will put a target of like what they wanna fix to and that will mess the objective up, as it would not really be caught anywhere (if the target is say a single number, as it would be broadcastable to the dim_f size)

YigitElma:
This would help a lot for multiple GPU stuff. When I put different `ForceBalance` objectives to different GPUs, optimizer thinks there are multiple different equilibrium (because jax.device_put() creates a copy of the equilibrium, so although they are the same at that moment, their memory address is different). I tried to prevent this by overriding the `obj.things[0] = eq_replicated_to_all_devices` but Jax complains about this saying that inputs of the jitted function should live on the same device (since we close on self, different portions of the objective live on different device). This PR would help to just use different equilibrium but link their parameters.

dpanici:
Another issue that I know the reason now but am unsure of the resolution, tho it is an edge case: for things like `normal` where we normalize the vector when the attribute is set, if I for instance do `ShareParameters([coil1, coil2], {"normal":True})` everything works perfect, if I do though `ShareParameters([coil1, coil2], {"normal":np.array([1])})` to only make the index-1 compenent of the normal the same, the constraint works correctly, but when the attributes are then set at the end of the optimization, they are normalized in the setter, and the 0th and 2nd components may differ (as they were not shared), and thus the resulting normalized vector may have a different 1-component. 

This is sort of a useless edge case, but still unsure of how to fully resolve a case like this.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ New Features
 ------------
 - New basis vector and metric elements derivatives in PEST coordinates and quantities useful for a global MHD stability solver.
 - Adds keyword argument `normalize` to plot_1d, plot_3d. `normalize` is a string to use for normalization.
+- Adds new linear objective `ShareParameters` which can enforce that two or more objects of the same type remain identical during an optimization. Potentially useful for flexible stellarator optimization, where one has two coilsets with the same geometry but differing currents, and attempts to optimize for two different stellarator equilibria.
```
YigitElma: Maybe instead of remain identical, say the chosen parameters of 2 or more objects remain identical ....

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,183 @@ def update_target(self, thing):
...
+            Shared degree of freedom errors.
+
+        """
+        # basically, just subtract the first things' params
+        # and every subsequent thing (adding on rows to dim_f if
+        # more than 2 total things) so that the Jacobian of this
+        # ends up being just rows with 1 in the first object's params
+        # indices and -1 in the second objects params indices,
+        # repeated vertically for each additional object
+        # i.e. for a a size-1 param being shared among 4 objects, the
```
YigitElma:
```suggestion
        # i.e. for a size-1 param being shared among 4 objects, the
```

tests/test_linear_objectives.py
```diff
@@ -1085,6 +1087,78 @@ def test_linear_objective_from_user_on_collection(DummyCoilSet):
...
+    assert J.shape[0] == subobj.dim_f
+    # make sure Jacobian is not trivial
+    assert not np.allclose(J, 0)
+    # now, check that each row sums to zero, and abs(J) rows sum to 2,
+    # meaning each row has only 2 nonzero elements which are 1 and -1,
+    J_row_sums = J.sum(axis=1)
+    abs_J_row_sums = np.abs(J).sum(axis=1)
+    np.testing.assert_allclose(abs_J_row_sums, 2)
+    np.testing.assert_allclose(J_row_sums, 0)
+    subobj = ShareParameters([eq1, eq2, eq3, eq4], {"p_l": [1, 3]})
```
YigitElma: What is the rest for?

dpanici: not needed, I will delete

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,183 @@ def update_target(self, thing):
...
+        Must be at least length 2, but may be of arbitrary length.
+        Every object must be of the same type, and have the same size array for the
+        desired parameter to be fixed (e.g. same geometric resolution if fixing
+         ``R_lmn``, or same pressure profile resolution if fixing ``p_l``)
+    params : nested list of dicts
+        Dict keys are the names of parameters to fix (str), and dict values are the
+        indices to fix for each corresponding parameter (int array).
+        Use True (False) instead of an int array to fix all (none) of the indices
+        for that parameter.
+        Must have the same pytree structure as thing.params_dict.
```
f0uriest: do you mean `things[0].params_dict`? ie if I'm fixing 2 coils should `params` be
`{"X_n": True}` or `[{"X_n": True}, {"X_n": True}]`

I assume the former, since I'm not sure how to parse the 2nd one if the keys are different. 

And the type for this: `nested list of dicts` I think in most cases it should just be a single dict? the only case where it would be a list is if `things` is an `OptimizeableCollection`?

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,183 @@ def update_target(self, thing):
...
+
+    def __init__(
+        self,
+        things,
+        params=None,
+        target=0,
+        bounds=None,
+        weight=1,
+        normalize=True,
+        normalize_target=True,
```
f0uriest: Doesn't this objective only make sense if `target=0`? so do we want these to be options? I think we can just remove them from the API and hard code the values passed to `super().__init__`

f0uriest: We do the same thing for the `*SelfConsistency` objectives

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,183 @@ def update_target(self, thing):
...
+        name="shared parameters",
+    ):
+        self._params = params
+        assert len(things) > 1, "only makes sense for >1 thing"
+        assert np.all([isinstance(things[0], type(t)) for t in things[1:]])
```
f0uriest: would be good to include an error message here that says which thing doesn't match, like
`expected things to all be the same type, got types {type(t) for t in things}`

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,183 @@ def update_target(self, thing):
...
+
+        # ensure things are the same resolution
+        # TODO: might be too strict? could we only try to ensure
+        #  that the desired params passed are same res?
+        for t in things[1:]:
+            assert np.all([things[0].dimensions == t.dimensions])
```
f0uriest: same thing, would be good to include an informative error message

desc/objectives/objective_funs.py
```diff
@@ -247,6 +247,7 @@ class ObjectiveFunction(IOAble):
         "_name",
         "_things_per_objective_idx",
         "_use_jit",
+        "_static_attrs",
```
f0uriest, line 250: do you know why this is needed now? I don't see anything that should change how static attrs get parsed

f0uriest:
Also don't forget to add the new objective to the docs

desc/objectives/linear_objectives.py
dpanici: Yea, specifically something like `MixedCoilSet` where you need to go into the nested structure.

for all else it can be just one dict, so I will change this, thanks

desc/objectives/objective_funs.py
dpanici: I don't know tbh, but after #1662 was merged, this then became an issue. Yigit had similar issue

> @dpanici I had a similar error in multidevice PR and I ended up adding "_static_attrs" to the "_static_attrs" list. For both objective function and _Objective. I also couldn't find the exact reason, but that seemed to solve the problem

It is specifcially for the case I mentioned in my comment above, where if you build subobj then obj it throws the error.

But if you run this similar code on master, no issue occurs:

```python
import numpy as np
from desc.coils import CoilSet, FourierXYZCoil
from desc.objectives import (
    ObjectiveFunction,
    FixParameters,
)

coils1 = CoilSet.linspaced_angular(FourierXYZCoil(), n=2)
coils2 = coils1.copy()

subobj = FixParameters([coils1, coils2], {"X_n": True, "Y_n": True, "Z_n": [2]})
subobj.build()
obj = ObjectiveFunction(subobj)
obj.build()

```

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,191 @@ def update_target(self, thing):
...
+        Use True (False) instead of an int array to fix all (none) of the indices
+        for that parameter.
+        Must have the same pytree structure as things[0].params_dict.
+        The default is to fix all indices of all parameters.
+    name : str, optional
+        Name of the objective function.
+
+    Examples
+    --------
+    .. code-block:: python
```
YigitElma, line 281: You might need a blank line after this

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,191 @@ def update_target(self, thing):
...
+        name="shared parameters",
+    ):
+        self._params = params
+        assert len(things) > 1, "only makes sense for >1 thing"
+        assert np.all(
+            [isinstance(things[0], type(t)) for t in things[1:]]
+        ), f"expected same type for all things, got types {[type(t) for t in things]}"
+
+        # ensure things are the same resolution
+        # TODO: might be too strict? could we only try to ensure
```
YigitElma: ~~I am thinking of something recursive like this~~
```python
def check_list(dim1, dim2, param_list):
    for d1, d2, p in zip(dim1, dim2, param_list):
        if isinstance(p, dict):
            for key, val in p.items():
                assert d1[key] == d2[key], (
                    f"expected same dimensions for {key}, but {d2[key]} != {d1[key]}. "
                    + "Make sure that each thing is at the same resolution."
                )
        elif isinstance(p, list):
            check_list(d1, d2, p)

for dim1, dim2, param in zip(t1.dimensions, t2.dimensions, params):
    if isinstance(param, dict):
        for key, val in param.items():
            assert dim1[key] == dim2[key], (
                f"expected same dimensions for {key}, but {dim2[key]} != {dim1[key]}. "
                + "Make sure that each thing is at the same resolution."
            )
    elif isinstance(param, list):
        check_list(dim1, dim2, param)
```

~~Not sure if we can use some kind of `jax.tree_util` because in general `params` doesn't have the same pytree structure as `thing.dimensions`.~~

~~Or we can just have a skip check option?~~

YigitElma: Actually forget about top one, if you move the check to build method, it is simpler,
```python
...
assert tree_structure(self._params) == tree_structure(default_params)

def look(d1, d2, p):
    # p.size is 0 if user didn't pass them in params, they shouldn't cause failure
    if p.size == 0:
        return True
    else:
        assert d1 == d2, "At least one dimension that is being fixed does not match!"
        return d1==d2

_= tree_map(look, t1.dimensions, t2.dimensions, params)
```

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,200 @@ def update_target(self, thing):
...
+            if p.size == 0:
+                return True
+            else:
+                assert (
+                    d1 == d2
+                ), "At least one dimension that is being fixed does not match!"
+                return d1 == d2
+
+        [
+            tree_map(look, thing.dimensions, t2.dimensions, self._params)
```
YigitElma: nitpick but why not just a for loop, it is even shorter

tests/test_linear_objectives.py
```diff
@@ -1085,6 +1087,74 @@ def test_linear_objective_from_user_on_collection(DummyCoilSet):
...
 
 
+@pytest.mark.unit
+def test_share_parameters_four_objects():
+    """Tests ShareParameters with 4 objects."""
+    eq1 = desc.examples.get("SOLOVEV")
+    eq2 = eq1.copy()
+    eq3 = eq1.copy()
+    eq4 = eq1.copy()
+
```
YigitElma, line 1097: if you want to test the logic for the resolutions, you can add 
```python
eq5 = eq1.copy()
eq5.pressure.change_resolution(eq.L+1)
# different pressure resolution shouldn't cause error if not fixed
s0 = ShareParameters([eq1, eq5], {"i_l": [1, 2]})
with pytest.raises(SssertionError):
    s0 = ShareParameters([eq1, eq5], {"p_l": [1, 2]})
    s0.build()
```

YigitElma (approved):
Both comments are minor suggestions, approving anyway.

tests/test_linear_objectives.py
dpanici: I think I cover this in the coil test when I do 

```python
    coils2[-1].change_resolution(N=xyz_coil2.N + 1)
   params = [
        [  # share the "current" of the 1st TF coil and 1st component of "center"
            {"current": True, "center": np.array([0])},
            # share "center" and  "normal" for the 2nd TF coil
            {"center": True, "normal": True},
            {"r_n": True},  # share radius of the 3rd TF coil
            {"r_n": False},  # dont share the r_n of 4th tf coil
        ],
        # share "shift" & "rotmat" for all VF coils
        {"shift": True, "rotmat": True},
        # share specified indices of "X_n" and "Z_n",
        # but not "Y_n", for other coil
        {"X_n": np.array([1, 2]), "Y_n": False, "Z_n": np.array([0])},
        {"current": True},  # only current for the last coil, which has different N in
        # second coilset
    ]
```

desc/objectives/linear_objectives.py
dpanici: this was your code originally haha I just copied it from your [comment](https://github.com/PlasmaControl/DESC/pull/1320#discussion_r2421716307)

YigitElma: No, I mean instead of 
```python
[
    tree_map(look, thing.dimensions, t2.dimensions, self._params)
    for t2 in self.things[1:]
]
```
Just do
```python
for t2 in self.things[1:]:
    tree_map(look, thing.dimensions, t2.dimensions, self._params)
```

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,200 @@ def update_target(self, thing):
...
+            for t2 in self.things[1:]
+        ]
+
+        self._dim_f = sum(idx.size for idx in self._indices) * (len(self.things) - 1)
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, *params, constants=None):
+        """Compute fixed degree of freedom errors.
```
rahulgaur104: Maybe make it "Compute shared degree of freedom errors."

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,200 @@ def update_target(self, thing):
...
+        def look(d1, d2, p):
+            # p.size is 0 if user didn't pass them in as params to share,
+            # they shouldn't cause failure
+            if p.size == 0:
+                return True
+            else:
+                assert (
+                    d1 == d2
+                ), "At least one dimension that is being fixed does not match!"
+                return d1 == d2
```
rahulgaur104, line 388: else:
      assert (
          d1 == d2
      ), f"Shared-parameter dimension mismatch: expected {d1}, got {d2}"
      return True

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +236,200 @@ def update_target(self, thing):
...
+        # ends up being just rows with 1 in the first object's params
+        # indices and -1 in the second objects params indices,
+        # repeated vertically for each additional object
+        # i.e. for a size-1 param being shared among 4 objects, the
+        # Jacobian looks like
+        #  [ 1 -1  0  0]
+        #  [ 1 0  -1  0]
+        #  [ 1 0   0 -1]
+        params_1 = params[0]
+        return jnp.concatenate(
```
rahulgaur104, line 432: This looks like your are making a reference list multiple times and then subtracting stuff from that list multiple times.
Wouldn't it be effective to make the reference list only once?

```python3
# build reference vector once
ref_vec = jnp.concatenate([
    jnp.atleast_1d(param[idx])
    for param, idx in zip(tree_leaves(params[0]), self._indices)
])

# build each comparison vector once, then subtract
diffs = []
for p in params[1:]:
    vec = jnp.concatenate([
        jnp.atleast_1d(param[idx])
        for param, idx in zip(tree_leaves(p), self._indices)
    ])
    diffs.append(ref_vec - vec)
return jnp.concatenate(diffs)
```

desc/objectives/linear_objectives.py
dpanici: oh sure, yea I can do that

desc/objectives/linear_objectives.py
dpanici: updated message

desc/objectives/linear_objectives.py
dpanici: true, I only make the reference array once now

desc/objectives/linear_objectives.py
```diff
@@ -236,6 +243,206 @@ def update_target(self, thing):
...
+        Every object must be of the same type, and have the same size array for the
+        desired parameter to be fixed (e.g. same geometric resolution if fixing
+        ``R_lmn``, or same pressure profile resolution if fixing ``p_l``)
+    params : dict
+        Dict keys are the names of parameters to fix (str), and dict values are the
+        indices to fix for each corresponding parameter (int array).
+        Use True (False) instead of an int array to fix all (none) of the indices
+        for that parameter.
+        Must have the same pytree structure as things[0].params_dict.
+        The default is to fix all indices of all parameters.
```
f0uriest, line 263: so this means even if I only want to fix one parameter, I still have to pass a dict with `"param":False` for all the other ones? If only ~one parameter is being shared it would probably be easier to just leave out the unshared ones which may also be less error prone in case people forget all the params that a given object has. Though if fixing many params maybe its better to always be explicit in case people forget one that should be fixed?

YigitElma: `Must have the same pytree structure as things[0].params_dict.` is not very clear but no, you can just specify which one to fix and the rest will be free. The default mentioned here is if you don't give any `params`.

```python
params = broadcast_tree(params, default_params)
```

Sets the missing params with empty arrays.

YigitElma: For example, the example code in docs has
```python
params = [
    [ # share the "current" of the 1st TF coil and 1st center component
        {"current": True, "center":np.array([0])},
        # share "center" and  "normal" for the 2nd TF coil
        {"center": True, "normal": True},
        {"r_n": True},  # share radius of the 3rd TF coil
        {},  # share nothing in the 4th TF coil
    ],
    # share "shift" & "rotmat" for all VF coils
    {"shift": True, "rotmat": True},
    # share specified indices of "X_n" and "Z_n",
    # but not "Y_n", for other coil
    {"X_n": np.array([1, 2]), "Y_n": False, "Z_n": np.array([0])},
]
```
After build with some broadcasting it becomes,
```python
self._params = [
    [
        {
            "current": array([0]),
            "center": array([0]),
            "normal": array([], dtype=int64),
            "r_n": array([], dtype=int64),
            "rotmat": array([], dtype=int64),
            "shift": array([], dtype=int64),
        },
        {
            "center": array([0, 1, 2]),
            "normal": array([0, 1, 2]),
            "current": array([], dtype=int64),
            "r_n": array([], dtype=int64),
            "rotmat": array([], dtype=int64),
            "shift": array([], dtype=int64),
        },
        {
            "r_n": array([0]),
            "center": array([], dtype=int64),
            "current": array([], dtype=int64),
            "normal": array([], dtype=int64),
            "rotmat": array([], dtype=int64),
            "shift": array([], dtype=int64),
        },
        {
            "center": array([], dtype=int64),
            "current": array([], dtype=int64),
            "normal": array([], dtype=int64),
            "r_n": array([], dtype=int64),
            "rotmat": array([], dtype=int64),
            "shift": array([], dtype=int64),
        },
    ],
    [
        {
            "shift": array([0, 1, 2]),
            "rotmat": array([0, 1, 2, 3, 4, 5, 6, 7, 8]),
            "R_n": array([], dtype=int64),
            "Z_n": array([], dtype=int64),
            "current": array([], dtype=int64),
        },
        {
            "shift": array([0, 1, 2]),
            "rotmat": array([0, 1, 2, 3, 4, 5, 6, 7, 8]),
            "R_n": array([], dtype=int64),
            "Z_n": array([], dtype=int64),
            "current": array([], dtype=int64),
        },
        {
            "shift": array([0, 1, 2]),
            "rotmat": array([0, 1, 2, 3, 4, 5, 6, 7, 8]),
            "R_n": array([], dtype=int64),
            "Z_n": array([], dtype=int64),
            "current": array([], dtype=int64),
        },
    ],
    {
        "X_n": array([1, 2]),
        "Y_n": array([], dtype=int64),
        "Z_n": array([0]),
        "current": array([], dtype=int64),
        "rotmat": array([], dtype=int64),
        "shift": array([], dtype=int64),
    },
]
```

dpanici: For the record, the default is only if no params is passed. Sort of like we do in FixParameters

PR #1323, opened 2024-10-24
Fix grid L, M, N attributes from being set incorrectly in certain cases

### LINKED ISSUE #1313: Full grid is overriden to symmetric grid
When I make a compute function with
```python
resolution_requirement="tz",
grid_requirement={"coordinates": "rtz", "is_meshgrid": True, "sym": False}
```
and use `eq.compute` for that function, the full grid I supply is overridden to a symmetric one. The compute function catches this error, complaining that a full grid is required.

In general, I should not need to specify `override_grid=False` here because I am fine with the grid being overriden when computing dependencies.

### LINKED ISSUE #1329: Automatically computing on grid with required resolution fails for grids made under jit
Right now the following code is called when making a grid under JIT.
```python
self._L = self.num_nodes
self._M = self.num_nodes
self._N = self.num_nodes
```
When we call ``eq.compute`` this will usually pass the following check, even if it shouldn't
```python
if grid.L >= self.L_grid and grid.M >= self.M_grid and grid.N >= self.N_grid:
    # then don't automatically compute on grid with required resolution
```

unalmis:
Also resolves #1313 and resolves #1329

### DIFF
--- a/desc/equilibrium/equilibrium.py
+++ b/desc/equilibrium/equilibrium.py
@@ -1234,7 +1234,7 @@ class Equilibrium(IOAble, Optimizable):
             **kwargs,
         )
 
-    def _get_rtz_grid(
+    def get_rtz_grid(
         self,
         radial,
         poloidal,
--- a/desc/equilibrium/equilibrium.py
+++ b/desc/equilibrium/equilibrium.py
@@ -861,6 +861,7 @@ class Equilibrium(IOAble, Optimizable):
                 "v": "theta_PEST",
                 "a": "alpha",
                 "z": "zeta",
+                "p": "phi",
             }
             rtz_nodes = self.map_coordinates(
                 grid.nodes,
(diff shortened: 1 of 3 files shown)


dpanici:
This is specifically for when rho theta zeta are passed in?

desc/equilibrium/equilibrium.py
```diff
@@ -990,23 +991,23 @@ def need_src(name):
...
+                    msg("poloidal") + f" got M_grid={grid.M} < {self._M_grid}.",
                 )
                 warnif(
                     # if need more toroidal resolution
                     "z" in req and grid.N < self.N_grid
                     # and won't override grid to one with more toroidal resolution
                     and not (override_grid and coords in {"r", ""}),
                     ResolutionWarning,
-                    msg("toroidal"),
+                    msg("radial") + f" got N_grid={grid.N} < {self._N_grid}.",
```
f0uriest, line 1010: msg Should be toroidal

desc/grid.py
```diff
@@ -187,17 +187,25 @@ def _scale_weights(self):
...
+            self._L = self.num_rho - 1
+        return self._L
 
     @property
     def M(self):
         """int: Poloidal grid resolution."""
-        return self.__dict__.setdefault("_M", 0)
+        if getattr(self, "_M", None) is None:
+            self._M = (
+                (self.num_theta // 2 - 1) if self.sym else (self.num_theta - 1) // 2
```
f0uriest, line 199: I think these are reversed? If sym is true then there are fewer nodes for a given M

unalmis: The logic in create nodes is the opposite. Due to grid spacing stuff iirc
```python
theta = 2 * (M + 1) if self.sym else 2 * M + 1
```

desc/grid.py
```diff
@@ -187,17 +187,25 @@ def _scale_weights(self):
     @property
     def L(self):
         """int: Radial grid resolution."""
-        return self.__dict__.setdefault("_L", 0)
+        if getattr(self, "_L", None) is None:
+            self._L = self.num_rho - 1
+        return self._L
```
unalmis, line 192: i used convention for linear grid. quadrature and concentric grid always define these attributes, so they won't get overwritten.

if it's preferred to move this logic do lineargrid only by overriding these property functions, feel free to

desc/grid.py
dpanici: With this way rn,

```python
g = LinearGrid(theta=3,rho=1,sym=True)
```
results in `g.M` being `0`, which feels odd since it does have some sort of theta resolution, it is not just a single point.

desc/grid.py
```diff
@@ -283,7 +295,7 @@ def unique_rho_idx(self):
         errorif(
             not hasattr(self, "_unique_rho_idx"),
             AttributeError,
-            f"{self} does not have unique indices assigned. "
+            "Grid does not have unique indices assigned. "
```
f0uriest, line 298: I kind of prefer having it print `self` in case there are multiple grids so its clear which one its referring to

unalmis: The issue is that calls the print / to string method of the grid object, which calls the L, M, N attributes, which calls this method again, and you get stuck recursing.

dpanici (approved):

PR #1326, opened 2024-10-27
Fix bug in PlasmaVesselDistance signed distance error checks 

dpanici:
when the grids have differing number of zeta points,  np.allclose errors out instead of the correct error passing. This fixes the bug


YigitElma:
Can this be the longest branch name lol

dpanici:
> Can this be the longest branch name lol

there's no character limit, it's fine. Just approve it

rahulgaur104 (approved):

PR #1327, opened 2024-10-28
Implement `from_input_file` for `Equilibrium`, update docs with this info

### LINKED ISSUE #1298: VMEC inputs
Clearer instructions to load a vmec input file (not from the command line, and without specifying e.g. profiles in code) would be useful:)

dpanici:
resolves #1298


tests/test_input_output.py
```diff
@@ -226,6 +226,73 @@ def test_near_axis_input_files():
...
+
+@pytest.mark.unit
+def test_from_input_file_equilibrium_desc_vmec():
+    """Test that from_input_file gives same eq for DESC and VMEC input files."""
+    vmec_path = ".//tests//inputs//input.QSC_r2_5.5_vmec"
+    desc_path = ".//tests//inputs//input.QSC_r2_5.5_desc"
+    kwargs = {"L": 10, "M": 10, "N": 14}
+    eq = Equilibrium.from_input_file(desc_path, **kwargs)
+    # change surface resolution to match the one from after thresholding of VMEC's bdry
+    eq.surface.change_resolution(M=10, N=14)
```
YigitElma, line 275: Is this necessary?

docs/api_equilibrium.rst
```diff
@@ -16,6 +16,17 @@ An ``EquilibriaFamily`` is a ``list`` like object for storing multiple equilibri
     desc.equilibrium.Equilibrium
     desc.equilibrium.EquilibriaFamily
 
+The ``Equilibrium`` class may be instantiated in a couple of ways in addition to providing inputs to its constructor.
+- from an existing DESC or VMEC input file with its ``from_input_file`` method
+- from a ``pyQSC`` ``Qsc``  or ``pyQIC`` ``Qic`` near-axis equilibrium with the ``Equilibrium``'s ``from_near_axis`` method
+
+.. autosummary::
+    :toctree: _api/equilibrium
+    :recursive:
```
YigitElma, line 25:
```suggestion
    :recursive:
    :no-index:
```
may solve the docs problem

docs/api_equilibrium.rst
```diff
@@ -35,6 +46,15 @@ the magnetic axis, cross section, and various space curves.
     desc.geometry.SplineXYZCurve
     desc.geometry.ZernikeRZToroidalSection
 
+The ``FourierRZToroidalSurface`` and the ``FourierRZCurve`` classes may be instantiated from an existing DESC or VMEC input file with their ``from_input_file`` method.
+
+.. autosummary::
+   :toctree: _api/geometry/
+   :recursive:
```
YigitElma, line 53: same as above

tests/inputs/input.DSHAPE_desc
```diff
@@ -1,11 +1,12 @@
 # This DESC input file was auto generated from the VMEC input file
-# /DESC/tests/inputs/input.DSHAPE
+# .//tests//inputs//input.DSHAPE
+# on 10/27/2024 at 20:00:46.
```
YigitElma, line 3: with this additional line, you may need to change the compared lines in `test_vmec_input`

tests/test_input_output.py
dpanici: yes, otherwise the DESC eq has N=75 and the VMEC one has N=14 and we cannot compare. When we load in a VMEC input file we have this threshold option for the bdry modes which drops any modes with an amplitude less than threshold (which maybe I should add to both the DESC/VMEC one tbh and have as a kwarg here...)

dpanici:
Should make it no threshold by default

tests/test_input_output.py
dpanici, line 275: check why this is needed, it should be made self consistent in the eq init

CHANGELOG.md
f0uriest, line 1: can you move this to its own PR?

desc/equilibrium/equilibrium.py
```diff
@@ -2015,6 +2017,57 @@ def from_near_axis(
...
+            Equilibrium being created.
+
+        Returns
+        -------
+        Equilibrium : Equilibrium
+            Equilibrium generated from the given input file.
+
+        """
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
f0uriest, line 2039: won't this ignore possibly important warnings if using a vmec input with incompatible options?

f0uriest: we ignore them in the `FourierRZToroidalSurface.from_input_file` because we know we only care about the boundary shape. But for the full equilibrium we also care about profiles, flux, etc which vmec has a lot of options for

f0uriest (changes requested):

desc/equilibrium/equilibrium.py
dpanici: Good point

YigitElma (approved):
The non-zero l modes in the input file warning can be added in some other PR. Seems fine to me

desc/equilibrium/equilibrium.py
```diff
@@ -2015,6 +2017,54 @@ def from_near_axis(
...
+            inputs = InputReader().parse_inputs(path)[-1]
+        if (inputs["bdry_ratio"] is not None) and (inputs["bdry_ratio"] != 1):
+            warnings.warn(
+                "boundary_ratio = {} != 1, surface may not be as expected".format(
+                    inputs["bdry_ratio"]
+                )
+            )
+        inputs["surface"][:, 1:3] = inputs["surface"][:, 1:3].astype(int)
```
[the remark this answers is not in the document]
f0uriest: `equilibrium.utils.parse_surface` ensures that the surface array either has all `l==0` or all `n==0`:
https://github.com/PlasmaControl/DESC/blob/f5abd60cbc56dab4bd5a8c9f18378725f5cb769e/desc/equilibrium/utils.py#L62

f0uriest (approved):

PR #1337, opened 2024-10-30
Update CHANGELOG

dpanici:


YigitElma (approved):

f0uriest (dismissed):

f0uriest:
We could think about adding a GH action that checks if the changelog has been updated on each PR (with maybe some flag we can set to ignore it when needed)

f0uriest (approved):

PR #1338, opened 2024-10-30
Run Unmarked Test pre-commit only if all other hooks pass

dpanici:
the unmarked test pre-commit hook (at least on my laptop) can take anywhere from 5-10 seconds to run, so I'd rather it only run if everything else is passing.


.pre-commit-config.yaml
```diff
@@ -3,6 +3,7 @@ repos:
     rev: 24.3.0
     hooks:
       - id: black-jupyter
+        log_file: devtools/pre-commit.log
```
dpanici, line 6: this log is only created if black fails

devtools/clean_logfiles.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+if [ -f "devtools/pre-commit.log" ]; then
+   echo "delete devtools/pre-commit.log"
```
YigitElma, line 4:
```suggestion
   echo "deleting devtools/pre-commit.log file"
```

YigitElma (approved):

YigitElma (approved):

.pre-commit-config.yaml
f0uriest: if it fails on the first pass, the run again does it delete the old log file? or does it just overwrite it with a blank one?

YigitElma: I think the last one deletes it at the end of each run.

f0uriest (approved):

PR #1349, opened 2024-11-09
Umbilic fourier bounce

rahulgaur104:
This is a dummy  PR  that combines fourier bounce with rg/NFP_fac so we can find umbilic equilibria with good omnigenity. 
The purpose right now is to see if I can do such an optimization. The optimization throws a cryptic error when I combine combine umbilic with ripple (spline field.)


dpanici:
can we close this?

PR #1353, opened 2024-11-11
Fix softmin and softmax implementation

### LINKED ISSUE #1317: Remove entropy term from desc `softmax`
[Updated issue](https://github.com/PlasmaControl/DESC/issues/1317#issuecomment-2434352292)

I a while back implemented softmax [here](https://github.com/PlasmaControl/DESC/blob/d138a4990d7a1c4d50e0f9e781c43b5d24c16b02/desc/objectives/utils.py#L285)

but I realize we also have [since grabbed the one from `jax.nn`](https://github.com/PlasmaControl/DESC/blob/d138a4990d7a1c4d50e0f9e781c43b5d24c16b02/desc/backend.py#L74)

We should probably use just the one from `jax.nn` assuming they are interchangable

dpanici:
Resolves #1317


desc/objectives/utils.py
```diff
@@ -288,11 +288,6 @@ def softmax(arr, alpha):
     Inspired by https://www.johndcook.com/blog/2010/01/13/soft-maximum/
     and https://www.johndcook.com/blog/2010/01/20/how-to-compute-the-soft-maximum/
```
unalmis, line 289: please remove this

f0uriest (approved):

PR #1345, opened 2024-11-04
Add support for benchmark test from forks

YigitElma:
Benchmark action requires write access for commenting which forks don't have. This change makes the benchmark workflow coming from a fork run on our CI and thus have proper access. 

This shouldn't change the general permissions for fork repos, so someone in the DESC team must still approve to test to run for such PRs.

https://github.com/thollander/actions-comment-pull-request?tab=readme-ov-file#permissions

- [ ] Make sure this is secure! Check from a fork repo!

PR #1356, opened 2024-11-11
Fix Extra Benchmark and cancelation 

YigitElma:
For some reason, changes from #1345 try to initialize 2 separate benchmarks for a single PR and cancel the second one with `Canceling since a higher priority waiting request for 'Benchmarks-1344' exists`. The second one seems to run properly but this is still annoying.


dpanici (approved):

f0uriest (approved):

YigitElma:
Can someone merge this? @dpanici @f0uriest

PR #1358, opened 2024-11-12
Add action for checking changelog updates

### LINKED ISSUE #1340: Add GH action that checks if the changelog has been updated on each PR
>We could think about adding a GH action that checks if the changelog has been updated >on each PR (with maybe some flag we can set to ignore it when needed)

_Originally posted by @f0uriest in https://github.com/PlasmaControl/DESC/issues/1337#issuecomment-2448695957_

f0uriest:
Adds a GH action that ensures `CHANGELOG.md` is updated with each PR, unless the PR is labelled `skip_changelog`

Resolves #1340

### DIFF
--- a/.github/workflows/changelog_update.yml
+++ b/.github/workflows/changelog_update.yml
@@ -1,0 +1,19 @@
+name: Check changelog updated
+
+on:
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  check_changelog_updated:
+    if: ${{contains(github.event.pull_request.labels.*.name, 'skip_changelog') == false}}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: danieljimeneznz/ensure-files-changed@v4.1.0
+        with:
+          require-changes-to: |
+            CHANGELOG.md
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog_update.yml
```diff
@@ -0,0 +1,19 @@
...
+    branches:
+      - master
+    types: [opened, synchronize, labeled, unlabeled]
+
+jobs:
+  check_changelog_updated:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip_changelog') }}
```
YigitElma, line 14: Adding filter for changed files could save time here
```
- name: Filter changes
        id: changes
        uses: dorny/paths-filter@v3
        with:
          filters: |
            has_changes:
              - 'desc/**'
```
For the changelog, we don't want to include the changes in other folders anyway.

YigitElma (changes requested):

.github/workflows/changelog_update.yml
f0uriest: this whole action takes a few seconds so don't think its really worth it.

YigitElma: By time, I mean adding the flag. If this test runs for every PR and fails, then that small red cross could be annoying :)

YigitElma (approved):

dpanici (approved):

Issue #1208, opened 2024-08-20
Add tests for optimization subproblms


unalmis:
Excerpt:
We should add unit tests which check that these deterministic subproblems are solved correctly.

Right now, I only see optimization test for strictly convex functions in `test_optimizer`. A basic gradient descent would work for that; in general stuff could go wrong in more specialized methods that would not be detected by minimizing a convex function.

_Originally posted by @unalmis in https://github.com/PlasmaControl/DESC/pull/1181#pullrequestreview-2235943780_

unalmis:
Could be cause of #1894

PR #1360, opened 2024-11-13
non-singular Laplace BIE solver, improved FFT interpolation for singular integral, and new magnetic field API

### LINKED ISSUE #890: Better FFT interpolation for singular integration
FFT interpolation can't handle non-symmetric grids. We will probably always need the source grid to be symmetric to capture the integral correctly, but in theory the evaluation grid could be symmetric if we get the interpolation to work correctly.

_Originally posted by @ddudt in https://github.com/PlasmaControl/DESC/pull/725#discussion_r1495256689_

### LINKED ISSUE #1807: New magnetic field API
This issue documents the new API for magnetic field computations that I discussed at the last developer meeting. The other developers agreed to it. The current API makes analysis difficult, especially for quantities that are derived from the magnetic field, and slows feature development.

The new API in #1805 will register computable quantities in the same manner that the equilibrium class does. The old magnetic field API for existing classes will be updated to the new ones in a later PR.

### LINKED ISSUE #1819: document tangential assumption of current potential classes
These magnetic field classes assume that grad Phi is tangential which is only true in special cases. I mentioned  this last fall but we have yet to merge the changes . This issue is to document this for others.

unalmis:
PR Age

All the solvers are thoroughly tested. See the shared paper to explain this PR.

This is the only open source JAX AD compatible BIE solver.

The robustness of the free surface optimization is outside scope of this pull request. See #1894 and #1208 for more details. There are plenty of tests that illustrate correctness, and tuning optimization hyper parameter defaults which is problem dependent anyway should not delay merging this PR.

This PR includes my FFT improvements significantly improve free surface optimization (both old and new) and anything that involves singular integrals.

The quadrature in master has been observed to have unfortunately slow convergence for some test cases. It has been shown the improved quadrature in the linked paper will converge faster especially for the non singular kernel of this system. It can be added in a different PR.


unalmis:
I'll push the simpler method we talked about. Regardless it is not expected for the current method to not satisfy the lcfs condition. 

Am I missing something in the implementation? If you are available, but would like to be caught up on the code or math let me know.

tests/test_integrals.py
```diff
@@ -667,38 +696,346 @@ def test_singular_integral_vac_estell(self, interpolator):
...
+
+        Phi_mn, Phi_transform = compute_Phi_mn(
+            eq=eq,
+            B0n=B0n,
+            eval_grid=grid,
+            source_grid=grid,
+            check=False,
+            chunk_size=chunk_size,
+        )
+        assert np.isfinite(Phi_mn).all()
```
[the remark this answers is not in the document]
dpanici: try with finite diff, vmap ID matrix, or build the pinv yourself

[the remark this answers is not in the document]
unalmis: vmap id matrix is better than building pinv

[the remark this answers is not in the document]
unalmis: Changing to avoid the Jacobian still had nan. After more debugging, the `nan` was another JAX bug. Once the matrix is large enough, JAX just gives up and fills linear algebra solve with nan. I fixed this by chunking the computation, just like in #1522 .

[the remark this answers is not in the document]
unalmis: Will push fixed commit soon.

tests/test_integrals.py
```diff
@@ -667,38 +695,334 @@ def test_singular_integral_vac_estell(self, interpolator):
...
+            eval_grid=eval_grid,
+            source_grid=source_grid,
+            Phi_mn=Phi_mn,
+            basis=Phi_transform.basis,
+            chunk_size=chunk_size,
+            warn_fft=False,
+        )
+        Bn = B0n + dPhi_dn
+        # FIXME: This fails to converge to zero, only converges to 0.03.
+        np.testing.assert_allclose(Bn, 0, atol=0.03)
```
unalmis, line 755: @f0uriest Can you push the axissymetric bugfix to the anisotropy branch? Then we can merge it to this branch. Maybe that will fix the issue.

Here you can see above that `Phi` is computed correctly, but the result does not converge to zero. This is just applying the shielding theorem.

desc/integrals/_laplace.py
```diff
@@ -0,0 +1,572 @@
...
+        _kernel_biot_savart,
+        interpolator,
+        chunk_size,
+    )
+    # Biot-Savart kernel assumes Φ in amperes, so we account for that.
+    return -dot(bs, evl_data["n_rho"]) * 2 / mu_0
+
+
+# TODO: remove below stuff before merge.
+# alternative methods to check for agreement
```
unalmis, line 333: can ignore below

tests/test_integrals.py
unalmis: It didn't change anything, but thank you.

tests/test_integrals.py
```diff
@@ -667,38 +692,334 @@ def test_singular_integral_vac_estell(self, interpolator):
...
+            eq=eq,
+            eval_grid=eval_grid,
+            source_grid=source_grid,
+            Phi_mn=Phi_mn,
+            basis=Phi_transform.basis,
+            chunk_size=chunk_size,
+            warn_fft=False,
+        )
+        Bn = B0n + dPhi_dn
+        # FIXME: This fails to converge to zero, only converges to 0.03.
```
unalmis, line 751: Analytically, we know that $\nabla \Phi \cdot n = - sin(\theta)$. (You can also write the integral done by `compute_dPhi_dn` analytically and it comes out to this). However the numerical result converges to $0.03 + sin(\theta)$.

The only part of the numerical computation that is not performed analytically is the singular integral and the quadrature for the Fourier harmonics of $\Phi$ (what `jnp.linalg.solve` is basically doing under the hood). The latter is shown to converge to the correct result, so I can only conclude that the integration of the kernel is struggling. But there is no implementation bug in the kernel integration.

desc/integrals/_laplace.py
```diff
@@ -0,0 +1,401 @@
...
+        ----------
+        chunk_size : int or None
+            Size to split computation into chunks.
+            If no chunking should be done or the chunk size is the full input
+            then supply ``None``. Default is ``None``.
+
+        Returns
+        -------
+        data : dict
+             Vacuum field ∇Φ stored in ``data["evl"]["grad(Phi)"]``.
```
unalmis, line 292: if you prefer some other api let me know. i was taking care to make these functions pure (recall that is a technical term) so that they play nice with jax

unalmis: ideally each method would be a registered compute function

desc/integrals/singularities.py
```diff
@@ -546,24 +543,13 @@ def eval_pt(eval_data_i):
...
-    # we need the higher dummy NFP value, but the error message is more
-    # helpful with the nfp loop docstring.
-    errorif(
-        source_grid.num_zeta == 1 and source_grid.NFP == 1,
-        msg="Source grid cannot compute toroidal effects.\n"
-        "Increase NFP of source grid to e.g. 64.\n"
-        "This is required to " + nfp_loop.__doc__,
-    )
-    f = jnp.zeros((eval_grid.num_nodes, kernel.ndim))
-    f, _ = fori_loop(0, source_grid.NFP, nfp_loop, (f, source_data))
```
unalmis, line 566: this seems to work without passing source_data in the f_data tuple. pointing out in case I'm missing something about pure functions

README.rst
```diff
@@ -97,7 +97,7 @@ Contribute
 - `Contributing guidelines <https://github.com/PlasmaControl/DESC/blob/master/CONTRIBUTING.rst>`_
 - `Issue Tracker <https://github.com/PlasmaControl/DESC/issues>`_
 - `Source Code <https://github.com/PlasmaControl/DESC/>`_
-- `Documentation <https://desc-docs.readthedocs.io/>`_
+- `Documentation <https://desc-docs.readthedocs.io/en/latest/>`_
```
dpanici, line 100: I'd prefer this to be stable as then it is consistent with our pip package

desc/integrals/singularities.py
```diff
@@ -805,12 +806,15 @@ def _kernel_1_over_r(eval_data, source_data, diag=False):
 def _kernel_nr_over_r3(eval_data, source_data, diag=False):
     """Returns n ⋅ −∇G(x,x') = n ⋅ (x−x')|x−x'|⁻³."""
     dx = _dx(eval_data, source_data, diag)
-    n = rpz2xyz_vec(source_data["n_rho"], phi=source_data["phi"])
+    # Need to use e^rho*sqrt(g) to pass Green's ID test.
+    # Fourier spectrum is much more concentrated than n_rho for some reason.
```
unalmis, line 810: otherwise the w7x one fails, which is interesting because there's many nodes and no truncation.

README.rst
unalmis: ok

desc/basis.py
```diff
@@ -657,57 +657,19 @@ def evaluate(self, grid, derivatives=np.array([0, 0, 0]), modes=None):
         _, t, z = grid.nodes.T
         _, m, n = modes.T
 
-        t = t[tidx]
-        z = z[zidx]
+        t = kwargs["t"] if "t" in kwargs else t[tidx]
```
dpanici, line 660: `kwargs.get` would be cleaner here

dpanici (dismissed):
I still should dig deeper but we need to decide on an API for this stuff. Currently it is its own class and creates/stores stuff in its own dictionaries. We don't have a good API for compute magnetic field stuff right now so there is not necessarily something to go off of, but I think if we can make it clear what this class takes and also how to access things (like, giving it a compute_magnetic_field function that would let it compute things off-surface and on-surface depending on a flag? or a compute_on_surface and compute_off_surface method) for computing B on and off-surface would be useful. I'd like it to at least have a `compute_magnetic_field` method which has similar API to our current stuff.

Potentially we could also make use of FourierCurrentPotentialField for the vacuum part of this field, since it is essentially similar to that representation, and that would allow re-use of those methods for the off-surface evaluations of B from the vacuum part.

unalmis:
> Potentially we could also make use of FourierCurrentPotentialField for the vacuum part of this field, since it is essentially similar to that representation, and that would allow re-use of those methods for the off-surface evaluations of B from the vacuum part.

FYI, the vacuum field cannot be generated from a surface current distribution on the boundary, i.e. a FourierCurrentPotentialField. The latter can only capture magnetic fields that are orthogonal to the boundary.

desc/integrals/singularities.py
```diff
@@ -765,11 +651,19 @@ def polar_pt(i):
     return f
 
 
+def _add_reduce(x):
+    # https://github.com/jax-ml/jax/issues/23493
```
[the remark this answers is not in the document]
unalmis: only in new versions of jax it is resolved

desc/integrals/singularities.py
unalmis, line 1: The primary reason for some of the changes here, it to ensure that only one large tensor product is computed when inverting the system for Phi. Otherwise you need to compute many large tensor products and that makes it slow. Recall JAX and JIT cannot optimize such matrix multiplication order.

desc/integrals/singularities.py
```diff
@@ -810,10 +712,16 @@ def singular_integral(
...
-        ``DFTInterpolator``
+    known_map : (str, callable)
+        Optional. If map used in kernel of singular integral is known,
+        then may provide a callable to compute to avoid inefficient
+        interpolation and function approximation.
+        First index should store the name of the map used in the kernel
+        e.g. "Phi", and the second index should store the Python callable
+        that accepts a grid argument.
+    ndim : int
+        Default is kernel.ndim.
```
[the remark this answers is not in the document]
unalmis: it is needed to make the integrals work for inversion without computing large tensor products

desc/compute/_surface.py
```diff
@@ -21,14 +21,14 @@
...
     profiles=[],
     coordinates="tz",
-    data=[],
+    data=["zeta", "omega"],
     parameterization="desc.geometry.surface.FourierRZToroidalSurface",
 )
 def _x_FourierRZToroidalSurface(params, transforms, profiles, data, **kwargs):
     R = transforms["R"].transform(params["R_lmn"])
     Z = transforms["Z"].transform(params["Z_lmn"])
     # TODO(#568): change when zeta no longer equals phi
```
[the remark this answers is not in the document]
unalmis: most of the curve compute quantities don't work for general toroidal angle so i am just keeping them. this change was made to avoid some errors as the singular integral stuff assumed that if phi was computed then so was zeta and omega.

desc/objectives/_free_boundary.py
```diff
@@ -504,51 +504,20 @@ def build(self, use_jit=True, verbose=1):
...
-        except AssertionError as e:
-            warnif(
-                True,
-                msg="Could not build fft interpolator, switching to dft which is slow."
-                "\nReason: " + str(e),
-            )
-            interpolator = DFTInterpolator(
-                eval_grid, source_grid, self._st, self._sz, self._q
-            )
+        del self._st
```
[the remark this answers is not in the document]
unalmis: they may not be the correct st and sz anymore. defensive programming

README.rst
dpanici:
```suggestion
- `Documentation <https://desc-docs.readthedocs.io/en/stable/>`_
```

desc/basis.py
unalmis: then both branches would execute regardless

desc/integrals/singularities.py
unalmis: And also so that I can compute integral of a non-periodic integrand.

desc/integrals/_vacuum.py
```diff
@@ -0,0 +1,693 @@
...
+        )
+        errorif(
+            Phi_N is not None and Phi_N > Phi_grid.N,
+            msg=f"Got Phi_N={Phi_N} > {Phi_grid.N}=Phi_grid.N.",
+        )
+        basis = DoubleFourierSeries(
+            M=setdefault(Phi_M, Phi_grid.M),
+            N=setdefault(Phi_N, Phi_grid.N),
+            NFP=surface.NFP,
+            sym=setdefault(sym, False) and surface.sym,
```
dpanici, line 144: This will fail if not False, as the options are False, cos, sin

tests/test_integrals.py
```diff
@@ -692,28 +695,474 @@ def test_vanilla_params(self):
...
+        data = vac.compute_vacuum_field(chunk_size)["evl"]
+        grad_Phi = rpz2xyz_vec(data["grad(Phi)"], phi=evl_data["phi"])
+        np.testing.assert_allclose(
+            dot(grad_G(evl_data["x"]) - grad_Phi, evl_data["n_rho"]),
+            0,
+            atol=atol,
+        )
+
+    @pytest.mark.unit
+    def test_secular_splitting(self, chunk_size=100, I=0, Y=12):  # noqa: E741
```
unalmis, line 878: @dpanici 

I am going to table this approach for now. When we want to add the effect of secular terms we can just compute the `B0` field that generates that secular term and use that method of splitting the potential as described in my report. That is the standard way of doing it and our functionality to that is already working and tested.

desc/integrals/_vacuum.py
```diff
@@ -444,6 +445,38 @@ def compute_magnetic_field(self, chunk_size=None, coords=None):
...
+        if coords is not None:
+            self._set_new_evl_coords(coords)
+        elif "B0+grad(Phi)" in self._data["evl"]:
+            return self._data
+
+        self._data = self.compute_current_field(chunk_size)
+        self._data = self.compute_vacuum_field(chunk_size)
+        self._data["evl"]["B0+grad(Phi)"] = (
+            self._data["evl"]["B0"] + self._data["evl"]["grad(Phi)"]
+        )
```
unalmis, line 475: This is the same as above function, maybe you copied accidentily?

desc/integrals/_vacuum.py
```diff
@@ -146,6 +146,7 @@ def __init__(
 
         # Compute data on source grid.
         self._src_transform = Transform(src_grid, basis, derivs=1)
+        self._evl_transform = Transform(evl_grid, basis, derivs=1)
```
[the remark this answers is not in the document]
dpanici: This was for use inside of the free boundary so we can compute the data needed for the evl_data on the new eq surface at the current iteration in a way that works within our jitted compute functions (i..e to be passed to compute_fun inside of objective.compute)

desc/magnetic_fields/_laplace.py
```diff
@@ -0,0 +1,398 @@
...
+        surface,
+        M,
+        N,
+        sym=False,
+        M_coil=None,
+        N_coil=None,
+        sym_coil=None,
+        B_coil=None,
+        Y_coil=None,
+        I_plasma=0.0,  # noqa: E741
```
unalmis, line 295:
```suggestion
        I_plasma=0.0,
```

desc/objectives/_free_boundary.py
```diff
@@ -884,6 +854,277 @@ def _print(fmt, fmax, fmin, fmean, f0max, f0min, f0mean, norm, unit):
...
+    """
+
+    __doc__ = __doc__.rstrip() + collect_docs(
+        target_default="``target=0``.",
+        bounds_default="``target=0``.",
+    )
+
+    _scalar = False
+    _print_value_fmt = "Free surface Error: "
+    _units = "T^2"
```
unalmis, line 899:
```suggestion
    _units = "T^2 m^2"
```

desc/compute/_laplace.py
```diff
@@ -0,0 +1,926 @@
...
+    B0="_MagneticField : Field object to compute with.",
+    surface="Surface : Surface to compute on.",
+    chunk_size=_doc["chunk_size"],
+)
+def _B0_dot_n(params, transforms, profiles, data, **kwargs):
+    grid = transforms["grid"]
+    data["B0*n"] = kwargs["B0"].compute_Bnormal(
+        kwargs["surface"],
+        eval_grid=grid,
+        source_grid=grid,
```
unalmis, line 652: I am assuming all magnetic field objects copy the source grid to each field period and integrate over the full domain [0, 2pi).

desc/compute/_laplace.py
```diff
@@ -0,0 +1,926 @@
...
+    parameterization="desc.magnetic_fields._laplace.SourceFreeField",
+    chunk_size=_doc["chunk_size"],
+    B0="_MagneticField : Field object to compute with.",
+    public=False,
+)
+def _B0_field(params, transforms, profiles, data, RpZ_data, **kwargs):
+    coords = jnp.column_stack([RpZ_data["R"], RpZ_data["phi"], RpZ_data["Z"]])
+    RpZ_data["B0"] = kwargs["B0"].compute_magnetic_field(
+        coords=coords,
+        source_grid=transforms["grid"],
```
unalmis, line 680: I am assuming all magnetic field objects copy the source grid to each field period and integrate over the full domain [0, 2pi).

desc/compute/_laplace.py
```diff
@@ -0,0 +1,926 @@
...
+    profiles=[],
+    data=["x"],
+    parameterization="desc.magnetic_fields._laplace.SourceFreeField",
+    chunk_size=_doc["chunk_size"],
+    B_coil="_MagneticField : Field object to compute with.",
+)
+def _B_coil_field(params, transforms, profiles, data, **kwargs):
+    data["B_coil"] = kwargs["B_coil"].compute_magnetic_field(
+        coords=data["x"],
+        source_grid=transforms["grid"],
```
unalmis, line 705: I am assuming all magnetic field objects copy the source grid to each field period and integrate over the full domain [0, 2pi).

unalmis:
`nan` gradient is due to jax issue #14868

desc/objectives/_free_boundary.py
```diff
@@ -901,6 +871,369 @@ def _print(fmt, fmax, fmin, fmean, f0max, f0min, f0mean, norm, unit):
...
+        data, _ = self._field.compute(
+            ["interpolator"] if self._is_neumann else ["interpolator", "Y_coil"],
+            grid=self._grid,
+            q=self._q,
+            transforms=source_transforms,
+            B_coil=self._B_coil,
+            B_coil_chunk_size=self._B_coil_chunk_size,
+        )
+        # No net poloidal current in equation 4.13 of [1].
+        self._field.Y = 0.0 if self._is_neumann else data["Y_coil"]
```
unalmis, line 1088: As `obj.build` is not jitted, this line should execute and permanently modify `self._field.Y` as expected.

desc/objectives/_free_boundary.py
```diff
@@ -901,6 +871,369 @@ def _print(fmt, fmax, fmin, fmean, f0max, f0min, f0mean, norm, unit):
...
+        field,
+        *,
+        eval_grid=None,
+        grid=None,
+        coil_grid=None,
+        q=None,
+        maxiter=-1,
+        chunk_size=None,
+        B_coil_chunk_size=None,
+        I_sheet=0.0,
```
unalmis, line 976: Reviewer should make this a flag that sets I_sheet to an optimizable parameter. By default, I_sheet should be not optimizable and fixed to zero.

dpanici (dismissed):
unicode characters cannot be easily typed, I would prefer changing all ∇φ to grad(Phi) etc so they can be written out easily

rahulgaur104:
> All the solvers are tested for correctness. See the shared paper to explain this PR.
How can I access the shared paper? 
I have some simple questions about this process.

rahulgaur104:
I am interested because there is an older paper that uses an almost similar idea (in terms of theory) to solve for a perturbed potential phi_1.
Screenshot from 2025-08-22 13-58-11

unalmis:
There aren't any automatically differentiable boundary integral equation solvers in Python. Hence, I would still like to merge this as the solver works fine.

dpanici:
@dpanici  try using [jaxopt's anderson acceleration](https://jaxopt.github.io/stable/fixed_point.html#anderson-acceleration)

desc/backend.py
```diff
@@ -78,27 +79,98 @@ def g(x):
...
+            body_fun,
+            (x0, err0, 0, full_ps, full_fs),
+        )
+    else:
+
+        def cond_fun_m(state):
+            _, err, i, _, _ = state
+            return (i < m) & (~is_converged(err, tol))
+
+        ## run for m steps first with del2 then do anderson
```
unalmis, line 165: ### Caution with some of these changes

you must use iteration methods that guarentee convergence for this problem; no heuristics. Atkinson's `del2` method is a heuristic accelerator. When I tested it alone, it did not converge for our integral system as can be seen by choosing that option and running `test_convergence_run_fixed_point`, and generating the plots with `test_convergence_plot_fixed_point`.

1. Standard fixed point contraction guarantees convergence.
2. Anderson method is guaranteed to converge, which is why I recommended it.
3. (If you know the eigenspectrum, a fully iterative Chebyshev method works, but eigenspectrum of this operator is open mathematics problem as I discussed in the shared paper. So don't use chebyshev iteration).

## Improvements that can be made
There are two main improvements that can be done in the fixed point method. The first is reducing number of iteration. We should see with Anderson because the function we iterate `f` calls singular integrals which are are so expensive, so doing some extra work with Anderson to reduce calls to `f` should be worthwile.

The second is that when we autodiff through this, the function `_tangent_solve` is computing the inverse on the right hand side of the implicit function theorem.
image


That inversion is likely why optimization was slow/used memory with this implementation, even though computing the objective was fast. When we do normal root solve like with `map_pest_coordinates`, the Jacobian is 1x1 so its inverse is just the reciprocal. In this case the Jacobian is `mxm` where `m` is the number of output points the singular integral is evaluated `m~10^4`. We were using standard inversion so that's grows as m^3 so the cost was `10^12`! 

The way to avoid this is to also use fixed_point iterative to compute that inverse as discussed [here](https://docs.jax.dev/en/latest/advanced-autodiff.html#example-implicit-function-differentiation-of-iterative-implementations). Note this fixed point iteration to compute our vector Jacobian product doesn't involve any calls to singular integrals, so it should be super cheap. You can use Anderson here too. The jaxopt anderson method already has this built in with the methods `implicit_diff=True` and `implicit_diff_solve=None`. The latter refers to how that matrix is inverted. I did not check whether the default is fixed point for the latter. https://jaxopt.github.io/stable/_autosummary/jaxopt.AndersonAcceleration.html#jaxopt.AndersonAcceleration

unalmis: That might also be why the optimizer was stalling? Accumulating error from inverting such a large matrix directly . I still don't trust jax to do large scale computations due to the open bugs i reported with batching so that might also have been an issue.

dpanici: That comment is inaccurate actually, my bad, if you look at the code I am using `"simple"` for the first few iterations. I did not have `"del2"` used either for the iterations for any of the prior work on this PR or on the discretization patch one either

Thank you for checking this though, as these are good points. My next thing to try is jaxopt's version as it incorporates both of these improvements, and is likely a better impleemntation than my attempt at anderson here

dpanici: It was not straightforward to get JAXOPT working, as they don't seem to like us closing over all the optimizable params in our iteration function I am also not so familiar at this. Maybe it is doable, more likely we should attempt something like the linked JAX docs kaya posted, but also that example is not using custom_root so it needs some changing around to work with our setup, I think.

Also, while the fixed point iteration for the jacobian does not need any singuar integrals, wouldn't we also still require one of the jacobians of f, either w.r.t a or x, which would still need many singular integral evaluations? like the "A" and "B" things in the [linked jax docs ](https://docs.jax.dev/en/latest/advanced-autodiff.html#example-implicit-function-differentiation-of-iterative-implementations) which are themselves derivatives of the function f

dpanici: Could it be as simple as a change like this (just replacing _tangent_solve with the below in the custom_root call of fixed_point):

```python
def _tangent_solve_fixed_pt(g, y):
    # map f(u) = y + u @ A
    # where A is jax.jacfwd(fxn) wrt x

    # g here is fxn(x)-x
    # jac.jacfwd(g) = A-I
    # so jac.jacfwd(g)+I = A
    # then we can use fixed_pt to find
    # fixed pt of f(u=y+u@A)
    # and return that?
    A = jax.jacfwd(g)(y)
    A = A + jnp.eye(A.shape[0], A.shape[1])

    def f(u):
        return y + u @ A

    u0 = y
    p, _, _, _, _ = _fixed_point(
        f,
        u0,
        tol=1e-8,
        maxiter=100,
        method="anderson",
        is_converged=_is_converged,
        m=4,
        beta=0.25,
    )
    return p


```

dpanici: the above does not seem to solve either the speed or the non-convergence in vacuum, have not looked into the memory use. 

On this PR it seems to change behavior of the optimization, but on the discretization patch PR it does not seem to impact the final solution much. Still need to understand why

unalmis:
https://github.com/jax-ml/jax/pull/36383 was causing GPU results to be wrong on some machines. I have added a patch in an upstream library to bypass the issue.

desc/backend.py
unalmis: > Could it be as simple as a change like this

not quite, but its the right idea.  I implemented it properly using lineax/optimistix in #2213 for big improvement in memory and compute.

PR #1362, opened 2024-11-13
Refactor coil test to use parameterize

### LINKED ISSUE #1104: Refactor coil tests to use `pytest.mark.parametrize` which treats it as different tests rather than 1 big one so easier to debug and run specific ones in case of failure.

f0uriest:
Resolves #1104


dpanici:
#1210 _related?_

dpanici (approved):

YigitElma (approved):

PR #1363, opened 2024-11-13
Update `trust_region_step_exact_` solvers

YigitElma:
- Updates the initial guess for $\alpha$ to be 0 instead of 0.001* $\alpha$ _upper. I found out that the solution of $\alpha$ is usually pretty small. This reduces the total amount of QR's in the while loop.
- Removes the last QR outside the while loop of `trust_region_step_exact_qr`. The loop condition already satisfies the step norm to be around trust radius.

### Misc
- Adds `execute_on_cpu` flag to `desc.examples.get`


f0uriest:
isn't this the same as the `cho` option but using LU instead of cholesky?

YigitElma:
> isn't this the same as the `cho` option but using LU instead of cholesky?

Yes, pretty similar. `jax.scipy.linlag.solve` with symmetric A should call LAPACK `sysv` ([docs](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-fortran/2023-0/sysv.html)), and it is something like LU. For finding alpha, I used Secant method.

Cholesky is also really fast (maybe faster than this one), but maybe this is more accurate, that's why I wanted to test it. If I remember correctly the problem with Cholesky was the accuracy.

dpanici:
run on multiple threads

desc/optimize/tr_subproblems.py
```diff
@@ -421,26 +421,20 @@ def truefun(*_):
...
-            alpha, alpha_lower, alpha_upper, phi, k = state
+            p, alpha, alpha_lower, alpha_upper, phi, k = state
 
-            alpha = jnp.where(
-                (alpha < alpha_lower) | (alpha > alpha_upper),
-                jnp.maximum(0.001 * alpha_upper, (alpha_lower * alpha_upper) ** 0.5),
-                alpha,
-            )
+            alpha = jnp.where((alpha < alpha_lower), alpha_lower, alpha)
+            alpha = jnp.where((alpha > alpha_upper), alpha_upper, alpha)
```
f0uriest, line 437: do these changes help reduce the number of iterations of the subproblem? If so by how much? 

We may want to make the same changes to the other trust region subproblems (svd, cho etc) to keep them consistent.

f0uriest: also this could be simplified to `alpha = jnp.clip(alpha, alpha_lower, alpha_upper)`

YigitElma: I don't know the exact effect of clipping alpha rather than setting it to `0.001*alpha_upper`. My reasoning for these changes was to make the initial guess of alpha 0, and whenever alpha goes out of the range setting it to `0.001*alpha_upper` seemed to be taking alpha away from the actual solution. So, it seemed necessary to clip alpha when I changed the initial guess.

YigitElma: The total number iterations for root finding depends on the case but these changes help to reduce it. Previously a root finding was taking 5-10 iterations, now around 3-6.

f0uriest: ok in that case can you make the same changes to `trust_region_step_exact_{svd,cho}`

tests/test_optimizer.py
```diff
@@ -1414,7 +1414,8 @@ def test_optimize_coil_currents(DummyCoilSet):
         verbose=2,
         copy=True,
     )
-    # check that optimized coil currents changed by more than 15% from initial values
+    # check that on average optimized coil currents changed by more than
```
YigitElma, line 1417: #1406 made the same change I used here, the change of comment is from my version.

dpanici (dismissed):

YigitElma:
It is a bit hard to give proper speed improvement from this PR. It depends on the case. For example, currently our longest regression test, `test_ATF_results` used to take 4.5 mins on my laptop with GPU, now it takes 3mins. The final equilibrium is a bit different but the force error is lower.
ATF_continuation_pr_vs_master
```
Force error after pr: 1.1110e-04
Force error after master: 2.7513e-04
```

However, the HELIOTRON test is not affected as much as ATF. 2min 42 sec to 2min 30sec.

YigitElma:
I tried to apply the same changes to SVD subproblem solver but it caused `test_qh_optimization` test to fail. For the third step, it throws a bad approximation warning and quits early. 

I think the difference is, now, we call QR version in optimizer and SVD in perturb. Optimizer calls subproblem solver in a loop, so, if there is a bad result, it gets resolved. Perturb uses the first one. I don't know why it only affects the result of that specific test and not others but I can look into it later. I was planning to look at subproblem solver of perturb, mainly changing it from SVD to QR, also related to #1050 and #1409. For now, I will keep it as is.

Moreover, QR version takes QR decomposition inside the loop, so reducing the iteration number improves speed more. But the iterations of SVD are very fast (less than a millisecond), so, I don't think this change is necessary.

dpanici:
> It is a bit hard to give proper speed improvement from this PR. It depends on the case. For example, currently our longest regression test, `test_ATF_results` used to take 4.5 mins on my laptop with GPU, now it takes 3mins. The final equilibrium is a bit different but the force error is lower. ATF_continuation_pr_vs_master
> 
> ```
> Force error after pr: 1.1110e-04
> Force error after master: 2.7513e-04
> ```
> 
> However, the HELIOTRON test is not affected as much as ATF. 2min 42 sec to 2min 30sec.

was the difference in ATF due to some bug?

YigitElma:
> was the difference in ATF due to some bug?

I don't think it is a bug. Just slight numerical differences cause small discrepancy around axis region of ATF.

f0uriest (approved):

dpanici (approved):

PR #1366, opened 2024-11-14
Add mirror ratio objective

### LINKED ISSUE #848: Mirror Ratio objective
Would be useful to have an objective to target a particular mirror ratio. There is already the `"mirror ratio"` compute quantity, so this is trivial to implement for regular equilibria (can already do it with `GenericObjective`). The other use case is to implement something similar that can work for `OmnigenousField` after PR #535 is merged in.

f0uriest:
Resolves #848


tests/test_objective_funs.py
```diff
@@ -1423,6 +1424,53 @@ def test_signed_plasma_vessel_distance(self):
...
+        # axisymmetry, no iota, so B ~ B0/R
+        eq = Equilibrium()
+        eq.solve()
+        # R0 = 10, a=1, so Bmax = B0/9, Bmin = B0/11
+        mirror_ratio = (1 / 9 - 1 / 11) / (1 / 9 + 1 / 11)
+        obj = MirrorRatio(eq)
+        obj.build()
+        f = obj.compute(eq.params_dict)
+        # not perfect agreement bc eq is low res, so B isnt exactly B0/R
+        np.testing.assert_allclose(f, mirror_ratio, rtol=0.13)
```
ddudt, line 1439: It might be better to use the Solovev or D-shape examples. They aren't perfectly circular but are higher-resolution equilibria, so might give a more accurate solution anyways

ddudt (dismissed):
Approving, but would like to see if you can change the test to work with a lower tolerance

tests/test_objective_funs.py
```diff
@@ -1423,6 +1424,53 @@ def test_signed_plasma_vessel_distance(self):
             )
             obj.build()
 
+    @pytest.mark.unit
+    def test_mirror_ratio_equilibrium(self):
+        """Test mirror ratio objective for Equilibrium."""
+        # axisymmetry, no iota, so B ~ B0/R
+        eq = Equilibrium()
```
YigitElma, line 1431:
```suggestion
        eq = Equilibrium(L=8, M=8)
```

tests/test_objective_funs.py
YigitElma, line 1439:
```suggestion
        np.testing.assert_allclose(f, mirror_ratio, rtol=3e-3)
```

YigitElma: I guess increasing the resolution slightly can give better results without taking too much time.

YigitElma (approved):
I agree with @ddudt, the tolerance is a bit high, but not too important if you don't change it. I made some suggestions that should work.

YigitElma (changes requested):
Update changelog

YigitElma (approved):

desc/objectives/_geometry.py
```diff
@@ -1355,3 +1355,146 @@ def compute(self, params, constants=None):
...
+        }
+
+        timer.stop("Precomputing transforms")
+        if verbose > 1:
+            timer.disp("Precomputing transforms")
+
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, params, constants=None):
+        """Compute mirro ratio.
```
dpanici, line 1474:
```suggestion
        """Compute mirror ratio.
```

dpanici (dismissed):

YigitElma (approved):

YigitElma (approved):

dpanici (approved):

PR #1367, opened 2024-11-14
Allow compute kwargs to be passed to ``GenericObjective`` and ``ObjectiveFromUser``

### LINKED ISSUE #1224: Can't pass compute kwargs to `GenericObjective` or `ObjectiveFromUser`

f0uriest:
Resolves #1224


f0uriest:
@ddudt you made the linked issue, do you have a use case we can use as a test?

dpanici:
ISS04 compute quantity

desc/examples/__init__.py
```diff
@@ -54,6 +55,8 @@ def get(name, data=None):
     assert os.path.exists(path)
 
     eqf = desc.io.load(path)
+    if not isinstance(eqf, EquilibriaFamily):
```
ddudt, line 58: the new "reactor_QA" example is only an Equilibrium, not a family

desc/examples/reactor_QA.py
ddudt, line 1: We need an equilibrium with kinetic profiles for tests. Originally I was going to put this in `/tests/conftest.py`, but it takes several minutes to run so I decided to make it an example instead. It is probably useful for other things anyways.

dpanici:
setup.cfg to exclude the newly added script

dpanici (approved):

YigitElma (approved):

desc/examples/reactor_QA.py
```diff
@@ -0,0 +1,87 @@
+"""Example script for recreating the QA configuration of Landreman, Buller & Drevlak."""
```
dpanici, line 1: add more description, and emphasize that it is not the exact same as that paper (same for precise QA/QH)

YigitElma (approved):

dpanici (dismissed):

YigitElma (approved):

dpanici (approved):

PR #1369, opened 2024-11-14
Change `FixLambdagauge` to fix FSA of lambda to be zero

dpanici:
Our lambda gauge constraint previously enforced that $\lambda(\rho, 0,0)=0$ for asymmetric equilibria, however this seemed to be overly restrictive as it enforces that the SFL theta zeroline lies along the $\theta=0$ line.

The constraint now enforces that $<\lambda>$=0 on each flux surface (which amounts to setting every $m=n=0$ mode of $\lambda_{lmn}$ to zero) which should be sufficient for fixing the gauge freedom (and is consistent with the constraints on $\lambda$ outlined in the original 1983 Hirshman paper)

- [x] redo asym save test


desc/objectives/linear_objectives.py
```diff
@@ -793,88 +793,36 @@ class FixLambdaGauge(_Objective):
...
-    _scalar = False
-    _linear = True
-    _fixed = False  # not "diagonal", since it is fixing a sum
     _units = "(rad)"
     _print_value_fmt = "lambda gauge error: "
 
     def __init__(
         self,
         eq,
+        target=0,
```
f0uriest, line 802: I think for the other ones we don't even allow the user to pick target bounds or weight

f0uriest (commented):
Should probably also mention this in the changelog

rahulgaur104 (approved):
Merge after passing test_vmec_save_asym

tests/test_vmec.py
```diff
@@ -1166,66 +1176,64 @@ def test(
...
-    test("bsubvmn", "B_zeta", rtol_desc_desc_wout=3e-4)
+    test("bsubvmn", "B_zeta", grid=vol_half_grid, is_half_grid=True)
 
     # hard to compare to VMEC for the currents, since
     # VMEC F error is worse and equilibria are not exactly similar
     # just compare back to DESC
     test("currumn", "J^theta", atol_vmec_desc_wout=1e4)
     test("currvmn", "J^zeta", negate_DESC_quant=True, atol_vmec_desc_wout=1e5)
 
-    # can only compare lambda, sqrt(g) B_psi B^theta and B_theta at bdry
```
f0uriest, line 1188: Isn't this still true? Even with the same gauge freedom there's still no guarantee they end up with the same poloidal angle

f0uriest (approved):

tests/test_vmec.py
dpanici: Yes but the particular solution I have, they end up finding similar lambdas

PR #1373, opened 2024-11-18
Functionality to rotate an Equilibrium around Z axis

### LINKED ISSUE #1368: Rotating Equilibrium in Space
Hi All.

I have two different equilibria, and want to compare their boundaries using _plot_boundaries_, or _plot_comparison_, etc.
However their angle $\phi$ is not aligned each other, which makes it difficult for me to do it.

Can I rotate the equilibrium in space by 90 degress (pi/4) so that they are aligned with each other?

image

YigitElma:
Resolves #1368 

Adds a utility function to rotate an equilibrium around Z axis.

$$f(\rho, \theta, \zeta) = \sum_{lmn} f_{lmn} \mathcal{R}_l^{|m|} \mathcal{F}(m\theta) cos(|n|\zeta)$$

$$f(\rho, \theta, \zeta+dz) = \sum_{lmn} f_{lmn} \mathcal{R}_l^{|m|} \mathcal{F}(m\theta) \[cos(|n|\zeta)cos(|n|dz) - sin(|n|\zeta)sin(|n|dz)\]$$

$$f(\rho, \theta, \zeta) = \sum_{lmn} f_{lmn} \mathcal{R}_l^{|m|} \mathcal{F}(m\theta) sin(|n|\zeta)$$

$$f(\rho, \theta, \zeta+dz) = \sum_{lmn} f_{lmn} \mathcal{R}_l^{|m|} \mathcal{F}(m\theta) \[sin(|n|\zeta)cos(|n|dz) + cos(|n|\zeta)sin(|n|dz)\]$$

Take $cos(|n|\zeta)$ or $sin(|n|\zeta)$ from both and find new spectral coefficients. This requires modes $(l,m,n)$ and $(l,m,-n)$ to exist, that's why we need to set stellarator symmetry to false.


YigitElma:
```python
eq = get("precise_QH")
eq_rotated = eq.rotate_zeta(np.pi/2, copy=True)
fig = plot_3d(eq, "|B|")
plot_3d(eq_rotated, "|B|", fig=fig)
```
image
```python
eq = get("precise_QH")
eq_rotated = eq.rotate_zeta(np.pi, copy=True)
fig = plot_3d(eq, "|B|")
plot_3d(eq_rotated, "|B|", fig=fig)
```
image

f0uriest:
Would this be better as a function in desc.compat rather than a class method?

YigitElma:
Can we also change the file docstring of `compat.py` with this PR? I feel like those functions do more than just old DESC compatibility

desc/compat.py
```diff
@@ -261,3 +261,66 @@ def rescale(
...
+            f_lmn = np.array(eq_rotated.Z_lmn)
+            modes = eq_rotated.Z_basis.modes
+        elif fun == "L":
+            f_lmn = np.array(eq_rotated.L_lmn)
+            modes = eq_rotated.L_basis.modes
+        else:
+            raise ValueError("fun must be 'R', 'Z' or 'L'")
+
+        new_coeffs = f_lmn.copy()
+        mode_lookup = {(l, m, n): idx for idx, (l, m, n) in enumerate(modes)}
```
f0uriest, line 307: can use `Basis.get_idx` instead of this

tests/test_compat.py
```diff
@@ -277,3 +277,25 @@ def fun(eq):
     np.testing.assert_allclose(new_vals["B_max"], 2)
     np.testing.assert_allclose(new_vals["R0/a"], old_vals["R0/a"])
     np.testing.assert_allclose(new_vals["err"], old_vals["err"], atol=1e-10)
+
+
+@pytest.mark.unit
+@pytest.mark.solve
+def test_rotate_zeta():
```
f0uriest, line 284: would also be good to add a test that evaluates R,Z at zeta=0 for the original eq and then at zeta=zeta0 for an eq rotated by zero0 and ensure they're the same

desc/compat.py
```diff
@@ -261,3 +261,66 @@ def rescale(
...
+def rotate_zeta(eq, angle, copy=False):
+    """Rotate the equilibrium about the toroidal direction.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium to rotate.
+    angle : float
+        Angle to rotate the equilibrium in radians. The actual physical rotation
+        is by angle / self.NFP.
```
f0uriest, line 275: i'd prefer if angle is the actual rotation, and the user should include nfp explicitly

tests/test_compat.py
f0uriest: also a test rotating by pi/nfp for a symmetric equilibria to make sure the result is also symmetric

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Changelog
 
 New Features
 - Add ``from_input_file`` method to ``Equilibrium`` class to generate an ``Equilibrium`` object with boundary, profiles, resolution and flux specified in a given DESC or VMEC input file
+- Add ``rotate_zeta`` method to ``Equilibrium`` class to rotate an ``Equilibrium`` around Z axis.
```
f0uriest, line 6: function, not method

desc/compat.py
YigitElma: @dpanici also wanted this. I will change it.

tests/test_compat.py
YigitElma: are the new ones qood or you prefer to evaluate R and Z positions?

desc/compat.py
```diff
@@ -261,3 +261,66 @@ def rescale(
...
+    copy : bool, optional
+        Whether to update the existing equilibrium or make a copy (Default).
+
+    Returns
+    -------
+    eq_rotated : Equilibrium
+        Equilibrium rotated about the toroidal direction
+    """
+    eq_rotated = eq.copy() if copy else eq
+    if eq.sym and not (angle % np.pi == 0) and eq.N != 0:
```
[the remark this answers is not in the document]
unalmis: That won't solve the issue unless its over pi. you'll need something like this https://github.com/PlasmaControl/DESC/blob/555a95929cc3831c5d7ff059b97fd842e426b4ee/tests/test_compute_everything.py#L68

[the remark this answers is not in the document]
YigitElma: I think the better way to use just degrees. There is some weird numerical issue here,
```python
k = 11
(k*np.pi % np.pi)
```
```
3.1415926535897896
```
But this  error is not consistent. For example, 12 and 13 don't have but 15 has.

desc/compat.py
```diff
@@ -261,3 +261,67 @@ def rescale(
...
+    eq_rotated : Equilibrium
+        Equilibrium rotated about the toroidal direction
+    """
+    eq_rotated = eq.copy() if copy else eq
+    # We will apply the rotation in NFP domain
+    angle = angle * eq.NFP
+    if eq.sym and not angle % np.pi == 0 and eq.N != 0:
+        warnings.warn(
+            "Rotating a stellarator symmetric equilibrium by an angle "
+            "that is not a multiple of pi will break the symmetry. "
```
dpanici, line 290: change pi to pi/NFP

desc/compat.py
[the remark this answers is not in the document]
YigitElma: image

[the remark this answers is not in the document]
dpanici: Can't you check with something like

```python
angle_div_pi = angle / np.pi
angle_is_int_mult_of_pi = np.isclose(angle_div_pi, int(angle_div_pi),atol=1e-8)
```

desc/compat.py
```diff
@@ -261,3 +261,72 @@ def rescale(
...
+    Returns
+    -------
+    eq_rotated : Equilibrium
+        Equilibrium rotated about the toroidal direction
+    """
+    eq_rotated = eq.copy() if copy else eq
+    # We will apply the rotation in NFP domain
+    angle = angle * eq.NFP
+    # Check if the angle is a multiple of pi/NFP
+    kpi = np.isclose(angle % np.pi, 0, 1e-8, 1e-8) or np.isclose(
```
dpanici, line 289: one more question, what if `angle` is negative? should this not have an abs value on the angle?

YigitElma: I am pretty sure `mode` operator `%` only returns positive numbers.

dpanici: add neg angle test

dpanici (approved):

f0uriest (commented):
Should also add new function to docs

YigitElma:
> Should also add new function to docs

You mean the full path to CHANGELOG or some other documentation?

f0uriest:
> > Should also add new function to docs
> 
> You mean the full path to CHANGELOG or some other documentation?

Both, should add it to `docs/api.rst` and `docs/api_equilibrium.rst`

dpanici (approved):

f0uriest (approved):

PR #1374, opened 2024-11-18
Make `factorize_linear_constraints` faster

YigitElma:
- Replaces the for loop inside `factorize_linear_constrains` for degenerate constraints by `np.unique` along `axis=0`
- Adds benchmark test for `LinearConstraintProjection` build. Considering that initialization of the optimization problem takes around 20 seconds, this is a good case to test every PR.
- Refactors `solve_fixed_iter` benchmark to 2 parts, one with everything compiled, the other with cleared cache.
- Changes null-space and particular solution calculations to use `jax.numpy.linalg.svd`.

#1300 added the check for degenerate constraints check to the `factorize_linear_constraints` function. However, the loop over rows and calling `np.where` which is basically another loop makes it very slow (at least on GPUs). I wouldn't expect it to be that slow but apparently, it takes more than actual null-space and a particular solution calculations. Given that this check is not that necessary for most cases, this is a big burden.

## GPU Benchmarks
Here is the old version,
image

The new version with `jax.svd` and updated degenerate constraint handler,
image


YigitElma:
If anyone is interested here are the benchmarks for the second time `factorize_linear_constraints` is called (everything compiled)
image
image

dpanici (dismissed):

YigitElma:
For future reference, I have also tried `jnp.unique` but on CPU that is so slow that after 10mins of profiling I quit. Also some benchmarks fail with that (probably time limit). This issue was mentioned [here](https://github.com/jax-ml/jax/issues/17370). So, probably sticking with `np.unique` is enough for now.

dpanici:
> Maybe with this PR, I can also make the null-space calculation part faster too. Because A matrix is pretty sparse, there should be a better way. linear_constraint_projection_A

https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.svds.html
could use this with `k` equal to the number of rows of A?

YigitElma:
We can also get the null-space from QR or other decompositions which should be faster and accurate enough. The condition number of A is very low. For this test case it is 2.

f0uriest:
> We can also get the null-space from QR or other decompositions which should be faster and accurate enough. The condition number of A is very low. For this test case it is 2.

worth a try. you'd have to use rank-revealing qr because in general A isn't full rank. i know jax doesn't have that option but I think scipy does.

f0uriest:
> Maybe with this PR, I can also make the null-space calculation part faster too. Because A matrix is pretty sparse, there should be a better way. linear_constraint_projection_A

is this including the self consistency constraints? it looks like its just fixing some dofs? iirc its the self consistency stuff that makes it poorly conditioned at high res

YigitElma:
> > Maybe with this PR, I can also make the null-space calculation part faster too. Because A matrix is pretty sparse, there should be a better way. linear_constraint_projection_A
> 
> is this including the self consistency constraints? it looks like its just fixing some dofs? iirc its the self consistency stuff that makes it poorly conditioned at high res

I used this script for it
```python

con = get_fixed_boundary_constraints(eq)
con = maybe_add_self_consistency(eq, con)
con = ObjectiveFunction(con)
obj.build()
con.build()

lc = LinearConstraintProjection(obj, con)
lc.build()

plt.imshow(lc._A, aspect='auto', cmap='binary')
plt.colorbar()
```

desc/objectives/utils.py
```diff
@@ -98,37 +98,24 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
...
-        ):
-            row_idx_to_delete = np.append(row_idx_to_delete, rows_equal_to_this_row[1:])
-    # delete the affected rows, and also the corresponding rows of b
-    A_augmented = np.delete(A_augmented, row_idx_to_delete, axis=0)
+
+    # Find unique rows of A_augmented
+    unique_rows, unique_indices = np.unique(A_augmented, axis=0, return_index=True)
+
+    # Sort the indices to preserve the order of appearance
+    unique_indices = np.sort(unique_indices)
```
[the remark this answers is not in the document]
unalmis: i think its fine. np.unique returns unique elements in sorted order. Yigit's code wants the unique elements but without the sorting, so this sort allows him to unsort

desc/objectives/utils.py
```diff
@@ -185,11 +170,12 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
     # null space & particular solution
     A = A * D[None, unfixed_idx]
     if A.size:
```
YigitElma, line 172: I will debug the tests but I am confused about this if statement. I assume this is for single constraint vs multiple constraints? What exactly we try to achieve by `A.size`? It should give the total number of elements in A, right? Or no constraint vs with constraints?

desc/utils.py
```diff
@@ -431,6 +431,41 @@ def svd_inv_null(A):
     return Ainv, Z
 
 
+def qr_inv_null(A, b, tol=1e-10):
+    """Compute pseudo-inverse and null space of a matrix using QR.
```
YigitElma, line 435: Need to change this.

desc/objectives/utils.py
```diff
@@ -185,11 +170,12 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
...
-        A_inv, Z = svd_inv_null(A)
+        x_p, Z = qr_inv_null(A, b)
+        xp = put(xp, unfixed_idx, x_p)
     else:
         A_inv = A.T
         Z = np.eye(A.shape[1])
-    xp = put(xp, unfixed_idx, A_inv @ b)
+        xp = put(xp, unfixed_idx, A_inv @ b)
```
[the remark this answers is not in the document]
dpanici: or if there is a reason to move it, what is the reason?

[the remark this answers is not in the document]
YigitElma: With SVD it is easier to calculate `A_inv` but with QR, it is easier to solve Rx=b kind of system. So, instead of returning `A_inv` the new function `qr_xp_null` returns the particular solution and the null-space. I could also return `Q, R` and calculate `x_p` here, but it is more convenient to calculate it in `qr_xp_null`.

desc/objectives/utils.py
YigitElma: Yes, I was also confused about that. I feel like without the if statement we should be able to handle both cases. Inside the `qr_xp_null`, we can check if  `A` is rank 0 matrix or empty. If empty return empty `xp`, if rank 0 return 0.

desc/utils.py
```diff
@@ -431,6 +431,41 @@ def svd_inv_null(A):
...
+    Z : ndarray
+        Null space of A such that AZ=0 and Z.T@Z=I.
+
+    """
+    # Linear constraint matrix A is usually wide
+    # QR decomposition of A^T
+    Q, R = qr(A.T)
+    # Determine rank
+    diag = jnp.abs(jnp.diag(R))
+    rank = jnp.sum(diag > tol)
```
YigitElma, line 457: This is a naive way to find the rank, but it seems to be working. I don't know if our linear constraints result in this kind of matrix but, we may need to consider,

$$
R = \begin{bmatrix}
x & x & x & x & x \\
0 & 0 & x & x & x \\
0 & 0 & 0 & 0 & x \\
0 & 0 & 0 & 0 & 0 \\
0 & 0 & 0 & 0 & 0 \\
\end{bmatrix}
$$

which should have rank=3 but my current implementation will give 1.

desc/objectives/utils.py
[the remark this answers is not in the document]
YigitElma: Btw, we can consider using `pnadas.unique` too. It seems to be way faster, and returns in order of appearance.

desc/utils.py
dpanici: Have you checked this against rank from SVD for some examples in DESC? this is an approximate way to get the rank... I am not comfortable doing this to grab the matrix rank here.

desc/utils.py
```diff
@@ -431,6 +431,41 @@ def svd_inv_null(A):
...
+    -------
+    x_p : ndarray
+        Particular solution to Ax = b.
+    Z : ndarray
+        Null space of A such that AZ=0 and Z.T@Z=I.
+
+    """
+    # Linear constraint matrix A is usually wide
+    # QR decomposition of A^T
+    Q, R = qr(A.T)
```
dpanici, line 454: we can check this can't we? in case it ends up not being wide?

though I guess if it is not wide that means it is overdetermined

YigitElma: Yes, if it is tall then we have overdetermined and I don't think in DESC you can create that scenario by existing constraints, and since we clean degenerate stuff and fixed constraints already, we remove a lot of rows. So, I believe it is impossible to get tall A here, but I can add a check if there is a possibility.

desc/utils.py
f0uriest: you need to use pivoting qr to get the rank, the qr that you're using doesn't support that. try the scipy one

desc/objectives/utils.py
f0uriest: If there are no constraints A will be shape (0, dim_x) which means size is 0 but shape[1] is the dimension of x that we want

desc/objectives/utils.py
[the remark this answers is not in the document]
f0uriest: id rather not add another dependency just to save a tiny bit of time here

desc/utils.py
f0uriest, line 457: should make tol a relative tolerances based on the size of the matrix or largest diagonal or something, like for svd

Also the default should probably be smaller, like 10*eps or something

f0uriest: should also add a test that this works correctly for linearly dependent constraints. The unique logic will remove actual duplicates but won't catch stuff like

x1 + x2 = 4
2*x1 + 2*x2 = 8

YigitElma: I will try the scipy one when I profile again. But my hesitation is due to not being able to use GPU.

desc/objectives/utils.py
YigitElma: So, A_inv@b for this case is 0, right? To me multiplying empty arrays is very confusing.

dpanici:
Try throwing execute on cpu decorator

YigitElma:
New results with 32 cores. Overall effect is the same. When Rory said 2-3 seconds, those were for low res appearing in `solve_continuation`, My script is different and uses full res.

These are SVD and for loop for degeneracy, on CPU and GPU respectively, (and I show the second time it is called so everything is compiled)
image
image
GPU version is slower for the first run, but it seems to be the same on later.

Since QR is not done yet, I won't share the results of it but it is impressive on GPU :) So here are the results for SVD+np.unique on same settings
image
image

desc/utils.py
YigitElma: Nevermind, I just use JAX svd ans it is fast enough.

YigitElma: I switched to SVD again, but I changed the `svd_inv_null` to use JAX's SVD. The CPU benchmarks won't show a lot but on GPU it is very fast for consecutive runs.

YigitElma: I can still try QR but without JAX implementing RRQR I don't think it can be performant. Because scipy thing is slow.

dpanici (approved):
does using jnp make it faster on gpu now?

YigitElma:
> does using jnp make it faster on gpu now?

Yes, I have updated the benchmarks in the description. The CPU is the same, but it is much faster on the GPU. Especially once the SVD is compiled (yes some internal jax functions get compiled even outside of jit), the speed up is extreme for high res.

f0uriest (approved):

Issue #1346, opened 2024-11-04
`solve_continuation_automatic` can decrease L resolution incorrectly if final eq has `L` higher than `M`


dpanici:
```python
from desc.equilibrium import Equilibrium
from desc.continuation import solve_continuation_automatic
eq = Equilibrium(L=8,M=6,N=6)
eqf = solve_continuation_automatic(eq)
```

results in the first step having `L=12` and `M=N=6`, then the second step decreases `L` down to `8`, this is not intended behavior, we want the first step to be `L=6` probably, since that is `mres_step` default.

dpanici:
Also ends in an error

```python
{
	"name": "ValueError",
	"message": "params should have the same size as the basis, got 7 for basis with 5 modes",
	"stack": "---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[8], line 4
      2 from desc.continuation import solve_continuation_automatic
      3 eq = Equilibrium( Psi=-0.035, L=8,M=6,N=6)
----> 4 eq = solve_continuation_automatic(eq, verbose=1, bdry_step=0.34)[-1]

File ~/Research/DESC/desc/continuation.py:614, in solve_continuation_automatic(eq, objective, optimizer, pert_order, ftol, xtol, gtol, maxiter, verbose, checkpoint_path, jac_chunk_size, **kwargs)
    583     eqfam = _add_pressure(
    584         eq,
    585         eqfam,
   (...)
    596         jac_chunk_size=jac_chunk_size,
    597     )
    599     eqfam = _add_shaping(
    600         eq,
    601         eqfam,
   (...)
    612         jac_chunk_size=jac_chunk_size,
    613     )
--> 614 eq.params_dict = eqfam[-1].params_dict
    615 eqfam[-1] = eq
    616 timer.stop(\"Total time\")

File ~/Research/DESC/desc/optimizable.py:53, in Optimizable.params_dict(self, d)
     51 for key, val in d.items():
     52     if jnp.asarray(val).size:
---> 53         setattr(self, key, val)

File ~/Research/DESC/desc/equilibrium/equilibrium.py:1647, in Equilibrium.p_l(self, p_l)
   1640 @p_l.setter
   1641 def p_l(self, p_l):
   1642     errorif(
   1643         self.pressure is None,
   1644         ValueError,
   1645         \"Attempt to set pressure on an equilibrium with fixed kinetic profiles\",
   1646     )
-> 1647     self.pressure.params = p_l

[...]

Issue #914, opened 2024-02-29
Reduce testing time/memory


f0uriest:
Running our basic unit tests now takes 1.5+ hrs for me locally, and fairly often it will get killed due to OOM issues if I'm trying to do anything else at the same time.

A few suggestions:
- Reduce resolution as much as possible wherever possible
- Drop resolution after doing `desc.examples.get` since in most cases we don't need all the detail
- See if we can eliminate some tests are are now redundant with others. There's ways to see which tests cover which lines, so we can cut tests that don't cover new ground: https://pytest-cov.readthedocs.io/en/latest/contexts.html

f0uriest:
For reference, here's a breakdown of all the unit tests and how long they take
[test_durations.txt](https://github.com/PlasmaControl/DESC/files/14505446/test_durations.txt)

dpanici:
https://github.com/PlasmaControl/DESC/blob/95612fa625bd5ba939569004c767ed02594586f0/tests/test_objective_funs.py#L984

we could make this test use the `conftest` coils instead of making new ones, which is more expensive bc of intersection checks

YigitElma:
![image](https://github.com/user-attachments/assets/a8330280-14bd-4b6e-8964-15081947953c)
Here are some of the longest regression(bottom) and unit(top) tests.

dpanici:
maybe we could reduce the number of examples we have? or move some to "not tested" part of examples so it still exists but people know not to expect it to work all the time?

f0uriest:
We don't do that many actual examples, I think its just DSHAPE, HELIOTRON and ATF. I don't think we can get rid of any of those without missing some important code paths

YigitElma:
I think a couple of recent PRs improved the speed of the tests significantly. At least on CI, unit and regression tests take, on average 20-25 mins per worker, the longest notebook worker takes around 35 mins to complete, and benchmarks also complete under 40mins. We still need to clean up some tests to make these run in a reasonable time locally. But just giving some updates on the issue.

PR #1406, opened 2024-11-21
Update equilibria in `desc.examples`

### LINKED ISSUE #1185: Some examples do not run as intended in `desc/examples`
- Should we add the examples to the CI, to run them and ensure the results are nested and match some reference result? 
  - Or add a flag for `test_examples` we can add to a PR like `test_jax`, required to merge BUT does not run automatically so must be added each time.

### LINKED ISSUE #1186: Check + Update old notebooks on a recurring basis
- [ ] make a recurring issue to check notebooks for syntax/up-to-dateness with master and current API
- [ ] add assert statements to notebooks for correctness so nbtests also check result correctness

f0uriest:
Resolves #1185 
Resolves #1186 

Re-ran them all and manually verified they look good by eye and the usual metrics (low force error etc).


dpanici:
This might change/need checking again once #1073 is in (or maybe base this off of #1073?)

dpanici:
I will approve this once merge conflict is resolved, but I also want to note that I'd like each notebook to be re-ran as well. The 3D plot-containing ones may have to be each separate PRs to avoid the issues we have with the git diffs being too large and failing our tests. I will make an issue for this

dpanici:
Could try to reduce test time as well #914 by reducing res or not doing continuation for some of the examples

unalmis (approved):

tests/baseline/test_section_F.png
YigitElma, line 1: Is this normal? Looks weirdly uniform to me.

desc/examples/ATF
YigitElma, line 1: I think we can also change the radial resolution of ATF. Also, I am pretty sure this input file has a problem like #1346, because the first step will use L=6 and M=12 which will automatically set L=18, but the second step will solve it for L=M=12. I will check it tho

f0uriest: what do you mean it will set L=18? Pretty sure for the manual continuation method it should use exactly what you tell it (6 in this case)

tests/baseline/test_section_F.png
f0uriest: yeah it just means theres more variation at the boundary so the middle looks relatively flat

desc/examples/ATF
YigitElma: This maybe a different issue than the one I mentioned but if I run this,
```python
eqfam = desc.io.load("../desc/examples/ATF_output.h5")
for eq in eq_fam:
    print(eq)
    print(np.max(eq.R_basis.modes[:,0]))
    print(np.max(eq.R_basis.modes[:,1]))
```
I get
```
Equilibrium at 0x707624133770 (L=6, M=12, N=0, NFP=12, sym=True, spectral_indexing=fringe)
15
12
Equilibrium at 0x707624826150 (L=12, M=12, N=0, NFP=12, sym=True, spectral_indexing=fringe)
18
12
Equilibrium at 0x707625c04350 (L=12, M=12, N=0, NFP=12, sym=True, spectral_indexing=fringe)
18
12
Equilibrium at 0x7076cc258350 (L=12, M=12, N=0, NFP=12, sym=True, spectral_indexing=fringe)
18
12
Equilibrium at 0x70761be6bf80 (L=12, M=12, N=2, NFP=12, sym=True, spectral_indexing=fringe)
18
12
```

dpanici: This is due to how the fringe indexing works and the fact that `L` is the max difference btwn `m` and `l` for a given mode, and also is the maximum radial-only basis degree. It is weird though yes and is a bit unintuitive esp. when L < M

YigitElma: > what do you mean it will set L=18? Pretty sure for the manual continuation method it should use exactly what you tell it (6 in this case)

By L, I was actually meaning max l mode in the basis. And if the spectral indexing is ansi, then I think you get max l 18 for L=6 and M=12.

f0uriest: yeah, `L` for a basis/equilibrium is not actually the maximum value of the radial mode number. As dario mentioned its a bit trickier with the zernike indexing. 

In any event, it has nothing to do with the continuation method, you get the same result if you just create a basis with a given L,M

```python
basis = desc.basis.ZernikePolynomial(L=6, M=12, spectral_indexing="ansi")
desc.plotting.plot_basis(basis);
```
image

YigitElma: Yeah, I don't know why but I thought there was a conditional inside modes creation like L<M and L>M. that's why I made that comment. Yes, this version works. Sorry

desc/equilibrium/utils.py
```diff
@@ -141,6 +141,7 @@ def parse_axis(axis, NFP=1, sym=True, surface=None):
             axis[:, 1],
             axis[:, 2],
             axis[:, 0].astype(int),
+            axis[:, 0].astype(int),
```
dpanici, line 144: what is this doing? seems to be passing in the same modes for Rn as Zn?

dpanici: That seems off? or do we ensure before this that we have the full mode spectrum in `n` and just set the non-symmetric modes to zero if it is stell symmetric?

dpanici (dismissed):

YigitElma (dismissed):
I think we can also reduce the radial resolution of ATF and HELITRON (also change some solver tolerances) to get the same equilibrium, as you did here for W7-X. Those are the simplest examples but have the highest radial resolution. It would help reducing the CI testing time too.

YigitElma:
I found these to be also working. For `ATF`,
```
# spectral resolution
L_rad  =  8, 16
M_pol  =  12
N_tor  =  0x4, 4
M_grid =  12, 16
N_grid =  0x4, 6
```
For `HELIOTRON`,
```
# spectral resolution
L_rad  =  8, 18
M_pol  =  6, 12
N_tor  =  0,  0,  3
M_grid =  12, 18
N_grid =  0,  0,  6
```

f0uriest:
> I found these to be also working. For `ATF`,
> 
> ```
> # spectral resolution
> L_rad  =  8, 16
> M_pol  =  12
> N_tor  =  0x4, 4
> M_grid =  12, 16
> N_grid =  0x4, 6
> ```
> 
> For `HELIOTRON`,
> 
> ```
> # spectral resolution
> L_rad  =  8, 18
> M_pol  =  6, 12
> N_tor  =  0,  0,  3
> M_grid =  12, 18
> N_grid =  0,  0,  6
> ```

Reducing the resolution causes some tests to fail so I'm keeping the old one for now. You can make a new PR for those if you want.

dpanici (approved):

YigitElma (approved):

YigitElma:
> > I found these to be also working. For `ATF`,
> > 
> > ```
> > # spectral resolution
> > L_rad  =  8, 16
> > M_pol  =  12
> > N_tor  =  0x4, 4
> > M_grid =  12, 16
> > N_grid =  0x4, 6
> > ```
> > 
> > For `HELIOTRON`,
> > 
> > ```
> > # spectral resolution
> > L_rad  =  8, 18
> > M_pol  =  6, 12
> > N_tor  =  0,  0,  3
> > M_grid =  12, 18
> > N_grid =  0,  0,  6
> > ```
> 
> Reducing the resolution causes some tests to fail so I'm keeping the old one for now. You can make a new PR for those if you want.

Thanks for trying🙏

PR #1409, opened 2024-11-27
Use single `LinearConstraintProjection` in `ProximalProjection`

### LINKED ISSUE #1404: Don't call `factorize_linear_constraint` every time in `_update_equilibrium` of `ProximalProjection`?
https://github.com/PlasmaControl/DESC/blob/555a95929cc3831c5d7ff059b97fd842e426b4ee/desc/optimize/_constraint_wrappers.py#L734

- [ ] profile an optimization and see if this is actually a bottleneck

YigitElma:
- Adds a `_eq_solve_objective` property to `ProximalProjection` which will be `LinearConstraintProjection` of `ForceBalance` and fixed boundary constraints. 
- Updates the fixed parameter constraint targets of  `_eq_solve_objective`, and calculate particular solution without factorization
- Passes  `_eq_solve_objective` to `eq.perturb` and `eq.solve`

- Updates proximal-... benchmarks to include solve and perturb part. Previously, force balance was not solved since the previous x value was used. I slightly change the x using random number to include that part too

Resolves #1404


CHANGELOG.md
```diff
@@ -18,6 +18,11 @@ New Features
 - Changes hessian computation to use chunked ``jacfwd`` and ``jacrev``, allowing ``jac_chunk_size`` to now reduce hessian memory usage as well.
 - Adds an option to ``VMECIO.save`` to specify the grid resolution in real space.
 - Adds a new objective ``desc.objectives.CoilIntegratedCurvature`` for targeting convex coils.
+- `eq.solve` and `eq.perturb` now accept `LinearConstraintProjection` as objective. This option must be used without any constraints.
+
+Possible Breaking Change
+
+- `perturb` function takes `constraints` as keyword-only argument. With this change, codes like `perturb(eq, objective, constraints, deltas)` will not work properly, you need to change it to `perturb(eq, objective, deltas, constraints=constraints)`.
```
YigitElma, line 25: I don't need to change this probably. Maybe pass `constraints=None` for the special cases.

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -377,13 +379,15 @@ def test_proximal_freeb_jac(benchmark):
...
-    prox = ProximalProjection(objective, constraint, eq)
+    prox = ProximalProjection(objective, constraint, eq, solve_options={"maxiter": 1})
     obj = LinearConstraintProjection(
         prox, ObjectiveFunction((FixCurrent(eq), FixPressure(eq), FixPsi(eq)))
     )
     obj.build()
     obj.compile()
     x = obj.x(eq)
+    # we change x slightly to profile solve/perturb equilibrium too
+    x = x.at[0].add(0.001)
```
YigitElma, line 390: Inside `ProximalProjection._update_equilibrium()`, there is a check for previous x values, since this x is the same as the first x, it was not solving or perturbing the equilibrium. If we want to profile that, I can revert but I was surprised to see this PR slowed this test, and then I realised this and made the change, now, it only solves for 1 step, and perturb second order.

YigitElma: I used a random number to prevent following runs to have the same problem

YigitElma:
Figure out why `test_bootstrap_consistency_current` fails and nothing else...

desc/optimize/_constraint_wrappers.py
```diff
@@ -164,6 +173,40 @@ def unpack_state(self, x, per_objective=True):
         x = self.recover(x)
         return self._objective.unpack_state(x, per_objective)
 
+    def update_constraint_target(self, eq_new):
```
f0uriest, line 176: give this a full docstring

desc/optimize/_constraint_wrappers.py
```diff
@@ -790,9 +843,8 @@ def _update_equilibrium(self, x, store=False):
             # reset to last good params
             self._eq.params_dict = self.history[-1][self._eq_idx]
 
-        for con in self._linear_constraints:
-            if hasattr(con, "update_target"):
-                con.update_target(self._eq)
+        if (x != self._allx[-1]).all():
+            self._eq_solve_objective.update_constraint_target(self._eq)
```
f0uriest, line 847: i think this needs to be inside the else block above. if we reset the equilibrium we should reset the constraints as well

desc/optimize/optimizer.py
```diff
@@ -98,8 +98,10 @@ def optimize(  # noqa: C901
         ----------
         things : Optimizable or tuple/list of Optimizable
             Things to optimize, eg Equilibrium.
-        objective : ObjectiveFunction
-            Objective function to optimize.
+        objective : ObjectiveFunction or LinearConstraintProjection
+            Objective function to optimize. If a LinearConstraintProjection is passed,
+            no constraints is necessary. This is useful for using the ProximalProjection
```
f0uriest, line 103: see previous comment

desc/perturbations.py
```diff
@@ -109,8 +110,9 @@ def perturb(  # noqa: C901
     ----------
     eq : Equilibrium
         Equilibrium to perturb.
-    objective : ObjectiveFunction
-        Objective function to satisfy.
+    objective : ObjectiveFunction or LinearConstraintProjection
+        Objective function to satisfy. If a LinearConstraintProjection is passed,
```
f0uriest, line 114: see previous comment

desc/perturbations.py
```diff
@@ -186,24 +204,39 @@ def perturb(  # noqa: C901
...
     if verbose > 1:
         timer.disp("linear constraint factorize")
 
     # state vector
-    x = objective.x(eq)
+    if is_linear_proj:
+        obj = objective._objective
+    else:
+        obj = objective
+    x = obj.x(eq)
```
f0uriest, line 229: rather than renaming everything you could just do
```python
if is_linear_proj:
    obj = objective
    objective = obj._objective
```

desc/optimize/optimizer.py
```diff
@@ -199,72 +205,80 @@ def optimize(  # noqa: C901
...
-        if linear_constraint is not None and not linear_constraint.built:
-            linear_constraint.build(verbose=verbose)
-        if nonlinear_constraint is not None and not nonlinear_constraint.built:
-            nonlinear_constraint.build(verbose=verbose)
-
-        # combine arguments from all three objective functions
-        if linear_constraint is not None and nonlinear_constraint is not None:
-            objective, linear_constraint, nonlinear_constraint = combine_args(
-                objective, linear_constraint, nonlinear_constraint
+        if not is_linear_proj:
```
f0uriest, line 208: might be worth moving all the stuff in this block to a helper function rather than adding another level of conditionals ie
```python
if not is_linear_proj:
    obj, cons = _helper_func_to_combine_objs_and_check_stuff(obj, cons)
else:
   nonlinear_constraint = None
   if not objective.built:

**Review by f0uriest — CHANGES_REQUESTED** [review]

File: desc/optimize/_constraint_wrappers.py [hunk 9]
```diff
@@ -164,6 +173,40 @@ def unpack_state(self, x, per_objective=True):
...
+        # remove fixed parameters from A and b again by the same loop as in factorize
+        A, b, xp, unfixed_idx, fixed_idx = remove_fixed_parameters(A, b, xp)
+
+        # compute x_scale
+        x_scale = self._objective.x(*self._objective.things)
+        Dnew = jnp.where(jnp.abs(x_scale) < 1e2, 1, jnp.abs(x_scale))
+
+        # since D has changed, we need to update the ADinv
+        self._ADinv = (1 / Dnew)[unfixed_idx, None] * self._Ainv
+        self._D = Dnew
```
YigitElma, line 200: I also need to update the null space, like I did for Ainv. Previously, when I missed updating Ainv, `test_bootstrap_consistency_current` was failing.

desc/perturbations.py
YigitElma: Thanks

desc/optimize/optimizer.py
YigitElma: I am not sure about how to proceed with `x_scale`? Do people use it often? The constraint updater I have just uses the `D` for scaling.

desc/optimize/_constraint_wrappers.py
YigitElma: I am not sure I understand the previous if statement completely. If `store=True` then we store the intermediate equilibrium in `history`, but if it is `False`, `history` only has the first params_dict, right? Why do we go back to the initial one?

YigitElma:
> ```diff
>  test_build_transform_fft_midres         |     +3.72 +/- 3.09     | +2.36e-02 +/- 1.96e-02 |  6.57e-01 +/- 9.1e-03  |  6.33e-01 +/- 1.7e-02  |
>  test_build_transform_fft_highres        |     +3.22 +/- 2.72     | +3.17e-02 +/- 2.68e-02 |  1.02e+00 +/- 1.5e-02  |  9.85e-01 +/- 2.2e-02  |
> -test_equilibrium_init_lowres            |     +8.65 +/- 2.60     | +3.50e-01 +/- 1.05e-01 |  4.40e+00 +/- 4.9e-02  |  4.05e+00 +/- 9.3e-02  |
>  test_objective_compile_atf              |     +3.80 +/- 2.80     | +3.25e-01 +/- 2.39e-01 |  8.87e+00 +/- 5.9e-02  |  8.54e+00 +/- 2.3e-01  |
>  test_objective_compute_atf              |     +6.09 +/- 4.04     | +9.67e-04 +/- 6.41e-04 |  1.68e-02 +/- 3.0e-04  |  1.59e-02 +/- 5.7e-04  |
>  test_objective_jac_atf                  |     +0.29 +/- 1.44     | +5.80e-03 +/- 2.86e-02 |  1.98e+00 +/- 1.7e-02  |  1.98e+00 +/- 2.3e-02  |
>  test_perturb_1                          |     -4.12 +/- 3.48     | -6.70e-01 +/- 5.66e-01 |  1.56e+01 +/- 5.6e-01  |  1.63e+01 +/- 7.8e-02  |
> +test_proximal_jac_atf                   |     -3.00 +/- 0.89     | -2.56e-01 +/- 7.57e-02 |  8.27e+00 +/- 5.2e-02  |  8.52e+00 +/- 5.5e-02  |
> +test_proximal_freeb_compute             |    -48.76 +/- 1.22     | -9.92e-02 +/- 2.48e-03 |  1.04e-01 +/- 1.0e-03  |  2.03e-01 +/- 2.3e-03  |
> +test_solve_fixed_iter                   |     -5.42 +/- 1.79     | -1.88e+00 +/- 6.21e-01 |  3.27e+01 +/- 5.1e-01  |  3.46e+01 +/- 3.5e-01  |
> ```

Don't be misled by the 50% speed improvement in test_proximal_freeb_compute. It is because I made the constraint update part optional.

desc/optimize/_constraint_wrappers.py
f0uriest: the idea is suppose we try a step that gets rejected because it increases the cost function. In that case we want to perturb the next step from the last accepted step, not the bad one which may be unnested etc. The jacobian only gets computed after accepted steps, so thats how we know which ones are accepted, and computing the jacobian sets `store=True`. Otherwise its false and we assume that we should reset the equilibrium after trying the step.

desc/optimize/optimizer.py
f0uriest: what do you mean? x_scale will either be an array or string. If its a string it just gets passed through, if its an array you need to project it down to the optimization subspace

desc/equilibrium/equilibrium.py
```diff
@@ -2140,6 +2140,7 @@ def solve(
         ----------
         objective : {"force", "forces", "energy"}
             Objective function to solve. Default = force balance on unified grid.
+            ObjectiveFunction or LinearConstraintProjection can also be passed.
```
f0uriest, line 2143: I don't think `LinearConstraintProjection` is public so shouldn't mention it here in a public docstring. Using it here is only for internal purposes anyways.

desc/equilibrium/equilibrium.py
```diff
@@ -2324,8 +2325,7 @@ def perturb(
         ----------
         objective : ObjectiveFunction or LinearConstraintProjection
             Objective function to satisfy. Default = force balance. If
-            LinearConstraintProjection is used, no constraints are needed. This is
-            useful for eq update in ProximalProjection.
+            LinearConstraintProjection is used, no constraints are needed.
```
f0uriest, line 2328: see above. Basically I don't think these docstrings should change at all, since from a user standpoint the functions/methods are exactly the same, we're just adding some private stuff behind the scenes. You can move it to a commend in the function so other devs know what's going on.

desc/optimize/_constraint_wrappers.py
```diff
@@ -189,18 +209,34 @@ def update_constraint_target(self, eq_new):
...
         # There is probably a more clever way of doing this, but for now we just
         # remove fixed parameters from A and b again by the same loop as in factorize
+        # Actually A (unscaled linear constraint matrix without any degenerate rows)
+        # does not change here, but still recompute it while updating others
         A, b, xp, unfixed_idx, fixed_idx = remove_fixed_parameters(A, b, xp)
 
         # compute x_scale
-        x_scale = self._objective.x(*self._objective.things)
-        Dnew = jnp.where(jnp.abs(x_scale) < 1e2, 1, jnp.abs(x_scale))
+        if x_scale == "auto":
```
f0uriest, line 217: ohhh i tihnk i understood you before. I thought you meant `x_scale` argument to `optimize` which is internal to the optimizer, so shouldn't affect any of this stuff? I think you can get rid of this

desc/optimize/_constraint_wrappers.py
```diff
@@ -842,9 +879,12 @@ def _update_equilibrium(self, x, store=False):
         else:
             # reset to last good params
             self._eq.params_dict = self.history[-1][self._eq_idx]
-
-        if (x != self._allx[-1]).all():
-            self._eq_solve_objective.update_constraint_target(self._eq)
+            x_scale = (
+                self._solve_options["x_scale"]
```
f0uriest, line 883: i think you can remove this. `self._solve_options["x_scale"]` only affects stuff internal to the optimizer, it shouldn't affect anything here (ie, the x passed in to `_update_equilibrium` will always be in physical units)

desc/optimize/optimizer.py
```diff
@@ -100,8 +100,7 @@ def optimize(  # noqa: C901
             Things to optimize, eg Equilibrium.
         objective : ObjectiveFunction or LinearConstraintProjection
             Objective function to optimize. If a LinearConstraintProjection is passed,
-            no constraints is necessary. This is useful for using the ProximalProjection
-            equilibrium updates.
+            no constraints is necessary.
```
f0uriest, line 103: see previous comment

desc/optimize/optimizer.py
```diff
@@ -206,75 +205,19 @@ def optimize(  # noqa: C901
...
-                    )
+            objective, nonlinear_constraint, x_scale = (
+                get_combined_constraint_objectives(
+                    eq,
+                    constraints,
+                    objective,
+                    things,
+                    x_scale,
+                    self.method,
+                    method,
```
f0uriest, line 216: i think you just need to pass in self.method, not both

desc/perturbations.py
```diff
@@ -112,7 +112,7 @@ def perturb(  # noqa: C901
         Equilibrium to perturb.
     objective : ObjectiveFunction or LinearConstraintProjection
         Objective function to satisfy. If a LinearConstraintProjection is passed,
-        no constraints is necessary. This will be more efficient for proximal optimizers
+        no constraints is necessary.
```
f0uriest, line 115: see previous comment

desc/optimize/_constraint_wrappers.py
```diff
@@ -789,10 +879,12 @@ def _update_equilibrium(self, x, store=False):
...
-
-        for con in self._linear_constraints:
-            if hasattr(con, "update_target"):
-                con.update_target(self._eq)
+            x_scale = (
+                self._solve_options["x_scale"]
+                if "x_scale" in self._solve_options
+                else "auto"
+            )
+            self._eq_solve_objective.update_constraint_target(self._eq, x_scale=x_scale)
```
f0uriest, line 887: Just to check, regardless of if `store` is true or false, at the end of this function the equilibrium and the constraints will be self consistent correct?

Previously we always updated the constraint target, but I think if store=True then its not necessary, bc the equilibrium and constraints are both updated in perturb and because store=True we don't reset the equilibrium. We only need this if store=False bc in that case we're resetting the equilibrium to the previous step and hence need to also reset the constraint targets. 

Does that sound right?

desc/optimize/_constraint_wrappers.py
YigitElma: Yes, the internal equilibrium update will have same units, but if a user specifically wants to use different scaling for the equilibrium update, shouldn't we allow it? `self._solve_options["x_scale"]` is different than big `optimizer`s.

desc/optimize/optimizer.py
YigitElma: `method` is the parsed version of `self.method`, I didn't want to call it twice.

desc/optimize/_constraint_wrappers.py
YigitElma: I think the easiest way is to update the constraints if the new optimizable parameters are different than the last ones, like I had before. I reverted to that. There will probably be some redundancies, but not much.

desc/optimize/_constraint_wrappers.py
```diff
@@ -1109,13 +1206,8 @@ def _proximal_jvp_f_pure(constraint, xf, constants, dc, unfixed_idx, Z, D, dxdc,
...
     Fx_reduced = Fx @ jnp.diag(D)[:, unfixed_idx] @ Z
     Fc = Fx @ (dxdc @ dc)
-    Fxh = Fx_reduced
-    cutoff = jnp.finfo(Fxh.dtype).eps * max(Fxh.shape)
-    uf, sf, vtf = jnp.linalg.svd(Fxh, full_matrices=False)
-    sf += sf[-1]  # add a tiny bit of regularization
-    sfi = jnp.where(sf < cutoff * sf[0], 0, 1 / sf)
-    Fxh_inv = vtf.T @ (sfi[..., None] * uf.T)
-    return Fxh_inv @ Fc
+    Q, R = qr(Fx_reduced, mode="economic")
```
YigitElma, line 1209: I am just experimenting

YigitElma: I want to believe there is a way to do this but may need some additional changes to some tests. Benchmarks for this,
```diff
|             benchmark_name             |         dt(%)          |         dt(s)          |        t_new(s)        |        t_old(s)        | 
| -------------------------------------- | ---------------------- | ---------------------- | ---------------------- | ---------------------- |
 test_build_transform_fft_lowres         |     +0.15 +/- 9.10     | +8.40e-04 +/- 4.95e-02 |  5.45e-01 +/- 3.3e-02  |  5.45e-01 +/- 3.7e-02  |
 test_equilibrium_init_medres            |     -2.49 +/- 3.23     | -1.16e-01 +/- 1.51e-01 |  4.54e+00 +/- 5.8e-02  |  4.66e+00 +/- 1.4e-01  |
 test_equilibrium_init_highres           |     -0.12 +/- 1.95     | -7.13e-03 +/- 1.15e-01 |  5.87e+00 +/- 8.5e-02  |  5.87e+00 +/- 7.7e-02  |
 test_objective_compile_dshape_current   |     -0.88 +/- 3.60     | -3.70e-02 +/- 1.51e-01 |  4.17e+00 +/- 1.2e-01  |  4.21e+00 +/- 9.1e-02  |
 test_objective_compute_dshape_current   |     +3.03 +/- 4.06     | +1.57e-04 +/- 2.10e-04 |  5.33e-03 +/- 2.0e-04  |  5.17e-03 +/- 6.3e-05  |
 test_objective_jac_dshape_current       |     +1.88 +/- 7.02     | +8.16e-04 +/- 3.05e-03 |  4.43e-02 +/- 1.7e-03  |  4.35e-02 +/- 2.5e-03  |
 test_perturb_2                          |     -0.86 +/- 1.87     | -1.74e-01 +/- 3.79e-01 |  2.01e+01 +/- 3.0e-01  |  2.02e+01 +/- 2.3e-01  |
+test_proximal_jac_atf_with_eq_update    |    -11.98 +/- 1.23     | -5.31e+00 +/- 5.46e-01 |  3.90e+01 +/- 3.9e-01  |  4.44e+01 +/- 3.8e-01  |
 test_proximal_freeb_jac                 |     -1.28 +/- 1.26     | -9.58e-02 +/- 9.47e-02 |  7.40e+00 +/- 4.2e-02  |  7.50e+00 +/- 8.5e-02  |
 test_solve_fixed_iter_compiled          |     +0.53 +/- 2.79     | +1.10e-01 +/- 5.85e-01 |  2.11e+01 +/- 5.1e-01  |  2.10e+01 +/- 2.9e-01  |
 test_LinearConstraintProjection_build   |     +1.59 +/- 2.14     | +1.73e-01 +/- 2.32e-01 |  1.11e+01 +/- 2.3e-01  |  1.09e+01 +/- 3.6e-02  |
 test_build_transform_fft_midres         |     +0.31 +/- 2.92     | +1.84e-03 +/- 1.75e-02 |  6.02e-01 +/- 1.5e-02  |  6.00e-01 +/- 9.8e-03  |
 test_build_transform_fft_highres        |     -0.33 +/- 2.30     | -3.20e-03 +/- 2.20e-02 |  9.52e-01 +/- 2.0e-02  |  9.56e-01 +/- 8.1e-03  |
 test_equilibrium_init_lowres            |     +0.12 +/- 0.92     | +4.87e-03 +/- 3.73e-02 |  4.07e+00 +/- 3.2e-02  |  4.06e+00 +/- 2.0e-02  |
 test_objective_compile_atf              |     -0.16 +/- 1.34     | -1.30e-02 +/- 1.11e-01 |  8.27e+00 +/- 8.4e-02  |  8.28e+00 +/- 7.3e-02  |
 test_objective_compute_atf              |     -0.19 +/- 1.37     | -3.01e-05 +/- 2.15e-04 |  1.57e-02 +/- 1.5e-04  |  1.58e-02 +/- 1.5e-04  |
 test_objective_jac_atf                  |     +1.68 +/- 2.13     | +3.27e-02 +/- 4.14e-02 |  1.98e+00 +/- 3.2e-02  |  1.95e+00 +/- 2.7e-02  |
 test_perturb_1                          |     +0.10 +/- 1.02     | +1.41e-02 +/- 1.48e-01 |  1.45e+01 +/- 1.3e-01  |  1.45e+01 +/- 7.6e-02  |
+test_proximal_jac_atf                   |    -32.47 +/- 0.61     | -2.69e+00 +/- 5.09e-02 |  5.59e+00 +/- 3.1e-02  |  8.28e+00 +/- 4.1e-02  |
+test_proximal_freeb_compute             |    -46.95 +/- 0.56     | -9.54e-02 +/- 1.13e-03 |  1.08e-01 +/- 6.5e-04  |  2.03e-01 +/- 9.2e-04  |
 test_solve_fixed_iter                   |     +0.34 +/- 1.77     | +1.07e-01 +/- 5.65e-01 |  3.20e+01 +/- 2.9e-01  |  3.19e+01 +/- 4.9e-01  |
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -164,6 +177,72 @@ def unpack_state(self, x, per_objective=True):
...
+        the inverse of the scaled linear constraint matrix (ADinv) to reflect the new
+        equilibrium a.k.a. the new target of the constraint of system Ax=b. This
+        also updates the project and recover methods. Updating quantities in this way
+        is faster than calling factorize_linear_constraints again.
+
+        Parameters
+        ----------
+        eq_new : Equilibrium
+            New equilibrium to target for the constraints.
+        x_scale : ndarray or str, optional
```
f0uriest, line 193: still dont understand why x_scale is needed here?

desc/optimize/_constraint_wrappers.py
f0uriest: yeah but `self._solve_options["x_scale"]` doesn't affect anything in this function. The inputs and outputs are both in the original unscaled units

desc/optimize/_constraint_wrappers.py
```diff
@@ -789,10 +882,14 @@ def _update_equilibrium(self, x, store=False):
         else:
             # reset to last good params
             self._eq.params_dict = self.history[-1][self._eq_idx]
-
-        for con in self._linear_constraints:
-            if hasattr(con, "update_target"):
-                con.update_target(self._eq)
+        # if the optimization variables have changed, we need to update the constraints
+        if not (x == last_x).all():
```
f0uriest, line 886: this isn't the check we want. This will always be false, since we always append the new `x` to `allx` in line 871. What we want is to only reset the equilibrium to the last *good* step if store=False

desc/optimize/least_squares.py
```diff
@@ -397,7 +397,7 @@ def lsqtr(  # noqa: C901
             )
 
             if g_norm < gtol:
-                success, message = True, STATUS_MESSAGES["gtol"]
+                success, message = True, STATUS_MESSAGES["gtol"] + f" {gtol:.2e}"
```
f0uriest, line 400: why do we need to print out gtol? the norm of the gradient is already printed

desc/optimize/least_squares.py
```diff
@@ -412,7 +412,7 @@ def lsqtr(  # noqa: C901
             )
 
     if g_norm < gtol:
-        success, message = True, STATUS_MESSAGES["gtol"]
+        success, message = True, STATUS_MESSAGES["gtol"] + f" {gtol:.2e}"
```
f0uriest, line 415: see above

desc/optimize/_constraint_wrappers.py
YigitElma: I can get rid of it, but just wanted to keep the functionality we already have. Previously, for proximal if you use solve_option["x_scale"] this would be what was happening, right?

desc/optimize/_constraint_wrappers.py
YigitElma: `last_x` is the last element of `allx` at the beginning of the function, so no, this is not always false. This is different than equilibrium update because there we store the params_dict and replace it. Here we don't store the constraint, the logic is, if the optimizable params haven't changed from the last time this function is called, then constraints are the same so, they don't have to be updated. For example, the first iteration of the eq.update function doesn't need to call constraint target update.

desc/objectives/utils.py
```diff
@@ -100,61 +100,22 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
     A_augmented = np.hstack([A, np.reshape(b, (A.shape[0], 1))])
```
dpanici, line 100: rename xscale

desc/optimize/_constraint_wrappers.py
f0uriest: no, this is a different x_scale. The one the user passes to `optimize` isn't passed through to here. This is just a heuristic scaling to normalize very large qtys like pressure and current.

desc/optimize/_constraint_wrappers.py
f0uriest: I still don't follow. When would `last_x` be the same as `x`? It seems like that would only happen if the optimizer doesn't take a step.

f0uriest: You need something in the `if store` block above to ensure that when we change the equilibrium parameters we also update the constraints

desc/objectives/utils.py
YigitElma: added a note

desc/optimize/least_squares.py
YigitElma: Sometimes I want to reduce the gtol to run the optimization a little bit longer, but if it is set to default, I don't its value. So, I thought this would help.

desc/optimize/_constraint_wrappers.py
YigitElma: I think this should work as is. I can actually just remove the checks, because it is pretty fast. For now, if you call `compute_` or `jac_` outside of an optimization function, this will prevent updating the target. Inside the optimization loop, this will only block the first call since `allx` already has the initial state and the constraints are built accordingly. Besides that additional updates shouldn't cause too much time (even for high res this takes 300-400ms)

desc/optimize/least_squares.py
f0uriest: in that case we should modify the actual message to include the tol, this will currently print something like 
```
`gtol` condition satisfied. 1.00e-4
```
where its not clear what the numbers mean. We should make it somethin like
```
gtol(=1.00e-4) condition satisfied.
```
or something like that

desc/optimize/_constraint_wrappers.py
f0uriest: i think we just need to have 
```
self._eq_solve_objective.update_constraint_target(self._eq)
```

inside the `else` block.

f0uriest (approved):

YigitElma:
We can change one of the perturb benchmarks to use `LinearConstraintProjection` after this PR, like
```python
@pytest.mark.slow
@pytest.mark.benchmark
def test_perturb_2(benchmark):
    """Benchmark 2nd order perturbations with LinearConstraintProjection."""
    eq = desc.examples.get("SOLOVEV")
    objective = get_equilibrium_objective(eq)
    constraints = maybe_add_self_consistency(eq, constraints)
    con = ObjectiveFunction(constraints)
    lc = LinearConstraintProjection(objective, con)
    lc.build()
    tr_ratio = [0.01, 0.25, 0.25]
    dp = np.zeros_like(eq.p_l)
    dp[np.array([0, 2])] = 8e3 * np.array([1, -1])
    deltas = {"p_l": dp}

    args = (eq, lc)
    kwargs = {
        "deltas": deltas,
        "tr_ratio": tr_ratio,
        "order": 2,
        "verbose": 2,
        "copy": True,
    }

    def run(*args, **kwargs):
        jax.clear_caches()
        perturb(*args, **kwargs)

    benchmark.pedantic(run, args=args, kwargs=kwargs, rounds=10, iterations=1)
```
Most of the perturb calls result from `eq.update`.

f0uriest (dismissed):

desc/equilibrium/equilibrium.py
```diff
@@ -2177,13 +2178,19 @@ def solve(
             `OptimizeResult` for a description of other attributes.
 
         """
-        if constraints is None:
+        is_linear_proj = isinstance(objective, LinearConstraintProjection)
```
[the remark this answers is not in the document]
YigitElma: @f0uriest requested previously that this shouldn't be a part of the Public API (because `LinearConstraintProjection` is kind of hidden and constructed automatically for the user). Though, technically this should work with everything, the underlying operations are the same.

desc/optimize/least_squares.py
```diff
@@ -412,7 +412,7 @@ def lsqtr(  # noqa: C901
             )
 
     if g_norm < gtol:
-        success, message = True, STATUS_MESSAGES["gtol"]
+        success, message = True, STATUS_MESSAGES["gtol"] + f" ({gtol=:.2e})"
```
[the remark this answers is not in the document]
YigitElma: When the optimization terminates due to some tolerance, it prints what is the value of the tolerance

PR #1410, opened 2024-11-27
Jb/dps - Differential Particle Solver

joaopedrobiu6:
This adds an Objective `ParticleTracer` that uses the guiding center equations of motion to trace particles' trajectories given a DESC Equilibrium. The goal of this objective is to optimize the equilibrium directly from quantities measured from the trajectories traced, e.g. the deviation from the initial flux surface. 
It currently supports one-particle tracing and optimization of the equilibrium from that trajectory, and multiple-particle tracing. The optimization of the equilibrium from the multi-particle tracing is not yet fixed

TO DO LIST:
- [ ] Run with fwd mode for deriv mode, and see if that fixes the NaNs with multiple particles
- [ ] Run with the same particle twice, see if NaN occurs (make sure the conditions used worked with a single particle)
- [ ] Look at the trajectories for the particles in the case where the optimization gets NaN on the first iteration. Did some exit the equilibrium (psi > 1)?
- [ ] How to deal with particle trajectories that exit plasma (psi > 1)? Probably should just cut off integration there, since past the boundary any field info from DESC is nonsense and we do not want to be integrating the particles there
- [ ] Make a tutorial notebook showing how to use
- [ ] Move from `jax.experimental` to `diffrax`


dpanici:
Close once #1820 is in

PR #1419, opened 2024-12-02
bug fix for proximal scalar optimization

### LINKED ISSUE #1403: `TracerArrayConversionError` with proximal scalar optimization
Minimal working example:

```
eq = get("precise_QA")
optimizer = Optimizer("proximal-fmintr")
objective = ObjectiveFunction(AspectRatio(eq=eq, target=5))
constraints = (
    FixCurrent(eq=eq),
    FixPressure(eq=eq),
    FixPsi(eq=eq),
    ForceBalance(eq=eq),
)
[eq], _ = optimizer.optimize(things=eq, objective=objective, constraints=constraints)
```

Error message:

```
  File "/home/dudt/DESC/debug.py", line 21, in <module>
    [eq], _ = optimizer.optimize(things=eq, objective=objective, constraints=constraints)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dudt/DESC/desc/optimize/optimizer.py", line 314, in optimize
    result = optimizers[method]["fun"](
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dudt/DESC/desc/optimize/_desc_wrappers.py", line 360, in _optimize_desc_fmin_scalar
    result = fmintr(
             ^^^^^^^
  File "/home/dudt/DESC/desc/optimize/fmin_scalar.py", line 191, in fmintr
    f = fun(x, *args)
        ^^^^^^^^^^^^^
  File "/home/dudt/DESC/desc/optimize/_constraint_wrappers.py", line 244, in compute_scalar
    return self._objective.compute_scalar(x, constants)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dudt/DESC/desc/objectives/objective_funs.py", line 518, in compute_scalar
    f = jnp.sum(self.compute_scaled_error(x, constants=constants) ** 2) / 2
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dudt/DESC/desc/optimize/_constraint_wrappers.py", line 836, in compute_scaled_error
    xopt, _ = self._update_equilibrium(x, store=False)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dudt/DESC/desc/optimize/_constraint_wrappers.py", line 752, in _update_equilibrium
    xopt = f_where_x(x, self._allx, self._allxopt)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dudt/DESC/desc/optimize/utils.py", line 523, in f_where_x
    x, xs, fs = map(np.asarray, (x, xs, fs))
    ^^^^^^^^^
jax.errors.TracerArrayConversionError: The numpy.ndarray conversion method __array__() was called on traced array with shape float64[300]
The error occurred while tracing the function compute_scalar at /home/dudt/DESC/desc/objectives/objective_funs.py:501 for jit. 
See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.TracerArrayConversionError
```

Similar error with other scalar optimizers, such as the SciPy ones. Error does not occur with `"proximal-lsq-exact"`. 

Using JAX v0.4.35

ddudt:
Resolves #1403


desc/optimize/_constraint_wrappers.py
```diff
@@ -836,6 +836,25 @@ def compute_scaled_error(self, x, constants=None):
         xopt, _ = self._update_equilibrium(x, store=False)
         return self._objective.compute_scaled_error(xopt, constants[0])
 
+    def compute_scalar(self, x, constants=None):
```
YigitElma, line 839: Docs of `grad` and `hess` methods of `ProximalProjection` say that it computes the gradient of `self.compute_scalar` which this PR just adds. Should we change them?

tests/test_optimizer.py
```diff
@@ -324,6 +324,44 @@ def test_no_iterations():
     np.testing.assert_allclose(x0, out2["x"])
 
 
+@pytest.mark.regression
+@pytest.mark.optimize
+def test_proximal_scalar():
```
f0uriest, line 329: Do we need a full regression test for this? Can't we just check that the function works?

desc/optimize/_constraint_wrappers.py
ddudt: I think the docs are correct. `compute_scalar` computes the sum-of-squares error $\frac{1}{2} \Sigma f^2$, `compute_grad` computes the gradient of that scalar error as $f^T \cdot J$, and `compute_hess` computes the Hessian of that scalar error as $J^T \cdot J$.

f0uriest (approved):

desc/optimize/_constraint_wrappers.py
YigitElma: Ah ok, wrappers have different syntax. I was expecting smth more like `ObjectiveFunction`s grad method,
```python
    @jit
    def grad(self, x, constants=None):
        """Compute gradient vector of self.compute_scalar wrt x."""
        if constants is None:
            constants = self.constants
        return jnp.atleast_1d(
            Derivative(self.compute_scalar, mode="grad")(x, constants).squeeze()
        )
```

YigitElma (approved):

desc/optimize/_constraint_wrappers.py
ddudt: Both implementations should be equivalent in theory. @f0uriest would changing this to take the gradient of the new `compute_scalar` be any faster than the existing implementation?

PR #1421, opened 2024-12-03
Finite Build Coils

ryanwu4:


dpanici:
See `CoilCoilMinDistance` objective for masking logic for avoiding doing biot savart of a coil over itself.

ryanwu4:
I identified the 2 issues preventing JAX grad:
1. `safenorm` is necessary if eval points and coil points are the same, which is the case for the centerline self field. I changed the biot savart integral to do that, and also implemented an improved singularity-subtracted integral from the paper which should require fewer quad points, specifically for the compute_self_field function.
2. NaNs appearing in the unregularized biot savart integral were screwing with the gradient as well. I now have `compute_self_field` output a mask of where centerline points are, and the maxNormB objective function ignores those points (in a kinda gross way). I decided to do this over having a biot-savart safemode where it regularizes for points close to a centerline since the latter is more computationally expensive and might not play well with JAX control flow logic.

desc/coils.py
```diff
@@ -936,6 +1116,595 @@ def _check_type(coil0, coil):
...
+            current,
+            abdelta,
+        )
+
+        # pack the parameters for the self field computation
+        B_fb_params = {"current": current, "cross_section_dims": cross_section_dims}
+
+        names = ["p_frame", "q_frame", "curv1_frame", "curv2_frame", "x"]
+
+        if params is not None:
```
ryanwu4, line 1565: Is the way that I'm packing in params and transforms here correct? This is kind of hacky because the terms `p_frame`, `curv1_frame`, etc should (and are?) only being calculated with the curve implementation, which means that they don't vary over the cross-section dimensions of the grid and only over the centerline dimension. This is the intended behavior, as these variables are only dependent upon the centerline filament, even if the finite build components are dependent on the full cross section geometry.

ryanwu4: Getting the params with `get_params` and computing over the full grid with `compute_fun` instead of `self.compute()` basically allows me to pass in the full 3D coil grid without overwriting the base behavior of `compute` as implemented in the `Curve` class.

desc/compute/_field.py
```diff
@@ -3516,3 +3516,362 @@ def _L_grad_B(params, transforms, profiles, data, **kwargs):
...
+    data=["u_fb", "v_fb", "p_frame", "q_frame"],
+    parameterization="desc.coils._FiniteBuildCoil",
+)
+def _B_0_fb(params, transforms, profiles, data, **kwargs):
+    # scalars
+    current = params["current"]
+    a = params["cross_section_dims"][0]
+    b = params["cross_section_dims"][1]
+
+    # kind of a trick, the p and q frame computations are only done on the centerline
```
ryanwu4, line 3687: This is what I was referring to about the centerline computations being done in a weird way, since technically `p_frame` only varies over the centerline dimension, while `u_fb` only varies over the cross section dimension, but the parameterization correctly preserves this convention even with a 3D grid.

ryanwu4:
I'm adding final unit tests on the associated objective functions shortly, but otherwise this PR is basically ready to review.

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+
+        filament_count = finite_build_grid.num_nodes // finite_build_grid.num_zeta
+
+        x_fb = x_fb.reshape(finite_build_grid.num_zeta, filament_count, 3)
+        x_fb = x_fb.transpose(1, 0, 2)
+
+        def biot_savart_subfilament(x):
+            dx = (
+                jnp.roll(x, -1, axis=0) - x
+            )  # using finite diff instead of analytic tangent
```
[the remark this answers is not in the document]
f0uriest: (the main reason for this is that using approximate tangents will ruin the exponential convergence of the integral making it really innaccurate. If you don't have exact analytic tangents you should use `biot_savart_hh` instead)

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+
+        B_vec = vmap(biot_savart_subfilament, in_axes=(0,), out_axes=0)(x_fb)
+
+        B = jnp.sum(B_vec, axis=0)
+
+        if basis.lower() == "rpz":
+            B = xyz2rpz_vec(B, phi=phi)
+        return B
+
+    def compute_magnetic_field_multifilament_circ(
```
f0uriest, line 1836: `_private`

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+        ----------
+        xsection_grid : LinearGrid, int or None
+            Refer to `prep_grid` method. If `prep_grid` is False, this is assumed to be
+            the full 3D finite build grid. Otherwise, this is assumed to be the 2D
+            grid of the cross section.
+        centerline_grid : LinearGrid or int, optional
+            Refer to `prep_grid` method. If `prep_grid` is False, this is assumed to
+            have the same toroidal dimension as the xsection grid.
+        coil_frame : bool, optional
+            Whether to compute the field in the t, p ,q coil frame. Default is False,
```
f0uriest, line 1875: not sure what this means?

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+        if self.cross_section_shape == "circular":
+            return self.compute_self_field_circ(
+                finite_build_grid, centerline_grid, coil_frame, params, transforms
+            )
+        if self.cross_section_shape == "rectangular":
+            return self.compute_self_field_rect(
+                finite_build_grid, centerline_grid, coil_frame, params, transforms
+            )
+
+    def compute_self_field_rect(
```
f0uriest, line 1911: `_private`

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+        # find mask for the grid axis to prevent downstream singularities with
+        # biot savart
+        centerline_mask = jnp.zeros(len(finite_build_grid.nodes), dtype=bool)
+        centerline_mask = centerline_mask.at[
+            (L + 1) * (2 * M + 1) // 2 :: (L + 1) * (2 * M + 1)
+        ].set(True)
+
+        return B_self, x_fb, centerline_mask
+
+    def compute_self_field_circ(
```
f0uriest, line 2048: `_private`

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+        centerline_grid,
+        coil_frame=False,
+        params=None,
+        transforms=None,
+    ):
+        return NotImplementedError(
+            "Circular cross section self field not implemented yet"
+        )
+
+    def project_coil_frame(
```
f0uriest, line 2060: Is this method needed? I feel like comptuting projections is pretty straightforward?

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+            ), "Rectangular cross section coils can only have exactly two dimensions"
+        self._cross_section_dims = new
+
+    @property
+    def cross_section_shape(self):
+        """str: Shape of the coil cross section, either 'circular' or 'rectangular'."""
+        return self._cross_section_shape
+
+    @classmethod
+    def prep_grid(cls, xsection_grid, centerline_grid=None):
```
f0uriest, line 1610: I'm not crazy about having two different grids for finite build coils that need to get combined in a particular way. I think it would make more sense to just require the user to give a 3d grid if they want 3d effects (and hence if they pass in a 1d grid it just does the usual centerline only stuff)

desc/coils.py
f0uriest: in theory this is fine, but shouldn't params already have the current and centerline dims?

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+        self,
+        cross_section_dims=[0.1, 0.1],
+        cross_section_shape="rectangular",
+        current=1,
+        center=[10, 0, 0],
+        normal=[0, 1, 0],
+        r_n=2,
+        modes=None,
+        basis="xyz",
+        name="",
```
f0uriest, line 2165: minor point but i'd prefer if the ordering kept the regular FourierPlanarCoil params first and added the finite build params on at the end

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+            center,
+            normal,
+            r_n,
+            modes,
+            basis,
+            name,
+        )
+
+    @classmethod
+    def from_FourierPlanarCoil(
```
f0uriest, line 2184: would also be useful to have a `to_FourierPlanarCoil` method to give you back the centerline

f0uriest: we should also think about how to convert between other finite build coils, like how the regular `FourierPlanarCoil` has a `to/from_FourierXYZ` etc.

desc/objectives/_coils.py
```diff
@@ -850,6 +858,233 @@ def body(k):
...
+    magnitude or components projected in the coil frame. All coils in the coilset must
+    inherit from _FiniteBuildCoil.
+
+    Parameters
+    ----------
+    coil : CoilSet
+        Coil(s) that are to be optimized.
+    component :  {'mag', 'p', 'q'}, optional
+        Component of magnetic field to be targeted. Can be 'mag' for field magnitude,
+        or 'p' or 'q' for the respective dimensions of the conductor cross section.
```
f0uriest, line 877: do we also want an option to project onto the tangent direction?

f0uriest (commented):
This is a first pass, mostly focused on API stuff. There's a lot of good things here but I think it can be cleaned up/simplified.

One thing that felt kind of clunky was the grids for centerline vs cross section, and I wonder if we should make some new dedicated grid classes?

Right now all of our grid classes are basically assuming the coordinate system is rho,theta,zeta in an equilibrium, and we sort of hackily make that work for surfaces and curves.

A better option might be to make a dedicated `CurveGrid` class that only has a 1d set of nodes along the centerline, a `SurfaceGrid` that only has 2d nodes on a surface etc. Then we could also make a `FiniteBuildCoilGrid` or something with nodes along the centerline and cross section?

desc/compute/_field.py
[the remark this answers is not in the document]
ddudt: I think the first few functions about curve coordinates/components can be moved to `_curve.py`, but the last few functions related to `B` can stay here.

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+    ----------
+    cross_section_dims : array-like
+        Dimensions of the coil cross section, with 1 or 2 dimensions depending on
+        the cross section shape (circular or rectangular).
+        For circular cross sections, this should be a single value representing
+        the radius. For rectangular cross sections, this should be a 2-element array
+        representing the dimensions in the p and q directions, respectively.
+        Refer to fig 3 in https://arxiv.org/pdf/2310.12087 for the p and q directions.
+
+    cross_section_shape : str
```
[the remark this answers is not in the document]
ddudt: Yeah I don't think this input is necessary, since we can assume it is circular if `cross_section_dims.size == 1` and rectangular if `cross_section_dims.size == 2` (and throw an error if size > 2)

desc/coils.py
ddudt: Didn't we decide that all new classes should require keyword arguments going forward? So let's enforce that by adding `*` after `self`.  (And also make this change for all the other new classes added in this PR.)

desc/coils.py
ddudt: Ideas:

- circle -> rectangle can convert to a square cross-section with width=radius
- rectangle -> circle can use the average width/height as the radius

desc/compute/_field.py
```diff
@@ -3837,3 +3838,368 @@ def _L_grad_B(params, transforms, profiles, data, **kwargs):
...
+    name="u_fb",
+    label="u_{FB}",
+    units="",
+    units_long="",
+    description="U coordinate for finite build expansion",
+    dim=1,
+    params=[],
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="rt",  # just two degrees of freedom
```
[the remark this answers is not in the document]
ryanwu4: Is there a clearer notation I should use the denote the coordinate frame of the coil? I was mostly just trying to signify that it varies only over the first two dimensions of the grid.

desc/coils.py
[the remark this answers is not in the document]
ryanwu4: I believe that the tangent is not exactly the same as the centerline, as in the case of a straight segment of coil with the cross section profile rotating, the tangent vectors of current chunks at the outer edges of the bundle will have a rotational component which the centerline does not have. I'll use `biot_savart_hh `instead.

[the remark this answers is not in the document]
f0uriest: If it's possible to get the analytic tangents that would be the best option, `biot_savart_quad` is generally a lot more accurate than `biot_savart_hh` for a given resolution, provided the tangents are correct. 

It sounds like the correct tangent would be the centerline tangent + rotation which should be possible.

desc/coils.py
f0uriest: that was more for objectives, I don't think we decided for that for all classes

desc/coils.py
ryanwu4: I've added an overridden `to_FourierPlanarCoil`. For the latter two conversion cases (to finite build coils with other parameterizations and between cross sections), are these used commonly enough? The conversion to other finite build coils could be doable via converting to the underlying coil class and then adding finite build dimennsions again, or perhaps I can add those methods to the generic `AbstractFiniteBuildCoil` once we have more parameterizations. 

For circle <-> rectangle, I feel like the user would usually want to just reset the cross section dims with the setter method, rather than have an implicit conversion. Is that a fair assumption?

desc/objectives/_coils.py
```diff
@@ -850,6 +858,233 @@ def body(k):
...
+        centerline_grid=None,
+        name="field on coil",
+        jac_chunk_size=None,
+        use_softmax=False,
+        softmax_alpha=1.0,
+    ):
+        from desc.coils import CoilSet, _FiniteBuildCoil
+
+        if target is None and bounds is None:
+            bounds = (-0.0, 20.0)
```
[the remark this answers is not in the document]
ryanwu4: I did choose that default just based on the 20 T max field as a reasonable limit for current HTS. Just changed it to `target=0` for simplicity.

desc/coils.py
```diff
@@ -1328,6 +1481,756 @@ def _check_type(coil0, coil):
...
+        if params is not None:
+            multifilament = params.pop("multifilament", False)
+            xsection_grid = params.pop("xsection_grid", None)
+            prep_grid = params.pop("prep_grid", True)
+            finite_build_grid = params.pop("finite_build_grid", None)
+
+            # make params None if it's empty to correctly handle compute functions
+            if not params:
+                params = None
+        if not multifilament:
```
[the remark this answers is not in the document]
ryanwu4: I'm going to implement the change you suggested where the filamentary operation is used if a 1D grid is passed and the multifilament is used if a 3D grid is passed.

tests/test_compute_everything.py
```diff
@@ -171,6 +177,15 @@ def test_compute_everything():
...
             current=5, X=[5, 10, 2, 5], Y=[1, 2, 3, 1], Z=[-4, -5, -6, -4]
         ),
+        "desc.coils.FourierPlanarFiniteBuildCoil": FourierPlanarFiniteBuildCoil(
+            cross_section_dims=[0.1, 0.2],
+            cross_section_shape="rectangular",
+            current=5,
+            center=[10, 1, 3],
+            normal=[1, 2, 3],
+            r_n=[1, 2, 3],
+            modes=[0, 1, 2],
```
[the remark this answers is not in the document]
ryanwu4: I just matched the `FourierPlanarCoil` implementation, I can fix all of these if its preferred but I want to be sure that I don't break things when I do that

ryanwu4:
This latest commit should address all the comments above except for the analytic tangents. There's now only a 3D grid used for all the relevant compute function, but I've kept the `prep_grids ` class method just for utility. The multifilament integral and the coilset max field on coil objective are now benchmarked more rigorously as well.

desc/coils.py
```diff
@@ -240,6 +257,134 @@ def biot_savart_vector_potential_quad(
...
+    )
+
+    B += (
+        0.5
+        * curvatures[:, jnp.newaxis]
+        * binormals
+        * (-2 + jnp.log(64 * eval_prime_squared / regularization))[:, jnp.newaxis]
+    )
+
+    # 1e-7 == mu_0/(4 pi)
```
dpanici, line 382: They actually have updated this constant in recent years (#1513) and this is no longer so accurate, I would prefer if we used the `mu_0` imported from `scipy.constants`

desc/coils.py
```diff
@@ -1456,6 +1602,770 @@ def _check_type(coil0, coil):
...
+        elif isinstance(xsection_grid, numbers.Integral):
+            L = xsection_grid * 2
+            M = xsection_grid
+        elif isinstance(xsection_grid, LinearGrid):
+            if not xsection_grid.endpoint:
+                raise ValueError(
+                    "The cross section grid must have endpoint=True to compute "
+                    "the self field"
+                )
+            elif xsection_grid.N > 1:
```
dpanici, line 1764: I am not sure if N=1 is what you want, don't you actually want `grid.num_zeta==1`?

an `N=1` LinearGrid has `2N+1` zeta planes in it

desc/coils.py
```diff
@@ -1456,6 +1602,770 @@ def _check_type(coil0, coil):
...
+                L = xsection_grid.L
+                M = xsection_grid.M
+
+        # set centerline grid if not provided for phi cross sectional coordinates
+        if centerline_grid is None:
+            centerline_grid = LinearGrid(N=50)
+        elif isinstance(centerline_grid, numbers.Integral):
+            centerline_grid = LinearGrid(N=centerline_grid)
+        elif isinstance(centerline_grid, LinearGrid):
+            if centerline_grid.L > 1 or centerline_grid.M > 1:
```
dpanici, line 1776: similar to above, should be num_rho==1 and num_theta==1 I think

dpanici: or L==0 and M==0

desc/coils.py
```diff
@@ -1456,6 +1602,770 @@ def _check_type(coil0, coil):
...
+        elif isinstance(xsection_grid, LinearGrid):
+            if not xsection_grid.endpoint:
+                raise ValueError(
+                    "The cross section grid must have endpoint=True to compute "
+                    "the self field"
+                )
+            elif xsection_grid.N > 1:
+                raise ValueError("The cross section grid must have dimension N=1")
+            else:
+                L = xsection_grid.L
```
dpanici, line 1767: Might be better to throw a warning or even an error here, as this means they passed in a grid that is not LinearGrid, so the meaning of L and M may be different

Like I could pass a ConcentricGrid into this thinking that is how the XS will be discretized, and it will silently be instead changed to have linear spacing in that dimension instead of the jacobi roots that ConcentricGrid has for radial spacing

dpanici: Or just mark it clearly in the docstring

desc/coils.py
```diff
@@ -1456,6 +1602,770 @@ def _check_type(coil0, coil):
...
+        Uses direct quadrature of the Biot-Savart integral for filamentary coils with
+        tangents provided by the underlying curve class. Convergence should be
+        exponential in the number of points used to discretize the curve, though curl(B)
+        may not be zero if not fully converged.
+
+        """
+        multifilament = False
+        if source_grid is not None:
+            if isinstance(source_grid, LinearGrid):
+                if source_grid.L > 1 or source_grid.M > 1:
```
dpanici, line 1839: Should be >0

desc/coils.py
```diff
@@ -1456,6 +1602,770 @@ def _check_type(coil0, coil):
...
+            cross_section_dims = self.cross_section_dims
+            alpha_n = self.alpha_n
+        else:
+            current = params.pop("current", self.current)
+            cross_section_dims = params.pop(
+                "cross_section_dims", self.cross_section_dims
+            )
+            alpha_n = params.pop("alpha_n", self.alpha_n)
+
+        # this is a somewhat hacky solution, but the points for the cross section must
```
dpanici, line 1892: can you explain this to me a bit? I dont think I understand, cross_section_dims are the height/width of the rectangle, why do they need to be rescaled?

desc/coils.py
```diff
@@ -1456,6 +1602,770 @@ def _check_type(coil0, coil):
...
+        self,
+        coords,
+        params=None,
+        basis="rpz",
+        finite_build_grid=None,
+        transforms=None,
+        chunk_size=None,
+    ):
+        return NotImplementedError(
+            "Multifilament circular cross section coils are not yet implemented."
```
dpanici, line 1973: add issue for this

desc/coils.py
```diff
@@ -1456,6 +1602,770 @@ def _check_type(coil0, coil):
...
+                obj=self,
+                grid=finite_build_grid,
+                jitable=True,
+            )
+
+        compute_kwargs = {}
+        # pass the basis_in arg to compute_fun if necessary
+        try:
+            compute_kwargs["basis_in"] = self.basis
+        except KeyError:
```
dpanici, line 2100: I think this would be an AttributeError, not a KeyError

desc/coils.py
dpanici: Yea I think that for converting from circle <> rectangle, I would require them to set the specific things instead of assuming defaults

desc/objectives/_coils.py
```diff
@@ -850,6 +858,258 @@ def body(k):
...
+    magnetic field on that coil in the component chosen. Options for magnetic field are
+    magnitude or components projected in the coil frame. All coils in the coilset must
+    inherit from AbstractFiniteBuildCoil.
+
+    Parameters
+    ----------
+    coil : CoilSet
+        Coil(s) that are to be optimized.
+    field : MagneticField, optional
+        External field to be added to the coil field. Defaults to None.
```
dpanici, line 876: specify that this is not included in the optimizable things

desc/objectives/_coils.py
```diff
@@ -850,6 +858,258 @@ def body(k):
...
+        from desc.magnetic_fields import SumMagneticField
+
+        coilset = self.things[0]
+
+        finite_build_grid = AbstractFiniteBuildCoil.prep_grid(
+            self._xsection_grid, self._centerline_grid
+        )
+
+        self._dim_f = coilset.num_coils
+        self._constants = {
```
dpanici, line 999: I think we are trying to move away from constants #1758 , so would prefer if instead these were all just attributes of the objective self (though I guess `"quad_weights"` is special as we check for it later right in`compute_scaled_error`? so maybe just leave that one but put rest as attributes of self)

desc/objectives/_coils.py
```diff
@@ -850,6 +858,258 @@ def body(k):
...
+            "coilset": coilset,
+            "finite_build_grid": finite_build_grid,
+            "centerline_grid": self._centerline_grid,
+            "quad_weights": 1.0,
+        }
+
+        if self._has_field:
+            self._constants["field"] = SumMagneticField(self._field)
```
dpanici, line 1007: same for here (also why is it wrapped with SumMagneticField? to allow self._field to be a list of fields? That should probably be something that the user does themselves I would think, unless you have a good reason for doing so? the docstring only says a single MagneticField is passed in)

desc/objectives/_coils.py
```diff
@@ -850,6 +858,258 @@ def body(k):
...
+            # compute self field and other coil field to find maximum field magnitude
+            # on the coil k
+            coil = constants["coilset"][0]
+            self_field, quad_points, centerline_mask = coil.compute_self_field(
+                constants["finite_build_grid"],
+                coil_frame=False,
+                params=x,  # just this coil's parameters
+            )
+
+            quad_points = jnp.where(
```
dpanici, line 1048: what is this line doing?

desc/objectives/_coils.py
```diff
@@ -850,6 +858,258 @@ def body(k):
...
+            # compute field from other coils and subtract the coil computation to
+            # prevent double counting
+            other_field = constants["coilset"].compute_magnetic_field(
+                quad_points,
+                params=params,  # all coils' parameters
+                basis="xyz",
+                source_grid=constants["centerline_grid"],
+                chunk_size=self._bs_chunk_size,
+            )
+            other_field -= coil.compute_magnetic_field(
```
dpanici, line 1061: This is technically a singular computation that you are adding, then subtracting right? Have you found this to be stable numerically?

desc/objectives/_coils.py
```diff
@@ -850,6 +858,258 @@ def body(k):
...
+        Returns
+        -------
+        f : array of floats
+            Maximum field on coil for each coil in the coilset.
+
+        """
+        if constants is None:
+            constants = self.constants
+
+        def body(dummy, x):
```
dpanici, line 1038: small nit but can you make this like "this_coil_param" or "param_one_coil" instead of "x"? since "x" is used in coil stuff a lot

PR #1422, opened 2024-12-03
new profile class `PoweredProfile`

ddudt:
Adds a new profile class to raise profiles to a power, of the form: $p_{new} = p_{old}^{\alpha}$. This class is conceptually similar to `ScaledProfile`, etc. 

This extends the existing profile arithmetic operations we already offer. One example use case is if you want to create kinetic profiles from an existing pressure profile of the form $n = p^{\alpha}$, $T = p^{1-\alpha}$.


desc/profiles.py
```diff
@@ -244,6 +244,17 @@ def __sub__(self, x):
         """Subtract another profile from this one."""
         return self.__add__(-x)
 
+    def __pow__(self, x):
+        """Raise this profile to a power."""
+        if np.isscalar(x):
+            return PoweredProfile(x, self)
+        else:
+            raise NotImplementedError()
```
f0uriest, line 252: the pythonic thing here would be `return NotImplemented`

docs/api_equilibrium.rst
```diff
@@ -53,12 +53,15 @@ profiles together by addition, multiplication, or scaling.
     :recursive:
     :template: class.rst
 
-    desc.profiles.PowerSeriesProfile
-    desc.profiles.SplineProfile
+    desc.profiles.HermiteSplineProfile
```
f0uriest, line 56: I'd prefer to keep the old order here, which was meant to be ordered by importance/how often they get used. The one in `api.rst` is a full alphabetical list.

desc/profiles.py
ddudt: I followed the format of our existing dunder methods. If we make this change to all of them, then doing stuff like `PowerSeriesProfile() + 4` will raise a `TypeError` instead of the `NotImplementedError`, and we would have to change our tests accordingly. I'm not sure that's what we want.

desc/profiles.py
```diff
@@ -335,6 +346,128 @@ def __repr__(self):
         return s
 
 
+class PoweredProfile(_Profile):
```
f0uriest, line 349: not crazy about the name. I think the proper adjective here would be "exponentiated" but thats pretty clunky. maybe just `PowerProfile`? That would fit with `SumProfile` and `ProductProfile`

CHANGELOG.md
```diff
@@ -14,12 +14,12 @@ New Features
     * use of both this and the ``QuadraticFlux`` objective allows for REGCOIL solutions to be obtained through the optimization framework, and combined with other objectives as well.
 - Changes local area weighting of Bn in QuadraticFlux objective to be the square root of the local area element (Note that any existing optimizations using this objective may need different weights to achieve the same result now.)
 - Adds a new tutorial showing how to use``REGCOIL`` features.
+- Adds a new profile class ``PoweredProfile`` for rasing profiles to a power.
```
dpanici, line 17:
```suggestion
- Adds a new profile class ``PoweredProfile`` for raising profiles to a power.
```

desc/profiles.py
dpanici: `PowerProfile` I could see being mixed up with `PowerSeriesProfile` actually, maybe `ExponentProfile`?

dpanici: `RaiseProfile`

ddudt: I decided that `PowerProfile` is the least bad of all the options

f0uriest (approved):

dpanici (approved):

PR #1430, opened 2024-12-05
Using quadrature grid for the normalizing length a

### LINKED ISSUE #1424: Apparent mismatch between ballooning objective output and direct calculation of growth rates
setup:

```python

eq0 = desc.examples.get("HELIOTRON")

nalpha = 8  # Number of field lines

# Field lines on which to evaluate ballooning stability
alpha = np.linspace(0, np.pi, nalpha, endpoint=False)

# Number of toroidal transits of the field line
nturns = 3

# Number of point along a field line per transit
nzetaperturn = 200

rho=0.8

obj2 = desc.objectives.BallooningStability(
            eq=eq0,
            rho=rho,
            alpha=alpha,
            nturns=nturns,
            nzetaperturn=nzetaperturn,
            weight=1,
            target=0,
        )
obj2.build()
```

We expect the following 3 to give the same answer since the objective has target=0 and weight=1
```python
print(obj2.compute(eq0.params_dict, constants=obj2.constants))
print(obj2.compute_scaled_error(eq0.params_dict, constants=obj2.constants))
print(obj2.compute_scalar(eq0.params_dict, constants=obj2.constants))
```
which gives
```
221.8172472905935
[221.81724729]
221.81724729059223
```

which seems fine

But if we then compute the growth rates directly and apply the same reduction as in the objective function:

```python
# range of the ballooning coordinate zeta
zeta = np.linspace(-np.pi * nturns, np.pi * nturns, nzetaperturn*nturns)

grid = desc.grid.Grid.create_meshgrid(
)

# make sure its on the same points
np.testing.assert_allclose(alpha, obj2.constants['alpha'], atol=1e-14)
np.testing.assert_allclose(zeta, obj2.constants['zeta'], atol=1e-14)

ball_data0 = eq0.compute(["ideal ballooning lambda"], grid=grid)
lam = ball_data0["ideal ballooning lambda"]

# same as in obj2.compute
lambda0, w0, w1 = obj2.constants["lambda0"], obj2.constants["w0"], obj2.constants["w1"]

# Shifted ReLU operation
data = (lam - lambda0) * (lam >= lambda0)

results = w0 * np.sum(data) + w1 * np.max(data)
print(results)
```

gives 
```
14.325405153826495
```

which is way smaller. Am I missing something?

rahulgaur104:
Fixes #1424


tests/test_stability_funs.py
```diff
@@ -596,6 +596,8 @@ def test_ballooning_stability_eval():
         # different numerics than "ideal ballooning lambda" so that we can verify them
         # against one another
         psi_b = data01["Psi"][-1] / (2 * jnp.pi)
+        # Calculating a_N accurately requires a QuadratureGrid
+        # but for some reason it gives the correct value even with a LinearGrid here.
```
unalmis, line 600: `eq.compute` will automatically override the grid to the correct value in case the user forgets, which is why it gives the correct value even though you give it a lineargrid.

unalmis (approved):

tests/test_stability_funs.py
dpanici, line 600:
```suggestion
        # which is automatically accounted for inside of eq.compute
```

dpanici (changes requested):
Approved once comment is updated

unalmis (approved):

PR #1436, opened 2024-12-08
Add point B measurement error objective

dpanici:
- Adds objective for B measurements at points (outside the plasma, though this is not something I check for explicitly, it just does not make sense if plasma currents are present as the virtual casing principle will get the plasma contribution incorrect).
- Currently not added in the public objective API, as we may want to think of having an overrarching "MagneticDiagnostic" objective that tries to minimize the re-computation of the magnetic field (think, concatenate all of the eval coords for both point B measurements for B probes, as well as the points along the curves for flux/mirnov and partial/full rogowski loops, to better take advantage of vectorization. I will think of a possible API for this

- [x] change print value to be more intuitive for this objective (show error in matching targets, not just the compute value)


tests/test_objective_funs.py
```diff
@@ -3321,6 +3472,35 @@ def test_compute_scalar_resolution_omnigenity(self):
...
+        coords_Z = np.zeros_like(coords_R)
+        coords = np.vstack([coords_R, coords_phi, coords_Z]).T
+
+        f = np.zeros_like(self.res_array, dtype=float)
+        for i, res in enumerate(self.res_array):  # omnigenity needs higher res
+            eq.change_resolution(
+                L_grid=int(eq.L * res), M_grid=int(eq.M * res), N_grid=int(eq.N * res)
+            )
+            obj = ObjectiveFunction(
+                PointBMeasurement(
```
YigitElma, line 3494: need to add `target` too. I guess having a default 0 target for this objective is not too practical

desc/objectives/_reconstruction.py
```diff
@@ -0,0 +1,343 @@
...
+                target = target.reshape(measurement_coords.shape)
+                target = xyz2rpz_vec(
+                    target, x=measurement_coords[:, 0], y=measurement_coords[:, 1]
+                )
+                target = target.flatten()
+
+            measurement_coords = xyz2rpz(measurement_coords)
+
+        self._measurement_coords = measurement_coords
+        self._directions = directions
```
YigitElma, line 141: As a sanity check, might worth normalizing `directions`, or use normalized `directions` in the dot products.

dpanici: good catch, I had not thuoght of this

desc/objectives/_reconstruction.py
```diff
@@ -0,0 +1,343 @@
...
+        bounds=None,
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        deriv_mode="auto",
+        name="Magnetic-Point-Measurement",
+        jac_chunk_size=None,
+    ):
+        # TODO: I want the output to be like "measurement error", but that means
```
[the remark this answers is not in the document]
dpanici: yea but that may mess up stuff with target in the superclass objective logic

desc/objectives/_reconstruction.py
```diff
@@ -0,0 +1,339 @@
...
+                basis="rpz",
+                source_grid=self._field_grid,
+            )
+
+        self._constants = {
+            "quad_weights": 1.0,
+            "measurement_coords": self._measurement_coords,
+            "equil_transforms": eq_transforms,
+            "equil_profiles": eq_profiles,
+            "directions": self._directions,
```
YigitElma, line 231: Having `measurements` and `directions`  in constants can cause unexpected behavior. For example, if `self._field_fixed`, `B_from_field` will be computed for the given direction and measurement coords, then if you call `obj.compute(..,.., new_constants_with_new_coords_directions)`, will use the data that is computed for something else. So, having them in constants might be dangerous (and really not necessary)

desc/objectives/_reconstruction.py
```diff
@@ -0,0 +1,339 @@
...
+        }
+        if self._field_fixed:
+            self._constants["B_from_field"] = (
```
YigitElma, line 234: I still think moving the dot product to the compute function itself is better, but I'm fine with both

desc/objectives/_reconstruction.py
```diff
@@ -0,0 +1,346 @@
...
+        Grid containing the nodes to evaluate field source.
+        Defaults to the default for the
+        given field, see the docstring of the field object for the specific default.
+    vc_source_grid : LinearGrid
+        LinearGrid to use for the (non-singular) integral over the virtual casing
+        principle current to calculate the flux loop contribution from the
+        plasma currents. Must have endpoint=False and sym=False and be linearly
+        spaced in theta and zeta, with nodes only at rho=1.0
+    basis : {"rpz","xyz"}
+        the basis used for ``measurement_coords``, ``directions`` and ``target``,
```
YigitElma, line 57:
```suggestion
        The basis used for ``measurement_coords``, ``directions`` and ``target``,
```

desc/objectives/_reconstruction.py
```diff
@@ -0,0 +1,346 @@
...
+        assumed to be "rpz" by default. if "xyz", will convert the input arrays
```
YigitElma, line 58:
```suggestion
        assumed to be "rpz" by default. If "xyz", will convert the input arrays
```

desc/objectives/_reconstruction.py
```diff
@@ -0,0 +1,346 @@
...
+        be calculated.
+    directions : array (n,3), optional
+        the directions of the measured B field for each of the n sensors, i.e. if a
+        sensor is measuring the poloidal field and is located at (R,phi,Z) = (1,0,0),
+        its ``direction`` might be [0,0,+/- 1]. If not passed, will default to instead
+        comparing the entire magnetic field vector at the points passed in.
+
+    """
+
+    __doc__ = __doc__.rstrip() + collect_docs()
```
YigitElma, line 75: You probably need to override the `target`

desc/objectives/_reconstruction.py
```diff
@@ -0,0 +1,346 @@
...
+                )
+                target = target.flatten()
+
+            measurement_coords = xyz2rpz(measurement_coords)
+        self._measurement_coords = measurement_coords
+        self._directions = directions
+        self._vacuum = vacuum
+        self._sheet_current = hasattr(eq.surface, "Phi_mn")
+        things = [eq]
+        self._field_fixed = field_fixed
```
YigitElma, line 158: `field_fixed` and `vacuum` result in a trivial thing, right? Maybe throw a warning since objective value will never change for that combination

YigitElma: Can vacuum equilibrium have sheet current? I feel like if `vacuum=True` then `eq` don't need to be in the `things`

YigitElma (changes requested):
Need to add the objective to docs

PR #1439, opened 2024-12-09
Add dependabot config to manage dependency versions

f0uriest:
Rather than us having to manually check compatible versions and update things accordingly, this PR pins the maximum versions of our dependencies to their current version, and enables githubs [dependabot](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide) to regularly update the requirements file. 

Right now its set to check for new versions every monday and automatically opens a PR to update `requirements.txt` if new versions are available. This will then trigger our regular CI so we can see if the new version works correctly, and make any necessary fixes if not before merging.

Note this only pins the main requirements, not dev-requirements, which I figured we don't need to keep up to date as rigorously so we can manage that ourselves.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -55,7 +55,7 @@ Option 2: Using conda to install packages (this will only install DESC + JAX wit
 .. code-block:: sh
 
     conda env create --file requirements_conda.yml
-    # option B: with developer requirements (if you want to run tests)
+    # optionally: install developer requirements (if you want to run tests)
     conda install --file devtools/dev-requirements_conda.yml
 
     # to add DESC to your Python path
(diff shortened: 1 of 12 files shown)


docs/installation.rst
```diff
@@ -54,11 +54,9 @@ Option 2: Using conda to install packages (this will only install DESC + JAX wit
 
 .. code-block:: sh
 
-    # only need to do one of these conda env create commands, not both
-    # option A: without developer requirements
     conda env create --file requirements_conda.yml
     # option B: with developer requirements (if you want to run tests)
```
dpanici, line 58:
```suggestion
    # optionally: install developer requirements (if you want to run tests)
```

dpanici (approved):

rahulgaur104 (approved):

YigitElma (approved):
Btw I am also ok with the idea to completely remove conda requirement files. With this action , we will need to change these files a lot, and it will probably be annoying.

PR #1440, opened 2024-12-09
Debugging singular integrals

### LINKED ISSUE #1519: FFTInterpolator restriction to avoid frequency spectrum truncation
The interpolation algorithm used by the `FFTInterpolator` to integrate kernels by shifting from the rectangular Fourier grid to the polar (Fourier-Legendre) grids surrounding each evaluation point will lose frequency content of the signal anytime there are less evaluation points than than source points.

The reason the existing tests passed with the given tolerance is that the equilibria used in those tests do not have sufficiently high frequency content, and the resulting errors from loss of high frequencies were interpreted as approximation error (e.g. force balance error or configurations that are imperfect vacuum configurations due to some current, etc.).

- [x] Update ``FFTInterpolator`` with this restriction. The other, less performant, Fourier interpolator `DFTInterpolator` does not have this limitation.
- [x] Fix tests/example code so that evaluation grid has resolution at least that of the source grid.

### LINKED ISSUE #1522: DFTInterpolator for singular integral gives bad results
When the `DFTInterpolator` is used in the singular integral routine the results are not correct, and also not deterministic on my machine. Running this test multiple times gives different results. [Attached](https://github.com/user-attachments/files/18454163/stochastic_interpolator.txt) is example output.

```python
@pytest.mark.unit
def test_singular_integral_vac_estell(self):
    """Test calculating Bplasma for vacuum estell, which should be near 0."""
    eq = get("ESTELL")
    grid = LinearGrid(M=25, N=25, NFP=eq.NFP)
    keys = [
        "K_vc",
        "B",
        "|B|^2",
        "R",
        "phi",
        "Z",
        "e^rho",
        "n_rho",
        "|e_theta x e_zeta|",
    ]
    data = eq.compute(keys, grid=grid)
    k = min(grid.num_theta, grid.num_zeta * grid.NFP)
    s = k - 1
    q = k // 2 + int(np.sqrt(k))
    interpolator = DFTInterpolator(grid, grid, s, q)
    Bplasma = virtual_casing_biot_savart(data, data, interpolator, loop=True)
    # need extra factor of B/2 bc we're evaluating on plasma surface
    Bplasma += data["B"] / 2
    Bplasma = np.linalg.norm(Bplasma, axis=-1)
    # scale by total field magnitude
    B = Bplasma / np.linalg.norm(data["B"], axis=-1).mean()
    np.testing.assert_allclose(B, 0, atol=0.005)
```

unalmis:
- [x]  Resolves #1522
- [x] Resolves #1519 
- [x] Reduces memory usage of `DFTInterpolator`.
- [x] Adds `chunk_size` option to free boundary.
- [x] #1536 
- [x] Add guards to prevent  choosing `st`, `sz` larger than partition size for the integration.
- [x] Applies #1530 to the `Bounce2D` bounce integral methods and refactors the code to reuse the Vandermonde interpolation matrix while retaining JIT's ability to fuse the matrix multiply. Benchmarking shows this
  - Doubles the speed performance of computing `effective ripple` and `Gamma_c`.
  - Halves the memory of reverse mode bounce integrals. Forward mode memory increased.
- Make singular partition shape circle in real space moved to other pull request.

### motivation for pr (resolved by the above updates)
I have some tests  to incrementally test what feature of shaping causes the issue
  - (torision, rotation, 3d, etc.).
  - non constant harmonic function

That should help search for the issue. For now it seems 3D shaping fails the green's id test for $\phi = 1$. Likewise, the `FFT` interpolators and `DFT` interpolators disagree in their output, indicating some bug in one or either, although the implementation here and on interpax looks correct.


dpanici:
Might be related, but if you change the source grid NFP, you get different convergence/no convergence for the test for the axisymmetric case

dpanici:
Maybe check basis being used?

dpanici:
https://arxiv.org/pdf/2404.02799
could be useful as another test, to check an axisymmetric plasma's Bfield against this implementation of virtual casing for axisymmetry. I will reach out to the author to ask if they have the code/examples

desc/integrals/singularities.py
```diff
@@ -7,12 +7,20 @@
...
+from desc.batching import batch_map
 from desc.compute.geom_utils import rpz2xyz, rpz2xyz_vec, xyz2rpz_vec
 from desc.grid import LinearGrid
 from desc.io import IOAble
-from desc.utils import isalmostequal, islinspaced, safediv, safenorm
+from desc.utils import safediv, safenorm, warnif
+
+
+def _get_default_sq(grid):
+    k = max(min(grid.num_theta, grid.num_zeta * grid.NFP), 2)
```
unalmis, line 20: note to self: check if expanding polar grid beyond a single period of the source grid is fine to do here. check both in general and the edge case where `source_grid.NFP != eval_grid.NFP` for axissymetric device.

tests/test_integrals.py
```diff
@@ -617,42 +621,38 @@ def test_singular_integral_greens_id(self):
...
-            q = qs[i]
-            interpolator = FFTInterpolator(eval_grid, source_grid, s, q)
-
-            err = singular_integral(
-                eval_data,
-                source_data,
-                "nr_over_r3",
-                interpolator,
-                loop=True,
+        Nu = [13, 13, 13, 19, 19, 25, 37]
```
dpanici, line 624: @dpanici check for test with changing NFP

desc/integrals/singularities.py
unalmis, line 1: I recommend clicking the settings icon and switching the diff view to `split`

desc/integrals/singularities.py
```diff
@@ -327,104 +427,94 @@ def __call__(self, f, i):
...
+            ``self.fourier``. Default is false.
+        vander : jnp.ndarray
+            Cached value for ``self.vander_polar(i)``.
 
         Returns
         -------
         fi : ndarray
             Source data interpolated to ith polar node.
-        """
-        return self._mat[:, i] @ f
```
unalmis, line 336: JAX would not perform this matrix multiplication correctly.

desc/integrals/singularities.py
unalmis: I checked this. I think there are subtleties that are only resolved by generalizing the coordinate transformation for the singular integral as:  rectangular toroidal grid to elliptical polar grid. (Before it was a circular polar grid).

Therefore, we now request two parameters from the user $s_{\theta}$ and $s_{\zeta}$ instead of just the single parameter $s$ ($M$ in Mahlotora).

unalmis: I am holding off from editing the default settings for the free boundary stuff to take two parameters $s_{\theta}$ and $s_{\zeta}$ discussed above until you guys give a thumbs up on this.

desc/basis.py
unalmis, line 1: I changed some of the mode number constructors. They are the same as before. I changed this while debugging 1522 and decided to keep the modifications because it uses the simpler conventions we use elsewhere in the code.

desc/integrals/singularities.py
```diff
@@ -2,17 +2,143 @@
...
+    Since kernels are 2π periodic, the choice for s₂ should be multiplied by NFP.
+    Then the support lies on an s₁ × s₂ subset of the full domain. For large NFP
+    devices such as Heliotron or tokamaks it is typical that s₁ ≪ s₂.
+
+    Parameters
+    ----------
+    theta, zeta : jnp.ndarray
+        Coordinates of points to evaluate partition function η₀(θ,ζ).
+    theta0, zeta0 : jnp.ndarray
+        Origin (θ₀,ζ₀) where partition η₀ unity.
```
unalmis, line 88:
```suggestion
        Origin (θ₀,ζ₀) where partition η₀ is unity.
```

desc/basis.py
```diff
@@ -261,9 +261,9 @@ def _get_modes(self, L):
             Each row is one basis function with modes (l,m,n).
 
         """
-        l = np.arange(L + 1).reshape((-1, 1))
-        z = np.zeros((L + 1, 2))
-        return np.hstack([l, z])
+        l = np.arange(L + 1)
+        z = np.zeros_like(l)
+        return np.array([l, z, z]).T
```
unalmis, line 266: I changed some of the mode number constructors. They are the same as before. I changed this while debugging github issue 1522 and decided to keep the modifications because it uses the simpler conventions we use elsewhere in the code.

desc/integrals/singularities.py
unalmis: I know that sounds suspicious, since it is a fundamental operation but that is what is observed. At first I presumed the pseudoinverse had a numerical issue, but even after replacing the `Ainv` matrix with a fourier transform the issue is persisted. Only once the `B` matrix was computed in chunks (self.polar_vander(i)) and multiplied against the Fourier coefficients was the issue resolved.

Since the $B$ matrix had to be computed in chunks to resolve the issue, it didn't make sense anymore to precompute the $A_{inv}$ matrix because evaluating the DFT interpolator then requires two matrix multiplies and then  $B[i] @ A_{inv} @ f$ which is basically the same cost as just $B[i] @ fourier_transform(f)$.

desc/integrals/singularities.py
f0uriest:
```python

def _chi(rho):
    """Partition of unity function. Eq 39 in [2]."""
    return jnp.exp(-36 * jnp.abs(rho) ** 8)

def _rho(theta, zeta, theta0, zeta0, dtheta, dzeta, s):
    """Polar grid radial coordinate. Argument of Chi in eq. 36 in [2]."""
    dt = abs(theta - theta0)
    dz = abs(zeta - zeta0)
    dt = jnp.minimum(dt, 2 * np.pi - dt)
    dz = jnp.minimum(dz, 2 * np.pi - dz)
    return 2 / s * jnp.sqrt((dt / dtheta) ** 2 + (dz / dzeta) ** 2)

source_grid = LinearGrid(M=5, N=5, NFP=10)
eval_grid = LinearGrid(M=5, N=5, NFP=1)

source_theta = source_grid.nodes[:, 1]
source_zeta = source_grid.nodes[:, 2]
source_dtheta = source_grid.spacing[:, 1]
source_dzeta = source_grid.spacing[:, 2] / source_grid.NFP
eval_theta = jnp.asarray(eval_grid.nodes[:, 1])
eval_zeta = jnp.asarray(eval_grid.nodes[:, 2])
h_t = jnp.mean(source_dtheta)
h_z = jnp.mean(source_dzeta)

unique_theta = source_theta[source_grid.unique_theta_idx]
unique_zeta = source_zeta[source_grid.unique_zeta_idx]


i = 200
k = min(source_grid.num_theta, source_grid.num_zeta * source_grid.NFP)
s = k - 1
eta = []

for j in range(source_grid.NFP):
    
    r = _rho(
        source_theta,
        (source_zeta + j * 2 * np.pi / source_grid.NFP) % (
                2 * np.pi
            ),  # to account for different field periods
        eval_theta[i],
        eval_zeta[i],
        h_t,
        h_z,
        s,
    )
    e = _chi(r)
    eta.append(source_grid.meshgrid_reshape(e, "rtz").squeeze())
eta = np.concatenate(eta, axis=1)

fig, ax = plt.subplots(2)
ax[0].contourf(np.concatenate([(unique_zeta + j * 2 * np.pi / source_grid.NFP) % (
                2 * np.pi
            ) for j in range(source_grid.NFP)]), unique_theta, eta)
ax[0].axis("equal")
ax[0].set_xlabel("zeta")
ax[0].set_ylabel("theta")

ax[1].contourf(eta)
ax[1].axis("equal")
ax[1].set_xlabel("iz")
ax[1].set_ylabel("it")
```
image

Top one is what the partitioned region looks like in theta/zeta space, bottom is what it looks like in "point space". Currently we end up with a circle in "point space" though its an ellipse in theta/zeta. I think both might be a bit wrong though, I think what we actually want is a circle in real space, since the singularity is in real space. This suggests that we also want to account for the aspect ratio of the device along with the anisotropy in the grid, or we could change things such that `source_grid.num_zeta * source_grid.NFP / source_grid.num_theta` is ~= aspect ratio?

unalmis: The "point space" is polar coordinate space right, $\rho, \omega$? That will always be a circle by construction.

>  though its an ellipse in theta/zeta.  I think both might be a bit wrong though, I think what we actually want is a circle in real space, since the singularity is in real space. 

Yes, exactly. In this PR I have generalized the coordinate transformation to be a circle in $\theta, \zeta$ space. With claim with $s_{\theta}$ = `source_grid.num_theta` and $s_{\zeta}$ = `source_grid.num_zeta * source_grid.NFP` that the partitioned region in $\theta, \zeta$ space is the desired circle.

>  the singularity is in real space

I agree that it would be even better to have the partitioned region be a circle in real space. I think the way to do that is to choose $s_{\theta}$ and $s_{\zeta}$ appropriately, but I don't know the the missing factors in the relation $s_{\zeta} ~ NFP * s_{\theta}$. (We shouldn't touch the grid resolution).

desc/integrals/singularities.py
dpanici: if this is reproducible on a smaller MWE, it would be a good thing to make an issue for in the JAX repo

desc/integrals/bounce_integral.py
unalmis, line 1: @rahulgaur104 mentioned there was some regression in memory of bounce integrals after perl mutter's recent maintenance.. I think it might be related to #1484 . Still I took a look and managed the following:
 
This pr
- Halves the memory of reverse mode AD of bounce integrals. The example optimization notebook has peak memory of 17gb reverse mode while on master it had 32gb.
- Doubles the compute speed of bounce integrals  
- Doubles the memory of forward mode bounce integrals

Not much more to do besides the issues listed on the objective docstring, but those are not my priority right now.

tests/inputs/master_compute_data_rpz.pkl
unalmis, line 1: https://github.com/PlasmaControl/DESC/pull/1440/commits/7c3720ca62a5ea996633a133c45c4942e7a91371

desc/basis.py
```diff
@@ -400,6 +399,9 @@ def evaluate(
         -------
         y : ndarray, shape(num_nodes,num_modes)
             Basis functions evaluated at nodes.
+            The Vandermonde matrix when ``modes is None`` is
+            given by ``y.reshape(-1,2*N+1)`` and is ordered
+            [sin(Nx), ..., sin(x), 1, cos(x), ..., cos(Nx)].
```
YigitElma, line 404:
```suggestion
            [
              [sin(Nx_1), ..., sin(x_1), 1, cos(x_1), ..., cos(Nx_1)]
              [sin(Nx_2), ..., sin(x_2), 1, cos(x_2), ..., cos(Nx_2)]
              ...
              [sin(Nx_K), ..., sin(x_K), 1, cos(x_K), ..., cos(Nx_K)]
            ]
            where K = num_nodes
```

desc/integrals/_interp_utils.py
```diff
@@ -425,10 +463,9 @@ def idct_non_uniform(xq, a, n, axis=-1):
         Real function value at query points.
 
     """
-    # |a| << |xq|, so move a instead
-    a = jnp.moveaxis(a, axis, -1)
-    # Same as Clenshaw recursion ``chebval(xq,a,tensor=False)`` but better on GPU.
```
YigitElma, line 430: This comment seems useful, maybe keep it?

desc/objectives/_fast_ion.py
```diff
@@ -61,9 +61,9 @@ class GammaC(_Objective):
     eq : Equilibrium
         ``Equilibrium`` to be optimized.
     grid : Grid
-        Optional, tensor-product grid in (ρ, θ, ζ) with uniformly spaced nodes
-        (θ, ζ) ∈ [0, 2π) × [0, 2π/NFP). Number of poloidal and toroidal nodes
-        preferably rounded down to powers of two.
+        Tensor-product grid in (ρ, θ, ζ) with uniformly spaced nodes
```
YigitElma, line 64: This is still "optional" parameter, right?

desc/objectives/_free_boundary.py
```diff
@@ -12,10 +10,21 @@
...
+    errorif,
+    parse_argname_change,
+    setdefault,
+    warnif,
+)
+
+from ..integrals.singularities import _get_default_params
 from .normalization import compute_scaling_factors
 
+# TODO: Often eval grid and source grid are the same, so don't recompute
```
YigitElma, line 25: Can you create and link the issue here?

desc/objectives/_neoclassical.py
```diff
@@ -60,9 +60,9 @@ class EffectiveRipple(_Objective):
     eq : Equilibrium
         ``Equilibrium`` to be optimized.
     grid : Grid
-        Optional, tensor-product grid in (ρ, θ, ζ) with uniformly spaced nodes
-        (θ, ζ) ∈ [0, 2π) × [0, 2π/NFP). Number of poloidal and toroidal nodes
-        preferably rounded down to powers of two.
+        Tensor-product grid in (ρ, θ, ζ) with uniformly spaced nodes
```
YigitElma, line 63: Again, if this is still optional, I would prefer to have it.

YigitElma (commented):
I will take a look at it again, for the mean time some small comments.

desc/basis.py
unalmis: I can update DoubleFourier with this; but note that x is a constant in this array, so we shouldn't label $x$ with a subscript.

unalmis: I think that's why it's technically called the psuedo-vandermonde matrix?

YigitElma: By $x_i$ I mean each grid point, so in this case different angles. Isn't each row of the matrix is different? Then when you multiply this matrix with the Fourier coefficients, you get the value of the function at the point.

unalmis: For each point you have a matrix of size N^2. For k points it's  tensor of size N^2 k

unalmis: Oh I see what you are saying

YigitElma: Yes, I think we are saying the same thing, in your case x is a vector of grid points, in mine I specifically write them as individual components of the vector which shows the grid points. Just wanted to express it as matrix with scalar values.

desc/objectives/_neoclassical.py
unalmis: Every parameter is optional though.

desc/objectives/_free_boundary.py
unalmis: just did the pr instead

desc/integrals/singularities.py
```diff
@@ -433,91 +554,110 @@ def eval_pt_loop(i, fj):
...
         }
-
-        # can't interpolate phi directly since its not periodic, so we interpolate
-        # omega and add it back in
-        source_data_polar["zeta"] = zeta_i
-        source_data_polar["theta"] = theta_i
-        if "phi" in keys:
-            source_data_polar["phi"] = zeta_i + source_data_polar["omega"]
+        # The (θ, ζ) coordinates at which the maps above were evaluated.
+        # For FFT interpolator on functions with toroidal variation used on
```
unalmis, line 621: Commit this before approving the PR
```suggestion
        # For FFT interpolator on functions with no toroidal variation used on
```

desc/integrals/singularities.py
```diff
@@ -433,91 +554,110 @@ def eval_pt_loop(i, fj):
...
+        # grids of different NFP, we still use eval grid to compute coordinates.
+        # These are the points the above maps were evaluated, but the values
```
unalmis, line 623:
```suggestion
        # In that case, the eval grid points are not where the maps were interpolated,
```

desc/integrals/singularities.py
```diff
@@ -433,91 +554,110 @@ def eval_pt_loop(i, fj):
...
+        # of the vector coordinates of the orthonormal polar basis are the same.
```
unalmis, line 624:
```suggestion
        # but the components of maps in coordinates of the orthonormal polar basis are still the same.
```

desc/basis.py
unalmis: do you still want this, i made the x bold for vector

unalmis:
It was requested that I move the bounce integral memory improvement to #1537 , so doing that now.

desc/objectives/_neoclassical.py
YigitElma: I think in most of the objectives docs, we write optional next to the data type of the argument, like `grid: Grid, optional`. But not too important just for consistency

desc/integrals/basis.py
```diff
@@ -227,14 +227,22 @@ def harmonics(self):
         Transform Fourier interpolant harmonics to Nyquist trigonometric
         interpolant harmonics so that the coefficients are all real.
 
+        The order of the returned coefficient array
+        matches the Vandermonde matrix formed by an outer
+        product of Fourier and Chebyshev matrices with order
+        [sin(k𝐱), ..., sin(𝐱), 1, cos(𝐱), ..., cos(k𝐱)]
+        ⊗ [T₀(𝐲), T₁(𝐲), ..., Tₙ(𝐲)]
+
+        When ``self.X`` is even the sin(k𝐱) coefficient is zero and is excluded.
```
dpanici, line 236:
```suggestion
        When ``self.X`` is even the sin(k𝐱) coefficients are zero and is excluded.
```
right?

desc/integrals/singularities.py
```diff
@@ -7,12 +7,177 @@
...
+    # TODO: Account for real space grid anisotropy by choosing
+    #   ratio = |e_zeta|/|e_theta|
+    #   ratio_avg = surface_average(ratio, jacobian=|e_theta x e_zeta|)
+    #   st/(sz*NFP) (at evaluation point i) = mean (ratio(i), ratio_avg).
+    #   Then we have ~circle in real space around each singular point.
+    #   The grid resolution can then still be chosen independently according
+    #   to the frequency content of R, Z, λ to net the best function approximation.
+    Nt = grid.num_theta
+    Nz = grid.num_zeta * grid.NFP
+    st = int(min(1 + jnp.sqrt(Nt), Nt))
```
dpanici, line 177: going to ask my stupid question but why this particular choice for `st`? it is basically always the former option (1 + sqrt(Nt)) unless Nt is 1 or 2

desc/objectives/_free_boundary.py
```diff
@@ -364,6 +377,13 @@ class BoundaryError(_Objective):
     loop : bool
         If True, evaluate integral using loops, as opposed to vmap. Slower, but uses
         less memory.
+    sz : int
```
dpanici, line 380: can this go next to `st`? or do we want it last bc it should be set by defauly or something?

desc/integrals/basis.py
unalmis: sure, feel free to press commit before approval

desc/objectives/_free_boundary.py
unalmis: That would be better to change after #1458 . Don't want people to mix q and sz

desc/integrals/singularities.py
unalmis: It can't be empty region (>=0) more than full domain (<=Nt) and Malhotra et al. chose the sqrt

desc/vmec_utils.py
```diff
@@ -273,6 +274,83 @@ def ptolemy_linear_transform(desc_modes, vmec_modes=None, helicity=None, NFP=Non
...
+        [sin(k𝐱), ..., sin(𝐱), 1, cos(𝐱), ..., cos(k𝐱)].
+        When ``n`` is even the sin(k𝐱) coefficient is zero and is excluded.
+
+    """
+    is_even = (n % 2) == 0
+    # sin(nx) coefficients
+    an = -2 * jnp.flip(
+        take(
+            a.imag,
+            jnp.arange(1, a.shape[axis] - is_even),
```
dpanici, line 305: I don't follow this logic completely, so when `is_even` is True, are there any sin coefficients?

unalmis: When `n` is even, instead of returning coefficients that can be dotted with
`[sin(k𝐱), ..., sin(𝐱), 1, cos(𝐱), ..., cos(k𝐱)]`
we return
`[sin([k-1]𝐱), ..., sin(𝐱), 1, cos(𝐱), ..., cos(k𝐱)]`
because the coefficient for `sin(k𝐱)` is always zero in that case. By doing there are always the same number of returned coefficients `n` as there were interpolated parameters `n`.

tests/test_integrals.py
dpanici: Maybe you can check what I am misunderstanding/doing wrong with this test. I am seeing how the convergence of this greens ID error goes for the Equilibrium with higher NFP. 

Convergence is good with NFP=1. NFP=19 and changing nothing else about the test is bad/no convergence, as is NFP=19 and using _default_params. 

Does `Nu` need to be higher in the higher NFP case?


```python

from functools import partial

import numpy as np
import pytest
from jax import grad
from numpy.polynomial.chebyshev import chebinterpolate, chebroots
from numpy.polynomial.legendre import leggauss
from scipy import integrate
from scipy.interpolate import CubicHermiteSpline
from scipy.special import ellipe, ellipkm1
# from tests.test_interp_utils import _f_1d, _f_1d_nyquist_freq, _f_2d, _f_2d_nyquist_freq
# from tests.test_plotting import tol_1d

from desc.backend import jnp, vmap
from desc.basis import FourierZernikeBasis
from desc.equilibrium.coords import get_rtz_grid
from desc.examples import get
from desc.grid import ConcentricGrid, Grid, LinearGrid, QuadratureGrid
from desc.integrals import (
    Bounce1D,
    Bounce2D,
    DFTInterpolator,
    FFTInterpolator,
    line_integrals,
    singular_integral,
    surface_averages,
    surface_integrals,
    surface_integrals_transform,
    surface_max,
    surface_min,
    surface_variance,
    virtual_casing_biot_savart,
)
from desc.integrals._bounce_utils import (
    _check_bounce_points,
    _get_extrema,
    bounce_points,
)
from desc.integrals._interp_utils import fourier_pts
from desc.integrals.basis import FourierChebyshevSeries
from desc.integrals.quad_utils import (
    automorphism_sin,
    bijection_from_disc,
    chebgauss1,
    chebgauss2,
    get_quadrature,
    grad_automorphism_sin,
    grad_bijection_from_disc,
    leggauss_lob,
    tanh_sinh,
)
from desc.integrals.singularities import (
    _get_default_params,
    _get_quadrature_nodes,
    _kernel_nr_over_r3,
)
from desc.integrals.surface_integral import _get_grid_surface
from desc.transform import Transform
from desc.utils import dot, errorif, safediv

# @pytest.mark.unit
#     def test_singular_integral_greens_id(self):
"""Test high order singular integration using greens identity.

Any harmonic function can be represented as the sum of a single layer and double
layer potential:

Φ(r) = -1/2π ∫ Φ(r) n⋅(r-r')/|r-r'|³ da + 1/2π ∫ dΦ/dn 1/|r-r'| da

If we choose Φ(r) == 1, then we get

1 + 1/2π ∫ n⋅(r-r')/|r-r'|³ da = 0

So we integrate the kernel n⋅(r-r')/|r-r'|³ and can benchmark the residual.

"""



eq = Equilibrium(NFP=1)
Nu = [13, 13, 13, 19, 19, 25, 37]
Nv = [30, 45, 60, 90, 120, 150, 240]
ss = [13, 13, 13, 19, 19, 25, 37]
qs = [10, 12, 12, 16, 18, 20, 24]
es = [0.4, 2e-2, 3e-3, 5e-5, 4e-6, 1e-6, 1e-9]
errs = []
for i in range(len(Nu)):
    grid = LinearGrid(M=Nu[i] // 2, N=Nv[i] // 2, NFP=eq.NFP)
    interpolator = FFTInterpolator(grid, grid, ss[i], ss[i], qs[i])
    data = eq.compute(
        _kernel_nr_over_r3.keys + ["|e_theta x e_zeta|"], grid=grid
    )
    err = singular_integral(data, data, "nr_over_r3", interpolator, loop=i != 0)
    np.testing.assert_array_less(np.abs(2 * np.pi + err), es[i])
    errs.append(np.mean(np.abs(2 * np.pi + err)))

eq = Equilibrium(NFP=19)
errsNFP19 = []
for i in range(len(Nu)):
    grid = LinearGrid(M=Nu[i] // 2, N=Nv[i] // 2, NFP=eq.NFP)
    interpolator = FFTInterpolator(grid, grid, ss[i], ss[i], qs[i])
    data = eq.compute(
        _kernel_nr_over_r3.keys + ["|e_theta x e_zeta|"], grid=grid
    )
    err = singular_integral(data, data, "nr_over_r3", interpolator, loop=i != 0)
    # np.testing.assert_array_less(np.abs(2 * np.pi + err), es[i])
    errsNFP19.append(np.mean(np.abs(2 * np.pi + err)))

errsNFP19_with_default = []
for i in range(len(Nu)):
    grid = LinearGrid(M=Nu[i] // 2, N=Nv[i] // 2, NFP=eq.NFP)
    st, sz, q = _get_default_params(grid)
    interpolator = FFTInterpolator(grid, grid, st, sz, q)
    data = eq.compute(
        _kernel_nr_over_r3.keys + ["|e_theta x e_zeta|"], grid=grid
    )
    err = singular_integral(data, data, "nr_over_r3", interpolator, loop=i != 0)
    # np.testing.assert_array_less(np.abs(2 * np.pi + err), es[i])
    errsNFP19_with_default.append(np.mean(np.abs(2 * np.pi + err)))

plt.figure()
total_N = np.array(Nu) * np.array(Nv)
plt.plot(total_N, errs,".-",label="NFP=1")
plt.plot(total_N, errsNFP19,".-",label="NFP=19 same params as NFP=1")
plt.plot(total_N, errsNFP19_with_default,".-",label="NFP=19 using _default_params")


plt.yscale("log")
plt.legend()
# eq = get("W7-X")
# Nu = 100
# Nv = 100
# es = 6e-7
# grid = LinearGrid(M=Nu // 2, N=Nv // 2, NFP=eq.NFP)
# st, sz, q = _get_default_params(grid)
# interpolator = FFTInterpolator(grid, grid, st, sz, q)
# data = eq.compute(_kernel_nr_over_r3.keys + ["|e_theta x e_zeta|"], grid=grid)
# err = singular_integral(data, data, "nr_over_r3", interpolator, loop=True)
# np.testing.assert_array_less(np.abs(2 * np.pi + err), es)
```

image

unalmis: You are not doing anything wrong. The issue is the convergence of the integration depends on the shape and size of the support of the singular region is assigned. When using the default params on `master` (`_get_default_params`), changing NFP changes the elongates the shape and the size of that region. There is also some lower order effects that I'm chatting with Rory about.

This is especially confusing for axissymetric configurations, because from the end user perspective the results should be indepenent of NFP. For comparison here is using the heuristic params, although note that this is not a circle in theta zeta space, because the heuristic params assume the grid resolution is the same (i.e. the scaling has to go with the parameter `t` mentioned in https://github.com/PlasmaControl/DESC/pull/1440#discussion_r1926070739).

image

f0uriest:
Another thing to check: with the changes to DFTInterpolator, is it still a lot slower? It might be competitive, especially since it allows you to use a coarser eval grid

tests/test_integrals.py
f0uriest: it seems like it is something with the aspect ratio of the grid, because if I increase NFP, but keep Nv*NFP the same, things seem to work ok:

image

Note that the heuristic params which attempt to compensate for the grid anisotropy seem to do worse...

Also strange that for NFP=1 and those values of Nu and Nv, the grid anisotropy varies quite a bit:
```python
[nv/nu for nv, nu in zip(Nv,Nu)]
```
```
[2.3076923076923075,
 3.4615384615384617,
 4.615384615384615,
 4.7368421052631575,
 6.315789473684211,
 6.0,
 6.486486486486487]
```

desc/integrals/singularities.py
```diff
@@ -433,91 +582,111 @@ def eval_pt_loop(i, fj):
...
+        on that eval point. Polar grids from other singularities have no effect.
+        """
+        # The ``FFTInterpolator`` interpolates to polar node i of the source grid, while
+        # the ``DFTInterpolator`` interpolates to polar node i of the eval   grid.
+        # When the source grid and eval grid have different NFP, the values taken
+        # by a function that is periodic with source grid NFP will in general
+        # be different at these points. For functions with no toroidal variation
+        # there will be no difference, and that is the only time the
+        # source grid and eval grid may have different NFP.
+        vander = interpolator.vander_polar(i)
```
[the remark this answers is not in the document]
unalmis: There is no recalculation because the vandermonde being constructed is different for each iteration i.

Computing it and caching it would consume more memory (and slow down evaluation because you now also need to index and select a matrix from a larger tensor, but that's not a dominant effect).

If you do multiple calls to singular integral on the same grids then sure caching would avoid recomputing then but it's probably not worth memory increase and prevents the compiler from optimizing the reduction of the array.

the other issue is precomputing it is what causes the results to be inaccurate in the linked issue

desc/integrals/singularities.py
```diff
@@ -327,104 +470,110 @@ def __call__(self, f, i):
...
-    dz = abs(zeta - zeta0)
-    dt = jnp.minimum(dt, 2 * np.pi - dt)
-    dz = jnp.minimum(dz, 2 * np.pi - dz)
-    return 2 / s * jnp.sqrt((dt / dtheta) ** 2 + (dz / dzeta) ** 2)
+        """
+        if not is_fourier:
+            f = self.fourier(f)
+        if vander is None:
+            vander = self.vander_polar(i)
+        return jnp.real(vander @ f)
```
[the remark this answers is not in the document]
unalmis: is fourier is always true in the routines we wrote to avoid recomputing the fourier transform at each iteration i  discussed below.

likewise with the vander. the reason i did not require the arguments is because i do want to break user code as this is a public function

desc/integrals/singularities.py
```diff
@@ -206,26 +389,27 @@ def __call__(self, f, i):
             Data at source grid points to interpolate.
         i : int
             Index of polar node.
+        is_fourier : bool
```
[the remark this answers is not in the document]
unalmis: to avoid recomputing fourier transforms

desc/integrals/singularities.py
```diff
@@ -108,96 +354,33 @@ class FFTInterpolator(_BIESTInterpolator):
...
-        source_theta = source_grid.nodes[:, 1].reshape(
-            (source_grid.num_zeta, source_grid.num_theta)
-        )
-        source_zeta = source_grid.nodes[:, 2].reshape(
-            (source_grid.num_zeta, source_grid.num_theta)
+    def __init__(self, eval_grid, source_grid, st, sz, q, **kwargs):
+        st = parse_argname_change(st, kwargs, "s", "st")
+        assert eval_grid.can_fft2, "Got False for eval_grid.can_fft2."
+        # Otherwise frequency spectrum is truncated.
+        assert eval_grid.num_theta >= source_grid.num_theta, (
```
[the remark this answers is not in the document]
unalmis: Ok but may I ask what is the use case for fitting a fourier series to an overdetermined system? Fourier basis is the basis that benefit least from such fitting and that's why people just use interpolation. 

with FFT the speed gain from truncation also isn't big.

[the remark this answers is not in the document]
unalmis: if there is no use case then an error is better

desc/integrals/singularities.py
```diff
@@ -433,91 +582,111 @@ def eval_pt_loop(i, fj):
...
-            source_data_polar["phi"] = zeta_i + source_data_polar["omega"]
+        # For FFT interpolator on functions with no toroidal variation used on
+        # grids of different NFP, we still use eval grid to compute below coordinates.
+        # In that case, the eval grid points are not where the maps were interpolated,
+        # but the components of maps in coordinates of the orthonormal polar basis are
+        # the same. Since the polar basis vectors between the evaluation point
+        # and the point that was interpolated to point in different directions,
+        # it is important to use eval grid to compute the coordinates below.
+        source_data_polar["theta"] = eval_theta + interpolator.shift_t[i]
+        source_data_polar["zeta"] = eval_zeta + interpolator.shift_z[i]
```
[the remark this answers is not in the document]
unalmis: you want the polar node to be a shift around the evaluation point.You also want the function to have been evaluated at the polar points around the evaluation point. But the FFT interpolator interpolates to the polar node around the source point. When NFP is different these are not the same function value unless there is no toroidal variation

unalmis:
> Another thing to check: with the changes to DFTInterpolator, is it still a lot slower? It might be competitive, especially since it allows you to use a coarser eval grid

It is competitive with courser evaluation grid. For example `test_singular_integral_vac_estell` with DFT with `evl_grid = LinearGrid(M=10, N=10, NFP=eq.NFP, sym=eq.sym)` takes about same time as FFT interpolator.

desc/integrals/singularities.py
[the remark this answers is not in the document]
f0uriest: Both interpolators should be interpolating from the source grid to the polar nodes around the evaluation points, which is I believe what they're doing. Yes technically the FFT interpolator only knows about the number of pts in the eval grid, but the checks for constructing the FFT interpolator ensure that knowing that is enough to know where the eval pts are and interpolate to them correctly. My main point is that as far as the rest of the code is concerned, the interpolators should be interchangeable, any logic specific to one or the other should only be in those classes, so this comment is kind of confusing

desc/integrals/singularities.py
```diff
@@ -433,91 +582,111 @@ def eval_pt_loop(i, fj):
...
-        zeta_i = eval_zeta + dz
+    def polar_pt(i):
+        """See sec 3.2.2 of [2].
 
-        # data interpolated to each eval pt offset by dt,dz
+        Evaluate the effect from a single polar node around each eval point
+        on that eval point. Polar grids from other singularities have no effect.
+        """
+        # The ``FFTInterpolator`` interpolates to polar node i of the source grid, while
+        # the ``DFTInterpolator`` interpolates to polar node i of the eval   grid.
```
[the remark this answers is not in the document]
f0uriest: but the FFT interpolator already checks that NFP is the same. See my other comment, the rest of the code shouldn't care what interpolator is used.

desc/integrals/singularities.py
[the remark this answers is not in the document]
f0uriest: My idea for the use case would be having the eval grid have sufficient resolution to resolve stuff like R, Z, J etc (ie, eval_grid.M = eq.M_grid). But this still may not be enough to accurately resolve the singular behavior, so the source grid would need higher resolution. Even though the eval grid is lower resolution, it should still have enough to resolve the things that are interpolated to that grid, even with the truncation.

desc/integrals/singularities.py
```diff
@@ -327,104 +446,110 @@ def __call__(self, f, i):
...
-
-        # vmap for inner part found more efficient than fori_loop, especially on gpu,
-        # but for jacobian looped seems to be better and less memory
-        if loop:
-            fj = fori_loop(0, eval_grid.num_nodes, eval_pt_loop, jnp.zeros_like(f))
-        else:
-            fj = vmap(eval_pt_vmap)(jnp.arange(eval_grid.num_nodes))
-
-        f += fj.reshape((eval_grid.num_nodes, kernel.ndim))
+        f += batch_map(eval_pt, eval_data, 1 if loop else None).reshape(
```
f0uriest, line 537: I think we should allow arbitrary batch sizes while we're at it. Just have loop default to batch size=1

desc/integrals/singularities.py
```diff
@@ -433,91 +558,109 @@ def eval_pt_loop(i, fj):
...
-        keys += ["omega"]
+        keys.remove("phi")
+        keys.add("omega")
+    keys = list(keys)
     fsource = [source_data[key] for key in keys]
+    # Note that it is necessary to take the Fourier transforms of the
+    # vector components of the orthonormal polar basis vectors R̂, ϕ̂, Ẑ.
+    # Vector components of the Cartesian basis are not NFP periodic.
+    if isinstance(interpolator, DFTInterpolator):
+        fsource = [interpolator.fourier(val) for val in fsource]
```
f0uriest, line 599: did you test if having this outside the loop/vmap is actually necessary? f is fixed, so if you vmap over `i` I'm pretty sure jax shouldn't duplicate doing the fft.

desc/integrals/singularities.py
```diff
@@ -433,91 +558,109 @@ def eval_pt_loop(i, fj):
...
-        # in a single field period. loop this to get all pts
-        f_temp = polar_pt_vmap(i)
-        return f + f_temp.reshape((eval_grid.num_nodes, kernel.ndim))
+        return f + polar_pt(i)
 
-    f = jnp.zeros((eval_grid.num_nodes, kernel.ndim))
     # vmap found more efficient than fori_loop, esp on gpu, but uses more memory
     if loop:
-        f = fori_loop(0, v.size, polar_pt_loop, f)
+        f = fori_loop(
```
f0uriest, line 659: i think also worth using vmap_chunked or batch_map or whatever here

desc/objectives/_free_boundary.py
```diff
@@ -471,12 +484,12 @@ def build(self, use_jit=True, verbose=1):
...
 
         if self._eval_grid is None:
-            eval_grid = LinearGrid(
-                rho=np.array([1.0]), M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False
-            )
+            eval_grid = source_grid
         else:
             eval_grid = self._eval_grid
 
+        self._eval_grid_is_source_grid = eval_grid == source_grid
```
f0uriest, line 491: this is basically checking if they are the same object which might be too restrictive, ie if the user passed in `eval_grid=LinearGrid(4,5,6), source_grid=LinearGrid(4,5,6)` this would be False. I think you want `eval_grid.equiv(source_grid)` which will check if they have the same nodes etc.

desc/integrals/singularities.py
f0uriest: In general im not a fan of having this logic that depends on which method of interpolation is being used. If you want to keep it we should make a method of both that gets called in either case, and the fft method just returns f as is.

desc/integrals/singularities.py
unalmis: This is a case where chunking does not generalize the for loop because the loop reduces as the sum is computed (so max array size is 1), while chunking does not.

Should I heed your old code comment and interpret chunk size=1 as `fori_loop` and `chunk_size!=1` as a chunked vmap? Or just the latter?

desc/integrals/singularities.py
unalmis: No, I tried this way first. I don't like to leave potentially large optimizations up to a JIT compiler. We saved 16gb of memory by assigning a vandermonde matrix to a local variable in the bounce integrals.

f0uriest: ok, can we at least do my second comment, and always do this but for FFTInterpolator its a no-op so we can have a cleaner API

desc/integrals/singularities.py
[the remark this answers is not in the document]
unalmis: When NFP is different, the interpolators do different things. In that case the FFT interpolator is wrong unless the source grid has a single zeta node. I have added checks to ensure that is the only time NFP can be different.

I will move this comment to the FFTInterpolator

desc/integrals/singularities.py
unalmis: Yes of course

desc/integrals/singularities.py
f0uriest: ahh good point. Then yeah I think leave the loop for chunksize=1 but use vmap_chunked for the else

f0uriest: not for this PR, but at some point might be worth implementing some sort of batched reduction `desc.batching`

f0uriest (approved):

unalmis:
recommend squash merge after someone overrides codecov

rahulgaur104 (dismissed):

f0uriest (approved):

PR #1457, opened 2024-12-11
Fix `nan` in reverse mode gradient caused by `rotation_matrix`

### LINKED ISSUE #1456: nan reverse mode gradient when `FourierPlanarCurve` `normal` is parallel or anti-parallel to `Zaxis`
When I design TF and PF coils for the Solovev equilibrium everything works fine but when I add a helical coil to the coilset, I get a nan optimality.

Here's a minimal failing example

```python
from desc.backend import jnp
from desc.examples import get
from scipy.constants import mu_0
from desc.plotting import *
from desc.magnetic_fields import ToroidalMagneticField,SumMagneticField,  VerticalMagneticField
from desc.coils import (
    CoilSet,
    FourierPlanarCoil,
    FourierRZCoil,
    MixedCoilSet,
)
from desc.objectives import (
        BoundaryError,
        FixBoundaryR,
        FixBoundaryZ,
        FixPressure,
        FixSumCoilCurrent,
        ObjectiveFunction, 
        ForceBalance,
        QuadraticFlux, 
        FixParameters,
        FixIota, 
        CoilSetMinDistance,
        CoilLength,
)

from desc.examples import get

eq = get("SOLOVEV")
grid_at_surf = LinearGrid(rho=1.0, M=eq.M_grid, N=eq.N_grid)
G_tot = 2*jnp.pi*eq.compute("G", grid=grid_at_surf)["G"][0] / mu_0
field = ToroidalMagneticField(R0=1, B0=mu_0*G_tot/2/jnp.pi) + VerticalMagneticField(mu_0*G_tot/2/jnp.pi)  #+ PoloidalMagneticField(mu_0*G_tot/2/jnp.pi,R0=1,iota=eq.compute("iota")["iota"].mean()) 

coil1 = FourierPlanarCoil(current=1e5,center=[0,0, 2.4],normal=[0,0,1],r_n=4.5)
coil2 = FourierPlanarCoil(current=1e5,center=[0,0,-2.4],normal=[0,0,1],r_n=4.5)
n_coils = 12
TFcoil=FourierPlanarCoil(current=G_tot/n_coils,center=[4,0,0],normal=[0,1,0],r_n=2.2)
TFcoil.change_resolution(N=0)
coil3 = CoilSet(TFcoil,NFP=n_coils)

minor_radius = eq.compute("a")["a"]
offset = 1.5 * minor_radius
num_coils = 1  # coils per half field period per coilset
zeta = jnp.linspace(0, 2*jnp.pi, 41)
grid = LinearGrid(rho=[0.0], zeta=zeta, NFP=eq.NFP)
data = eq.axis.compute(["x", "x_s"], grid=grid, basis="rpz")
helical_coils = []
helical_offset = 0.0
R0 = 4.
R = R0 + offset* jnp.cos(zeta - helical_offset)
Z = offset * jnp.sin(zeta - helical_offset)
data = jnp.vstack([R, zeta, Z]).T
coil4 = FourierRZCoil.from_values(
    current=2.1e2,
    coords=data,
    #N = 4,
    basis="rpz",  # we are giving the center and normal in cylindrical
)

field = MixedCoilSet(coil1, coil2, coil3, coil4)
#field = MixedCoilSet(coil1, coil2, coil3)

# first coil opt to find correct toroidal field needed
R_modes = eq.surface.R_basis.modes[jnp.max(jnp.abs(eq.surface.R_basis.modes), 1) > 0, :]
Z_modes = eq.surface.Z_basis.modes[jnp.max(jnp.abs(eq.surface.Z_basis.modes), 1) > 0, :]
bdry_constraints = (
    FixBoundaryZ(eq=eq, modes=Z_modes),
)
coil_grid = LinearGrid(N=50)
constraints = (
    FixParameters(eq),
    FixParameters(field,[{"r_n":True,"center": [0, 1, 2],"normal":True}, {"r_n":True,"center": [0, 1, 2],"normal":True},{"normal":True}, {}]),
    #FixParameters(field,[{"r_n":True,"center": [0, 1, 2],"normal":True}, {"r_n":True,"center": [0, 1, 2],"normal":True},{"normal":True}]),
    FixSumCoilCurrent(field,indices=[False,False,True,False],target = G_tot/n_coils),
    #FixSumCoilCurrent(field,indices=[False,False,True],target = G_tot/n_coils),
)
objective = ObjectiveFunction((BoundaryError(eq,field),
                               CoilSetMinDistance(field,bounds=(0.2,100),grid=LinearGrid(N=100)),
                               #CoilLength(field, bounds=(0, 2 * np.pi * (0.5)), normalize_target=True, weight=100, grid=coil_grid,)
                              ))
(eq,field_opt), out = optimizer.optimize([eq,field],
    objective,
    constraints + bdry_constraints,
    verbose=3,
    options={},copy=True,
     ftol=1e-4, xtol=1e-6,maxiter=5
)


```
The nans go away if you remove coil4 from the coilset. 
Another way to fix this(as suggested by @dpanici ) while including the helical coil is to add deriv_mode="fwd" to all the objectives.  
What might be the reason for this?

dpanici:
Daniel's update:

Resolves #1456. There are really 2 different issues dealt with:

1. NaN with reverse-mode AD, which was caused by `jnp.arccos`. This is fixed with the new `safearccos` utility function. 
2. Incorrect behavior when the planar curve normal was anti-parallel to the +Z axis. This is fixed with special logic for this edge case. 

~~Resolves #1456 by adding different method of computing `rotation_matrix` (from [here](https://math.stackexchange.com/questions/180418/calculate-rotation-matrix-to-align-vector-a-to-vector-b-in-3d))~~

- [x] ~~Do we want to keep the old `rotation_matrix` still?~~
- [x] ~~I am a bit confused on why [this test](https://github.com/PlasmaControl/DESC/blob/61e5034321b89fdd9ce1d69aa381e7f24babdafc/tests/test_compute_utils.py#L12) passes on master even though this seems to be the cause of #1456 ?~~


desc/compute/geom_utils.py
```diff
@@ -28,6 +28,10 @@ def reflection_matrix(normal):
 def rotation_matrix(axis, angle=None):
     """Matrix to rotate points about axis by given angle.
 
+    NOTE: not correct if a and b are antiparallel, will
```
f0uriest, line 31: dont think this comment applies here

desc/compute/geom_utils.py
```diff
@@ -53,6 +57,45 @@ def rotation_matrix(axis, angle=None):
...
 
 
+def _skew_matrix(a):
+    return jnp.array([[0, -a[2], a[1]], [a[2], 0, -a[0]], [-a[1], a[0], 0]])
+
+
+def rotation_matrix_vector_vector(a, b):
+    """Matrix to rotate vector a onto b.
+
+    NOTE: not correct if a and b are antiparallel, will
```
f0uriest, line 67: could add something like `jnp.where(a+b==0, -R, R)`

tests/test_objective_funs.py
```diff
@@ -3076,7 +3076,7 @@ def test_objective_no_nangrad_quadratic_flux(self):
     @pytest.mark.unit
     def test_objective_no_nangrad_quadratic_flux_minimizing(self):
         """SurfaceQuadraticFlux."""
-        ext_field = ToroidalMagneticField(1.0, 1.0)
+        ext_field = FourierPlanarCoil(normal=[0, 0, 1])
```
f0uriest, line 3079: does this intersect the plasma?

f0uriest:
That existing test passes because the issue isnt with the rotation matrix itself, but in the `arccos` that is needed to compute the angle

desc/compute/geom_utils.py
```diff
@@ -53,6 +57,45 @@ def rotation_matrix(axis, angle=None):
...
+    b = safenormalize(b)
+    axis = cross(a, b)
+    norm = safenorm(axis)
+    axis = safenormalize(axis)
+    eps = 1e2 * jnp.finfo(axis.dtype).eps
+    skew = _skew_matrix(axis)
+    R1 = jnp.eye(3)
+    R2 = skew
+    R3 = (skew @ skew) * safediv(1, 1 + dot(a, b))
+    return jnp.where(norm < eps, jnp.eye(3), R1 + R2 + R3)  # if axis=0, no rotation
```
f0uriest, line 96: Not sure this is correct, I seem to get weird answers in a few cases I've tried. 
The determinant of a rotation matrix should always be 1, but for example

```python
np.linalg.det(rotation_matrix_vector_vector(np.array([1,0,0]), np.array([1,1,0])))
```
gives 1.17

and if they are almost antiparallel:
```python
np.linalg.det(rotation_matrix_vector_vector(np.array([1,0,0]), np.array([-1,1e-4,0])))
```
gives like 4e16

desc/compute/_curve.py
```diff
@@ -345,9 +345,7 @@ def _x_sss_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
     coords = jnp.array([d3X, d3Y, d3Z]).T
     # rotate into place
     Zaxis = jnp.array([0.0, 0.0, 1.0])  # 2D curve in X-Y plane has normal = +Z axis
-    axis = cross(Zaxis, normal)
```
f0uriest, line 348: another way to rotate `a` into `b` would be to rotate by `pi` around the axis `(a+b)/2` which might be easier. 
```python
a = Zaxis
b = safenormalize(normal)
axis = safenormalize((a+b)/2)
A = rotation_matrix(axis, angle=np.pi)
```

That avoids the arccos and I think should be fine since its all safenormalized?

Would still need to special case if they are exactly antiparallel but seems to work ok if they are ~nearly antiparallel

ddudt (commented):
It seems like any method for computing the rotation matrix will have edge cases that require special logic to handle. Therefore I vote to keep the existing method, and just add the logic to make the reverse mode work when `angle = arccos(1) = 0`

desc/compute/geom_utils.py
dpanici: I am certain it does, I checked it and the rotation matrix is wrong for this case with this function too

dpanici:
safe arccos
check for sign of dot product of Zaxis and normal to assign correct sign

desc/compute/geom_utils.py
dpanici: thanks, I ended up using this and I think it basically fixed it, at least my tests pass so I think this is correct, and we get no nans in the tests either for parallel/antiparalel or nearly parallel/antiparallel

desc/compute/geom_utils.py
dpanici: yea I think I was wrongly normalizing the axis before, removing that fixes the issue.

desc/compute/geom_utils.py
f0uriest: I just meant that there's no a or b in this function.

desc/compute/geom_utils.py
f0uriest: Ok so this is then correct in all cases? If so we can remove the comment

f0uriest (commented):
If the original rotation matrix function is now unused I'm fine removing it

dpanici:
> If the original rotation matrix function is now unused I'm fine removing it

it is used in some of our coils for the `rotate` method, right now this PR does not change any of the public-facing ways for doing this. I guess I can change just internally how the rotmat is computed

dpanici:
> > If the original rotation matrix function is now unused I'm fine removing it
> 
> it is used in some of our coils for the `rotate` method, right now this PR does not change any of the public-facing ways for doing this. I guess I can change just internally how the rotmat is computed, but I am pretty sure I would have to use the same formula as the one implemented now anyways to compute the correct a/b matrix to pass to the new function. 

I can just make sure the old function correctly handles reflections and keep it for now.

dpanici:
This PR now works fine, but seems the new way I implemented is noisier for nearly-antiparallel, so if we can make safe arccos work or I will try the other way Rory mentioned, that would be best.

```python
from desc.compute.geom_utils import rotation_matrix_vector_vector, rotation_matrix
from desc.utils import safenormalize
import numpy as np
print("\nnew way (vector->vector rotation)")
a = [0,1,0]
b = np.array([0,-1,1e-6])
A=rotation_matrix_vector_vector(a, b)
print(A)
print(np.linalg.det(A))
aa = np.array([1,1,1])
print(A@aa)
print("\nold way (axis-angle rotation)")
axis = np.cross(a, b)
angle = np.arccos(np.dot(a, safenormalize(b)))
A=rotation_matrix(axis,angle)
print(A)
print(np.linalg.det(A))
aa = np.array([1,1,1])
print(A@aa)

```
yields

```
new way (vector->vector rotation)
[[ 1.00000000e+00  0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00 -9.99822215e-01 -1.00000000e-06]
 [ 0.00000000e+00  1.00000000e-06 -9.99822215e-01]]
0.9996444608857149
[ 1.         -0.99982321 -0.99982121]

old way (axis-angle rotation)
[[ 1.00000000e+00  0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00 -1.00000000e+00 -1.00004445e-06]
 [-0.00000000e+00  1.00004445e-06 -1.00000000e+00]]
1.0
[ 1.       -1.000001 -0.999999]
```

YigitElma:
Don't know if related but nice trick https://docs.kidger.site/equinox/api/debug/#common-sources-of-nans

desc/compute/_curve.py
```diff
@@ -212,8 +212,11 @@ def _x_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
     # rotate into place
     Zaxis = jnp.array([0.0, 0.0, 1.0])  # 2D curve in X-Y plane has normal = +Z axis
     axis = cross(Zaxis, normal)
-    angle = jnp.arccos(dot(Zaxis, safenormalize(normal)))
-    A = rotation_matrix(axis=axis, angle=angle)
+    angle = safearccos(dot(Zaxis, safenormalize(normal)))
+    if jnp.allclose(Zaxis + normal, 0.0):  # normal is anti-parallel to +Z axis
+        A = jnp.diag(jnp.array([1.0, -1.0, -1.0]))
+    else:
+        A = rotation_matrix(axis=axis, angle=angle)
```
ddudt, line 219: Dario's old solution had this logic in the `rotation_matrix` function. I want to keep that function general for arbitrary rotations, without assuming the edge case of being anti-parallel is relative to the Z-axis. The downside is that now we have to handle this logic in each of these compute functions.

desc/geometry/curve.py
```diff
@@ -1217,8 +1222,13 @@ def from_values(cls, coords, N=10, s=None, basis="xyz", name=""):
...
         Z_axis = np.array([0, 0, 1])
         axis = np.cross(Z_axis, normal)
-        angle = np.arccos(np.dot(Z_axis, normal))
-        rotmat = rotation_matrix(axis, angle)
+        dotprod = np.dot(Z_axis, normal)
+        angle = np.arccos(dotprod)
+        rotmat = np.where(
+            np.allclose(dotprod, -1.0),
+            np.diag(np.array([1.0, -1.0, -1.0])),
+            rotation_matrix(axis, angle),
```
YigitElma, line 1230: If `axis` is [0,0,0], `rotation_matrix` function just returns identity which is not enough even if we don't differentiate over it.

ddudt (approved):

YigitElma (dismissed):

desc/geometry/curve.py
f0uriest: what do you mean not enough?

desc/compute/_curve.py
```diff
@@ -212,8 +212,13 @@ def _x_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
     # rotate into place
     Zaxis = jnp.array([0.0, 0.0, 1.0])  # 2D curve in X-Y plane has normal = +Z axis
     axis = cross(Zaxis, normal)
-    angle = jnp.arccos(dot(Zaxis, safenormalize(normal)))
-    A = rotation_matrix(axis=axis, angle=angle)
+    dotprod = dot(Zaxis, safenormalize(normal))
+    angle = safearccos(dotprod)
+    A = jnp.where(  # handle the case where normal is aligned with the -Z axis
+        jnp.allclose(dotprod, -1.0),
```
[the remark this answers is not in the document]
YigitElma: We can make it tighter. However, the current case will fail if the angle between 2 unit vectors is less than

$$
arccos(1-1e-5) = 0.25 ^\degree = 0.0045 \text{ radian}
$$

which is, I think, good enough.

desc/geometry/curve.py
YigitElma: Yes yes, I was just explaining why I had this change.

YigitElma:
@dpanici

PR #1462, opened 2024-12-12
Fix unit tests

YigitElma:
The new releases for Python 3.12 and 3.11 introduced some changes to the activation of virtual environments that made the `venv`s incompatible between Python patched 3.12.8 and 3.12.7 which Github runners use. This caused the unit tests to fail (since they are the only ones that use 3.12. and 3.11).

This PR changes the caching procedure to include the full Python version and use it if it is the same as the runners use. Otherwise, create a new environment for that job, so it will run either way.

Note: I believe this problem will be resolved automatically once all Github runners use the same Python version (they released new patch last week so it might take another week). But these changes will make it ready for future releases.


YigitElma:
Probably related to https://github.com/python/cpython/issues/124651
They changed `venv` activation for python `3.11.11` and `3.12.8` which the job runner recently started to use

f0uriest:
If we can't get this working can we just drop the caching and go back to installing each time? It only takes like ~1 minute

.github/workflows/regression_tests.yml
```diff
@@ -45,25 +44,25 @@ jobs:
...
-          python-version: ${{ matrix.python-version }}
+          python-version: '3.10'
 
       - name: Restore Python environment cache
         if: env.has_changes == 'true'
         id: restore-env
         uses: actions/cache/restore@v4
         with:
-          path: .venv-${{ matrix.python-version }}
-          key: ${{ runner.os }}-venv-${{ matrix.python-version }}-${{ hashFiles('devtools/dev-requirements.txt', 'requirements.txt') }}
```
f0uriest, line 59: instead of using `matrix.python-version` as the key why can't we use the actual version from `python --version`

YigitElma: Btw this problem will probably be solved when every Github runner starts using 3.12.8 and 3.11.11. For now, some of them use 3.12.7 and this causes the issue. They made some security changes regarding the `venv` activation with the latest patch

f0uriest (approved):

dpanici (approved):

PR #1463, opened 2024-12-12
Set preferred coordinate direction for conversion from points to FourierPlanarCurve

ryanwu4:
Previously, the `from_values()` method in `FourierPlanarCurve` would convert the coords provided to an r(s) representation with s sorted to be monotonically increasing (counterclockwise). However, this means that if the original coordinates were parameterized "backwards" (clockwise), the new `FourierPlanarCurve` would return coordinates in a reversed order, which is problematic if you are trying to run something like a Biot Savart integral where the coordinate order is the implicit current direction in the loop. This PR fixes the issue by flipping the normal vector of the `FourierPlanarCurve` when the provided coordinates are in the clockwise direction.

### DIFF
--- a/desc/geometry/curve.py
+++ b/desc/geometry/curve.py
@@ -806,10 +806,10 @@ class FourierPlanarCurve(Curve):
 
         # center
         center = np.mean(coords, axis=0)
-        coords = coords - center  # shift to origin
+        coords_centered = coords - center  # shift to origin
 
         # normal
-        U, _, _ = np.linalg.svd(coords.T)
+        U, _, _ = np.linalg.svd(coords_centered.T)
         normal = U[:, -1].T  # left singular vector of the least singular value
 
         # axis and angle of rotation
@@ -817,17 +817,30 @@ class FourierPlanarCurve(Curve):
         axis = np.cross(Z_axis, normal)
         angle = np.arccos(np.dot(Z_axis, normal))
         rotmat = rotation_matrix(axis, angle)
-        coords = coords @ rotmat  # rotate to X-Y plane
+        coords_rotated = coords_centered @ rotmat  # rotate to X-Y plane
 
         warnif(
-            np.max(np.abs(coords[:, 2])) > 1e-14,  # check that Z=0 for all points
+            np.max(np.abs(coords_rotated[:, 2])) > 1e-14,  # check that Z=0 for all points
             UserWarning,
             "Curve values are not planar! Using the projection onto a plane.",
         )
 
+        # polar angle
+        s = np.arctan2(coords_rotated[:, 1], coords_rotated[:, 0])
+        unwrapped_s = np.unwrap(s)
+        curve_sign = np.sign(unwrapped_s[-1] - unwrapped_s[0])
+
+        if curve_sign == -1: #original curve was parameterized "backwards" (clockwise) compared to FourierPlanarCurve assumption
+            normal = -normal # flip normal vector direction
+            axis = np.cross(Z_axis, normal)
+            angle = np.arccos(np.dot(Z_axis, normal))
+
+            rotmat = rotation_matrix(axis, angle)
+            coords_rotated = coords_centered @ rotmat  # rotate to X-Y plane
+
         # polar radius and angle
-        r = np.sqrt(coords[:, 0] ** 2 + coords[:, 1] ** 2)
-        s = np.arctan2(coords[:, 1], coords[:, 0])
+        r = np.sqrt(coords_rotated[:, 0] ** 2 + coords_rotated[:, 1] ** 2)
+        s = np.arctan2(coords_rotated[:, 1], coords_rotated[:, 0])
         s = np.mod(s + 2 * np.pi, 2 * np.pi)  # mod angle to range [0, 2*pi)
         idx = np.argsort(s)  # sort angle to be monotonically increasing
         r = r[idx]
--- a/tests/test_curves.py
+++ b/tests/test_curves.py
@@ -657,6 +657,56 @@ class TestFourierPlanarCurve:
         with pytest.raises(AssertionError):
             _ = FourierPlanarCurve(r_n=[1], modes=[1, 2])
 
+    @pytest.mark.unit
+    def test_to_FourierPlanarCurve(self):
+        """Test converting SplineXYZCurve to FourierPlanarCurve object."""
+        npts = 1000
+        N = 5
+
+        # Create a SplineXYZCurve of a planar circle
+        s = np.linspace(0, 2 * np.pi, npts)
+        X = 2 * np.cos(s)
+        Y = np.ones(npts)
+        Z = 2 * np.sin(s)
+        c = SplineXYZCurve(X=X, Y=Y, Z=Z)
+
+        # Create a backwards SplineXYZCurve by flipping the coordinates
+        c_backwards = SplineXYZCurve(
+            X=np.flip(X),
+            Y=np.flip(Y),
+            Z=np.flip(Z),
+        )
+
+        # Convert to FourierPlanarCurve
+        c_planar = c.to_FourierPlanar(N=N, grid=npts, basis="xyz")
+        c_backwards_planar = c_backwards.to_FourierPlanar(N=N, grid=npts, basis="xyz")
+
+        grid = LinearGrid(N=20, endpoint=True)
+
+        coords_spline = c.compute("x", grid=grid)["x"]
+        coords_planar = c_planar.compute("x", grid=grid)["x"]
+        coords_backwards_spline = c_backwards.compute("x", grid=grid)["x"]
+        coords_backwards_planar = c_backwards_planar.compute("x", grid=grid)["x"]
+
+        # Assertions for point positions
+        np.testing.assert_allclose(coords_spline, coords_planar, atol=1e-10)
+        np.testing.assert_allclose(
+            coords_backwards_spline,
+            coords_backwards_planar,
+            atol=1e-10,
+        )
+        # backwards coordinate order is import for Biot-Savart implicit current direction
+        np.testing.assert_allclose(
+            coords_planar, np.flip(coords_backwards_planar, axis=0), atol=1e-10
+        )
+
+        # Verify that the normal vectors are in opposite directions
+        np.testing.assert_allclose(
+            (c_planar.normal / c_backwards_planar.normal)[1],
+            np.array([-1]),
+            atol=1e-10,
+        )
+
 
 class TestSplineXYZCurve:
     """Tests for SplineXYZCurve class."""


desc/geometry/curve.py
```diff
@@ -806,28 +806,41 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
+        unwrapped_s = np.unwrap(s)
+        curve_sign = np.sign(unwrapped_s[-1] - unwrapped_s[0])
+
+        if curve_sign == -1: #original curve was parameterized "backwards" (clockwise) compared to FourierPlanarCurve assumption
+            normal = -normal # flip normal vector direction
+            axis = np.cross(Z_axis, normal)
+            angle = np.arccos(np.dot(Z_axis, normal))
+
+            rotmat = rotation_matrix(axis, angle)
+            coords_rotated = coords_centered @ rotmat  # rotate to X-Y plane
```
ddudt, line 839: Why do we need the additional rotation and not just flipping the normal vector direction? I think we only need to flip the normal vector, since that would effectively reverse the parameterization in real space

desc/geometry/curve.py
```diff
@@ -806,28 +806,41 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
 
         warnif(
-            np.max(np.abs(coords[:, 2])) > 1e-14,  # check that Z=0 for all points
+            np.max(np.abs(coords_rotated[:, 2])) > 1e-14,  # check that Z=0 for all points
             UserWarning,
             "Curve values are not planar! Using the projection onto a plane.",
         )
 
+        # polar angle
+        s = np.arctan2(coords_rotated[:, 1], coords_rotated[:, 0])
```
ddudt, line 829: If you don't need to rotate the coordinates, then this calculation of the polar angle is redundant and you can use the existing angles `s` below.

desc/geometry/curve.py
```diff
@@ -806,28 +806,41 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
+        unwrapped_s = np.unwrap(s)
+        curve_sign = np.sign(unwrapped_s[-1] - unwrapped_s[0])
```
ddudt, line 831: What if the endpoint of the curve is repeated? It might be better to use some sort of average over all the angles, instead of comparing only the first and last points (they should either be monotonically increasing or decreasing)

desc/geometry/curve.py
ddudt: Alternatively, I think we could flip the sign of the Fourier coefficients `r_n` where $m < 0$, since $\cos(-\theta)=\cos(\theta)$ and $\sin(-\theta)=-\sin(\theta)$

tests/test_curves.py
```diff
@@ -657,6 +657,56 @@ def test_asserts(self):
...
+        np.testing.assert_allclose(coords_spline, coords_planar, atol=1e-10)
+        np.testing.assert_allclose(
+            coords_backwards_spline,
+            coords_backwards_planar,
+            atol=1e-10,
+        )
+        # backwards coordinate order is import for Biot-Savart implicit current direction
+        np.testing.assert_allclose(
+            coords_planar, np.flip(coords_backwards_planar, axis=0), atol=1e-10
+        )
```
ddudt, line 701: A better test here would be to compute $\mathbf{B}$ at several points in real space and check that both coils give the same result.

desc/geometry/curve.py
ryanwu4: The weird thing is that the curve points test passes with just flipping the sign of the normal, but the fields are different between spline and normal-flipped-FourierPlanar _unless_ I redo the rotation matrix with the flipped normal. I'll look into this a bit more.

ryanwu4: So I looked at the coil shapes with and without re-rotating, and it appears that if you only flip the sign of the normal without rotating as well, you mirror the curve as well -- this is because you swap the chirality of the curve. If you drew the curve on a piece of paper, flipping the normal would be looking at the sheet of paper from the back, rather than flipping the piece of paper itself.

ryanwu4: For some reason this doesn't work, no matter what it seems to need both an axis reversal and a rotation (pancake flip) to get the curves to line up again -- I clarified in the code what the rotation is now doing, but I couldn't find the best way to do it without another rotation.

ryanwu4:
In the process of working on this , I noticed the following behavior: 
If the normal vector of a `FourierPlanarCurve` is  defined as [0,0,-1], the curve direction will be incorrect (clockwise about the normal instead of counterclockwise). This is due to the `axis` term in the corresponding `x` compute function being undefined for axes parallel to the Z axis, so the rotation matrix defaults to the unit diagonal matrix. I added this as an assertion check, though conceivably this can probably be fixed.

The problematic code in the compute function:
```
    Zaxis = jnp.array([0.0, 0.0, 1.0])  # 2D curve in X-Y plane has normal = +Z axis
    axis = cross(Zaxis, normal)
    angle = jnp.arccos(dot(Zaxis, safenormalize(normal)))
    A = rotation_matrix(axis=axis, angle=angle)
```

Minimal Example:
```
coil = FourierPlanarCoil(
            current=1e6,
            center=[0, 0, 0],
            normal=[0, 0, 1],
            r_n=2,
            modes=[0],
        )

coil2 = FourierPlanarCoil(
            current=1e6,
            center=[0, 0, 0],
            normal=[0, 0, -1],
            r_n=2,
            modes=[0],
        )

field1 = coil.compute_magnetic_field([0, 0, 0], source_grid=LinearGrid(N=100), basis = "xyz")
field2 = coil2.compute_magnetic_field([0, 0, 0], source_grid=LinearGrid(N=100), basis = "xyz")

np.testing.assert_allclose(field1, field2, rtol=1e-3, atol=1e-3) # these fields shouldn't be equal but they are
```

The expected behavior of the assertion should be to fail, as `field1` and `field2` should have equal and opposite magnitudes. Instead, they are actually the same number, and the assertion succeeds. Compare this to if you perturb the normal vectors just a tiny bit (eg. [0, 0.1, +-1]), and the field correctly has opposite signs.

ryanwu4:
This branch should be ready for PR pending resolution of #1456

ryanwu4:
This actually has been fixed in a more elegant way in https://github.com/PlasmaControl/DESC/pull/1504/commits/c101095a51a45f1733715ec0d933c1571ce7a081. If that PR is merged, I can just port the tests over from this and close this PR.

dpanici:
> This actually has been fixed in a more elegant way in [c101095](https://github.com/PlasmaControl/DESC/commit/c101095a51a45f1733715ec0d933c1571ce7a081). If that PR is merged, I can just port the tests over from this and close this PR.

That PR actually breaks the `from_values` of `FourierPlanarCurve` i think

My preference is to merge this one in without waiting for #1456 to be resolved as this fixes different issues than #1456

dpanici (changes requested):
If you can update the changelog, I would approve this, something like


- Fixes issue in ``desc.geometry.curve.FourierPlanarCurve.from_values`` where the orientation of the fitted curve can be the reverse of the original curve, which can be problematic for coils (the current is not negated, so the resulting fitted coil would have field opposite of the initial)

desc/geometry/curve.py
```diff
@@ -826,28 +826,54 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
         )
 
+        # polar angle
+        s = np.arctan2(coords_rotated[:, 1], coords_rotated[:, 0])
+        unwrapped_s = np.unwrap(s)
+        # Determine if the sequence is monotonically increasing or decreasing
+        if np.all(np.diff(unwrapped_s) > 0):
+            curve_sign = 1
+        elif np.all(np.diff(unwrapped_s) < 0):
+            curve_sign = -1
```
dpanici, line 856: unclear why codecov is complaining as these clearly are checked by the added tests

dpanici (dismissed):

ddudt (approved):

desc/geometry/curve.py
```diff
@@ -826,28 +826,54 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
 
         # center
         center = np.mean(coords, axis=0)
-        coords = coords - center  # shift to origin
+        coords_centered = coords - center  # shift to origin
 
         # normal
-        U, _, _ = np.linalg.svd(coords.T)
+        U, _, _ = np.linalg.svd(coords_centered.T)
```
YigitElma, line 832: General question:
Is there a specific reason why we don't use just 3 points to get the normal? but instead use a full SVD?
I am pretty sure just this would work
```python
N = int(coords.shape[0]/3)
x1 = coords[0]
x2 = coords[N]
x3 = coords[2*N]
AB = x2-x1
BC = x3-x2
normal = np.cross(AB, BC)
normal /= np.linalg.norm(normal)
```

YigitElma: Maybe as a safety check, look if their parallel or not but it is really unlikely imo

YigitElma: Ok, I see the older version was using something similar, but they have changed it to account for non-planar points.

YigitElma (dismissed):

desc/geometry/curve.py
ryanwu4: Yea I think it functions as like a best-fit plane for a nonplanar starting shape.

YigitElma:
@dpanici @ddudt feel free to merge. Memory benchmark fails because this is a fork (it will be fixed by #1688)

tests/test_coils.py
```diff
@@ -454,6 +454,63 @@ def test_convert_type(self):
...
+
+        # Create a SplineXYZCoil of a planar ellipse
+        s = np.linspace(0, 2 * np.pi, npts)
+        X = 2 * np.cos(s)
+        Y = np.ones(npts)
+        Z = 1 * np.sin(s)
+        c = SplineXYZCoil(X=X, Y=Y, Z=Z, current=1e6)
+
+        # Create a backwards SplineXYZCoil by flipping the coordinates
+        c_backwards = SplineXYZCoil(
```
YigitElma, line 472: Do we expect these tests to fail on master? I don't think they do

ryanwu4: I thought that they do, at least the magnetic field computation is backwards I think

YigitElma: Can you check it again, please? I might have made a mistake while copying the code, but it was passing on my laptop

ryanwu4: Strange, these do pass on master now...

ryanwu4: Dario's MWE on https://github.com/PlasmaControl/DESC/issues/1715 is still incorrect though, maybe I'll stick with that one and edit the test slightly

ryanwu4:
Looks like something in how `SplineXYZ` computes points is slightly different from when I wrote this test a few months ago, and the tests do pass on master. As such, I just subbed in the test Dario wrote on #1716 , which correctly fails on master and is fixed on this branch.

desc/geometry/curve.py
YigitElma: I think for the previous test, it was not covering these lines

tests/test_coils.py
dpanici: These tests do fail on master... I am confused now. At least the following code on master fails. It passes on my other fix branch #1715, and passes on this PR as well


```python
import numpy as np
from desc.coils import *
from desc.grid import *
"""Test converting SplineXYZCoil to FourierPlanarCoil object."""
# Necessary as SplineXYZ can be either clockwise or counterclockwise
npts = 1000
N = 50

# Create a SplineXYZCoil of a planar ellipse
s = np.linspace(0, 2 * np.pi, npts)
X = 2 * np.cos(s)
Y = np.ones(npts)
Z = 1 * np.sin(s)
c = SplineXYZCoil(X=X, Y=Y, Z=Z, current=1e6)

# Create a backwards SplineXYZCoil by flipping the coordinates
c_backwards = SplineXYZCoil(
    X=np.flip(X),
    Y=np.flip(Y),
    Z=np.flip(Z),
    current=1e6,
)

# Convert to FourierPlanarCoil
c_planar = c.to_FourierPlanar(N=N, grid=npts, basis="xyz")
c_backwards_planar = c_backwards.to_FourierPlanar(N=N, grid=npts, basis="xyz")

grid = LinearGrid(zeta=100)

field_spline = c.compute_magnetic_field(
    np.zeros((1, 3)), source_grid=grid, basis="xyz"
)
field_planar = c_planar.compute_magnetic_field(
    np.zeros((1, 3)), source_grid=grid, basis="xyz"
)
field_backwards_spline = c_backwards.compute_magnetic_field(
    np.zeros((1, 3)), source_grid=grid, basis="xyz"
)
field_backwards_planar = c_backwards_planar.compute_magnetic_field(
    np.zeros((1, 3)), source_grid=grid, basis="xyz"
)

np.testing.assert_allclose(
    field_spline,
    field_planar,
    atol=1e-5,
)
np.testing.assert_allclose(
    field_backwards_spline,
    field_backwards_planar,
    atol=1e-5,
)
np.testing.assert_allclose(
    field_planar,
    -field_backwards_planar,
    atol=1e-5,
)
```

dpanici (approved):
Not sure what that weirdness with the other test was but this test is fine.

tests/test_curves.py
```diff
@@ -684,6 +684,49 @@ def test_asserts(self):
         with pytest.raises(AssertionError):
             _ = FourierPlanarCurve(r_n=[1], modes=[1, 2])
 
+    @pytest.mark.unit
+    def test_to_FourierPlanarCurve(self):
```
YigitElma, line 688: This one also passes on master branch

YigitElma (approved):
Second test seems to work on master, but if you want to keep it I am fine

tests/test_coils.py
dpanici: I never figured this out, only for myself locally did this fail on master so maybe it is a mac thing (I updated my environment). regardless on this PR the test passes everywhere so it is fine

PR #1465, opened 2024-12-12
bug fix for Profile I/O attributes

### LINKED ISSUE #1448: I/O warning for profiles without `params` attribute
The `_Profile` class defines the I/O attribute `"_params"` that all other profiles inherit. But `PowerProfile`, `ProductProfile`, `ScaledProfile`, and `SumProfile` do not have a `self._params` attribute, they only define a `self.params` "getter" method. This leads to a `RuntimeWarning` when saving and/or loading these profiles.

ddudt:
Resolves #1448


dpanici (approved):

YigitElma (approved):
I know this is not related to this PR, but can we add a ``__all__`` attribute to `profiles.py` file? Or is there a reason we have for most other files and not for this one?

f0uriest (approved):

ddudt:
> I know this is not related to this PR, but can we add a `__all__` attribute to `profiles.py` file? Or is there a reason we have for most other files and not for this one?

I think defining `__all__` only matters for the behavior of `from desc.profiles import *`. By default (if we don't specify `__all__`) it will import all of the classes without a leading underscore. Since `profiles.py` doesn't contain any local utility functions like most other files, this default behavior is fine and will make all the profile classes public. [SE source](https://stackoverflow.com/questions/44834/what-does-all-mean-in-python)

YigitElma:
> > I know this is not related to this PR, but can we add a `__all__` attribute to `profiles.py` file? Or is there a reason we have for most other files and not for this one?
> 
> I think defining `__all__` only matters for the behavior of `from desc.profiles import *`. By default (if we don't specify `__all__`) it will import all of the classes without a leading underscore. Since `profiles.py` doesn't contain any local utility functions like most other files, this default behavior is fine and will make all the profile classes public. [SE source](https://stackoverflow.com/questions/44834/what-does-all-mean-in-python)

Hmm, yeah. I sometimes use it to see which options we have in file as a quick scan. What profiles/grid/basis etc. types we have implemented.

PR #1472, opened 2024-12-16
Neo wrapper

dpanici:
eventually add to #1028


dpanici:
use existing NEOIO in test_neoclassica.py

unalmis:
closing since #2170 exists and neo_jax exists.

PR #1483, opened 2024-12-20
Fix `objective.compile` in Benchmarks and Miscellaneous stuff

### LINKED ISSUE #1402: Export compiled objectives for common equilibrium resolutions?
https://jax.readthedocs.io/en/latest/export/export.html

YigitElma:
Some benchmarks were using `obj.compile` and `jac_scaled_error` together but `obj.compile` only compiles `jac_scaled` and `compute_scaled`. This cause some benchmarks to have very different min/max values, like,
image
With the change it is more consistent,
image

## Misc

- Adjusts number of rounds for tests
- Adds `Action Details` step to tests
- Resolves #1402 
- Switches from `scipy.linalg.pinv` to `jnp.linalg.pinv` for building pseudoinverse of the transform matrix.


sinaatalay (approved):
This PR looks good to me, although others should review it as I am looking at it with only my Python and GitHub knowledge, without DESC knowledge.

- Workflows (`benchmark.yaml`, `notebook_tests.yaml`, and `regression_tests.yaml` works the same way as its previous version, except a new step is added, `Action Details`, which moves all the debugging-related commands to a separate step. It makes sense.
- There is a change in `desc.objectives.objective_funcs.ObjectiveFunction.compile` method, which uses `compute_scaled_error` method instead of `compute_scaled` in `lsq` and `all` modes. This hasn't been explained in the PR or commit messages. Maybe @YigitElma should explain it, but I am sure it's okay.
- The documentation is updated and seems okay.
- Tests haven't been changed algorithmically (except changing `rounds` in benchmarks) but have been cleaned. It looks better, I don't see any errors/

docs/performance_tips.rst
```diff
@@ -1,3 +1,23 @@
...
+    import jax
+    import jax.numpy as jnp
+
+    jax.config.update("jax_compilation_cache_dir", "../jax-caches")
+    jax.config.update("jax_persistent_cache_min_entry_size_bytes", -1)
+    jax.config.update("jax_persistent_cache_min_compile_time_secs", 0)
+
+This will use a directory called ``jax-caches`` in the parent directory of the script to store the compiled code. The ``jax_persistent_cache_min_entry_size_bytes`` and ``jax_persistent_cache_min_compile_time_secs`` parameters are set to -1 and 0, respectively, to ensure that all compiled code is cached. For more details on caching, refer to official JAX documentation `here <https://jax.readthedocs.io/en/latest/persistent_compilation_cache.html#persistent-compilation-cache>`__.
+
+Note: Updating JAX version might re-compile some previously cached code, and thi might increase the cache size. Every once in a while, you might need to clear your cache directory.
```
f0uriest, line 20: `thi` -> `this`

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -131,67 +131,53 @@ def build():
         N = 25
         _ = Equilibrium(L=L, M=M, N=N)
 
-    benchmark.pedantic(build, setup=setup, iterations=1, rounds=50)
+    benchmark.pedantic(build, setup=setup, iterations=1, rounds=10)
```
f0uriest, line 134: reducing the number of rounds will make the statistics more noisy, and may lead to more false positives

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -355,17 +338,16 @@ def test_proximal_jac_atf(benchmark):
...
 
-    benchmark.pedantic(run, args=(x,), rounds=15, iterations=1)
+    benchmark.pedantic(run, args=(x, prox), rounds=10, iterations=1)
 
 
 @pytest.mark.slow
 @pytest.mark.benchmark
 def test_proximal_freeb_compute(benchmark):
     """Benchmark computing free boundary objective with proximal constraint."""
-    jax.clear_caches()
```
f0uriest, line 368: why remove this?

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -455,26 +436,19 @@ def run(eq):
 @pytest.mark.benchmark
 def test_LinearConstraintProjection_build(benchmark):
     """Benchmark LinearConstraintProjection build."""
+    eq = desc.examples.get("W7-X")
 
-    def setup():
+    def run():
```
f0uriest, line 441: moving everything to the run function is changing what its actually profiling, this now includes a bunch of other stuff besides building the linear constraints, is that what we want?

tests/benchmarks/benchmark_cpu_small.py
f0uriest, line 1: any changes made to this file should also be made to `benchmark_gpu_small.py`

tests/benchmarks/benchmark_cpu_small.py
YigitElma: I agree. For example, in the last benchmark, this PR shows speed improvement for `perturb` but it shouldn't. We can decide on the exact number of rounds, but my intention is to balance the time spent on tests. Previously, these equilibrium initialization tests took more time than `fixed_iter_solve` and `perturb` tests. Given that benchmark workflow started to take around 50mins, I wanted to reduce them from 50, which is a bit overkill.

tests/benchmarks/benchmark_cpu_small.py
YigitElma: This doesn't cause too much difference,(I can re add it) but technically only `run` is benchmarked, so clearing the cache here doesn't have much purpose.

tests/benchmarks/benchmark_cpu_small.py
YigitElma: The extra stuff is just building individual constraints and objectives, right? Previously, this was effectively just benchmarking `factorize_linear_constraints`. Do we usually pass built constraints to the `LinearConstraintProjection`? Then I can revert it.

docs/performance_tips.rst
```diff
@@ -1,3 +1,24 @@
+================
+Performance Tips
+================
+
+Caching the Compiled (Jitted) Code
+----------------------------------
+Although the compiled code is fast, it still takes time to compile. If you are running the same optimization, or some similar optimization, multiple times, you can save time by caching the compiled code. This automatically happens for a single session (for example, until you restart your kernel in Jupyter Notebook) but once you start using another session, the code will need to be recompiled. Fortunately, there is a way to bypass this. First create a cache directory (i.e. ``jax-caches``), and put the following code at the beginning of your script:
```
dpanici, line 7: you have this locally in your backend.py right? so you would not usually need this in every script you run if it is there?

YigitElma: Yeah, if you put it to `backend.py`, you shouldn't need it every time. But it can be messy for users, so I didn't mention that.

tests/benchmarks/benchmark_cpu_small.py
dpanici: I think we usually pass built constraints to it, yes

dpanici: https://github.com/PlasmaControl/DESC/blob/f8c66c724544380ec31e57db047ee08c54fc142c/desc/optimize/optimizer.py#L218

docs/index.rst
```diff
@@ -21,6 +21,7 @@
    :caption: Tutorials
 
    installation
+   performance_tips
```
f0uriest, line 24: for ordering id prefer to keep this at the end, so that going through them in order makes more sense

YigitElma: This will be fixed by #1498

sinaatalay (approved):

f0uriest (approved):

PR #1485, opened 2024-12-20
`CoilIntegratedCurvature` objective

ddudt:
A scalar version of convexity for a curve: $\int \kappa |\partial_s \mathbf{x}| ds = 2 \pi$

Similar to our existing `CoilCurvature` objective, which is signed to indicate convexity.


desc/objectives/_coils.py
```diff
@@ -134,6 +139,11 @@ def _prune_coilset_tree(coilset):
             ValueError,
             "Only use toroidal resolution for coil grids.",
         )
+        errorif(
```
dpanici, line 142: I don't love this, can you instead do something like giving the CoilConvexity objective a `self._endpoint` attribute, and in this build check if hasattr(self,_endpoint) and usethat, else set it to False.

Then have this error check statement be after the `super.build()` statement inside of `CoilConvexity`'s build method

desc/objectives/_coils.py
```diff
@@ -602,6 +612,107 @@ def compute(self, params, constants=None):
         return out
 
 
+class CoilConvexity(_CoilObjective):
+    """Coil convexity.
```
dpanici, line 616: make it more clear that it is targeting that this objective is targeting to make the coil convex by making the that integral quantity minus 2pi = 0, that way the target of this objective being zero makes sense.

As is, it is ambiguous as to whether one should make the target 2pi or 0.

desc/objectives/_coils.py
```diff
@@ -602,6 +612,107 @@ def compute(self, params, constants=None):
...
+        coil,
+        target=None,
+        bounds=None,
+        weight=1,
+        normalize=True,
+        normalize_target=True,
+        loss_function=None,
+        deriv_mode="auto",
+        grid=None,
+        name="coil convexity",
```
dpanici, line 649: should be "coil convexity error"  I think, since you are subtracting by 2pi.

dpanici (dismissed):
- give a `self._endpoint` attribute to `CoilConvexity` to be True
- check for this attribute in `_CoilObjective.build` to decide if `endpoint` should be true or not (False by default if `hasattr(self._endpoint)` is False
- move errorif statement to after `super.build()` in `build` of `CoilConvexity`
- make docstring clearer for `CoilConvexity`
- Fix docs

dpanici:
Call it `IntegratedCurvature`

desc/objectives/_coils.py
```diff
@@ -602,6 +611,111 @@ def compute(self, params, constants=None):
         return out
 
 
+class IntegratedCurvature(_CoilObjective):
```
f0uriest, line 614: Do we want to include the word `Coil` somewhere in the name? (ie to distinguish from the integrated curvature of the plasma boundary or something)
 `IntegratedCoilCurvature`? `CoilIntegratedCurvature`? `CoilCurvatureIntegrated`?

desc/objectives/_coils.py
```diff
@@ -602,6 +611,111 @@ def compute(self, params, constants=None):
...
+        -------
+        f : array of floats
+            Integrated curvature.
+
+        """
+        data = super().compute(params, constants=constants)
+        data = tree_leaves(data, is_leaf=lambda x: isinstance(x, dict))
+        out = jnp.array(
+            [
+                trapezoid(
```
f0uriest, line 710: I think you can get the same thing if you leave off the endpoint and do `(curvature*x_s*ds).sum()`. The integrand is periodic so the endpoint and start are the same.

desc/objectives/_coils.py
```diff
@@ -602,6 +611,111 @@ def compute(self, params, constants=None):
         return out
 
 
+class IntegratedCurvature(_CoilObjective):
+    """Curvature integrated along a coil.
+
+    If a curve is convex, then the following condition must be true: ∫ κ |∂ₛx| ds = 2π.
```
f0uriest, line 617: should mention what kappa, x_s etc mean here

tests/test_objective_funs.py
```diff
@@ -2662,6 +2703,7 @@ class TestComputeScalarResolution:
         BootstrapRedlConsistency,
         BoundaryError,
         CoilArclengthVariance,
+        IntegratedCurvature,
```
f0uriest, line 2706: keep this alphabetized

tests/test_objective_funs.py
```diff
@@ -3084,6 +3126,7 @@ def test_compute_scalar_resolution_others(self, objective):
         "objective",
         [
             CoilArclengthVariance,
+            IntegratedCurvature,
```
f0uriest, line 3129: keep this alphabetized

tests/test_objective_funs.py
```diff
@@ -3144,9 +3191,10 @@ class TestObjectiveNaNGrad:
         BootstrapRedlConsistency,
         BoundaryError,
         CoilArclengthVariance,
-        CoilLength,
+        IntegratedCurvature,
```
f0uriest, line 3194: keep this alphabetized

tests/test_objective_funs.py
```diff
@@ -3374,6 +3422,7 @@ def test_objective_no_nangrad(self, objective):
         "objective",
         [
             CoilArclengthVariance,
+            IntegratedCurvature,
```
f0uriest, line 3425: keep this alphabetized

f0uriest (changes requested):

desc/objectives/_coils.py
YigitElma: Although I like `CoilIntegratedCurvature` more, for sake of not having l and capital i next to each other, I vote for `CoilCurvatureIntegrated`.

desc/objectives/_coils.py
```diff
@@ -602,6 +611,111 @@ def compute(self, params, constants=None):
         return out
 
 
+class IntegratedCurvature(_CoilObjective):
+    """Curvature integrated along a coil.
```
YigitElma:
```suggestion
    """Curvature integrated along each coil.
```

desc/objectives/_coils.py
```diff
@@ -602,6 +611,111 @@ def compute(self, params, constants=None):
...
+
+    def compute(self, params, constants=None):
+        """Compute integrated curvature.
+
+        Parameters
+        ----------
+        params : dict
+            Dictionary of the coil's degrees of freedom.
+        constants : dict
+            Dictionary of constant data, eg transforms, profiles etc. Defaults to
```
YigitElma, line 689:
```suggestion
            Dictionary of constant data, e.g. transforms, profiles etc. Defaults to
```

desc/objectives/_coils.py
```diff
@@ -602,6 +611,111 @@ def compute(self, params, constants=None):
...
+        f : array of floats
+            Integrated curvature.
+
+        """
+        data = super().compute(params, constants=constants)
+        data = tree_leaves(data, is_leaf=lambda x: isinstance(x, dict))
+        out = jnp.array(
+            [
+                trapezoid(
+                    jnp.abs(dat["curvature"]) * safenorm(dat["x_s"], axis=1), dat["s"]
```
YigitElma: I don't know the actual formula but from the integral, I didn't understand why there `jnp.abs(kappa)` here? Can you explain?

desc/objectives/_coils.py
ddudt: This is slightly less accurate then the trapezoid integration (`atol=2e-5` vs `atol=1e-6` at the grid resolution used in the tests) but I still like this change for the simplicity.

tests/test_objective_funs.py
```diff
@@ -3099,7 +3142,11 @@ def test_compute_scalar_resolution_coils(self, objective):
         f = np.zeros_like(self.res_array, dtype=float)
         for i, res in enumerate(self.res_array):
             obj = ObjectiveFunction(
-                objective(coilset, grid=LinearGrid(N=int(5 + 3 * res))), use_jit=False
+                objective(
+                    coilset,
+                    grid=LinearGrid(N=int(5 + 3 * res), endpoint=objective._endpoint),
```
YigitElma, line 3147: `CoilSetLinkingNumber` and `CoilSetMinDistance` are not a subclass of `_CoilObjective`, so don't have `_endpoint` attribute.

desc/objectives/_coils.py
ddudt: We want to integrate the unsigned curvature, but our compute quantity `"kappa"` is the signed curvature, where the sign denotes the local convexity.

YigitElma: Thanks, the explanation in the docs makes it clear now.

YigitElma (approved):

YigitElma (dismissed):

f0uriest (approved):

desc/objectives/_coils.py
```diff
@@ -602,6 +600,112 @@ def compute(self, params, constants=None):
         return out
 
 
+class CoilIntegratedCurvature(_CoilObjective):
+    """Coil integrated curvature.
+
+    If a curve is convex, then the following condition must be true: ∫ κ |∂ₛx| ds = 2π
```
f0uriest, line 606: needs to be `||∂ₛx||` to avoid rst issues

f0uriest (approved):

YigitElma (approved):

PR #1488, opened 2024-12-22
Safediv to avoid nan generation in conngould trust radius...

unalmis:
This is necessary to avoid tripng the jax nan debugger in a dummy objective. Useful for future as well.

### DIFF
--- a/desc/optimize/least_squares.py
+++ b/desc/optimize/least_squares.py
@@ -208,12 +208,12 @@ def lsqtr(  # noqa: C901
     # conngould : norm of the cauchy point, as recommended in ch17 of Conn & Gould
     # scipy : norm of the scaled x, as used in scipy
     # mix : geometric mean of conngould and scipy
-    scipy = jnp.linalg.norm(x * scale_inv / v**0.5)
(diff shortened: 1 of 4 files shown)


desc/optimize/least_squares.py
```diff
@@ -208,14 +208,12 @@ def lsqtr(  # noqa: C901
     # conngould : norm of the cauchy point, as recommended in ch17 of Conn & Gould
     # scipy : norm of the scaled x, as used in scipy
     # mix : geometric mean of conngould and scipy
+    scipy = jnp.linalg.norm(x * scale_inv / v**0.5)
```
YigitElma, line 211: I will be annoying but I don't like the idea of having `scipy` as variable name, especially since this file is mostly optimization stuff and future we might add some actual scipy functions.

Maybe something like `tr_scipy`?

YigitElma (changes requested):

YigitElma:
Also, there are some other instances of `init_tr`, i.e. `aug_lagrangian_ls.py`. The ones in `aug_lagrangian.py` and `fmin_scalar.py` are slightly different but can use `safediv` too.

YigitElma (approved):

PR #1490, opened 2024-12-23
enable selecting gpu

brokenhammer:
Add a parameter "gpuid" in "set_device" to manually assign a GPU.


desc/__init__.py
```diff
@@ -127,11 +127,15 @@ def set_device(kind="cpu"):
...
-            mem = dev["mem_total"] - dev["mem_used"]
-            if mem > maxmem:
-                maxmem = mem
-                selected_gpu = dev
+
+        if (not (gpuid == None)) and (str(gpuid) in gpu_ids):
+            selected_gpu = [dev for dev in devices if dev["index"] == str(gpuid)][0]
+        else:
+            for dev in devices:
+                mem = dev["mem_total"] - dev["mem_used"]
```
dpanici, line 135: we should still have a check on the memory of the selected gpu so we print an accurate memory later in this function

dpanici: ah nevermind I see this is just checking which device has the max memory and choosing that one

dpanici (dismissed):
Thanks for the PR! Approving but I'd like @f0uriest to also look over before merging

f0uriest:
My main question is what benefit this has over setting `CUDA_VISIBLE_DEVICES=...` which is what we currently recommend. I could also see there being possible issues in cases where a user tries to do both.

We'll also need to test this a bit manually since our CI doesn't do any GPU stuff.

desc/__init__.py
```diff
@@ -127,11 +127,15 @@ def set_device(kind="cpu"):
...
             return
         devices = [dev for dev in devices if dev["index"] in gpu_ids]
-        for dev in devices:
-            mem = dev["mem_total"] - dev["mem_used"]
-            if mem > maxmem:
-                maxmem = mem
-                selected_gpu = dev
+
+        if (not (gpuid == None)) and (str(gpuid) in gpu_ids):
+            selected_gpu = [dev for dev in devices if dev["index"] == str(gpuid)][0]
```
YigitElma, line 132: What if the user tries to assign a wrong ID? Shouldn't we throw an error?

desc/__init__.py
```diff
@@ -61,7 +61,7 @@ def __getattr__(name):
 config = {"device": None, "avail_mem": None, "kind": None}
 
 
-def set_device(kind="cpu"):
```
YigitElma, line 64: The docs should be updated.

desc/__init__.py
```diff
@@ -61,7 +61,7 @@ def __getattr__(name):
...
+def set_device(kind="cpu",gpuid=None):
```
YigitElma, line 64:
```suggestion
def set_device(kind="cpu", gpuid=None):
```

YigitElma (changes requested):
To @f0uriest's point, I believe 
```python
os.environ["CUDA_VISIBLE_DEVICES"] = str(gpuid)
```
is also fine, but having this functionality in `set_device` will help printing correct device type and available memory at the initialization. Also, probably better for the default values of `jac_chunk_size` which checks the `config["avail_mem"]`, iirc.

desc/__init__.py
```diff
@@ -127,11 +127,15 @@ def set_device(kind="cpu"):
             set_device(kind="cpu")
             return
         devices = [dev for dev in devices if dev["index"] in gpu_ids]
-        for dev in devices:
-            mem = dev["mem_total"] - dev["mem_used"]
-            if mem > maxmem:
-                maxmem = mem
-                selected_gpu = dev
+
+        if (not (gpuid == None)) and (str(gpuid) in gpu_ids):
```
unalmis, line 131:
```suggestion
        if gpuid is not None and (str(gpuid) in gpu_ids):
```

desc/__init__.py
brokenhammer: Thank you for suggesting. If the wrong ID is provided, it should fall back to the previous method: selecting the GPU with maximum free memory. I added a warning message for this case.

desc/__init__.py
brokenhammer: The docs are updated now.

desc/__init__.py
brokenhammer: Updates are made according to suggestion.

brokenhammer:
> To @f0uriest's point, I believe
> 
> ```python
> os.environ["CUDA_VISIBLE_DEVICES"] = str(gpuid)
> ```
> 
> is also fine, but having this functionality in `set_device` will help printing correct device type and available memory at the initialization. Also, probably better for the default values of `jac_chunk_size` which checks the `config["avail_mem"]`, iirc.

Besides, adding the parameter makes it convenient to assign different GPUs to different processes in a multi-process run.

desc/__init__.py
```diff
@@ -61,11 +61,12 @@ def __getattr__(name):
...
 
 
-def set_device(kind="cpu"):
+def set_device(kind="cpu", gpuid=None):
     """Sets the device to use for computation.
 
-    If kind==``'gpu'``, checks available GPUs and selects the one with the most
-    available memory.
+    If kind==``'gpu'`` and a gpuid is specified, uses the specified GPU. If 
+    gpud==``None`` or a wrong GPU id is given, checks available GPUs and selects the
```
YigitElma, line 68:
```suggestion
    gpuid==``None`` or a wrong GPU id is given, checks available GPUs and selects the
```

YigitElma (approved):

YigitElma (changes requested):
The black formatting test seems to fail. I believe if you fix that this should be good to go.

YigitElma (approved):

dpanici (approved):

PR #1494, opened 2024-12-25
Ku/bounce alpha

unalmis:
- [x] Enable tracking multiple fieldlines with Bounce2D 
- [x] simplify ability to analyze drift over alpha as requested by a few people.
- [x] other changes requested on previous pull request.


desc/compute/_fast_ion.py
```diff
@@ -180,52 +182,49 @@ def fun(pitch_inv):
                     ),
                 )
             )
-            return jnp.sum(v_tau * gamma_c**2, axis=-1)
+            return jnp.sum(v_tau * gamma_c**2, axis=-1).mean(axis=-2)
```
unalmis, line 185: the tldr is to replace this mean over the alpha axis with the logic on `gh/Gamma_c`.

dpanici: which logic from that branch specifically?

unalmis: for gamma_alpha and gamma_d you have to perform a different operation over the poloidal angle than `.mean()`

desc/integrals/quad_utils.py
```diff
@@ -298,7 +298,7 @@ def chebgauss2(deg):
     # github.com/scipy/scipy/blob/v1.14.1/scipy/special/_orthogonal.py#L1803-L1851.
     t = jnp.arange(deg, 0, -1) * jnp.pi / (deg + 1)
     x = jnp.cos(t)
-    w = jnp.pi * jnp.abs(jnp.sin(t)) / (deg + 1)
+    w = jnp.pi * jnp.sin(t) / (deg + 1)
```
dpanici, line 301: why the removal of the abs? shouldnt this be positive?

unalmis: the domain `t` makes this positive regardless. less operations means less gradients so less memory

tests/test_fast_ion.py
```diff
@@ -65,15 +66,15 @@ def test_Gamma_c_Nemov_1D():
     Y_B = 100
     num_transit = 10
     num_well = 20 * num_transit
-    rho = np.linspace(0, 1, 10)
+    rho = np.linspace(1e-12, 1, 10)
```
dpanici, line 69: what changed at 0? or was it just analytically 0 before and you want to check it is not blowing up?

unalmis: It's zero analytically on axis, which I confirmed when I wrote the quantity. The reason for the change is here
- For the 2D methods, including zero will compute the axis limits of the dependencies, which will slow down the tests for no reason.
- For the 1D methods, `get_rtz_grid` makes a jitable grid by default which [moves the nodes to 1e-12](https://github.com/PlasmaControl/DESC/blob/6b069f3f2994de2f0297dce876b223eb8aedd6a1/desc/grid.py#L764) anyway.

dpanici (changes requested):
just the one comment on the plotting when grid has multiple rho for check_points, either
- change check_points to have a rho/alpha label somewhere in the title
- or just make a comment in that notebook that they are at different rho, same alpha, and make a TODO for changing check_points

unalmis:
<p>Yes typically there is a finer structure near axis, but the integration domain is also much smaller (max B - min B is small)</p><p> </p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1494/discussion/#comment-2568311789'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

dpanici (approved):

rahulgaur104:
Here's what I am running to test this

```python

#!/usr/bin/env python3
import numpy as np
from desc.examples import get
from desc.integrals import Bounce2D

import pdb

eq = get("W7-X")

rho = 0.5
X, Y = 16, 32
theta = Bounce2D.compute_theta(eq, X, Y, rho)
num_transit = 3
num_well = 10 * num_transit
num_quad = 32

Y_B = 32

N_alpha = int(4)
num_pitch = int(32)

alpha_arr = np.linspace(0, 2*np.pi, N_alpha)
grid = LinearGrid(rho=rho, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False)

data = eq.compute(
            "Gamma_c",
            grid=grid,
            theta=theta,
            Y_B=Y_B,
            num_transit=num_transit,
            num_well=num_well,
            num_quad=num_quad,
            num_pitch=num_pitch,
            alpha =alpha_arr,
            # Can also specify ``pitch_batch_size`` which determines the
            # number of pitch values to compute simultaneously.
            # Reduce this if insufficient memory. If insufficient memory is detected
            # early then the code will exit and return ε = 0 everywhere. If not detected
            # early then typical OOM errors will occur.
            )

pdb.set_trace()

```
but the shape of gamma_c is still (1, num_pitch)

rahulgaur104 (approved):

rahulgaur104:
Codecov short by 0.65%. Merging now.

PR #1495, opened 2024-12-25
Support for Multiple Devices and Parallelism via MPI

### LINKED ISSUE #1071: Implement multi-gpu functionality for separate objectives with MPI4Jax
The general idea is

- Physics objectives are put onto separate GPUs
- Constraints are taken out
- Then combine jacobians on a single GPU to create the $A$ matrix, and perform SVD on this $A$ matrix
- From this a new `eq` is created, which is then fed back to each GPU

### LINKED ISSUE #1601: Parallelization options
This is not an issue but rather a discussion. In #1495, I have been testing different parallelization options for the Jacobian calculation. A quick recap,
What #1495 does is, it adds a device option to objectives, and if there are multiple devices in an optimization problem, each objective runs on the corresponding device. This uses `blocked` strategy for the Jacobian, so concatenates individual Jacobian calls to a final matrix.

Then, the question is how to run each objective's `compute_` and `jac_` methods in parallel? This is not possible with native JAX currently. So, I tried some general strategies,

- MPI
  - If the part you want to use MPI for is not jitted, you can still use `mpi4py`. Since we will run objectives on different devices, and jitted functions have to run on a single device ( [related issue](https://github.com/jax-ml/jax/discussions/18789#discussioncomment-8487343) ), we already eliminated hitting so `mpi4py` can be used. 
  - The design problem is then, how to create a bunch of workers but only use them during Jacobian call? This is possible with some conditionals like `if rank == 0:` but this might look ugly. Basically, we need a general way of achieving this for all optimizers and perturb (where jacobian is called)
  - `mpi4py` requires the transferred array to be numpy, jax arrays cause issue, so we should first cast them to numpy and then send the message.

- Multiprocessing
  - In terms of clean code, this is much better, because you can create processes when needed. But the problem is JAX doesn't allow the use of `fork` in multiprocessing, with the `spawn` method creating the processes take a long time. For example, all of the workers need to import `desc.compute` at some point, and even that takes 2 seconds....! This is definitely more than the jacobian calculation itself. 
  - So, we should create a process (ideally pass the built objectives once at the beginning, then only pass the `x`). This is basically `mpi` though 😄 

Any feedback is appreciated 🙏

YigitElma:
Support for multi-device optimization.

- Updates `set_device` and `desc.backend` for initialization of multiple devices. These are the cause of slightly low coverage (around 90% for patch)
- Adds `_device_id` and `_device` to `_Objective` class (defaults to 0) for making parallelization work with other objectives 
- Uses `mpi4py` for compute and jvp methods of parallel objective functions (there is no separate class but if user passes objectives with different devices and an mpi communicator, parallel strategy will be used.) To make this work a user has to use context manager properly
- Adds `_rank` to `_Objective` that specifies which objective should run on which rank. Default is `None`, and if all sub-objectives have `None`, `ObjectiveFunction` defaults to `np.arange(len(objs))`. There are some limitations for the user. For example, 2 different rank cannot use the same device (maybe it can, but it shouldn't in practice). One tricky part about this is, for multiple node case, same `device_id` can mean different devices. The reason we have `rank` is to be able to distinguish that (MPI.size is equal to total number of devices in all nodes, whereas max(device_id) can be only the number of GPUs connected to a single node)

Things to consider in future PRs,
- Maybe implement a new optimizer that uses distributed matrix operations instead of QR and SVD.
- Find good distributed linear algebra routines for SVD and QR (necessary for speed and low memory usage since the current implementation will use CPU if the Jacobian is big)
  - Something like TSQR implementation via sharding? This should reduce QR of (C*M)xN matrix into C MxN QRs plus a single (C*N)xN QR. By some careful implementation (without storing Q, like in our custom `qr_multiply`), this can be faster.
- Use CUDA-aware MPI (possibly without `mpi4jax`). I added `desc_config["mpi-cuda"]` for related parts, it should be easy to implement in the future
- Incorporate sharding if possible?

Note for reviewers:
- Actual code changes live in `objective_funs.py`, `_constraint_wrappers.py` and `optimizer.py`
- All the changes in `_Objective` class are adding `_device_id` and `_rank`
- There are around 1100 lines of code just for documentation and ~750 lines for testing

Resolves #1071 (but with `mpi4py`)
Resolves #1601


YigitElma:
Some test scripts,
```python
import nvgpu 
devices = nvgpu.gpu_info()
print([dev["type"] for dev in devices])

from desc import set_device
set_device("gpu", num_device=3)

eq = get("W7-X")
cons = get_fixed_boundary_constraints(eq)
cons = maybe_add_self_consistency(eq, cons)
cons = ObjectiveFunction(cons)
objective = LinearConstraintProjection(objective=obj, constraint=cons)
objective.build()

print(objective.compute_scaled_error(objective.x(eq)).shape)
print(objective.jac_scaled_error(objective.x(eq)).shape)
jac = objective.jac_scaled_error(objective.x(eq))
_, _ = qr(jac, mode="economic") 
%timeit _ = objective.compute_scaled_error(objective.x(eq)).block_until_ready()
%timeit _ = objective.jac_scaled_error(objective.x(eq)).block_until_ready()
%timeit _, _ = qr(jac, mode="economic") 
jax.debug.visualize_array_sharding(jac)
```

YigitElma:
```python
@jax.jit
def dummy_fun(x):
    return x*jnp.ones(N)

@jax.jit
def dummy_fun2(x):
    return x[0]*jnp.ones(N)

N = 55000
# x = jnp.ones(N)
# v = jnp.eye(N)[:,1]
# f, df = jax.jvp(dummy_fun2, (x,), (v,))

x = 1.
v = 1.
f, df = jax.jvp(dummy_fun, (x,), (v,))
```

dpanici:
#763
check for overlap with this one

YigitElma:
This is not yet finalized, but I think this is the main structure I would like to have. Let me know your feedback.

desc/objectives/objective_funs.py
```diff
@@ -436,12 +453,23 @@ def compute_unscaled(self, x, constants=None):
...
+            f = jnp.concatenate(
+                [
+                    obj.compute_unscaled(*par, constants=const)
+                    for par, obj, const in zip(params, self.objectives, constants)
+                ]
+            )
+        else:
+            f = pconcat(
+                [
+                    obj.compute_unscaled(
```
dpanici, line 466: see if we can change this around to make JAX do it asynchronously instead of serially

dpanici: maybe removing this from list comprehension will help?

desc/__init__.py
```diff
@@ -58,10 +58,10 @@ def __getattr__(name):
 BANNER = colored(_BANNER, "magenta")
 
 
-config = {"device": None, "avail_mem": None, "kind": None}
+config = {"device": None, "avail_mem": None, "kind": None, "num_device": None}
 
 
-def set_device(kind="cpu", gpuid=None):
+def set_device(kind="cpu", gpuid=None, num_device=1):  # noqa : C901
```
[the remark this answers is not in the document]
f0uriest: though we may also want to allow for multiple CPU devices, useful for testing this stuff. You can pretend a single cpu is multiple devices with
```python
xla_flags = os.getenv("XLA_FLAGS", "")
xla_flags = re.sub(r"--xla_force_host_platform_device_count=\S+", "", xla_flags).split()
os.environ["XLA_FLAGS"] = " ".join(
    [f"--xla_force_host_platform_device_count=8"] + xla_flags
)
```

This would allow us to test this code on CI

desc/backend.py
```diff
@@ -32,6 +32,15 @@
             from jax import config as jax_config
 
             jax_config.update("jax_enable_x64", True)
+            if desc_config["num_device"] != 1:
+                # for now, these are not used. Delete them if they are not needed.
+                mesh = jax.make_mesh((desc_config["num_device"],), ("grid"))
```
f0uriest, line 37: is this needed? I thought you said you weren't sharding anything?

desc/backend.py
```diff
@@ -425,6 +443,67 @@ def tangent_solve(g, y):
...
+        -------
+        out : jnp.ndarray
+            Concatenated array that lives on CPU.
+        """
+        # we will use either CPU or GPU[0] for the matrix decompositions, so the
+        # array of float64 should fit into single device
+        size = jnp.array([x.size for x in arrays])
+        size = jnp.sum(size)
+        if size * 8 / (1024**3) > desc_config["avail_mems"][0]:
+            device = jax.devices("cpu")[0]
```
f0uriest, line 466: should at least throw a warning here, or possibly an error.

desc/objectives/_equilibrium.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         grid=None,
         name="force",
         jac_chunk_size=None,
+        device_id=0,
```
f0uriest, line 64: This might be better to pass an instance of `jax.device` or something?

desc/objectives/getters.py
```diff
@@ -344,3 +344,78 @@ def add_if_multiple(constraints, cls):
         constraints = add_if_multiple(constraints, FixCurveRotation)
 
     return constraints
+
+
+def get_parallel_forcebalance(
```
f0uriest, line 349: not sure if we really want/need this function. I think it's better for now to make the user control objectives and devices more explicitly.

(also I think needing multiple GPUs for a single force objective seems like a rather uncommon use case)

desc/objectives/objective_funs.py
```diff
@@ -292,19 +297,28 @@ def _unjit(self):
...
     @execute_on_cpu
-    def build(self, use_jit=None, verbose=1):
+    def build(self, use_jit=None, use_jit_wrapper=True, verbose=1):  # noqa:  C901
         """Build the objective.
 
         Parameters
         ----------
         use_jit : bool, optional
             Whether to just-in-time compile the objective and derivatives.
+        use_jit_wrapper : bool, optional
```
f0uriest, line 307: does this need to be an argument? shouldn't it be done automatically?

desc/objectives/objective_funs.py
```diff
@@ -292,19 +297,28 @@ def _unjit(self):
...
+            Whether to use the jit wrapper for the objective. If multiple GPUs are
+            used, this will be set to False.
         verbose : int, optional
             Level of output.
 
         """
         if use_jit is not None:
             self._use_jit = use_jit
+            if use_jit is False:
+                use_jit_wrapper = False
```
f0uriest, line 317: need to also set this to something outside of the double if block

f0uriest (commented):
some initial thoughts

desc/__init__.py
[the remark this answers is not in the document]
YigitElma: Yes, I just saw this on your old PR. I can add a test when I am done with GPU version.

desc/backend.py
YigitElma: I will delete them eventually. These were very useful for reducing the amount of code when I was experimenting with it.

desc/objectives/_equilibrium.py
YigitElma: I think from a user standpoint just passing an integer is easier. Otherwise, they have to type `jax.devices("gpu")[0]`

desc/backend.py
YigitElma: It now throws a warning that can be suppressed by some configuration parameter. I don't know the proper way of setting flags, but I tried something.

YigitElma: The reason I want to be able to suppress it is that each jacobian evaluation will throw this error and this might get annoying.

desc/objectives/_equilibrium.py
YigitElma: But I can add another attribute `objective._device` which can be jax.device, something like
```python
from desc import config as desc_config
self._device = jax.devices(desc_config['kind'])[self._device_id)]
```
This can make device_put calls cleaner, since currently I assume multiple GPU not CPU.

desc/objectives/getters.py
YigitElma: This was mainly a helper function for testing, but I think it could be useful where you have a high res equilibrium you try to optimize using Proximal. Even thought the outer Jacobian is rather small, the ForceBalance part might limit the optimization. So, we can have a this kind of convenience function that is used like,

```python
parallel_force = get_parallel_force(eq, num_device) # assume this returns a tuple 
constraints = (
    *prallel_force,
    FixBoundaryZ(eq=eq, modes=Z_modes),
    FixPressure(eq=eq),
    FixPsi(eq=eq),
    FixCurrent(eq=eq),
)
```
Basically, this can make grid creation and putting objectives to GPUs automatically.

desc/objectives/_equilibrium.py
YigitElma: I implemented above comment.

desc/objectives/utils.py
```diff
@@ -139,6 +139,7 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
     xp = put(xp, unfixed_idx, A_inv @ b)
     xp = put(xp, fixed_idx, ((1 / D) * xp)[fixed_idx])
     # cast to jnp arrays
+    # TODO: might consider sharding these
```
YigitElma, line 142: I mean for multiple GPU case, if we use sharding for Z, then `project` and `recover` (which are called a lot) will be faster. It is just a matrix vector product but still...

YigitElma:
@f0uriest @dpanici @rahulgaur104 @ddudt One more review before hackathon would be great.
Updates:
- Each rank can now have multiple objectives. These objectives will be computed in a for-loop as if they are blocked
- For catching some future problems, I set the `ObjectiveFunction._jac_chunk_size` of objectives with multiple objective and sub-objectives have different `jac_chunk_size` to an array (like an error message). If someone ever tries to use it, it will throw an error, and show the error message.

desc/__init__.py
```diff
@@ -100,55 +178,62 @@ def set_device(kind="cpu", gpuid=None):
...
-            selected_gpu["mem_total"] - selected_gpu["mem_used"]
-        ) / 1024  # in GB
-        os.environ["CUDA_VISIBLE_DEVICES"] = str(selected_gpu["index"])
+            if gpuid is not None:
+                # TODO: implement multiple GPU selection
+                raise ValueError("Cannot specify `gpuid` when requesting multiple GPUs")
+
+        config["avail_mems"] = [
+            memories[dev["index"]] / 1024 for dev in devices[:num_device]
+        ]  # in GB
```
rahulgaur104, line 233: Is the default value in MB here?

YigitElma: Yeap

desc/backend.py
```diff
@@ -452,11 +472,82 @@ def bodyfun(state):
...
+                    "GPU[id=0]. Moving the array to CPU. This may cause performance "
+                    "degredation. To suppress this warning, use "
+                    "`from desc import config as desc_config` \n"
+                    "`desc_config['SUPPRESS_CPU_WARNING'] = True`"
+                )
+            device = jax.devices("cpu")[0]
+        else:
+            device = jax.devices("gpu")[0]
+
+        if mode == "concat":
```
rahulgaur104, line 514: Does this cove all the cases of stacking?
I guess it's only used for objectives or jacobians which can never have more than 2 dimensions?

desc/objectives/getters.py
```diff
@@ -367,3 +367,63 @@ def add_if_multiple(constraints, cls):
...
+            f"Number of devices in desc_config ({desc_config['num_device']}) "
+            f"is less than the number of devices in input ({num_device})."
+        )
+    if grid is not None:
+        if len(grid) != num_device:
+            raise ValueError(
+                f"Number of grids and num_device must be the same! Got "
+                f"{len(grid)=} and {num_device=}."
+            )
+    if eq.L_grid % num_device == 0:
```
rahulgaur104, line 402: This idea of splitting the grid is great only for radially local objectives, right?
If I have an objective where I need the full grid for the compute function this won't work.

desc/backend.py
YigitElma: I think for the parallel stacking applications, it covers all, at least I have never seen a function other than `concatenate`, `hstack` and `vstack`.

desc/objectives/objective_funs.py
```diff
@@ -267,6 +285,206 @@ def __init__(
...
+                break
+            elif "jvp" in message[0] and "proximal" not in message[0]:
+                print(
+                    f"Rank {self.rank} : {message[0]} for objectives ids: "
+                    + f"{obj_idx_rank}"
+                )
+                # inputs to jitted functions must live on the same device. Need to
+                # put xi and vi on the same device as the objective
+                J_rank = []
+                for idx in obj_idx_rank:
```
rahulgaur104, line 427: I am assuming each rank will have it's own value of obj_idx_rank. 
But are all the objectives shared across all the rank?
Because when you do self.objectives[idx], that operation is happening at all ranks,
so each rank must have self.objectives.

desc/objectives/getters.py
YigitElma: Yes, not off the shelf. But maybe one can go around this limitation by changing the objective's compute function. For example, if a quantity requires a special grid, compute it with full grid on each instance (rank/process), then pass it to `compute_fun` through `data`.  But it requires changes to some objectives. I was thinking of using this kind of grid splitting for free-boundary (like obj1->phi=[0,pi/NFP], obj2->phi=[pi/NFP, 2pi/NFP]), we can introduce parallel version or add some keywords to them.

desc/objectives/objective_funs.py
YigitElma: Yes, each rank has the built objectives that are on the device they need to be. Only the computation happens on different ranks.

`obj_idx_rank = self._obj_per_rank[self.rank]` is the indices of objectives that belong to this rank, so they are different for each rank.

desc/__init__.py
```diff
@@ -58,36 +61,114 @@ def __getattr__(name):
...
+            config["avail_mems"] = [cpu_mem]
+        else:
+            try:
+                import jax
+
+                jax_cpu = jax.devices("cpu")
+                assert len(jax_cpu) == num_device
+                # These CPUs doesn't have to be the same model, but I think slurm will
+                # always give same model
+                config["devices"] = [f"{cpu_info + " " + str(dev)}" for dev in jax_cpu]
```
YigitElma, line 162:
```suggestion
                config["devices"] = [f"{cpu_info + ' ' + str(dev)}" for dev in jax_cpu]
```

YigitElma:
Ok I think instead of

```python
 from desc import _set_cpu_count, set_device
 
 num_device = 2
 # These will be used for diving the single CPU into multiple virtual CPUs
 # such that JAX and XLA thinks there are multiple devices
 # If you have multiple CPUs, this will just use 1 of them, and split it into multiple fake versions
 _set_cpu_count(num_device)
 set_device("cpu", num_device=num_device)
```
I should have this for initializing multiple cpu cases. But just for multiple cpu and 0 gpu case. If there is a GPU current version is fine,

```python
 # to be able to see all CPUs
 # You should start this script using srun
 import jax
 jax.distributed.initialize()
 
 from desc import set_device
 
 num_device = 2
 set_device("cpu", num_device=num_device)
```

docs/notebooks/tutorials/mpi-tutorials/mpi-eq-solve.py
```diff
@@ -0,0 +1,92 @@
...
+        print(f"====== TOTAL OF {size} RANKS ======")
+
+    # see which rank is running on which device
+    # Note: JAX has 2 functions for this: `jax.devices()` and `jax.local_devices()`
+    # `jax.devices()` will return all devices available to JAX, while `jax.local_devices()`
+    # will return only the devices that are available to the current process. This is
+    # useful when you have multiple processes running on multiple nodes and you want
+    # to see which devices are available to each process.
+    if desc_config["kind"] == "gpu":
+        print(
```
rahulgaur104, line 52: This will throw a syntax error because both your outer and inner quotes are ". Make the outer quotes '.

rahulgaur104: Make the same changes to the tutorial notebook as well.

docs/notebooks/tutorials/mpi-tutorials/mpi-proximal.py
```diff
@@ -0,0 +1,148 @@
...
+
+    # see which rank is running on which device
+    # Note: JAX has 2 functions for this: `jax.devices()` and `jax.local_devices()`
+    # `jax.devices()` will return all devices available to JAX, while `jax.local_devices()`
+    # will return only the devices that are available to the current process. This is
+    # useful when you have multiple processes running on multiple nodes and you want
+    # to see which devices are available to each process.
+    if desc_config["kind"] == "gpu":
+        print(
+            f"Rank {rank} is running on {jax.local_devices(backend="gpu")} "
```
rahulgaur104, line 64: Same as above.

docs/notebooks/tutorials/multi_device.ipynb
rahulgaur104, line 1: There should be a check for mpi4py installation and maybe a simple test. Because even after installing mpi4py it may not yield the expected behaviour.

Maybe this could be a test
subprocess.call[;'python3 -c "import mpi4py" ']
and use the status to define whther mpi4py exists.

docs/notebooks/tutorials/mpi-tutorials/mpi-eq-solve.py
```diff
@@ -0,0 +1,92 @@
...
+import sys
+
+# Add the path to the parent directory to augment search for module
+sys.path.insert(0, os.path.abspath("."))
+sys.path.append(os.path.abspath("../../../"))
+
+from desc import _set_cpu_count, set_device
+
+# ====== Using CPUs ======
+num_device = 4
```
rahulgaur104, line 11: when I set num_device = 1, I get the following error

```python3

mpirun -n 1 python3 -u mpi-eq-solve.py 
====== TOTAL OF 1 RANKS ======
Rank 0 is running on [CpuDevice(id=0)]

====== BACKEND INFO ======
DESC version=0.14.2+133.g50e5239ed.dirty.
Using JAX backend: jax version=0.4.31, jaxlib version=0.4.31, dtype=float64.
CPU Info:  AMD Ryzen 7 7735U with Radeon Graphics CPU with 21.98 GB available memory


Starting timer!
Building objective: force
Precomputing transforms
------------------------------------------------------------
Rank 0 will run objective(s): ['ForceBalance']
------------------------------------------------------------
Building objective: lcfs R
Building objective: lcfs Z
Building objective: fixed Psi
Building objective: fixed pressure
Building objective: fixed iota
Building objective: fixed sheet current
Building objective: self_consistency R
Building objective: self_consistency Z
Building objective: lambda gauge
Building objective: axis R self consistency
Building objective: axis Z self consistency
Timer: Objective build = 1.03 sec
Timer: LinearConstraintProjection build = 4.17 sec
Number of parameters: 551
Number of objectives: 8424
Timer: Initializing the optimization = 5.24 sec

Starting optimization
Using method: lsq-exact
  File "/home/rgaur/DESC/docs/notebooks/tutorials/mpi-tutorials/mpi-eq-solve.py", line 84, in <module>
    eq.solve(
  File "/home/rgaur/DESC/desc/equilibrium/equilibrium.py", line 2254, in solve
    things, result = optimizer.optimize(
                     ^^^^^^^^^^^^^^^^^^^
  File "/home/rgaur/DESC/desc/optimize/optimizer.py", line 278, in optimize
    result = optimizers[method]["fun"](
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rgaur/DESC/desc/optimize/_desc_wrappers.py", line 270, in _optimize_desc_least_squares
    result = lsqtr(
             ^^^^^^
  File "/home/rgaur/DESC/desc/optimize/least_squares.py", line 178, in lsqtr
    f = fun(x, *args)
        ^^^^^^^^^^^^^
  File "/home/rgaur/DESC/desc/optimize/_constraint_wrappers.py", line 314, in compute_scaled_error
    f = self._objective.compute_scaled_error(x, constants)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Cannot interpret value of type <class 'mpi4py.MPI.Intracomm'> as an abstract array; it does not have a dtype attribute
--------------------
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[38784,1],0]
  Exit code:    1
--------------------------------------------------------------------------

```

rahulgaur104:
So I only compared the mpi-eq-solve script timings with 2 and 4 processes. With 2 processes it takes ~ 31 s but with 4 it takes ~ 180s. This is probably because a lot of time is being spent on gather compute_scaled_error data.

Maybe this will change when we use multiple GPUs.

YigitElma:
> So I only compared the mpi-eq-solve script timings with 2 and 4 processes. With 2 processes it takes ~ 31 s but with 4 it takes ~ 180s. This is probably because a lot of time is being spent on gather compute_scaled_error data.
> 
> Maybe this is happening because the grid cannot be divided across 4 CPU and not the communication.
> 

What are the number of parameters and number of objectives for both cases?

If it is the same, yes probably the transfer overhead is high. If they are not, then it is peobably the QR decomposition difference. Because QR is still happening on single device.

docs/notebooks/tutorials/multi_device.ipynb
YigitElma: I think this might cause other problems. When you import `mpi4py`, I think it creates the communications, and in some future cases, we might not want that.

Using multiple devices will be definitely a lot trickier than normal version, so I assume the user has proper `mpi4py` installation.

We can consider adding that during hackathon though

docs/notebooks/tutorials/mpi-tutorials/mpi-eq-solve.py
YigitElma: I added an error for this case.

YigitElma:
Annoying thing about `jax.distributed.initialize()`:

It allows you to see other devices that are not local to the process. However, for some reason, makes the default device, the CPU of the rank 0 process. So, rank 0 is fine, but other processes are actually using a remote CPU that is not local to them. This throws an error `jaxlib.xla_extension.XlaRuntimeError: INVALID_ARGUMENT: Cannot copy array to non-addressable device: BasicDeviceList([CpuDevice(id=0)])` which is triggered by `jnp.array` and `jnp.asarray` operations. Unless someone is willing to add the device flag to each of these functions (and make sure that future codes also have that flag), I would recommend not starting distributed initialize.

YigitElma:
@rahulgaur104 I made some updates. Here is the lean-up output I get for 4 CPUs equilibrium solve (you need to comment out `_set_cpu_count()` line)

```bash
====== TOTAL OF 4 RANKS ======
Rank 0 can see [CpuDevice(id=0)]

====== BACKEND INFO ======
DESC version=0.14.2+144.ge93d0c937.dirty.
Using JAX backend: jax version=0.6.0, jaxlib version=0.6.0, dtype=float64.
Using 4 CPUs with 1376.06 GB total available memory:
	 CPU : 0  AMD EPYC 9654 96-Core Processor
	 CPU : 1  Intel(R) Xeon(R) Gold 6242R CPU @ 3.10GHz
	 CPU : 2  Intel(R) Xeon(R) Gold 6242R CPU @ 3.10GHz
	 CPU : 3  Intel(R) Xeon(R) Gold 6242R CPU @ 3.10GHz

Note: The backend information assumes that the user has 1 process per CPU (node). Using multiple processes per CPU (node) is not the most efficient way to use MPI with purely CPUs.

------------------------------------------------------------
Rank 0 will run objective(s): ['ForceBalance']
Rank 1 will run objective(s): ['ForceBalance']
Rank 2 will run objective(s): ['ForceBalance']
Rank 3 will run objective(s): ['ForceBalance']
------------------------------------------------------------
Timer: Objective build = 1.07 sec
Building objective: force
Timer: LinearConstraintProjection build = 3.86 sec
Number of parameters: 551
Number of objectives: 8424
Timer: Initializing the optimization = 4.97 sec

Starting optimization
Using method: lsq-exact
       0              1          2.500e+00                                    1.228e+00   
       1              2          8.226e-01      1.678e+00      2.256e-01      5.198e-01   
       2              3          2.473e-02      7.978e-01      1.993e-01      7.096e-02   
       3              5          3.451e-03      2.128e-02      8.908e-02      3.526e-02   
Warning: Maximum number of iterations has been exceeded.
         Current function value: 3.451e-03
         Total delta_x: 2.505e-01
         Iterations: 3
         Function evaluations: 5
         Jacobian evaluations: 4
Timer: Solution time = 17.8 sec
Timer: Avg time per step = 4.46 sec
================================================================================================
                                                                 Start  -->   End
Total (sum of squares):                                      2.500e+00  -->   3.451e-03, 
Maximum absolute Force error:                                6.794e+04  -->   6.655e+03 (N)
Minimum absolute Force error:                                1.059e-10  -->   1.232e-10 (N)
Average absolute Force error:                                1.304e+04  -->   1.696e+03 (N)
Maximum absolute Force error:                                5.464e-03  -->   5.352e-04 (normalized)
Minimum absolute Force error:                                8.514e-18  -->   9.912e-18 (normalized)
Average absolute Force error:                                1.049e-03  -->   1.364e-04 (normalized)
Maximum absolute Force error:                                2.032e+05  -->   1.704e+04 (N)
Minimum absolute Force error:                                1.502e-10  -->   1.535e-10 (N)
Average absolute Force error:                                4.494e+04  -->   2.436e+03 (N)
Maximum absolute Force error:                                1.634e-02  -->   1.370e-03 (normalized)
Minimum absolute Force error:                                1.208e-17  -->   1.235e-17 (normalized)
Average absolute Force error:                                3.614e-03  -->   1.959e-04 (normalized)
Maximum absolute Force error:                                6.807e+05  -->   2.238e+04 (N)
Minimum absolute Force error:                                8.896e-11  -->   6.677e-11 (N)
Average absolute Force error:                                7.775e+04  -->   2.548e+03 (N)
Maximum absolute Force error:                                5.474e-02  -->   1.800e-03 (normalized)
Minimum absolute Force error:                                7.155e-18  -->   5.370e-18 (normalized)
Average absolute Force error:                                6.253e-03  -->   2.049e-04 (normalized)
Maximum absolute Force error:                                1.149e+07  -->   1.001e+06 (N)
Minimum absolute Force error:                                3.304e-12  -->   6.674e-13 (N)
Average absolute Force error:                                1.493e+05  -->   6.575e+03 (N)
Maximum absolute Force error:                                9.238e-01  -->   8.054e-02 (normalized)
Minimum absolute Force error:                                2.657e-19  -->   5.368e-20 (normalized)
Average absolute Force error:                                1.200e-02  -->   5.288e-04 (normalized)
R boundary error:                                            0.000e+00  -->   0.000e+00 (m)
Z boundary error:                                            0.000e+00  -->   0.000e+00 (m)
Fixed Psi error:                                             0.000e+00  -->   0.000e+00 (Wb)
Fixed pressure profile error:                                0.000e+00  -->   0.000e+00 (Pa)
Fixed iota profile error:                                    0.000e+00  -->   0.000e+00 (dimensionless)
Fixed sheet current error:                                   0.000e+00  -->   0.000e+00 (~)
================================================================================================

Rank 3 STOPPING
Rank 1 STOPPING
Rank 2 STOPPING
```

YigitElma:
This is for 2 CPUs,

```bash
====== TOTAL OF 2 RANKS ======
Rank 0 can see [CpuDevice(id=0)]

====== BACKEND INFO ======
DESC version=0.14.2+144.ge93d0c937.dirty.
Using JAX backend: jax version=0.6.0, jaxlib version=0.6.0, dtype=float64.
Using 2 CPUs with 1378.87 GB total available memory:
	 CPU : 0  AMD EPYC 9654 96-Core Processor
	 CPU : 1  Intel(R) Xeon(R) Gold 6242R CPU @ 3.10GHz

Note: The backend information assumes that the user has 1 process per CPU (node). Using multiple processes per CPU (node) is not the most efficient way to use MPI with purely CPUs.

------------------------------------------------------------
Rank 0 will run objective(s): ['ForceBalance']
Rank 1 will run objective(s): ['ForceBalance']
------------------------------------------------------------
Timer: Objective build = 1.04 sec
Timer: LinearConstraintProjection build = 3.70 sec
Number of parameters: 551
Number of objectives: 8424
Timer: Initializing the optimization = 4.78 sec

Starting optimization
Using method: lsq-exact
       0              1          8.248e-01                                    6.170e-01   
       1              2          3.863e-01      4.385e-01      4.568e-01      2.795e-01   
       2              3          1.176e-01      2.687e-01      2.927e-01      1.966e-01   
       3              4          8.549e-02      3.208e-02      3.424e-01      1.955e-01   
Warning: Maximum number of iterations has been exceeded.
         Current function value: 8.549e-02
         Total delta_x: 5.986e-01
         Iterations: 3
         Function evaluations: 4
         Jacobian evaluations: 4
Timer: Solution time = 17.9 sec
Timer: Avg time per step = 4.48 sec
================================================================================================
                                                                 Start  -->   End
Total (sum of squares):                                      8.248e-01  -->   8.549e-02, 
Maximum absolute Force error:                                2.032e+05  -->   7.521e+04 (N)
Minimum absolute Force error:                                1.059e-10  -->   1.375e-10 (N)
Average absolute Force error:                                4.102e+04  -->   1.600e+04 (N)
Maximum absolute Force error:                                1.634e-02  -->   6.049e-03 (normalized)
Minimum absolute Force error:                                8.514e-18  -->   1.106e-17 (normalized)
Average absolute Force error:                                3.299e-03  -->   1.287e-03 (normalized)
Maximum absolute Force error:                                1.149e+07  -->   1.580e+06 (N)
Minimum absolute Force error:                                3.304e-12  -->   1.282e-12 (N)
Average absolute Force error:                                1.029e+05  -->   2.220e+04 (N)
Maximum absolute Force error:                                9.238e-01  -->   1.270e-01 (normalized)
Minimum absolute Force error:                                2.657e-19  -->   1.031e-19 (normalized)
Average absolute Force error:                                8.279e-03  -->   1.785e-03 (normalized)
R boundary error:                                            0.000e+00  -->   0.000e+00 (m)
Z boundary error:                                            0.000e+00  -->   0.000e+00 (m)
Fixed Psi error:                                             0.000e+00  -->   0.000e+00 (Wb)
Fixed pressure profile error:                                0.000e+00  -->   0.000e+00 (Pa)
Fixed iota profile error:                                    0.000e+00  -->   0.000e+00 (dimensionless)
Fixed sheet current error:                                   0.000e+00  -->   0.000e+00 (~)
================================================================================================
Rank 1 STOPPING

```

dpanici:
Note that when running on cluster using a command like `srun -n 2 python mpi-eq-solve.py`, the `-n X` part seems like it must match what you requested (and what your number of devices is in the code) in order to avoid errors/hanging code (in my brief experience).

Perhaps we could use some env flags to avoid having to include this in the script, like using `import os; num_devices = os.environ["SLURM_CPUS_PER_TASK"]`?

YigitElma:
> Note that when running on cluster using a command like `srun -n 2 python mpi-eq-solve.py`, the `-n X` part seems like it must match what you requested (and what your number of devices is in the script itself) in order to avoid errors/hanging code (in my brief experience).
> 
> Perhaps we could use some env flags to avoid having to include this in the script, like using `import os; num_devices = os.environ["SLURM_CPUS_PER_TASK"]`?
> 
> EDIT: I think this actually was due to slow compile times so I just thought it was hanging. I still think the second point is important

Yes, the script and `srun -n X ... ` must match but you can use less than or equal to number of processes you previously requested during slurm file. For example, requesting 4 and using only 2 (with proper script) works fine. 

Other combinations, like not using all the requested ranks (supplied by srun) will throw an error during the ObjectiFunction phase, saying that you should put at least one objective to all ranks.

dpanici:
- utility to split up a single objective and single grid into multiple obj and multiple grids for parallel distribution

YigitElma:
Some profiles using JAX 0.6.2 (MPI without CUDA) for `eq-solve` example 2 processes
image
Close up on `compute`,
image
Close up on `jacobian`,
image


Maybe the MPI communication overhead can be reduced by CUDA-aware MPI (we didn't spend much time on that during hackathon). But still, there will be some overhead anyway due to shear array sizes (i.e. say portion of the Jacobian is 1GB, then it will take 100ms with 10GB/s NVLink connection at best, tho I am not sure about the exact speeds))

dpanici:
@dpanici find use case where it speeds things up (auglag to avoid eq solves)

dpanici:
Dependency stuff: add MPI as optional dependency (in pip too), would also be good to point to instructions for openMPI maybe
make it so DESC works fine without MPI installed.

desc/objectives/objective_funs.py
```diff
@@ -1103,6 +1476,9 @@ class _Objective(IOAble, ABC):
         "_scalar",
         "_units",
     ]
+    # _device is of type 'jaxlib.xla_extension.Device' which cannot be an argument
+    # to a jitted function.
+    _static_attrs = ["_device"]
```
YigitElma:
```suggestion
        "_scalar",
        "_units",
        "_device",
    ]
```

desc/__init__.py
```diff
@@ -58,36 +61,126 @@ def __getattr__(name):
...
-        config["device"] = "CPU"
-        config["avail_mem"] = cpu_mem
+        if num_device == 1:
+            config["devices"] = [f"{cpu_info} CPU"]
+            config["avail_mems"] = [cpu_mem]
+        else:
+            try:
+                if mpi is None:
+                    warnings.warn(
+                        "To get the fu list of CPUs, provide the MPI communicator.",
```
YigitElma:
```suggestion
                        "To get the full list of CPUs, provide the MPI communicator.",
```

desc/__init__.py
```diff
@@ -58,36 +61,126 @@ def __getattr__(name):
...
+                    )
+                    cpu_names = [
+                        f"{str(i) + ' ' + cpu_info}" for i in range(num_device)
+                    ]
+                else:
+                    comm = mpi.COMM_WORLD
+                    rank = comm.Get_rank()
+                    cpu_name = f"{str(rank) + ' ' + cpu_info}"
+                    cpu_names = comm.allgather(cpu_name)
+                # These CPUs might not be the same model, but I think slurm will
```
YigitElma: Remove outdated comment

desc/objectives/objective_funs.py
```diff
@@ -283,6 +302,180 @@ def __init__(
...
+            message = ("STOP", None, None)
+            self.comm.bcast(message, root=0)
+        self.running = False
+
+    def _worker_loop(self):
+        """Worker loop for MPI parallelization.
+
+        This function is called when the ObjectiveFunction is used as a context manager.
+
+            with ObjectiveFunction(...) as obj:
```
YigitElma: This is a bit misleading since the syntax has an unbuilt objective function. Fix it

desc/objectives/objective_funs.py
```diff
@@ -314,30 +507,52 @@ def _unjit(self):
...
+    def build(self, use_jit=None, verbose=1):  # noqa:  C901
         """Build the objective.
 
         Parameters
         ----------
         use_jit : bool, optional
-            Whether to just-in-time compile the objective and derivatives.
+            Whether to just-in-time compile the objective and derivatives. If using
+            multiple GPUs, instead of jitting the ObjectiveFunction, the sub-objectives
+            will be jitted individually, independent of the value of `use_jit`.
```
YigitElma: It is not independent of use_jit. If use jit False, nothing will be jitted.

Consider to add use jit wrapper logic to master for selectively jitting objectives. Currently, jit is all or none, might be good to jit some of them, need to use blocked mode though.

desc/objectives/objective_funs.py
```diff
@@ -314,30 +507,52 @@ def _unjit(self):
...
                 objective.build(use_jit=self.use_jit, verbose=verbose)
             self._dim_f += objective.dim_f
+            if objective._device_id != 0:
+                if verbose > 0 and self.rank == 0:
+                    print(
+                        f"Putting objective {objective.name} on device "
+                        f"{objective._device_id}"
+                    )
+                objective = jax.device_put(objective, objective._device)
+                objective._things = obj_things
```
YigitElma, line 799: Explain this with a comment. 

We do this to prevent different copies of the same thing in different devices. This is necessary for finding proper unique things.

desc/objectives/objective_funs.py
```diff
@@ -475,12 +720,21 @@ def compute_unscaled(self, x, constants=None):
         if constants is None:
             constants = self.constants
         assert len(params) == len(constants) == len(self.objectives)
-        f = jnp.concatenate(
-            [
-                obj.compute_unscaled(*par, constants=const)
-                for par, obj, const in zip(params, self.objectives, constants)
-            ]
-        )
+        if not self._is_mpi:
```
YigitElma: We can even have a single _compute_op function that takes op and couple other stuff and do the computation which is kind of repeat now

desc/objectives/objective_funs.py
```diff
@@ -504,12 +758,17 @@ def compute_scaled(self, x, constants=None):
         if constants is None:
             constants = self.constants
         assert len(params) == len(constants) == len(self.objectives)
-        f = jnp.concatenate(
-            [
-                obj.compute_scaled(*par, constants=const)
-                for par, obj, const in zip(params, self.objectives, constants)
-            ]
-        )
+        if not self._is_mpi:
```
YigitElma: See above

desc/objectives/objective_funs.py
```diff
@@ -533,12 +792,21 @@ def compute_scaled_error(self, x, constants=None):
         if constants is None:
             constants = self.constants
         assert len(params) == len(constants) == len(self.objectives)
-        f = jnp.concatenate(
-            [
-                obj.compute_scaled_error(*par, constants=const)
-                for par, obj, const in zip(params, self.objectives, constants)
-            ]
-        )
+        if not self._is_mpi:
```
YigitElma: See above

desc/objectives/objective_funs.py
```diff
@@ -695,18 +969,28 @@ def grad(self, x, constants=None):
...
-        return jnp.atleast_1d(
-            Derivative(self.compute_scalar, mode="grad")(x, constants).squeeze()
-        )
+        if not self._is_mpi:
+            return jnp.atleast_1d(
+                Derivative(self.compute_scalar, mode="grad")(x, constants).squeeze()
+            )
+        else:
+            raise NotImplementedError(
+                "Gradient computation is not implemented for MPI ObjectiveFunction."
```
YigitElma, line 1338: This is probably easy to implement via f.T@Jacobian where both f and Jacobian can be computed in parallel

desc/objectives/objective_funs.py
```diff
@@ -1375,6 +1701,8 @@ def jac_unscaled(self, *args, **kwargs):
         )(*args, **kwargs)
 
     def _jvp(self, v, x, constants=None, op="scaled"):
+        if constants is None:
```
YigitElma: Need this in master too

desc/optimize/_constraint_wrappers.py
```diff
@@ -1198,23 +1199,34 @@ def _jvp(self, v, x, constants=None, op="scaled_error"):
...
+            tangents = batched_vectorize(
+                jvpfun,
+                signature="(n)->(k)",
+                chunk_size=self._constraint._jac_chunk_size,
+            )(v)
+        else:
+            # TODO: implement parallel constraint for ProximalProjection
+            raise NotImplementedError(
+                "Parallel constraint for ProximalProjection not implemented yet. Note: "
+                "This would require putting workers into a second infinite loop which "
```
YigitElma: Leave the Note: part as comment, user doesnt need to know all of this

docs/notebooks/tutorials/mpi-tutorials/mpi-eq-solve.py
```diff
@@ -0,0 +1,116 @@
...
+    obj.build()
+    cons = get_fixed_boundary_constraints(eq)
+
+    # Until this line, the code is performed on all ranks, so it might print some
+    # information multiple times. The following part will only be performed on the
+    # master rank
+
+    # this context manager will put the workers in a loop to listen to the master
+    # to compute the objective function and its derivatives
+    with obj as obj:
```
YigitElma: Do we need 'as obj' part?

docs/notebooks/tutorials/mpi-tutorials/mpi-proximal.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    optimizer = Optimizer("proximal-lsq-exact")
+
+    # Until this line, the code is performed on all ranks, so it might print some
+    # information multiple times. The following part will only be performed on the
+    # master rank
+
+    # this context manager will put the workers in a loop to listen to the master
+    # to compute the objective function and its derivatives
+    with objective as objective:
```
YigitElma: Same as above

docs/notebooks/tutorials/multi_device.ipynb
```diff
@@ -0,0 +1,793 @@
...
+   "source": [
+    "import sys\n",
+    "import os\n",
+    "\n",
+    "sys.path.insert(0, os.path.abspath(\".\"))\n",
+    "sys.path.append(os.path.abspath(\"../../../\"))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
```
YigitElma: Is there a better way to take the content of files here and display as code? It is very tedious to keep them in sync

docs/notebooks/tutorials/multi_device.ipynb
```diff
@@ -0,0 +1,793 @@
...
+   "metadata": {},
+   "source": [
+    "## Using other Objectives\n",
+    "Above we used the convenience function for force balance objective, but we can also use other objectives with this approach."
+   ]
+  },
+  {
+   "cell_type": "markdown",
```
YigitElma: Same as above

desc/objectives/objective_funs.py
```diff
@@ -283,6 +302,180 @@ def __init__(
...
+            # for multiple node cases, each process sees 1 CPU, for those cases,
+            # we cannot put objectives to different devices. Instead, we will
+            # run each objective on the given rank.
+            self._is_mpi = True
+            self._rank_per_objective = (
+                rank_per_objective
+                if rank_per_objective is not None
+                else np.arange(len(objectives))
+            )
+            errorif(
```
YigitElma, line 438: One can probably find more efficient way to validate inputs, I kind of added more errors as I came up with them. Appreciate any feedback!

YigitElma:
I am about to make some changes that might break some code!

The main goal is to use capitalized versions of `bcast` and `gather` to reduce overhead. I have read that `bcast` and `gather` (lower case versions) are meant for general Python objects, so they use pickling. Capitalized versions are told to be more efficient.

desc/backend.py
```diff
@@ -521,6 +554,69 @@ def bodyfun(state):
...
+                and desc_config["kind"] == "gpu"
+            ):
+                warnings.warn(
+                    "The total size of the arrays exceeds the available memory of the "
+                    "GPU[id=0]. Moving the array to CPU. This may cause performance "
+                    "degredation. To suppress this warning, use \n"
+                    "`from desc import config as desc_config` \n"
+                    "`desc_config['SUPPRESS_GPU_MEMORY_WARNING'] = True`"
+                )
+            return jnp.asarray(arr, device=jax.devices("cpu")[0])
```
YigitElma, line 734: Thinking about this again, I don't think this is enough to conduct  trust region operations on CPU. I probably need to add `@execute_on_cpu` to those functions, optionally. So, things will get slightly messy.

YigitElma: Based on the [JAX discussion](https://github.com/jax-ml/jax/discussions/34942) and some testing, as long as we don't commit other arrays to a device, all the computations should follow the device of jacobian (because I explicitly commit it to CPU). This shouldn't be a issue

desc/optimize/_constraint_wrappers.py
```diff
@@ -1067,15 +1075,27 @@ def grad(self, x, constants=None):
...
+            g_vjp = self._objective.vjp_scaled_error(g, xg, constants[0])
+            return tangents @ g_vjp
+        elif self._constraint._is_mpi:
+            # TODO: implement parallel constraint for ProximalProjection
+            raise NotImplementedError(
+                "Parallel constraint for ProximalProjection not implemented yet. "
+                "Please use only one Equilibrium constraint."
+            )
+        else:
+            # TODO: apply vjp for multidevice similar to #2030
```
YigitElma: I am not sure if this makes sense. VJP computes the jacobian row by row, so the objectives that are on different devices must be computed together, the code logic would be convoluted probably.

desc/__init__.py
```diff
@@ -58,37 +61,134 @@ def __getattr__(name):
...
+                else:
+                    comm = mpi.COMM_WORLD
+                    rank = comm.Get_rank()
+                    cpu_name = f"{str(rank) + ' ' + cpu_info}"
+                    cpu_names = comm.allgather(cpu_name)
+                config["devices"] = [name for name in cpu_names]
+                # This memory is not individual but the total memory
+                config["avail_mems"] = [cpu_mem for _ in range(num_device)]
+            except ModuleNotFoundError:
+                raise ValueError(
```
YigitElma: I don't think this is necessary

ddudt:
> @rahulgaur104 @ddudt I could make this work on NERSC Perlmutter with the following steps,

I tested this today and it worked. Thanks!

desc/objectives/_coils.py
```diff
@@ -1391,6 +1407,7 @@ def __init__(
         use_softmin=False,
         softmin_alpha=1.0,
         dist_chunk_size=None,
+        device_id=0,
```
ddudt, line 1438: I noticed `device_id` was missing as an argument for `PlasmaCoilSetMinDistance` so I added it here. We should test or check that this works for every objective in case it is missing anywhere else.

ddudt (changes requested):

desc/objectives/_coils.py
YigitElma: Oh thanks, I added these to all objectives a year ago, so newer objectives might be missing. For general stuff like chunk size, device id, target etc, we might actually pass them as kwargs, but a low priority

desc/objectives/utils.py
YigitElma: all sharding options can be tested on a new pr

desc/__init__.py
```diff
@@ -58,37 +61,134 @@ def __getattr__(name):
...
+    xla_flags = os.getenv("XLA_FLAGS", "")
+    xla_flags = re.sub(
+        r"--xla_force_host_platform_device_count=\S+", "", xla_flags
+    ).split()
+    os.environ["XLA_FLAGS"] = " ".join(
+        [f"--xla_force_host_platform_device_count={n}"] + xla_flags
+    )
+
+
+def set_device(kind="cpu", gpuid=None, num_device=1, mpi=None):  # noqa: C901
```
ddudt, line 128: [todo(blocking)](https://pullpo.io/cc?l=todo&d=blocking) 
`mpi` argument is not documented. It looks like it is only used if `kind="cpu"` and not for the GPU?

.github/workflows/notebook_tests.yml
ddudt, line 1: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
Is this a necessary change to increase the indentation? It looks like the only real change is to add the new notebook

docs/notebooks/tutorials/mpi-tutorials/mpi-eq-solve.py
```diff
@@ -0,0 +1,116 @@
...
+
+import numpy as np
+from mpi4py import MPI
+
+from desc import _set_cpu_count, set_device
+
+kind = "cpu"  # or "gpu"
+num_device = 2
+# ====== Using CPUs ======
+# These will be used for diving the single CPU into multiple virtual CPUs
```
ddudt: diving -> dividing

desc/objectives/objective_funs.py
```diff
@@ -348,6 +358,12 @@ class ObjectiveFunction(IOAble):
         accurately estimate the available device memory, so the "auto" chunk_size
         option will yield a larger chunk size than may be needed. It is recommended
         to manually choose a chunk_size if an OOM error is experienced in this case.
+    mpi : MPI object, optional
+        MPI communicator. Required when using multiple devices.
+    rank_per_objective : ndarray, optional
```
ddudt: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
Could we allow this to also be a list? We could cast the user input to an array if they give a list/tuple.

docs/notebooks/tutorials/multi_device.ipynb
ddudt, line 1: [suggestion](https://pullpo.io/cc?l=suggestion) 
Note about the last part of this notebook, where you show an example of running 6 objectives across 2 nodes with 3 GPUs per node: 

> By default, the initializer will assign different ranks for each sub-objective.

It is not obvious from the example how it specifies which objectives are assigned to node 1 vs node 2. This is done with `rank_per_objective`, assuming the default value of `np.arange(len(objectives))`. I recommend showing this `rank_per_objective` input in the example, to clarify that is how it is using both nodes.

desc/objectives/objective_funs.py
```diff
@@ -404,6 +423,211 @@ def __init__(
...
+            errorif(
+                (
+                    np.mod(self._rank_per_objective, max(device_ids) + 1) != device_ids
+                ).any(),
+                ValueError,
+                "Same rank objectives should also have the same device id. Supplied "
+                f"ranks {self._rank_per_objective} and device ids {device_ids} are "
+                "not compatible.",
+            )
+            # TODO: should this throw an Error?
```
ddudt: [thought](https://pullpo.io/cc?l=thought) 
I think this is fine as a warning, as long as it does not break the code and it will still run. Some nodes have 8 GPUs and maybe the user only has a few different objectives, so it is inefficient but we should allow it.

tests/test_multidevice.py
```diff
@@ -0,0 +1,499 @@
...
+            f0 = objective.compute_scalar(objective.x(eq))
+            eq.optimize(
+                objective=objective,
+                constraints=constraints,
+                optimizer=optimizer,
+                maxiter=1,
+                verbose=3,
+            )
+            f1 = objective.compute_scalar(objective.x(eq))
+            assert f1 < f0
```
ddudt, line 620: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
This test is only checking that the optimization runs and decreases the cost. We could also run the same optimization on a single device and check that both optimizations give the same result. It would take longer to run but be a more robust test.

ddudt (commented):
[praise](https://pullpo.io/cc?l=praise) 
I've tested this PR on Perlmutter with a single node and it works very well! I have a few minor comments, otherwise I will approve. I will also try to test with multiple nodes.

.github/workflows/notebook_tests.yml
YigitElma: I don't think so. I just remember adding the notebook, maybe black formatted it later?

tests/test_multidevice.py
YigitElma: I can write the same optimization with a single device, again in blocked mode, and run it before this. Splitting the objectives into multiple sub-objectives is necessary in the single device case because otherwise, the ordering of the grid nodes differs and causes different numerics. That is why I was using this strategy for the other tests above.

ddudt:
I tested this on multiple nodes and it works as expected. It wasn't very slow with 2 nodes on Perlmutter.

desc/__init__.py
YigitElma: Yes, it is only used for CPU backend. I updated the docstring. It is actually not too useful, but for users without GPU access can give better information. Supplying `mpi` can only change the output of `print_backend_info()`, otherwise it doesn't affect resource usage.

desc/objectives/objective_funs.py
```diff
@@ -1565,6 +1951,8 @@ def jac_unscaled(self, *args, **kwargs):
         )(*args, **kwargs)
 
     def _jvp(self, v, x, constants=None, op="scaled"):
+        if constants is None:
+            constants = self.constants
```
YigitElma:
```suggestion
            constants = self._get_deprecated_constants(constants)
```

desc/objectives/objective_funs.py
```diff
@@ -2002,3 +2390,48 @@ def __call__(self, things):
...
+            # matmul manually. This is slightly inefficient, but usually with rev mode
+            # dim_f << dim_x so it is not too bad.
+            Ji = getattr(obj, "jac_" + op)(*x[idx])
+            J_rank.append(
+                jnp.array([Jii @ vii.T for Jii, vii in zip(Ji, v[idx])]).sum(axis=0)
+            )
+        else:
+            J_rank.append(
+                getattr(obj, "jvp_" + op)(
+                    [_vi for _vi in v[idx]], x[idx], constants=obj.constants
```
YigitElma:
```suggestion
                    [_vi for _vi in v[idx]], x[idx], constants=None
```

desc/objectives/objective_funs.py
```diff
@@ -2002,3 +2390,48 @@ def __call__(self, things):
...
+        ]
+    )
+
+
+@functools.partial(jit, static_argnames="op")
+def jvp_per_process(x, v, objectives, op):
+    """Compute the Jacobian-vector product on each process."""
+    return jnp.hstack(
+        [
+            getattr(obj, op)(v[idx], x[idx], constants=obj.constants)
```
YigitElma:
```suggestion
            getattr(obj, op)(v[idx], x[idx], constants=None)
```

desc/objectives/objective_funs.py
```diff
@@ -465,20 +469,27 @@ def __init__(
...
-            errorif(
-                max(self._rank_per_objective) + 1 != self.size,
-                ValueError,
+            msg = (
                 "The maximum value of rank_per_objective "
                 f"({max(self._rank_per_objective)+1}, supplied as "
-                f"({max(self._rank_per_objective)} in the array) "
-                f"is not equal to the number of ranks ({self.size}). There "
-                "should be at least 1 objective per rank.",
+                f"{max(self._rank_per_objective)} in the array) "
```
ddudt: [question](https://pullpo.io/cc?l=question) 
Why is the max value one more than the actual value? Say we have two ranks, then `rank_per_objective` should be `[0, 1]` and the max value is 1. Why would we print that the max value is 2?

docs/notebooks/tutorials/multi_device.ipynb
```diff
@@ -677,9 +704,49 @@
     "\n",
     "When you write your script for multiple nodes, the number of devices and the device IDs must be selected as if there is only 1 node and only the local GPUs are visible. Other nodes will be used through `rank` of MPI communicator.\n",
     "\n",
-    "**Note: Most clusters have multiple GPUs connected to each node, so before using multiple nodes, use all the GPUs available to that node. Multi-node communication is significantly slower and your script will be easier to write properly.**\n",
+    "**Note: Most clusters have multiple GPUs connected to each node, so before using multiple nodes, use all the GPUs available to that node. Multi-node communication is slower and your script will be easier to write properly.**\n",
+    "\n",
+    "Note: You should have at least 6 objectives, so at least 1 objective per device. If you want to run multiple objectives on the same device, you can specify the ``rank_per_objective`` in the `ObjectiveFunction` keywords. By default, the initializer will assign different ranks for each sub-objective, in above example it defaults to `np.arange(len(objs))`. An example optimization setup where each rank has 1 or more objectives,\n",
```
ddudt: [question](https://pullpo.io/cc?l=question) 
Don't we allow the using fewer ranks than the total available? For example, if you have a node with 4 GPUs I thought you could run an optimization with only 3 objectives using 3 ranks on only 3 of the 4 devices. I thought this would throw a warning but still run correctly, with the extra GPU being unused.

desc/objectives/objective_funs.py
YigitElma: Since `self.size` is 2, printing both `max()` and `max()+1` seemed clearer. For example, let's say we have 3 ranks and `rank_per_objective` is [0, 1, 1], it will print
```
The maximum value of rank_per_objective (2, supplied as 1 in the 
array) is not equal to the number of ranks (3).
```
Your example won't trigger any errors.

YigitElma: I am open to cleaner wording suggestions.

docs/notebooks/tutorials/multi_device.ipynb
YigitElma: Yeah, you are correct, this is only a warning. I updated the explanation. The reason for 6 objectives is the number of ranks (above I say there will be 6 ranks) not number of devices

desc/objectives/objective_funs.py
YigitElma: I updated most of the error/warning messages

ddudt (dismissed):

docs/notebooks/tutorials/multi_device.ipynb
YigitElma, line 1: add a note that GPU 0 is filled with null-space and all other matrices, and does all the heavy linear algebra.

YigitElma:
If you have been using this branch, I made a recent change. Now, instead of giving rank_per_objective to ObjectiveFunction, you should give it to each sub-objective. Default behavior of not having any rank is the same.

YigitElma:
I can finalize this PR for parallel constraints in Proximal too after #2239

desc/__init__.py
YigitElma, line 1: #2324 will probably simplify this part

PR #1496, opened 2024-12-26
Implement Symbolic Recursive Integration for Dommaschk

### LINKED ISSUE #1278: Make Dommaschk factorials more accurate
n! / m!

n! = gamma(n+1)
gammaln(n+1) = log(n!)
exp(gammaln(n+1) - gammaln(m+1)) = n!/m!

### LINKED ISSUE #1467: Dommaschk Potential Issues
It seems that for certain Dommaschk potentials, the results do not align with past publications.

Examples are in [this paper](https://iopscience.iop.org/article/10.1088/0029-5515/24/6/013/meta), HELIOTRON-E apparently gives some surfaces but with helical excursion which is incorrect, and the other two give no flux surfaces at all.

Things to check 
- [x] Implement the HELIOTRON-E myself and check - seems to have a helical axis when it should be planar
- [ ] check AD B versus finite difference B for different step sizes (maybe I did not do the gradient right, though I doubt it)
- [ ] Implement optimization with the potentials to see if that helps uncover any issues (if this does not work at all maybe there are some issues numerically with implementation?)
- [ ] look at algorithm closer and compare to how Dommaschk originally implemented them, see if there are parts prone to numerical error (in orig. paper, he mentions that the factorial can cause very large denominators for some terms)
  - [ ] could try implementing with a higher precision library (mpmath or sympy maybe) to see if it is precision issues, I have them in Sympy somewhere already I think
  - [x] implement recursive version of the potentials (currently have nonrecursive implemented)

dpanici:
Resolves #1467 
Resolves #1278 

Changes implementation of Dommaschk potentials to use `sympy` and symbolic integration to recursively compute the potentials with symbolic algebra. I then use the `sympy.lambdify` function to turn this into a jax-differentiable expression usable with our `ScalarPotentialField` class.


desc/magnetic_fields/_dommaschk.py
```diff
@@ -5,7 +5,13 @@
 
 """
 
-from desc.backend import cond, fori_loop, gammaln, jit, jnp
+import time
+
+import sympy as sp
```
dpanici, line 10: import this stuff in backend not here

desc/backend.py
```diff
@@ -5,6 +5,7 @@
 import warnings
 
 import numpy as np
+import sympy as sp  # noqa: F401
```
f0uriest, line 8: i don't think this needs to be in backend, is there any reason we don't just import it normally where its needed?

desc/magnetic_fields/_dommaschk.py
```diff
@@ -158,6 +229,9 @@ def fit_magnetic_field(  # noqa: C901
         # and max_l and max_m should probably be both nonzero anyways,
         # this is not an issue right now
 
+        # TODO: Add REGCOIL-like option to perform a least squares minimization
```
f0uriest, line 232: prefer to make issues for these rather than TODO in source

desc/magnetic_fields/_dommaschk.py
```diff
@@ -176,6 +250,7 @@ def fit_magnetic_field(  # noqa: C901
         ]  # indices that should be 0 due to symmetry
         # if sym is True, when l is even then we need a=d=0
         # and if l is odd then b=c=0
+        # TODO: does the sym hold if the eq is not "centered" about R=1?
```
f0uriest, line 253: prefer to make issues for these rather than TODO in source

desc/magnetic_fields/_dommaschk.py
```diff
@@ -217,18 +292,19 @@ def fit_magnetic_field(  # noqa: C901
             round(num_modes - 1) / 4
         )  # how many l-m mode pairs there are, also is len(a_s)
         n = int(n)
-        domm_field = DommaschkPotentialField(0, 0, 0, 0, 0, 0, 1)
+        t0 = time.time()
```
f0uriest, line 295: use `desc.utils.Timer` for this so all the timing info is done in one place

desc/magnetic_fields/_dommaschk.py
```diff
@@ -248,21 +324,25 @@ def get_B_dom(coords, X, ms, ls):
             else:
                 X += [obj]
         X = jnp.asarray(X)
-
-        jac = jit(Derivative(get_B_dom, argnum=1))(
-            coords, X, params["ms"], params["ls"]
-        )
+        t0 = time.time()
```
f0uriest, line 327: use desc.utils.Timer for this so all the timing info is done in one place

desc/magnetic_fields/_dommaschk.py
```diff
@@ -248,21 +324,25 @@ def get_B_dom(coords, X, ms, ls):
...
         # TODO (#928): use min singular value to give sense of cond number?
-        c, res, _, _ = jnp.linalg.lstsq(A, rhs)
+        c, res, rank, s = jnp.linalg.lstsq(A, rhs)
 
         if verbose > 0:
             # res is a list of len(1) so index into it
-            print(f"Sum of Squares Residual of fit: {res[0]:1.4e} T")
+            print(f"Sum of Squares Residual of fit: {res[0]:1.4e} T^2")
+            res_at_pts = A @ c - rhs
+            print(f"Max Residual: {jnp.max(abs(res_at_pts)):1.4e} T")
```
f0uriest, line 343: max _absolute_ residual, same for min & mean

desc/magnetic_fields/_dommaschk.py
```diff
@@ -478,12 +504,11 @@ def V_m_l(R, phi, Z, m, l, a, b, c, d):
...
     """
-    return (a * jnp.cos(m * phi) + b * jnp.sin(m * phi)) * D_m_n(R, Z, m, l) + (
-        c * jnp.cos(m * phi) + d * jnp.sin(m * phi)
+    return (a * sp.cos(m * phi) + b * sp.sin(m * phi)) * D_m_n(R, Z, m, l) + (
+        c * sp.cos(m * phi) + d * sp.sin(m * phi)
     ) * N_m_n(R, Z, m, l - 1)
 
 
-@jit
 def dommaschk_potential(R, phi, Z, ms, ls, a_arr, b_arr, c_arr, d_arr, B0=1):
```
f0uriest, line 512: the arguments to this are now expected to be sympy variables, not numpy or jax arrays right? Should update the docstring

tests/test_magnetic_fields.py
```diff
@@ -1558,3 +1561,227 @@ def test_domm_field_is_nonzero_and_continuous_across_Z_0():
...
+
+@pytest.mark.mpl_image_compare(remove_text=True, tolerance=10)
+@pytest.mark.unit
+def test_domm_W7AS():
+    """Test can get flux surfaces from W7-AS Dommashk coefficients."""
+    # W7-AS
+    # The coefficients are taken from IPP-Report IPP_0_48 by Dommaschk et al.
+    B0 = 1.00
+    # toroidal harmonics
+    ms = [
```
f0uriest, line 1574: if you want you can turn off black for these sections so it won't put everything on its own line: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#ignoring-sections

f0uriest (changes requested):

desc/backend.py
dpanici: I had moved it just to be in backend but I can move it back. Just thought I remembered someone saying we should import stuff like this in backend.

f0uriest (approved):

dpanici:
Moving notes to separate comment so it is not in PR commit comment

Notes
------
I've found that using sympy and symbolic integration to use the recursive form of the Dommaschk potentials to be a much more accurate way to use them (accurate as in, recreating poincare plots from the IPP report linked [here](https://pure.mpg.de/rest/items/item_2132035_1/component/file_2132034/content), which previously I failed to find any flux surface using the nonrecursive algorithm)

Previous implementation was non-recursive and suffered from issues for certain more complex potentials, which were either due to numerical issues in the algorithm (precision issues) or due to an incorrect implementation on my part (which I never was able to track down). Either way, this symbolic method seems to work much better.

tests/test_magnetic_fields.py
```diff
@@ -1558,3 +1561,95 @@ def test_domm_field_is_nonzero_and_continuous_across_Z_0():
     Bnew = new_field.compute_magnetic_field(np.vstack([Rs, phis, Zs]).T)
 
     assert not np.any(np.isclose(Bnew[:, 1], 0))
+
+
+@pytest.mark.mpl_image_compare(remove_text=True, tolerance=10)
+@pytest.mark.unit
+def test_domm_W7AS():
+    """Test can get flux surfaces from W7-AS Dommashk coefficients."""
```
YigitElma, line 1569:
```suggestion
    """Test getting flux surfaces from W7-AS Dommaschk coefficients."""
```

YigitElma (changes requested):

desc/magnetic_fields/_dommaschk.py
```diff
@@ -248,21 +322,26 @@ def get_B_dom(coords, X, ms, ls):
             else:
                 X += [obj]
         X = jnp.asarray(X)
```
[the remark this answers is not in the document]
dpanici: this is capital X, not lowercase x, but ok yea I can change that it is nbd

f0uriest (approved):

YigitElma (approved):

PR #1425, opened 2024-12-04
Update notebooks with new stopping criteria

f0uriest:
Updates all the notebooks with new stopping criteria and new example equilibria (and misc other changes since they were last ran forever ago). No significant changes, just some small numerical differences in a few places.

- coil optimization tutorial will be redone in #1361 

Resolves #1423 
Resolves #1468

Issue #1215, opened 2024-08-21
Strip kernel info from jupyter notebooks


dpanici:

dpanici:
pre-commit hook to check if a notebook was changed and if so, to strip the kernel info from it

PR #1498, opened 2024-12-27
Minor fixes for docs

YigitElma:


docs/notebooks/tutorials/basic_optimization.ipynb
```diff
@@ -1061,7 +1061,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "desc-env",
```
unalmis, line 1064: did you forgot to switch back to the python kernel

YigitElma: Are we doing that? This happened automatically, I haven't noticed.

dpanici: yea you should, I think, technically this change is fine but if any local user has renamed the kernel differently than `desc-env`, they would not be able to run locally, or if your kernel was not `desc-env`, the notebook would not run on CI

YigitElma: Is there an easy way to do this? I thought we were waiting for #1215. When I change the kernel to my default Python, it shows `"display_name": "Python 3"` not `"display_name": "Python 3 (ipykernel)"`. There are a lot of notebooks with different kernel names, I can also change them but it is really cumbersome for future since this change happens automatically if you just click on the notebook.

YigitElma: I believe the problem with CI was only happening after cluster use (for some reason). If you look at free boundary tutorial etc, they also have `desc-env` as kernel.

ddudt: I use conda environments that are not named `desc-env`

docs/notebooks/tutorials/coil_stage_two_optimization.ipynb
```diff
@@ -818598,7 +818598,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "desc-env-latest",
+   "display_name": "Python 3",
```
YigitElma, line 818601: For example, this was working fine!?

docs/notebooks/tutorials/basic_optimization.ipynb
YigitElma: See comment below for `coil_stage_two_optimization.ipynb`. It previously had `desc-env-latest` but working fine.

docs/notebooks/tutorials/coil_stage_two_optimization.ipynb
dpanici: if this worked fine I guess that's ok... but when you download that notebook, it might not work immediately unless the  local env name is desc-env-latest

dpanici (changes requested):
make requested change in notebook

docs/notebooks/tutorials/coil_stage_two_optimization.ipynb
YigitElma: I don't know about you but when I try to use these kinds of notebooks, VS code automatically directs me to select a kernel. Apart from that one time on CI error, I have never faced an issue regarding different environment name.

dpanici (changes requested):
change VMEC load part again

f0uriest:
Can we do this after #1425

YigitElma:
Note about kernel stuff: The important name for CI tests to pass is the `name`, so as long as the `name` in `kernelspec` is `python3`, it is fine, whatever the `display_name` is.

This problem probably arise from the difference between VS Code's jupyter notebook extension and jupyter lab. VS-CODE doesn't change the `name` only changes `display_name` so it works.

rahulgaur104 (approved):

f0uriest (dismissed):

unalmis (approved):

PR #1504, opened 2024-12-30
`FourierXYCurve` and `FourierXYCoil` classes

ddudt:
Adds a new parameterization for planar curves/coils. Instead of representing $r(\theta)$ as a Fourier series like the `FourierPlanarCurve`, this represents $X(s)$ and $Y(s)$ as Fourier series where $r = \sqrt{X^2 + Y^2}$ and $s$ is an arbitrary curve parameter. 

There are some simple coil shapes like an ellipse that cannot be perfectly represented with the `FourierPlanarCoil` class (at finite resolution). This parameterization can represent more coil geometries, at the cost of more optimization parameters and the possibility of self-intersecting curves. 

- [x] Tests
- [x] Update master_compute_data


dpanici:
Add symmetry to FourierSeries basis class and use here to delete the n=0 mode

desc/geometry/curve.py
```diff
@@ -858,6 +866,341 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
+    Parameterized by a point (the center of the curve), a vector (normal to the plane),
+    and Fourier series defining the X and Y coordinates in the plane.
+
+    Parameters
+    ----------
+    center : array-like, shape(3,)
+        Coordinates of center of curve, in system determined by basis.
+    normal : array-like, shape(3,)
+        Components of normal vector to planar surface, in system determined by basis.
+    X_n : array-like
```
YigitElma, line 881: I guess it doesn't matter if the user gives n=0 modes or not, the shape is the same, right?

desc/geometry/curve.py
```diff
@@ -858,6 +866,341 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
+            self._basis = new
+
+    @property
+    def N(self):
+        """Maximum mode number."""
+        return max(self.X_basis.N, self.Y_basis.N)
+
+    def change_resolution(self, N=None):
+        """Change the maximum angular resolution."""
+        N = check_nonnegint(N, "N")
```
YigitElma, line 1039: Since N cannot be 0, would `check_posint()` be better?

desc/basis.py
```diff
@@ -435,9 +438,10 @@ class FourierSeries(_Basis):
         Maximum toroidal resolution.
     NFP : int
         number of field periods
-    sym : {``'cos'``, ``'sin'``, False}
-        * ``'cos'`` for cos(m*t-n*z) symmetry
-        * ``'sin'`` for sin(m*t-n*z) symmetry
+    sym : {``'cos'``, ``'sin'``, ``'n0'``, False}
+        * ``'cos'`` for cos(n*z) symmetry
+        * ``'sin'`` for sin(n*z) symmetry
+        * ``'n0'`` for no n=0 mode
```
dpanici, line 444: I don't love this name. I agree that having a basis that lacks a zero mode is useful, but can we be more verbose and say like `"no n=0"` or something?

desc/geometry/curve.py
```diff
@@ -293,15 +299,17 @@ def from_values(cls, coords, N=10, NFP=1, sym=False, basis="rpz", name=""):
         )
 
         grid = LinearGrid(zeta=phi, NFP=1, sym=sym)
-        basis = FourierSeries(N=N, NFP=NFP, sym=sym)
-        transform = Transform(grid, basis, build_pinv=True)
-        R_n = transform.fit(R)
-        Z_n = transform.fit(Z)
+        R_basis = FourierSeries(N=N, NFP=NFP, sym="cos" if sym else False)
```
dpanici, line 302: was this incorrect before then?

desc/geometry/curve.py
```diff
@@ -858,6 +866,341 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
+    center : array-like, shape(3,)
+        Coordinates of center of curve, in system determined by basis.
+    normal : array-like, shape(3,)
+        Components of normal vector to planar surface, in system determined by basis.
+    X_n : array-like
+        Fourier coefficients of the X coordinate in the plane.
+    Y_n : array-like
+        Fourier coefficients of the Y coordinate in the plane.
+    modes : array-like
+        mode numbers associated with X_n and Y_n.
```
dpanici, line 886: can you say the default assumption is [-X_n.size//2 , X_n.size//2], with the n=0 mode included even tho it will be taken out later?

desc/transform.py
```diff
@@ -204,16 +204,13 @@ def _check_inputs_fft(self, grid, basis):
...
             (basis.modes[:, None, :2] == self.lm_modes[None, :, :]).all(axis=-1)
         )[1]
         col = np.where(
-            basis.modes[None, :, 2] == basis.modes[basis.unique_N_idx, None, 2]
+            basis.modes[None, :, 2] == np.arange(-basis.N, basis.N + 1)[:, None, None]
         )[0]
-        self.fft_index = np.atleast_1d(
-            np.squeeze(self.num_n_modes * row + col + offset)
-        )
+        self.fft_index = np.atleast_1d(np.squeeze(self.num_n_modes * row + col))
```
dpanici, line 213: can you explain this change?

dpanici (changes requested):
I don't really like `n0` as the new basis sym name, can it be more verbose?

left a question or two as well

desc/geometry/curve.py
ddudt: Yes I think it was. `FourierSeries` has no `sym=True` option, so we probably just never tested that case.

desc/geometry/curve.py
ddudt: The user should not be giving n=0 modes, you can't give coefficients that are not in the basis.

ddudt: Should we throw a warning in this case if the user passes in a n=0 mode? Another option is that we could add the n=0 coefficients to the `center` parameter, but then we are overwriting the user inputs. I would rather force the user to give self-consistent inputs.

dpanici (changes requested):
make docstring of new curve class clearer for what it expects when `modes` is not explicitly passed in (no n0 mode is assumed, modes go from from [-X_n.size//2, X_n.size//2] but assuming n=0 is skipped)

desc/geometry/curve.py
ddudt: @YigitElma @dpanici What are your opinions on how to handle the case if the user inputs the n=0 mode? Options:

1. User should not give the n=0 coefficient, will error if they do (present code)
2. Throw a warning if the user does give the n=0 coefficient, saying that it is being ignored (I think this is my vote)
3. Allow the user to give the n=0 coefficient, and add it to the `center` input (more general but also more complicated)

f0uriest: i don't see it throwing an error if `modes` includes n=0 currently? I'd be in favor of this, or warning and dropping it.

YigitElma: I also like to drop the n=0 modes (I think they are already ignored), but additionally to throw a warning/error like
```
Ignoring n=0 modes. Please explicitly change the center argument if you are trying to shift the center.
```

desc/geometry/curve.py
ddudt: FYI the default value for `modes` now handles both cases: 

- If `len(X_n)` is odd, then it assumes the n=0 mode is included like the other curve classes but gets ignored. 
- If `len(X_n)` is even, then it assumes the n=0 mode is not included since it is unused for this class.

dpanici:
I think changes made to `FourierPlanarCurve.from_values` generalizing its curve parmeter actually break the curve fitting, as you allow for an arbitrary curve parameter (not the polar angle) but we assume the polar angle in the compute functions for it (i.e. we say X=r cos(s) and Y = rsin(s) which implies tan(s) = Y/X, which is not used/enforced in the `from_values` method)

```python
import numpy as np
from desc.coils import FourierPlanarCoil, FourierXYZCoil, FourierRZCoil

coil_XYZ = FourierXYZCoil(
    current=1e6, X_n=[0, 10, 1],
    Y_n=[-2, 0, 0],
    Z_n=[0, 0, 0]
)
coil_planar = coil_XYZ.to_FourierPlanar(N=10)
d_xyz = coil_XYZ.compute(["x"],basis="xyz")
d_planar = coil_planar.compute(["x"],basis="xyz")

plt.plot(d_xyz["x"][:,0], d_xyz["x"][:,1], label="xyz")
plt.plot(d_planar["x"][:,0], d_planar["x"][:,1],"--", label="planar fit")
plt.legend()
plt.xlabel("x")
plt.ylabel("y")
plt.axis("equal")
```
image
image (1)

dpanici (changes requested):
undo the changes to `FourierPlanarCurve.from_values` which allow arbitrary curve parameter, as we cannot allow that given the way `"x"` is computed assuming that `tan(s) = Y/X` in the plane which enforces a specific choice of curve parameter

dpanici:
was the update for master compute data only for the `FourierPlanarCurve` stuff?

ddudt:
> was the update for master compute data only for the `FourierPlanarCurve` stuff?

No. This adds new classes and `test_compute_everything` checks the compute quantities for each class. So I had to add the data for those new classes.

desc/coils.py
```diff
@@ -1003,6 +1037,86 @@ def from_values(cls, current, coords, N=10, basis="xyz", name=""):
         )
 
 
+class FourierXYCoil(_Coil, FourierXYCurve):
+    """Coil that lies in a plane.
+
+    Parameterized by a point (the center of the coil), a vector (normal to the plane),
+    and a Fourier series defining the X and Y coordinates in the plane as a function of
+    the polar angle theta.
```
dpanici, line 1045:
```suggestion
    an arbitrary curve parameter ``s``
```

desc/geometry/curve.py
```diff
@@ -895,6 +902,375 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
         )
 
 
+class FourierXYCurve(Curve):
+    """Curve that lies in a plane.
+
+    Parameterized by a point (the center of the curve), a vector (normal to the plane),
+    and Fourier series defining the X and Y coordinates in the plane.
```
dpanici, line 909: mention that the Fourier series is in an arbitrary angle

desc/geometry/curve.py
```diff
@@ -895,6 +902,375 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
+        self._Y_n = copy_coeffs(Y_n, modes, self.Y_basis.modes[:, 2])
+
+        self._basis = basis
+        self.normal = normal
+        self.center = center
+
+    @optimizable_parameter
+    @property
+    def center(self):
+        """Center of planar curve polar coordinates."""
```
dpanici, line 990: they are not polar coordinates right? not for this one

desc/objectives/_coils.py
```diff
@@ -1108,10 +1108,10 @@ class CoilArclengthVariance(_CoilObjective):
 
     This objective is meant to combat any issues corresponding to non-uniqueness of
     the representation of a curve, in that the same physical curve can be represented
-    by different parametrizations by changing the curve parameter [1]_. Note that this
-    objective has no effect for ``FourierRZCoil`` and ``FourierPlanarCoil`` which have
-    a single unique parameterization (the objective will always return 0 for these
-    types).
+    by different parametrizations by changing the curve parameter [1]_.
+    Note that this objective has no effect for ``FourierRZCoil``, ``FourierPlanarCoil``,
+    and ``FourierXYCoil`` which have a single unique parameterization
```
dpanici, line 1113: not correct about the FourierXYCoil

tests/test_coils.py
```diff
@@ -1505,18 +1575,21 @@ def test_initialize_helical():
...
+    """Test that fitting to a planar coils preserves orientation."""
     # easiest to check this is working using a coil, as can use
     # biot savart to see if B is same or not
     # tests fix for GH #1715
     coil_XYZ = FourierXYZCoil(
         current=1e6, X_n=[0, 10, 2], Y_n=[-2, 0, 0], Z_n=[0, 0, 0.1]
     )
-    coil_planar = coil_XYZ.to_FourierPlanar(N=1)
+    coil_planar1 = coil_XYZ.to_FourierPlanar(N=1)
+    coil_planar2 = coil_XYZ.to_FourierXY(N=1)
```
dpanici, line 1587: Since `FourierXY` uses an arbitrary angle, idt this is actually an issue. But am fine with keeping the test

desc/geometry/curve.py
YigitElma, line 990:
```suggestion
        """Center of planar curve in `self.basis` coordinates."""
```

desc/geometry/curve.py
```diff
@@ -895,6 +902,375 @@ def from_values(cls, coords, N=10, basis="xyz", name=""):
...
+        # convert to xyz basis
+        if basis == "rpz":
+            coords = rpz2xyz(coords)
+        coords = np.atleast_2d(coords)
+
+        # center
+        center = np.mean(coords, axis=0)
+        coords = coords - center  # shift to origin
+
+        # normal
```
dpanici, line 1212: Question for this curve class: 

technically it works totally fine if the curve orientation and the normal direction are opposite (in the sense that you can have a curve which is oriented left-handed w.r.t. the normal vector). Unless we include a line like Ryan has in the `from_values` of `FourierPlanarCurve` this will be allowed. This does not break anything as far as I can tell. Just want to note this as could be slightly counterintuitive

desc/geometry/curve.py
YigitElma: Or just delete that part as in `self.normal`s doc.

dpanici (changes requested):
some minor changes requested. Also see note on the normal vector for FourierXY, Idt it matters but should think about for a sec

desc/geometry/curve.py
ddudt: Hmm good point. There is a "right-handed" orientation wrt the normal vector, but it doesn't actually affect any of the computations. The only thing that matters is the physical direction of the current flowing in the coil. I think we should keep it general so that the user can choose the orientation they find the most convenient (e.g., they might want a "left-handed" curve if that allows them to make the current positive instead of negative.)

YigitElma (approved):

dpanici (approved):

YigitElma (approved):

PR #1507, opened 2025-01-06
Add puddle objective to avoid local min in current potential

dpanici:


dpanici:
@dpanici look at Fu paper and replace with different objective

PR #1508, opened 2025-01-06
Poloidal FFT Implementation

### LINKED ISSUE #641: Use FFTs in poloidal direction where possible
Several objectives now use `QuadratureGrid` or `LinearGrid` for their default (to get accurate volume or surface averages), and because they are uniform in the poloidal direction we could possibly use FFTs to do that part of the transform. This would like see a decent speed bump and memory savings, similar to using FFTs for the poloidal direction. This would be added as another option to the `Transform` class.

dpanici:
Resolves #641

### DIFF
--- a/desc/transform.py
+++ b/desc/transform.py
@@ -1,5 +1,6 @@
 """Class to transform from spectral basis to real space."""
 
+import pdb
 import warnings
 
 import numpy as np
@@ -262,6 +263,7 @@ class Transform(IOAble):
             self.method = "direct2"
             return
 
+        pdb.set_trace()
         self._method = "fft"
         self.lm_modes = np.unique(basis.modes[:, :2], axis=0)
         self.num_lm_modes = self.lm_modes.shape[0]  # number of radial/poloidal modes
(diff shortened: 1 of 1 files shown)


dpanici:
may need to change code to move grid checks after #1405

unalmis:
(if you're hooking in ffts for chebyshev double fourier, the order for transform from real to spectral should coeffs = rfft2(dct(f)) since dct returns a real output).

dpanici:
- Add util for converting to `numpy` fft basis
- make `ptolemy` identity work with JAX, also matrix free method?

f0uriest:
for the current ptolemy identity stuff we use some dynamic sizing and `np.where` operations that won't work with jax. I think one possible fix would be to pass in a `Basis` instead of just the `modes` array. Then eg `Basis.N` is statically known, so we can always do
```python
modes_full = jnp.arange(-basis.N, basis.N+1)
x_full = jnp.zeros(2*basis.N+1)
x_full = x_full.at[basis.modes[:,2]].set(x)
```
then we have a statically sized array with all of the mode numbers and coefficients that can more easily be manipulated without conditionals

f0uriest:
To get this to work properly I think we decided the best option would be to have a way to convert our fourier representation to the vmec one, since that can be directly used with 2d ffts. I played around with a few different ways to do this here: https://gist.github.com/f0uriest/b12be9d5f22b845463158793fefba82b

Here's some timing for a `DoubleFourierBasis` with M=N=20

```python
# existing method, np only, for comparison
%timeit _, _, s2, c2 = desc.vmec_utils.ptolemy_identity_rev(basis.modes[:,1], basis.modes[:,2], x[None])
# 297 ms ± 29.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit _, _, s3, c3 = jax.block_until_ready(ptolemy_identity_rev_jax(basis, x))
# 30.2 ms ± 2.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit _, _, s3, c3 = jax.block_until_ready(ptolemy_identity_rev_jax_vmap(basis, x))
# 9.38 ms ± 1.68 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

However, an even better option seems to be to just precompute the linear transform matrix eg

```python
A = jnp.array(desc.vmec_utils.ptolemy_linear_transform(basis.modes, vmec_modes=None, helicity=None, NFP=None)[0])
%timeit _ = jax.block_until_ready(jnp.dot(A, x))
# 1.01 ms ± 18.9 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

And we can do even better if we use a sparse representation for A:
```python
import jax.experimental.sparse
As = jax.experimental.sparse.BCOO.fromdense(A)

@jax.jit
def spdot(A, x):
    return A@x

%timeit _ = jax.block_until_ready(spdot(As, x))
# 32.4 μs ± 1.68 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```


For pretty much all resolutions, just multiplying by A seems to be the fastest option, and for high res using a sparse A seems the best bet.

f0uriest:
So it from some profiling it seems like using FFTs might not even be worth it compared to just using a factorized direct method?
```python
for M in [5, 10, 15, 20, 25, 30, 40, 50, 60, 70, 80, 90, 100]:
    m = 2*M+1
    x = np.random.random((m,m,m))
    A = np.random.random((4*M+1,m))
    print("M=",M)
    print("dot time")
    %timeit _ = jnp.dot(A, x).block_until_ready()
    print("fft time")
    %timeit _ = jax.jit(jnp.fft.ifft, static_argnames=["n", "axis"])(x, n=4*M+1, axis=0).block_until_ready()
```
```
M= 5
dot time
23.7 μs ± 2.83 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
fft time
93.9 μs ± 1.8 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
M= 10
dot time
77.4 μs ± 7.4 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
fft time
433 μs ± 162 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
M= 15
dot time
263 μs ± 12 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
fft time
1.99 ms ± 37.9 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
M= 20
dot time
377 μs ± 12.6 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
fft time
2.14 ms ± 39.9 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
M= 25
dot time
824 μs ± 17 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
fft time
12.2 ms ± 297 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
M= 30
dot time
1.53 ms ± 18 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
fft time
7.17 ms ± 65.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
M= 40
dot time
4.25 ms ± 339 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
fft time
20.6 ms ± 4.39 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
M= 50
dot time
6.65 ms ± 76.1 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
fft time
53.6 ms ± 6.67 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
M= 60
dot time
10.8 ms ± 94.1 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
fft time
87.5 ms ± 4.93 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
M= 70
dot time
18.3 ms ± 244 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
fft time
169 ms ± 12.4 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
M= 80
dot time
34.7 ms ± 716 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
fft time
231 ms ± 15.4 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
M= 90
dot time
82.4 ms ± 622 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
fft time
259 ms ± 47.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
M= 100
dot time
120 ms ± 3.73 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
fft time
417 ms ± 84.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

though I'm a bit suspicious that the FFT doesn't get faster even at crazy high resolutions. Can someone else run this and let me know?

unalmis:
Won't the speed improve a bunch if we do `rfft2(dct(f, radial axis), flux axes)` and the inverse `idct(irfft2)`

f0uriest:
This seems to suggest the opposite. Factorizing the 3d transform into 3 1d transforms is definitely worth it, but doing fft/dct doesn't seem to be any faster than matrix multiplication for the sizes we care about (and much larger sizes as well?)

f0uriest:
Also note that the time cost of converting our double fourier basis to the form expected by `np.fft2` is comparable to the cost of doing the fft. Plus, in the common case that we have poloidal up/down symmetry and we use a symmetric grid, I don't think there's any easy way to do FFT's in that case, other than doing the full domain and then truncating.

All that said, I think it probably makes more sense to just do direct matrix transforms poloidally. We can still get some speedup by separating the radial/poloidal parts, but still using dense matrices for both.

Issue #1294, opened 2024-10-05
Nonuniform fast transforms


unalmis:
In general, the computation of locally defined or singular quantities will be more efficient with non-uniform fast transforms. In particular, I am referring to the type 2 nonuniform fast transform, which evaluates spectral coefficients at non-uniformly spaced points.

Although a fixed grid with nodes uniformly spaced throughout the plasma is fine for optimizing continuous functions defined throughout the volume, it is of course not well-suited for locally defined quantities such as bounce integrals. There are two hurdles with computing these quantities efficiently: 

1. Require computation along field lines.
    - #1119 should cover this, improving upon #854.
3. They are defined by singular integrals.
    - This requires evaluating the function at non-uniform points. For flux surface average of bounce integrals in particular, there are infinitely many integrals to be done, so regardless of the efficiency of the quadrature, there will be many function evaluations. Hence, linearithmic performance of fast non-uniform transform is desired over the quadratic performance of MMTs.

We should pick a JAX compatible NFFT package and write drop in replacements for the currently used [MMTs](https://github.com/PlasmaControl/DESC/pull/1119/files#r1788686172). We can check accuracy of these external NFFT libraries by comparing to those utilities on the tests in that PR. (Note that for strongly singular integrals $1 / \sqrt(1 - \lambda |B|)$, we will maybe still use MMT's for |B| since the integrands are not continuous functions of |B|, and the nodes are packed more tightly near the edges. This should be fine since |B| is parameterized by a series expansion in a single variable along each cut of the field line, so it is cheaper. The other quantities in the integrand are 2D functions where this would be more useful. For the weakly singular ones, all the quantities should be able to use NFFTs without issue).

Here are some candidates:
1. https://github.com/jakevdp/nfft
2. https://github.com/flatironinstitute/jax-finufft

acerfon:
For what it's worth, we have been very happy with the Flatiron's NUFFT, which we have used with much success for a new Particle in Fourier scheme: 
https://www.sciencedirect.com/science/article/pii/S0021999124006387#:~:text=Journal%20of%20Computational%20Physics.%20Volume

I highly recommend that library!

acerfon:
As another 2-cents, for the singular integrals you are talking about, I very highly recommend the Kapur-Rokhlin quadrature schemes, which we have also found easy to use, and super accurate, in our case for the evaluation of the virtual casing principle:
https://www.jstor.org/stable/2952054

f0uriest:
[...]

Issue #1303, opened 2024-10-15
Differentiable code with dynamic shapes


unalmis:
A fundamental limitation of our auto diff tool JAX is it is [unable to handle dynamic size arrays](https://github.com/jax-ml/jax/discussions/18790) or jagged tensor operations. Both of these operations are supported by PyTorch, which I reference to better document what these operations are:
1. [Dynamic shapes](https://pytorch.org/docs/stable/torch.compiler_dynamic_shapes.html)
2. [Jagged tensor operations](https://pytorch.org/FBGEMM/fbgemm_gpu-overview/jagged-tensor-ops/JaggedTensorOps.html) 

So, unlike `c` code, in JAX we can't do `jnp.nonzero(a)[0]`, which constrains the algorithms we write. Sometimes it is possible to work around this by using a shape independent algorithm. For example, performing least squares regression to a variable number of points through equation 12 instead of 13: [least squares example](https://mathworld.wolfram.com/LeastSquaresFittingPolynomial.html). In other cases, it is not possible to write a shape independent algorithm without sacrificing performance.

It might be worthwile to implement a patch in `desc.backend` to work around JAX not being able to do 1. For example, we can make a differentiable `jnp.nonzero(a)[0]` with the following code

```python
import jax
import jax.numpy as jnp
import numpy as np

@jax.jit
def expensive_function(x):
    # pretend this is some non-trivial operation
    return jnp.dot(x, x) * x

grad_expensive_function = jnp.vectorize(jax.grad(expensive_function))

def nonzero_where(x):
    dummy = -1.0
    idx = np.nonzero(x)[0]
    y = jnp.full(x.shape, fill_value=dummy)
    y = y.at[idx].set(expensive_function(x[idx]))
    return y

def forward(x):
    return nonzero_where(x), x

def backward(x, gradient):
    idx = np.nonzero(x)[0]
    grad = jnp.zeros(x.shape)
    grad = grad.at[idx].set(grad_expensive_function(x[idx]) * gradient[idx])
    return (grad,)

fun = jax.custom_vjp(nonzero_where)
fun.defvjp(forward, backward)

x = jnp.array([1.0, 0.0, 3.0])
y = fun(x)

example_optimization_gradient = jax.grad(lambda x: fun(x).sum())
np.testing.assert_allclose(example_optimization_gradient(x), [3, 0, 27])
```

where the intention is to have a function in `desc.backend`  so that we can replace this logic
```python
dummy = -1.0
[...]

Issue #1206, opened 2024-08-20
Utility to upsample data computed above midplane to full volume assuming stellarator symmetry


unalmis:
This would be useful for FFT's or integration over a subset of the volume as discussed in #1101 . Currently algorithms that rely on such computations are forced to compute on the full volume, which impacts performance (memory and speed). Once we have a utility like

```python
def map2fullgrid(grid, data):
   """Return data computed on grid as if it were computed on full grid."""
```
We can just call this inside the compute functions rather than raising warnings that computations are incorrect for quantities that compute integrals over subsets of the volume in #1094 or for those that need to perform FFT's on data.

dpanici:
Need to consider symmetry of the underlying data

dpanici:
Could check if at `theta=0, zeta=0` the value is zero, then it is `sin` sym, but I would prefer to manually specify the symmetry

dpanici:
https://github.com/PlasmaControl/DESC/blob/95612fa625bd5ba939569004c767ed02594586f0/tests/test_axis_limits.py#L194

PR #1514, opened 2025-01-09
Adding old objectives

unalmis:
until the nan in forward for GammaC is resolved (1489) and nuffts are used (#1294)


desc/objectives/_neoclassical.py
```diff
@@ -283,3 +283,120 @@ def compute(self, params, constants=None):
             **self._hyperparam,
         )
         return constants["transforms"]["grid"].compress(data["effective ripple"])
+
+
+class OldEffectiveRipple(EffectiveRipple):  # noqa: D101
```
dpanici, line 288: maybe keep it a single class but add a `interpolation_method` flag to be `1D` or `2D`

unalmis: that would be clean but we did not know if combining the objectives would have disadvantage for jit compilation

desc/objectives/_fast_ion.py
```diff
@@ -295,3 +303,105 @@ def compute(self, params, constants=None):
             **self._hyperparam,
         )
         return constants["transforms"]["grid"].compress(data[self._key])
+
+
+class GammaC_Spline(GammaC):  # noqa: D101
```
dpanici, line 308: use a flag to choose method

f0uriest:
This would also be a good place to add benchmarks for the bounce integral objectives (compute and jacobian, ideally using both 1d and 2d versions)

desc/compute/_old.py
f0uriest, line 1: maybe something like `_neoclassical_1d.py` rather than just `old` or `deprecated`

unalmis:
> This would also be a good place to add benchmarks for the bounce integral objectives (compute and jacobian, ideally using both 1d and 2d versions)

That is additional work for a separate pull request. You'd need to scan through different parameter settings (num alpha, num transit, num quad, eq.M, eq.N). Also compare forward, reverse and the current jax issues prevent that. Then re-run and reinterpret the results with with each of the remaining major updates, #1154, #1294, #1303, #1206, #1034, and the two open PRs for faster transforms and partial summation/vandermonde stuff.

f0uriest:
> > This would also be a good place to add benchmarks for the bounce integral objectives (compute and jacobian, ideally using both 1d and 2d versions)
> 
> That is additional work for a separate pull request. You'd need to scan through different parameter settings (num alpha, num transit, num quad, eq.M, eq.N). Also compare forward, reverse and the current jax issues prevent that. Then re-run and reinterpret the results with with each of the remaining major updates, #1154, #1294, #1303, #1206, #1034, and the two open PRs for faster transforms and partial summation/vandermonde stuff.

I don't think we need to scan over every parameter, the default for each objective is fine. But it would be useful to have some sort of benchmark before doing those other improvements so we know what helps and what doesn't

unalmis:
> I don't think we need to scan over every parameter, the default for each objective is fine. But it would be useful to have some sort of benchmark before doing those other improvements so we know what helps and what doesn't

#1575 can be merged here upon review

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -458,3 +459,58 @@ def run(obj, con):
...
+        ]
+    )
+    objective.build(eq)
+    objective.compile()
+    x = objective.x(eq)
+
+    def run(x, objective):
+        objective.compute_scaled_error(x, objective.constants).block_until_ready()
+
+    benchmark.pedantic(run, args=(x, objective), rounds=3, iterations=1)
```
YigitElma, line 488: If this doesn't take really long time, I would suggest to increase the number of rounds to at least 10 ish. Github runners seems to be noisy.

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -458,3 +459,58 @@ def run(obj, con):
...
```
YigitElma, line 516: Same as above

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -458,3 +459,58 @@ def run(obj, con):
         rounds=10,
         iterations=1,
     )
+
+
+@pytest.mark.slow
+@pytest.mark.benchmark
```
YigitElma, line 465: We don't run the gpu benchmarks on CI but still we try to keep cpu and gpu files the same. Can you add these to that file too? When you commit you need to put `--no-verify` to prevent pre-commit from complaining `set_device("gpu")` line.

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -458,3 +459,60 @@ def run(obj, con):
...
+            EffectiveRipple(
+                eq,
+                num_transit=num_transit,
+                num_well=10 * num_transit,
+                num_quad=16,
+            )
+        ]
+    )
+    objective.build(eq)
+    objective.compile(mode="bfgs")
```
YigitElma, line 483: `bfgs` will compile `objective.compute_scalar` and `objective.grad`, so, in this case compiling doesn't help. Personally, I favor calling `objective.compute_scaled_error(x, objective.constants).block_until_ready()` here, instead of `objective.compile`. One, it prevents this kind of confusion, and two, it doesn't spend time compiling the derivatives.

unalmis: I thought we were timing this?

YigitElma: It depends on what you want to profile. For other benchmarks with the same structure, the part we time is individual calls to `objective.compute_scaled_error(x, objective.constants).block_until_ready()`, so only the function `run`. Since in real applications, the compute function is compiled once, we profile the following runtimes. If you want to also benchmark how long it takes to compile the function, add `jax,clear_caches()` inside `run` before the `compute` call.

unalmis: i'll add objective.grad to run

YigitElma: The current version will call `run` 10 times, and the first one will take very long since it also needs to compile, but the other 9 calls will be only the evaluation of compiled `compute`. If you look at the details, you can see this,
image
For printing benchmarks, we use the median value, hence this won't change things much, but still having all rounds do the same thing is good practice.

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -458,3 +459,60 @@ def run(obj, con):
...
+
+
+@pytest.mark.slow
+@pytest.mark.benchmark
+def test_ripple_objective_1D(benchmark):
+    """Benchmark computing objective for effective ripple."""
+    eq = desc.examples.get("W7-X")
+    with pytest.warns(UserWarning, match="Reducing radial"):
+        eq.change_resolution(L=eq.L // 2, M=eq.M // 2, N=eq.N // 2)
+    num_transit = 5
```
YigitElma, line 499: Also, if you think this is unrealistically low res, you can increase it. The benchmarks say that `compute` takes 80ms, we might not see improvements in this resolution.

tests/benchmarks/benchmark_cpu_small.py
YigitElma: Instead of `objective.grad` maybe use similar thing like `test_proximal_freeb_compute` and `test_proximal_freeb_jac`. Those are the actual derivatives taken during the optimization (not including the proximal.eq_update` part.

unalmis: [4be6703](https://github.com/PlasmaControl/DESC/pull/1514/commits/4be67032f64c39af4cf62cdd06c8656fee62ae57)?

unalmis: Why don't we just add benchmarking to the computations in test_neoclassical and test_fast_ion?

unalmis:
The benchmarks on this pull request run fine. However the benchmarks for master are actually running the benchmarks from commit https://github.com/PlasmaControl/DESC/pull/1514/commits/e6d806b167d4d657d84c1fb0a0d2052d33968104 on this branch, which is what fails the CI. 
@YigitElma Do you know why master benchmarks attempt to run a test from a commit that is not even in master?

tests/benchmarks/benchmark_cpu_small.py
YigitElma: I was thinking like,
```python
@pytest.mark.slow
@pytest.mark.benchmark
def test_proximal_effective_ripple_jac(benchmark):
    eq = desc.examples.get("ESTELL")
    with pytest.warns(UserWarning, match="Reducing radial"):
        eq.change_resolution(8, 8, 8, 16, 16, 16)
    num_transit = 20
        [
            EffectiveRipple(
                eq,
                num_transit=num_transit,
                num_well=10 * num_transit,
                num_quad=16,
            )
        ]
    )
    prox = ProximalProjection(objective, constraint, eq)
    prox.build()
    x = prox.x(eq)
    prox.jac_scaled_error(x, prox.constants).block_until_ready()

    def run(x, prox):
        prox.jac_scaled_error(x, prox.constants).block_until_ready()

    benchmark.pedantic(run, args=(x, prox), rounds=10, iterations=1)


@pytest.mark.slow
@pytest.mark.benchmark
def test_proximal_effective_ripple_compute(benchmark):
    eq = desc.examples.get("ESTELL")
    with pytest.warns(UserWarning, match="Reducing radial"):
        eq.change_resolution(8, 8, 8, 16, 16, 16)
    num_transit = 20
        [
            EffectiveRipple(
                eq,
                num_transit=num_transit,
                num_well=10 * num_transit,
                num_quad=16,
            )
        ]
    )
    prox = ProximalProjection(objective, constraint, eq)
    prox.build()
    x = prox.x(eq)
    prox.compute_scaled_error(x, prox.constants).block_until_ready()

    def run(x, prox):
        prox.compute_scaled_error(x, prox.constants).block_until_ready()

    benchmark.pedantic(run, args=(x, prox), rounds=10, iterations=1)


@pytest.mark.slow
@pytest.mark.benchmark
def test_proximal_effective_ripple_compute_compilation(benchmark):
    eq = desc.examples.get("ESTELL")
    with pytest.warns(UserWarning, match="Reducing radial"):
        eq.change_resolution(8, 8, 8, 16, 16, 16)
    num_transit = 20
        [
            EffectiveRipple(
                eq,
                num_transit=num_transit,
                num_well=10 * num_transit,
                num_quad=16,
            )
        ]
    )
    prox = ProximalProjection(objective, constraint, eq)
    prox.build()
    x = prox.x(eq)

    def run(x, prox):
        jax.clear_caches()
        prox.compute_scaled_error(x, prox.constants).block_until_ready()

    benchmark.pedantic(run, args=(x, prox), rounds=5, iterations=1)
```

Feel free to change resolutions and stuff. But the main point, if you don't create a `ProximalProjection`, the derivative is computed for each equilibrium parameter R_lmn, Rb_lmn etc. In optimization, we rarely take the full derivative, `ProximalProjection` reduces the derivatives to be only in boundary parameters Rb_lmn. For compute benchmarks, proximal shouldn't change anything, but I just copy/pasted jacobian cases. You can write your own version for that.

YigitElma: > Why don't we just add benchmarking to the computations in test_neoclassical and test_fast_ion?

For compute, you can use one of them I guess.

YigitElma:
> The benchmarks on this pull request run fine. However the benchmarks for master are actually running the benchmarks from commit [4200972](https://github.com/PlasmaControl/DESC/commit/42009726a41659c73d71bd6144401decc31b5acb) on this branch, which is what fails the CI. @YigitElma Do you know why master benchmarks attempt to run a test from a commit that is not even in master?

Yeah, this is expected. Our benchmarks basically run the same `benchmarks_cpu_small.py` file for both master and PR branches. So, we checkout to master inside benchmarks but we keep the `benchmarks_cpu_small.py` the same as the feature branch.

unalmis:
> The benchmarks on this pull request run fine. However the benchmarks for master are actually running the benchmarks from commit [4200972](https://github.com/PlasmaControl/DESC/commit/42009726a41659c73d71bd6144401decc31b5acb) on this branch, which is what fails the CI. @YigitElma Do you know why master benchmarks attempt to run a test from a commit that is not even in master?

I turned on more verbose debugging. What is exactly happening is that master is attempting to run a benchmark added on this branch, however it fails because running that benchmark requires the old objectives added on this branch.

YigitElma:
If you want benchmark some code that doesn't exist in master, you should wait for this one to be merged and then add benchmarks for that unfortunately.

tests/benchmarks/benchmark_cpu_small.py
unalmis: The tests fail when I make the suggested change to use proximal instead of just using the raw objective.gradient. Am I missing something?

unalmis:
> If you want to benchmark some code that doesn't exist in master, you should wait for this one to be merged and then add benchmarks for that unfortunately.

In that case, this PR is ready for merge.

YigitElma (approved):
Ideally, I would like to have benchmarks in a separate PR, but given the high number of existing PRs and since they pass for this PR, I don't think this would cause an issue.

desc/compute/_old.py
unalmis: I think in the longer term, once all the todo's are done, these will be removed anyway.

f0uriest (approved):

PR #1515, opened 2025-01-10
Change normalization of force balance to use correct units

dpanici:
Currently we use `scales["f"]` for normalizing the force balance objectives, but they have units of Newton, not Newton per meter cubed, so this normalization technically is wrong.

- [ ] change weights for qh opt test
- [ ] rerun notebooks

### DIFF
--- a/desc/objectives/_equilibrium.py
+++ b/desc/objectives/_equilibrium.py
@@ -130,7 +130,7 @@ class ForceBalance(_Objective):
 
         if self._normalize:
             scales = compute_scaling_factors(eq)
-            self._normalization = scales["f"]
+            self._normalization = scales["F"]
 
         super().build(use_jit=use_jit, verbose=verbose)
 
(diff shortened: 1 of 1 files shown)


f0uriest:
The residual returned by the objectives is also newtons, not per volume, so I think it's correct as is?

dpanici:
> The residual returned by the objectives is also newtons, not per volume, so I think it's correct as is?

On master we use `f` as the normalization which has units of N/m^3

the objectives return units of `N` like you say. This PR is changing the normalization from `f` to `F` to match the units

f0uriest:
https://github.com/PlasmaControl/DESC/blob/bc0654ec6f8423ffced8f83b895fd59032fb91c7/desc/objectives/normalization.py#L45-L50

`scales['F']` is Newtons/m^3, `scales['f']` is Newtons, so I think its correct as is.

YigitElma:
> https://github.com/PlasmaControl/DESC/blob/bc0654ec6f8423ffced8f83b895fd59032fb91c7/desc/objectives/normalization.py#L45-L50
> 
> `scales['F']` is Newtons/m^3, `scales['f']` is Newtons, so I think its correct as is.

I think @f0uriest is correct. We are already using the correct normalization. Maybe the confusion comes from the notation, usually the small case letters represent per volume, but here we use `F` for force per volume and `f` for force.

dpanici:
ah embarrassing, I definitely read the "f" line as being / Volume not * Volume.

PR #1520, opened 2025-01-16
Robust Coil Optimization (Wechsung)

dpanici:


desc/io/optimizable_io.py
```diff
@@ -86,7 +86,10 @@ def _unjittable(x):
     if isinstance(x, dict):
         return all([_unjittable(y) or y is None for y in x.values()])
     if hasattr(x, "dtype") and np.ndim(x) == 0:
-        return np.issubdtype(x.dtype, np.bool_) or np.issubdtype(x.dtype, np.int_)
+        try:
+            return np.issubdtype(x.dtype, np.bool_) or np.issubdtype(x.dtype, np.int_)
```
dpanici, line 90: this change was, if I recall correctly, to handle setting the `key` in `stochastic` dict to be static, I dont remember exactly the issue but I think that whatever type of object the JAX key is does not have a `.dtype` attribute, maybe

sinaatalay:
Is `perturbation` the same in each step during optimization? It would be better if it was the same.

sinaatalay:
Perturbations seems to be working:
image

- The blue line represents the perturbation in the x-direction as a function of $\theta$ (ranging from $0$ to $2\pi$, covering a complete revolution of the coil).
- The orange and green lines show the derivatives of the perturbation with respect to $\theta$, which will be used to perturb the tangent at each point.
    - The orange line corresponds to samples drawn from a normal distribution using the covariance matrix (which is what we are going to use for the optimization)
    - The green line computed numerically through finite differences, for validation.

The perturbation is periodic (perturbation at $\theta=0$ and perturbation at $\theta=2\pi$ is the same), a crucial aspect since the coil’s position must be identical at $\theta$ = 0 and $\theta = 2\pi$.

dpanici:
> Perturbations seems to be working:
> 
> X	Y	Z
> image	image	image
> * The blue line represents the position perturbation as a function of 
>     θ
>    (ranging from 
>     0
>    to 
>     2
>     π
>   , covering a complete revolution of the coil).
> * The orange and green lines show the derivatives of the position perturbation with respect to 
>     θ
>   , which will be used to perturb the tangent at each point.
>   
>   * The orange line corresponds to samples drawn from a normal distribution using the covariance matrix (which is what we are going to use for the optimization)
>   * The green line computed numerically through finite differences, for validation.
> 
> The perturbation is periodic (position perturbation at θ = 0 and perturbation at θ = 2 π is the same), a crucial aspect since the coil’s position must be identical at θ = 0 and θ = 2 π .
> 
> Here is an example of a coil and its perturbed version (large standard deviation is used for illustration):
> 
> image
> 
> Here is an optimization example (black one is the result of the regular optimization, red one is the result of the stochastic optimization):
> 
> image
> 
> Here is what the API will look like:
> 
> ```python
> QuadraticFlux(
>     eq,
>     field=coilset,
>     eval_grid=plasma_grid,
>     field_grid=coil_grid,
>     vacuum=True,
>     weight=weights_dict["quadratic flux"],
>     stochastic_optimization_settings={
>         "number_of_samples": 1,
>         "length_scale": 0.2,
>         "standard_deviation": 0.01,
>     },
> )
> ```
> 
> Some questions:
> 
> 1. Do we want to add `stochastic_optimization_settings` key to all the coil objectives?

@sinaatalay how were these figures made?

desc/objectives/_coils.py
```diff
@@ -1513,6 +1666,13 @@ def __init__(
             jac_chunk_size=jac_chunk_size,
         )
 
+        if stochastic_optimization_settings:
+            # Validate the settings
+            StochasticOptimizationSettings(**stochastic_optimization_settings)
```
YigitElma, line 1673: I guess better api would be to pass this class directly to the objective rather than using it as a checker.

YigitElma: We probably make it a subclass of `IOAble` and add static attrs to make it jittable

desc/objectives/_coils.py
```diff
@@ -1621,17 +1801,35 @@ def compute(self, *field_params, constants=None):
...
+                params=field_params,
+                perturbations=perturbations,
+                chunk_size=self._bs_chunk_size,
+            )
+            B_ext = jnp.sum(B_ext * eval_data["n_rho"], axis=-1)
+            f = (B_ext + B_plasma) * jnp.sqrt(eval_data["|e_theta x e_zeta|"])
+            return f
+
+        if self._stochastic_settings:
+            stochastic_settings = StochasticOptimizationSettings(
```
YigitElma, line 1821: I mean for no reason we create this class multiple times. Having a single input and modify it would be better, and also make it easier to apply for other classes.

desc/objectives/_free_boundary.py
```diff
@@ -236,14 +278,25 @@ def compute(self, eq_params, *field_params, constants=None):
...
-            basis="rpz",
-            params=field_params,
-            chunk_size=self._bs_chunk_size,
-        )
+        if not self._use_perturbation:
+            Bext = constants["field"].compute_magnetic_field(
+                x,
+                source_grid=self._field_grid,
+                basis="rpz",
+                params=field_params,
```
YigitElma, line 286: This should work with perturbations=None (set self._perturbations default as None) to reduce redundant code.

desc/objectives/_free_boundary.py
```diff
@@ -757,13 +851,23 @@ def compute(self, eq_params, *field_params, constants=None):
...
         # can always pass in field params. If they're None, it just uses the
         # defaults for the given field.
-        Bext = constants["field"].compute_magnetic_field(
-            x,
-            source_grid=self._field_grid,
-            basis="rpz",
-            params=field_params,
-            chunk_size=self._bs_chunk_size,
-        )
+        if not self._use_perturbation:
```
YigitElma, line 854: Same as above

dpanici:
Maybe instead of current API make a "perturbed coilset" class which would have `perturbations` or something related to that as its params (though it would NOT be optimizable... or just force use of `MixedCoilset`?)

PR #1536, opened 2025-01-26
Free boundary speed

unalmis:
The evaluation grid and source grid should be chosen to be the same to use the more efficient interpolator (#1519). This PR adds logic for that case so that stuff isn't recomputed on both grids.

### DIFF
--- a/desc/objectives/_free_boundary.py
+++ b/desc/objectives/_free_boundary.py
@@ -22,9 +22,6 @@ from desc.utils import (
 from ..integrals.singularities import _get_default_params
 from .normalization import compute_scaling_factors
 
-# TODO: Often eval grid and source grid are the same, so don't recompute
-#       everything twice. Just check if eval grid == source grid first.
-
 
 class VacuumBoundaryError(_Objective):
     """Target for free boundary conditions on LCFS for vacuum equilibrium.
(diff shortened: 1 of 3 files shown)


dpanici (changes requested):
- put a note somewhere in the objective docstrings that if the eval grid and source grid are the same, some computations can be saved

desc/objectives/_free_boundary.py
```diff
@@ -496,12 +495,18 @@ def build(self, use_jit=True, verbose=1):
             source_grid = self._source_grid
 
         if self._eval_grid is None:
-            eval_grid = LinearGrid(
-                rho=np.array([1.0]), M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False
+            eval_grid = (
+                LinearGrid(
+                    rho=np.array([1.0]), M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False
+                )
+                if eq.N == 0
```
YigitElma, line 502: I have a question, why do we have this special case?

YigitElma (approved):

desc/objectives/_free_boundary.py
unalmis: This deserves more explanation so I have pushed a commit to address it.

In that special case the default source grid is chosen differently (1/64 of the toroidal domain) than the default evaluation grid (a single field period), and I did not want to change the default settings for free boundary in this PR.

Issue #1530, opened 2025-01-20
Take advantage of separable basis functions


unalmis:
Right now, anytime we compute a series expansion, for all of the basis classes in DESC e.g. `ZernikeSeries`, `DoubleFourierSeries`, the resulting Vandermonde matrix is computed by first forming an outer product of the mode numbers then evaluating the basis functions.

For each point $x$, this generates a large mode matrix $A(x)$ of size $L M N$. If there are $P$ points then the matrix $A$ has size $P L M N$. From this matrix all the basis functions $f \colon A \mapsto f(A)$ are evaluated. In the performance analysis for this type of stuff, the dominant expense is by far evaluating $f$. Hence, this is an inefficient approach to computing the Vandermonde matrix because $f$ must be evaluated at $P L M N$ times.

Now, because the basis function associated with any given spectral coefficient $a_{l,m,n}$ are separable in the spatial dimensions, more efficient methods can be used to generate the Vandermonde matrix. In particular for any basis the $K$ dimensional Vandermonde matrix should be computed by 
1. first forming $K$ one-dimensional Vandermonde matrices $\{V_i | i \in (1,...,K)\} $
2. Computing the basis functions on this matrices $\{f(V_i) | i \in (1,...,K)\} $
3. Generate the $K$ dimensional Vandermonde by taking the outer product $f(V_1) \otimes f(V_2) \otimes, \dots, \otimes f(V_K)$

This works for any grid; there does not need to be a special structure. This can be done complement to the partial summation techniques discussed in #1154 and #1405 . I have already updated the `DoubleFourierSeries` basis in #1440 with this change and the code should be able to be copied to the other basis.

f0uriest:
I think the `unique` option of `Basis.evaluate` already sort of does this, and is improved in #1405 to not need repeated calls to `np.unique`

unalmis:
1. That option is not JIT compilable or differentiable.
2. Calling `np.unique` and then indexing the unique modes and expanding them with gather operations are unnecessary and somewhat expensive. It also involves storing more things in memory.

Issue #1243, opened 2024-09-03
Making transforms more efficient


f0uriest:
Trying to coalesce a few things we've been discussing regarding making transforms more efficient.

## 1. Partial `basis.evaluate`
When calling `basis.evaluate` you should be able to specify which coordinates to actually evaluate. Default would be all of them, but evaluating only radial eg would make the next step easier. It would also modularize things better (right now we directly call `zernike_radial` several places where we really mean `basis.evaluate` but only in the `r` coordinate), making it easier to swap in different basis types (finite element, chebyshev, etc)
## 2. Partial `transform.transform`
Similarly, `transform` should allow you to transform only one or two coordinates at a time. For example, instead of `lmn -> rtz` you could do `lmn -> rmn` to only evaluate the radial part. This would mean breaking the transform matrices into individual parts, which we sort of already do because for FFTs we only use matrices for the radial/poloidal part. This should also make it easier to do poloidal FFTs (#641). 
## 3. Allow you to promise "same nodes" when updating `transform.grid`
ie, `transform.update_grid(new_grid, promise_same="r")` you would be saying this grid is the same as the old one in the radial direction, so that you only need to recompute the poloidal and toroidal part of the transform matrices. So for example, in `map_coordinates` assuming rho is both an input and output coordinate:
```python
transforms = get_transforms(names..., grid=initial_guess_grid)
rho = coords[:,0]

def rootfun(x):
    nodes = np.stack([rho, x]) # x is only theta/zeta
    new_grid = Grid(nodes)
    transforms.update_grid(new_grid, promise_same="r")
    data = _compute_fun(names, params, transforms, ...)
```
this is more along the lines of "partial evaluation" rather than "partial summation" but would likely see similar improvements (#1154)
## 4. Move logic from `transform._check_inputs_fft` to `Grid` and `Basis`
For most grid classes we don't really need to do these checks since we know by construction that certain things are/are not satisfied. This would simplify the transforms a bit and speed up construction a bit if we don't need to do a bunch of checks. (It would also be nice if we can get rid of some of the indexing operations there in favor of just reshaping, but im not sure how that plays with symmetry where certain modes don't exist)
## 5. Extend notion of "meshgrid" to partial meshgrids/bases
eg, `ConcentricGrid` has the same rho/theta at each zeta, so its sort of a tensor product grid (with some accompanying simplifications). Similarly, a `FourierZer nikeBasis` is a partial tensor product.

unalmis:
[...]

PR #1537, opened 2025-01-27
Partial summation in bounce integrals

unalmis:
- [x] Applies (#1530) in the remaining places in the bounce integrals.
- [x] Partial summation (#1154, #1243) for the field line integral and chebyshev series.
- [x] Partial summation via FFT for the cubic spline.
- [x] Fold over field period for the cubic spline


desc/integrals/_bounce_utils.py
```diff
@@ -779,26 +766,31 @@ def cubic_spline(T, f, Y, n0, n1, NFP=1, check=False):
...
+    num_transit = T.X
+    zeta = jnp.linspace(-1, 1, Y, endpoint=False)
+    # θ at uniformly spaced points in ζ along field line
+    theta = idct_non_uniform(zeta, T.cheb[..., jnp.newaxis, :], T.Y)
+    zeta = bijection_from_disc(zeta, T.domain[0], T.domain[1])
+
+    # f at uniformly spaced points in ζ along field line
+    f = _irfft2_non_uniform(
+        theta,
+        zeta,
```
unalmis, line 782: Take 5 min tomorrow to use a uniform fft in zeta here so that we only need 1d non uniform transform

desc/integrals/_bounce_utils.py
```diff
@@ -691,79 +679,93 @@ def fourier_chebyshev(theta, iota, alpha, num_transit):
...
+        θ over one toroidal transit.
     f : jnp.ndarray
+        Shape broadcasts with (num rho, 1, n_modes.size, m_modes.size).
         Fourier transform of f(θ, ζ) as returned by ``Bounce2D.fourier``.
-        ``n0=grid.num_theta``, ``n1=grid.num_zeta``, ``NFP=grid.NFP``.
     Y : int
-        Number of knots per transit to interpolate ``f``.
-    n0 : int
+        Number of knots per toroidal transit to interpolate ``f``.
+        This number will be rounded down to an integer multiple of NFP.
```
unalmis, line 762: I confirmed this is the reason the baseline images for the neoclassical and fast ion confinement stuff changed. This can be confirmed easily by setting NFP to 1 everywhere in this function except for the line of code with `domain=(0, 2 * jnp.pi / NFP)`.

desc/integrals/bounce_integral.py
unalmis, line 1: None of these changes affect public API. Just internal stuff that was necessary to do the partial summation.

desc/integrals/_interp_utils.py
```diff
@@ -137,24 +137,59 @@ def irfft_non_uniform(xq, a, n, domain=(0, 2 * jnp.pi), axis=-1):
...
+        if (n % 2) == 0:
+            i = (0, -1)
+        else:
+            i = 0
+        a = jnp.moveaxis(a, axis, -1).at[..., i].divide(2) * 2
+    vander = jnp.exp(1j * _modes * (xq - domain[0])[..., jnp.newaxis])
     return (vander * a).real.sum(axis=-1)
 
 
+def ifft_non_uniform(xq, a, domain=(0, 2 * jnp.pi), axis=-1, _modes=None):
```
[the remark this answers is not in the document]
unalmis: identical except the modes are different

desc/integrals/_bounce_utils.py
```diff
@@ -779,26 +781,45 @@ def cubic_spline(T, f, Y, n0, n1, NFP=1, check=False):
...
+
+    # Let m, n denote the poloidal and toroidal Fourier resolution. We need to
+    # compute a set of 2D Fourier series each on uniform (non-uniform) in ζ (θ)
+    # tensor product grids of size
+    #   |𝛉|×|𝛇| where |𝛉| = num alpha × num transit × NFP and |𝛇| = Y/NFP.
+    # Partial summation via FFT is more efficient than direct evaluation when
+    # mn|𝛉||𝛇| > m log(|𝛇|) |𝛇| + m|𝛉||𝛇| or equivalently n|𝛉| > log|𝛇| + |𝛉|.
+
+    if num_zeta >= f.shape[-2] and num_zeta == f.shape[-2]:
+        # FIXME: This does not work unless num_zeta == f.shape[-2],
```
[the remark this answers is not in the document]
unalmis: Yes. I did not figure it out.

desc/integrals/bounce_integral.py
```diff
@@ -681,7 +691,36 @@ def integrate(
...
+    #     This is now possible.
+    #  4. Uiform FFT can be used in toroidal direction. Combined with partial
+    #     summation the interpolation becomes cheap.
+    #     (Same code as ``desc/integrals/_bounce_utils.py::cubic_spline``).
+    #  5. The quadrature points are no longer functions of the solutions
+    #     to the nonlinear equation λB = 1. In particular, all the ζ values
+    #     are constants throughout optimization. This makes AD cheaper.
+    #  6. Because the interpolation is now purely a function of ``num_transit``
+    #     and θ, rather than the apriori unknown number of bounce points, the
+    #     expensive JAX limitation in GitHub issue #1303 is avoided.
```
unalmis, line 717: @f0uriest thoughts?

desc/integrals/_bounce_utils.py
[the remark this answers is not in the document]
f0uriest: Can you make an issue and add the number to the comment so we can track it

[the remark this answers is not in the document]
unalmis: done

desc/integrals/bounce_integral.py
f0uriest: It can work well as long as you can find a good function to subtract. The usual issue is that "easy" functions to subtract aren't particularly smooth so they slow convergence of the now nonsingular portion of the integral.

I think knosos does something similar to what you're describing by analytically integrating some approximation of the singularity at the bounce point? But they then still need to do adaptive quadrature to compute the non-singular part.

f0uriest (approved):

desc/integrals/bounce_integral.py
unalmis: To clarify, for strongly singular integrals, I think fast convergence requires a change of variable like we do. Looking at the convergence plots in the bounce averaging paper, I can conclude that the above would not work well unless the singularity can be subtracted to high order accuracy and that subtracted quantity can be analytically integrated somehow. (So I do not intend for this alternative to be used for `Gamma_c` etc.).

For weakly singular integrals, however, (e.g. effective ripple, and the invariant J), suppose we subtracted the singularity with just first order accuracy (same taylor expansion in comment above). Because the original singularity was weak, some math shows that the resulting integrand is smooth (bounded derivative). Also, for symmetric wells, the resulting integrand is periodic. Of course, the derivative of the resulting integrand is **not** smooth.

Looking at the convergence plots in the bounce averaging paper for weakly singular integrals, I think that the penalty from applying a uniform quadrature to an integrand that is smooth (but whose derivative is not smooth) is probably less than ``num_pitch``>=50, so it could work better for the reasons discussed in the comments.

unalmis:
Could someone override codecov to merge? thanks

rahulgaur104 (approved):

rahulgaur104:
After the benchmarks run successfully.

unalmis:
and make sure to click the down arrow adjacent to the merge button and select squash & merge please

PR #1543, opened 2025-01-27
update VMECIO.save defaults

ddudt:
Allows for setting the grid resolution in real space, so that quantities can be oversampled before fitting to a Fourier series with the Nyquist resolution.


desc/vmec.py
```diff
@@ -208,7 +208,18 @@ def load(
         return eq
 
     @classmethod
-    def save(cls, eq, path, surfs=128, verbose=1, M_nyq=None, N_nyq=None):  # noqa: C901
+    def save(  # noqa: C901
+        cls,
+        eq,
+        path,
+        *,
+        surfs=128,
```
ddudt, line 216: allow for positional arg?

desc/vmec.py
```diff
@@ -208,7 +208,18 @@ def load(
...
+        M_nyq=None,
+        N_nyq=None,
+        M_grid=None,
+        N_grid=None,
```
ddudt, line 220: Is this equivalent to one of the VMEC inputs? NTHETA, NZETA?

f0uriest (approved):

desc/vmec.py
```diff
@@ -250,10 +264,24 @@ def save(cls, eq, path, surfs=128, verbose=1, M_nyq=None, N_nyq=None):  # noqa:
...
-            "Passed in N_nyq but equilibrium is axisymmetric, setting N_nyq to zero",
+            "Passed in N_nyq but equilibrium is axisymmetric, setting N_nyq to zero.",
         )
         N_nyq = N + 2 if N_nyq is None else N_nyq
         N_nyq = 0 if int(N) == 0 else N_nyq
+        M_grid = eq.M_grid if M_grid is None else M_grid
+        N_grid = eq.N_grid if N_grid is None else N_grid
+        warnif(
+            M_grid < M_nyq,
+            UserWarning,
```
YigitElma, line 275: We need to add warning exception for some tests after this change...

YigitElma: Also, this will throw a warning for M=2 and M_grid=4, but that case is technically correct, right? It is double the highest frequency.

YigitElma: Failing tests as far as I see,
`test_vmec_save_kinetic`
`test_vmec_save_asym`

YigitElma (changes requested):

desc/vmec.py
ddudt: The grid resolution should be at least as high as the Nyquist resolution. So if M_nyq=2 and M_grid=4, that is fine. But in this case it was M=2, M_nyq=6, M_grid=4 and that would result in under-sampling.

YigitElma (approved):

f0uriest (approved):

PR #1545, opened 2025-01-28
newcomb compute parallel in rho

a957924278:
parallization in rho, in computation of "Newcomb ballooning metric"; Now u may directly compute it on different rhos, without "for":
```python
rho= np.linspace(0.1, 0.9, 9)
grid = Grid.create_meshgrid(
    [rho, alpha, zeta],
    coordinates="raz",
    period=(np.inf, 2 * np.pi, np.inf),
    NFP=eq.NFP,
)
```


a957924278:
sorry for opening a pr rudely; I've got no much exp in git, maybe i should fix ideal ballooning mode and notebooks, then commit at the same time?

dpanici:
Hi! Thanks for the PR, I have not reviewed in detail yet, but can you
- Add a test for it (can just make the [current test](https://github.com/PlasmaControl/DESC/blob/b3b6dcc495716418ea01b7c0ff489f66ff578aac/tests/test_stability_funs.py#L530) for this compute variable compute it on more than one rho, with no for loop)
- change the notebook to do the same
Also, how do you have an idea of the impact on memory use this change has? I imagine it may make the computation of the value and its derivative take more memory possibly

dpanici:
Also, make sure to install the pre-commit hook detailed in our [Contributing page](https://github.com/PlasmaControl/DESC/blob/master/CONTRIBUTING.rst#python-styleguide), this will catch any formatting inconsistencies before you commit!

Issue #1534, opened 2025-01-23
Can not use gpu to run desc on perlmutter.


WDDJGWDDJG:
Hi, 

I followed the installation instruction to install perlmutter on NERSC. But I cannot use gpu to run desc on perlmutter

If I run `import desc.io`, There will be long error output. The last part of it is, 

> RuntimeError: Unable to initialize backend 'cuda': INVALID_ARGUMENT: Unexpected PJRT_Client_Create_Args size: expected 88, got 72. The plugin is likely built with a later version than the framework. This plugin is built with PJRT API version 0.64. (set JAX_PLATFORMS='' to automatically choose an available backend).

But if I run `desc.set_device('gpu')`, there will be no error output.

rahulgaur104:
Hi WDDJGWDDJG,

Can you run conda list and tell me your jax version? I suspect it has jumped to 0.5.0 which may be a potential source of the issue.
The recent maintenance on Perlmutter could also be another issue.

WDDJGWDDJG:
I run conda list and it shows
```
jax                       0.4.38                   pypi_0    pypi
jax-cuda12-pjrt           0.5.0                    pypi_0    pypi
jax-cuda12-plugin         0.5.0                    pypi_0    pypi
jaxlib                    0.4.38                   pypi_0    pypi
```
Is it the problem that the version of jax and jaxlib are not the same as jax-cuda12-pjrt and jax-cuda12-plugin ?

rahulgaur104:
That's one possibility. You can change the version of cuda12-plugin by using

`pip upgrade "jax-cuda12-plugin==0.4.38"`

If that version is not available, try a version where all the jax libraries have the same version.

WDDJGWDDJG:
I run `pip install --upgrade "jax[cuda12]==0.4.38"` and reinstall desc using `pip install --editable .` and It works now. Thank you for help

rahulgaur104:
Make sure that your python script is running as well before we close the issue. Sometimes set_device works correctly but the code throws an error later.

If your script works, please let us know so we can close this issue.

WDDJGWDDJG:
I think my python scipts is running well. I tried the precise QA example case and nothing goes wrong. Thank you

PR #1547, opened 2025-01-29
update della install instructions

dpanici:
Can probably be reversed once #1539 is resolved

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -141,25 +141,24 @@ Della and Stellar Clusters (Princeton)
 First, install JAX for the latest version of `jaxlib` available on the Princeton clusters.
 
 We base our instructions below off of `this tutorial <https://github.com/PrincetonUniversity/intro_ml_libs/tree/master/jax>`__, if the below instructions do not work please
-check the link to install JAX with the most up-to-date recommendations from the Princeton computing services:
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -147,7 +147,7 @@ check the link to install JAX with the most up-to-date recommendations from the
 
     conda create --name desc-env python=3.12 -y
     conda activate desc-env
-    pip install -U "jax[cuda12]"
+    pip install -U "jax[cuda12]==0.4.38"
```
YigitElma, line 150: Can we add a note here like,
Note: We try to keep DESC compatible with the most recent JAX version but sometimes, dependencies may require lower JAX versions or some cuda tools are not available for the clusters, in these cases, installed `jax, jax-cuda12-pjrt, jax-cuda12-plugin, jaxlib` packages may have different versions. To prevent problem for using GPUs, you might need to specify a specific version of JAX here. See Github issue #1534.

YigitElma: Actually, I think the package manager should be more intelligent about this issue, but this is frequent enough to mention in the docs imo.

dpanici:
change command to install requirements first, then overwrite the cpu jax installed to instead get the GPU version matching the cpu-installed jax version.

ddudt (approved):

YigitElma (approved):

PR #1555, opened 2025-02-01
Add utility for initializing helical coils

### LINKED ISSUE #1160: Utilities for initializing coilsets
It would be nice to make a few utils for initializing coilsets from a given equilibrium.

- [x] #1182
    - Package what we do in the tutorial notebook, with planar coils aligned with the axis
    - method of coilset? or standalone function?
- [x] #1183
     - Package stuff from tutorial, with windowpane offset in n x direction
- [ ] #1184
    - For helical coils I thought we could evaluate `x` from the equilibrium along a helical curve `theta = helicity * zeta` on the LCFS, then offset that curve by some fixed distance in the `n_rho` direction and fit that with splines/fourier etc?

### LINKED ISSUE #1184: Utility for creating helical coils from equilibrium/surface

f0uriest:
Resolves #1160 
Resolves #1184


desc/coils.py
```diff
@@ -3050,3 +3050,64 @@ def initialize_saddle_coils(eq, num_coils, r_over_a=0.5, offset=2.0, position="o
...
+
+    s = np.linspace(0, 2 * np.pi, npts, endpoint=False)
+
+    theta = M * s
+    zeta = N * s
+
+    theta %= 2 * np.pi
+    zeta %= 2 * np.pi
+
+    theta_offset = np.linspace(0, 2 * np.pi, num_coils, endpoint=False)
```
f0uriest, line 3101: this will always have a coil near theta=zeta=0, is that something we need an option for?

desc/coils.py
```diff
@@ -3050,3 +3050,64 @@ def initialize_saddle_coils(eq, num_coils, r_over_a=0.5, offset=2.0, position="o
...
+        may need more points.
+
+    Returns
+    -------
+    coilset : CoilSet of FourierPlanarCoil
+        Planar coils centered on magnetic axis, with appropriate symmetry.
+    """
+    a = eq.compute("a")["a"]
+    G = eq.compute("G", grid=LinearGrid(rho=1.0))["G"]
+    M = helicity[0] * eq.NFP
```
f0uriest, line 3090: i know this looks wrong but im pretty sure its right. For example, heliotron with helicity=(1,1), we'd expect the coil to follow the ridges of the plasma, which would mean it links the plasma NFP times poloidally

desc/objectives/_coils.py
f0uriest, line 1: in writing tests for the other stuff i noticed this was a bit wrong (in that it the default grid had the wrong NFP, and would throw warnings). This saves some computation by only calculating over 1 period (ensuring we have enough resolution for that) then duplicates the points to the other periods bc thats needed for accurate distance calculations.

desc/coils.py
dpanici: I think it is fine, they could always rotate the coil around if they don't want it there

dpanici: our symmetry stuff does not reduce effort for helical coils anyways right now iirc

desc/objectives/_coils.py
```diff
@@ -835,6 +835,15 @@ def body(k):
         return min_dist_per_coil
 
 
+def _copy_rpz_periods(rpz, NFP):
```
dpanici, line 838: this is done in another PR as well right? the distance one?

tests/test_coils.py
```diff
@@ -1386,3 +1389,32 @@ def test_initialize_saddle():
     np.testing.assert_allclose(x[:, 2], -offset)  # Z ~ -3
     np.testing.assert_allclose(x[:, 0].min(), -1, rtol=1e-2)  # xmin
     np.testing.assert_allclose(x[:, 0].max(), 1, rtol=1e-2)  # xmax
+
+
+@pytest.mark.unit
+def test_initialize_helical():
+    """Test initializing a modular coilset."""
```
dpanici, line 1396:
```suggestion
    """Test initializing a helical coilset."""
```

dpanici (approved):

desc/objectives/_coils.py
f0uriest: Yeah whichever gets second will need to be rebased a bit

desc/coils.py
```diff
@@ -3050,3 +3050,64 @@ def initialize_saddle_coils(eq, num_coils, r_over_a=0.5, offset=2.0, position="o
...
+    npts : int
+        How many points to use when creating the coils. Equilibria with very high NFP
+        may need more points.
+
+    Returns
+    -------
+    coilset : CoilSet of FourierPlanarCoil
+        Planar coils centered on magnetic axis, with appropriate symmetry.
+    """
+    a = eq.compute("a")["a"]
```
YigitElma, line 3088: Might need to check if M and N are ints first. For example,

```python
eq = get("NCSX")
coilset = initialize_helical_coils(eq, 1, r_over_a=2.0, helicity=(1, 0.5), npts=100)
fig = plot_3d(eq, "|B|")
fig = plot_coils(coilset, fig=fig)
fig.show()
```
image

YigitElma: Also, if `N=0`, we might ask the user to use another function, or say this is not helical coil, reducing M to make the coil single loop, or something on that front

desc/coils.py
```diff
@@ -3050,3 +3050,64 @@ def initialize_saddle_coils(eq, num_coils, r_over_a=0.5, offset=2.0, position="o
...
+    M = helicity[0] * eq.NFP
+    N = helicity[1]
+
+    s = np.linspace(0, 2 * np.pi, npts, endpoint=False)
+
+    theta = M * s
+    zeta = N * s
+
+    theta %= 2 * np.pi
+    zeta %= 2 * np.pi
```
YigitElma, line 3099: I know this is just an initializer, so it is not too important, but with high NFP cases, we might get some weird results like this,
image
This is HELIOTRON with M=1 and N=0.

YigitElma (changes requested):
Might need to add some checks for M and N

desc/coils.py
dpanici: That's a good point

desc/coils.py
f0uriest: this is because you're basically making a modular coil (N=0) that links NFP times. I've added warnings to check for that.

YigitElma (dismissed):

YigitElma (dismissed):

YigitElma (approved):

dpanici (approved):

PR #1556, opened 2025-02-01
Some improvements to distance calculation objectives

### LINKED ISSUE #652: Make plasma-surface distance calculations robust when using NFP
Currently, if we want to compute the `PlasmaVesselDistance` objective for two surfaces with `NFP>1`, and let's say the surface is such that for the point on the plasma surface at $(\theta=0, \zeta=0)$ the closest point on the winding surface is at $(\theta=0,\zeta=\epsilon$ where $\epsilon>0$ is some small number of radians. Then due to the symmetry, we can say that the closest point on the surface for the plasma point $(\theta=0, \zeta=2\pi/NFP)$ is on the surface at $(\theta=0, \zeta=2\pi/NFP+\epsilon)$, which is NOT included on the grids we by default use when both surfaces have `NFP>1`, resulting in instead a closest point which is not the true closest point.

To remedy this, we can calculate distance not by using cartesian distance formula (which we do now) but instead by calculating the cylindrical coordinates of each, and calculating the distance using some modding of the toroidal angle with $2\pi/NFP$ to account for this

### LINKED ISSUE #1063: Add softmin for coil coil and coil plasma distance
might also want an option for softmin?

_Originally posted by @f0uriest in https://github.com/PlasmaControl/DESC/pull/977#discussion_r1645117792_

### LINKED ISSUE #1064: Add option for vmap + reduction in coil coil  and coil plasma distance objectives
>does this need to be a loop or can we do vmap + reduction? It might be good to have both options with a flag like for the free boundary objective? or a batched vmap or something?

_Originally posted by @f0uriest in https://github.com/PlasmaControl/DESC/pull/977#discussion_r1645117235_

f0uriest:
- Uses full torus when measuring distance in `PlasmaVesselDistance`, rather than just a single field period
- Adds option to use softmin to `CoilSetMinDistance` and `PlasmaCoilSetMinDistance`
- Adds `dist_chunk_size` to `CoilSetMinDistance` and `PlasmaCoilSetMinDistance` for breaking the distance calculations into smaller pieces (or really larger pieces since previously it used a loop so chunk_size=1). Default is no chunking which will generally be fastest, and the memory requirements shouldn't be too bad since the coils are just 1d.

Resolves #652 
Resolves #1063 
Resolves #1064

### DIFF
--- a/desc/compute/geom_utils.py
+++ b/desc/compute/geom_utils.py
@@ -161,3 +161,12 @@ def rpz2xyz_vec(vec, x=None, y=None, phi=None):
         return cart
 
     return inner(vec, phi)
+
+
+def copy_rpz_periods(rpz, NFP):
+    """Copy a rpz position vector into multiple field periods."""
+    r, p, z = rpz.T
+    r = jnp.tile(r, NFP)
+    z = jnp.tile(z, NFP)
+    p = p[None, :] + jnp.linspace(0, 2 * jnp.pi, NFP, endpoint=False)[:, None]
+    return jnp.array([r, p.flatten(), z]).T
--- a/desc/objectives/_coils.py
+++ b/desc/objectives/_coils.py
@@ -891,6 +891,14 @@ class PlasmaCoilSetMinDistance(_Objective):
         during optimization, and self.things = [eq] only.
         If False, the coil coordinates are computed at every iteration.
         False by default, so that self.things = [coil, eq].
+    use_softmin: bool, optional
+        Use softmin or hard min. Softmin is a smooth approximation to the actual minimum
+        distance that may give smoother gradients, at the expense of being slightly more
+        expensive and only an approximate minimum.
+    softmin_alpha: float, optional
+        Parameter used for softmin. The larger ``softmin_alpha``, the closer the
+        softmin approximates the hardmin. softmin -> hardmin as
+        ``softmin_alpha`` -> infinity.
 
     """
 
@@ -921,6 +929,8 @@ class PlasmaCoilSetMinDistance(_Objective):
         coils_fixed=False,
         name="plasma-coil minimum distance",
         jac_chunk_size=None,
+        use_softmin=False,
+        softmin_alpha=1.0,
     ):
         if target is None and bounds is None:
             bounds = (1, np.inf)
@@ -930,6 +940,8 @@ class PlasmaCoilSetMinDistance(_Objective):
         self._coil_grid = coil_grid
         self._eq_fixed = eq_fixed
         self._coils_fixed = coils_fixed
+        self._use_softmin = use_softmin
+        self._softmin_alpha = softmin_alpha
         errorif(eq_fixed and coils_fixed, ValueError, "Cannot fix both eq and coil")
         things = []
         if not eq_fixed:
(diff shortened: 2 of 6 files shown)


unalmis (approved):

desc/objectives/_coils.py
```diff
@@ -818,20 +834,19 @@ def compute(self, params, constants=None):
...
-            mask = jnp.ones(self.dim_f).at[k].set(0)[:, None, None]
-            return jnp.min(dist, where=mask, initial=jnp.inf)
-
-        min_dist_per_coil = fori_loop(
-            0,
-            self.dim_f,
-            lambda k, min_dist: min_dist.at[k].set(body(k)),
-            jnp.zeros(self.dim_f),
-        )
+            coil_pts = pts[k]
```
dpanici, line 841: I've worked on this code before but I forget how the broadcasting here works for the `dist` line, can you remind me?

```python
        def body(k):
            # pts shape (ncoils, num_nodes, 3)
            # dist btwn all pts; shape(ncoils,num_nodes,num_nodes)
            # dist[i,j,n] is the distance from the jth point on the kth coil
            # to the nth point on the ith coil
            coil_pts = pts[k] # (k,num_nodes,3)
            other_pts = jnp.delete(pts, k, axis=0, assume_unique_indices=True) # (ncoils - k, num_nodes, 3)
            # (1,         k,     1,    num_nodes   ,     3 )
            # (     ncoils - k,  1,     num_nodes,       3)
            # ?? should be ( something like ( k, ncoils-k, num_nodes, 3) which gets passed into safenorm
            # to yield (k, ncoils-k, 3)? or no something else...
            dist = safenorm(coil_pts[None, :, None] - other_pts[:, None], axis=-1) # 
            if self._use_softmin:
                return softmin(dist, self._softmin_alpha)
            return jnp.min(dist)```

File: desc/compute/geom_utils.py [hunk 2]
```diff
@@ -161,3 +161,12 @@ def inner(vec, phi):
         return cart
 
     return inner(vec, phi)
+
+
+def copy_rpz_periods(rpz, NFP):
```
dpanici, line 166: when #1555 is in, should unify this utility and the one of the same name from that PR (since they are the same I think?)

desc/objectives/_coils.py
f0uriest: `k` is just an integer
`coil_pts` is shape `(num_nodes, 3)`
`other_pts` is shape `(num_coils-1, num_nodes, 3)`
`coil_pts[None,:,None]` is shape `(1, num_nodes, 1, 3)`
`other_pts[:,None]` is shape `(num_coils-1, 1, num_nodes, 3)`

so the argument to safenorm is shape `(num_coils-1, num_nodes, num_nodes, 3)`
safenorm contracts along the last dimension so dist is shape `(num_coils-1, num_nodes, num_nodes)`
so `dist[i,j,n]` is the distance from the jth point on the kth coil to the nth point on the ith coil`

ddudt (approved):

desc/objectives/_geometry.py
```diff
@@ -461,13 +462,6 @@ class PlasmaVesselDistance(_Objective):
     points on surface corresponding to the grid that the plasma-vessel distance
     is evaluated at, which can cause cusps or regions of very large curvature.
 
-    NOTE: When use_softmin=True, ensures that softmin_alpha*values passed in is
-    at least >1, otherwise the softmin will return inaccurate approximations
-    of the minimum. Will automatically multiply array values by 2 / min_val if the min
-    of softmin_alpha*array is <1. This is to avoid inaccuracies when values <1
-    are present in the softmin, which can cause inaccurate mins or even incorrect
-    signs of the softmin versus the actual min.
```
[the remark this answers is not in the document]
dpanici: yea I forgot to remove, when Kaya corrected my use of softmin this no longer became necessary.

PR #1557, opened 2025-02-02
Switch compute from recursive to looped

### LINKED ISSUE #1336: Change compute function algorithm from recursive to loops over tiers
#1147

f0uriest:
This PR builds upon #2156 

- `_compute` now uses a for loop instead of recursion
  - We still need to call `_get_deps` to consider given data
- Cleans up data dependency getters and removes the redundant functions

Resolves #1336


desc/compute/utils.py
```diff
@@ -184,39 +184,20 @@ def _compute(
     if data is None:
         data = {}
 
-    for name in names:
+    names += get_data_deps(
```
YigitElma: If we decide to go with #2156, we can make have something like,
```python
deps_type = "full_with_axis_dependencies" if transforms["grid"].axis.size else "full_dependencies"
all_deps = []
for name in names:
    all_deps += data_index[p][name][deps_type ]["data"]
all_deps = list(set(all_deps)) # remove repeated stuff
all_deps = sorted(all_deps, key=topological_order[p])
```
We should store the topological_order. This can be a global variable created in `_build_data_index` using `topo_index`. Or we can add it to `tier` but they are not exactly equivalent, since `topo_index` just gives an order, whereas `tier` specifically means how many dependency levels there are.

I can remove the `set_tier` in this PR.

YigitElma: This removes `get_data_deps` call, but not sure about the effect on performance.

YigitElma:
The logic should take into account that a given data may make certain sub dependencies obselete

desc/compute/utils.py
```diff
@@ -307,8 +245,6 @@ def _get_deps(parameterization, names, deps, data=None, has_axis=False, check_fu
         Type of object to compute for, eg Equilibrium, Curve, etc.
     names : str or array-like of str
         Name(s) of the quantity(s) to compute.
-    deps : set[str]
```
YigitElma, line 310: Previously this function was recursive, so this deps was needed but not anymore

desc/compute/utils.py
```diff
@@ -404,25 +342,11 @@ def get_derivs(keys, obj, has_axis=False, basis="rpz"):
     """
     p = _parse_parameterization(obj)
     keys = [keys] if isinstance(keys, str) else keys
-
-    def _get_derivs_1_key(key):
-        if has_axis:
-            if "full_with_axis_dependencies" in data_index[p][key]:
```
YigitElma, line 410: We used to call this function in `_build_data_index`, but now `data_index` is already built when we call this function

desc/compute/__init__.py
```diff
@@ -43,7 +43,7 @@
     _stability,
     _surface,
 )
-from .data_index import all_kwargs, allowed_kwargs, data_index
+from .data_index import _topological_order, all_kwargs, allowed_kwargs, data_index
```
YigitElma, line 46: Putting this to `./data_index.py` prevent circular import

YigitElma:
I am not sure if this PR is necessary, given that there is no performance difference. I guess it is cleaner than before, anyway it is ready for review.

desc/compute/__init__.py
ddudt: [question](https://pullpo.io/cc?l=question) 
I want to understand what is happening:

1. `desc/compute/data_index.py` creates an empty dict: `_topological_order = {}`
2. `desc/compute/__init__.py` imports then populates this dict with values: `_topological_order[p] = topo_index`
3. `desc/compute/utils.py` imports this dict from `data_index` then uses it

Why does the empty dict need to be created in `data_index.py` and not in `__init__.py`?

ddudt (commented):
need to fix the failing test, otherwise this looks good

desc/compute/__init__.py
YigitElma: Yes, your summary is correct.
If I created it here (`__init__.py`), it was causing circular imports in other `compute` submodules. I can go around it by importing it inside the function, but I decided to define it in `data_index.py` to be consistent with `data_index`.

YigitElma:
> need to fix the failing test

Btw, I cannot recreate the failing test locally.

YigitElma:
- Try fixing the iota profile instead of the bounds
- Maybe also remove the bounds for the force balance constraint too

YigitElma:
@f0uriest let me know if you approve, since you opened the PR, I can approve for you

ddudt (approved):

tests/test_optimizer.py
```diff
@@ -1045,12 +1045,13 @@ def con(x):
 @pytest.mark.optimize
 def test_constrained_AL_lsq():
     """Tests that the least squares augmented Lagrangian optimizer does something."""
+    # Note: This test is known to be sensitive to small numerical differences
```
YigitElma, line 1048: Related comment about this test
https://github.com/PlasmaControl/DESC/pull/2247#issuecomment-4753257966

YigitElma:
@dpanici can you check this once you have some time?

desc/compute/utils.py
```diff
@@ -496,12 +423,18 @@ def get_params(keys, obj, has_axis=False, basis="rpz"):
...
-    deps = list(keys) + get_data_deps(keys, p, has_axis=has_axis, basis=basis)
-    params = []
-    for key in deps:
-        params += data_index[p][key]["dependencies"]["params"]
+    deps_type = "full_with_axis_dependencies" if has_axis else "full_dependencies"
+    params = set()
+    # below loop doesn't consider extra "phi" in basis="xyz" case
+    # but since "phi" doesn't have any params, no problem
+    # this way we skip calling get_data_deps again
+    # TODO (#568): This will probably need w_lmn
```
dpanici, line 431: what do you mean here?

dpanici (approved):

desc/compute/utils.py
dpanici: oh nvm. We add phi when a vector quantity is computed so that we can rotate to xyz basis if basis is xyz. In generalized toroidal angle #465 , we will also have to add `W_lmn` since `phi` will require the toroidal stream angle to compute

PR #1560, opened 2025-02-03
Improve warning for grids with missing quadrature weights

### LINKED ISSUE #1193: Improve missing quadrature weight warning to lead users to compute on uniform grid
> The changes aren't good because currently iota_r will be computed on your non-uniform straight field line grid sfl_grid, which will be signficantly less accurate do to worse convergence properties of the quadrature.
> 
> To notify you of this, the code raised a warning saying that sfl_grid doesn't have the proper quadrature weights to perform the integration, that is the _spacing attribute is missing. The patch of adding quadrature weights to the grid will suppress this warning, but it's a bad choice to do the quadrature on a grid with non-uniform spacing.
> 
> I have created a new issue to update the warning to suggest  users toward computing things on a uniform grid first when they see this warning, rather than trying to construct quadrature weights on a non-uniform grid.

_Originally posted by @unalmis in https://github.com/PlasmaControl/DESC/pull/1170#discussion_r1717355554

unalmis:
The current warning leads users to construct quadrature spacing and weights on their custom grids, i.e. following field lines or in other coordinate systems. Such grids are almost always ill-suited for quadrature. This PR updates the warning to resolve #1193.

### DIFF
--- a/desc/grid.py
+++ b/desc/grid.py
@@ -442,7 +442,13 @@ class _Grid(IOAble, ABC):
         errorif(
             self._spacing is None,
             AttributeError,
-            "Custom grids must have spacing specified by user.",
+            "Custom grids must have spacing specified by user.\n"
+            "Recall that the accurate computation of surface integral quantities "
+            "requires a specific set of quadrature nodes.\n"
+            "In particular, flux surface integrals are best performed on grids with "
+            "uniform spacing in (θ,ζ).\n"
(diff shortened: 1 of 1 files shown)


f0uriest (commented):
Can you make the PR title more descriptive?

f0uriest (approved):

dpanici (approved):

PR #1571, opened 2025-02-05
Change `maybe_add_self_consistency` to account for multiple objects of same type

dpanici:
Adds some checks to allow `maybe_add_self_consistency` to work with multiple equilibrium and curve class optimizations. 

- If the passed-in constraints have self-consistency constraint, it now checks for which thing that objective is created and adds a new one if there is no self-consistency constraint for that thing


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ New Features
 - Changes hessian computation to use chunked ``jacfwd`` and ``jacrev``, allowing ``jac_chunk_size`` to now reduce hessian memory usage as well.
 - Adds an option to ``VMECIO.save`` to specify the grid resolution in real space.
 - Adds a new objective ``desc.objectives.CoilIntegratedCurvature`` for targeting convex coils.
+- Allow non-proximal optimizers to  handle optimizing more than one ``Equilibrium`` object simultaneously.
```
YigitElma, line 21:
```suggestion
- Allows non-proximal optimizers to  handle optimizing more than one ``Equilibrium`` object simultaneously.
```

desc/objectives/getters.py
```diff
@@ -315,25 +321,62 @@ def maybe_add_self_consistency(thing, constraints):
         constraints, BoundaryRSelfConsistency
     ):
         constraints += (BoundaryRSelfConsistency(eq=thing),)
```
YigitElma, line 323: I think you again need `params` condition to check if the thing is an equilibrium. Also, it would be nice to have them under the same if like,
```python
if {"R_lmn", "Rb_lmn"} in params:
    if not is_any_instance(constraints, BoundaryRSelfConsistency):
        .... just add
    else:
        .... check if the constraint belongs to that thing
````

File: tests/test_optimizer.py [hunk 3]
```diff
@@ -1422,3 +1422,31 @@ def test_optimize_coil_currents(DummyCoilSet):
         np.asarray(coils.current).mean() * 0.15,
         np.abs(np.asarray(coils_opt.current) - np.asarray(coils.current)).mean(),
     )
+
+
+@pytest.mark.optimize
+@pytest.mark.unit
+def test_optimize_three_eq_at_once():
+    """Test optimizing 3 equilibria at the same time."""
+    eq1 = Equilibrium()
```
YigitElma, line 1431: add a comment here saying that `default equilibrium constructor produces an eq with R0=10`

YigitElma (changes requested):

desc/objectives/getters.py
```diff
@@ -306,34 +312,59 @@ def _get_NAE_constraints(
...
 
-def maybe_add_self_consistency(thing, constraints):
+def maybe_add_self_consistency(thing, constraints):  # noqa: C901
     """Add self consistency constraints if needed."""
+
+    def add_if_multiple(constraints, cls):
+        cons = get_all_instances(constraints, cls)
+        cons_on_this_thing = [con for con in cons if con.things[0] == thing]
+        if not len(cons_on_this_thing):
+            constraints += (cls(eq=thing),)
```
f0uriest, line 322: i dont think this works for the coil self consistency stuff?

desc/objectives/getters.py
```diff
@@ -306,34 +312,59 @@ def _get_NAE_constraints(
...
+        else:
+            constraints = add_if_multiple(constraints, AxisZSelfConsistency)
 
     # Curve
-    if {"shift"} <= params and not is_any_instance(constraints, FixCurveShift):
-        constraints += (FixCurveShift(curve=thing),)
-    if {"rotmat"} <= params and not is_any_instance(constraints, FixCurveRotation):
-        constraints += (FixCurveRotation(curve=thing),)
+    if {"shift"} <= params:
+        if not is_any_instance(constraints, FixCurveShift):
```
f0uriest, line 360: i think this also needs to be modified, since we should be fixing the shift/rotmat of every curve

tests/test_optimizer.py
```diff
@@ -1422,3 +1430,60 @@ def test_optimize_coil_currents(DummyCoilSet):
...
+        eq3,
+    ], _ = opt.optimize([eq1, eq2, eq3], objective=obj, constraints=cons, maxiter=1)
+
+    assert eq1.equiv(eq2)
+    assert eq3.compute(["R0"])["R0"] < eq1.compute(["R0"])["R0"]
+
+
+@pytest.mark.optimize
+@pytest.mark.unit
+def test_optimize_three_coil_at_once():
```
f0uriest, line 1474: This isn't actually optimizing anything. Should actually run a few steps and make sure shift and rotmat don't change

f0uriest (changes requested):

tests/test_optimizer.py
YigitElma: @dpanici Can you add one?

dpanici: I will note that this whole not fixing rotmat or shift when multiple coils are present, likely has not popped up anywhere because
- all of our objectives that need external fields only take in a single external field object, like one `MixedCoilSet` for instance
- the problem only would pop up if `things` contained more than one `Coil` or `Curve` or `CoilSet` object (which would not happen because of the above usually, the only times I think this occured is for flexible optimization where 2 coilsets with same geometry are optimized with different currents so like #1320 )

f0uriest (approved):

dpanici (commented):
I approve but it's my own PR. Maybe should have another person who isn't me/Yigit look at it

unalmis (approved):

PR #1575, opened 2025-02-07
Bounce benchmark

unalmis:


tests/benchmarks/benchmark_cpu_large.py
```diff
@@ -5,6 +5,7 @@
 import pytest
 
 import desc
+from desc.objectives import EffectiveRipple, ObjectiveFunction
```
f0uriest, line 8: This line needs to be moved after set_device

tests/benchmarks/benchmark_cpu_large.py
f0uriest, line 1: The large benchmarks aren't automatically run, only the small ones. Will this test not work on GitHub runners?

unalmis: It will work; I will document that in a docstring for this file

f0uriest: In that case it should be moved to the small benchmarks

PR #1577, opened 2025-02-09
Add batched reduction

### LINKED ISSUE #1564: Add some sort of batched reduction to `desc.batching`
not for this PR, but at some point might be worth implementing some sort of batched reduction `desc.batching`

_Originally posted by @f0uriest in https://github.com/PlasmaControl/DESC/pull/1440#discussion_r1940450193_
            

Basically, if we need to evaluate some expensive function over a bunch of inputs and then reduce it (eg sum). Currently the two main options are a single loop which is slow on GPU, or a full vmap which means materializing the full array in memory (or perhaps hoping the the compiler fuses stuff so it isn't but still). One could potentially improve the performance of the loop by unrolling part of it though that increases compile time. Ideally you want something that takes advantage of the fact that the loop iterations are independent.

semantically should be equivalent to the following:

```python
def unbatched_reduce(fun, x, reduction=jnp.add):
    out = 0
    for xi in x:
        out = reduction(out, fun(xi))
    return out

### PR DESCRIPTION
**Pull request by unalmis**
Resolves #1564 , which is needed to speed up testing for the laplace stuff.

### CODE REVIEW

File: desc/batching.py [hunk 1]
```diff
@@ -147,40 +172,42 @@ def _get_rest(x):
...
             y_rest = vmapped_fun(*args_rest, **kwargs)
-            y = jax.tree_util.tree_map(
-                lambda y1, y2: jnp.concatenate((y1, y2)), y_chunks, y_rest
-            )
+            if reduction is None:
+                y = jax.tree_util.tree_map(
+                    lambda y1, y2: jnp.concatenate((y1, y2)), y_chunks, y_rest
+                )
+            else:
+                y = reduction(y_chunks, y_rest)
```
unalmis, line 195: check if input needs to be tuple

desc/batching.py
```diff
@@ -58,145 +69,213 @@ def _unchunk(x):
...
+    # to avoid having to rely on xla/llvm to optimize the overhead away
+    if jax.tree_util.tree_leaves(x)[0].shape[0] == 1:
+        return _multimap(
+            lambda c, x: jnp.expand_dims(x, 0) if c else x,
+            reduction is None,
+            f(x0),
+        )
+
+    # the original idea was to use pytrees, however for now just operate on the
+    # return value tuple
```
unalmis, line 121: i kept this clunky pattern in case they follow through on this comment, so we can port it easier

desc/batching.py
```diff
@@ -1,8 +1,25 @@
...
+----------
+The chunking utility functions are adapted from the NetKet project.
+The original copyright notice is as follows
+Copyright 2021 The NetKet Authors - All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+https://github.com/netket/netket/blob/master/netket/jax/_chunk_utils.py
+https://github.com/netket/netket/blob/master/netket/jax/_scanmap.py
+https://github.com/netket/netket/blob/master/netket/jax/_vmap_chunked.py
+
+A benefit of keeping close to the original is that improvements can be ported easily.
```
unalmis, line 13: but i couldn't copy theirs because theirs doesn't work like we want.

unalmis:
ok now it's ready

PR #1580, opened 2025-02-10
Have ``obj.print_value`` return values in a dict

### LINKED ISSUE #1433: Option to save `.optimize` results to a .txt or .json file
It could be useful to save the optimization results in a friendly format so that it can be read with existing packages, `json` is one such option I think. We could add a `save_json=False` flag to `optimize` that will save a `json` file with the name of each objective and their min/max/avg values (both normalized and not), so that one need not scrape the print output in order to get that info.

f0uriest:
Resolves #1433 

The keys in the dict are the objectives `print_value_fmt`, and the values are a list of dict of arrays. (The list is needed because there may be multiple objectives of the same type). The innermost dict contains different keys depending on the objective eg `f`, `f_min`, `f_max_norm` etc, corresponding to the values that are printed for each objective.

- Adds example usage to 2 tutorial notebooks


dpanici:
possibly could use this in `optimizer.py`
```python
import contextlib
import io

def some_function_with_output():
    print("This will be printed to stdout by default.")

def suppress_output():
    with io.StringIO() as buf, contextlib.redirect_stdout(buf):
        some_function_with_output()
        # The output from some_function_with_output is now in 'buf'
        hidden_output = buf.getvalue()
    # Output is restored after the 'with' block
    # hidden_output can be used or discarded, depending on the need

print("Before suppression:")
some_function_with_output()

print("\nDuring suppression:")
suppress_output()
print("Output suppressed!")

print("\nAfter suppression:")
some_function_with_output()
```

YigitElma (changes requested):
Some of the objectives define their own `print_value.  For example, `VacuumBoundaryError` and `BoundaryError` which don't return a value. Will they work with this?

YigitElma:
I added some examples to the documentation. It looks like it is working but should we add a test just in case?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ New Features
 - Adds a new utility function ``desc.compat.contract_equilibrium`` which takes in an ``Equilibrium`` object and an argument ``inner_rho``, and returns a new ``Equilibrium`` with original ``Equilibrium``'s ``inner_rho`` flux surface as its boundary.
 Optionally can also contract the profiles of the original ``Equilibrium`` so that the new ``Equilibrium``'s profiles match the original's in real space.
 - Adds second-order NAE constraints, accessible by passing ``order=2`` to ``desc.objectives.get_NAE_constraints``.
+- - `result` dictionary returned by `Optimizer.optimize` or `eq.optimize` now includes sub-dictionary `"Objective values"` containing summary info of objective values before and after optimization, the same info that is printed to the terminal by default.
```
YigitElma, line 9:
```suggestion
- `result` dictionary returned by `Optimizer.optimize` or `eq.optimize` now includes sub-dictionary `"Objective values"` containing summary info of objective values before and after optimization, the same info that is printed to the terminal by default.
```

YigitElma (approved):

YigitElma (approved):

dpanici (approved):

YigitElma:
@f0uriest this one is ready when you want to merge it.

PR #1582, opened 2025-02-12
Make basis modes as static attribute

YigitElma:
According to the explanation `_AutoRegisterPytree` metaclass, basis modes should be static and stored in `aux_data` but if you run,
```python
eq = desc.examples.get("ATF")
child, aux_data = tree_flatten(eq)
child
```
you can see that modes are stored in children.

This PR adds the basis arguments of some classes to `_static_attrs` such that they are not traced during `jit` decorated functions. One important point to consider when making variables static is that that variable shouldn't be used in the computations. For example, we store `R_basis` in `Equilibrium` class but we never use it directly inside compute functions. The basis related stuff are stored in `constants`. That being said, if we set `_modes` of `_Basis` class as static, we will get an error with the latest changes of #1405 since full set of modes is sliced by unique indices.

During testing, I have seen that this doesn't affect memory usage, and in contrast to my expectations, it doesn't reduce the compilation time either. The speed up mainly comes from the `tree_flatten` operation. So, very simple compute functions (like ForceBalance) gets a major speed up. But others not much.


f0uriest:
We went through this a while ago and found that marking things as static generally increased compile time and memory cost, bc Jax then tries to fold the big constant array into the compiled program. The fix was to keep most things dynamic unless there was a strong reason not to

YigitElma:
> We went through this a while ago and found that marking things as static generally increased compile time and memory cost, bc Jax then tries to fold the big constant array into the compiled program. The fix was to keep most things dynamic unless there was a strong reason not to

I think this explains why making any of the `transform` stuff static either causes errors or makes some benchmarks slower. Just equilibrium basis seems to have no bad effect but I will experiment more and if it increases memory or compile time, I will close this PR.

YigitElma:
Some local benchmarks,
```python
eq = get("W7-X")
obj.build()

f = obj.compute_scaled_error(obj.x(eq)).block_until_ready() 
J = obj.jac_scaled_error(obj.x(eq)).block_until_ready()

%timeit -n 100 obj.compute_scaled_error(obj.x(eq)).block_until_ready()
%timeit obj.jac_scaled_error(obj.x(eq)).block_until_ready()

%timeit -n 10 jax.clear_caches(); obj.compute_scaled_error(obj.x(eq)).block_until_ready()
%timeit jax.clear_caches(); obj.jac_scaled_error(obj.x(eq)).block_until_ready()

%timeit jax.clear_caches(); a, b = tree_flatten(obj)
```

**MASTER**
```
12.5 ms ± 456 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
9.14 s ± 1.43 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.14 s ± 117 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
12.3 s ± 876 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
6.94 ms ± 305 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

**PR**
```
7.59 ms ± 528 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
8.82 s ± 246 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.01 s ± 44.6 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
11.4 s ± 173 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
2.5 ms ± 36.1 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I ran the tests a couple of times, memory usage looks the same. To be fair, compiled `compute` is faster but the difference fluctuates, I would believe the CI benchmarks for that.

f0uriest (approved):

unalmis (approved):

PR #1587, opened 2025-02-16
Fix bug in `desc_output_to_input` utility function for asymmetric equilibria profiles

dpanici:


tests/inputs/iotest_HELIOTRON
f0uriest, line 1: What are these changes for?

dpanici: The old input file did not have the full pressure profile resolution and thus was not actually correct (it was not representative of what the function should create). The test was not good enough to catch this as the >0 indexing made it so it did not catch that the loaded profile resolution was different than what it actually needed to be given the equilibrium resolution.

f0uriest (approved):

YigitElma (approved):

YigitElma (approved):

YigitElma (approved):

f0uriest (approved):

PR #1588, opened 2025-02-17
Replacing the incomplete integrals for shifted circle test with complete elliptic integrals

rahulgaur104:
These tests typically use complete elliptic integrals. We were using incomplete integrals.
In this PR both the calculations are done but for the final drift calculation, we use complete integrals.


unalmis (approved):

rahulgaur104:
The changes here are really small so one approval should be enough.

PR #1600, opened 2025-02-23
Simplified Bout computation in free boundary

unalmis:
This PR goal is to separate the computation of the inner and outer magnetic fields in the free boundary problem.


unalmis:
@dpanici still needs a bit of clean up, but if you wanted to start taking a look to compare to the paper i shared it it ready for that

unalmis:
we can worry about api later

PR #1603, opened 2025-02-25
Bug Fix for `ObjectiveFunction._things_per_objective_idx`

### LINKED ISSUE #1602: `ObjectiveFunction._things_per_objective_idx` bug
MWE:
```
eq = desc.examples.get("reactor_QA")
coils = initialize_modular_coils(eq, num_coils=3, r_over_a=3.0)

    (
        PlasmaCoilSetMinDistance(eq=eq, coil=coils, eq_fixed=False),
        LinkingCurrentConsistency(eq=eq, coil=coils, eq_fixed=False),
    )
)
objective.build()
print(objective._things_per_objective_idx)
```
Actual output:
`[[0, 1], [0, 1]]`
Desired output:
`[[0, 1], [1, 0]]`

`PlasmaCoilSetMinDistance` expects params for `things=[eq, coils]`, but `LinkingCurrentConsistency` expects params for `things=[coils, eq]`. The logic to compute `ObjectiveFunction._things_per_objective_idx` always returns the `things` sorted in the same order for each objective, regardless of the unique order defined by each objective.

ddudt:
Resolves #1602


desc/objectives/objective_funs.py
```diff
@@ -409,7 +409,7 @@ def _set_things(self, things=None):
         self._things_per_objective_idx = []
         for obj in self.objectives:
             self._things_per_objective_idx.append(
-                [i for i, t in enumerate(unique_) if t in obj.things]
+                [unique_.index(t) for t in obj.things if t in unique_]
```
f0uriest, line 412: you can get rid of the `if t in unique_` part. In theory `t` should always be in `unique_` unless something goes horribly wrong, and in that case its better to have the `unique_.index(t)` throw an error rather than silently truncating the list

PR #1606, opened 2025-02-26
Fix bug that occurs when setting coilset current with a 0-D array

dpanici:
Bug was popping up in `test_things_per_objective_idx(self)` locally for me, specifically

```python
linking_current = 2 * np.pi * eq.compute("G", grid=grid)["G"][0] / mu_0
coils.current = linking_current / coils.num_coils
```

the second line here, the error was that zip cannot deal with 0-D arrays. Upon debugging, it seemed that for some reason `jnp.isscalar(new)` was False for the new current being assigned here, which was weird. Putting the at least 1D fixes this now, and the broadcast_to should still raise an error if a weird length array is passed in.

### DIFF
--- a/desc/coils.py
+++ b/desc/coils.py
@@ -1285,9 +1285,10 @@ class CoilSet(OptimizableCollection, _Coil, MutableSequence):
 
     @current.setter
     def current(self, new):
-        if jnp.isscalar(new):
-            new = [new] * len(self)
-        for coil, cur in zip(self.coils, jnp.atleast_1d(new)):
+        new = jnp.atleast_1d(new)
+        if new.size == 1:
+            new = jnp.broadcast_to(new, (len(self),))
+        for coil, cur in zip(self.coils, jnp.atleast_1d(new), strict=True):
             coil.current = cur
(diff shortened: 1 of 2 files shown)


rahulgaur104 (approved):

desc/coils.py
```diff
@@ -1287,7 +1287,7 @@ def current(self):
     def current(self, new):
         if jnp.isscalar(new):
             new = [new] * len(self)
-        for coil, cur in zip(self.coils, new):
+        for coil, cur in zip(self.coils, jnp.atleast_1d(new)):
```
dpanici, line 1290: add flag for strict=True

desc/coils.py
```diff
@@ -1287,7 +1287,7 @@ def current(self):
     def current(self, new):
         if jnp.isscalar(new):
```
dpanici, line 1288: make atleast_1d new
broadcast to same length as self.coils if size is 1

YigitElma (approved):

f0uriest (approved):

PR #1613, opened 2025-03-01
Ensure correct grid is used for coil biot-savart computation

unalmis:
Should always enforce the right grid is used


desc/coils.py
```diff
@@ -1609,6 +1609,10 @@ def _compute_A_or_B(
             ValueError,
             f'Expected "A" or "B" for compute_A_or_B, instead got {compute_A_or_B}',
         )
+        if source_grid is not None:
+            assert (
+                source_grid.NFP == self.NFP
+            ), f"{source_grid.NFP} does not equal {self.NFP}"
```
unalmis, line 1615: This fails. Should it?

YigitElma:
```
source_grid : Grid, int or None, optional
            Grid used to discretize coils. If an integer, uses that many equally spaced
            points. Should NOT include endpoint at 2pi.
```
I think you should change the condition to `if source_grid is not None and isinstance(source_grid, _Grid):`

unalmis: Yes, but the tests fail because the NFP differs so that does not resolve the issue.

f0uriest: I think usually we don't throw errors if the grid has NFP 1 even if the basis doesn't. As long as nzeta is enough then it's still correct in that case

unalmis: for example, every NFP needs to be the same for the biot-savart in the singular integrals, so this may also be the case here.

f0uriest: for eg FourierXYZCoil, if you use a grid with NFP!=1 the results will be incorrect. And if you have a coilset made of them, even if the coilset has NFP=2, you still should use NFP=1 grids

unalmis: Then could you please add checks for these cases then and justify/state which field period the comments refer to?

YigitElma (approved):

desc/coils.py
```diff
@@ -1609,6 +1631,13 @@ def _compute_A_or_B(
             ValueError,
             f'Expected "A" or "B" for compute_A_or_B, instead got {compute_A_or_B}',
         )
+        # NFP symmetry applies to coilset as a whole, not individual coils, so the grid
+        # should have NFP=1.
+        errorif(
+            getattr(source_grid, "NFP", 1) != 1,
+            ValueError,
+            "source_grid for CoilSet must have NFP=1",
+        )
```
unalmis, line 1640: If you scroll down, the nfp loop for the biot-savart uses ``self.NFP`` not source_grid.NFP. it may be useful to be defensive and not pass in the source_grid as input to ``op``.

f0uriest: Not sure what you mean? That loop should use the coilset NFP

unalmis: The arguments given to the `op` function are ``coords`` which take ``NFP`` as self.NFP``, and ``source_grid``, which may have a different ``NFP`` than the one associated to ``coords``. Can this lead to problems in the future? If so, I don't think ``source_grid`` is used for the ``op`` function anyway, so maybe it's better to not pass it in.

unalmis:
This is Rory's PR and it has mine and Yigit's approval.

PR #1615, opened 2025-03-01
Explain to users why variable computations may differ

### LINKED ISSUE #1424: Apparent mismatch between ballooning objective output and direct calculation of growth rates

### LINKED ISSUE #1614: Question: why <|B|>_axis and R0/a Depend on rho in compute_fun but Not in eq.compute
Hi guys,

I have a question to ask when I using desc to do some calculations.

When computing <|B|>_axis and R0/a using compute_fun, the results vary with different rho values. However, using eq.compute('<|B|>_axis', grid=grid) or eq.compute('R0/a', grid=grid), the results remain independent of rho.

```python

data_keys = ["drho/B^2", "G", "I", "iota", "<|B|>_axis", "R0/a"]
grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=eq.sym, rho=[rho])

profiles = get_profiles(data_keys, obj=eq, grid=grid)
transforms = get_transforms(data_keys, obj=eq, grid=grid)
constants = {"transforms": transforms, "profiles": profiles, "helicity": (1, eq.NFP)}

data = compute_fun(
    data_keys, params=eq.params_dict,
    transforms=constants["transforms"],
    helicity=constants["helicity"],
)

data["<|B|>_axis"], data["R0/a"]  # Varies with `rho`
eq.compute("<|B|>_axis", grid=grid)  # Constant across `rho`
eq.compute("R0/a", grid=grid)  # Constant across `rho`

```

Questions
	1.	Why do <|B|>_axis and R0/a from compute_fun depend on rho, while eq.compute results do not?
	2.	How should these quantities be computed correctly to ensure consistency?

Thanks for your assistance.

unalmis:
Resolves #1424 (https://github.com/PlasmaControl/DESC/issues/1424#issuecomment-2521311229) and resolves #1614.


f0uriest (dismissed):

desc/compute/_field.py
```diff
@@ -2488,7 +2488,8 @@ def _B_sub_zeta_rz(params, transforms, profiles, data, **kwargs):
     label="\\langle |\\mathbf{B}| \\rangle_{axis}",
     units="T",
     units_long="Tesla",
-    description="Average magnitude of magnetic field on the magnetic axis",
+    description="Average magnitude of magnetic field on the flux surface "
```
dpanici, line 2491: Can you mention somewhere (on the given grid) or something? Sounds a bit weird as is, like we don't evaluate things on axis or something

dpanici (changes requested):

YigitElma (changes requested):
I think for #1614 s sake, making this kind of clarification to `eq.compute`s `grid` docs would be nice, too. Currently, it states,
```
grid : Grid, optional
       Grid of coordinates to evaluate at. Defaults to the quadrature grid.
```
maybe something like,
```
grid : Grid, optional
       Grid of coordinates to evaluate at. Defaults to the quadrature grid. Note that this is the grid that final output values belong to, but not the grid to use during computing the quantities. To use this grid for the compute functions, set `override_grid=False`
```

unalmis:
I do not think there is anything confusing with the override grid documentation, so I do not feel obligated to duplicate it ten lines above. As a general principle the more writing that is added, the less people will read it

YigitElma:
> I do not think there is anything confusing with the override grid documentation, so I do not feel obligated to duplicate it ten lines above. As a general principle the more writing that is added, the less people will read it

I think most people just read the first 2 3 arguments docs and not the rest, since they are optional. In this case, this detail is stated in the optional part. But anyway

docs/write_variables.py
```diff
@@ -86,6 +86,13 @@ def write_csv(parameterization):
 those of the DESC :math:`\rho, \theta, \zeta` coordinate system. For example, ``|B|_z``
 is :math:`(\partial \vert B \vert / \partial\zeta)|_{\rho, \theta}`.
 
+Many quantities require special grids to compute accurately.
+To not burden users with such bookkeeping,
+when an object method such as ``eq.compute`` is called,
+DESC will automatically use a set of best grids for the computation.
+However, when writing objectives developers must perform the bookkeeping
+and ensure everything can be computed accurately on the chosen grid.
```
YigitElma, line 94: mention `override_grid` here

unalmis: done

YigitElma (approved):

Issue #811, opened 2023-12-24
Reduce Memory Usage by performing operations component-wise for Equilibrium Objectives


dpanici:
In looking at the traceback for an OOM error, I saw that a lot of the memory is being taken up by just concatenation operations where we are putting together 3 arrays to form a vector, like in our basis vectors or vector B. It could be helpful, memory wise, to try to write out things component-wise only instead, at least for the equilibrium objectives, which may end up reducing memory usage, since there won't be these concatenation actions taking up memory

example traceback:

```python
2023-12-23 21:03:51.178436: W external/xla/xla/service/hlo_rematerialization.cc:2202] Can't reduce memory use below 59.49GiB (63880937472 bytes) by rematerialization; only reduced to 64.19GiB (68923960570 bytes), down from 71.23GiB (76482936842 bytes) originally
2023-12-23 21:04:02.619504: W external/tsl/tsl/framework/bfc_allocator.cc:485] Allocator (GPU_0_bfc) ran out of memory trying to allocate 69.48GiB (rounded to 74602902016)requested by op 
2023-12-23 21:04:02.619827: W external/tsl/tsl/framework/bfc_allocator.cc:497] ******************__________________________________________________________________________________
2023-12-23 21:04:02.620390: E external/xla/xla/pjrt/pjrt_stream_executor_client.cc:2593] Execution of replica 0 failed: RESOURCE_EXHAUSTED: Out of memory while trying to allocate 74602901848 bytes.
BufferAssignment OOM Debugging.
BufferAssignment stats:
             parameter allocation:    2.46MiB
              constant allocation:   192.5KiB
        maybe_live_out allocation:  900.36MiB
     preallocated temp allocation:   69.48GiB
  preallocated temp fragmentation:    1.77GiB (2.54%)
                 total allocation:   70.36GiB
              total fragmentation:    1.77GiB (2.51%)
Peak buffers:
	Buffer 1:
		Size: 2.64GiB
		Operator: op_name="jit(jac_scaled)/jit(main)/vmap(jvp(jit(compute_scaled)))/mul" source_file="/home/dpanici/DESC/desc/compute/_field.py" source_line=513
		XLA Label: broadcast
		Shape: f64[3,8375,14091]
		==========================

	Buffer 2:
		Size: 2.64GiB
[...]

PR #1616, opened 2025-03-03
Use singular integral chunking in objectives

unalmis:
It is faster so best to let users explicitly choose if oom.


YigitElma:
Seems like coil notebook is still getting OOM.

unalmis:
> Seems like coil notebook is still getting OOM.

The chunk size is 1, which is what it was before.

desc/magnetic_fields/_core.py
```diff
@@ -359,7 +360,11 @@ def compute_Bnormal(
         chunk_size : int or None
             Size to split computation into chunks of evaluation points.
             If no chunking should be done or the chunk size is the full input
-            then supply ``None``. Default is ``None``.
+            then supply ``None``.
+        B_plasma_chunk_size : int or None
```
f0uriest, line 364: It's not immediately clear how this is different from the other chunk size parameter. Might be clearer if when both are present, we make clear that `chunk_size` is for the biot-savart calculation

desc/magnetic_fields/_current_potential.py
```diff
@@ -1222,7 +1223,11 @@ def solve_regularized_surface_current(  # noqa: C901 fxn too complex
     chunk_size : int or None
         Size to split computation into chunks of evaluation points.
         If no chunking should be done or the chunk size is the full input
-        then supply ``None``. Default is ``None``.
+        then supply ``None``.
+    B_plasma_chunk_size : int or None
```
f0uriest, line 1227: see above

desc/magnetic_fields/_core.py
```diff
@@ -450,14 +464,23 @@ def save_BNORM_file(
...
         scale_by_curpol : bool, optional
-            Whether or not to scale the Bnormal coefficients by curpol
+            Whether to scale the Bnormal coefficients by curpol
             which is expected by most other codes that accept BNORM files,
             by default True
+        chunk_size : int or None
+            Size to split computation into chunks of evaluation points.
+            If no chunking should be done or the chunk size is the full input
+            then supply ``None``.
+        B_plasma_chunk_size : int or None
```
f0uriest, line 474: see above

desc/magnetic_fields/_current_potential.py
```diff
@@ -1472,19 +1479,30 @@ def surfint_Bnsqd(phi_mn, I, G):
...
-    if not vacuum:  # get Bn from plasma contribution
-        Bn_plasma = compute_B_plasma(eq, eval_grid, vc_source_grid, normal_only=True)
-        Bn_plasma = Bn_plasma * ne_mag * eval_grid.weights
-
-    else:
-        Bn_plasma = jnp.zeros_like(Bn_GI)  # from plasma current, currently assume is 0
+    # get Bn from plasma current contribution
+    Bn_plasma = (
+        jnp.zeros_like(Bn_GI)
+        if vacuum
```
f0uriest, line 1485: nit: im not crazy about this kind of inline if/else if it still takes up multiple lines

desc/plotting.py
```diff
@@ -320,7 +320,9 @@ def _compute(eq, name, grid, component=None, reshape=True):
     return data, label
 
 
-def _compute_Bn(eq, field, plot_grid, field_grid):
+def _compute_Bn(
+    eq, field, plot_grid, field_grid, chunk_size=None, B_plasma_chunk_size=None
```
f0uriest, line 324: do we need to add a pass through kwarg to `plot_xx` so these can actually be set by the calling function?

f0uriest (approved):

unalmis:
override codecov and merge please

f0uriest:
> ```diff
>  test_build_transform_fft_lowres         |     +4.87 +/- 5.76     | +2.95e-02 +/- 3.49e-02 |  6.34e-01 +/- 3.4e-02  |  6.05e-01 +/- 7.3e-03  |
>  test_equilibrium_init_medres            |     -1.06 +/- 1.33     | -4.49e-02 +/- 5.62e-02 |  4.17e+00 +/- 4.0e-02  |  4.22e+00 +/- 4.0e-02  |
>  test_equilibrium_init_highres           |     -0.79 +/- 1.29     | -4.11e-02 +/- 6.71e-02 |  5.17e+00 +/- 5.7e-02  |  5.21e+00 +/- 3.6e-02  |
>  test_objective_compile_dshape_current   |     +0.40 +/- 4.58     | +1.63e-02 +/- 1.89e-01 |  4.15e+00 +/- 1.4e-01  |  4.13e+00 +/- 1.3e-01  |
>  test_objective_compute_dshape_current   |     -0.97 +/- 1.69     | -5.26e-05 +/- 9.17e-05 |  5.39e-03 +/- 6.6e-05  |  5.44e-03 +/- 6.4e-05  |
>  test_objective_jac_dshape_current       |     +1.26 +/- 6.88     | +5.34e-04 +/- 2.91e-03 |  4.29e-02 +/- 2.1e-03  |  4.24e-02 +/- 2.0e-03  |
>  test_perturb_2                          |     +0.07 +/- 2.28     | +1.48e-02 +/- 4.63e-01 |  2.03e+01 +/- 3.6e-01  |  2.03e+01 +/- 3.0e-01  |
>  test_proximal_jac_atf_with_eq_update    |     +0.08 +/- 0.86     | +1.28e-02 +/- 1.45e-01 |  1.69e+01 +/- 1.2e-01  |  1.69e+01 +/- 8.8e-02  |
> -test_proximal_freeb_jac                 |    +31.19 +/- 1.24     | +1.54e+00 +/- 6.09e-02 |  6.46e+00 +/- 5.1e-02  |  4.93e+00 +/- 3.4e-02  |
>  test_solve_fixed_iter_compiled          |     +0.42 +/- 0.87     | +8.80e-02 +/- 1.80e-01 |  2.08e+01 +/- 1.5e-01  |  2.07e+01 +/- 1.0e-01  |
>  test_LinearConstraintProjection_build   |     +0.16 +/- 2.19     | +1.81e-02 +/- 2.43e-01 |  1.11e+01 +/- 1.6e-01  |  1.11e+01 +/- 1.9e-01  |
>  test_objective_compute_ripple_spline    |     -0.09 +/- 4.91     | -3.06e-04 +/- 1.70e-02 |  3.46e-01 +/- 1.4e-02  |  3.47e-01 +/- 9.1e-03  |
>  test_objective_grad_ripple_spline       |     +0.89 +/- 2.38     | +1.25e-02 +/- 3.33e-02 |  1.41e+00 +/- 2.7e-02  |  1.40e+00 +/- 1.9e-02  |
>  test_build_transform_fft_midres         |     +2.54 +/- 7.47     | +1.64e-02 +/- 4.83e-02 |  6.63e-01 +/- 4.7e-02  |  6.47e-01 +/- 1.3e-02  |
>  test_build_transform_fft_highres        |     +0.12 +/- 3.78     | +1.17e-03 +/- 3.54e-02 |  9.37e-01 +/- 2.8e-02  |  9.36e-01 +/- 2.2e-02  |
>  test_equilibrium_init_lowres            |     +0.15 +/- 1.72     | +5.84e-03 +/- 6.94e-02 |  4.03e+00 +/- 5.9e-02  |  4.03e+00 +/- 3.7e-02  |
>  test_objective_compile_atf              |     -1.02 +/- 3.53     | -8.58e-02 +/- 2.96e-01 |  8.29e+00 +/- 2.2e-01  |  8.37e+00 +/- 2.0e-01  |
>  test_objective_compute_atf              |     +1.16 +/- 2.11     | +1.86e-04 +/- 3.37e-04 |  1.62e-02 +/- 2.4e-04  |  1.60e-02 +/- 2.3e-04  |
>  test_objective_jac_atf                  |     -1.17 +/- 1.43     | -2.29e-02 +/- 2.82e-02 |  1.94e+00 +/- 2.5e-02  |  1.97e+00 +/- 1.3e-02  |
>  test_perturb_1                          |     +0.67 +/- 2.04     | +1.02e-01 +/- 3.10e-01 |  1.53e+01 +/- 1.9e-01  |  1.52e+01 +/- 2.5e-01  |
>  test_proximal_jac_atf                   |     +0.37 +/- 0.95     | +3.00e-02 +/- 7.62e-02 |  8.04e+00 +/- 5.7e-02  |  8.01e+00 +/- 5.0e-02  |
> +test_proximal_freeb_compute             |     -9.46 +/- 1.79     | -1.77e-02 +/- 3.35e-03 |  1.70e-01 +/- 3.2e-03  |  1.87e-01 +/- 1.1e-03  |
>  test_solve_fixed_iter                   |     +0.96 +/- 1.99     | +3.12e-01 +/- 6.49e-01 |  3.29e+01 +/- 5.4e-01  |  3.26e+01 +/- 3.6e-01  |
>  test_objective_compute_ripple           |     +0.30 +/- 2.37     | +2.09e-03 +/- 1.65e-02 |  6.97e-01 +/- 6.5e-03  |  6.95e-01 +/- 1.5e-02  |
>  test_objective_grad_ripple              |     +2.65 +/- 1.28     | +6.94e-02 +/- 3.35e-02 |  2.68e+00 +/- 2.9e-02  |  2.61e+00 +/- 1.6e-02  |
> ```
> 
> 

This makes it look like the Jacobian is now a lot slower?

unalmis:
> > ```diff
> > |             benchmark_name             |         dt(%)          |         dt(s)          |        t_new(s)        |        t_old(s)        | 
> > | -------------------------------------- | ---------------------- | ---------------------- | ---------------------- | ---------------------- |
> >  test_build_transform_fft_lowres         |     +4.87 +/- 5.76     | +2.95e-02 +/- 3.49e-02 |  6.34e-01 +/- 3.4e-02  |  6.05e-01 +/- 7.3e-03  |
> >  test_equilibrium_init_medres            |     -1.06 +/- 1.33     | -4.49e-02 +/- 5.62e-02 |  4.17e+00 +/- 4.0e-02  |  4.22e+00 +/- 4.0e-02  |
> >  test_equilibrium_init_highres           |     -0.79 +/- 1.29     | -4.11e-02 +/- 6.71e-02 |  5.17e+00 +/- 5.7e-02  |  5.21e+00 +/- 3.6e-02  |
> >  test_objective_compile_dshape_current   |     +0.40 +/- 4.58     | +1.63e-02 +/- 1.89e-01 |  4.15e+00 +/- 1.4e-01  |  4.13e+00 +/- 1.3e-01  |
> >  test_objective_compute_dshape_current   |     -0.97 +/- 1.69     | -5.26e-05 +/- 9.17e-05 |  5.39e-03 +/- 6.6e-05  |  5.44e-03 +/- 6.4e-05  |
> >  test_objective_jac_dshape_current       |     +1.26 +/- 6.88     | +5.34e-04 +/- 2.91e-03 |  4.29e-02 +/- 2.1e-03  |  4.24e-02 +/- 2.0e-03  |
> >  test_perturb_2                          |     +0.07 +/- 2.28     | +1.48e-02 +/- 4.63e-01 |  2.03e+01 +/- 3.6e-01  |  2.03e+01 +/- 3.0e-01  |
> >  test_proximal_jac_atf_with_eq_update    |     +0.08 +/- 0.86     | +1.28e-02 +/- 1.45e-01 |  1.69e+01 +/- 1.2e-01  |  1.69e+01 +/- 8.8e-02  |
> > -test_proximal_freeb_jac                 |    +31.19 +/- 1.24     | +1.54e+00 +/- 6.09e-02 |  6.46e+00 +/- 5.1e-02  |  4.93e+00 +/- 3.4e-02  |
> >  test_solve_fixed_iter_compiled          |     +0.42 +/- 0.87     | +8.80e-02 +/- 1.80e-01 |  2.08e+01 +/- 1.5e-01  |  2.07e+01 +/- 1.0e-01  |
> >  test_LinearConstraintProjection_build   |     +0.16 +/- 2.19     | +1.81e-02 +/- 2.43e-01 |  1.11e+01 +/- 1.6e-01  |  1.11e+01 +/- 1.9e-01  |
> >  test_objective_compute_ripple_spline    |     -0.09 +/- 4.91     | -3.06e-04 +/- 1.70e-02 |  3.46e-01 +/- 1.4e-02  |  3.47e-01 +/- 9.1e-03  |
> >  test_objective_grad_ripple_spline       |     +0.89 +/- 2.38     | +1.25e-02 +/- 3.33e-02 |  1.41e+00 +/- 2.7e-02  |  1.40e+00 +/- 1.9e-02  |
> >  test_build_transform_fft_midres         |     +2.54 +/- 7.47     | +1.64e-02 +/- 4.83e-02 |  6.63e-01 +/- 4.7e-02  |  6.47e-01 +/- 1.3e-02  |
> >  test_build_transform_fft_highres        |     +0.12 +/- 3.78     | +1.17e-03 +/- 3.54e-02 |  9.37e-01 +/- 2.8e-02  |  9.36e-01 +/- 2.2e-02  |
> >  test_equilibrium_init_lowres            |     +0.15 +/- 1.72     | +5.84e-03 +/- 6.94e-02 |  4.03e+00 +/- 5.9e-02  |  4.03e+00 +/- 3.7e-02  |
> >  test_objective_compile_atf              |     -1.02 +/- 3.53     | -8.58e-02 +/- 2.96e-01 |  8.29e+00 +/- 2.2e-01  |  8.37e+00 +/- 2.0e-01  |
> >  test_objective_compute_atf              |     +1.16 +/- 2.11     | +1.86e-04 +/- 3.37e-04 |  1.62e-02 +/- 2.4e-04  |  1.60e-02 +/- 2.3e-04  |
> >  test_objective_jac_atf                  |     -1.17 +/- 1.43     | -2.29e-02 +/- 2.82e-02 |  1.94e+00 +/- 2.5e-02  |  1.97e+00 +/- 1.3e-02  |
> >  test_perturb_1                          |     +0.67 +/- 2.04     | +1.02e-01 +/- 3.10e-01 |  1.53e+01 +/- 1.9e-01  |  1.52e+01 +/- 2.5e-01  |
> >  test_proximal_jac_atf                   |     +0.37 +/- 0.95     | +3.00e-02 +/- 7.62e-02 |  8.04e+00 +/- 5.7e-02  |  8.01e+00 +/- 5.0e-02  |
> > +test_proximal_freeb_compute             |     -9.46 +/- 1.79     | -1.77e-02 +/- 3.35e-03 |  1.70e-01 +/- 3.2e-03  |  1.87e-01 +/- 1.1e-03  |
> >  test_solve_fixed_iter                   |     +0.96 +/- 1.99     | +3.12e-01 +/- 6.49e-01 |  3.29e+01 +/- 5.4e-01  |  3.26e+01 +/- 3.6e-01  |
> >  test_objective_compute_ripple           |     +0.30 +/- 2.37     | +2.09e-03 +/- 1.65e-02 |  6.97e-01 +/- 6.5e-03  |  6.95e-01 +/- 1.5e-02  |
> >  test_objective_grad_ripple              |     +2.65 +/- 1.28     | +6.94e-02 +/- 3.35e-02 |  2.68e+00 +/- 2.9e-02  |  2.61e+00 +/- 1.6e-02  |
> > ```
> 
> This makes it look like the Jacobian is now a lot slower?

I think this is expected. If a computation is large enough such that the computer does not have sufficient ability to take advantage of doing the computation in one batch, the end result will be slower. Basically, it may be more difficult to fuse operations if the array is larger. The Jacobian a larger computation then the function itself so it makes since that is shows up there.

unalmis:
unrelated but i think performance stuff like this where fusing can reduce compute and memory could be wortwhile to look at for hackathon #811 . it helped in #1440 for example.

YigitElma (commented):
I don't have other comments but can you please run some benchmarks on GPU? Jacobian of free boundary getting 50% slower doesn't make me comfortable. Maybe for that objective there are a lot of nested vmap's? So, changing the default for it can help.

With this setting, speed up obtained by `compute` will be cancelled (and probably slowed down) by `Jacobian`. I would like to be sure before changing the default

unalmis:
> I don't have other comments but can you please run some benchmarks on GPU? Jacobian of free boundary getting 50% slower doesn't make me comfortable. Maybe for that objective there are a lot of nested vmap's? So, changing the default for it can help.
> 
> With this setting, speed up obtained by `compute` will be cancelled (and probably slowed down) by `Jacobian`. I would like to be sure before changing the default

The slow down seen by the github ci will probably vanish after #1623 . So merge that first. And don't we default to no chunking anyway, even when it is slower than some chunking, because we assume the user chooses the chunk size that is best for their computer.

YigitElma:
> The slow down seen by the github ci will probably vanish after #1623

#1623 only makes the `jac` of `ForceBalance` part faster, it doesn't do anything about the objectives jacobian. The slow-down here is purely due to free boundary objectives jacobian. 

Some general note, `ProximalProjection` takes the derivative of `ForceBalance` and `Objective`. It uses the derivative of `ForceBalance` to find the derivative direction and takes the derivative of  `Objective`s using that. This PR only changes the second part (#1623 changes the first part). So, we are making one part faster but the other part slower.

> we assume the user chooses the chunk size that is best for their computer.

It would be nice to see what the scaling is. For example, how many cores do you need to make no chunking faster (because to me it is a bit counter-intuitive). Even though there are not enough resources to execute the function in a single batch, it should still be faster than executing the operation sequentially.

unalmis:
> Even though there are not enough resources to execute the function in a single batch, it should still be faster than executing the operation sequentially.

I don't think that is necessarily true. For example, consider when chunking materializes a large  matrix. if that matrix is large enough that operations needs to be performed sequentially, i.e. performing N matrix vector products rather than 1 matrix matrix product, then it is slower, because a large intermediate matrix needs to be materialized and operations are less likely to be fused.

YigitElma:
Ok, I ran some benchmarks on A100s, I used the same benchmarks,
```python
eq = desc.examples.get("ESTELL")
# eq.change_resolution(6, 6, 6, 12, 12, 12)
field = ToroidalMagneticField(1.0, 1.0)  # just a dummy field for benchmarking
objective = ObjectiveFunction(BoundaryError(eq, field=field))
prox = ProximalProjection(objective, constraint, eq)
obj = LinearConstraintProjection(
    prox, ObjectiveFunction((FixCurrent(eq), FixPressure(eq), FixPsi(eq)))
)
obj.build(verbose=0)
x = obj.x(eq)
f = obj.compute_scaled_error(x, obj.constants).block_until_ready()
%timeit -n 5 _ = obj.compute_scaled_error(x, obj.constants).block_until_ready()
J = obj.jac_scaled_error(x, prox.constants).block_until_ready()
%timeit -n 5 _ = obj.jac_scaled_error(x, prox.constants).block_until_ready()
```
# Results for PR

Full resolution of ESTELL (without eq.change_resolution)
```
compute:  241 ms ± 854 μs per loop (mean ± std. dev. of 7 runs, 5 loops each)
jacobian:  1.02 s ± 8.66 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
```
Reduced resolution,
```
compute:  203 ms ± 1.37 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
jacobian:  334 ms ± 1.02 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
```

# Results for Master

Full resolution of ESTELL (without eq.change_resolution)
```
compute:  367 ms ± 741 μs per loop (mean ± std. dev. of 7 runs, 5 loops each)
jacobian:  1.41 s ± 1.9 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
```
Reduced resolution,
```
compute:  263 ms ± 1.29 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
jacobian:  480 ms ± 1.57 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
```

YigitElma (approved):

unalmis:
Someone will need to override codecov for squash merge

PR #1621, opened 2025-03-03
Dump Python 3.9, add support for 3.13

YigitElma:
Before updating our dependencies, we should dump the Python 3.9 (and optionally add 3.13). The latest JAX version supported by Python 3.9 is 0.4.30 and those unit, regression and benchmarks don't use the latest JAX version, so we might miss some things.

Misc

- Adds a flag to skip the CI tests called `only-docs-comments`, if this label is present none, of the tests will run.
- Removes `requirement` changes from benchmarks (so if the only change is requirements file, benchmarks will not run). I changed this because, in benchmarks we use the same environment (the requirement file from feature branch) for both tests, so this is redundant. Optionally, we might create new environments for both tests, let me know if you want that to be the case.


YigitElma:
As far as I see, only important change is how the tabs in docstrings are handled. Now, there is no dummy tab for docstrings. If we change the docs builder python version, this probably cause some issue but I commented the fix for that.

YigitElma:
Once this is merged in, the benchmarks will run using Python 3.12 instead of 3.9. The reason why the benchmarks of this branch still use 3.9 is that we use `pull_request_target` which forces benchmarks to use `master`s definition (this was added to run benchmarks from forks)

unalmis (dismissed):

.github/workflows/benchmark.yml
```diff
@@ -44,14 +46,12 @@ jobs:
...
               - 'tests/benchmarks/**'
-              - 'requirements.txt'
-              - 'devtools/dev-requirements.txt'
               - 'setup.cfg'
               - '.github/workflows/benchmark.yml'
 
       - name: Check for relevant changes
         id: check_changes
-        run: echo "has_changes=${{ steps.changes.outputs.has_changes }}" >> $GITHUB_ENV
+        run: echo "has_changes=${{ !contains(github.event.pull_request.labels.*.name, 'only-docs-comments') && env.has_changes == 'true'}}" >> $GITHUB_ENV
```
YigitElma, line 54: Fix this

unalmis (approved):

tests/test_objective_funs.py
```diff
@@ -3673,13 +3675,22 @@ def test_objective_print_widths():
...
     """Test that the objective docstring and collect_docs are consistent."""
     objective_docs = _Objective.__doc__.rstrip()
-    doc_header = (
-        "Objective (or constraint) used in the optimization of an Equilibrium.\n\n"
-        + "    Parameters\n"
-        + "    ----------\n"
-        + "    things : Optimizable or tuple/list of Optimizable\n"
-        + "        Objects that will be optimized to satisfy the Objective.\n"
-    )
+    if Version(platform.python_version()) >= Version("3.13"):
```
f0uriest, line 3678: Why is this needed if the docs are still built with an older version of Python?

And will we need to reformat docstrings everywhere if we want to use newer python for building docs?

rahulgaur104 (approved):

YigitElma:
The expected tests doesn't exist anymore. Actions check the required workflows from master, but since this PR renamed some of the tests (only the python versions of them), they won't run (cause they don't exist anymore).

PR #1623, opened 2025-03-06
Add `blocked` option to `ProximalProjection`

### LINKED ISSUE #1589: `"lsq-exact"` optimizer hangs when computing Jacobian from `jaxify`
The `"lsq-exact"` optimizer hangs (takes an infinite amount of time to execute) when evaluating the Jacobian of an `ExternalObjective` for the first time. Our hack to fix this was to add `.block_until_ready()` on that Jacobian evaluation line. See discussion in #1028 for more details. 

This might be a JAX issue on their end?

YigitElma:
This separates the force balance part and objective part of `ProximalProjection` `jac` methods and applies batch/blocked to both of them. Basically, moves the `batched_vectorized` to objective part (but optionally, where previously we were always batching it).

Additionally, instead of taking the full jacobian and then applying the matrix vector multiplication (in this case this is exactly equivalent to jvp), replaces it with jvps in proper directions. This reduces the compute time for CPU (not for GPU).

This will be useful for concurrently calculating the jacobian of each objective in #1495 

- Fixes the bug where `blocked` mode of proximal was not using each sub-objectives `jac_chunk_size`
- Adds a lot of code explanations (especially for variable dimensions)
- Renames `_unfixed_idx_mat` to `_feasible_tangents`
- Adds `_eq_` prefix to some variable names to make it clear that they belong to the `Equilibrium`
 
Resolves #1589


unalmis:
Maybe also keep https://github.com/PlasmaControl/DESC/issues/1564#issuecomment-2654920791 in mind for potential improvements.

YigitElma:
> Maybe also keep [#1564 (comment)](https://github.com/PlasmaControl/DESC/issues/1564#issuecomment-2654920791) in mind for potential improvements.

What do you mean by this exactly?

If you say we have a big Jacobian, but at the end of each optimization iteration, what we get is a single descent direction, so having the full Jacobian is unnecessary. Yes, this is correct but with our current trust-region optimizers, we have to use some kind of matrix decomposition (hence we need the full matrix materialized). There are some iterative matrix solvers that I am currently looking to see if they are viable (for example, conjugate gradient, gmres etc). Those don't need the full matrix but rows/columns of it at a time. I believe the biggest speed up and memory reduction to DESC would come from that kind of linear system solver (making sure that the accuracy is preserved)

For this PR, I will try some other tricks too, it will be just some re-ordering.

unalmis:
> If you say we have a big Jacobian, but at the end of each optimization iteration, what we get is a single descent direction, so having the full Jacobian is unnecessary.

Yes, that. So you could partially construct it, reduce it according to one of many algorithms. Maybe https://arxiv.org/pdf/2403.16341 will be useful if you are looking into this

rahulgaur104 (approved):

desc/optimize/_constraint_wrappers.py
```diff
@@ -1181,16 +1192,17 @@ def __getattr__(self, name):
...
+        (jnp.diag(D)[:, unfixed_idx] @ Z).T, xf, constants
+    ).T
+    Fc = getattr(constraint, "jvp_" + op)(dxdc @ dc, xf, constants)
     cutoff = jnp.finfo(Fxh.dtype).eps * max(Fxh.shape)
     uf, sf, vtf = jnp.linalg.svd(Fxh, full_matrices=False)
     sf += sf[-1]  # add a tiny bit of regularization
     sfi = jnp.where(sf < cutoff * sf[0], 0, 1 / sf)
-    Fxh_inv = vtf.T @ (sfi[..., None] * uf.T)
-    return Fxh_inv @ Fc
+    return vtf.T @ (sfi * (uf.T @ Fc))
```
unalmis, line 1205: the order of matrix operations can have significant effect on compute. it it possible to add comments mentioning the dimensions of these matrices. that way we know in future whether this is better than the naiive  `Fxh_inv @ Fc` without working through the code to understand

desc/optimize/_constraint_wrappers.py
```diff
@@ -1213,5 +1225,4 @@ def _proximal_jvp_blocked_pure(objective, vgs, xgs, op):
         else:
             outi = getattr(obj, "jvp_" + op)([_vi for _vi in vi], xi, constants=const).T
             out.append(outi)
-    out = jnp.concatenate(out)
-    return out
+    return -jnp.concatenate(out)
```
unalmis, line 1228: was this wrong before?

desc/optimize/_constraint_wrappers.py
```diff
@@ -1154,13 +1171,7 @@ def _jvp(self, v, xf, xg, constants, op):
             ]
         )
         tangent = self._unfixed_idx_mat @ dfdc - dxdcv
-        if self._objective._deriv_mode in ["batched"]:
-            out = getattr(self._objective, "jvp_" + op)(tangent, xg, constants[0])
-        else:  # deriv_mode == "blocked"
-            vgs = jnp.split(tangent, np.cumsum(self._dimx_per_thing))
-            xgs = jnp.split(xg, np.cumsum(self._dimx_per_thing))
-            out = _proximal_jvp_blocked_pure(self._objective, vgs, xgs, op)
-        return -out
```
YigitElma, line 1163: There is no specific reason why I moved the - sign to inside, but basically instead of negating -out, I return the value immediately (without storing it in a dummy variable `out`. So, the results are the same. The only computational difference introduced in this PR is the use of jvp's instead of full jacobian (but again they are equivalent)

desc/optimize/_constraint_wrappers.py
YigitElma: No, it was correct. I just moved the minus sign. See comment above.

desc/optimize/_constraint_wrappers.py
YigitElma: Yeah sure. In short, here we try to find Fxh^-1 Fc. If Fxh is NxM matrix, then it's singular value decomposition will have u ∈ NxN, vt=v.T ∈ MxM and singular values s is a vector (not a matrix) of size M. Apart from some regularization (this is kind of necessary for ill-conditioned matrices where max(s)/min(s) is high), we just compute the inverse by simple transpositions.

unalmis: okay; just keep in mind /recall that
``(vtf.T @ (sfi[..., None] * uf.T)) @ Fc`` might be faster than ``vtf.T @ (sfi * (uf.T @ Fc))`` depending on the dimensions of each variable / how they broadcast

YigitElma:
I ran some of the benchmarks on the cluster GPU yesterday, there is no speed difference for jacobian calculation. So, the speed increase in the CI benchmark is only for CPU.

desc/optimize/_constraint_wrappers.py
YigitElma: I assume the jit compiler does that optimization. But since we `batch_vectorize` this operation, `Fc` is a vector, so `vtf.T @ (sfi * (uf.T @ Fc))` doesn't require any matrix-matrix product, whereas `(vtf.T @ (sfi[..., None] * uf.T)) @ Fc` this needs matrix-matrix product (`vtf.T` and `sfi[..., None] * uf.T` are matrices).

unalmis: > I assume the jit compiler does that optimization.

Neither JAX's JIT compiler nor XLA can optimize this.

Also, operation fusion, which is XLA's most important optimization for memory and speed, is not a general thing; it doesn't magically reduce all computations to be more efficient. 

If the sizes are dynamic, we should be using ``jnp.linalg.multi_dot``.

> Fc is a vector, so vtf.T @ (sfi * (uf.T @ Fc)) doesn't require any matrix-matrix product, whereas (vtf.T @ (sfi[..., None] * uf.T)) @ Fc this needs matrix-matrix product (vtf.T and sfi[..., None] * uf.T are matrices).

Then in this case we know the best order and don't need to use multi_dot.

f0uriest (commented):
I didn't see any tests verifying the two approaches give the same result?

Also, although we can't jit the whole jvp step due to the `_update_equilibrium` part, we could still probably jit smaller bits of it to improve performance a bit.

YigitElma:
> I didn't see any tests verifying the two approaches give the same result?

Do you want me to? I took tests running fine as a consistency check.



> Also, although we can't jit the whole jvp step due to the `_update_equilibrium` part, we could still probably jit smaller bits of it to improve performance a bit.

I can play around with it, but we might need to take more code out of the class methods, and pass bunch of variable to the external function.

YigitElma:
@f0uriest https://github.com/PlasmaControl/DESC/blob/de4f85318cd63ee7b7cf03c9880569357d7d0b64/tests/test_optimizer.py#L1202 tests both `blocked` and `batched` versions with a 2 previous version, which was equivalent to the previous version.

desc/optimize/_constraint_wrappers.py
```diff
@@ -1181,16 +1209,17 @@ def __getattr__(self, name):
 
 @functools.partial(jit, static_argnames=["op"])
 def _proximal_jvp_f_pure(constraint, xf, constants, dc, unfixed_idx, Z, D, dxdc, op):
-    Fx = getattr(constraint, "jac_" + op)(xf, constants)
-    Fx_reduced = Fx @ jnp.diag(D)[:, unfixed_idx] @ Z
-    Fc = Fx @ (dxdc @ dc)
-    Fxh = Fx_reduced
+    # here we are forming (dF/dx)^-1 @ dF/dc
+    # where Fxh is dF/dx and Fc is dF/dc
+    Fxh = getattr(constraint, "jvp_" + op)(
```
YigitElma, line 1214: I thought about using the final jacobian of internal eq solve for this ( I changed the returned_data to include jacobian optionally), but the problem is that one is always `jvp_scaled_error` but this one is general. If anyone has an idea lmk.

For GPU this would probably only reduce the memory usage, but for CPU it could have helped for the speed too.

dpanici: Hm, it is general BUT where in the code do we use this? prox is only ever used inside of `.optimize`, and inside here the only `jac_XXX` ever used is `jac_scaled_error`. We could either
- just assume it is jvp_scaled_error and delete the other jvp/jac stuff
- make a separate function for this that assumes `scaled_error` as `op` and then re-uses the jacobian of the last iteration of the eq solve, and use that one for the `jvp_scaled_error` and have the rest of the stuff be general

or just leave the savings on the table but since this is internal I'm fine with slightly messier code to make this faster, as long as it's well commented

desc/optimize/_constraint_wrappers.py
```diff
@@ -1107,34 +1105,61 @@ def jvp_unscaled(self, v, x, constants=None):
...
         """
+        op = "unscaled"
+        return self._jvp(v, x, constants, op)
+
+    def _jvp(self, v, x, constants=None, op="scaled_error"):
+        # The goal is to compute the Jacobian of the objective function with respect to
+        # the optimization variables (c). Before taking the jacobian, we update the
+        # equilibrium such that
+        # F(x+dx, c+dc) = 0 = F(x, c) + dF/dx * dx + dF/dc * dc
+        # where we already have F(x, c) = 0, so we can solve for dx and get
```
dpanici, line 1116:
```suggestion
        # so that we can set F(x, c) = 0, from here we can solve for dx and get
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -1181,16 +1209,17 @@ def __getattr__(self, name):
...
 @functools.partial(jit, static_argnames=["op"])
 def _proximal_jvp_f_pure(constraint, xf, constants, dc, unfixed_idx, Z, D, dxdc, op):
-    Fx = getattr(constraint, "jac_" + op)(xf, constants)
-    Fx_reduced = Fx @ jnp.diag(D)[:, unfixed_idx] @ Z
-    Fc = Fx @ (dxdc @ dc)
-    Fxh = Fx_reduced
+    # here we are forming (dF/dx)^-1 @ dF/dc
+    # where Fxh is dF/dx and Fc is dF/dc
+    Fxh = getattr(constraint, "jvp_" + op)(
+        (jnp.diag(D)[:, unfixed_idx] @ Z).T, xf, constants
```
dpanici, line 1215: is this not similar to ._unfixed_idx_mat ? could pass that in instead of Z and D and unfixed_idx

ryanwu4:
This PR also appears to resolve the issue observed in #1701 .

desc/optimize/_constraint_wrappers.py
```diff
@@ -1181,16 +1209,17 @@ def __getattr__(self, name):
 
 @functools.partial(jit, static_argnames=["op"])
 def _proximal_jvp_f_pure(constraint, xf, constants, dc, unfixed_idx, Z, D, dxdc, op):
-    Fx = getattr(constraint, "jac_" + op)(xf, constants)
-    Fx_reduced = Fx @ jnp.diag(D)[:, unfixed_idx] @ Z
-    Fc = Fx @ (dxdc @ dc)
-    Fxh = Fx_reduced
+    # here we are forming (dF/dx)^-1 @ dF/dc
```
dpanici, line 1212: including some info on matrix sizes here would be useful, like we are forming full matrices within here, but then returning only a vector right?

YigitElma:
> This PR also appears to resolve the issue observed in #1701 .

@ryanwu4 nice to hear that! But I think there were some other changes in the last year that caused it in the first place. I will look at the memory effect of this PR too.

YigitElma:
@dpanici I will address your comments this weekend, and explain some parts better

desc/optimize/_constraint_wrappers.py
```diff
@@ -1107,34 +1105,61 @@ def jvp_unscaled(self, v, x, constants=None):
...
-        # want jvp_f to only get parts from equilibrium, not other things
+        if self._objective._deriv_mode == "batched":
+            # objective's method already know about its jac_chunk_size
+            return getattr(self._objective, "jvp_" + op)(tangents, xg, constants[0])
+        else:
+            xgs = jnp.split(xg, np.cumsum(self._dimx_per_thing))
+            jvpfun = lambda u: _proximal_jvp_blocked_pure(
+                self._objective, jnp.split(u, np.cumsum(self._dimx_per_thing)), xgs, op
+            )
+            return batched_vectorize(
```
YigitElma, line 1144: I will fix this part today. The problem is if the `ObjectiveFunction` is blocked, `self._objective.jac_chunk_size` is the one set by available memory, but this is different than `self._objective.objectives` each chunk size. I thought this would be handled in the inner part, but seems to be not working.

desc/objectives/objective_funs.py
```diff
@@ -379,6 +379,8 @@ def build(self, use_jit=None, verbose=1):
                 for obj in self.objectives:
                     if obj._jac_chunk_size is None:
                         obj._jac_chunk_size = self._jac_chunk_size
+                sub_obj_chunk_sizes = [obj._jac_chunk_size for obj in self.objectives]
+                self._jac_chunk_size = min(sub_obj_chunk_sizes)
```
YigitElma, line 383: This was causing a silent bug where the `blocked` mode of the objective was not using proper chunk size. Since we were wrapping the `_proximal_jvp_f_blocked` by `batched_vectorized` with `proximal._objective._jac_chunk_size`, it was using the very high chunk size even if the user was asking for a lower.

Some clarification. Let's say we have nested vectorized functions,
```python
batched_fun_1(batched_fun_2, batch_size=1)

def batched_fun_2(x):
   return do_something_in_batches(batch_size=100)
```

`batched_fun_1` calls `batched_fun_2` inside with its argument, but since its argument is batched, `batched_fun_2` is always called by 1D arrays and effectively the `batch_size=100` is never used. This behavior doesn't change even if you increase `batch_size` of the first function, because it never passes 2D arrays to the second one, this just changes how many of those function needs to run in a vectorized manner.

I completely removed any `batched_vectorize` calls in the path of `blocked`, so this change is not necessary, but it wouldn't hurt to prevent such other cases.

dpanici: so in the context of the example you have, it makes no difference to have batched fun 2 just assume it always takes in 1D arrays instead of trying to do batched vectorize?

desc/optimize/_constraint_wrappers.py
```diff
@@ -709,9 +720,9 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
             self._unfixed_idx_mat, np.cumsum([t.dim_x for t in self.things]), axis=-1
         )
         self._unfixed_idx_mat[self._eq_idx] = self._unfixed_idx_mat[self._eq_idx][
-            :, self._unfixed_idx
-        ] @ (self._Z * self._D[self._unfixed_idx, None])
-        self._unfixed_idx_mat = np.concatenate(
+            :, self._eq_unfixed_idx
+        ] @ (self._eq_Z * self._eq_D[self._eq_unfixed_idx, None])
```
dpanici, line 724: `self._eq_z` here is actually already scaled by `_eq_D` (there is another matrix, [`self._eq_solve_objective._ZA`](https://github.com/PlasmaControl/DESC/blob/3df4e270f4a690b8f7159bfe3826c815c1de9b9d/desc/optimize/_constraint_wrappers.py#L136) which is the unscaled nullspace of the constraint matrix `A`)

Should we then NOT be multiplying Z by `eq_D` here?

desc/objectives/objective_funs.py
YigitElma: Yes, the inner function always gets 1D arrays, so there is nothing to vectorize over

desc/optimize/_constraint_wrappers.py
```diff
@@ -607,9 +618,8 @@ def _set_eq_state_vector(self):
         self._args = self._eq.optimizable_params.copy()
         for arg in ["R_lmn", "Z_lmn", "L_lmn", "Ra_n", "Za_n"]:
             self._args.remove(arg)
-        linear_constraint = ObjectiveFunction(self._linear_constraints)
-        linear_constraint.build()
```
YigitElma, line 611: Thanks!

desc/optimize/_constraint_wrappers.py
```diff
@@ -709,9 +725,11 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
             self._unfixed_idx_mat, np.cumsum([t.dim_x for t in self.things]), axis=-1
         )
         self._unfixed_idx_mat[self._eq_idx] = self._unfixed_idx_mat[self._eq_idx][
-            :, self._unfixed_idx
-        ] @ (self._Z * self._D[self._unfixed_idx, None])
-        self._unfixed_idx_mat = np.concatenate(
+            :, self._eq_unfixed_idx
+        ] @ (
+            self._eq_Z
+        )  # .eq_Z is already scaled by D, so can just use it alone here
```
YigitElma, line 731: I think you are right about this, but just in case, we should run the optimization notebooks to see what the effect is, because this changes the numerics

desc/optimize/_constraint_wrappers.py
```diff
@@ -609,7 +620,7 @@ def _set_eq_state_vector(self):
             self._args.remove(arg)
         linear_constraint = ObjectiveFunction(self._linear_constraints)
         linear_constraint.build()
```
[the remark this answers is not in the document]
YigitElma: I might have forget this during #1409. Thanks!

desc/optimize/_constraint_wrappers.py
YigitElma: I think you are right about this, but just in case, we should run the optimization notebooks to see what the effect is, because this changes the numerics

dpanici: opt tests pass at least so that's good

desc/objectives/objective_funs.py
```diff
@@ -379,6 +379,8 @@ def build(self, use_jit=None, verbose=1):
                 for obj in self.objectives:
                     if obj._jac_chunk_size is None:
                         obj._jac_chunk_size = self._jac_chunk_size
```
ddudt, line 381: The default for `_Objective.jac_chunk_size` is `None`, which implies that it should not do any chunking. But if `ObjectiveFunction.deriv_mode="blocked"` and all other defaults are used, then this line will overwrite all the `_Objective.jac_chunk_size` to use the heuristic. This can only be avoided by setting `ObjectiveFunction.jac_chunk_size=None`, but that is weird to have to set when using blocked mode and is not a documented option. We should add documentation for the `ObjectiveFunction.jac_chunk_size=None` option, and possibly also change the defaults (I'm not sure if the heuristic or no chunking is a better default).

desc/objectives/objective_funs.py
ddudt: `obj._jac_chunk_size` could be `None` for some/all objectives (see my comment above). This logic has to be updated to handle that case, since `min` cannot handle `None` values.

ddudt (changes requested):

desc/optimize/least_squares.py
```diff
@@ -178,7 +178,9 @@ def lsqtr(  # noqa: C901
     f = fun(x, *args)
     nfev += 1
     cost = 0.5 * jnp.dot(f, f)
-    J = jac(x, *args).block_until_ready()  # FIXME: block is needed for jaxify util
+    # block is needed for jaxify util which uses jax functions inside
+    # jax.pure_callback and gets stuck due to async dispatch
+    J = jac(x, *args).block_until_ready()
```
ddudt, line 183: You say this PR resolves #1589, but you didn't change this line. If it is actually resolved we should be able to remove the `block_until_ready()` call

YigitElma: It's been a while since we talked about this (https://github.com/PlasmaControl/DESC/issues/1589#issuecomment-2702010778), but in summary, this is expected behavior for `pure_callback`, so we said that we should remove the `FIXME`. This PR resolves that issue by removing the `FIXME`

desc/objectives/objective_funs.py
YigitElma: Ok, I made some changes. Now,
- This PR fixes the problem of using `objfun._jac_chunk_size` for objective case. So, it shouldn't be used again
- Only case that is necessary is the `ForceBalance` constraint `ProximalProjection`, which uses blocked mode (due to `_maybe_wrap_nonlinear_constraints`), that shouldn't be a problem because if there is only 1 sub-objective `batched==blocked`. So, only in that case, `blocked` mode will have `jac_chunk_size` for the `ObjectiveFunction`.

See the changes for more

desc/optimize/_constraint_wrappers.py
```diff
@@ -607,9 +618,8 @@ def _set_eq_state_vector(self):
         self._args = self._eq.optimizable_params.copy()
         for arg in ["R_lmn", "Z_lmn", "L_lmn", "Ra_n", "Za_n"]:
```
dpanici, line 619: maybe commenting here why we do this (these are parts of the state which our ForceBalance constraint projection (i.e. `eq.solve`) are responsible for handling, and they are a consequence of that projection, thus we don't want them considered in our DOFs (they are effectively removed by the ForceBalance constraint)

desc/optimize/_constraint_wrappers.py
```diff
@@ -649,12 +659,10 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
             Level of output.
 
         """
-        eq = self._eq
         timer = Timer()
         timer.start("Proximal projection build")
 
-        self._eq = eq
-        self._linear_constraints = get_fixed_boundary_constraints(eq=eq)
+        self._linear_constraints = get_fixed_boundary_constraints(eq=self._eq)
```
dpanici, line 665: really these are `self._eq_linear_constraints`, might be clearer like that to separate from potential linear constraints that exist (outside of this Projection) on the actual bdry or other DOFs

desc/optimize/_constraint_wrappers.py
```diff
@@ -1116,60 +1120,96 @@ def jvp_unscaled(self, v, x, constants=None):
...
-            self._Z,
-            self._D,
+            self._eq_solve_objective._unfixed_idx_mat,
             self._dxdc,
             op,
         )
         # broadcasting against multiple things
         dfdcs = [jnp.zeros(dim) for dim in self._dimc_per_thing]
         dfdcs[self._eq_idx] = dfdc
+        # note that size.dfdc != size.vs[self._eq_idx]
```
dpanici, line 1186:
```suggestion
        # note that size of dfdc != size of vs[self._eq_idx]
```

or dfdc.size and vs[...].size

as size. is not proper python so just makes the comment more confusing imo

desc/optimize/_constraint_wrappers.py
```diff
@@ -1116,60 +1120,96 @@ def jvp_unscaled(self, v, x, constants=None):
...
+        # so, the above line is not a simple assignment where you just change the value
+        # of the array, but we actually change the size of it too
+        # the size change will be handled by multiplication by the unfixed_idx_mat later
```
dpanici, line 1189: can you explain this? really it is that `dfdc.size` is NOT the same as `self.dimc_per_thing[self._eq_idx]` right?

But I am not sure why, is this bc it also has like Ra_n and Za_n? then the unfixed idx mat will remove those?

desc/optimize/_constraint_wrappers.py
```diff
@@ -1189,21 +1229,40 @@ def __getattr__(self, name):
...
-    return Fxh_inv @ Fc
+    return vtf.T @ (sfi * (uf.T @ Fc))
 
 
 @functools.partial(jit, static_argnames=["op"])
 def _proximal_jvp_blocked_pure(objective, vgs, xgs, op):
+    # Note: This function is not vectorized and takes the full set of tangents, and
+    # returns a matrix.
+
+    # vgs and xgs are list of arrays (not same size necessarily), that are split by the
```
dpanici, line 1259: but they are lists of same length, but diff size arrays inside? please clarify

dpanici (changes requested):
some clarifying comments. I also will run locally some of the opt notebooks and if I see major differences can commit changes, tho since tests pass I doubt it

desc/objectives/utils.py
```diff
@@ -78,6 +78,7 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
 
     if isinstance(objective, ProximalProjection):
         # remove cols of A corresponding to ["R_lmn", "Z_lmn", "L_lmn", "Ra_n", "Za_n"]
+        # see ProximalProjection._set_eq_state_vector
```
dpanici, line 81:
```suggestion
        # see desc.optimize._constraint_wrappers.ProximalProjection._set_eq_state_vector
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -605,11 +616,17 @@ def __init__(
     def _set_eq_state_vector(self):
         full_args = self._eq.optimizable_params.copy()
         self._args = self._eq.optimizable_params.copy()
+        # the optimizable variables for proximal are the Rb, Zb and profile
+        # coefficients. Once these are chosen, we will solve the equilibrium to
+        # find the R_lmn, Z_lmn, L_lmn, Ra_n, Za_n. That is why we remove them
+        # from the list of optimizable variables. This is accompanied by not including
+        # self-consistency constraints (see get_combined_constraint_objectives) and
```
dpanici, line 623:
```suggestion
        # self-consistency constraints (see desc.optimize.optimizer.get_combined_constraint_objectives) and
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -605,11 +616,17 @@ def __init__(
...
+        # also removing columns corresponding to these variables from the constraint
+        # matrix A in factorize_linear_constraints.
```
dpanici, line 625:
```suggestion
        # matrix A in desc.objectives.utils.factorize_linear_constraints.
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -832,18 +860,25 @@ def _update_equilibrium(self, x, store=False):
         solution when store was True
 
         """
+        # xopt is the full state vector of all the things
+        # xeq is the full state vector of the equilibrium only
+
+        # TODO: We don't need to check the whole state vector, just the
```
dpanici, line 866: add an issue for this TODO

dpanici, line 866:
```suggestion
        # TODO (#1720): We don't need to check the whole state vector, just the
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -624,18 +642,24 @@ def _set_eq_state_vector(self):
...
                 dxdc.append(dxdRb)
             if arg == "Zb_lmn":
-                c = get_instance(self._linear_constraints, BoundaryZSelfConsistency)
+                c = get_instance(self._eq_linear_constraints, BoundaryZSelfConsistency)
                 A = c.jac_unscaled(xz)[0]["Z_lmn"]
                 Ainv = np.linalg.pinv(A)
                 dxdZb = np.eye(self._eq.dim_x)[:, self._eq.x_idx["Z_lmn"]] @ Ainv
                 dxdc.append(dxdZb)
-        self._dxdc = np.hstack(dxdc)
+        # dxdc is a matrix when multiplied by the optimization variables
```
dpanici, line 659:
```suggestion
        # dxdc is a matrix that when multiplied by the optimization variables
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -624,18 +642,24 @@ def _set_eq_state_vector(self):
...
             if arg == "Zb_lmn":
-                c = get_instance(self._linear_constraints, BoundaryZSelfConsistency)
+                c = get_instance(self._eq_linear_constraints, BoundaryZSelfConsistency)
                 A = c.jac_unscaled(xz)[0]["Z_lmn"]
                 Ainv = np.linalg.pinv(A)
                 dxdZb = np.eye(self._eq.dim_x)[:, self._eq.x_idx["Z_lmn"]] @ Ainv
                 dxdc.append(dxdZb)
-        self._dxdc = np.hstack(dxdc)
+        # dxdc is a matrix that when multiplied by the optimization variables
+        # gives the full state vector (Rb_lmn and Zb_lmn part will be 0, but they will
```
dpanici, line 660: is "full_state_vector" here referring to `thing.x` for every `thing`, and so for eq includes R_lmn and Rb_lmn?

dpanici (approved):

desc/optimize/_constraint_wrappers.py
YigitElma: I will update the comment. It just returns the full_state_vector of the equilibrium, not the other things.
image

dpanici (dismissed):

YigitElma:
I explain the details of Proximal Projection in this document [Proximal_Projection_and_Force_Balance_Constraint.pdf](https://github.com/user-attachments/files/20193795/Proximal_Projection_and_Force_Balance_Constraint.pdf)

For reference, I have a big theory/code explanation document in Overleaf which can be accessed from this [repo](https://github.com/YigitElma/Introduction-to-DESC-and-Stellarators).

ddudt (approved):
I think now the `ForceBalance` objective in a `ProximalProjection` will default to no chunking, but you can specify a chunk size manually. Compared to before, where I think the default was using the heuristic chunk size. That is fine, I just want to understand if default behavior has changed.

dpanici:
> I think now the `ForceBalance` objective in a `ProximalProjection` will default to no chunking, but you can specify a chunk size manually. Compared to before, where I think the default was using the heuristic chunk size. That is fine, I just want to understand if default behavior has changed.

It is used when `._constraint._jac_chunk_size` is called in `get_tangents` (where the df/dx is computed), which is `ObjectiveFunction(ForceBalance)`, so it will still use the default if I understand the code correctly.

why do you say it changes this behavior?

dpanici (approved):

desc/optimize/_constraint_wrappers.py
```diff
@@ -137,8 +137,22 @@ def build(self, use_jit=None, verbose=1):
...
+        # these tangents.
+        # For example, let's say the full state vector X has constraints X1=X2 and
+        # X = [X1 X2 X3]. The reduced state vector of this is Y = [Y1 Y2]. We can take
+        # Y1=X1=X2 and Y2=X3. Then df/dY1 = df/dX1 + df/dX2 and df/dY2 = df/dX3.
+        # in this case, feasible_tangents = [ [1 , 0], [1, 0], [0,1]]
+        # and is a shape 3x2 matrix equivalent to dx/dy
+        # s.t. df/dy = df/dx @ dx/dy
+
+        # TODO (1721):: Z is already scaled by D, so we don't need to scale it again.
+        # Update the weights of the coil tests
```
f0uriest, line 154: is this related to 1721 or is this a separate issue?

f0uriest (approved):

desc/optimize/_constraint_wrappers.py
YigitElma: We believe we are double scaling but when I removed the second one, some coil tests were failing. We believe that if we change the weights (which are probably too low/high now), this can be resolved. I should add this explanation to the issue

f0uriest (approved):

desc/optimize/_constraint_wrappers.py
dpanici: I'm not completely sure we are double scaling tbh but the issue is a reminder to check

dpanici (approved):

PR #1629, opened 2025-03-08
Update jdotb formula when saving to wout, and don't negate current profile in `ensure_nested_jacobian`

dpanici:
- No longer negates the current profile in `ensure_nested_jacobian` when flipping from a left-handed to a right-handed coordinate system
- changes the formula for the `jdotb` variable to better match VMEC outputted `jdotb`

### DIFF
--- a/desc/vmec.py
+++ b/desc/vmec.py
@@ -362,6 +362,12 @@ class VMECIO:
                 "sqrt(g)",
                 "<|B|^2>",
                 "<J*B>",
+                "G_r",
+                "G",
+                "I_r",
+                "I",
+                "psi_r",
+                "sqrt(g)_Boozer",
             ],
             grid=grid_full,
         )
@@ -605,6 +611,7 @@ class VMECIO:
         buco = file.createVariable("buco", np.float64, ("radius",))
         buco.long_name = "Boozer toroidal current I, on half mesh"
         buco.units = "T*m"
+
         buco[1:] = -grid_half.compress(data_half["I"])  # - for negative Jacobian
         buco[0] = 0
 
--- a/desc/vmec.py
+++ b/desc/vmec.py
@@ -708,8 +708,29 @@ class VMECIO:
         jdotb = file.createVariable("jdotb", np.float64, ("radius",))
         jdotb.long_name = "flux surface average of J*B, on full mesh"
         jdotb.units = "N/m^3"
-        jdotb[:] = grid_full.compress(data_full["<J*B>"])
-        jdotb[0] = 0
+        # in VMEC, they use the form of parallel current from
+        # assuming Boozer coordinates, which is what we will also use here.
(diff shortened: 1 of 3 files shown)


dpanici:
- check for stellarator if not negating current keeps iota profile magnitude the same

dpanici:
- Should add documentation to the ensure nested jacobian saying that it will assume your inputs for current/Psi are what you physically want to preserve

dpanici:
> * check for stellarator if not negating current keeps iota profile magnitude the same

This seems to work fine, at least going from VMEC input to DESC input, I check it comparing to the wout for [this NCSX](https://github.com/landreman/vmec_equilibria/tree/master/NCSX/fixed_boundary_from_free_boundary) (which has substantial iota from current of almost 0.2 at the edge) and we are close to vmec wout's iota magnitude (within like 0.02), if this was an issue we would see a big difference between the two. Also checked it when loading the VMEC eq into desc with `VMECIO.load(fname, profile="iota")` (which fits the eq, and will flip the current sign it reads from VMEC since bdry is left-handed and thus the `buco` variable sign is negative of the toroidal current sign, and the iota profile matches up well too.

I also loaded in the surface from the input file without changing the orientaiton (so it is left-handed bdry in DESC), solved, checked that iota mag against the others (matches), then called `ensure_positive_jacobian` on it (which swaps the bdry but does NOT swap the current profile sign), and checked the iota magnitude (it remained unchanged) while the iota sign was negated (as needed to keep the magnetic field physically pointing in the same direction, poloidal direction was flipped so iota had to be negated). Force balance in the left hand and right hand eq were equal as well (after accounting for checking the F at nodes where theta_left = - theta_right)

dpanici:
However, the magnetic field poloidal direction seems to be flipped

> * Should add documentation to the ensure nested jacobian saying that it will assume your inputs for current/Psi are what you physically want to preserve

Right now this would have to involve negating `Psi` (and maybe also the current) #1710

desc/vmec.py
```diff
@@ -712,8 +736,29 @@ def save(  # noqa: C901
...
+        JB = (
+            surface_averages(
+                grid_full, JB, sqrt_g=data_full["sqrt(g)"], expand_out=False
+            )
+            / mu_0
+        )
+
+        jdotb[:] = JB
+        jdotb[0] = (
+            0  # NOTE: This does not actually match VMEC wouts,
```
dpanici, line 759: is this zero physically?

dpanici: image

Since we can compute at 0 and it is not NaN, means that the things it depends on have non-nan axis limits and so seems it exists, I don't really see why it should be zero on axis anyways as J itself is a current density, not a net current. There can exist current flowing along the axis, net current is just zero due to there being vanishing surface area there but this is not a net current.

PR #1630, opened 2025-03-09
Fix `test_data_index_deps` for Python 3.13

### LINKED ISSUE #1628: Breaking changes to tests in Python 3.13
Python 3.13 breaks how we use the `inspect` module. After some debugging, it appears in Python 3.13 only the source code of objects that are functions as classified by `inspect.isfunction` can be retrieved. In particular, JITed functions can no longer have their source code inspected.

One test that fails is ``tests/test_data_index.py``.
To resolve that failure, one solution is, in that file, to replace
```for _, fun in inspect.getmembers(module, callable):```
with 
```for _, fun in inspect.getmembers(module, inspect.isfunction):```

and also unjit all compute functions in `desc/_compute`. It would be nice to fix this without unjitting the compute functions though.

Otherwise, the error message that appears is
```python
    def findsource(object):
        """Return the entire source file and starting line number for an object.
    
        The argument may be a module, class, method, function, traceback, frame,
        or code object.  The source code is returned as a list of all the lines
        in the file and the line number indexes a line in that list.  An OSError
        is raised if the source code cannot be retrieved."""
    
        file = getsourcefile(object)
        if file:
            # Invalidate cache if needed.
            linecache.checkcache(file)
        else:
            file = getfile(object)
            # Allow filenames in form of "<something>" to pass through.
            # `doctest` monkeypatches `linecache` module to enable
            # inspection, so let `linecache.getlines` to be called.
            if (not (file.startswith('<') and file.endswith('>'))) or file.endswith('.fwork'):
                raise OSError('source code not available')
    
        module = getmodule(object, file)
        if module:
            lines = linecache.getlines(file, module.__dict__)
        else:
            lines = linecache.getlines(file)
        if not lines:
            raise OSError('could not get source code')
    
        if ismodule(object):
            return lines, 0
    
        if isclass(object):
            try:
                lnum = vars(object)['__firstlineno__'] - 1
            except (TypeError, KeyError):
>               raise OSError('source code not available')
E               OSError: source code not available


```

YigitElma:
Updated by @f0uriest 

There are two fixes here: 

- [466142e](https://github.com/PlasmaControl/DESC/pull/1630/commits/466142e3bbe0867852425c28f3e8220dd1cbdf32) uses a stricter filter when getting member functions to avoid picking up imported objects, see https://github.com/PlasmaControl/DESC/issues/1628#issuecomment-2711748669
- [fccd186](https://github.com/PlasmaControl/DESC/pull/1630/commits/fccd186bebe3f4a05a3a3a462e12dffb8ddc7fdb) Adds a helper function `getsource` to handle `functools.partial` objects. This isn't strictly necessary and isn't related to any changes in python 3.13, but also doesn't hurt. More explanation:
  - `inspect.getsource(fun)` works normally and always did
  - `inspect.getsource(jax.jit(fun))` also works fine and always did
  - `inspect.getsource(functools.partial(jax.jit, static_argnums=...)(fun))` also works fine and always did (note that the partial acts on jit, not on the function itself)
  - `inspect.getsource(functools.partial(fun, ...))` doesn't work (partial acting directly on fun), and never did. This doesn't occur in practice in our code right now but in theory may in the future, so this second change allows that to work correctly.

Resolves #1628


unalmis (dismissed):

f0uriest:
I think instead, we should make a little wrapper for `inspect.getsource`:

```python
def getsource(obj):
    if isinstance(obj, jaxlib.xla_extension.PjitFunction):
        return getsource(obj._fun)
    if isinstance(obj, functools.partial):
        return getsource(obj.func)
    return inspect.getsource(obj)

YigitElma (commented):
This is a lot cleaner than my fix, thanks! I have also realised that `inspect` was getting every function not just those that are defined in the file (like if you put a print in the inner loop, it was showing fori_loop etc), now it looks fine. I have also tested this fix on Python 3.13. Mostly, when test runs on specific CI group, it tends to stay in that group for other workflows, but this one is changing.

dpanici (approved):

YigitElma:
I would approve this PR.

PR #1643, opened 2025-03-13
Initial Guess for Non-Convex Shapes

### LINKED ISSUE #1384: Add `surface.get_axis` to get centroid of a surface
- [ ] https://github.com/PlasmaControl/DESC/blob/95612fa625bd5ba939569004c767ed02594586f0/desc/equilibrium/utils.py#L149

YigitElma:
For non-convex cross-sections, our initial guess function cannot form a set of nested surfaces. This PR changes the axis parser to compute the midpoint between theta=0 and theta=pi, and create an `FourierRZCurve` from those. 

Ideally, one could add the Shafranov shift with some heuristic on the pressure profile but for now, I won't add since for Stellarators, the shift is not as straightforward as tokamaks.

Resolves #1384


YigitElma:
Previous behavior with `precise_QA`,
```python
eq = get("precise_QA")
eq1 = Equilibrium(
        surface=eq.surface,
        pressure=eq.pressure,
        current=eq.current,
        Psi=eq.Psi,
        NFP=eq.NFP,
        L=eq.L,
        M=eq.M,
        N=eq.N,
        sym=eq.sym,
        spectral_indexing=eq.spectral_indexing,
        iota=eq.iota,
       ensure_nested=False,
    )
plot_surfaces(eq1, phi=3)
```
image

and new one,
image

dpanici:
I think this should go in without the shafranov shift. It is hard to pick a good heuristic for that for stellarators and as long as the surfaces are nested initially, the solver can figure out the rest.

unalmis (commented):
I think you should do either 

- compute the actual mean R and Z values by calling `line_integrals(R, line_label="theta")`, for data computed on the last flux surface. likewise for Z
- Or at least include theta=0,pi/2,pi,3pi/2 in your averaging.

You're reducing a three variable function R(rho, theta, zeta) to a one variable parametrization R(zeta), so basically averaging over 2D cross section, and 4 points would be better to define that plane

unalmis (approved):

rahulgaur104 (approved):

desc/geometry/surface.py
```diff
@@ -783,6 +783,43 @@ def fun_jax(zeta_hat, theta, zeta):
         else:
             return offset_surface
 
+    def get_axis(self):
+        """Get the axis of the surface.
```
f0uriest, line 787: should add some explanation that this is the geometric centroid (and define how its computed), not a magnetic axis.

desc/geometry/surface.py
```diff
@@ -783,6 +783,43 @@ def fun_jax(zeta_hat, theta, zeta):
...
+        Zin = Z[1::2]
+        Rmid = (Rout + Rin) / 2
+        Zmid = (Zout + Zin) / 2
+        phis = (
+            jnp.linspace(0, 2 * np.pi / self.NFP, len(Rmid), endpoint=False)
+            if self.N > 0
+            else jnp.zeros_like(Rmid)
+        )
+        # FourierRZCurve constructor will usually throw this warning, so ignore it
+        with warnings.catch_warnings():
```
f0uriest, line 816: what warning is this catching specifically?

Ideally we'd give inputs that don't throw a warning, or failing that should filter only for the specific warning we expect, not all `UserWarning`

desc/geometry/surface.py
```diff
@@ -1070,3 +1107,35 @@ def set_coeffs(self, l, m=0, R=None, Z=None):
             if ZZ is not None:
                 idxZ = self.Z_basis.get_idx(ll, mm, 0)
                 self.Z_lmn = put(self.Z_lmn, idxZ, ZZ)
+
+    def get_axis(self):
```
f0uriest, line 1111: for a ZernikeSection isn't the axis always just a cirlce at rho=0?

tests/test_configuration.py
```diff
@@ -429,8 +429,7 @@ def test_guess_from_coordinate_mapping(self):
         Zb = np.array([0.0, 0.16, -2])
         Z_modes = np.array([[-3, 0], [-2, 0], [-1, 0]])
         surf = FourierRZToroidalSurface(Rb, Zb, R_modes, Z_modes)
-        with pytest.warns(UserWarning):
```
f0uriest, line 432: is this because of the suppressed warnings?

tests/test_random.py
```diff
@@ -73,9 +72,6 @@ def test_random_surface():
     eq = Equilibrium(surface=surf)
     R0 = eq.compute("R0")["R0"]
     assert 5 <= R0 <= 10
-    Z0 = eq.compute("Z", grid=LinearGrid(rho=np.array([0]), M=0, N=8, NFP=eq.NFP))["Z"]
-    # should be centered around Z=0
-    np.testing.assert_allclose(np.mean(Z0), 0, atol=1e-14)
```
f0uriest, line 78: why remove this check?

YigitElma:
> I think you should do either
> 
> * compute the actual mean R and Z values by calling `line_integrals(R, line_label="theta")`, for data computed on the last flux surface. likewise for Z
> * Or at least include theta=0,pi/2,pi,3pi/2 in your averaging.
> 
> You're reducing a three variable function R(rho, theta, zeta) to a one variable parametrization R(zeta), so basically averaging over 2D cross section, and 4 points would be better to define that plane

Maybe I should use some other explanation in the docs. This is technically not the geometric center. For non-convex shapes, the geometric center might not be in the given shape. But since we want to form nested surfaces by this guess, it is better to choose an axis that lies inside the boundary. What I do is to compute midpoint of theta=0 and theta=pi. I feel like there should be a name for this, but I couldn't find any. Also, there could be a more proper implementation (like some mini-optimization), this was the easiest thing I could think of.

tests/test_random.py
YigitElma: maybe I am missing something, but why does this have to hold? When I plot the flux surface shapes, I cannot see any visual difference, but the test fails due to Z being slightly shifted, for this test 0.0012

desc/geometry/surface.py
YigitElma: Yeap

desc/geometry/surface.py
YigitElma: I think this was a bug in the `FourierRZCurve.from_values` method. It creates a grid with NFP 1 for the transform. I will fix it.

tests/test_configuration.py
YigitElma: No, these used to trigger `eq.solve` in the initialisation step because it was not nested. Now, the initial guess is nested without needing to solve.

YigitElma:
> Is it kind of like the "visual center" as described [here](https://blog.mapbox.com/a-new-algorithm-for-finding-a-visual-center-of-a-polygon-7c77e6492fbc)

Yes, this is an option. I also thought about finding the center of the biggest circle you can fit in, however this is another optimization problem you will solve for each eq initialisation (and for couple toroidal sections), and as long as you don't have a "U" like cross section where the concave region is in the middle of theta=0 and pi line, the changes in this PR should work.

@rahulgaur104 can you send one of your figure 8 like equilibria? I think this PR should work for them too, but to be sure, I would like to try.

tests/test_random.py
f0uriest: If it's stellarator symmetric it should have zero Z average

f0uriest: This might be because the axis is no longer just the M=0 modes of the boundary. You should check that your method still respects stellarator symmetry

f0uriest: wait this one isnt symmetric, I think this check was supposed to go with the surface above

f0uriest: Though I think even for the non-symmetric stuff it should still be centered around Z=0 in some sense. Without symmetry you can shift the whole device in Z without really changing anything but the way the random surface is created means it should still have <Z>~0

YigitElma: The axis Z position for the problematic one is 
image

YigitElma: And here I plot the Z values at different zeta cross sections.
image

```python
for zeta in np.linspace(0, 2 * np.pi, 20, endpoint=True):
    grid0 = LinearGrid(
        zeta=zeta, rho=1, theta=np.linspace(0, 2 * np.pi, 20, endpoint=True), NFP=eq.NFP
    )
    data_zeta0 = eq.compute(["R", "Z"], grid=grid0)
    if zeta == 0 or zeta == 2 * np.pi:
        plt.plot(data_zeta0["Z"], "-k", linewidth=5, label=f"Z zeta={zeta:.2f}")
    else:
        plt.plot(data_zeta0["Z"], label=f"Z zeta={zeta:.2f}")
    print(np.mean(data_zeta0["Z"]))
plt.plot(np.zeros_like(data_zeta0["R"]), "r--", label="0")
```

I have never used `random_surface` function but I think `get_axis` function works as expected

rahulgaur104 (approved):

desc/geometry/curve.py
```diff
@@ -292,7 +292,7 @@ def from_values(cls, coords, N=10, NFP=1, sym=False, basis="rpz", name=""):
             not np.all(np.diff(phi) > 0), ValueError, "Supplied phi must be monotonic"
         )
 
-        grid = LinearGrid(zeta=phi, NFP=1, sym=sym)
+        grid = LinearGrid(zeta=phi, NFP=NFP, sym=sym)
```
f0uriest, line 295: I forget, what happens if phi>2pi/NFP here? I vaguely recall there was a reason we had NFP=1.

desc/geometry/surface.py
f0uriest: In that case this can be simplified a lot, and I don't think the current implementation is even correct. Eg, if the cross section has large shafranov shift, then the average of R,Z at theta=0,pi will be in the center, but rho=0 won't be.

f0uriest (changes requested):
Can you split the docs improvements into another PR?

desc/geometry/curve.py
YigitElma: Maybe @dpanici knows?
This was throwing the warning that suppressed previously. The tests seems to be fine, but if there is a specific reason, I can move back to suppressing the specific warning.

desc/geometry/surface.py
YigitElma: Do you suggest computing the actual axis for this case and then use it to form a single term FourierRZCurve? I just kept this similar to the one above.

tests/test_random.py
YigitElma: This implementation might not be ideal for some cases but I think it still guarantees the initial guess for the axis to be inside the surface. Even now, I use theta=0,pi for zeta angles that give non-symmetric cross-sections, for example the `precise_qa` example I sent earlier. Yes the axis position is biased but as long as we don't have a cross-section like "U" where there is a gap between theta=0 and pi, the axis will reside inside the surface.

f0uriest (dismissed):

desc/geometry/curve.py
dpanici: So I think technically, we assume the curve input was closed, so I kept NFP=1 so that we dont have to deal with any weirdness from fitting the closed curve points (which have phi from 0 to 2pi) but now the points have actually been assigned to all lie within 0 to 2pi/NFP

I am a bit surprised all the tests pass so I guess it is nbd but I would really prefer if this was kept to NFP=NFP to avoid any edge cases where periodicity truncating causes issues

desc/geometry/surface.py
```diff
@@ -783,6 +783,49 @@ def fun_jax(zeta_hat, theta, zeta):
         else:
             return offset_surface
 
+    def get_axis(self):
+        """Get the axis of the surface.
+
+        This method calculates the axis of the surface by finding the mid point of the
+        R and Z values at theta=0 and pi degrees for a bunch of toroidal angles and
+        fitting a Fourier curve to it.
+        For general non-convex surfaces, the geometric center (aka centroid) might not
```
dpanici, line 792: your explanation seems to say we don't use the geometric centroid, but is this not something like the geometric centroid?

YigitElma: What I do here is find some axis that is roughly at the center (not centroid but visual center, I don't think there is a name for this, maybe center of biggest inner circle) and always inside the surface and also low computational burden.

desc/geometry/curve.py
```diff
@@ -292,7 +292,7 @@ def from_values(cls, coords, N=10, NFP=1, sym=False, basis="rpz", name=""):
...
         basis = FourierSeries(N=N, NFP=NFP, sym=sym)
         transform = Transform(grid, basis, build_pinv=True)
```
dpanici, line 297: let's just locally suppress the warning here

dpanici:
```suggestion
        with warnings.catch_warnings():
            warnings.simplefilter("ignore", category=UserWarning)
            transform = Transform(grid, basis, build_pinv=True)
```

desc/geometry/curve.py
```diff
@@ -294,7 +294,13 @@ def from_values(cls, coords, N=10, NFP=1, sym=False, basis="rpz", name=""):
 
         grid = LinearGrid(zeta=phi, NFP=1, sym=sym)
         basis = FourierSeries(N=N, NFP=NFP, sym=sym)
-        transform = Transform(grid, basis, build_pinv=True)
+        with warnings.catch_warnings():
+            # grid and basis have uneven NFP because we want to allow the user to either
+            # pass in an entire curve (in which case phi : 0-> 2pi) which has some
+            # field-periodicity, or only a portion of the curve
+            # (in which case phi: 0->2pi/NFP) and still have this fit work.
+            warnings.simplefilter("ignore", category=UserWarning)
```
YigitElma, line 302: I tried to filter based on the message but weirdly it doesn't work because the error message is `colored`

dpanici: I'd be in favor of removing the colored from that warning then, it is kinda random

dpanici (dismissed):

dpanici (approved):

f0uriest (approved):

rahulgaur104 (approved):

PR #1651, opened 2025-03-19
Fix Redl consistency with spline profiles

### LINKED ISSUE #1650: Redl current does not work with SplineProfile current
Rewriting the final function in _bootstrap.py as seen below should work, but it doesn't as it gives nan in the objective when solving for the equilibrium. In addition, if the spline has a knot at zero, it will result in a nan which will cause the cumulative_trapezoid to return all nans.
```python
if isinstance(profiles["current"],PowerSeriesProfile):
        degree = kwargs.get(
            "degree",
            min(
                (
                    profiles["current"].basis.L
                    if profiles["current"] is not None
                    else transforms["grid"].num_rho - 1
                ),
                transforms["grid"].num_rho - 1,
            ),
        )
        XX = jnp.vander(rho, degree + 1)[:, :-1]  # remove constant term
        c_l_r = jnp.pad(jnp.linalg.lstsq(XX, current_r)[0], (0, 1))  # manual polyfit
        c_l = jnp.polyint(c_l_r)
        current = jnp.polyval(c_l, rho)

    elif isinstance(profiles["current"], SplineProfile):
        current = cumulative_trapezoid(y=current_r, x=rho, initial=0.0)
```

dpanici:
- Adds error for incorrect grid with Redl, allow Redl to use SplineProfile, 
- Update Redl tutorial to include a spline profile optimization

Resolves #1650


dpanici:
We also should add a guard against incorrectly computing `current Redl` in `eq.compute` maybe, since I only added errors for the Objective

desc/objectives/_bootstrap.py
```diff
@@ -125,6 +133,18 @@ def build(self, use_jit=True, verbose=1):
...
+            np.any(np.isclose(rho, 0)),
+            ValueError,
+            "Redl formula is undefined" "at rho=0, but grid has grid points at rho=0",
+        )
+
+        errorif(
+            np.any(np.isclose(rho, 1)),
+            ValueError,
+            "Redl formula is undefined" "at rho=1, but grid has grid points at rho=1",
+        )
```
ddudt, line 147: The Redl formula is defined at $\rho=1$, but not where $p=0$ which is often the case at the boundary. We should change this to check if $n$ or $T$ vanish at any of the grids points. Or at the very least change this from an error to a warning.

tests/test_objective_funs.py
```diff
@@ -1824,6 +1824,27 @@ def test_things_per_objective_idx(self):
...
+        )
+        grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, L=2, axis=True)
+        obj = BootstrapRedlConsistency(eq, grid=grid)
+        with pytest.raises(ValueError, match="rho=0"):
+            obj.build()
+
+        grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, L=2, axis=False)
+        obj = BootstrapRedlConsistency(eq, grid=grid)
+        with pytest.raises(ValueError, match="rho=1"):
+            obj.build()
```
ddudt, line 1846: If you address my previous comment, this will also have to be updated.

ddudt (changes requested):

desc/objectives/_bootstrap.py
```diff
@@ -149,6 +163,35 @@ def build(self, use_jit=True, verbose=1):
...
         # wrong units. Densities should be ~ 10^20, temperatures are ~ 10^3.
         rho = eq.compute("rho", grid=grid)["rho"]
+        rhomax = np.max(rho)
+
+        # check last rho point, it may be at rho=1 and some profiles may go to zero
+        # there, if they are exactly zero this would cause NaNs since the profiles
+        # vanish.
+        errorif(
+            np.isclose(rhomax, 1.0)
+            and np.isclose(eq.electron_density(rhomax), 0.0, atol=1e-8),
```
f0uriest, line 173: Wouldn't it be better to just evaluate all the profiles at all the rho values to catch any other weirdness?

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+- Adds error for incorrect grids in Redl bootstrap current consistency objective
```
f0uriest, line 4: Should use the actual objective name and path in all these

desc/equilibrium/equilibrium.py
```diff
@@ -858,6 +858,42 @@ def compute(  # noqa: C901
             msg="must pass in a Grid object for argument grid!"
             f" instead got type {type(grid)}",
         )
+        # a check for Redl to prevent computing on-axis or
+        # at a rho=1.0 point where profiles vanish
+        if any(["Redl" in name for name in names]):
```
ddudt, line 863: I don't think this check is necessary or desired. We have plenty of other compute quantities that are sometimes undefined and we don't bother to check for them all here. `eq.compute` is mostly used for post-processing so a few NaN values for plotting etc. isn't problematic like it is in the optimization. We should let users compute the quantities they want and handle NaN values on their own, since that could be useful in some cases.

tests/test_bootstrap.py
```diff
@@ -1069,6 +1069,13 @@ def test_Redl_sfincs_QH(self):
 
         grid = LinearGrid(rho=rho, M=eq.M, N=eq.N, NFP=eq.NFP)
         data = eq.compute("<J*B> Redl", grid=grid, helicity=helicity)
+        grid2 = LinearGrid(rho=0.0, M=eq.M, N=eq.N, NFP=eq.NFP)
+        grid3 = LinearGrid(rho=1.0, M=eq.M, N=eq.N, NFP=eq.NFP)
+        with pytest.raises(ValueError, match="rho=0"):
+            eq.compute("<J*B> Redl", grid=grid2, helicity=helicity)
+        with pytest.raises(ValueError, match="vanish"):
+            eq.compute("<J*B> Redl", grid=grid3, helicity=helicity)
+
```
ddudt, line 1078: See previous comment.

ddudt (changes requested):

desc/equilibrium/equilibrium.py
dpanici: my rationale is that Redl is often used for the iterative method which bypasses all the checks we put in the `BootstrapRedlConsistency` objective, so I wanted to also have the checks in the compute method. 

I alternatively could make it a warning?

ddudt: Definitely don't throw an error, because otherwise you can't even call the compute method with that grid. 

I just feel like it is adding a lot of code for a really marginal use case. In practice I think the only reason to use the iterative method over an optimization is if you have a spline profile. And if people are copying the example from the tutorial, they should also be reading the explanation about what grid to use. 

If you really want to throw a warning in `eq.compute`, I would prefer something more general that covers all quantities. Like we could check if any values in `data` are NaN, and throw a warning saying something like "Some of the computed quantities are not defined, check the Grid you are computing on."

dpanici: I'll do a warning but still just for Redl. I think using a generic NaN catch will cause a lot of areas of the code to start throwing the warning which I don't want to deal with in this PR. Making an issue for it

desc/equilibrium/equilibrium.py
```diff
@@ -858,6 +858,46 @@ def compute(  # noqa: C901
...
+                "but grid has grid points at rho=0, note that on-axis"
+                "current will be NaN.",
+            )
+            rho = grid.compress(grid.nodes[:, 0], "rho")
+
+            # check if profiles may go to zero
+            # if they are exactly zero this would cause NaNs since the profiles
+            # vanish.
+            warnif(
+                np.any(np.isclose(self.electron_density(rho), 0.0, atol=1e-8)),
```
f0uriest, line 880: do we need atol here? I thought the issue was only if the profile is actually zero. If its small but finite I thought it should work? (though accuracy may be degraded)

f0uriest (approved):

desc/equilibrium/equilibrium.py
dpanici: True, I put atol here considering if at some point the density is lower than 1e-8, it is more likely that we are close to an actual zero of the profile, rather than just near a point of small but non-vanishing electron density.

This of course may be a poor assumption but until it becomes an issue later I am fine with keeping this as is.

desc/equilibrium/equilibrium.py
```diff
@@ -858,6 +858,46 @@ def compute(  # noqa: C901
...
+                UserWarning,
+                "Redl formula is undefined where kinetic profiles vanish, "
+                "but given electron density vanishes at at least one provided"
+                "rho grid point.",
+            )
+            Warning(
```
YigitElma, line 886: Is this a typo? `warnif` instead of `Warning`

desc/objectives/_bootstrap.py
```diff
@@ -125,6 +134,12 @@ def build(self, use_jit=True, verbose=1):
             ValueError,
             "Helicity toroidal mode number should be 0 (QA) or +/- NFP (QH)",
         )
+        rho = grid.nodes[grid.unique_rho_idx, 0]
+        errorif(
+            np.any(np.isclose(rho, 0)),
```
YigitElma, line 139: You might again want to specify `atol` here. With this, unless `rho<1e-7`, it will return False.

desc/objectives/_bootstrap.py
```diff
@@ -166,7 +205,9 @@ def build(self, use_jit=True, verbose=1):
             UserWarning,
             "Ion temperature is surprisingly high. It should have units of eV",
         )
-        # Profiles may go to 0 at rho=1, so exclude the last few grid points from lower
+
+        # Profiles may go to 0 at rho=1 (and we've already checked if our
+        #  grid has points there), so exclude the last few grid points from lower
```
YigitElma, line 210:
```suggestion
        # grid has points there), so exclude the last few grid points from lower
```

YigitElma (changes requested):

docs/notebooks/tutorials/bootstrap_current.ipynb
```diff
@@ -62,7 +62,7 @@
   },
```
YigitElma, line 62: <pre class="ql-syntax" spellcheck="false">Line #2.    # in our case here it it NaN at the axis and at rho=1.0
</pre><p>typo "it is NaN"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1651/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

YigitElma, line 62: <pre class="ql-syntax" spellcheck="false">Line #4.    # fig, ax = plot_1d(fam2[3], "current", linecolor="r", lw=2, label="3", ax=ax)
</pre><p>I think you can delete this?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/1651/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='43'/>

YigitElma:
I think for the failing lsq test, Rory had a fix in one of his PR's

YigitElma (approved):

f0uriest (approved):

YigitElma:
@ddudt Does this look fine now?

ddudt (approved):

PR #1656, opened 2025-03-25
Fix error in AD with default grid for splineXYZCoil

dpanici:
When providing `zeta` to a `LinearGrid` that is built on-the-fly, the resulting manipulations trigger some array conversions which don't play well with AD. This was happening in the default grid for `SplineXYZCoil` so I've changed that here.

I checked through the code for other instances where this may occur but I did not find any.

### DIFF
--- a/tests/test_objective_funs.py
+++ b/tests/test_objective_funs.py
@@ -3458,7 +3458,7 @@ class TestObjectiveNaNGrad:
     @pytest.mark.unit
     def test_objective_no_nangrad_quadratic_flux_minimizing(self):
         """SurfaceQuadraticFlux."""
-        ext_field = ToroidalMagneticField(1.0, 1.0)
+        ext_field = FourierXYZCoil().to_SplineXYZ(grid=3)
 
         surf = FourierRZToroidalSurface(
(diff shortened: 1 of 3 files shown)


unalmis (dismissed):
looks like knots is a jax array and we wanted grids to have numpy arrays to bake the nodes into jitted functions as constants, so this fix should work

f0uriest (approved):

tests/test_objective_funs.py
```diff
@@ -3458,7 +3458,7 @@ def test_objective_no_nangrad_quadratic_flux(self):
     @pytest.mark.unit
     def test_objective_no_nangrad_quadratic_flux_minimizing(self):
         """SurfaceQuadraticFlux."""
-        ext_field = ToroidalMagneticField(1.0, 1.0)
+        ext_field = FourierXYZCoil().to_SplineXYZ(grid=3)
```
YigitElma, line 3461: Does this supposed to create the problem? It runs fine on master

f0uriest (approved):

YigitElma (approved):

PR #1660, opened 2025-03-28
fix bug in`save_in_makegrid_format` for `CoilSet` objects with `sym=True` or `NFP>1`

### LINKED ISSUE #1659: CoilSet to makegrid fails for symmetries
The `CoilSet.save_in_makegrid_format()` function fails (has assertion error) for a base coilset which has NFP symmetry. I believe that this is due to the `flatten_coils()` subfunction within `save_in_makegrid_format` getting called with `self.coils` instead of `self`, which means that a list of coils is input instead of a `CoilSet` object. 

Minimum working example:
```
from desc.coils import CoilSet, FourierPlanarCoil

coil = FourierPlanarCoil()
coil_list = [coil, coil]

coilset = CoilSet(coil_list, NFP = 2, sym = True)

coilset.save_in_makegrid_format("test.coils", grid = 24)

```
I expect this to save a makegrid list of 8 coils (with some warnings about intersections), but instead I get:

  File "/home/rwu4/DESC_convex/makegrid_dummy.py", line 8, in <module>
    coilset.save_in_makegrid_format("test.coils", grid = 24)
  File "/home/rwu4/DESC/desc/coils.py", line 3009, in save_in_makegrid_format
    assert len(coils) == self.num_coils
AssertionError

dpanici:
Fix bug in `save_in_makegrid_format` for `CoilSet` objects with `sym=True`or `NFP>1` caused by calling `flatten_coils` not on self but on self.coils

Resolves #1659

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Changelog
 Bug Fixes
 - Fixes bug in ``desc.vmec.VMECIO.write_vmec_input`` for current-constrained equilibria, where DESC was incorrectly writing the ``s**0`` mode, where VMEC actually assumes it is zero and starts at the  ``s**1`` (which is different than the usual convention VMEC uses for its current profile when it uses the current derivative, where it starts with the ``s**0`` mode).
 - Fixes error that occurs when using the default grid for ``SplineXYZCoil`` in an optimization.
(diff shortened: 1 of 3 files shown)


ryanwu4:
Confirming that this fixes the issue in my test code as well.

YigitElma (approved):

f0uriest (approved):

PR #1662, opened 2025-03-29
Manually marking attributes as static

f0uriest:


desc/io/optimizable_io.py
```diff
@@ -137,19 +121,11 @@ def _generic_tree_flatten(obj):
...
             static_attrs = getattr(obj, "_static_attrs", [])
-            dynamic_attrs = getattr(obj, "_dynamic_attrs", [])
-            assert set(static_attrs).isdisjoint(set(dynamic_attrs))
 
             for key, val in obj.__dict__.items():
                 if key in static_attrs:
                     aux_data += [(key, _make_hashable(val))]
-                elif key in dynamic_attrs:
-                    children[key] = val
-                elif _unjittable(val):
```
YigitElma, line 151: Can we keep this as backup, in case someone forgets to add them manually.

f0uriest: I'd prefer not to. This has caused some weird bugs in the past and I think it's always preferable to be explicit about what is static and what isn't.

YigitElma: Ok, I also prefer that tbh. My only concern is for new users/devs, we need to add a warning regarding these kinds of problems (`tree_flatten`/`tree_unflatten`) so that people know where to fix it. This part is extremely hidden

tests/test_backend.py
```diff
@@ -114,27 +114,29 @@ def find_root_full(a):
 @pytest.mark.unit
 def test_lstsq():
     """Test cholesky factorization of least squares solution."""
+    rng = np.random.default_rng(seed=0)
```
f0uriest, line 117: This test ran fine for me locally and on della, but would fail on CI.

Because the seed wasn't fixed before, the particular values would depend on which other tests were run and in what order, etc so its hard to debug. Manually fixing the seed seems to fix the issue (likely the rng was just spitting out some weird values leading to ill-conditioned matrices)

tests/test_examples.py
```diff
@@ -2065,7 +2065,7 @@ def test_coil_arclength_optimization():
     np.testing.assert_allclose(
         coil_opt_with_arc_obj.compute("length")["length"], target_length, rtol=1e-4
     )
-    np.testing.assert_allclose(np.var(np.linalg.norm(xs1, axis=1)), 0, atol=1e-5)
+    np.testing.assert_allclose(np.var(np.linalg.norm(xs1, axis=1)), 0, atol=1e-3)
```
f0uriest, line 2068: Don't fully understand why this changed, my hunch is that changing the static attributes would slightly change the computation, leading to a different optimized result. Visually they look fine, and the other metrics are good so I think it's within our expected fudge factor for optimization tests

tests/test_objective_funs.py
```diff
@@ -1571,15 +1571,15 @@ def test_linking_current(self):
         obj = LinkingCurrentConsistency(eq, coilset1)
         obj.build()
         f = obj.compute(coilset1.params_dict, eq.params_dict)
-        np.testing.assert_allclose(f, 0)
+        np.testing.assert_allclose(f, 0, atol=1e-8)
```
f0uriest, line 1574: really should have had an atol here before, since default atol is 0 which is super fragile when the target value is also 0

f0uriest:
@unalmis can you take a look at the `Bounce{1,2}D` stuff? I didn't mark any attributes as static and things seem to work, I think because the `Bounce` objects are never passed across jit, they're only ever created and used within a jit context.

unalmis:
> @unalmis can you take a look at the `Bounce{1,2}D` stuff? I didn't mark any attributes as static and things seem to work, I think because the `Bounce` objects are never passed across jit, they're only ever created and used within a jit context.

I agree with this. If there's no disadvantage to it, should we add static attributes to those to?

YigitElma:
I don't know why but when I flatten an equilibrium with master and this one I get different results, especially this PR looks like flatten an array into integers and also store the array itself. For example,

```python
eq = get("precise_QA")
eq.change_resolution(L=4, M=4, N=4)
from jax.tree_util import tree_flatten, tree_unflatten

child, tree = tree_flatten(eq)
print(child)
```

This PR give,
```
[Array([ 0.0002, -0.0011, -0.0002, -0.0002,  0.0012, -0.0002, -0.0001,  0.0003, -0.0001, -0.0025, -0.0002,  0.0029, -0.0016, -0.0002,  0.0004, -0.0002,  0.0003,  0.0001, -0.0051,  0.0031,  0.005 ,  0.0018,  0.0021, -0.0031, -0.0001,  0.0008, -0.0002,  0.0834, -0.1387, -0.0015, -0.0134,  0.0066, -0.0006, -0.0003, -0.0002,  0.    ,  0.2845,  0.0529, -0.0026, -0.0019,  0.0016, -0.0008,  0.0152,  0.0018,  0.0058, -0.0065, -0.0017, -0.0006, -0.0119, -0.0104,  0.0013, -0.0013,  0.0002, -0.0003,  0.0053,  0.0012, -0.0011, -0.0016, -0.0001, -0.0004,  0.0002, -0.    ,  0.0001, -0.0004,  0.    , -0.0001], dtype=float64),
 np.float64(0.087),
 Array([-0.0004,  0.0002, -0.0001,  0.0004,  0.0001, -0.0001, -0.0035,  0.0016,  0.0002,  0.0008, -0.0002, -0.0001, -0.0286,  0.0036, -0.0008,  0.0017,  0.0002, -0.0001, -0.1704,  0.0123,  0.0008,  0.0013,  0.0011,  0.0001,  1.01  ,  0.1481, -0.0098,  0.0322, -0.0016,  0.0034, -0.0002, -0.0001, -0.0004,  0.1872, -0.0736, -0.0096,  0.0231, -0.001 ,  0.0029,  0.0002, -0.0008,  0.0018,  0.0164, -0.0091, -0.0064,  0.0056,  0.0009, -0.0012,  0.0003,  0.0001,  0.0003,  0.0008, -0.0016, -0.0016,  0.0024,  0.0006,  0.0002,  0.0002, -0.0002, -0.0003,  0.0001, -0.0003, -0.0002,  0.0002,  0.0003, -0.0001,  0.    , -0.0001,  0.    ], dtype=float64),
 Array([ 0.    ,  0.0002,  0.0002, -0.0002, -0.0002,  0.0001, -0.0001, -0.    ,  0.    , -0.0009,  0.0019,  0.0014, -0.0018, -0.001 ,  0.0002, -0.0001,  0.    ,  0.0001, -0.0145,  0.0116,  0.006 , -0.0075, -0.0014,  0.0007, -0.0003, -0.0002,  0.0001, -0.1353,  0.0376,  0.0163,  0.0005, -0.0002,  0.0005, -0.0005, -0.0004, -0.0006, -0.2447, -0.0274, -0.0077,  0.0015, -0.0003,  0.0008, -0.1395, -0.0045, -0.0017,  0.0023, -0.    ,  0.0008, -0.0291,  0.006 , -0.0002,  0.002 ,  0.    ,  0.0003, -0.0036,  0.001 , -0.0001,  0.0011,  0.    ,  0.0001, -0.0003,  0.0001, -0.0001,  0.0003, -0.    ,  0.    ], dtype=float64),
 Array([1.0196, 0.1971, 0.0232, 0.0025, 0.0002], dtype=float64),
 Array([-0.0003, -0.0025, -0.0207, -0.1521], dtype=float64),
 array([1., 0., 0., 0., 1., 0., 0., 0., 1.]),
 array([0., 0., 0.]),
 0,
 1,
 2,
 3,
 4,
 array([0, 1, 2, 3, 4]),
 array([0, 1, 2, 3, 4]),
 array([0, 0, 0, 0, 0]),
 array([0, 0, 0, 0, 0]),
 array([[0, 0, 0],
        [2, 0, 0],
        [4, 0, 0],
        [6, 0, 0],
        [8, 0, 0]]),
 array([0, 1, 2, 3, 4]),
 array([0, 1, 2, 3, 4]),
 array([0]),
 array([0]),
 Array([0., 0., 0., 0., 0.], dtype=float64),
 0,
 1,
 2,
 3,
 4,
 array([0, 1, 2, 3, 4]),
 array([0, 1, 2, 3, 4]),
 array([0, 0, 0, 0, 0]),
 array([0, 0, 0, 0, 0]),
 array([[0, 0, 0],
        [2, 0, 0],
        [4, 0, 0],
        [6, 0, 0],
        [8, 0, 0]]),
 array([0, 1, 2, 3, 4]),
 array([0, 1, 2, 3, 4]),
 array([0]),
 array([0]),
 Array([0., 0., 0., 0., 0.], dtype=float64),
 Array([ 0.0001, -0.0001,  0.0001, -0.    , -0.0002,  0.0002,  0.0015, -0.0028,  0.0002, -0.0008,  0.0035, -0.0269,  0.0011,  0.0007,  0.0124, -0.1691,  1.    ,  0.1466,  0.0322,  0.0034, -0.0004,  0.1778, -0.0744,  0.0223,  0.0027,  0.0018,  0.0103, -0.0082,  0.0057, -0.0013,  0.0003, -0.0006, -0.001 ,  0.0022,  0.0001, -0.0003, -0.0001, -0.    ,  0.0001, -0.0001,  0.    ], dtype=float64),
 Array([ 0.0001,  0.    , -0.0002,  0.0001,  0.    ,  0.0004,  0.0009, -0.0017,  0.0002,  0.0001, -0.0088,  0.0102, -0.0077,  0.0007,  0.0001, -0.1195,  0.0374,  0.0001,  0.0006, -0.0006, -0.0003, -0.0076, -0.0266, -0.2431,  0.    , -0.0016, -0.0037, -0.1373,  0.0001, -0.0001,  0.0063, -0.0272,  0.    , -0.0001,  0.0011, -0.0025, -0.    , -0.0001,  0.0001, -0.    ], dtype=float64),
 np.int64(1)]
```

And MASTER gives,
```
[Array([ 0.0002, -0.0011, -0.0002, -0.0002,  0.0012, -0.0002, -0.0001,  0.0003, -0.0001, -0.0025, -0.0002,  0.0029, -0.0016, -0.0002,  0.0004, -0.0002,  0.0003,  0.0001, -0.0051,  0.0031,  0.005 ,  0.0018,  0.0021, -0.0031, -0.0001,  0.0008, -0.0002,  0.0834, -0.1387, -0.0015, -0.0134,  0.0066, -0.0006, -0.0003, -0.0002,  0.    ,  0.2845,  0.0529, -0.0026, -0.0019,  0.0016, -0.0008,  0.0152,  0.0018,  0.0058, -0.0065, -0.0017, -0.0006, -0.0119, -0.0104,  0.0013, -0.0013,  0.0002, -0.0003,  0.0053,  0.0012, -0.0011, -0.0016, -0.0001, -0.0004,  0.0002, -0.    ,  0.0001, -0.0004,  0.    , -0.0001], dtype=float64),
 np.float64(0.087),
 Array([-0.0004,  0.0002, -0.0001,  0.0004,  0.0001, -0.0001, -0.0035,  0.0016,  0.0002,  0.0008, -0.0002, -0.0001, -0.0286,  0.0036, -0.0008,  0.0017,  0.0002, -0.0001, -0.1704,  0.0123,  0.0008,  0.0013,  0.0011,  0.0001,  1.01  ,  0.1481, -0.0098,  0.0322, -0.0016,  0.0034, -0.0002, -0.0001, -0.0004,  0.1872, -0.0736, -0.0096,  0.0231, -0.001 ,  0.0029,  0.0002, -0.0008,  0.0018,  0.0164, -0.0091, -0.0064,  0.0056,  0.0009, -0.0012,  0.0003,  0.0001,  0.0003,  0.0008, -0.0016, -0.0016,  0.0024,  0.0006,  0.0002,  0.0002, -0.0002, -0.0003,  0.0001, -0.0003, -0.0002,  0.0002,  0.0003, -0.0001,  0.    , -0.0001,  0.    ], dtype=float64),
 Array([ 0.    ,  0.0002,  0.0002, -0.0002, -0.0002,  0.0001, -0.0001, -0.    ,  0.    , -0.0009,  0.0019,  0.0014, -0.0018, -0.001 ,  0.0002, -0.0001,  0.    ,  0.0001, -0.0145,  0.0116,  0.006 , -0.0075, -0.0014,  0.0007, -0.0003, -0.0002,  0.0001, -0.1353,  0.0376,  0.0163,  0.0005, -0.0002,  0.0005, -0.0005, -0.0004, -0.0006, -0.2447, -0.0274, -0.0077,  0.0015, -0.0003,  0.0008, -0.1395, -0.0045, -0.0017,  0.0023, -0.    ,  0.0008, -0.0291,  0.006 , -0.0002,  0.002 ,  0.    ,  0.0003, -0.0036,  0.001 , -0.0001,  0.0011,  0.    ,  0.0001, -0.0003,  0.0001, -0.0001,  0.0003, -0.    ,  0.    ], dtype=float64),
 Array([1.0196, 0.1971, 0.0232, 0.0025, 0.0002], dtype=float64),
 Array([-0.0003, -0.0025, -0.0207, -0.1521], dtype=float64),
 array([1., 0., 0., 0., 1., 0., 0., 0., 1.]),
 array([0., 0., 0.]),
 array([0, 1, 2, 3, 4]),
 array([0, 1, 2, 3, 4]),
 array([0, 0, 0, 0, 0]),
 array([0, 0, 0, 0, 0]),
 array([[0, 0, 0],
        [2, 0, 0],
        [4, 0, 0],
        [6, 0, 0],
        [8, 0, 0]]),
 array([0, 1, 2, 3, 4]),
 array([0, 1, 2, 3, 4]),
 array([0]),
 array([0]),
 Array([0., 0., 0., 0., 0.], dtype=float64),
 array([0, 1, 2, 3, 4]),
 array([0, 1, 2, 3, 4]),
 array([0, 0, 0, 0, 0]),
 array([0, 0, 0, 0, 0]),
 array([[0, 0, 0],
        [2, 0, 0],
        [4, 0, 0],
        [6, 0, 0],
        [8, 0, 0]]),
 array([0, 1, 2, 3, 4]),
 array([0, 1, 2, 3, 4]),
 array([0]),
 array([0]),
 Array([0., 0., 0., 0., 0.], dtype=float64),
 Array([ 0.0001, -0.0001,  0.0001, -0.    , -0.0002,  0.0002,  0.0015, -0.0028,  0.0002, -0.0008,  0.0035, -0.0269,  0.0011,  0.0007,  0.0124, -0.1691,  1.    ,  0.1466,  0.0322,  0.0034, -0.0004,  0.1778, -0.0744,  0.0223,  0.0027,  0.0018,  0.0103, -0.0082,  0.0057, -0.0013,  0.0003, -0.0006, -0.001 ,  0.0022,  0.0001, -0.0003, -0.0001, -0.    ,  0.0001, -0.0001,  0.    ], dtype=float64),
 Array([ 0.0001,  0.    , -0.0002,  0.0001,  0.    ,  0.0004,  0.0009, -0.0017,  0.0002,  0.0001, -0.0088,  0.0102, -0.0077,  0.0007,  0.0001, -0.1195,  0.0374,  0.0001,  0.0006, -0.0006, -0.0003, -0.0076, -0.0266, -0.2431,  0.    , -0.0016, -0.0037, -0.1373,  0.0001, -0.0001,  0.0063, -0.0272,  0.    , -0.0001,  0.0011, -0.0025, -0.    , -0.0001,  0.0001, -0.    ], dtype=float64)]
```

I don't understand specifically this part
```
 0,
 1,
 2,
 3,
 4,
 array([0, 1, 2, 3, 4]),
```

desc/basis.py
```diff
@@ -36,6 +36,17 @@ class _Basis(IOAble, ABC):
         "_spectral_indexing",
     ]
 
+    _static_attrs = [
```
dpanici, line 39: why do we not want `modes` also as a static_attr here? and the rest of the arrays that we know should be integers and not traced?

desc/equilibrium/equilibrium.py
```diff
@@ -163,7 +163,23 @@ class Equilibrium(IOAble, Optimizable):
...
+        "_Z_sym",
+        "_NFP",
+        "_L",
+        "_M",
+        "_N",
+        "_L_grid",
+        "_M_grid",
+        "_N_grid",
+        "_spectral_indexing",
+        "_bdry_mode",
```
dpanici, line 181:
```suggestion
        "_bdry_mode",
        "_resolution",
```

desc/geometry/core.py
```diff
@@ -359,6 +360,7 @@ class Surface(IOAble, Optimizable, ABC):
     """Abstract base class for 2d surfaces in 3d space."""
 
     _io_attrs_ = ["_name", "_sym", "_L", "_M", "_N"]
+    _static_attrs = Optimizable._static_attrs + ["_name", "_sym", "_L", "_M", "_N"]
```
dpanici, line 363:
```suggestion
    _static_attrs = Optimizable._static_attrs + ["_name", "_sym", "_L", "_M", "_N", "_R_basis","_Z_basis"]
```

dpanici: nvm you have this already in the child classes

desc/grid.py
```diff
@@ -48,6 +48,21 @@ class _Grid(IOAble, ABC):
...
+        "_coordinates",
+        "_fft_poloidal",
+        "_fft_toroidal",
+        "_is_meshgrid",
+        "_L",
+        "_M",
+        "_N",
+        "_NFP",
+        "_node_pattern",
+        "_sym",
```
dpanici, line 63: why would we not want `_period` also here?

desc/integrals/singularities.py
```diff
@@ -316,6 +316,8 @@ class _BIESTInterpolator(IOAble, ABC):
         "_shift_z",
     ]
 
+    _static_attrs = ["_q"]
```
dpanici, line 319: why not also have st sz here?

desc/objectives/_omnigenity.py
```diff
@@ -549,6 +549,13 @@ class Omnigenity(_Objective):
         target_default="``target=0``.", bounds_default="``target=0``."
     )
 
+    _static_attrs = _Objective._static_attrs + [
```
dpanici, line 552: `_helicity` should be static, I think, for the QS objectives in this fiile right? tuple of ints?

desc/transform.py
```diff
@@ -43,7 +43,16 @@ class Transform(IOAble):
...
     _io_attrs_ = ["_grid", "_basis", "_derivatives", "_rcond", "_method"]
-    _static_attrs = ["_derivatives"]
+    _static_attrs = [
+        "_derivatives",
+        "_method",
+        "_rcond",
+        "_built",
+        "_built_pinv",
+        "num_n_modes",
+        "num_lm_modes",
```
dpanici, line 53: why must these be static but `num_zeta` etc are not static for `grid`? those are defined as the size of an array, which is also what `num_lm_modes` is defined as

dpanici (changes requested):
left some questions and I think the QS objectives are missing the `helicity` attribute as static

desc/basis.py
YigitElma: I tried this previously in #1582 but it caused some stuff to fail. I think the changes introduced in #1405 . Specifically, setting all of the modes static breaks `modes=modes[idx,:]` kind of operations.

YigitElma: I think this is possible though, with some changes to that part of the code.

desc/objectives/_generic.py
```diff
@@ -78,7 +78,12 @@ def myfun(eq, path=""):
 
     _units = "(Unknown)"
     _print_value_fmt = "External objective value: "
-    _static_attrs = ["_fun_wrapped", "_fun_kwargs"]
+    _static_attrs = _Objective._static_attrs + [
+        "_fun",
```
YigitElma, line 82: `_abs_step`, `_rel_step`?

desc/objectives/linear_objectives.py
```diff
@@ -3126,7 +3126,7 @@ class FixNearAxisR(_FixedObjective):
 
     """
 
-    _static_attrs = ["_nae_eq"]
+    _static_attrs = _Objective._static_attrs + ["_nae_eq"]
```
YigitElma, line 3129: `_order`?

desc/objectives/linear_objectives.py
```diff
@@ -3277,7 +3277,7 @@ class FixNearAxisZ(_FixedObjective):
...
```
YigitElma, line 3280: `_order`

desc/objectives/linear_objectives.py
```diff
@@ -3434,7 +3434,7 @@ class FixNearAxisLambda(_FixedObjective):
...
```
YigitElma, line 3437: `_order`

desc/optimizable.py
```diff
@@ -16,6 +16,8 @@ class Optimizable(ABC):
     ``optimizable_parameter`` decorator.
     """
 
+    _static_attrs = ["_optimizable_params"]
```
YigitElma, line 19: maybe `dimensions`, `x_idx` and `_dim_x`

desc/profiles.py
```diff
@@ -36,6 +36,7 @@ class _Profile(IOAble, ABC):
     """
```
YigitElma, line 36: Remaining Profile classes can also have some static attributes. For example, `PowerSeriesProfile` can have `_sym`, `_basis`. I don't remember if I tested `_basis` before, but it might work.

desc/transform.py
```diff
@@ -43,7 +43,16 @@ class Transform(IOAble):
     """
 
     _io_attrs_ = ["_grid", "_basis", "_derivatives", "_rcond", "_method"]
-    _static_attrs = ["_derivatives"]
+    _static_attrs = [
```
YigitElma, line 46: can we also add `_basis`?

YigitElma (changes requested):

YigitElma:
> I think this PR needs to be in to fix the `test_lstsq` test thats failing on master rn (sorry I definitely forgot that that was something we had to wait for)

This fail every now and then, no problem. We can merge the fix with dependency ones. This PR needs couple more changes.

YigitElma:
I will approve this but before that I want to run some profiling just to be sure.

desc/grid.py
f0uriest: I don't think we ever do any conditionals based on `period` so there's no need for it to be static

desc/integrals/singularities.py
f0uriest: technically `st`/`sz` don't need to be static since they're not used for array sizes like `q`. In theory they don't even need to be integers

desc/transform.py
f0uriest: `num_zeta` isn't actually stored as an attribute of `Grid`, its just a property that dynamically grabs `grid.unique_zeta_idx.size`, and `unique_zeta_idx` is marked static

desc/geometry/curve.py
```diff
@@ -895,6 +895,8 @@ class SplineXYZCurve(Curve):
 
     _io_attrs_ = Curve._io_attrs_ + ["_X", "_Y", "_Z", "_knots", "_method"]
 
+    _static_attrs = Curve._static_attrs + ["_method"]
```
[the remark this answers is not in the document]
f0uriest: usually not but in theory it could at some point. And making jax/np arrays static can cause bugs so I'm trying not to mark any arrays as static unless we have very good reason to.

desc/geometry/surface.py
```diff
@@ -65,7 +65,7 @@ class FourierRZToroidalSurface(Surface):
         "_NFP",
         "_rho",
     ]
-    _static_attrs = ["_R_basis", "_Z_basis"]
+    _static_attrs = Surface._static_attrs + ["_R_basis", "_Z_basis", "_NFP"]
```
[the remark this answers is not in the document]
f0uriest: I don't think we have any conditionals that use rho so there shouldn't be any need to make it static

desc/geometry/surface.py
```diff
@@ -838,7 +838,11 @@ class ZernikeRZToroidalSection(Surface):
         "_zeta",
     ]
 
-    _static_attrs = ["_R_basis", "_Z_basis"]
+    _static_attrs = Surface._static_attrs + [
+        "_R_basis",
+        "_Z_basis",
+        "_spectral_indexing",
```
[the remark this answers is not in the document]
f0uriest: I don't think we have any conditionals that use zeta so there shouldn't be any need to make it static

desc/objectives/_generic.py
f0uriest: There's no conditionals on them so I don't think they need to be static

desc/optimizable.py
f0uriest: those aren't actually attributes, they're properties that get calculated dynamically so there's nothing to mark as static

desc/profiles.py
f0uriest: `sym` isn't stored as an attribute, its just passed through when creating basis, and its already marked as static within the basis class

YigitElma:
> ### Memory benchmark result
> ```diff
> |               Test Name                |      %Δ      |    Master (MB)     |      PR (MB)       |    Δ (MB)    |    Time PR (s)     |  Time Master (s)   |
> | -------------------------------------- | ------------ | ------------------ | ------------------ | ------------ | ------------------ | ------------------ |
>   test_objective_jac_w7x                 |   -1.67 %    |     3.953e+03      |     3.887e+03      |    -65.88    |       36.30        |       30.04        |
>   test_proximal_jac_w7x_with_eq_update   |    1.98 %    |     6.682e+03      |     6.815e+03      |    132.12    |       174.81       |       163.13       |
>   test_proximal_freeb_jac                |   -0.02 %    |     1.326e+04      |     1.326e+04      |    -2.48     |       75.71        |       73.89        |
>   test_proximal_freeb_jac_blocked        |    0.32 %    |     7.516e+03      |     7.540e+03      |    24.34     |       68.20        |       64.97        |
>   test_proximal_freeb_jac_batched        |    0.23 %    |     7.522e+03      |     7.540e+03      |    17.56     |       69.60        |       65.77        |
>   test_proximal_jac_ripple               |    1.66 %    |     7.471e+03      |     7.595e+03      |    124.26    |       74.53        |       71.21        |
>   test_proximal_jac_ripple_spline        |    2.10 %    |     3.494e+03      |     3.567e+03      |    73.29     |       78.25        |       75.34        |
>   test_eq_solve                          |    1.46 %    |     2.020e+03      |     2.049e+03      |    29.42     |       124.86       |       124.08       |
> ```
> 
> For the memory plots, go to the summary of `Memory Benchmarks` workflow and download the artifact.

I think the memory problem is resolved with new JAX release (or something else on DESC fixed this)... It might have been similar to issue seen in #1701

YigitElma:
As it's current state, this PR doesn't solve the dead region issue. Here is the profile for eq solve on GPU,
image

Same thing on master,
image

YigitElma:
@f0uriest Have you ever considered using https://docs.kidger.site/equinox/api/module/module/#equinox.Module?

f0uriest:
> @f0uriest Have you ever considered using https://docs.kidger.site/equinox/api/module/module/#equinox.Module?

I've used them in several other projects. But equinox modules are read-only dataclasses, ie you can't update them after creation, you have to create a new one instead, which would require us to rewrite a huge chunk of the codebase and change the API a lot.

YigitElma:
Maybe we can implement a similar flatten/unflatten method. Just mentioning

desc/equilibrium/equilibrium.py
```diff
@@ -163,7 +163,23 @@ class Equilibrium(IOAble, Optimizable):
         "_M_grid",
         "_N_grid",
     ]
-    _static_attrs = ["_R_basis", "_Z_basis", "_L_basis"]
+    _static_attrs = Optimizable._static_attrs + [
+        "_R_basis",
```
f0uriest: I don't think we want to be marking the entire basis as static. When you mark something as static, then it will trigger recompilation whenever the objects hash changes, and for arbitrary objects like this, the hash is not constant. ie
```python
basis1 = FourierZernikeBasis(4,4,4)
basis2 = FourierZernikeBasis(4,4,4)
print(hash(basis1))
print(hash(basis2))
# 8175726823468
# 8175726754217
```

compare to marking string/int/bool as static:
```python
print(hash(12))
print(hash(12))
# 12
# 12
print(hash("cos"))
print(hash("cos"))
-5689838175960262998
-5689838175960262998
print(hash(tuple((1,2,3))))
print(hash(tuple((1,2,3))))
# 529344067295497451
# 529344067295497451
```

Basically this means that 2 equilibria with the exact same resolution etc will be treated as different objects, which isn't what we want, especially with the persistent compilation cache turned on.

desc/basis.py
f0uriest: I think generally we should err on the side of not marking things as static unless they absolutely have to be. The more things we mark as static the greater the chance that we'll end up triggering recompilation when we don't mean to, as well as possibly slowing down the flatten/unflatten time and leading to longer compile times because it tries to over optimize

desc/transform.py
f0uriest: I don't think we want to: https://github.com/PlasmaControl/DESC/pull/1662#discussion_r2146325405

desc/equilibrium/equilibrium.py
YigitElma: We can create a custom `__hash__` method for the basis classes. Basically if the resolution, symmetry option and some other simple stuff is the same, the hash must be the same, right? This might also prevent looking at the whole pytree to check if some argument has the same tree structure as a already compiled executable

YigitElma: For the `_Basis` class, for example,
```python
def __hash__(self):
        """Get the hash of the object."""
        return hash(
            (
                self.__class__.__name__,
                self.L,
                self.M,
                self.N,
                self.NFP,
                self.sym,
                self.spectral_indexing,
            )
        )
```

YigitElma:
Not sure about this exactly but can we remove `things` for objective classes? We mostly use it for some resolution stuff like creating the default grid etc, we might not need to store the whole object. In terms of `optimizer.optimize` this should be relatively easy, we already pass `things` to it, all we need is the proper order of `things`. Also, objective functions don't call `thing.compute` so once we got the transforms, profiles etc we don't need the object. 

For example, to give some reasoning for this change,
```python
eq = get("precise_QA")
obj.build(verbose=0)
from jax.tree_util import tree_flatten
vals, treedef = tree_flatten(obj)
print(treedef.num_leaves, treedef.num_nodes)
print(treedef)
```
Prints (I have some other local changes for the output but idea is same)
```
122 241
PyTreeDef(CustomNode(ObjectiveFunction[[('_jac_chunk_size', 3282), ('_use_jit', True), ('_deriv_mode', 'batched'), ('_built', 
True), ('_compiled', False), ('_name', 'ObjectiveFunction'), ('_dim_f', 5346), ('_scalar', False), ('_things_per_objective_idx', 
[[0]])]], [{'_flatten': CustomNode(_ThingFlattener[[('length', 1), ('treedef', PyTreeDef([[*]]))]], [{}]), '_objectives': 
(CustomNode(ForceBalance[[('_target', ('ndarray', (), (np.int64(0),))), ('_weight', ('ndarray', (), (np.int64(1),))), ('_normalize', 
True), ('_normalize_target', True), ('_deriv_mode', 'fwd'), ('_name', 'force'), ('_use_jit', True), ('_built', True), ('_dim_f', 5346), 
('_data_keys', ['F_rho', '|grad(rho)|', 'sqrt(g)', 'F_helical', '|e^helical*sqrt(g)|'])]], [{'_bounds': None, '_constants': {'profiles': 
{'atomic_number': None, 'current': CustomNode(PowerSeriesProfile[[('_name', 'current'), ('_basis', PowerSeries at 
0x72b9abb55ee0 (L=8, M=0, N=0, NFP=1, sym=even, spectral_indexing=linear))]], [{'_params': *}]), 'electron_density': None, 
'electron_temperature': None, 'ion_temperature': None, 'iota': None, 'pressure': CustomNode(PowerSeriesProfile[[('_name', 
'pressure'), ('_basis', PowerSeries at 0x72b9aaae4e60 (L=8, M=0, N=0, NFP=1, sym=even, spectral_indexing=linear))]], 
[{'_params': *}])}, 'quad_weights': *, 'transforms': {'L': CustomNode(Transform[[('_basis', FourierZernikeBasis at 0x72b99f611a60 
(L=8, M=8, N=8, NFP=2, sym=sin, spectral_indexing=ansi)), ('_rcond', 'auto'), ('_built', True), ('_built_pinv', False), 
('_derivatives', ('ndarray', (10, 3), (np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(0), np.int64(0), np.int64(2), 
np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2), 
np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(1), np.int64(0), np.int64(1), np.int64(1), 
np.int64(0), np.int64(0), np.int64(2)))), ('_method', 'fft'), ('num_lm_modes', 45), ('num_n_modes', 17), ('pad_dim', 16)]], [{'_grid': 
CustomNode(ConcentricGrid[[('_L', 16), ('_M', 16), ('_N', 16), ('_NFP', 2), ('_sym', ('ndarray', (), (np.True_,))), ('_node_pattern', 
'jacobi'), ('_coordinates', 'rtz'), ('_is_meshgrid', False)]], [{'_axis': *, '_inverse_poloidal_idx': *, '_inverse_rho_idx': *, 
'_inverse_zeta_idx': *, '_nodes': *, '_period': (*, *, *), '_spacing': *, '_unique_poloidal_idx': *, '_unique_rho_idx': *, 
'_unique_zeta_idx': *, '_weights': *}]), '_matrices': {'fft': {0: {0: *, 1: *, 2: *}, 1: {0: *, 1: *, 2: {}}, 2: {0: *, 1: {}, 2: {}}}}, 'dk': *, 'fft_grid': CustomNode(Grid[[('_NFP', 1), ('_sym', False), ('_node_pattern', 'custom'), ('_coordinates', 'rtz'), ('_is_meshgrid', False), ('_L', 
0), ('_M', 0), ('_N', 0)]], [{'_axis': *, '_nodes': *, '_period': (*, *, *), '_source_grid': None, '_spacing': None, '_weights': None}]), 
'fft_index': *, 'lm_modes': *}]), 'R': CustomNode(Transform[[('_basis', FourierZernikeBasis at 0x72b99f611010 (L=8, M=8, N=8, 
NFP=2, sym=cos, spectral_indexing=ansi)), ('_rcond', 'auto'), ('_built', True), ('_built_pinv', False), ('_derivatives', ('ndarray', (10, 
3), (np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(0), np.int64(0), np.int64(2), np.int64(0), np.int64(0), np.int64(0), 
np.int64(1), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2), np.int64(0), np.int64(0), np.int64(0), 
np.int64(1), np.int64(1), np.int64(0), np.int64(1), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2)))), 
('_method', 'fft'), ('num_lm_modes', 45), ('num_n_modes', 17), ('pad_dim', 16)]], [{'_grid': CustomNode(ConcentricGrid[[('_L', 
16), ('_M', 16), ('_N', 16), ('_NFP', 2), ('_sym', ('ndarray', (), (np.True_,))), ('_node_pattern', 'jacobi'), ('_coordinates', 'rtz'), 
('_is_meshgrid', False)]], [{'_axis': *, '_inverse_poloidal_idx': *, '_inverse_rho_idx': *, '_inverse_zeta_idx': *, '_nodes': *, '_period':
 (*, *, *), '_spacing': *, '_unique_poloidal_idx': *, '_unique_rho_idx': *, '_unique_zeta_idx': *, '_weights': *}]), '_matrices': {'fft': {0: 
{0: *, 1: *, 2: *}, 1: {0: *, 1: *, 2: {}}, 2: {0: *, 1: {}, 2: {}}}}, 'dk': *, 'fft_grid': CustomNode(Grid[[('_NFP', 1), ('_sym', False), 
('_node_pattern', 'custom'), ('_coordinates', 'rtz'), ('_is_meshgrid', False), ('_L', 0), ('_M', 0), ('_N', 0)]], [{'_axis': *, '_nodes': *, 
'_period': (*, *, *), '_source_grid': None, '_spacing': None, '_weights': None}]), 'fft_index': *, 'lm_modes': *}]), 'Z': 
CustomNode(Transform[[('_basis', FourierZernikeBasis at 0x72b99f611a60 (L=8, M=8, N=8, NFP=2, sym=sin, 
spectral_indexing=ansi)), ('_rcond', 'auto'), ('_built', True), ('_built_pinv', False), ('_derivatives', ('ndarray', (10, 3), (np.int64(0), 
np.int64(0), np.int64(0), np.int64(1), np.int64(0), np.int64(0), np.int64(2), np.int64(0), np.int64(0), np.int64(0), np.int64(1), 
np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2), np.int64(0), np.int64(0), np.int64(0), np.int64(1), 
np.int64(1), np.int64(0), np.int64(1), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2)))), ('_method', 'fft'), 
('num_lm_modes', 45), ('num_n_modes', 17), ('pad_dim', 16)]], [{'_grid': CustomNode(ConcentricGrid[[('_L', 16), ('_M', 16), 
('_N', 16), ('_NFP', 2), ('_sym', ('ndarray', (), (np.True_,))), ('_node_pattern', 'jacobi'), ('_coordinates', 'rtz'), ('_is_meshgrid', 
False)]], [{'_axis': *, '_inverse_poloidal_idx': *, '_inverse_rho_idx': *, '_inverse_zeta_idx': *, '_nodes': *, '_period': (*, *, *), 
'_spacing': *, '_unique_poloidal_idx': *, '_unique_rho_idx': *, '_unique_zeta_idx': *, '_weights': *}]), '_matrices': {'fft': {0: {0: *, 1: *, 
2: *}, 1: {0: *, 1: *, 2: {}}, 2: {0: *, 1: {}, 2: {}}}}, 'dk': *, 'fft_grid': CustomNode(Grid[[('_NFP', 1), ('_sym', False), ('_node_pattern', 
'custom'), ('_coordinates', 'rtz'), ('_is_meshgrid', False), ('_L', 0), ('_M', 0), ('_N', 0)]], [{'_axis': *, '_nodes': *, '_period': (*, *, *), 
'_source_grid': None, '_spacing': None, '_weights': None}]), 'fft_index': *, 'lm_modes': *}]), 'grid': 
CustomNode(ConcentricGrid[[('_L', 16), ('_M', 16), ('_N', 16), ('_NFP', 2), ('_sym', ('ndarray', (), (np.True_,))), ('_node_pattern', 
'jacobi'), ('_coordinates', 'rtz'), ('_is_meshgrid', False)]], [{'_axis': *, '_inverse_poloidal_idx': *, '_inverse_rho_idx': *, 
'_inverse_zeta_idx': *, '_nodes': *, '_period': (*, *, *), '_spacing': *, '_unique_poloidal_idx': *, '_unique_rho_idx': *, 
'_unique_zeta_idx': *, '_weights': *}])}}, '_grid': None, '_jac_chunk_size': None, '_loss_function': None, '_normalization': *, 
'_things': [CustomNode(Equilibrium[[('_sym', ('ndarray', (), (np.True_,))), ('_R_sym', 'cos'), ('_Z_sym', 'sin'), ('_NFP', 2), ('_L', 8), 
('_M', 8), ('_N', 8), ('_R_basis', FourierZernikeBasis at 0x72b99f611010 (L=8, M=8, N=8, NFP=2, sym=cos, 
spectral_indexing=ansi)), ('_Z_basis', FourierZernikeBasis at 0x72b99f611a60 (L=8, M=8, N=8, NFP=2, sym=sin, 
spectral_indexing=ansi)), ('_spectral_indexing', 'ansi'), ('_bdry_mode', 'lcfs'), ('_L_grid', 16), ('_M_grid', 16), ('_N_grid', 16), 
('_optimizable_params', ['R_lmn', 'Z_lmn', 'L_lmn', 'p_l', 'i_l', 'c_l', 'Psi', 'Te_l', 'ne_l', 'Ti_l', 'Zeff_l', 'a_lmn', 'Ra_n', 'Za_n', 
'Rb_lmn', 'Zb_lmn', 'I', 'G', 'Phi_mn'])]], [{'_L_lmn': *, '_Psi': *, '_R_lmn': *, '_Z_lmn': *, '_anisotropy': None, '_atomic_number': 
None, '_axis': CustomNode(FourierRZCurve[[('_name', ''), ('_R_basis', FourierSeries at 0x72b9a8131580 (L=0, M=0, N=8, 
NFP=2, sym=cos, spectral_indexing=linear)), ('_Z_basis', FourierSeries at 0x72b9a814fe00 (L=0, M=0, N=8, NFP=2, sym=sin, 
spectral_indexing=linear)), ('_sym', ('ndarray', (), (np.True_,))), ('_NFP', 2)]], [{'_R_n': *, '_Z_n': *, '_rotmat': *, '_shift': *}]), 
'_current': CustomNode(PowerSeriesProfile[[('_name', 'current'), ('_basis', PowerSeries at 0x72b99f612420 (L=8, M=0, N=0, 
NFP=1, sym=even, spectral_indexing=linear))]], [{'_params': *}]), '_electron_density': None, '_electron_temperature': None, 
'_ion_temperature': None, '_iota': None, '_pressure': CustomNode(PowerSeriesProfile[[('_name', 'pressure'), ('_basis', 
PowerSeries at 0x72bad4057b00 (L=8, M=0, N=0, NFP=1, sym=even, spectral_indexing=linear))]], [{'_params': *}]), '_surface': 
CustomNode(FourierRZToroidalSurface[[('_name', ''), ('_sym', ('ndarray', (), (np.True_,))), ('_L', 0), ('_M', 8), ('_N', 8), 
('_R_basis', DoubleFourierSeries at 0x72b99f611b20 (L=0, M=8, N=8, NFP=2, sym=cos, spectral_indexing=linear)), 
('_Z_basis', DoubleFourierSeries at 0x72b9aaa138f0 (L=0, M=8, N=8, NFP=2, sym=sin, spectral_indexing=linear)), ('_NFP', 2), 
('_rho', ('ndarray', (), (np.int64(1),)))]], [{'_R_lmn': *, '_Z_lmn': *}])}])]}]),), '_things': [CustomNode(Equilibrium[[('_sym', ('ndarray', 
(), (np.True_,))), ('_R_sym', 'cos'), ('_Z_sym', 'sin'), ('_NFP', 2), ('_L', 8), ('_M', 8), ('_N', 8), ('_R_basis', FourierZernikeBasis at 
0x72b99f611010 (L=8, M=8, N=8, NFP=2, sym=cos, spectral_indexing=ansi)), ('_Z_basis', FourierZernikeBasis at 
0x72b99f611a60 (L=8, M=8, N=8, NFP=2, sym=sin, spectral_indexing=ansi)), ('_spectral_indexing', 'ansi'), ('_bdry_mode', 
'lcfs'), ('_L_grid', 16), ('_M_grid', 16), ('_N_grid', 16), ('_optimizable_params', ['R_lmn', 'Z_lmn', 'L_lmn', 'p_l', 'i_l', 'c_l', 'Psi', 
'Te_l', 'ne_l', 'Ti_l', 'Zeff_l', 'a_lmn', 'Ra_n', 'Za_n', 'Rb_lmn', 'Zb_lmn', 'I', 'G', 'Phi_mn'])]], [{'_L_lmn': *, '_Psi': *, '_R_lmn': *, 
'_Z_lmn': *, '_anisotropy': None, '_atomic_number': None, '_axis': CustomNode(FourierRZCurve[[('_name', ''), ('_R_basis', 
FourierSeries at 0x72b9a8131580 (L=0, M=0, N=8, NFP=2, sym=cos, spectral_indexing=linear)), ('_Z_basis', FourierSeries at 
0x72b9a814fe00 (L=0, M=0, N=8, NFP=2, sym=sin, spectral_indexing=linear)), ('_sym', ('ndarray', (), (np.True_,))), ('_NFP', 
2)]], [{'_R_n': *, '_Z_n': *, '_rotmat': *, '_shift': *}]), '_current': CustomNode(PowerSeriesProfile[[('_name', 'current'), ('_basis', 
PowerSeries at 0x72b99f612420 (L=8, M=0, N=0, NFP=1, sym=even, spectral_indexing=linear))]], [{'_params': *}]), 
'_electron_density': None, '_electron_temperature': None, '_ion_temperature': None, '_iota': None, '_pressure': 
CustomNode(PowerSeriesProfile[[('_name', 'pressure'), ('_basis', PowerSeries at 0x72bad4057b00 (L=8, M=0, N=0, NFP=1, 
sym=even, spectral_indexing=linear))]], [{'_params': *}]), '_surface': CustomNode(FourierRZToroidalSurface[[('_name', ''), 
('_sym', ('ndarray', (), (np.True_,))), ('_L', 0), ('_M', 8), ('_N', 8), ('_R_basis', DoubleFourierSeries at 0x72b99f611b20 (L=0, M=8, 
N=8, NFP=2, sym=cos, spectral_indexing=linear)), ('_Z_basis', DoubleFourierSeries at 0x72b9aaa138f0 (L=0, M=8, N=8, 
NFP=2, sym=sin, spectral_indexing=linear)), ('_NFP', 2), ('_rho', ('ndarray', (), (np.int64(1),)))]], [{'_R_lmn': *, '_Z_lmn': *}])}])], 
'_unflatten': CustomNode(_ThingUnflattener[[('length', 1), ('inds', [0]), ('treedef', PyTreeDef([[*]]))]], [{}])}]))

```

Assuming we don't store `things` anymore, this would reduce to,
```python
obj._things = []
obj._objectives[0]._things = []
from jax.tree_util import tree_flatten
vals, treedef = tree_flatten(obj)
print(treedef.num_leaves, treedef.num_nodes)
print(treedef)
```
```
98 183
PyTreeDef(CustomNode(ObjectiveFunction[[('_jac_chunk_size', 3282), ('_use_jit', True), ('_deriv_mode', 'batched'), ('_built', 
True), ('_compiled', False), ('_name', 'ObjectiveFunction'), ('_dim_f', 5346), ('_scalar', False), ('_things', []), 
('_things_per_objective_idx', [[0]])]], [{'_flatten': CustomNode(_ThingFlattener[[('length', 1), ('treedef', PyTreeDef([[*]]))]], [{}]), 
'_objectives': (CustomNode(ForceBalance[[('_target', ('ndarray', (), (np.int64(0),))), ('_weight', ('ndarray', (), (np.int64(1),))), 
('_normalize', True), ('_normalize_target', True), ('_deriv_mode', 'fwd'), ('_name', 'force'), ('_use_jit', True), ('_built', True), 
('_things', []), ('_dim_f', 5346), ('_data_keys', ['F_rho', '|grad(rho)|', 'sqrt(g)', 'F_helical', '|e^helical*sqrt(g)|'])]], [{'_bounds': None, 
'_constants': {'profiles': {'atomic_number': None, 'current': CustomNode(PowerSeriesProfile[[('_name', 'current'), ('_basis', 
PowerSeries at 0x72b9aaa2f9b0 (L=8, M=0, N=0, NFP=1, sym=even, spectral_indexing=linear))]], [{'_params': *}]), 
'electron_density': None, 'electron_temperature': None, 'ion_temperature': None, 'iota': None, 'pressure': 
CustomNode(PowerSeriesProfile[[('_name', 'pressure'), ('_basis', PowerSeries at 0x72b9aaa2f170 (L=8, M=0, N=0, NFP=1, 
sym=even, spectral_indexing=linear))]], [{'_params': *}])}, 'quad_weights': *, 'transforms': {'L': CustomNode(Transform[[('_basis', 
FourierZernikeBasis at 0x72b99f610ad0 (L=8, M=8, N=8, NFP=2, sym=sin, spectral_indexing=ansi)), ('_rcond', 'auto'), ('_built', 
True), ('_built_pinv', False), ('_derivatives', ('ndarray', (10, 3), (np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(0), 
np.int64(0), np.int64(2), np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(0), np.int64(1), np.int64(1), np.int64(0), 
np.int64(0), np.int64(2), np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(1), np.int64(0), 
np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2)))), ('_method', 'fft'), ('num_lm_modes', 45), ('num_n_modes', 17), 
('pad_dim', 16)]], [{'_grid': CustomNode(ConcentricGrid[[('_L', 16), ('_M', 16), ('_N', 16), ('_NFP', 2), ('_sym', ('ndarray', (), 
(np.True_,))), ('_node_pattern', 'jacobi'), ('_coordinates', 'rtz'), ('_is_meshgrid', False)]], [{'_axis': *, '_inverse_poloidal_idx': *, 
'_inverse_rho_idx': *, '_inverse_zeta_idx': *, '_nodes': *, '_period': (*, *, *), '_spacing': *, '_unique_poloidal_idx': *, 
'_unique_rho_idx': *, '_unique_zeta_idx': *, '_weights': *}]), '_matrices': {'fft': {0: {0: *, 1: *, 2: *}, 1: {0: *, 1: *, 2: {}}, 2: {0: *, 1: {}, 2: 
{}}}}, 'dk': *, 'fft_grid': CustomNode(Grid[[('_NFP', 1), ('_sym', False), ('_node_pattern', 'custom'), ('_coordinates', 'rtz'), 
('_is_meshgrid', False), ('_L', 0), ('_M', 0), ('_N', 0)]], [{'_axis': *, '_nodes': *, '_period': (*, *, *), '_source_grid': None, '_spacing': 
None, '_weights': None}]), 'fft_index': *, 'lm_modes': *}]), 'R': CustomNode(Transform[[('_basis', FourierZernikeBasis at 
0x72b99f611fa0 (L=8, M=8, N=8, NFP=2, sym=cos, spectral_indexing=ansi)), ('_rcond', 'auto'), ('_built', True), ('_built_pinv', 
False), ('_derivatives', ('ndarray', (10, 3), (np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(0), np.int64(0), np.int64(2), 
np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2), 
np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(1), np.int64(0), np.int64(1), np.int64(1), 
np.int64(0), np.int64(0), np.int64(2)))), ('_method', 'fft'), ('num_lm_modes', 45), ('num_n_modes', 17), ('pad_dim', 16)]], [{'_grid': 
CustomNode(ConcentricGrid[[('_L', 16), ('_M', 16), ('_N', 16), ('_NFP', 2), ('_sym', ('ndarray', (), (np.True_,))), ('_node_pattern', 
'jacobi'), ('_coordinates', 'rtz'), ('_is_meshgrid', False)]], [{'_axis': *, '_inverse_poloidal_idx': *, '_inverse_rho_idx': *, 
'_inverse_zeta_idx': *, '_nodes': *, '_period': (*, *, *), '_spacing': *, '_unique_poloidal_idx': *, '_unique_rho_idx': *, 
'_unique_zeta_idx': *, '_weights': *}]), '_matrices': {'fft': {0: {0: *, 1: *, 2: *}, 1: {0: *, 1: *, 2: {}}, 2: {0: *, 1: {}, 2: {}}}}, 'dk': *, 'fft_grid': 
CustomNode(Grid[[('_NFP', 1), ('_sym', False), ('_node_pattern', 'custom'), ('_coordinates', 'rtz'), ('_is_meshgrid', False), ('_L', 
0), ('_M', 0), ('_N', 0)]], [{'_axis': *, '_nodes': *, '_period': (*, *, *), '_source_grid': None, '_spacing': None, '_weights': None}]), 
'fft_index': *, 'lm_modes': *}]), 'Z': CustomNode(Transform[[('_basis', FourierZernikeBasis at 0x72b99f610ad0 (L=8, M=8, N=8, 
NFP=2, sym=sin, spectral_indexing=ansi)), ('_rcond', 'auto'), ('_built', True), ('_built_pinv', False), ('_derivatives', ('ndarray', (10, 
3), (np.int64(0), np.int64(0), np.int64(0), np.int64(1), np.int64(0), np.int64(0), np.int64(2), np.int64(0), np.int64(0), np.int64(0), 
np.int64(1), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2), np.int64(0), np.int64(0), np.int64(0), 
np.int64(1), np.int64(1), np.int64(0), np.int64(1), np.int64(0), np.int64(1), np.int64(1), np.int64(0), np.int64(0), np.int64(2)))), 
('_method', 'fft'), ('num_lm_modes', 45), ('num_n_modes', 17), ('pad_dim', 16)]], [{'_grid': CustomNode(ConcentricGrid[[('_L', 
16), ('_M', 16), ('_N', 16), ('_NFP', 2), ('_sym', ('ndarray', (), (np.True_,))), ('_node_pattern', 'jacobi'), ('_coordinates', 'rtz'), 
('_is_meshgrid', False)]], [{'_axis': *, '_inverse_poloidal_idx': *, '_inverse_rho_idx': *, '_inverse_zeta_idx': *, '_nodes': *, '_period': 
(*, *, *), '_spacing': *, '_unique_poloidal_idx': *, '_unique_rho_idx': *, '_unique_zeta_idx': *, '_weights': *}]), '_matrices': {'fft': {0: 
{0: *, 1: *, 2: *}, 1: {0: *, 1: *, 2: {}}, 2: {0: *, 1: {}, 2: {}}}}, 'dk': *, 'fft_grid': CustomNode(Grid[[('_NFP', 1), ('_sym', False), 
('_node_pattern', 'custom'), ('_coordinates', 'rtz'), ('_is_meshgrid', False), ('_L', 0), ('_M', 0), ('_N', 0)]], [{'_axis': *, '_nodes': *, 
'_period': (*, *, *), '_source_grid': None, '_spacing': None, '_weights': None}]), 'fft_index': *, 'lm_modes': *}]), 'grid': 
CustomNode(ConcentricGrid[[('_L', 16), ('_M', 16), ('_N', 16), ('_NFP', 2), ('_sym', ('ndarray', (), (np.True_,))), ('_node_pattern', 
'jacobi'), ('_coordinates', 'rtz'), ('_is_meshgrid', False)]], [{'_axis': *, '_inverse_poloidal_idx': *, '_inverse_rho_idx': *, 
'_inverse_zeta_idx': *, '_nodes': *, '_period': (*, *, *), '_spacing': *, '_unique_poloidal_idx': *, '_unique_rho_idx': *, 
'_unique_zeta_idx': *, '_weights': *}])}}, '_grid': None, '_jac_chunk_size': None, '_loss_function': None, '_normalization': *}]),), 
'_unflatten': CustomNode(_ThingUnflattener[[('length', 1), ('inds', [0]), ('treedef', PyTreeDef([[*]]))]], [{}])}]))

```

desc/objectives/_stability.py
```diff
@@ -392,6 +392,15 @@ class BallooningStability(_Objective):
         normalize_target_detail=" Note: Has no effect for this objective.",
     )
 
+    _static_attrs = _Objective._static_attrs + [
+        "_iota_keys",
+        "_len_keys",
```
unalmis:
```suggestion
```

desc/objectives/objective_funs.py
```diff
@@ -1206,6 +1244,10 @@ def build(self, use_jit=True, verbose=1):
         if not self._use_jit:
             self._unjit()
 
+        # put the constants to device as jax arrays
```
YigitElma: I think your previous comment had fair points. I need to test this better some other time, until then you can remove these lines.

tests/test_examples.py
YigitElma: Is this still necessary? Locally it passed with -5

desc/equilibrium/equilibrium.py
YigitElma: Do you still think this is an issue with the basis hash? Previous slow benchmark problem seem to be gone, but I don't know what caused that exactly...

YigitElma (changes requested):

desc/io/optimizable_io.py
```diff
@@ -79,20 +77,6 @@ def load(load_from, file_format=None):
...
-        return all([_unjittable(y) or y is None for y in x])
-    if isinstance(x, dict):
-        return all([_unjittable(y) or y is None for y in x.values()])
-    if hasattr(x, "dtype") and np.ndim(x) == 0:
-        return np.issubdtype(x.dtype, np.bool_) or np.issubdtype(x.dtype, np.int_)
-    return isinstance(
-        x, (str, types.FunctionType, functools.partial, bool, int, np.int_)
-    )
-
-
```
YigitElma, line 95: I have a question about `_unmake_hashable` function. In its current state, it converts all of the `ndarrays` to `numpy` array. I think this is generally fine, but we can have `jax` arrays as static too (by the `_make_hashable` function). So, maybe adding one more case for `jax` arrays could help.

desc/io/optimizable_io.py
```diff
@@ -138,19 +122,11 @@ def _generic_tree_flatten(obj):
...
             static_attrs = getattr(obj, "_static_attrs", [])
-            dynamic_attrs = getattr(obj, "_dynamic_attrs", [])
-            assert set(static_attrs).isdisjoint(set(dynamic_attrs))
 
             for key, val in obj.__dict__.items():
                 if key in static_attrs:
                     aux_data += [(key, _make_hashable(val))]
-                elif key in dynamic_attrs:
-                    children[key] = val
-                elif _unjittable(val):
```
YigitElma, line 152: I think for DESC code base it is fine, but if a user has a custom objective class that uses one of the class attributes that is an integer or string as static, for instance,
```python
self.number = 6
...
def compute():
    if self.number > 5:
        ...
```
This change is kind of breaking. Previously, any `IOAble` type convert them automatically, but now they need to be assigned manually. Probably `jax`s error message, `TracerValue encountered array comparison in jit compiled function ...` is self-explanatory enough but this might worth mentioning in the changelog.

desc/basis.py
```diff
@@ -317,6 +316,34 @@ def __repr__(self):
...
+                self._sym,
+                self._spectral_indexing,
+            )
+        )
+
+    def __eq__(self, other):
+        """Check if two basis objects are equal."""
+        if not isinstance(other, _Basis):
+            return False
+        return (
```
unalmis, line 337: usually the first condition checks reference equality e.g. `self is other` to short-circuit

desc/io/optimizable_io.py
f0uriest: marking jax arrays as static causes weird bugs. Basically, if you create a jax array inside a jitted function (eg by reconstituting an array we want to be static), jax always treats it as traced, so static operations (control flow etc) breaks. The jax recommended way is to use numpy arrays for static arguments which is what we do here

desc/objectives/_fast_ion.py
```diff
@@ -140,6 +140,15 @@ class GammaC(_Objective):
         overwrite=_bounce_overwrite,
     )
 
+    _static_attrs = _Objective._static_attrs + [
+        "_hyperparam",
```
unalmis, line 144: @YigitElma isn't this an issue because we call `pop` on `_hyperparam` in `build`? Same thing with ripple. The hash value of `_hyperparam` will then change after popping

YigitElma: I think it is fine if you just pop in build but not in compute function. You should be able to change the pytree structure as long as you do so before first compilation (for example, changing type in build is fair game). The first time we compile a function is after build.

YigitElma: For instance, we add some attributes to static_attrs that don't even exist before build, they would definitely cause a recompilation, but I haven't noticed them yet.

YigitElma: One thing I realized about recompilation that I am not sure if it worths the effort to even fix is the `print_value` function at the end of optimization. During optimization, we call ObjectiveFunction.compute_scaled_error but for printing we call individual Objective.compute_scaled_error. Technically, when we compile the former, you compile the latter, but unfortunately compilation cache doesnt work like that, hence you recompile the function at the very end 'again'. One fix could be to use ObjectiveFunction's method and use indexing based on dim_f to print. But again I am not sure if it worth spending time.

PR #1663, opened 2025-03-30
Add `plot_field_lines` function

YigitElma:
Adds 3D version of `poincare_plot` function to draw the field lines for number of transits.
image

### DIFF
--- a/desc/plotting.py
+++ b/desc/plotting.py
@@ -46,6 +46,7 @@ __all__ = [
     "plot_section",
     "plot_surfaces",
     "poincare_plot",
+    "plot_field_lines",
 ]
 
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+New Features
+
+- Adds `plot_field_lines` function
```
dpanici, line 6:
```suggestion
- Adds `desc.plotting.plot_field_lines` function, which integrates and then plots magnetic field line trajectories in 3D.
```

rahulgaur104 (dismissed):

f0uriest (changes requested):
Should also add a test that at least verifies the code runs without error (could also check with `return_data=True` to make sure the data looks correct since we can't easily do regression tests on plotly plots)

tests/test_plotting.py
```diff
@@ -960,3 +961,16 @@ def test_plot_poincare():
...
+def test_plot_field_lines():
+    """Test plotting field lines."""
+    field = ToroidalMagneticField(1.0, 1.0)
+    fig, data = plot_field_lines(
+        field, [1.0], [0.0], nphi_per_transit=10, ntransit=0.8, return_data=True
+    )
+    assert all(data["Z"][0] == 0)
+    assert np.allclose((data["X"][0] ** 2 + data["Y"][0] ** 2), 1)
+
+    return fig
```
YigitElma, line 976:
```suggestion
```

desc/plotting.py
```diff
@@ -3613,6 +3614,231 @@ def plot_basis(basis, return_data=False, **kwargs):
...
+    ), f"plot_field_lines got unexpected keyword argument: {kwargs.keys()}"
+
+    if not isinstance(lw, (list, tuple)):
+        lw = [lw]
+    if not isinstance(ls, (list, tuple)):
+        ls = [ls]
+    if not isinstance(color, (list, tuple)):
+        color = [color]
+
+    errorif(
```
dpanici, line 3730: https://github.com/PlasmaControl/DESC/blob/2ce4dc66d31691bd633fb855a1bff78fbc4689de/desc/utils.py#L583

we have a util function for this and checks positivity which your thing should probably check too

dpanici (changes requested):
tiny change to reuse existing we code we have and check for positivity of number of transits

rahulgaur104 (approved):

desc/plotting.py
```diff
@@ -3613,6 +3621,227 @@ def plot_basis(basis, return_data=False, **kwargs):
         return fig, ax
 
 
+def plot_field_lines(
+    field,
+    R0,
+    Z0,
+    Phi0=0,
```
f0uriest, line 3628: minor point, but in our notation the toroidal angle is `phi` (lowercase), `Phi` (uppercase) usually means a potential of some kind.

desc/plotting.py
```diff
@@ -3613,6 +3621,227 @@ def plot_basis(basis, return_data=False, **kwargs):
...
+    Phi0 : float
+        Starting value of phi for field line tracing in radians. Defaults to 0.
+    ntransit : int, float
+        Number of transits to trace field lines for. Defaults to 1.
+    nphi_per_transit : int
+        Number of phi values to use for each transit. This is the number of
+        points to plot for each field line per transit.
+        Defaults to 50.
+    endpoint : bool
+        If True, [Phi0, Phi0+2*Pi*ntransit] is included in the field line.
```
f0uriest, line 3653: this description is a bit confusing. I think you mean
```
If True, the point phi0+2*pi*ntransit is included, else the last point is phi0 + 2*np.pi*(ntransit - 1/nphi_per_transit)
```
or something similar.

Also do we want this to default to True? I think we generally include endpoints in other plotting functions, so that eg rational fieldlines will actually close on themselves

desc/plotting.py
```diff
@@ -3613,6 +3621,227 @@ def plot_basis(basis, return_data=False, **kwargs):
...
+        **fli_kwargs,
+    )
+
+    zs = fieldZ.reshape((npts, -1))
+    rs = fieldR.reshape((npts, -1))
+
+    signBT = np.sign(
+        field.compute_magnetic_field(np.array([R0.flat[0], 0.0, Z0.flat[0]]))[:, 1]
+    ).flat[0]
+    if signBT < 0:  # field lines are traced backwards when toroidal field < 0
```
f0uriest, line 3757: This might need some extra thought. If Bt<0 I think we might need to also reverse phi before integrating or something so that the plot actually matches what the user wants

desc/plotting.py
```diff
@@ -3613,6 +3621,227 @@ def plot_basis(basis, return_data=False, **kwargs):
...
+    Parameters
+    ----------
+    field : MagneticField
+        External field, coilset, current potential etc to plot from.
+    R0, Z0 : array-like
+        Starting points for field line tracing.
+    phi0 : float
+        Starting value of phi for field line tracing in radians. Defaults to 0.
+    ntransit : int, float
+        Number of transits to trace field lines for. Defaults to 1.
```
dpanici, line 3647: specify that this traces field lines forward in `phi`

dpanici (changes requested):
add negative phi test, update docstring to specify direction of phi tracing (right now as coded it only traces forward in phi right?)

desc/plotting.py
```diff
@@ -3613,6 +3630,236 @@ def plot_basis(basis, return_data=False, **kwargs):
...
+    plot_data["X"] = []
+    plot_data["Y"] = []
+    plot_data["Z"] = []
+    for i in range(rs.shape[1]):  # iterate over each field line
+        x = rs[:, i] * np.cos(phis)
+        y = rs[:, i] * np.sin(phis)
+        z = zs[:, i]
+        plot_data["X"].append(x)
+        plot_data["Y"].append(y)
+        plot_data["Z"].append(z)
```
dpanici, line 3792:
```suggestion
        plot_data["Z"].append(z)
        plot_data["R"].append(rs[:,i])
        plot_data["phi"].append(phis)
        
```

OR, mention in docstring what the returned data keys are ("X", "Y", "Z")

desc/plotting.py
```diff
@@ -3613,6 +3630,236 @@ def plot_basis(basis, return_data=False, **kwargs):
...
+          True by default.
+
+        Additionally, any other keyword arguments will be passed on to
+        ``desc.magnetic_fields.field_line_integrate``
+
+    Returns
+    -------
+    fig : plotly.graph_objs._figure.Figure
+        Figure being plotted to.
+    plot_data : dict
```
dpanici, line 3695: Add here what `plot_data` contains (keys) and their data format (list of length rs.size of arrays of length phis.size)

dpanici (dismissed):
Im convinced of the reverse stuff being fine

desc/plotting.py
YigitElma: I think the test I added make it clear

dpanici:
check that field_line_integrate's rhs odefun is doing what we want (is it tracing field line backwards? giving the field lien at desired phi or at negative of desired phi?)

dpanici:
tl;dr: either move all the logic for flipping `r,z` when `B_phi<0` into `field_line_integrate`, or have `field_line_integrate` also return `phis` as an output of the integration (so that they are negative when `Bphi<0` and then we have the actual physical locations without needing to do weird things like flipping orders).

Ok so just to state the math again

- $l$ is whatever field-line-following parameter we choose
-  is then how the position vector changes while incrementing along this field-line-following parameter.
- $\mathbf{dx}/{dl} = \frac{dR}{dl} \hat{R} + \frac{d\phi}{dl} R \hat{phi} + \frac{dZ}{dl} \hat{Z} +  $ is parallel to $\mathbf{B}$ by definition (as $l$ increases, we should move along the direction that the field line is pointing, $\mathbf{B}/|B|$)
- Then, the field line equations are obtained by taking $\mathbf{dx}/{dl} \times \mathbf{B}=0$
  -  $\frac{dR}{dl} = \frac{R B_R}{B_{\phi}} \frac{d\phi}{dl}$
  - $\frac{dZ}{dl} = \frac{R B_Z}{B_{\phi}} \frac{d\phi}{dl}$
  - $\frac{d\phi}{dl}$ is determined by our choice of $l$
 
Now if we want to identify $l$ with the toroidal angle $\phi$, then this determines the magnitude of $\frac{d\phi}{dl}$ (to be equal to 1), however the sign of $\frac{d\phi}{dl}$ is determined by remembering  $\frac{d\phi}{dl} = \hat{phi} \cdot \mathbf{dx}/{dl} / R$, now $R>0$ always, and $\mathbf{dx}/{dl} \cdot \mathbf{B} > 0$ since they are parallel. So determining the sign of $\frac{d\phi}{dl}$ can be done by considering the $sign(\hat{phi} \cdot \mathbf{B}) = sign(B_{\phi})$.

Our field line integration equations are consistent with the above, namely they consider the sign of $B_\phi$ in the odefun, and we are always integrating in the increasing $l$ direction. So, when $B_\phi>0$, this means that the time variable of our ode integration is indeed $\phi$, but when $B_\phi<0$, our time variable is actually $-\phi$.

This means that our field line integration is _always_ integrating along the field line in the direction the field line is facing. What it also means is that when $B_\phi<0$, (say $\mathbf{B} = 1/R \hat{phi}$ for example's sake) if you ask for `phis = np.linspace(0,2*np.pi)` from some `r0=R0` and `z0=0`, and then look at the coordinates (`x=R, phi, Z`) output by the solution of `field_line_integrate`, they will form a circle oriented in the CW direction.

However, we do NOT return the `phi` that is the output of the integration in `field_line_integrate`! So if you were to take the output `r,z` and plot them with you input `phis=np.linspace(0,2*np.pi)` , you'd think your field line is moving in the opposite direction than it actually is  (the soln has constant R, Z, but you'd think it is moving in direction of increasing phi when it is actually pointing the opposite way)

desc/plotting.py
```diff
@@ -1926,22 +1938,36 @@ def poincare_plot(
...
 
     zs = fieldZ.reshape((ntransit, nplanes, -1))
     rs = fieldR.reshape((ntransit, nplanes, -1))
 
     signBT = np.sign(
         field.compute_magnetic_field(np.array([R0.flat[0], 0.0, Z0.flat[0]]))[:, 1]
     ).flat[0]
-    if signBT < 0:  # field lines are traced backwards when toroidal field < 0
+    if signBT < 0:
+        warnings.warn(
```
YigitElma, line 1961: I added this new warning because `field_line_integrate`s warning message talks about `phis` that the users actually doesn't pass. So, this one is a bit more concise version.

desc/magnetic_fields/_core.py
```diff
@@ -2609,6 +2609,18 @@ def field_line_integrate(
     z0 = z0.flatten()
     x0 = jnp.array([r0, phis[0] * jnp.ones_like(r0), z0]).T
 
+    signBT = np.sign(field.compute_magnetic_field(x0)[0, 1])
+    if signBT < 0:
+        warnings.warn(
+            "Field has negative toroidal component. Field line integration is being "
+            "done in the negative toroidal direction, hence the output is not in phis "
+            "but in negative phis. Use the outputed phis to get the correct "
```
dpanici, line 2617: Trying to think of a clearer way to phrase this

Field toroidal component is negative, so the field line points in the negative toroidal direction. Field line integration is done along the field line vector direction, hence the output `phi` array of the toroidal angle of the integrated trajectory will be equal to negative of the input `phis` array

I am not sure about this last part, maybe reword as "if you want to integrate along the same field trajectory but in the positive toroidal direction, you can flip the sign of the field", It is not that they are doing something wrong, it is just that the output `phis` will be different (the negative) of the input `phi` array.

desc/magnetic_fields/_core.py
```diff
@@ -2598,8 +2598,8 @@ def field_line_integrate(
 
     Returns
     -------
-    r, z : ndarray
-        arrays of r, z coordinates at specified phi angles
+    r, z, phi : ndarray
+        arrays of r, z and phi coordinates of the field line.
```
dpanici, line 2602:
```suggestion
        arrays of r, z and phi coordinates of the field line, where `phi` is the geometric toroidal angle. 
        Note that `phi` will be equal to the negative of the input `phis` array if Bphi<0.
```

desc/magnetic_fields/_core.py
```diff
@@ -2609,6 +2611,21 @@ def field_line_integrate(
     z0 = z0.flatten()
     x0 = jnp.array([r0, phis[0] * jnp.ones_like(r0), z0]).T
 
+    signBT = np.sign(field.compute_magnetic_field(x0)[0, 1])
+    if signBT < 0:
+        warnings.warn(
+            "Toroidal component of the field is negative. Since field line integration "
+            "follows the direction of the magnetic field vector, the resulting "
+            "toroidal angle array (phi) from the integration will be in descending "
```
dpanici, line 2619:
```suggestion
            "toroidal angle array (phi) from the integration will be in descending order (i.e. phi = [-phis[0]... -phis[-1]]) "
```

dpanici (approved):

dpanici:
- internally convert the user `phi` if the `bphi<0` such that the integration returned is at the requested `phi`
- possibly file issue with diffrax for min_step_size needing to be negative

desc/magnetic_fields/_core.py
```diff
@@ -2617,11 +2616,25 @@ def field_line_integrate(
...
+
+    if not pos_BT:
+        # if Bphi is negative, we need to flip the field direction
+        # this doesn't change the field-line, but useful for integration
+        field_in_phi = ScaledMagneticField(-1.0, field)
+    else:
+        field_in_phi = field
+    if not pos_phi:
+        # if phis is decreasing, we need to use negative steps size
+        # to integrate in the right direction
```
dpanici: make an issue with diffrax to ask if min_step_size < 0 is actually supposed to be allowed

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ New Features
 - Adds a new utility function ``desc.compat.contract_equilibrium`` which takes in an ``Equilibrium`` object and an argument ``inner_rho``, and returns a new ``Equilibrium`` with original ``Equilibrium``'s ``inner_rho`` flux surface as its boundary.
 Optionally can also contract the profiles of the original ``Equilibrium`` so that the new ``Equilibrium``'s profiles match the original's in real space.
 - Adds second-order NAE constraints, accessible by passing ``order=2`` to ``desc.objectives.get_NAE_constraints``.
+- Adds `desc.plotting.plot_field_lines` function, which integrates and then plots magnetic field line trajectories in 3D.
+- `desc.magnetic_fields.field_line_integrate` now returns the `phi` values where the `r` and `z` values belong to. Although the documentation was mentioning that the output can be in negative `phis`, it was not clear and open to misinterpretation.
```
f0uriest: I think this is now outdated, but should mention that `field_line_integrate` now outputs at the correct phi for negative fields

docs/api_plotting.rst
```diff
@@ -48,6 +48,7 @@ Contour Plots of 2-D data
 
     desc.plotting.plot_3d
     desc.plotting.plot_coils
+    desc.plotting.plot_field_lines
```
f0uriest, line 51: also need to add this to the big alphabetical list of stuff in `api.rst`

desc/magnetic_fields/_core.py
```diff
@@ -2617,11 +2616,25 @@ def field_line_integrate(
...
+        min_step_size = -jnp.abs(min_step_size)
+
     @jit
     def odefun(s, rpz, args):
         rpz = rpz.reshape((3, -1)).T
         r = rpz[:, 0]
-        br, bp, bz = field.compute_magnetic_field(
+        br, bp, bz = field_in_phi.compute_magnetic_field(
```
dpanici: Add a private function _field_line_integrate that assumes positive Bt and monotonic phi, and lacks any checks that would hinder jit or AD

dpanici: and call this inside of field_line_integrate

f0uriest: could also just refactor this to work with jit by always scaling the field and stepsize, eg
```python
field_in_phi = ScaledMagneticField(jnp.sign(Bt), field)
min_step_size = jnp.sign(jnp.diff(phi)[0])*jnp.abs(min_step_size)
```

YigitElma: Done. I just multiplied the computed magnetic field by scale (sign of B toroidal), which is done in `ScaledMagneticField`, so no need to create a new object.

f0uriest (dismissed):

dpanici (approved):

f0uriest (dismissed):

YigitElma:
I think our `field_line_integrate` doesn't work properly with newer versions of diffrax!? On master, I cannot run the example of `poincare_plot`.
```python
field = desc.io.load("../tests/inputs/precise_QA_helical_coils.h5")
eq = desc.examples.get("precise_QA")

from desc.plotting import poincare_plot
grid_trace = LinearGrid(rho=np.linspace(0.4, 0.9, 7))
r0 = eq.compute("R", grid=grid_trace)["R"]
z0 = eq.compute("Z", grid=grid_trace)["Z"]
fig, ax = desc.plotting.poincare_plot(
    field,
    r0,
    z0,
    NFP=eq.NFP,
    color="k",
    size=0.5,
    ntransit=100
)
```

It throws,
```
TypeError: no unmapping handler for 0 of type <class 'int'>
```

YigitElma:
> I think our `field_line_integrate` doesn't work properly with newer versions of diffrax!? On master, I cannot run the example of `poincare_plot`.
> 
> ```python
> field = desc.io.load("../tests/inputs/precise_QA_helical_coils.h5")
> eq = desc.examples.get("precise_QA")
> 
> from desc.plotting import poincare_plot
> grid_trace = LinearGrid(rho=np.linspace(0.4, 0.9, 7))
> r0 = eq.compute("R", grid=grid_trace)["R"]
> z0 = eq.compute("Z", grid=grid_trace)["Z"]
> fig, ax = desc.plotting.poincare_plot(
>     field,
>     r0,
>     z0,
>     NFP=eq.NFP,
>     color="k",
>     size=0.5,
>     ntransit=100
> )
> ```
> 
> It throws,
> 
> ```
> TypeError: no unmapping handler for 0 of type <class 'int'>
> ```

Ok, this was caused by some messed up environment (or some of the older version dependencies has the issue). But when I created a fresh environment, it was gone.

YigitElma:
I added a test just in case!

f0uriest (approved):

PR #1668, opened 2025-04-01
Documentation Styling

### LINKED ISSUE #1658: proximal-lsq-exact not listed in table of supported optimizers
HI there, 

Method `proximal-lsq-exact` is not listed in your [table of supported optimizers](https://desc-docs.readthedocs.io/en/latest/optimizers.html). Though it appears to still be supported by DESC's Optimizer class. Just wondering if there was a reason for this? Would be more than happy to submit a PR but am unsure about the header values for `proximal-lsq-exact` (e.g. does it support inequality constraints).

Thanks

YigitElma:
- Adds explanation for `proximal-` prefix to List of Optimizers page
- Adds example plotly plots for `plot_coils` and `pot_3d` documentation
- Adds example `poincare_plot`
- Uses admonitions in some parts of the docs
- Makes the corners of code blocks rounded with slight hint of purple
- Reorganizes `installation` page with tabs and dropdowns

Resolves #1658


docs/installation.rst
```diff
@@ -56,197 +59,193 @@ These examples use conda environments.
...
+                # installation for users
+                pip install --editable .
+                # optionally install developer requirements (if you want to run tests)
+                pip install -r devtools/dev-requirements.txt
 
-If pre-built JAX binaries are not available, you will first need to build JAX from source.
-More info can be found here: https://jax.readthedocs.io/en/latest/developer.html
 
-These instructions were tested and confirmed to work on the Traverse supercomputer at Princeton as of Nov. 6, 2023.
+        .. dropdown:: On Clusters with IBM Power Architecture
```
f0uriest, line 206: i think we can probably get rid of this since Traverse is closed, and the version quoted is over a year old anyways

f0uriest (approved):

unalmis:
can we change the default option to render code back to light mode instead of dark mode?

YigitElma:
> can we change the default option to render code back to light mode instead of dark mode? it was changed to dark mode last year, but I think it looks worse

I also it looks weird, tbh, I would prefer everything to be in dark mode (an ideally toggle to switch between light and dark mode for whole website).

Do you remember if it was us or one of the packages that changed that?

unalmis:
dario changed

> > can we change the default option to render code back to light mode instead of dark mode? it was changed to dark mode last year, but I think it looks worse
> 
> I also it looks weird, tbh, I would prefer everything to be in dark mode (an ideally toggle to switch between light and dark mode for whole website).
> 
> Do you remember if it was us or one of the packages that changed that?

it was us, don't know why. see old versions https://desc-docs.readthedocs.io/en/v0.10.0/installation.html

unalmis (approved):

YigitElma:
> it was us, don't know why. see old versions

Ok, in future we can add a toggle for light/dark them to set this up :)

PR #1674, opened 2025-04-07
Add python/uv-relevant entries to .gitignore

cfe316:
Adds a few entries to the .gitignore.


YigitElma (approved):

dpanici (approved):

YigitElma:
PLEASE DON'T MERGE THIS UNTIL #1678 IS IN!

YigitElma:
> What's the standard procedure for DESC?

We usually require 2 approvals. Also, all of the unit/regression/notebook/benchmarks should pass. 

In some PR, there is no actual code change that needs to be tested; for example, if your change is only for documentation or some other file (like `.gitignore`), we don't need to run the tests (they take some time). There are individual checks for different workflows. Anyway, in those cases, tests don't run, but they are still checked as successful. Though, since no code coverage report is uploaded, `codecov` status gets stuck as `expected`, and that's when someone with admin privileges needs to overwrite it.

I will merge this now. Thanks!

Issue #1352, opened 2024-11-11
Allow plotting utitilities to accept compute quantity hyperparameters


unalmis:
would need another dictionary for compute kwargs

PR #1676, opened 2025-04-08
maxJ(J_s), adiabatic J, and direct omnigenity(J_alpha)

### LINKED ISSUE #1273: Maximum-J Objective
With the bounce averaging and bounce point finding utilities, it should be possible to do max-J optimization like is done recently by [Goodman](https://arxiv.org/pdf/2405.19860), also relevant paper by [Rodriguez](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/4691B14FC2713173CD8AB2490649761A/S0022377824000345a.pdf/maximumj_property_in_quasiisodynamic_stellarators.pdf)

### LINKED ISSUE #1675: max-J
Since all the bounce utilies are complete, it is now an ooportune time to add the max-J compute + objective function.

This can be added along with the other important quantities such as the second adiabatic invariant J_|| and direct omnigenity objective.

rahulgaur104:
Resolves #1675 

- [x] adiabatic_J, J_s, J_alpha compute functions
- [x] maxJ objective
- [x] plotting routine for adiabatic_J
- [x] tests to increase coverage
- [ ] actually useful tests for direct omnigenity dJpar_dalpha
- [x] finalizing normalization for all these compute functions


dpanici:
Can you do some verification of this against the simsopt implementation possibly?

rahulgaur104:
Here's a $\mathcal{J}\_{\parallel}$ plot that is used to understand how good omnigenity $\frac{\partial \mathcal{J}_{\parallel}}{\partial \alpha}$ is.

Adding this to the plotting routines. Little change in $\mathcal{J}\_{\parallel}$  along alpha means good omnigenity.

Jpar_UT225

rahulgaur104:
Here are the Jpar contour plots with the new plotting routine for an omnigenous equilibrium.

Jpar_contour_0 18
Jpar_contour_0 53
Jpar_contour_0 77
Jpar_contour_1 00

desc/compute/_fast_ion.py
```diff
@@ -354,3 +359,354 @@ def fun(pitch_inv):
...
+
+        return batch_map(fun, data["pitch_inv"], pitch_batch_size) / (
+            bounce.compute_fieldline_length(fl_quad)
+        )
+
+    grid = transforms["grid"]
+    data["Jpar"] = _compute(
+        Jpar0,
+        {
+            "B^zeta": data["B^zeta"],
```
rahulgaur104, line 466: @unalmis  This becomes a problem that causes unit test 3.12 to fail.
Basically, B^zeta is defined twice. I think the solution is straightforward for you to figure out but take me a couple of hours.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +359,354 @@ def fun(pitch_inv):
...
+    spline = kwargs.get("spline", True)
+    fl_quad = (
+        kwargs["fieldline_quad"] if "fieldline_quad" in kwargs else leggauss(Y_B // 2)
+    )
+    quad = (
+        kwargs["quad"]
+        if "quad" in kwargs
+        else get_quadrature(
+            leggauss(kwargs.get("num_quad", 32)),
+            (automorphism_sin, grad_automorphism_sin),
```
unalmis, line 424: My rule of thumb is to use `chebgauss2` instead of `leggauss` for weakly singular integrals.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +359,354 @@ def fun(pitch_inv):
...
+            Jpar = bounce.integrate(
+                [_Jpar_num],
+                pitch_inv,
+                data,
+                ["B^zeta"],
+                bounce.points(pitch_inv, num_well),
+                is_fourier=True,
+            )
+            # Take sum over wells, mean in alpha
+            return jnp.sum(Jpar, axis=-1)
```
unalmis, line 451: > Take sum over wells, mean in alpha

`J_par.sum(axis=-1).mean(axis=-2)`

but i think you want to remove the comment instead

desc/compute/_fast_ion.py
```diff
@@ -354,3 +359,354 @@ def fun(pitch_inv):
...
+
+        return batch_map(fun, data["pitch_inv"], pitch_batch_size) / (
+            bounce.compute_fieldline_length(fl_quad)
```
unalmis, line 454: Output of `batch_map(fun, data["pitch_inv"], pitch_batch_size)` is (rho, alpha, lambda)

so need to add dimensions before dividing:

``bounce.compute_fieldline_length(fl_quad)[:, jnp.newaxis, jnp.newaxis]``

desc/compute/_fast_ion.py
unalmis: > Basically, B^zeta is defined twice.

In the dependencies for the compute function, there was `B^zeta` + Bounce2D.requirednames = ["B^zeta", "B^zeta", ...]. The test was failing to warn of this.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +359,354 @@ def fun(pitch_inv):
...
+        "pitch_batch_size",
+        "surf_batch_size",
+        "spline",
+    ],
+)
+def _Jpar(params, transforms, profiles, data, **kwargs):
+    """Second adiabatic invariant of particle motion.
+
+    The normalization requires a length for which we have uesd the fieldline
+    length ∫ dl/B.
```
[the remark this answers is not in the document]
rahulgaur104: J_II = int dl v_||/sqrt(2E/m) = int dzeta * B/B^zeta sqrt(1 - lambda B)
 
So that's just bounce_integral of _J_par_num. Becasue _dI_ripple has an extra factor of B

All I need is integral dl which is like the bounce integral of 1.

Becasue the bounce integrate function is integral (function) dl.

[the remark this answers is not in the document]
unalmis: The inconsistency is the code currently is that the sum over the ripple wells is normalized by a quantity that has units of 1/webers, yet your J_|| is dimensionless.

The suggestion I gave above resolves this while retaining J_|| = int dl sqrt(1-lambda B).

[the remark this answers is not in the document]
rahulgaur104: Can you look at it again? I just pushed something.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +359,354 @@ def fun(pitch_inv):
         surf_batch_size,
     )
     return data
+
+
+@register_compute_fun(
+    name="Jpar",
```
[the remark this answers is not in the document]
rahulgaur104: I like that. Lets see what the rest of the devs think tomorrow.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +359,354 @@ def fun(pitch_inv):
...
+    label=(
+        # J_∥ = ∫ dl v_∥ /(√E ∫ dl)
+        "\\J_{\\parallel} = \\integrate v_{\\parallel} dl/\\integrate dl/B"
+    ),
+    units="~",
+    units_long="~",
+    description="Normalized second adiabatic invariant of motion.",
```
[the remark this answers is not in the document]
rahulgaur104: It should be, now

unalmis:
@rahulgaur104 use chebgauss2 from desc.integrals.quad utils.

rahulgaur104:
Yes, that's what I used.

dpanici:
- compute quantity name itself should be more specific like `"adiabatic J"`

desc/compute/_fast_ion.py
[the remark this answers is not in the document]
rahulgaur104: Bounce integraing a quantity x gives the units [x][m], right, because bounce integral multiplies a quantity with a length?
If so, I am just dividin by length of the field line.

rahulgaur104:
Ok, I'll reach out.


We always sum (all quantities) bounce integrals over multiple wells. Mean over wells doesn't make a lot of sense to me.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +361,341 @@ def fun(pitch_inv):
...
+    grid = transforms["grid"]
+    data["adiabatic J"] = _compute(
+        adiabatic_J0,
+        {},
+        data,
+        theta,
+        grid,
+        num_pitch,
+        surf_batch_size,
+    ) / (2 * jnp.pi * data["R0"])
```
unalmis, line 464: 2 pi * num transit

rahulgaur104: Good point.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +361,341 @@ def fun(pitch_inv):
...
+            is_fourier=True,
+            spline=spline,
+        )
+
+        def fun(pitch_inv):
+            adiabatic_J = bounce.integrate(
+                [_adiabatic_J_num],
+                pitch_inv,
+                data,
+                ["B^zeta"],
```
unalmis, line 444: This works fine, but if you give an empty list here, then you will save some compute time and memory. For example the note in the docstring
> Names in ``data`` to interpolate. Default is all keys in ``data``. Do not include ``|B|`` or ``B^zeta``.

rahulgaur104:
@dpanici  Actually the definition and implementation is pretty clear in Goodman's paper
which is pretty close to what I am doing (ReLU of the value - threshold)

Screenshot from 2025-04-21 10-10-14

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Optionally can also contract the profiles of the original ``Equilibrium`` so tha
 - Allows Redl compute quantities to use SplineProfile
 - Updates Redl bootstrap current consistency tutorial to include a ``SplineProfile`` optimization
 - Adds automatically generated header file showing date the input file was created with `desc.vmec.VMECIO.write_vmec_input`
+- Adds max-J(dJ\_ds), direct omnigenity(dJ\_dalpha), adiabaric J compute, plotting and optimization objectives
```
dpanici, line 14: Refer to the actual plotting functions and objectives added, with the module path (like the line above `desc.vmec.VMECIO.write_vmec_input`)

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Optionally can also contract the profiles of the original ``Equilibrium`` so tha
...
+
+Bug Fixes
```
dpanici, line 16: This part seems mistakenly added?

desc/compute/_fast_ion.py
```diff
@@ -354,3 +361,341 @@ def fun(pitch_inv):
...
+        grid,
+        num_pitch,
+        surf_batch_size,
+    ) / (2 * jnp.pi * num_transit * data["R0"])
+    return data
+
+
+@register_compute_fun(
+    name="dJ_dalpha",
+    label=(  # ∂_α J_∥/∮ dl/v_∥ =  ∮ dl/v_∥ (v_d ⋅ ∇ρ)/∮ dl/v_∥
```
unalmis, line 470: This label should say ∇ρ²

CHANGELOG.md
rahulgaur104: Yeah, I'll correct that.

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,263 @@ def _compute_spline(self, params, constants=None):
...
+            fieldline_quad=constants["fieldline quad"],
+            quad=constants["quad"],
+            **self._hyperparam,
+        )
+        dJ_ds = constants["transforms"]["grid"].compress(data[self._key])
+
+        thresh0, w0, w1 = constants["thresh0"], constants["w0"], constants["w1"]
+
+        # Shifted ReLU operation
+        dJ_ds_filtrd = (dJ_ds - thresh0) * (dJ_ds >= thresh0)
```
unalmis, line 615:
```suggestion
        dJ_ds_filtrd = jnp.maximum(dJ_ds - thresh0, 0)
```

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,263 @@ def _compute_spline(self, params, constants=None):
...
+        results = w0 * jnp.sum(dJ_ds_filtrd) + w1 * jnp.max(dJ_ds_filtrd)
```
unalmis, line 620: The docstring suggests that the sum and max should be performed along axes (-1 and -2) but not axis 0.

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,263 @@ def _compute_spline(self, params, constants=None):
...
+            Dictionary of equilibrium degrees of freedom, e.g.
+            ``Equilibrium.params_dict``.
+        constants : dict
+            Dictionary of constant data, e.g. transforms, profiles etc.
+            Defaults to ``self.constants``.
+
+        Returns
+        -------
+        dJ_ds : ndarray
+            maxJ proxy a function of the flux surface label.
```
unalmis, line 581: Right now this returns a global scalar instead of array of shape (grid.num rho)

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,263 @@ def _compute_spline(self, params, constants=None):
...
+            self._Y,
+            self._grid.compress(self._grid.nodes[:, 0]),
+            domain=(0, 2 * np.pi),
+        )
+        self._constants["fieldline quad"] = leggauss(self._hyperparam["Y_B"] // 2)
+        self._constants["quad"] = get_quadrature(
+            leggauss(self._hyperparam.pop("num_quad")),
+            (automorphism_sin, grad_automorphism_sin),
+        )
+        self._dim_f = self._grid.num_rho
```
unalmis, line 558: Make sure that this matches the size of the array returned in `objective.compute`

desc/plotting.py
```diff
@@ -3843,3 +3845,143 @@ def plot_logo(save_path=None, **kwargs):
...
+        )  # BUG: Array of len 1 takes all the mem and never finishes
+
+    if alphas is None:
+        alphas = np.linspace(0, 2 * np.pi, 32, endpoint=True)
+
+    if num_pitch is None:
+        num_pitch = 16
+
+    N_rho = len(rhos)
+    X, Y = 32, 64
```
unalmis, line 3907: Add a comment for issue 1352 

`# TODO(#1352)`

to shield yourself from the demands that the other developers will surely request here

desc/plotting.py
```diff
@@ -3843,3 +3845,143 @@ def plot_logo(save_path=None, **kwargs):
...
+        num_transit=1,
+        num_quad=24,
+        num_pitch=num_pitch,
+        alpha=alphas,
+    )
+    data_full = grid.compress(data0["adiabatic J"])
+
+    if mode == "single-surface" or N_rho == 1:
+        # Extract pitch angle range
+        minB = data0["min_tz |B|"][0]  # shape: (Pitch,)
```
unalmis, line 3926: This should just be a single float, so the comment is misleading

desc/plotting.py
```diff
@@ -3843,3 +3845,143 @@ def plot_logo(save_path=None, **kwargs):
...
+    The second adiabatic invariant J_|| is calculated by integrating the parallel
+    velocity along the magnetic field line over a complete bounce period.
+    """
+    if rhos is None:
+        rhos = np.linspace(0.1, 1, 10)
+    elif isinstance(rhos, float):
+        print("Single rho provided...\n Plotting J_||(α, λ)")
+        rhos = np.array(
+            [rhos, 1.0]
+        )  # BUG: Array of len 1 takes all the mem and never finishes
```
unalmis, line 3898: is this still an issue?

tests/test_fast_ion.py
```diff
@@ -96,3 +96,26 @@ def test_Gamma_c_Velasco_1D():
     fig, ax = plt.subplots()
     ax.plot(rho, grid.compress(data["old Gamma_c Velasco"]), marker="o")
     return fig
+
+
+@pytest.mark.unit
+@pytest.mark.slow
+@pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_1d)
+def test_direct_omnigenity():
+    """Test Γ_c Velasco with W7-X."""
```
unalmis, line 105:
```suggestion
    """Test dJ_alpha plot against old plot."""
```

unalmis (changes requested):

unalmis:
Left some comments. Main thing is the math is not correctly implemented.

desc/plotting.py
rahulgaur104: Yeah.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +361,341 @@ def fun(pitch_inv):
...
+        theta,
+        grid,
+        num_pitch,
+        surf_batch_size,
+    ) / (2 * jnp.pi * num_transit * data["R0"])
+    return data
+
+
+@register_compute_fun(
+    name="dJ_dalpha",
```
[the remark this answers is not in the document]
rahulgaur104: Sans the constants, it is the partial derivative of J w.r.t alpha. Here's a snippet from Calvo et al.
Screenshot from 2025-05-24 15-22-49

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,263 @@ def _compute_spline(self, params, constants=None):
...
+        Starting field line poloidal labels.
+        Default is single field line. To compute a surface average
+        on a rational surface, it is necessary to average over multiple
+        field lines until the surface is covered sufficiently.
+    num_transit : int
+        Number of toroidal transits to follow field line.
+        In an axisymmetric device, field line integration over a single poloidal
+        transit is sufficient to capture a surface average. For a 3D
+        configuration, more transits will approximate surface averages on an
+        irrational magnetic surface better, with diminishing returns.
```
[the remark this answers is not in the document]
rahulgaur104: Surface average of X, by definition should be int dl/B X/int dl/B.
Are you saying it's one of those issues where the denominator piles up and causes the average -> 0 for large num_transits.

tests/test_fast_ion.py
rahulgaur104: Done!

desc/plotting.py
rahulgaur104: Corrected.

desc/compute/_neoclassical.py
```diff
@@ -126,7 +126,7 @@ def _compute(
         simp=simp,
     )
     out = batch_map(fun, fun_data, surf_batch_size)
-    assert out.ndim == 1
+    # --no-verify assert out.ndim == 1
```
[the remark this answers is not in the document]
rahulgaur104: Removed

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,263 @@ def _compute_spline(self, params, constants=None):
...
+        Threshold for penalizing precession drift in metric above.
+    w0, w1 : float
+        Weights for sum and max terms in metric above.
+
+    """
+
+    __doc__ = __doc__.rstrip() + collect_docs(
+        target_default="``target=-1``.",
+        bounds_default="``target=-1``.",
+        normalize_detail=" Note: Has no effect for this objective.",
```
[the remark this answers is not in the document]
rahulgaur104: I think now it should be.

desc/compute/_fast_ion.py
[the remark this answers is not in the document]
unalmis: I think my comments are consistent with Calvo's paper. sans constants https://github.com/PlasmaControl/DESC/pull/1676#discussion_r2098778170
(∂_α J_∥)= ∮ dℓ/v_∥ (v_d ⋅ ∇ρ²).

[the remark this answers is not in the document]
rahulgaur104: Are you talking specifically about the extra factor of rho (because s=rho^2)? I agree with that and I changed the unicode equation in J_alpha. But if you are talking about something else, I am not sure I understand.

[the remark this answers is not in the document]
unalmis: Well before the recent commits it was computing the code was computing (∂_α J_∥) ~ ∮ dℓ/v_∥ (v_d ⋅ ∇ρ) / ∮ dℓ/v_∥. My comment was saying it should be (∂_α J_∥) ~ ∮ dℓ/v_∥ (v_d ⋅ ∇ρ²)

desc/compute/_fast_ion.py
rahulgaur104: Should be resolved now.

desc/compute/_fast_ion.py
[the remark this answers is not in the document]
rahulgaur104: Okay, I think it's resolved now.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +361,341 @@ def fun(pitch_inv):
...
+            is_fourier=True,
+            spline=spline,
+        )
+
+        def fun(pitch_inv):
+            v_tau, radial_drift = bounce.integrate(
+                [_v_tau, _radial_drift],
+                pitch_inv,
+                data,
+                ["cvdrift0", "gbdrift (periodic)", "gbdrift (secular)/phi"],
```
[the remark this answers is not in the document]
rahulgaur104: Resolved.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +361,341 @@ def fun(pitch_inv):
...
+            / data["pitch_inv"] ** 2,
+            axis=-1,
+        ) / bounce.compute_fieldline_length(fl_quad)
+
+    grid = transforms["grid"]
+    data["dJ_dalpha"] = _compute(
+        dJ_dalpha0,
+        {
+            "cvdrift0": data["cvdrift0"],
+            "gbdrift (periodic)": data["gbdrift (periodic)"],
```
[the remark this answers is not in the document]
rahulgaur104: Resolved.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +361,341 @@ def fun(pitch_inv):
...
+    dim=1,
+    profiles=[],
+    transforms={"grid": []},
+    params=[],
+    data=[
+        "min_tz |B|",
+        "max_tz |B|",
+        "cvdrift0",
+        "gbdrift (periodic)",
+        "gbdrift (secular)/phi",
```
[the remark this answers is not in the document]
rahulgaur104: Resolved now.

desc/plotting.py
rahulgaur104: Added.

desc/objectives/_fast_ion.py
rahulgaur104: It does. The test_objective_compute test ensures that.

desc/objectives/_fast_ion.py
rahulgaur104: Yes, no sum over rho. Only alpha and pitch.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +374,453 @@ def fun(pitch_inv):
...
+        grid,
+        num_pitch,
+        surf_batch_size,
+    )
+    # )--no-verify / (2 * jnp.pi * num_transit * data["R0"])
+    return data
+
+
+@register_compute_fun(
+    name="J_alpha",
```
f0uriest, line 602: `adiabatic J_alpha`?

desc/compute/_fast_ion.py
```diff
@@ -354,3 +374,453 @@ def fun(pitch_inv):
...
+        num_pitch,
+        surf_batch_size,
+        radial_drift=grid.compress(data["<v_dot_grads>"]),
+    )
+    # )--no-verify / (2 * jnp.pi * num_transit * data["R0"])
+    return data
+
+
+@register_compute_fun(
+    name="J_s",
```
f0uriest, line 708: `adiabatic J_s`?

Also, is there any strong reason to use `s` instead of `rho`? I guess `s` is more common in the literature, but `rho` would make it more consistent with the rest of the code.

desc/compute/_fast_ion.py
```diff
@@ -354,3 +374,453 @@ def fun(pitch_inv):
...
+        data,
+        theta,
+        grid,
+        num_pitch,
+        surf_batch_size,
+    )
+
+    # Since there isn't a standard method of normalization, I choose this.
+    # Gives the right result (J_s < 0 everywhere) for Goodman's elongated QI.
+    data["J_s"] = data["J_s"] / jnp.max(jnp.abs(data["J_s"]))
```
f0uriest, line 825: I'd prefer to move this sort of normalization to the relevant objective, and keep the compute qty as close to the "raw physics" as possible

desc/compute/_fast_ion.py
```diff
@@ -354,3 +374,453 @@ def fun(pitch_inv):
...
+
+            # max drift < 0 provides TEM(trapped electron mode)
+            # stability for all rhos and pitches
+            return dJ_ds
+
+        # Output dimension (rho, alpha, lambda)
+        return batch_map(fun, data["pitch_inv"], pitch_batch_size)
+
+    grid = transforms["grid"]
+    data["J_s"] = -1 * _compute(
```
f0uriest, line 809: why the -1?

desc/plotting.py
```diff
@@ -3843,3 +3845,144 @@ def plot_logo(save_path=None, **kwargs):
...
+    alphas : array_like, optional
+        Fieldline label values (toroidal angle).
+        Default: np.linspace(0, 2π, 32, endpoint=True)
+    num_pitch : int, optional
+        Number of pitch angle values for bounce integral calculation.
+        Default: 16
+    pitch_idx : int, optional
+        Index of pitch angle to use for cross-section mode.
+        Default: -1 (last pitch angle)
+    mode : str, optional
```
f0uriest, line 3876: do we need this or can it always be inferred from `len(rhos)`?

desc/plotting.py
```diff
@@ -3843,3 +3845,144 @@ def plot_logo(save_path=None, **kwargs):
...
+    rhos : array_like or float, optional
+        Flux surface radii. If float, plots single surface.
+        Default: np.linspace(0.1, 1, 10)
+    alphas : array_like, optional
+        Fieldline label values (toroidal angle).
+        Default: np.linspace(0, 2π, 32, endpoint=True)
+    num_pitch : int, optional
+        Number of pitch angle values for bounce integral calculation.
+        Default: 16
+    pitch_idx : int, optional
```
f0uriest, line 3873: rather than an index can the user specify a particular value of pitch angle to plot?

desc/plotting.py
```diff
@@ -3843,3 +3845,144 @@ def plot_logo(save_path=None, **kwargs):
...
+    ----------
+    eq : object
+        Equilibrium object containing magnetic field information
+    rhos : array_like or float, optional
+        Flux surface radii. If float, plots single surface.
+        Default: np.linspace(0.1, 1, 10)
+    alphas : array_like, optional
+        Fieldline label values (toroidal angle).
+        Default: np.linspace(0, 2π, 32, endpoint=True)
+    num_pitch : int, optional
```
f0uriest, line 3870: if doing the usual contour vs rho/alpha at a single pitch, do we still need to calculate for multiple pitch values?

desc/plotting.py
```diff
@@ -3843,3 +3845,144 @@ def plot_logo(save_path=None, **kwargs):
...
+    Returns
+    -------
+    fig : matplotlib.figure.Figure
+        The figure object containing the plot
+    ax : matplotlib.axes.Axes
+        The axes object for further customization
+
+    Notes
+    -----
+    The second adiabatic invariant J_|| is calculated by integrating the parallel
```
f0uriest, line 3889: may also want to note our convention for handling multiple wells is to sum them

desc/plotting.py
f0uriest: is it because rho is a python float type rather than an array of size 1 maybe? Python scalars sometimes cause some weird stuff in jax.

desc/plotting.py
```diff
@@ -3843,3 +3845,144 @@ def plot_logo(save_path=None, **kwargs):
...
+        inv_pitch = np.linspace(minB, maxB, num_pitch)
+
+        # Create figure and prepare colormap
+        fig = plt.figure(figsize=(6, 5))
+        cmap = plt.get_cmap("plasma").copy()
+        cmap.set_under("white")  # Make values below vmin display as white
+
+        # Plot J_|| as function of α and 1/λ
+        extent = [inv_pitch.min(), inv_pitch.max(), alphas.min(), alphas.max()]
+        plt.imshow(
```
f0uriest, line 3938: prefer using the axis api

tests/baseline/test_plot_adiabatic_invariant_cross_section.png
f0uriest, line 1: any sense of what causes the broken contours eg on the bottom outermost contour?

desc/plotting.py
```diff
@@ -3843,3 +3845,144 @@ def plot_logo(save_path=None, **kwargs):
...
+            data_full[0],
+            origin="lower",  # α increases upward
+            extent=extent,
+            aspect="auto",  # allows aspect ratio to adjust
+            cmap=cmap,
+            interpolation="nearest",  # no smoothing
```
f0uriest, line 3944: why no smoothing? Would it be better as a contour plot?

desc/plotting.py
```diff
@@ -3843,3 +3845,144 @@ def plot_logo(save_path=None, **kwargs):
...
 
     return fig, ax
+
+
+def plot_adiabatic_invariant(
+    eq, rhos=None, alphas=None, num_pitch=None, pitch_idx=-1, mode="single-surface"
+):
+    """Plot the second adiabatic invariant J_|| (parallel action).
+
+    This function visualizes the parallel action invariant J_|| in either of two modes:
```
f0uriest, line 3855: I could see possibly splitting this into 2 different functions for the cases below, similar to how we have our existing `plot_section` vs `plot_1d/2d/3d` etc.

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,260 @@ def _compute_spline(self, params, constants=None):
...
         return grid.compress(data[self._key])
+
+
+class maxJ(_Objective):
+    """Proxy for TEM(Trapped Electron Mode) stability.
+
+    TEMs are excited when a drift wave resonates with the precession of the electrons
+    along a flux surface. To avoid this resonance and further destabilitzation of
+    electrons, we want to ensure that the direction of precession of electrons is
+    opposite to that of the drift waves. This property is called max-J for some
```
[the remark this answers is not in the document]
rahulgaur104: dJ = dJ/ds ds + dJ/dalpha dalpha + J_E dE
Max J => dJ = 0 which would need both J_s and J_alpha but this objective
* only calculates J_s, and
* tries to minimize it (achieve << 0), not set to 0
* and has absolutely no dependence on J_alpha (or omnigenity). 
Hence the "some reason".

Also, a favorable J_s should be as low as possible throughout the volume. TEMs, in fact, would probably peak near the edge.

I guess it is related to neoclassical transport of energetic ions as J_s is one of the terms used in GammaC. Don't know about bootstrap.

desc/compute/_fast_ion.py
rahulgaur104: Ok, I'll make that change

desc/objectives/_fast_ion.py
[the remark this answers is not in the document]
f0uriest: > One would assume from the name that Max J => dJ = 0

Not sure I follow this? "max J" just means dJ/ds < 0, which implies a local maximum on axis (for all energies), which feels consistent with the name. It doesn't say anything about dJ/dalpha away from the axis (though its usually combined with omingenity dJ/dalpha=0)
image

desc/plotting.py
rahulgaur104: No, the idea is to show that gradient of J in alpha is actually small and doesn't seem small because of interpolation that happens inside plt.contour

desc/plotting.py
rahulgaur104: Ok, will change that.

desc/plotting.py
rahulgaur104: Good point.

desc/plotting.py
rahulgaur104: That could be confusing for the user. They would have to find the |B| on that surface, invert it to get the pitch angle and then pick a  value. What we can do is give them a variable t (like Caoxiang's paper) that goes from 0-> 1 linearly covering values from the lowest pitch to the highest

desc/objectives/_fast_ion.py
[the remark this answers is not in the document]
rahulgaur104: Please read my comment again. For a smooth 2D function X to be maximum, at the point of maximum, a necessary not sufficient condition is dX = 0. It's a local condition.
I am saying that what maxJ => dJ = 0 should mean that and yet it does not. Anyone who knows calculus would agree.  Again:

* The paper says max on-axis which is a global statement even though in practice it's a local property. I can easily call it min J and minimum on edge.
* Sufficient condition is never applied d^2J < 0
* maxJ knows nothing about J_alpha so dJ may or may not be zero. What J_s is doing is just trying to get as small a dJ as possible. 
* Even after optimizing for omnigenity J_alpha is never that small from(unless you have precise QS) and dalpha is finite. 

If you don't see how this is a misnomer, especially to a new grad student or an outsider, there's nothing more I can say to change your mind and we'll have to take a vote at the group meeting.

desc/objectives/_fast_ion.py
unalmis: No it does not. For example this will fail if I give a grid with multiple surfaces since the objective compute returns a single value.

desc/compute/_fast_ion.py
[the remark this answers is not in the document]
unalmis: Reminder for myself to check once I am pinged for final review

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,260 @@ def _compute_spline(self, params, constants=None):
             **self._hyperparam,
         )
         return grid.compress(data[self._key])
+
+
+class maxJ(_Objective):
```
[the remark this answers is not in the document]
rahulgaur104: Done. Now it MaximumJ

desc/objectives/_fast_ion.py
```diff
@@ -359,3 +360,260 @@ def _compute_spline(self, params, constants=None):
...
+    X : int
+        Poloidal Fourier grid resolution to interpolate the poloidal coordinate.
+        Preferably rounded down to power of 2.
+    Y : int
+        Toroidal Chebyshev grid resolution to interpolate the poloidal coordinate.
+        Preferably rounded down to power of 2.
+    Y_B : int
+        Desired resolution for algorithm to compute bounce points.
+        Default is double ``Y``. Something like 100 is usually sufficient.
+        Currently, this is the number of knots per toroidal transit over
```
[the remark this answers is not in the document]
rahulgaur104: Well, it is applicable but the description is from the time we were using splines so I'll change it.

Issue #1371, opened 2024-11-15
Normalization of plot_2d()


mohawk811:
`plot_2d(eq_out,'J',log=False, norm_F=True)`

Is there a way to normalize to grad B and not grad B^2?

<img width="338" alt="Screenshot 2024-11-15 at 12 41 29 PM" src="https://github.com/user-attachments/assets/441bb467-430f-4037-94a5-8731e66471e3">

YigitElma:
You can use `norm_name` argument to adjust the normalization. Something like,

```python
plot_2d(eq,'J',log=False, norm_F=True, norm_name="grad(|B|)")
```
But this might not give proper results, since (as far as I know) we don't have a compute function for `"|grad(|B|)|"` or something similar.

mohawk811:
Is there a way to still have the 1/2mu0 ?

YigitElma:
If you want to use `plot_2d` function, I don't think you can because we don't have a compute function for $|\nabla B|/2\mu_0$. If you want you can write your own compute function following [this page](https://desc-docs.readthedocs.io/en/latest/adding_compute_funs.html). But probably the easiest way is to compute $|\nabla B|$ and then divide it by $\mu_0$ and plot it using `plt.contourf` instead of `plot_2d`.

dpanici:
Yea I recommend just computing the stuff yourself and plotting yourself if you need a custom normalization. You can use the `return_data` flag of `plot_2d` if you want to use that to get the R,Z and |F| instead of doing it yourself, and you can look at the source for `plot_2d` if you need to check how the reshaping needs to be to plot the data.

YigitElma:
I believe this issue is resolved. If you should either,

- write your own compute function
- use existing compute functions to calculate your normalization, and use `plt.contourf`

Feel free to re-open, if you think those won't solve your problem.

PR #1683, opened 2025-04-12
Add support for Multiple things to `plot_1d`, `plot_fsa` and Make `norm_F`=True by default

YigitElma:
- Adds support to pass multiple same class objects to `plot_1d` and `plot_fsa`, this changes the values of the returned dictionaries to be list.
- Sets `norm_F` default value to True, if `name="|F|"`. Also, I renamed it to `normalize` since I think it would make it clear that users can pass any arbitrary normalization. 
- Adds `norm_name` to docstring. This was already supported but not well documented.


desc/plotting.py
```diff
@@ -1173,6 +1209,9 @@ def plot_fsa(  # noqa: C901
         Vacuum equilibria are normalized by the volume average of the gradient
         of magnetic pressure, while finite beta equilibria are normalized by the
         volume average of the pressure gradient.
+    norm_name : str, optional
```
YigitElma, line 1212: If people are Ok with it, I suggest renaming `norm_F` with `normalize` because this way it looks like it only supports normaizing |F|, but a user can specify any computable quantity to it through `norm_name`

desc/plotting.py
```diff
@@ -1366,10 +1458,12 @@ def plot_section(
         Grid of coordinates to plot at.
     log : bool, optional
         Whether to use a log scale.
-    norm_F : bool, optional
-        Whether to normalize a plot of force error to be unitless.
-        Vacuum equilibria are normalized by the gradient of magnetic pressure,
-        while finite beta equilibria are normalized by the pressure gradient.
+    normalize : bool, optional
+        Whether to normalize a plot to be unitless. If `name`=`|F|`, defaults to
```
YigitElma, line 1462: Previous docstring might be a bit misleading since the normalization is always based on the volume averaged magnetic pressure gradient. 

Also, I think I should remove `to be unitless` part since, depending on the normalization supplied by the user, this might not be true.

Depending on your feedback, I will replace the other docstrings of the same keyword.

tests/baseline/test_plot_basis_chebyshevdoublefourier.png
YigitElma, line 1: When I plot this on a notebook, it looks fine but pytest creates this image with bunch of intersections

dpanici: is the size of the image too small or something for the pytest?

f0uriest (changes requested):
Not sure I see the motivation/need for a lot of these changes.

desc/plotting.py
YigitElma: `norm_name` was already implemented; I just added it to the docstring. Previously in #1371, I realized that this existed but was not documented. So, given that a user can call `plot_section(eq, "|B|", norm_F=True, norm_name="|F|")` or something like that, the name of this normalisation being always `norm_F` makes it hidden imo.

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ Optionally can also contract the profiles of the original ``Equilibrium`` so tha
 - Allows Redl compute quantities to use SplineProfile
 - Updates Redl bootstrap current consistency tutorial to include a ``SplineProfile`` optimization
 - Adds automatically generated header file showing date the input file was created with `desc.vmec.VMECIO.write_vmec_input`
+- Multiple plotting related changes:
+    - `plot_1d` and `plot_fsa` can now take in multiple objects to plot for comparisons.
+    - `plot_section` when called with `name=|F|` is now defaults to show the normalized force error. To get the old behavor, user needs to pass in `normalize=False`.
```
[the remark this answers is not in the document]
YigitElma: This was initially requested by @dpanici. But I also agree with him. Calling `plot_section` with `"|F|"` is pretty common, and we usually want the normalized error. Yes, the docstring looks a bit complicated but I guess overall, this is helpful automation.

desc/plotting.py
```diff
@@ -591,43 +601,64 @@ def plot_1d(eq, name, grid=None, log=False, ax=None, return_data=False, **kwargs
...
-    data, ylabel = _compute(eq, name, grid, kwargs.pop("component", None))
-    label = kwargs.pop("label", None)
-
     fig, ax = _format_ax(ax, figsize=kwargs.pop("figsize", None))
 
-    # reshape data to 1D
-    data = data.flatten()
-    linecolor = kwargs.pop("linecolor", colorblind_colors[0])
+    ldata = []  # list of data to return
+    color = kwargs.pop("color", colorblind_colors[: len(eq)])
```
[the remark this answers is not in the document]
YigitElma: I don't see the problem. `color` gets a default value, the `parse_argname_change` changes it to the one supplied by the user and warns. 
image

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ Optionally can also contract the profiles of the original ``Equilibrium`` so tha
 - Allows Redl compute quantities to use SplineProfile
 - Updates Redl bootstrap current consistency tutorial to include a ``SplineProfile`` optimization
 - Adds automatically generated header file showing date the input file was created with `desc.vmec.VMECIO.write_vmec_input`
+- Multiple plotting related changes:
+    - `plot_1d` and `plot_fsa` can now take in multiple objects to plot for comparisons.
```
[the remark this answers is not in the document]
YigitElma: I think this is useful, and the complications are just making the arguments lists, the main plotting body is literally the same, just for looped. The part that annoys me with the for loop of `plot_1d`s is that you need to pass in `ax` etc, and for quick checks this is tedious. We can discuss during dev meeting, if people are against it I can revert.

dpanici:
- add deprecation for norm_F 
- add new compute quantity for |F|_normalized (for this, specifcy what the norm is so like |F|/ |grad(B^2)| or wahtever
- norm_name document and if not None, normalizes by that quanitty

desc/plotting.py
```diff
@@ -712,6 +711,34 @@ def plot_2d(
...
+            )
+        )
+        if normalize is None and norm_F:
+            name = "|F|_normalized"
+        elif normalize is not None and norm_F:
+            raise ValueError(
+                "Cannot use both norm_F and normalize keyword arguments at "
+                + "the same time."
+            )
+    # writing a compute function for this quantity is buggy because of the logic
```
YigitElma, line 732: I previously added a compute function for this 
```python
@register_compute_fun(
    name="|F|_normalized",
    label="|\\mathbf{J} \\times \\mathbf{B} - \\nabla p|/\\langle "
    + "|\\nabla |B|^{2}/(2\\mu_0)| \\rangle_{vol}",
    units="",
    units_long="",
    description="Magnitude of force balance error normalized by volume averaged "
    + "magnetic pressure gradient",
    dim=1,
    params=[],
    transforms={},
    profiles=[],
    coordinates="rtz",
    data=["|F|", "<|grad(|B|^2)|/2mu0>_vol"],
)
def _Fmag_normalized(params, transforms, profiles, data, **kwargs):
    data["|F|_normalized"] = data["|F|"] / jnp.nanmean(
        jnp.abs(data["<|grad(|B|^2)|/2mu0>_vol"])
    )
    return data
```

But because of logic https://github.com/PlasmaControl/DESC/blob/64cf78d472991abf83a0937b2af2c8af623ab149/desc/equilibrium/equilibrium.py#L987, the computation doesn't happen with default grid. `eq.compute` part is very messy and I don't want to break anything else, if you have cleaner solution let me know.

ddudt:
I vote to remove all unnecessary calls to `map_coordinates` from the plotting functions. Presently these calls do not do anything since $\zeta=\phi$, and they are preventing us from properly plotting some quantities. We can make a note to resolve this in #568 .

YigitElma:
We talked about the `|F|_normalized` issue during dev meeting with @ddudt. There are 2 possible fixes for that:

1. Remove the 
```python
        coords = map_coordinates(
            eq,
            grid.nodes,
            ["rho", "theta", "phi"],
            ["rho", "theta", "zeta"],
            period=(np.inf, 2 * np.pi, 2 * np.pi),
            guess=grid.nodes,
        )
        grid = Grid(coords, sort=False)
```
part from plotting, to preserve the grid spacing for surface integration etc. Currently, the current `Grid` might be preventing plotting of some quantities (no matter if the user supplies a grid or not)

2. The hacky way of not having a compute function for `|F|_normalized` but instead set the proper normalisation string internally
```python
if name == "|F|_normalized":
    name = "|F|"
    normalize = "<|grad(|B|^2)|/2mu0>_vol" 
```

I think we prefer the first option, but this will require a proper fix to the compute logic. The main reason is mentioned above. Note: `eq.compute("|F|_normalized")` works, but calling this with custom `Grid` doesn't. I think the 0D 1D dependency check needs to be more involved.

desc/plotting.py
```diff
@@ -1511,7 +1545,11 @@ def plot_section(
     ), f"plot section got unexpected keyword argument: {kwargs.keys()}"
 
     cax_kwargs = {"size": "5%", "pad": 0.05}
-
+    units = (
+        f"(${data_index["desc.equilibrium.equilibrium.Equilibrium"][name]["units"]}$)"
```
YigitElma, line 1549:
```suggestion
        f"(${data_index['desc.equilibrium.equilibrium.Equilibrium'][name]['units']}$)"
```

desc/plotting.py
```diff
@@ -759,14 +779,11 @@ def plot_2d(
     if log:
         data = np.abs(data)  # ensure data is positive for log plot
         contourf_kwargs["norm"] = matplotlib.colors.LogNorm()
-        if norm_F:
-            contourf_kwargs["levels"] = kwargs.pop("levels", np.logspace(-6, 0, 7))
```
f0uriest, line 763: I'd prefer to keep this logic when `name="|F|_normalized"`, as it gives a consistent picture (red=bad, blue=good). Otherwise if the error is flat it can make the colors look weird.

desc/plotting.py
```diff
@@ -1432,38 +1489,18 @@ def plot_section(
...
-            eq,
-            grid.nodes,
-            ["rho", "theta", "phi"],
-            ["rho", "theta", "zeta"],
-            period=(np.inf, 2 * np.pi, 2 * np.pi),
-            guess=grid.nodes,
-        )
-        grid = Grid(coords, sort=False)
+        # TODO(#568): Add conversion to rho, theta, zeta grid
+        # Using custom Grid Class will cause problems for "|F|_normalized"
```
f0uriest, line 1493: what issue does this cause? we're just calling `eq.compute` so it should handle arbitrary grids

desc/plotting.py
```diff
@@ -3114,7 +3140,6 @@ def plot_qs_error(  # noqa: 16 fxn too complex
         * ``legend_kw``: dict, any keyword arguments to be passed to ax.legend()
         * ``xlabel_fontsize``: float, fontsize of the xlabel
         * ``ylabel_fontsize``: float, fontsize of the ylabel
-        * ``labels``: list of strs of length 3, labels to apply to each QS error metric
```
f0uriest, line 3117: why remove this?

YigitElma: It is a duplicate of line 3110

desc/plotting.py
YigitElma: I assume the main reason for this was that 1e-2 error is good in general. The only problem I have with this is, when the error is above or below the default range, it just clips the color to red and dark blue, but in fact, the error might be 1e2 or 1e-9 etc. 

I can add it back, we talked about this last week @ddudt, he was ok with this change.

desc/plotting.py
YigitElma: Normal `eq.compute` works for `|F|_normalized` but if you call it with this specific grid that doesn't have the spacing and weight etc, it cannot compute due to some internal logic which I mention https://github.com/PlasmaControl/DESC/pull/1683#discussion_r2053023178 here

desc/plotting.py
dpanici: I'd prefer adding back, and I'd like the docstring to mention if normalized quantitiy is plotted it defaults to these levels

tests/test_plotting.py
```diff
@@ -448,7 +449,8 @@ def test_section_logF(self):
     def test_plot_normF_section(self):
         """Test Poincare section plot of normalized force on log scale."""
         eq = get("DSHAPE_CURRENT")
-        fig, ax = plot_section(eq, "|F|", norm_F=True, log=True)
+        levels = np.logspace(-6, 0, 7)
+        fig, ax = plot_section(eq, "|F|_normalized", log=True, levels=levels)
```
ddudt, line 453: `levels` is not needed here since it is the same as the default

desc/plotting.py
```diff
@@ -685,7 +684,9 @@ def plot_2d(
         * ``xlabel_fontsize``: float, fontsize of the xlabel
         * ``ylabel_fontsize``: float, fontsize of the ylabel
         * ``cmap``: str, matplotlib colormap scheme to use, passed to ax.contourf
-        * ``levels``: int or array-like, passed to contourf
+        * ``levels``: int or array-like, passed to contourf.
+          If ``name``="|F|_normalized" and ``log``==True, default is
+          np.logspace(-6, 0, 7). Pass None to use actual data min/max.
```
ddudt, line 689:
```suggestion
        * ``levels``: int or array-like, passed to contourf.
          If ``name="|F|_normalized"`` and ``log=True``, default is
          ``np.logspace(-6, 0, 7)``. Otherwise the default (``None``) uses the min/max
          values of the data.
```

desc/plotting.py
```diff
@@ -712,6 +713,29 @@ def plot_2d(
...
+        warnings.warn(
+            FutureWarning(
+                "Argument norm_F has been deprecated. If you are trying to "
+                + "normalize |F| by magnetic pressure gradient, use  "
+                + "`name=|F|_normalized` instead. If you want to normalize by "
+                + "another quantity, use the `normalize` keyword argument."
+            )
+        )
+        if normalize is None and norm_F:
+            name = "|F|_normalized"
```
ddudt, line 728: The way this is now, the user could do something dumb like `plot(eq, name="sqrt(g)", norm_F=True)` and it would end up plotting `"|F|_normalized"` for them. If we want to preserve the old behavior, we should actually be plotting `"sqrt(g)" / "<|grad(|B|^2)|/2mu0>_vol"`

desc/plotting.py
```diff
@@ -712,6 +713,29 @@ def plot_2d(
...
+        elif normalize is not None and norm_F:
+            raise ValueError(
+                "Cannot use both norm_F and normalize keyword arguments at "
+                + "the same time."
+            )
```
ddudt, line 733:
```suggestion
        if normalize is not None and norm_F:
            raise ValueError(
                "Cannot use both normalize and norm_F keyword arguments."
            )
        if norm_F:
            normalize = "<|grad(|B|^2)|/2mu0>_vol"
```

desc/plotting.py
ddudt: See suggestion below

desc/plotting.py
```diff
@@ -1389,7 +1428,9 @@ def plot_section(
         * ``xlabel_fontsize``: float, fontsize of the xlabel
         * ``ylabel_fontsize``: float, fontsize of the ylabel
         * ``cmap``: str, matplotlib colormap scheme to use, passed to ax.contourf
-        * ``levels``: int or array-like, passed to contourf
+        * ``levels``: int or array-like, passed to contourf.
+          If ``name``="|F|_normalized" and ``log``==True, default is
+          np.logspace(-6, 0, 7). Pass None to use actual data min/max.
```
ddudt, line 1433: see suggestion from `plot_2d`

desc/plotting.py
```diff
@@ -1415,14 +1456,37 @@ def plot_section(
...
+                + "another quantity, use the `normalize` keyword argument."
+            )
+        )
+        if normalize is None and norm_F:
+            name = "|F|_normalized"
+        elif normalize is not None and norm_F:
+            raise ValueError(
+                "Cannot use both norm_F and normalize keyword arguments at "
+                + "the same time."
+            )
```
ddudt, line 1476: see suggestion from `plot_2d`

ddudt (changes requested):

desc/plotting.py
YigitElma: Ok, I made the change to preserve the old behavior. Those kind of cases were the primary reason I renamed the argument

desc/plotting.py
f0uriest: I think we should figure out why it isn't working. If we can compute `|F|` on a custom grid, and `<B^2>`, then we should also be able to compute `|F|/<B^2>` without it breaking.

f0uriest: The issue seems to be that iota is needed by both `|F|` and `<B^2>`, but this line https://github.com/PlasmaControl/DESC/blob/7b57d38db0d2480716e34b871dc887cd3fad10ac/desc/equilibrium/equilibrium.py#L991 basically makes it compute iota on the correct grid when computing `<B^2>` but then doesn't realize it also needs to use a special grid to compute iota for `|F|`

I *think* we can safely get rid of the check in that linked line, at the expense of some occasional extra compute work, but would appreciate someone else taking a look at it, I think @unalmis might know more.

YigitElma: I don't know what @unalmis's intention was for this logic, but https://github.com/PlasmaControl/DESC/blob/64cf78d472991abf83a0937b2af2c8af623ab149/desc/equilibrium/equilibrium.py#L987 to line 1011 makes some checks for which quantity requires a new grid to compute. The problem is when someone computes just `|F|`, iota stuff gets such a new grid. Similar thing happens with the volume averaged stuffto compute volume, but when you try to compute both an `iota` related dependency and a volume related dependency, only one of them gets this special grid, and the other is expected to be computed at the end with the original grid.

I also agree that if we can do,
```python
Fmag = eq.compute("|F|", grid=grid)
Bvol = eq.compute("<|B|>", grid=grid)
out = Fmag/Bvol
```
the new compute quantity that basical just divides them should be also computable. But I won't change that logic until @unalmis tells me his reason in the first place

f0uriest: I'd like to figure this out before merging

unalmis: Ok summarizing: I made a 2 word edit to Rory's comment in this chain to describe the issue. To link other comments, in the language of https://github.com/PlasmaControl/DESC/pull/1024#discussion_r1663361996, the issue is `C` fails to get computed in the last sentence because it requires a special grid. 

I do not like the suggested patch [here](https://github.com/PlasmaControl/DESC/pull/1024#discussion_r2052993674) because for the reasons already documented in this code on the master branch.

The solution is to make this [filter](https://github.com/PlasmaControl/DESC/blob/e04234dfbc24fcd73b3493aa434cd1c0d7731872/desc/equilibrium/equilibrium.py#L990) weaker. My recommendation is to exclude `iota` from `just_dep0d_deps`.

YigitElma:
This function may be useful for visualizing the dependency trees

```python

from anytree import Node, RenderTree

def find_node_by_path(root, path):
    """Find a node in the tree by '/'-separated path."""
    parts = path.strip("::").split("::")
    node = root
    for part in parts:
        matches = [child for child in node.children if child.name == part]
        if not matches:
            raise ValueError(f"Path component '{part}' not found under '{node.name}'")
        node = matches[0]
    return node

def build_dependency_tree(name, data_index, p, visited=None, parent=None, contract_circular=True):
    if visited is None:
        visited = set()
    if contract_circular:
        if name in visited:
            return Node(f"{name} (circular)", parent=parent)
    
        visited.add(name)
    node = Node(name, parent=parent)

    deps = data_index[p][name]["dependencies"]["data"]
    for dep in deps:
        build_dependency_tree(dep, data_index, p, visited, parent=node, contract_circular=contract_circular)
    return node

def print_dependency_tree(name, data_index, p, contract_circular=True, subpath=None):
    root = build_dependency_tree(name, data_index, p, contract_circular=contract_circular)
    start_node = root
    if subpath:
        try:
            start_node = find_node_by_path(root, subpath)
        except ValueError as e:
            print(f"Error: {e}")
            return
    for pre, fill, node in RenderTree(start_node):
        print(f"{pre}{node.name}")

    return root

p = "desc.equilibrium.equilibrium.Equilibrium"
name = "sqrt(g)"

# subpath = "effective ripple 3/2::min_tz |B|::|B|::|B|^2::B"
subpath = None 
root = print_dependency_tree(name, data_index, p, contract_circular=False, subpath=subpath)
```

The output looks like
```
sqrt(g)
├── e_rho
│   ├── R
│   ├── R_r
│   ├── Z_r
│   └── omega_r
│       └── 0
├── e_theta
│   ├── R
│   ├── R_t
│   ├── Z_t
│   └── omega_t
│       └── 0
└── e_zeta
    ├── R
    ├── R_z
    ├── Z_z
    └── omega_z
        └── 0
```

desc/plotting.py
unalmis: On `master` I wrote this [comment](https://github.com/PlasmaControl/DESC/blob/e81791039d9b54986c5d2179f600aaf26cb0642a/desc/equilibrium/equilibrium.py#L1057). Since the changes also appear complicated, I am a bit skeptical of the correctness. Are you sure it correctly handles quantities like `max(effective ripple)` or `max(effective ripple) * R0` or `mean(effective ripple)*n_rho*A(z)`, and other quantities people will add in the future?

[Rory and Dario wanted to add](https://github.com/PlasmaControl/DESC/pull/1094#discussion_r1818247009) additional logic to `eq.compute` so that it correctly computes things with just `resolution_requirement="t"` (i.e. a single letter rather than two letters). With the edits in this PR, it is not immediately clear to me how to do that.

Basically, I think this logic is too complex to maintain. Maybe if it were the most general solution possible I would not complain, but as mentioned in the comment that seems difficult.

desc/equilibrium/equilibrium.py
```diff
@@ -968,42 +968,77 @@ def need_src(name):
...
+        #       ├── c
+        #       ├── d
+        #       └── e
+        #       b
+        #       ├── f
+        #       └── g
+        # Say,
+        #    - "X" has deps = {"a", "c", "d", "e", "f", "g"}
+        # Since "X" depends on "a", it will have "c", "d", "e" automatically. So, we
+        # can remove them once "a" is computed. Although, "X" does depend on "d" and
```
YigitElma:
```suggestion
        # can remove them once "a" is computed. Although, "X" does depend on "g" and
```

desc/equilibrium/equilibrium.py
```diff
@@ -968,42 +968,77 @@ def need_src(name):
...
+        # "f", it doesn't depend on "b", so we keep "d" and "f".
```
YigitElma:
```suggestion
        # "f", it doesn't depend on "b", so we keep "g" and "f".
```

desc/equilibrium/equilibrium.py
```diff
@@ -968,42 +968,77 @@ def need_src(name):
...
-        just_dep0d_dep = lambda name: name in dep0d_deps and name not in names
+        # dependencies of 0d stuff
+        dep0d_deps = (
+            set(get_data_deps(dep0d, obj=p, has_axis=grid.axis.size, data=data)) | dep0d
+        ) - need_src_deps
+        # find the dependencies that are not dependencies of 0d stuff and any other
+        # special case
+        other_deps = deps - dep0d_deps - set(names) - need_src_deps
+        deps_other_deps = set(
+            get_data_deps(other_deps, obj=p, has_axis=grid.axis.size, data=data)
```
dpanici: recall this function after computing 0D, 1D etc to re-build graph, using the up to date data dict after each

desc/equilibrium/equilibrium.py
```diff
@@ -965,46 +965,33 @@ def need_src(name):
...
-        dep0d_deps = set(
-            get_data_deps(dep0d, obj=p, has_axis=grid.axis.size, data=data)
-        )
-        # This filter is stronger than the name implies, but the false positives
-        # that are filtered out will still get computed with the logic in
-        # compute.utils.compute
-        # https://github.com/PlasmaControl/DESC/pull/1024#discussion_r1663080423.
-        just_dep0d_dep = lambda name: name in dep0d_deps and name not in names
+        for dep in dep0d:
+            dummy_data[dep] = "dummy_data_0d"
```
f0uriest: im confused why this is needed, I thought we were going to move these lines so that it actually computes the 0d stuff, then re-evaluates the needed dependencies, then computes 1d stuff etc. This looks like it's still trying to figure out all the dependencies at the beginning rather than as we go

desc/plotting.py
```diff
@@ -1527,30 +1593,21 @@ def plot_section(
...
-                    + data_index["desc.equilibrium.equilibrium.Equilibrium"][name][
-                        "label"
-                    ]
-                    + "$",
-                    "$"
-                    + data_index["desc.equilibrium.equilibrium.Equilibrium"][norm_name][
-                        "label"
-                    ]
-                    + "$",
+                    "$" + data_index_name[name]["label"] + "$",
```
f0uriest: I think this is wrong, you're doing `data_index[parameterization][name][name][label]` (`name` twice), same on the next line

desc/plotting.py
f0uriest, line 1: there are a lot of lines here that it says are not covered, can you check if they actually aren't and if so add tests to make sure we hit them?

Would also be good to add a test to make sure that passing `norm_F=True` throws the correct warning and still plots the right thing

f0uriest (changes requested):

desc/equilibrium/equilibrium.py
YigitElma: This is equivalent to that (just adding this `dummy_data` dict). The reason is we use `calc1dr` and `calc1dz` which depends on these special dependency lists `dep1dr` and `dep1dz` before calculation of 0D stuff, so we need to duplicate some parts of the code. Yes, this dummy data doesn't look beautiful, but I think it has less code and serve for the same purpose.

desc/plotting.py
YigitElma: I added tests for deprecated options, but rest of the uncovered lines are just `log=True` option or some minor options like that. Also, I think codecov is a bit confused, because `plot_basis` for `ChebyshevDoubleFourier` is definitely tested, but it says that's not. Let's see the new report.

desc/plotting.py
YigitElma: I also want to use this for `normalized` too, so I changed the others, not these ones.

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ Changelog
 
 New Features
 
+- Multiple plotting related changes:
+    - Renames `norm_F` keyword argument to `normalize` and removes `norm_name`. `normalize` is a string to use for normalisation. If you want to get the old behavior `plot_section(eq, "|F|", norm_F=True)`, use instead `plot_section(eq, "|F|_normalized")`.
```
dpanici: or it could be `"|F|", normalize="<|grad(|B|^2)|/2mu0>_vol"`

dpanici:
```suggestion
    - Renames `norm_F` keyword argument to `normalize` and removes `norm_name`. `normalize` is a string to use for normalisation. If you want to get the old behavior e.g. `plot_section(eq, "|F|", norm_F=True)`, use instead plot_section(eq, "|F|", normalize="<|grad(|B|^2)|/2mu0>_vol")` or the new compute quantity `plot_section(eq, "|F|_normalized")`.
```

desc/plotting.py
```diff
@@ -532,7 +532,7 @@ def plot_1d(eq, name, grid=None, log=False, ax=None, return_data=False, **kwargs
         * ``label``: str, label of the plotted line (e.g. to be shown with ax.legend())
         * ``xlabel_fontsize``: float, fontsize of the xlabel
         * ``ylabel_fontsize``: float, fontsize of the ylabel
-        * ``linecolor``: str or tuple, color to use for plot line
+        * ``color``: str or tuple, color to use for plot line
```
dpanici, line 535: to be annoying... we could add normalize to 1D as well, and maybe to plot_3D too (I would find that useful tbh)

dpanici (changes requested):
one minor request for changelog and also, if you dont mind, we can do this for plot_1d and plot_3d as well... but we can also just make an issue to add normalize to plot_1d and plot_3d, I would be fine with that as there are lots of needed changes in this and it is big already.

dpanici (approved):

f0uriest (dismissed):

YigitElma:
@ddudt do you have time for a quick review?

desc/compute/_equil.py
```diff
@@ -525,6 +525,26 @@ def _Fmag(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="|F|_normalized",
+    label="|\\mathbf{J} \\times \\mathbf{B} - \\nabla p|/\\langle "
+    + "|\\nabla |B|^{2}/(2\\mu_0)| \\rangle_{vol}",
+    units="",
+    units_long="",
```
unalmis: for dimensionless variables the other quantities have

units="~",
units_long="None",

desc/equilibrium/equilibrium.py
```diff
@@ -965,46 +965,33 @@ def need_src(name):
...
+        deps_after_0d = set(
+            get_data_deps(names, obj=p, has_axis=grid.axis.size, data=dummy_data)
```
unalmis: should pass in `data=dep0d` instead of `dummy_data`

desc/equilibrium/equilibrium.py
```diff
@@ -965,46 +965,33 @@ def need_src(name):
...
+            + names
+        )
         dep1dr = {
             dep
-            for dep in deps
-            if is_1dr_rad_grid(dep)
-            and not just_dep0d_dep(dep)
-            and dep not in need_src_deps
+            for dep in deps_after_0d
+            if is_1dr_rad_grid(dep) and dep not in need_src_deps
```
unalmis, line 980: This does not work. Example, `iota` is often in `dep0d`. Therefore, this logic will not place `iota` in `dep1dr`. That means most quantities will be computed incorrectly.

desc/equilibrium/equilibrium.py
```diff
@@ -965,46 +965,33 @@ def need_src(name):
...
         }
+        for dep in dep1dr:
+            dummy_data[dep] = "dummy_data_1dr"
+        deps_after_0d_1dr = set(
+            get_data_deps(names, obj=p, has_axis=grid.axis.size, data=dummy_data)
```
unalmis: should pass in `data=dep1dr | dep0d`

desc/equilibrium/equilibrium.py
```diff
@@ -965,46 +965,33 @@ def need_src(name):
...
             dep
-            for dep in deps
-            # By including the additional requirement that dep is not just a dependency
-            # of some scalar (0d) quantity, we are ensuring that we do not unnecessarily
-            # compute things like A(z) when it was only needed to compute R0, as in the
-            # example above.
-            if is_1dz_tor_grid(dep)
-            and not just_dep0d_dep(dep)
-            and dep not in need_src_deps
+            for dep in deps_after_0d_1dr
```
unalmis, line 990: Same issue as above.

unalmis (changes requested):
- [ ] I do not believe the new compute method works. Please let me know.
- [ ] It would be worthwhile to also add a regression test that computes everything one at a time and compares each one to `test_compute_everything_rpz.pkl`.
- [ ] Also left comments to remove dummy data.

desc/equilibrium/equilibrium.py
unalmis: I left comments to remove dummy data.

desc/equilibrium/equilibrium.py
YigitElma: I thought data should be a dictionary, if a list works, I can make the change

desc/equilibrium/equilibrium.py
YigitElma: Why `iota` is in dep0d? It must be in dep1dr, and I see that that is the case.

unalmis: need B or J to compute F average, need iota to compute B or J etc.

dpanici: A quick check of compute "B" and "W_B" on master and this branch and comparing the two for a non-trivial current-constrained equilibrium seem to agree. But worth checking further

unalmis: The point is that this branch leaves that up to pure chance. This branch is only correct if there is something that needs to be computed on a flux grid that somehow is not needed to compute a volume quantity , and this something also needs to have  its dependencies include all flux grid quantities that need to be computed.

unalmis: And then the issue cascades to the cross section grids.

YigitElma: I don't understand what you mean, `iota` can never be in dep0d since `is_0d_vol_grid` check will never return true for it. It doesn't have anything to do with this branch. If a dependency of iota needs a full grid, then they will be computed in proper grid.

desc/equilibrium/equilibrium.py
unalmis: yes it works

desc/equilibrium/equilibrium.py
unalmis: Ok yea it's fine

unalmis (approved):

unalmis:
@YigitElma i am sure you can merge this as you have 3 approvals

YigitElma:
@f0uriest @rahulgaur104 Can one of you approve again?

PR #1684, opened 2025-04-12
Update della install instructions

unalmis:


YigitElma (approved):

dpanici (approved):

dpanici:
Ok so seems we cannot upload this unless we 
- remove our branch protection bypass restriction for admins
- switch the way we interact with codecov to use their [CLI](https://docs.codecov.com/docs/the-codecov-cli) so we can use the ["empty upload" feature](https://docs.codecov.com/docs/cli-options#empty-upload) when certain parts of the code are touched (docs), though it seems to [not be well documented](https://github.com/codecov/feedback/issues/108)

dpanici:
Trying a simple change based off our codecov action docs which seems to have an `empty-upload` feature. we prob wiill have to force this thru still since the yaml on master is not updated but after that we can see if this works

YigitElma:
> Trying a simple change based off our codecov action docs which seems to have an `empty-upload` feature. we prob wiill have to force this thru still since the yaml on master is not updated but after that we can see if this works

If I remember correctly, I tried something similar to this when I first made the change that stopped running unit/regression tests for small PRs. The problem was (as @f0uriest  mentioned back then) it could mess up the coverage, we might see 0% coverage after this. I had tried uploading a fake coverage report with same coverage as master, but it didn't work or something was buggy.

dpanici:
> > Trying a simple change based off our codecov action docs which seems to have an `empty-upload` feature. we prob wiill have to force this thru still since the yaml on master is not updated but after that we can see if this works
> 
> If I remember correctly, I tried something similar to this when I first made the change that stopped running unit/regression tests for small PRs. The problem was (as @f0uriest mentioned back then) it could mess up the coverage, we might see 0% coverage after this. I had tried uploading a fake coverage report with same coverage as master, but it didn't work or something was buggy.

Dang. I guess we have to undo the master restrictions, maybe I will just do it temporarily whenever this pops up

YigitElma (approved):

dpanici (approved):

Issue #1686, opened 2025-04-14
Cumulative Memory Buildup


ddudt:
I've noticed some Out Of Memory issues that don't manifest until after several iterations. So the Jacobian matrix fits in memory initially, but after several iterations it does not anymore. This suggests there is intermediate data accumulating that is probably not needed to keep in memory. 

MWE on Perlmutter:
```
from desc import set_device

set_device("gpu")

from desc.backend import print_backend_info
from desc.examples import get

print_backend_info()

N = 2**4

eq = get("precise_QA")
eq.change_resolution(L=N, M=N, N=N, L_grid=2 * N, M_grid=2 * N, N_grid=2 * N)
eq.resolution_summary()
eq.set_initial_guess()
eq.solve(
    objective="force",
    constraints=None,
    optimizer="lsq-exact",
    ftol=1e-4,
    xtol=1e-6,
    gtol=1e-6,
    maxiter=10,
    x_scale="auto",
    verbose=2,
    copy=False,
)
```
Output:
```
DESC version=0.14.1+204.g35d190015.
Using JAX backend: jax version=0.4.38, jaxlib version=0.4.38, dtype=float64.
Using device: NVIDIA A100-SXM4-80GB (id=0), with 80.00 GB available memory.
Spectral indexing: ansi
Spectral resolution (L,M,N)=(16,16,16)
Node resolution (L,M,N)=(32,32,32)
Building objective: force
Precomputing transforms
Timer: Precomputing transforms = 847 ms
Timer: Objective build = 1.36 sec
Building objective: lcfs R
Building objective: lcfs Z
Building objective: fixed Psi
Building objective: fixed pressure
Building objective: fixed current
Building objective: fixed sheet current
Building objective: self_consistency R
Building objective: self_consistency Z
Building objective: lambda gauge
Building objective: axis R self consistency
Building objective: axis Z self consistency
Timer: Objective build = 884 ms
Timer: LinearConstraintProjection build = 7.40 sec
Number of parameters: 6480
Number of objectives: 37570
Timer: Initializing the optimization = 9.78 sec

Starting optimization
Using method: lsq-exact
   Iteration     Total nfev        Cost      Cost reduction    Step norm     Optimality   
       0              1          3.137e-01                                    3.441e-01   
       1              4          2.103e-02      2.926e-01      6.817e-02      1.095e-01   
       2              5          3.256e-04      2.070e-02      7.363e-02      1.062e-02   
[...]

Issue #1669, opened 2025-04-04
Make more of DESC jax-friendly


f0uriest:
There are certain portions of the code that currently don't work with jax but could likely be made to. The main reason things don't work with jax is because of error checking, but this could be done with `equinox.error_if` in a jax compatible way.

A few places / things that would be most useful
- `Equilibrium.compute`
- Non-custom grid creation
- `Objective.build`
- `vmec_utils` for converting between different fourier conventions
- perturbations
- equilibrium solve

YigitElma:
By making Jax friendly, do you mean making them jittable? Or something else?

f0uriest:
the usual `jit`/`vmap`/`grad` etc

dpanici:
After #1688 , we will have some in-place operations which may not actually be in-place unless the optimize functions are placed under `jit`, the files affected are

https://github.com/PlasmaControl/DESC/blob/master/desc/optimize/fmin_scalar.py
https://github.com/PlasmaControl/DESC/blob/master/desc/optimize/least_squares.py
https://github.com/PlasmaControl/DESC/blob/master/desc/optimize/aug_lagrangian.py
https://github.com/PlasmaControl/DESC/blob/master/desc/optimize/aug_lagrangian_ls.py

PR #1688, opened 2025-04-14
Reduce storage of big arrays in memory during optimization

dpanici:
Reuses the J variable in optimizers to reduce the amount of Jacobian-sized matrices held in memory. This can lead to tangible memory decreases (at L=M=N=16 fixed bdry solve with default grids, `J` is 4GB) which can be observed after the significant memory jump after the first Jacobian calculation.

- Updates docs to give better memory usage insight
- Fixes some benchmark run conditions. This should make the memory profiler run with fork pull requests, also run benchmark when `run_benchmarks` label is added.


YigitElma:
Some vram profiling on my laptop GPU with 12GB memory. The used script:
```python
import sys
import os

sys.path.append(os.path.abspath("../../"))


os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"

from desc import set_device

set_device("gpu")

from desc.backend import print_backend_info
from desc.examples import get

print_backend_info()

N = 14

eq = get("precise_QA")
eq.change_resolution(L=N, M=N, N=N, L_grid=2 * N, M_grid=2 * N, N_grid=2 * N)
eq.resolution_summary()
eq.set_initial_guess()
obj = ObjectiveFunction(ForceBalance(eq), jac_chunk_size=500, deriv_mode="batched")
obj.build()
print(f"Objective function deriv mode: {obj._deriv_mode}")
print(f"Objective function chunk size: {obj._jac_chunk_size}")

eq.solve(
    objective=obj,
    constraints=None,
    optimizer="lsq-exact",
    ftol=1e-4,
    xtol=1e-6,
    gtol=1e-6,
    maxiter=5,
    x_scale="auto",
    verbose=2,
    copy=False,
)
```
# Master
test-case-res14-master

# PR
test-case-res14-with1688

YigitElma:
In fact, we can even delete `Q,R` after Newton step since the subproblem solver uses a new QR decomposition with augmented stuff. This reduces the peak memory, even though in the next step we still form a new QR and delete it again.

test-case-res14-with1688-delQR

desc/optimize/least_squares.py
```diff
@@ -423,7 +435,7 @@ def lsqtr(  # noqa: C901
         fun=f,
         grad=g,
         v=v,
-        jac=J,
+        jac=J * 1 / d,  # after overwriting J_h, we have to revert back
```
YigitElma, line 438: currently, we delete this from result output but I will make some changes in #1623 to use that

dpanici: ah thanks for catching this

ddudt (commented):
This is great! I am in favor of making similar changes to the other optimizers in this PR, since it isn't an urgent fix.

YigitElma:
To be clear though, this doesn't resolve the memory leak issue #1686 but prevents storing big arrays at the same time.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Optionally can also contract the profiles of the original ``Equilibrium`` so tha
 - Updates Redl bootstrap current consistency tutorial to include a ``SplineProfile`` optimization
 - Adds automatically generated header file showing date the input file was created with `desc.vmec.VMECIO.write_vmec_input`
 
+Performance Improvements
+
+- Couple memory management improvements to reduce the base memory used during optimization.
```
dpanici, line 16:
```suggestion
- Improves memory management to reduce the base memory used during optimization while using `lsq-exact`, `lsq-auglag` and `fmin-auglag` optimizers.
```

f0uriest:
Yeah I don't see how this actually fixes the underlying issue in #1686. From the plots above it looks like this doesn't get rid of those spikes which are likely the cause of the OOM after a few iterations, we should figure out what's actually causing that.

f0uriest:
Also the plots only show a reduction of ~500 MB, shouldn't it be a lot more?

YigitElma:
@f0uriest The last plot is with the most recent changes, and there, the memory decrease is ~2GB. The peaks are the Jacobian evaluations, and those can be flattened out by `jac_chunk_size=1` (for my test case it was 500). 

I think #1686 was at the edge of the memory limit and some small leak like 20-40 MB according to @dpanici's further analysis caused the OOM. We realized these changes while looking at that issue but this PR is not for solving that issue.

dpanici:
Do it in place
garbage collection manually in btwn steps

dpanici:
check for `fmintr` as well

YigitElma:
I shared some profiling 
https://github.com/PlasmaControl/DESC/issues/1686#issuecomment-2814372946
https://github.com/PlasmaControl/DESC/issues/1686#issuecomment-2814409983

YigitElma:
I ran the profiling again before and after Garbage collection, it looks like during the optimization there is no change but, it basically clears some earlier stuff. It looks like for lower resolutions garbage collection has some speed effect, maybe we can just run it once before optimization?

test-case-memory-gpucpu-bagc

YigitElma:
Also, just for record here are the profiling scripts.
[Profiling.zip](https://github.com/user-attachments/files/19807684/Profiling.zip)

dpanici:
Yea maybe we don't `gc` if it seems to have negligible effect. The in-place stuff if it does not affect speed, we can keep in and then when #1669 is eventually done, we will actually see the benefits then.

YigitElma (dismissed):

desc/optimize/aug_lagrangian.py
```diff
@@ -342,7 +346,18 @@ def laghess(z, y, mu, *args):
     diag_h = g * dv * scale
 
     g_h = g * d
-    H_h = d * H * d[:, None]
+    if H_is_numpy_array:
```
f0uriest, line 349: why is this needed? the `*=` operator works fine for jax arrays as well

desc/optimize/aug_lagrangian.py
```diff
@@ -342,7 +346,18 @@ def laghess(z, y, mu, *args):
...
+        # doing operation H = d * H * d[:, None]
+        # with just in-place operations
+        H *= d[:, None]
+        H *= d
+    else:
+        H = H.at[:].set(d * H * d[:, None])
```
f0uriest, line 355: note that this won't actually be done in place unless its inside jit

desc/optimize/aug_lagrangian_ls.py
```diff
@@ -531,7 +539,11 @@ def lagjac(z, y, mu, *args):
             d = v**0.5 * scale
             diag_h = g * dv * scale
             g_h = g * d
-            J_h = J * d
+            J = J.at[:].set(J * d)
+            # we don't need unscaled J anymore this iteration, so we overwrite
+            # it with J_h = J * d to avoid carrying so many J-sized matrices
+            # in memory, which can be large
+            J_h = J
```
f0uriest, line 546: you may need to `del J` after this i think

desc/optimize/fmin_scalar.py
```diff
@@ -239,7 +239,20 @@ def fmintr(  # noqa: C901
...
+    if callable(hess):  # H is a jax.numpy array
+        H = H.at[:].set(d * H * d[:, None])
+    else:
+        # H is a numpy array since comes from BFGS, so can just do in-place
+        # operation
+        # doing operation H = d * H * d[:, None]
+        # with just in-place operations
+        H *= d[:, None]
+        H *= d
+    H_h = H
```
f0uriest, line 254: `del H`

desc/optimize/aug_lagrangian_ls.py
YigitElma: This looks to be working. I think this line just creates an alias for `J`

desc/optimize/fmin_scalar.py
dpanici: I feel like it may not be necessary but I can add it anyways

desc/optimize/aug_lagrangian.py
```diff
@@ -342,7 +343,19 @@ def laghess(z, y, mu, *args):
     diag_h = g * dv * scale
 
     g_h = g * d
-    H_h = d * H * d[:, None]
+
+    # doing operation H = d * H * d[:, None]
+    # with just in-place operations
+    # Note that this will not be in place with JAX unless
+    # this function is placed under JIT
```
YigitElma, line 350: since this one is not jitted, I guess having this comment is unnecessary

desc/optimize/aug_lagrangian.py
```diff
@@ -581,7 +594,17 @@ def laghess(z, y, mu, *args):
             d = v**0.5 * scale
             diag_h = g * dv * scale
             g_h = g * d
-            H_h = d * H * d[:, None]
+
+            # doing operation H = d * H * d[:, None]
+            # with just in-place operations
```
YigitElma, line 599: same as above

desc/optimize/aug_lagrangian.py
YigitElma: The comments below look better
```
# we don't need unscaled J anymore, so we overwrite
    # it with J_h = J * d to avoid carrying so many J-sized matrices
    # in memory, which can be large
    # Note that this will not be in place with JAX unless
    # this function is placed under JIT
```

desc/optimize/fmin_scalar.py
```diff
@@ -239,7 +239,16 @@ def fmintr(  # noqa: C901
     diag_h = g * dv * scale
 
     g_h = g * d
-    H_h = d * H * d[:, None]
+    # we don't need unscaled H anymore this iteration, so we overwrite
+    # it with H_h = d * H * d[:, None] to avoid carrying so many H-sized matrices
+    # in memory, which can be large
+    # doing operation H = d * H * d[:, None]
+    # with just in-place operations
```
YigitElma, line 246: again, this comment is unncessary

desc/optimize/fmin_scalar.py
```diff
@@ -413,7 +422,16 @@ def fmintr(  # noqa: C901
             diag_h = g * dv * scale
 
             g_h = g * d
-            H_h = d * H * d[:, None]
+
+            # we don't need unscaled H anymore this iteration, so we overwrite
+            # it with H_h = d * H * d[:, None] to avoid carrying so many H-sized
+            # matrices in memory, which can be large
+            # doing operation H = d * H * d[:, None]
+            # with just in-place operations
```
YigitElma, line 430: same

YigitElma (changes requested):

YigitElma:
> ### Memory benchmark result
> ```diff
> |               Test Name                |      %Δ      |    Master (MB)     |      PR (MB)       |    Δ (MB)    |    Time PR (s)     |  Time Master (s)   |
> | -------------------------------------- | ------------ | ------------------ | ------------------ | ------------ | ------------------ | ------------------ |
>   test_objective_jac_w7x                 |   -0.15 %    |     3.945e+03      |     3.939e+03      |    -6.04     |       31.84        |       28.83        |
>   test_proximal_jac_w7x_with_eq_update   |   -0.34 %    |     6.902e+03      |     6.879e+03      |    -23.60    |       174.22       |       172.89       |
>   test_proximal_freeb_jac                |    0.02 %    |     1.317e+04      |     1.318e+04      |     3.01     |       71.00        |       70.69        |
>   test_proximal_freeb_jac_blocked        |    0.54 %    |     1.319e+04      |     1.327e+04      |    71.57     |       69.99        |       71.58        |
>   test_proximal_freeb_jac_batched        |   -2.12 %    |     7.644e+03      |     7.482e+03      |   -161.99    |       109.36       |       108.75       |
>   test_proximal_jac_ripple               |   -0.07 %    |     7.328e+03      |     7.323e+03      |    -5.43     |       69.62        |       69.91        |
>   test_proximal_jac_ripple_spline        |    2.08 %    |     3.831e+03      |     3.911e+03      |    79.84     |       77.24        |       77.45        |
> ```
> 
> For the memory plots, go to the summary of `Memory Benchmarks` workflow and download the artifact.

These benchmarks don't have any optimization, so this shouldn't have any effect. If you want, I can add an optimization test with only 1 step.

desc/optimize/aug_lagrangian.py
dpanici: I don't understand your comment?

dpanici:
> > ### Memory benchmark result
> > ```diff
> > |               Test Name                |      %Δ      |    Master (MB)     |      PR (MB)       |    Δ (MB)    |    Time PR (s)     |  Time Master (s)   |
> > | -------------------------------------- | ------------ | ------------------ | ------------------ | ------------ | ------------------ | ------------------ |
> >   test_objective_jac_w7x                 |   -0.15 %    |     3.945e+03      |     3.939e+03      |    -6.04     |       31.84        |       28.83        |
> >   test_proximal_jac_w7x_with_eq_update   |   -0.34 %    |     6.902e+03      |     6.879e+03      |    -23.60    |       174.22       |       172.89       |
> >   test_proximal_freeb_jac                |    0.02 %    |     1.317e+04      |     1.318e+04      |     3.01     |       71.00        |       70.69        |
> >   test_proximal_freeb_jac_blocked        |    0.54 %    |     1.319e+04      |     1.327e+04      |    71.57     |       69.99        |       71.58        |
> >   test_proximal_freeb_jac_batched        |   -2.12 %    |     7.644e+03      |     7.482e+03      |   -161.99    |       109.36       |       108.75       |
> >   test_proximal_jac_ripple               |   -0.07 %    |     7.328e+03      |     7.323e+03      |    -5.43     |       69.62        |       69.91        |
> >   test_proximal_jac_ripple_spline        |    2.08 %    |     3.831e+03      |     3.911e+03      |    79.84     |       77.24        |       77.45        |
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For the memory plots, go to the summary of `Memory Benchmarks` workflow and download the artifact.
> 
> These benchmarks don't have any optimization, so this shouldn't have any effect. If you want, I can add an optimization test with only 1 step.

Yea that might be useful. Just for lsq exact I'd say

desc/optimize/aug_lagrangian.py
YigitElma: You mention the operation is in-place, but then you say actually, it is not. It causes more confusion I guess

dpanici: I'll make this part of #1669 and reference that issue and remove these comments then

YigitElma (dismissed):

desc/optimize/fmin_scalar.py
```diff
@@ -443,13 +462,16 @@ def fmintr(  # noqa: C901
...
+    H_h /= d[:, None]
+    H_h /= d
     result = OptimizeResult(
         x=x,
         success=success,
         fun=f,
         grad=g,
         v=v,
-        hess=H,
+        hess=H_h,
```
ddudt, line 474: This is really the unscaled version that should be called `H`, not `H_h`, right? For the least squares optimizer we do this in a single line to avoid having a variable with a "misleading" name like: `jac=J_h * 1 / d`. I recommend doing something similar here, like the following suggestion: 
```
hess=H_h / d[:, None] / d
```
(And then remove the changes in lines 465-467)

ddudt (commented):
One minor comment, otherwise I would approve.

desc/optimize/fmin_scalar.py
YigitElma: Yes, this is the unscaled hessian

dpanici: sorry nvm I see what you meant

YigitElma:
I will add a small tip to decide what jac chunk size will make a difference. One can use the stdout at the beginning of the oprimization ` Number of parameters: ....` to decide min chunk size that will help reducing memory

YigitElma (dismissed):

YigitElma:
@ddudt @rahulgaur104

desc/optimize/aug_lagrangian_ls.py
```diff
@@ -281,7 +281,13 @@ def lagjac(z, y, mu, *args):
     diag_h = g * dv * scale
 
     g_h = g * d
-    J_h = J * d
+    # TODO: place this function under JIT (#1669)
+    # we don't need unscaled J anymore, so we overwrite
+    # it with J_h = J * d to avoid carrying so many J-sized matrices
+    # in memory, which can be large
+    J = J.at[:].set(J * d)
```
f0uriest, line 288: can't this be `J *= d`?

desc/optimize/aug_lagrangian_ls.py
```diff
@@ -531,7 +541,12 @@ def lagjac(z, y, mu, *args):
             d = v**0.5 * scale
             diag_h = g * dv * scale
             g_h = g * d
-            J_h = J * d
+            # we don't need unscaled J anymore, so we overwrite
+            # it with J_h = J * d to avoid carrying so many J-sized matrices
+            # in memory, which can be large
+            J = J.at[:].set(J * d)
```
f0uriest, line 547: `J *= d`

desc/optimize/least_squares.py
```diff
@@ -200,7 +200,15 @@ def lsqtr(  # noqa: C901
     diag_h = g * dv * scale
 
     g_h = g * d
-    J_h = J * d
+    # TODO: place this function under JIT (#1669)
+    # using in-place operations to save memory
+    J = J.at[:].set(J * d)
```
f0uriest, line 205: `J *= d`

desc/optimize/least_squares.py
```diff
@@ -388,7 +400,12 @@ def lsqtr(  # noqa: C901
             diag_h = g * dv * scale
 
             g_h = g * d
-            J_h = J * d
+            J = J.at[:].set(J * d)
```
f0uriest, line 403: `J *= d`

docs/performance_tips.rst
```diff
@@ -2,6 +2,25 @@
 Performance Tips
 ================
 
+Using more of the Available GPU Memory
+--------------------------------------
+By default, JAX will use only 75% of the available GPU memory. This is to allow other processes to run on the GPU at the same time. However, if you are running a single process and want to use all of the available memory (or some other amount of it), you can set the following environment variable:
```
f0uriest, line 7: i don't think this is correct? It's not that it only uses 75% of available, its that it reserves 75% when first imported, and will sub-allocate within that, but it will grab more if it's needed.

docs/performance_tips.rst
```diff
@@ -2,6 +2,25 @@
...
+
+.. code-block:: python
+
+    import os
+    os.environ["XLA_PYTHON_CLIENT_MEM_FRACTION"] = "0.90"
+
+Alternatively, if you don't want to preallocate memory, you can set the following environment variable to allow JAX to allocate memory as needed:
```
f0uriest, line 14: should maybe also mention that the jax docs don't recommend this as it will slow things down

docs/performance_tips.rst
```diff
@@ -2,6 +2,25 @@
 Performance Tips
 ================
 
+Using more of the Available GPU Memory
```
f0uriest, line 5: I'd also recommend moving this section lower down with the other memory stuff

docs/performance_tips.rst
```diff
@@ -28,10 +47,33 @@ of the cost function. This memory cost comes from attempting to calculate the en
...
-matrix of shape [``obj.dim_f`` x ``obj.dim_x``], and the calculation of the Jacobian is vectorized over
-the columns (the ``obj.dim_x`` dimension), where ``obj`` is the ``ObjectiveFunction`` object. Passing in the ``jac_chunk_size`` attribute allows one to split up
-the vectorized computation into chunks of ``jac_chunk_size`` columns at a time, allowing one to compute the Jacobian
-in a slightly slower, but more memory-efficient manner. The memory usage of the Jacobian calculation is
+matrix of shape [``obj.dim_f`` x ``x_opt.size``] (to see the size of ``x_opt``, one can check the printed ``Number of parameters:`` line, before optimization), and the calculation of the Jacobian is vectorized over
+the columns (the ``x_opt.size`` dimension), where ``obj`` is the ``ObjectiveFunction`` object. Passing in the ``jac_chunk_size`` attribute allows one to split up
+the vectorized computation into chunks of ``jac_chunk_size`` columns at a time, allowing one to compute the Jacobian in a slightly slower, but more memory-efficient manner.
+
+.. code-block:: python
```
f0uriest, line 54: maybe a line of text to introduce what this code block is meant to show?

f0uriest (changes requested):

docs/performance_tips.rst
YigitElma: No, I am pretty sure it just uses 75%, because I had many cases where increasing this prevented OOM. See https://github.com/PlasmaControl/DESC/issues/1686#issuecomment-2802801245 for example

docs/performance_tips.rst
YigitElma: I added the link to JAX docs, and in practice, I haven't seen a significant slowdown. Jax docs say it may cause segmentation fault but I think we don'thave that problem (I used this flag a lot and never that error)

docs/performance_tips.rst
f0uriest: I've definitely seen it use more than 75% without setting any extra flags. I think what you're seeing in those cases is that the internal XLA allocator is more efficient than the system one, so giving more memory to the XLA allocator lets you do slightly bigger problems. They note this in the jax docs, that turning off pre-allocation, or turning down the fraction may be less efficient and lead to higher overall memory usage.

docs/performance_tips.rst
ddudt: I have also found that it does not noticeably slow down performance, but I still think we should mention that this is only recommended for memory profiling.

ddudt (changes requested):

docs/performance_tips.rst
YigitElma: Hmm, okay. I think increasing this allocation gives a lot of memory saving, for example if I run `test_proximal_freeb_jac_batched` test on my laptop GPU with 12GB of memory with `platform` flag, I see peak ~4 GB memory  usage. If I set it to `0.25` fraction, it is slightly lower than what it needs but still there is 8GB more memory it can use, however I get OOM in this case. So, I think we should emphasize using higher fraction without saying it just uses that much.

dpanici (commented):
I approve but I wrote half so can't actually approve

YigitElma:
> I approve but I wrote half so can't actually approve

Me the same @f0uriest @ddudt

ddudt (approved):

rahulgaur104 (approved):

f0uriest (approved):

PR #1689, opened 2025-04-15
Add center=True option for CoilSetMinDistance and PlasmaCoilSetMinDistance

dpanici:
For small shaping coils, the center of the coil can be used as a proxy for the position along the coil, and one can use this in the distance objectives instead of the points along the coil. This can be much cheaper than using positions along the coil if the number of coils is large.

This PR adds `center` flag to `CoilSetMinDistance` and `PlasmaCoilSetMinDistance` which will use the center of the coil to compute distances from each coil.

Waiting on @dpanici to look into segmented sum to make this work so that we can have some coils use centers and others use the full coil


tests/test_coils.py
```diff
@@ -865,6 +865,40 @@ def test_symmetry_position(DummyCoilSet):
...
+    )
+    coilset_mixed = MixedCoilSet(*coilset_asym)
+    grid = LinearGrid(N=30)
+
+    # check that positions of CoilSets are the same with xyz basis
+    x_sym = coilset_sym._compute_position(basis="xyz", grid=grid, center=True)
+    x_asym = coilset_asym._compute_position(basis="xyz", grid=grid, center=True)
+    x_mixed = coilset_mixed._compute_position(basis="xyz", grid=grid, center=True)
+
+    np.testing.assert_allclose(x_sym, x_asym)
```
YigitElma, line 886: Should we make giving `atol` to these checks a default practice?

desc/coils.py
```diff
@@ -294,21 +296,37 @@ def _compute_position(self, params=None, grid=None, dx1=False, **kwargs):
...
+            errorif(dx1, ValueError, "``dx1=True`` cannot be used with ``center=True``")
         data = self.compute(keys, grid=grid, params=params, **kwargs)
-        x = jnp.transpose(jnp.atleast_3d(data["x"]), [2, 0, 1])  # shape=(1,num_nodes,3)
+        x = jnp.transpose(
+            jnp.atleast_3d(data[keys[0]]), [2, 0, 1]
+        )  # shape=(1,num_nodes,3)
+        if center:
+            # make sure only carrying around 1 center point per coil
+            x = x[:, 0, :]
+            x = x[:, None, :]
```
ddudt, line 329: Why is this needed? Shouldn't it work the same as the `center = False` case with `num_nodes = 1` (since the center is a single position)?

desc/coils.py
```diff
@@ -1497,24 +1517,38 @@ def _compute_position(self, params=None, grid=None, dx1=False, **kwargs):
...
         if params is None:
             params = [get_params(keys, coil, basis=basis) for coil in self]
         data = self.compute(keys, grid=grid, params=params, basis=basis, **kwargs)
         data = tree_leaves(data, is_leaf=lambda x: isinstance(x, dict))
-        x = jnp.dstack([d["x"].T for d in data]).T  # shape=(ncoils,num_nodes,3)
+        x = jnp.dstack([d[keys[0]].T for d in data]).T  # shape=(ncoils,num_nodes,3)
+        if center:
+            # make sure only carrying around 1 center point per coil
+            x = x[:, 0, :]
+            x = x[:, None, :]
```
ddudt, line 1551: see previous comment

tests/test_coils.py
ddudt: The default is `atol=0`, so I think it's fine to only give it if necessary.

desc/coils.py
dpanici: same answer as previous

YigitElma (dismissed):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Optionally can also contract the profiles of the original ``Equilibrium`` so tha
 - Allows Redl compute quantities to use SplineProfile
 - Updates Redl bootstrap current consistency tutorial to include a ``SplineProfile`` optimization
 - Adds automatically generated header file showing date the input file was created with `desc.vmec.VMECIO.write_vmec_input`
+- Adds center flag to ``CoilSetMinDistance`` and ``PlasmaCoilSetMinDistance`` which will use the center of the coil to compute distances from each coil, useful if the coils in the optimization are small enough to be well-approximated by their centers.
```
f0uriest, line 13: add ``ticks`` around ``center`` since its an argument name

desc/objectives/_coils.py
```diff
@@ -730,6 +730,11 @@ class CoilSetMinDistance(_Objective):
         a large number of coils, or if the resolution is very high, setting this to a
         small value will reduce peak memory usage at the cost of slightly increased
         runtime.
+    use_centers : bool
+        Whether or not to only consider the coil centers, rather than the position
+        along the coil itself. This can reduce the computational cost in situations
+        where a coil's position is well-approximated by its center, for example
+        if the coils have a very small radius.
```
f0uriest, line 737: may want to mention that in this case the min distance target should be larger than the radius for this objective to actually work correctly. May also want to mention dipole coils specifically since that's the common use case

desc/objectives/_coils.py
```diff
@@ -903,6 +910,11 @@ class PlasmaCoilSetMinDistance(_Objective):
...
```
f0uriest, line 917: see above comment, and should also mention that the coils shouldn't link the plasma (is it possible to also check that the initial coils dont link? not sure how expensive that is)

dpanici: Might be a bit expensive, since users have to opt into this flag I think I will trust them to know when they can't use it.

dpanici:
This would be most useful if we could allow some of the coils to be treated fully while others are treated with their centers... this may make it a bit messier though

dpanici:
segmented sum

dpanici:
> segmented sum JAX

see https://github.com/PlasmaControl/DESC/blob/b20e730f1dc932afd51cf193aa2017fa07fefa32/desc/objectives/_reconstruction.py#L247 for example use

unalmis:
> > segmented sum JAX
> 
> 
> 
> see https://github.com/PlasmaControl/DESC/blob/b20e730f1dc932afd51cf193aa2017fa07fefa32/desc/objectives/_reconstruction.py#L247 for example use

last year i remember all the jax segment ops were significantly less efficient (both memory and speed) than constructing the sparse matrix and multiplying

dpanici:
> > > segmented sum JAX
> > 
> > 
> > see https://github.com/PlasmaControl/DESC/blob/b20e730f1dc932afd51cf193aa2017fa07fefa32/desc/objectives/_reconstruction.py#L247
> >  for example use
> 
> last year i remember all the jax segment ops were significantly less efficient (both memory and speed) than constructing the sparse matrix and multiplying

At least for me in this case, the ops were easy to use and way better than a for loop. I will think about making a sparse matrix but if I can get it working with ops I am fine with that for now.

By sparse matrix you mean enforcing the segmented sum by just doing a big matrix multiply and masking out the entries you don't want?

dpanici:
Think about closing this once #2140 is in, as that one is a better approach

YigitElma:
@dpanici decide if this is needed, otherwise we can close it

dpanici:
Made irrelevant by #2140 so closing

PR #1690, opened 2025-04-15
`SurfaceCurrentRegularization` regularization options

ddudt:
Adds regularization options to the `SurfaceCurrentRegularization` objective. This is similar to the `regularization_type` argument of the function `solve_regularized_surface_current`.

### DIFF
--- a/desc/objectives/_coils.py
+++ b/desc/objectives/_coils.py
@@ -2291,6 +2291,8 @@ class SurfaceCurrentRegularization(_Objective):
     surface_current_field : CurrentPotentialField
         Surface current which is producing the magnetic field, the parameters
         of this will be optimized to minimize the objective.
+    regularization : str, optional
+        Regularization method. One of {'K', 'Phi', 'sqrt(Phi)'}. Default = 'K'.
     source_grid : Grid, optional
         Collocation grid containing the nodes to evaluate current source at on
         the winding surface. If used in conjunction with the QuadraticFlux objective,
@@ -2328,6 +2330,7 @@ class SurfaceCurrentRegularization(_Objective):
         normalize_target=True,
         loss_function=None,
         deriv_mode="auto",
+        regularization="K",
         source_grid=None,
         name="surface-current-regularization",
     ):
@@ -2336,6 +2339,12 @@ class SurfaceCurrentRegularization(_Objective):
             FourierCurrentPotentialField,
         )
 
+        errorif(
+            regularization not in ["K", "Phi", "sqrt(Phi)"],
+            ValueError,
+            "regularization must be one of ['K', 'Phi', 'sqrt(Phi)'], "
+            + f"got {regularization}.",
+        )
         if target is None and bounds is None:
             target = 0
         assert isinstance(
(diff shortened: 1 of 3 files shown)


ddudt:
@ddudt remember to update units and normalization

desc/objectives/_coils.py
```diff
@@ -2457,5 +2467,10 @@ def compute(self, surface_params=None, constants=None):
             profiles={},
         )
 
-        K_mag = safenorm(surface_data["K"], axis=-1)
-        return K_mag * jnp.sqrt(surface_data["|e_theta x e_zeta|"])
+        if self._regularization == "K":
+            K = safenorm(surface_data["K"], axis=-1)
+        elif self._regularization == "Phi":
+            K = jnp.abs(surface_data["Phi"])
```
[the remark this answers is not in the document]
dpanici: I guess it doesn't matter and this way the printout is more legible

dpanici (approved):

desc/objectives/_coils.py
```diff
@@ -2292,6 +2292,8 @@ class SurfaceCurrentRegularization(_Objective):
     surface_current_field : CurrentPotentialField
         Surface current which is producing the magnetic field, the parameters
         of this will be optimized to minimize the objective.
+    regularization : str, optional
+        Regularization method. One of {'K', 'Phi', 'sqrt(Phi)'}. Default = 'K'.
```
f0uriest, line 2296: should also modify the formula above on 2267

desc/objectives/_coils.py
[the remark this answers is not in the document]
ddudt: It's not necessary and would probably be more general without the absolute value. But since it doesn't really matter and the other two regularization options are always positive, I prefer to have it.

YigitElma (dismissed):

desc/objectives/_coils.py
```diff
@@ -2262,6 +2262,8 @@ def compute(self, params, constants=None):
 class SurfaceCurrentRegularization(_Objective):
     """Target the surface current magnitude.
 
+    If ``regularization="k"``:
```
dpanici, line 2265:
```suggestion
    If ``regularization="K"``:
```

desc/objectives/_coils.py
```diff
@@ -2292,6 +2306,8 @@ class SurfaceCurrentRegularization(_Objective):
     surface_current_field : CurrentPotentialField
         Surface current which is producing the magnetic field, the parameters
         of this will be optimized to minimize the objective.
+    regularization : str, optional
+        Regularization method. One of {'K', 'Phi', 'sqrt(Phi)'}. Default = 'K'.
```
dpanici, line 2310:
```suggestion
        Regularization method. One of {'"K"', '"Phi"', '"sqrt(Phi)"'}. Default = '"K"'.
```

dpanici (approved):

dpanici (changes requested):
clarify docstring slightly, I left suggested commit

dpanici (approved):

YigitElma (approved):

PR #1691, opened 2025-04-16
Fix `deriv_mode="batched"`

### LINKED ISSUE #1687: `ObjectiveFunction` `deriv_mode` not being set to `blocked` when `jac_chunk_size` is passed to a sub-objective
The expected behavior for this line

```python
objective=ObjectiveFunction(ForceBalance(eq, jac_chunk_size=1))
```
is that `objective._deriv_mode` is `"blocked"` and then the resulting jacobian calculations end up using the `jac_chunk_size` of the sub-objectives, in this case `1`

However currently the `objective._deriv_mode` is `"batched"` and it sets its own `jac_chunk_size` based off the auto heuristic, which can be larger than intended, so something is going wrong here in the logic.

YigitElma:
Resolves #1687


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Optionally can also contract the profiles of the original ``Equilibrium`` so tha
 - Updates Redl bootstrap current consistency tutorial to include a ``SplineProfile`` optimization
 - Adds automatically generated header file showing date the input file was created with `desc.vmec.VMECIO.write_vmec_input`
 
+Bug Fixes
+
+- Fixes bug when ``jac_chunk_size`` is given to a sub-objective but the ``ObjectiveFunction`` is still using a different value which is set by an heuristic based on available memory.
```
dpanici, line 16:
```suggestion
- Fixes bug where ``ObjectiveFunction`` was incorrectly using ``deriv_mode="batched"`` and the heuristic-set ``jac_chunk_size`` when ``jac_chunk_size`` is given to a sub-objective, where it should have instead defaulted to ``deriv_mode="blocked"``.
```

idk how to succinctly explain this bug...

dpanici (dismissed):

CHANGELOG.md
YigitElma: yeah I didn't want to explain the whole thing, maybe I should add the issue number

dpanici (approved):

f0uriest (approved):

PR #1697, opened 2025-04-19
Add 'linear_constraint_options'

ddudt:
Allows the user to pass `x_scale` to `factorize_linear_constraints` through a new `Optimizer.optimize` kwarg. 

The default `x_scale="auto"` normalizes the optimization variables by the magnitudes of the initial guess `x0`. This is helpful when you have variables of different units, for example in a single-stage optimization where the coil currents (MA) are 6 orders of magnitude larger than the equilibrium boundary coefficients (m). This default does not work well if the initial guess for those variables is 0, in which case no normalization is done. This PR allows the user to override that default, so for example coil currents can be initialized to 0 but still normalized by 1e6.


desc/optimize/_constraint_wrappers.py
```diff
@@ -43,11 +43,18 @@ class LinearConstraintProjection(ObjectiveFunction):
         Objective function to optimize.
     constraint : ObjectiveFunction
         Objective function of linear constraints to enforce.
+    x_scale : array_like or ``'auto'``, optional
```
YigitElma, line 46: I remember I was confused by these 2 different `x_scale`s during #1409. First of all, I have a general theory question: what is the difference between them? One scales the null-space and the other scales the reduced state vector that optimizer sees (and usually based on jacobian etc). But isn't the overall effect very similar?

Also, can I recommend changing the name of one of them to something like `x_scale_constraint` or whatever, if they have very different effects?

YigitElma: This has been discussed in dev meeting. Basically, they have similar effect, the one for linear constraint scales the variables permanently, the one for optimizer scales the Jacobian for better conditioning.

YigitElma (approved):

dpanici (changes requested):
I think `linear_constraint_options` should be included in the docstring for `optimize` now so it is clearly documented

dpanici (approved):

desc/optimize/optimizer.py
```diff
@@ -137,9 +137,19 @@ def optimize(  # noqa: C901
...
-            settings. See the documentation page ``Optimizers Supported`` for
-            more details (https://desc-docs.readthedocs.io/en/stable/optimizers.html)
-            to check the options for each specific available optimizer.
+            settings. Options that are universal to all optimizers are:
+
+            - ``"linear_constraint_options"`` : Dictionary of keyword arguments to pass
+            to ``LinearConstraintProjection``.
+            - ``"perturb_options"`` : Dictionary of keyword arguments to pass to
+            ``ProximalProjection`` as its ``perturb_options``.
+            - ``"solve_options"`` : Dictionary of keyword arguments to pass to
```
YigitElma, line 146: I think a careful user will get that `ProximalProjection` stuff are only used for optimization not equilibrium solve.

ddudt: Are you suggesting a change? If you are only running an equilibrium solve and not an optimization, then `"solve_options"` doesn't get used. So it won't harm anything. 

The docstrings are meant to be a short summary, we don't want to write a long essay explaining the details of every possible option for every possible situation.

YigitElma: Yeah sorry, I was going to approve it, but got distracted, no change is necessary, just wanted to point it out.

desc/optimize/optimizer.py
```diff
@@ -137,9 +137,19 @@ def optimize(  # noqa: C901
             Maximum number of iterations. Defaults to 100.
         options : dict, optional
             Dictionary of optional keyword arguments to override default solver
-            settings. See the documentation page ``Optimizers Supported`` for
-            more details (https://desc-docs.readthedocs.io/en/stable/optimizers.html)
-            to check the options for each specific available optimizer.
+            settings. Options that are universal to all optimizers are:
+
+            - ``"linear_constraint_options"`` : Dictionary of keyword arguments to pass
+            to ``LinearConstraintProjection``.
```
YigitElma, line 143:
```suggestion
              to ``LinearConstraintProjection``.
```

desc/optimize/optimizer.py
```diff
@@ -137,9 +137,19 @@ def optimize(  # noqa: C901
...
+            - ``"perturb_options"`` : Dictionary of keyword arguments to pass to
+            ``ProximalProjection`` as its ``perturb_options``.
```
YigitElma, line 145:
```suggestion
              ``ProximalProjection`` as its ``perturb_options``.
```

desc/optimize/optimizer.py
```diff
@@ -137,9 +137,19 @@ def optimize(  # noqa: C901
...
+            - ``"solve_options"`` : Dictionary of keyword arguments to pass to
+            ``ProximalProjection`` as its ``solve_options``.
```
YigitElma, line 147:
```suggestion
              ``ProximalProjection`` as its ``solve_options``.
```

YigitElma (changes requested):
These should resolve the docs build errors.

YigitElma (approved):

dpanici (approved):
Approving but in PR description can you be more detailed on use case? we document here that you can pass these things in, but idt we have LinearConstraintProjection in our public API so I'd like at least one place to point people other than the source to say what these options could be used for, if ever asked

PR #1702, opened 2025-04-25
Optional memory profiler workflow

YigitElma:
This adds an optional profiling workflow which will profile the CPU memory usage of some memory-intensive routines (such as Free Boundary, Bounce Averages, Jacobian) and comment the results in a way similar to timing benchmarks. Also, this will plot the data for each function, as well as master plots and saves it to artifacts.

There are a couple of concerns. First, how reliable is this, which I think is good enough to see significant changes. Second, I just run them around 3 times, I think this is enough since memory profile is less volatile then timing.


YigitElma:
> the memory benchmark took 6 hours to setup the tmate session... something weird going on there

My bad, I deleted a part of it since I don't want to use it, but there was one more step that I had to delete. It should work now...

tests/benchmarks/compare_mem_results.py
```diff
@@ -0,0 +1,72 @@
...
+run_id = os.environ.get("GITHUB_RUN_ID")
+repo = os.environ.get("GITHUB_REPOSITORY")
+
+# GitHub artifact URL (raw links are not stable unless served via Pages)
+image_url = f"https://github.com/{repo}/actions/runs/{run_id}/artifacts"
+
+# ---------- commit message ----------
+msg = "### Memory benchmark result\n\n```diff\n"
+msg += (
+    "| {'Test Name':^38} | {'%Δ':^12} | {'Master (MB)':^18} | "
```
YigitElma, line 47:
```suggestion
    f"| {'Test Name':^38} | {'%Δ':^12} | {'Master (MB)':^18} | "
```

tests/benchmarks/compare_mem_results.py
```diff
@@ -0,0 +1,72 @@
...
+    + "{'PR (MB)':^18} | {'Δ (MB)':^12} |\n"
```
YigitElma, line 48:
```suggestion
    + f"{'PR (MB)':^18} | {'Δ (MB)':^12} |\n"
```

tests/benchmarks/compare_mem_results.py
```diff
@@ -0,0 +1,72 @@
...
+    + f"{'PR (MB)':^18} | {'Δ (MB)':^12} |\n"
+)
+msg += f"| {'-'*38} | {'-'*12} | {'-'*18} | {'-'*18} | {'-'*12} |\n"
+for i, name in enumerate(data_master.keys()):
+    peak_pr = data_pr[name]["mem"].max()
+    peak_ma = data_master[name]["mem"].max()
+    delta = peak_pr - peak_ma
+    sign = "+" if delta >= 0 else "-"
+    # only show color if the delta is significant
+    color = " " if abs((delta / peak_ma) * 100) < 7 else "-" if delta >= 0 else "+"
```
YigitElma, line 57:
```suggestion
    color = " " if abs((delta / peak_ma) * 100) < 10 else "-" if delta >= 0 else "+"
```

YigitElma:
Example plot if you download the artifact,
compare

tests/benchmarks/memory_funcs.py
```diff
@@ -0,0 +1,172 @@
...
+    QuasisymmetryTwoTerm,
+    get_equilibrium_objective,
+    get_fixed_boundary_constraints,
+    maybe_add_self_consistency,
+)
+from desc.optimize import LinearConstraintProjection, ProximalProjection
+
+
+@pytest.mark.skip()
+@pytest.mark.unit
```
f0uriest, line 44: probably better to just add a new mark for memory tests rather than marking them as unit + skip

tests/benchmarks/memory_funcs.py
```diff
@@ -0,0 +1,172 @@
...
+        ObjectiveFunction(
+            maybe_add_self_consistency(eq, get_fixed_boundary_constraints(eq)),
+        ),
+    )
+    objective.build(verbose=0)
+    x = objective.x(eq)
+
+    for _ in range(3):
+        objective.jac_scaled_error(x, objective.constants).block_until_ready()
+        gc.collect()
```
f0uriest, line 60: do we want this here? wouldn't this possibly reduce memory usage below what we'd see in practice?

tests/benchmarks/compare_mem_results.py
```diff
@@ -0,0 +1,65 @@
...
+    for i, name in enumerate(data_master.keys()):
+        peak_pr = data_pr[name]["mem"].max()
+        peak_ma = data_master[name]["mem"].max()
+        delta = peak_pr - peak_ma
+        # only show color if the delta is significant
+        color = " " if abs((delta / peak_ma) * 100) < 10 else "-" if delta >= 0 else "+"
+        percent_change = f"{(delta / peak_ma) * 100:.2f}" + " %"
+        delta = f"{delta:.2f}"
+        msg += (
+            f"{color} {name:<38} | {percent_change:^12} | {peak_ma:^18.1f} | "
```
f0uriest, line 54: minor point but I prefer exponential notation `{blah:.3e}` to floating point

tests/benchmarks/memory_funcs.py
YigitElma: Moved it to top with `jax.clear_caches()`

.github/workflows/jax_tests.yml
```diff
@@ -2,12 +2,12 @@ name: Dependency test JAX
...
 on:
   pull_request:
-    types: [labeled]
+    types: [labeled, synchronize]
   workflow_dispatch:
 
 jobs:
   jax_tests:
-    if: ${{ github.event.label.name == 'test_jax' && github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'test_jax') || github.event_name == 'workflow_dispatch' }}
```
YigitElma, line 10: We had to manually remove/add the label everytime we push a commit, this fixes that

dpanici:
Make this required (maybe ake benchmarks not auto)

ddudt (dismissed):
I agree with Rory's comment about scientific notion, but other wise looks good

dpanici (dismissed):

dpanici (approved):

rahulgaur104 (approved):

PR #1704, opened 2025-04-28
add small footnote to poincare_plot doc

### LINKED ISSUE #1695: field line tracing opposite to the magnetic field direction.
I think it would be a simple but useful feature to add the ability to perform/plot poincare trace in a direction opposite to the magnetic field. This can help users better understand the fieldline structure outside the stellarator boundary.

With modular coils, can be easily achieved by reversing the coil current in all the coils. This would reverse the magnetic field without changing the flux surface shape.

dpanici:
Resolves #1695 

- [ ] change docstring to not say strictly increasing but rather monotonic
- [ ] remove current change
- [ ] double check if Bphi neg means phi is reversed phi

np.linspace(0, -2pi)

### DIFF
--- a/desc/plotting.py
+++ b/desc/plotting.py
@@ -1895,6 +1895,10 @@ def poincare_plot(
         fig, ax = desc.plotting.poincare_plot(
             field, r0, z0, NFP=eq.NFP, ax=ax, color="r", size=0.5, ntransit=250
         )
+        # if one wants to follow the field lines in reverse, can simply use
+        field_reversed = ScaledMagneticField(-1.0, field)
+        # and use this field_reversed instead of field.
+
     """
     fli_kwargs = {}
     for key in inspect.signature(field_line_integrate).parameters:


YigitElma (approved):

rahulgaur104 (approved):

f0uriest:
I don't really get what case we're trying to cover here?

If B_phi < 0 then the field lines are traced backwards (ie, phi decreases as you go along the fieldline), so if you say you want output at phi=0 and pi/2, what happens? should it give you that or at phi=-pi/2?

rahulgaur104:
>If B_phi < 0 then the field lines are traced backwards (ie, phi decreases as you go along the fieldline), so if you say you want output at phi=0 and pi/2, what happens? should it give you that or at phi=-pi/2?

B_phi < 0 doesn't mean that fieldlines are traced antiparallel to B. If you just flip B_phi, you will stop following the fieldline.
But I guess you are asking what happens to NFP > 1 cases when the coil current is reversed which is a valid point.
Yes, we will probably end up at phi = -pi/2 instead of pi/2.

dpanici:
Superseded (potentially) by #1663

YigitElma:
I checked this with `np.linspace(0, -2*np.pi, 10)`. It doesn't work except you also specify negative `min_step_size` to `field_line_integrate`. Instead of having 2 checks for reverse field detection, I think just reversing the phis when plotting is enough. See the test I have in #1663

PR #1705, opened 2025-04-28
Add uv project install instructions to docs

### LINKED ISSUE #1680: Add `uv` block to installation docs

cfe316:
Resolves #1680 .

This method is under the On Your Local Machine section but it also works on clusters.

I'm not sure how to preview any documentation. Is there a way?


dpanici:
running the command `make html` in the `docs` folder should make the docs locally and you can view them in the `docs/_build` folder (I think)

but you can a so just wait for the docs action to complete here and check it there

docs/installation.rst
```diff
@@ -51,6 +51,39 @@ Now use pip to install packages (this will only install DESC + JAX with CPU capa
...
+
+    # the depth=1 option reduces the quantity of older data downloaded
+    git clone --depth=1 git@github.com:PlasmaControl/DESC.git
+
+    uv init myproject
+    cd myproject
+
+    # add dependencies
+    uv add --editable "../DESC"
+    uv add "jax[cpu]" # or jax[cuda12] or similar on a platform with a GPU
```
YigitElma, line 75: I have never used `uv`, but in some of the tutorials I saw, it was still required to install dependencies with some file etc, is this enough for DESC? We have a bunch of other dependencies too which this doesn't explicitly show.

YigitElma: A quick question: does this create  a virtual environment? For example, how do you select this on cluster interactive jupyter sessions?

docs/installation.rst
```diff
@@ -51,6 +51,43 @@ Now use pip to install packages (this will only install DESC + JAX with CPU capa
     # optionally install developer requirements (if you want to run tests)
     pip install -r devtools/dev-requirements.txt
 
+**Or using uv instead of pip**
+
+One could use `uv <https://docs.astral.sh/uv>`_, a new python package management tool, instead of pip.
+For a project that modifies DESC and also uses it to perform analysis,
+it can be nice to separate the DESC folder from the project's data, scripts, jupyter notebooks, etc.
+This will show how to set up a new ``uv`` project with DESC as an editable dependency,
```
dpanici, line 59:
```suggestion
This will show how to set up a new ``uv`` project with DESC as an editable dependency (Either on local machine or on the cluster, this method can work with both),
```

dpanici (dismissed):

docs/installation.rst
```diff
@@ -51,6 +51,43 @@ Now use pip to install packages (this will only install DESC + JAX with CPU capa
     # optionally install developer requirements (if you want to run tests)
     pip install -r devtools/dev-requirements.txt
 
+**Or using uv instead of pip**
+
+One could use `uv <https://docs.astral.sh/uv>`_, a new python package management tool, instead of pip.
+For a project that modifies DESC and also uses it to perform analysis,
+it can be nice to separate the DESC folder from the project's data, scripts, jupyter notebooks, etc.
+This will show how to set up a new ``uv`` project with DESC as an editable dependency (Either on local machine or on the cluster, this method can work with both),
```
YigitElma, line 59:
```suggestion
This will show how to set up a new ``uv`` project called ``myproject`` with DESC as an editable dependency (Either on local machine or on the cluster, this method can work with both),
```

docs/installation.rst
```diff
@@ -51,6 +51,43 @@ Now use pip to install packages (this will only install DESC + JAX with CPU capa
...
+    # initialize a project
+    uv init myproject
+    cd myproject
+
+    # add dependencies
+    uv add --editable "../DESC"
+    # to install a special jax version, for GPU:
+    # uv add "jax[cuda12]"
+
+    # run a python REPL
```
YigitElma, line 79: I don't think we need this here. Or we can add this as, "test you installation", then say run `from desc.backend import print_backend_info; print_backend_info()`.

Personally, I would prefer the latter.

docs/installation.rst
```diff
@@ -51,6 +51,43 @@ Now use pip to install packages (this will only install DESC + JAX with CPU capa
...
+    uv run python
+
+    # Jupyter Notebooks
+    # ----------------
+    # install a jupyter kernel
+    uv add --dev ipykernel
+    uv run ipython kernel install --user --env VIRTUAL_ENV $(pwd)/.venv --name=myproject
```
YigitElma, line 86: Maybe we can note that for VSCode jupyter extension, these are not necessary, after `uv add --dev ipykernel`, just selecting `.venv` is enough.

YigitElma: Forgive my lack of knowledge, but what does the second line do?

YigitElma (changes requested):

YigitElma:
@cfe316 I have some minor comments, but since `uv` is relatively new, I think it is worth having some additional explanations.

docs/installation.rst
cfe316: The second line installs the kernel into a place where jupyter can see it. For example on my linux installation the typical directory is `.local/share/jupyter/kernels`; I guess this is the `--user` directory.

 All different jupyter kernels have entries in this folder, for example julia, python, and other languages. Each kernel installed there has a subfolder with a `kernel.json` file (which lists the executable, any arguments (or virtual environments), and the display name of the kernel) and a few `logo.png` files.

YigitElma (approved):

f0uriest (approved):

PR #1712, opened 2025-05-07
`BoundaryError` flag for `finite_pressure`

### LINKED ISSUE #1711: `BoundaryError` Jacobian can be `NaN` with some kinetic profiles
The `BoundaryError` objective requires computing the pressure `"p"` at the boundary for the pressure balance equation. Only the pressure is used, not the pressure gradient. But if the pressure gradient is undefined at the boundary, then this results in `NaN` values for the Jacobian of the pressure balance equations. 

A common use case is when creating kinetic profiles from an existing pressure profile:
```
# equilibrium with pressure profile
eq = get("DSHAPE")
print(f"p(1) = {eq.compute('p', grid=LinearGrid(rho=1.0))['p'][0]}")
print(f"p'(1) = {eq.compute('p_r', grid=LinearGrid(rho=1.0))['p_r'][0]}")

# change to equivalent kinetic profiles
density = PowerProfile(0.5, ScaledProfile(0.5 / elementary_charge, eq.pressure))
temperature = PowerProfile(0.5, ScaledProfile(0.5 / elementary_charge, eq.pressure))
eq.pressure = None
eq.atomic_number = PowerSeriesProfile([1], [0])
eq.electron_density = density
eq.electron_temperature = temperature
eq.ion_temperature = temperature
print(f"p(1) = {eq.compute('p', grid=LinearGrid(rho=1.0))['p'][0]}")
print(f"p'(1) = {eq.compute('p_r', grid=LinearGrid(rho=1.0))['p_r'][0]}")

# BoundaryError objective
field = ToroidalMagneticField(B0=0.2, R0=3.5)
objective = ObjectiveFunction(BoundaryError(eq=eq, field=field, field_fixed=True))
objective.build()

f = objective.compute_unscaled(objective.x(eq))
J = objective.jac_unscaled(objective.x(eq))
print(np.any(np.isnan(f)))
print(np.any(np.isnan(J)))
```

ddudt:
Resolves #1711 

Adds a flag `finite_pressure` to the `BoundaryError` objective to include the plasma pressure term in the computation or not.

### DIFF
--- a/desc/objectives/_free_boundary.py
+++ b/desc/objectives/_free_boundary.py
@@ -359,7 +359,7 @@ class BoundaryError(_Objective):
     Computes the residual of the following:
 
     𝐁ₒᵤₜ ⋅ 𝐧 = 0
-    𝐁ₒᵤₜ² - 𝐁ᵢₙ² - p = 0
+    𝐁ₒᵤₜ² - 𝐁ᵢₙ² - 2μ₀p = 0
     μ₀∇Φ − 𝐧 × [𝐁ₒᵤₜ − 𝐁ᵢₙ]
 
     Where 𝐁ᵢₙ is the total field inside the LCFS (from fixed boundary calculation)
@@ -405,6 +405,10 @@ class BoundaryError(_Objective):
     field_fixed : bool
         Whether to assume the field is fixed. For free boundary solve, should
         be fixed. For single stage optimization, should be False (default).
+    finite_pressure : bool
+        Whether to assume the plasma pressure at the boundary is finite (True) or always
+        vanishes (Fasle). If False, the plasma pressure term is removed from the
+        pressure balance equation. Default is True.
     bs_chunk_size : int or None
         Size to split Biot-Savart computation into chunks of evaluation points.
         If no chunking should be done or the chunk size is the full input
@@ -464,6 +468,7 @@ class BoundaryError(_Objective):
         name="Boundary error",
         jac_chunk_size=None,
         *,
+        finite_pressure=True,
         bs_chunk_size=None,
         B_plasma_chunk_size=None,
         **kwargs,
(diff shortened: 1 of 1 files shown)


f0uriest (changes requested):
Not crazy about this since it seems like it's not fixing the underlying issue, its just avoiding the issue in one particular place. I'd prefer a more complete fix.

ddudt:
> Not crazy about this since it seems like it's not fixing the underlying issue, its just avoiding the issue in one particular place. I'd prefer a more complete fix.

I think the "underlying issue" is just the nature of math. We have a pressure profile like $p^\alpha$, so the derivative is $\partial_\alpha (p^\alpha) = p^\alpha \log(p)$. And this will always be undefined when $p=0$, which is the most common use case. 

So I don't see a clean solution while still allowing for these profile parameterizations. This was the simplest (but admittedly hacky) solution I could think of. I'm open to other ideas if you have one.

ddudt:
@f0uriest I updated this to make the solution more elegant, but the approach is the same: ignore the pressure term in the equation when it is unnecessary. 

As we discussed, the pressure derivatives are _supposed_ to be undefined sometimes and that mathematical fact is not an "issue" we can fix. I think this solution is simple and covers all of the common use cases.

desc/objectives/_free_boundary.py
```diff
@@ -756,7 +756,11 @@ def compute(self, eq_params, *field_params, constants=None):
 
         g = eval_data["|e_theta x e_zeta|"]
         Bn_err = Bn * g
-        Bsq_err = (bsq_in + eval_data["p"] * (2 * mu_0) - bsq_out) * g
+        Bsq_err = jnp.where(
```
f0uriest, line 759: Does this also work for reverse mode? Usually in that case you need a double where

ddudt: It does not work for reverse mode. I don't think it can be fixed with a double where because computing the pressure itself is the thing that is undefined, not some function evaluated at a pressure of 0. So we can't check "where p=0" without causing NaNs. 

My old method with the input argument flag would work with reverse mode, but I still like this new method better because it avoids any work for the user. It is unlikely that anyone would need/want to use reverse mode with this objective anyways. I could add a check to make sure `deriv_mode="fwd"`, at least if the equilibrium has kinetic profiles?

YigitElma (approved):

f0uriest (approved):

PR #1713, opened 2025-05-07
Add flag to not check intersection when plotting coils

dpanici:


YigitElma:
This just adds the existing option of `flatten_coils` to the plotting function, right?

ddudt (approved):

desc/plotting.py
```diff
@@ -2574,13 +2577,16 @@ def plot_coils(coils, grid=None, fig=None, return_data=False, **kwargs):
     if grid is None:
         grid = LinearGrid(N=400, endpoint=True)
 
-    def flatten_coils(coilset):
+    def flatten_coils(coilset, check_intersection=check_intersection):
```
f0uriest, line 2580: I think you don't need this here you can just close over the value but it's fine

f0uriest (approved):

YigitElma (approved):

PR #1714, opened 2025-05-10
Do not check intersection by default when adding coilsets

dpanici:


YigitElma:
`TestCoilSet.test_dunder_methods` is failing due to not throwing warning.

YigitElma (dismissed):

dpanici:
Should just use `MixedCoilSet(*c1, *c2, check_intersection=False)` to avoid this check

PR #1722, opened 2025-05-13
Add PlasmaCoilSetMaxDistance objective

ryanwu4:
Sometimes, the coils get too far away from the plasma, and I often need an objective function which constrains the distance between plasma and coil from above. This adds that objective by generalizing the PlasmaCoilSetDistance objective to find the maximum (over the coil) minimum (over the plasma) distance between coil and plasma. To do this, I also added functionality to the `softmax` function to allow it to perform maximization over a given axis rather than flattening the array.

### DIFF
--- a/desc/objectives/_coils.py
+++ b/desc/objectives/_coils.py
@@ -15,7 +15,7 @@ from desc.utils import Timer, broadcast_tree, errorif, safenorm, setdefault, war
 
 from .normalization import compute_scaling_factors
 from .objective_funs import _Objective, collect_docs
-from .utils import softmin
+from .utils import softmax, softmin
 
 
 class _CoilObjective(_Objective):
@@ -850,11 +850,13 @@ class CoilSetMinDistance(_Objective):
         return min_dist_per_coil
 
 
-class PlasmaCoilSetMinDistance(_Objective):
-    """Target the minimum distance between the plasma and coilset.
+class PlasmaCoilSetDistance(_Objective):
+    """Target the distance between the plasma and coilset.
 
-    Will yield one value per coil in the coilset, which is the minimum distance from
-    that coil to the plasma boundary surface.
+    Will yield one value per coil in the coilset, which is the distance from
+    that coil to the plasma boundary surface. Can be penalized as maximum
+    or minimum distance. Note that maximum plasma coil distance is the maximum over the
+    coil, not over the plasma.
 
     NOTE: By default, assumes the plasma boundary is not fixed and its coordinates are
     computed at every iteration, for example if the equilibrium is changing in a
@@ -869,6 +871,8 @@ class PlasmaCoilSetMinDistance(_Objective):
         to satisfy the Objective.
     coil : CoilSet
         Coil(s) that are to be optimized.
+    mode: string
+        Either `max` or `min` for targeting max or min plasma-coil distance.
     plasma_grid : Grid, optional
         Collocation grid containing the nodes to evaluate plasma geometry at.
         Defaults to ``LinearGrid(M=eq.M_grid, N=eq.N_grid)``.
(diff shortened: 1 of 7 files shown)


tests/test_objective_funs.py
```diff
@@ -1249,6 +1250,148 @@ def test(
             coils_fixed=True,
         )
 
+    @pytest.mark.unit
+    def test_plasma_coil_max_distance(self):
```
dpanici, line 1254: in these cases, is the max also equal to the min?

Might be worth to try shifting one of these slightly outward so that there is an asymmetry introduced in the problem, so that you can check the max stuff more explicitly

ryanwu4:
Not sure why codecov is freaking out over these lines which are actually covered in the tests ...

dpanici (dismissed):

ryanwu4:
Change this to remove Max option, keep Bounds and precompute both max and min so that the bounds objective works the "intuitive" way

desc/objectives/_coils.py
```diff
@@ -1095,12 +1110,135 @@ def body(k):
...
+                return max
+            if self._ismin:
+                return min
+            return jnp.array([min, max])
 
         k = jnp.arange(self.dim_f)
-        min_dist_per_coil = vmap_chunked(body, chunk_size=self._dist_chunk_size)(k)
-        return min_dist_per_coil
+        extreme_dist_per_coil = vmap_chunked(body, chunk_size=self._dist_chunk_size)(k)
+        return extreme_dist_per_coil
```
YigitElma, line 1131: For ``bound`` case, is this going to return a 2 by N array? I don't know how that would work out for optimization. Can you concatenate this into a single 1D array of size 2*N? Also, you should change the `dim_f` part to account for the special ``bound`` case, and most probably the target to be an array like [min, max, min, max, ...]

YigitElma (changes requested):

desc/objectives/_coils.py
YigitElma: The ``target`` can also be in [min, min, min,...., max, max, max ...] order. Whichever you choose

YigitElma (approved):

desc/objectives/_coils.py
```diff
@@ -931,16 +940,23 @@ def __init__(
...
         if target is None and bounds is None:
             bounds = (1, np.inf)
+        errorif(
+            mode not in ["bound", "max", "min"],
+            ValueError,
+            "mode must be 'bound', 'max', or 'min'",
+        )
         self._eq = eq
         self._coil = coil
+        self._ismax = mode == "max"
```
f0uriest, line 958: Why not just store mode itself instead of two bools?

desc/objectives/_coils.py
```diff
@@ -931,16 +940,23 @@ def __init__(
...
         eq_fixed=False,
         coils_fixed=False,
-        name="plasma-coil minimum distance",
+        name="plasma-coil distance",
         jac_chunk_size=None,
         use_softmin=False,
         softmin_alpha=1.0,
         dist_chunk_size=None,
     ):
         if target is None and bounds is None:
```
f0uriest, line 949: Do we want the default for this to be different from PlasmaCoilsetMinDistance?

desc/objectives/_coils.py
```diff
@@ -850,11 +850,15 @@ def body(k):
...
 
 
-class PlasmaCoilSetMinDistance(_Objective):
-    """Target the minimum distance between the plasma and coilset.
+class PlasmaCoilSetDistanceBound(_Objective):
+    """Target the distance between the plasma and coilset.
 
-    Will yield one value per coil in the coilset, which is the minimum distance from
-    that coil to the plasma boundary surface.
+    Will yield one or two values per coil in the coilset, depending on the mode
```
f0uriest, line 856: Should mention somewhere here that the minima is the first N values in the array, then the maxima, it's not interleaved

YigitElma:
Can we also consider solving #1733 here?

dpanici (approved):

f0uriest (approved):

PR #1723, opened 2025-05-13
Ensure `sgd` prints an iteration even if the first step satisfies the termination criterion

dpanici:
Move check_termination later so that sgd actually prints the iteration before terminating

### DIFF
--- a/desc/optimize/stochastic.py
+++ b/desc/optimize/stochastic.py
@@ -125,7 +125,19 @@ def sgd(
 
     allx = [x]
 
-    while True:
+    while iteration < maxiter and success is None:
+
+        v = beta * v + (1 - beta) * g
+        x = x - alpha * v
+        g = grad(x, *args)
+        ngev += 1
+        step_norm = jnp.linalg.norm(alpha * v, ord=2)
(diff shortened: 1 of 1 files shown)


desc/optimize/stochastic.py
```diff
@@ -163,7 +162,8 @@ def sgd(
             success, message = False, STATUS_MESSAGES["callback"]
 
         iteration += 1
-
+    if (iteration == maxiter) and success is None:
```
dpanici, line 165: just print first iteration

desc/optimize/stochastic.py
```diff
@@ -125,7 +125,19 @@ def sgd(
...
+        x = x - alpha * v
+        g = grad(x, *args)
+        ngev += 1
+        step_norm = jnp.linalg.norm(alpha * v, ord=2)
+        g_norm = jnp.linalg.norm(g, ord=jnp.inf)
+        fnew = fun(x, *args)
+        nfev += 1
+        df = f - fnew
+        f = fnew
+
```
YigitElma, line 140: You might need to add the below `check_termination` before the loop to prevent taking a step in case if it is already satisfied

rahulgaur104 (approved):

PR #1724, opened 2025-05-14
ensure surfaces only compute things that make sense for their parametrizations

### LINKED ISSUE #1127: Surfaces can compute more than they should

dpanici:
Resolves #1127


dpanici:
@unalmis are there any specific tests for this I should try? the test mentioned in #1127 seems to pass on master already

dpanici:
All tests pass and I think I removed the stuff that was problematic for #1094 , so I'd say we merge this to master then work on fixing #1094

desc/compute/_core.py
```diff
@@ -2270,7 +2270,7 @@ def _omega_rrrz(params, transforms, profiles, data, **kwargs):
     data=["0"],
     parameterization=[
         "desc.equilibrium.equilibrium.Equilibrium",
-        "desc.geometry.core.Surface",
+        "desc.geometry.surface.ZernikeRZToroidalSection",
```
unalmis, line 2273: Can we compute this zeta derivative on a cross-section?

desc/compute/_geometry.py
```diff
@@ -466,6 +466,46 @@ def _perimeter_of_z(params, transforms, profiles, data, **kwargs):
...
+        "desc.geometry.surface.ZernikeRZToroidalSection",
+    ],
+    resolution_requirement="rt",  # just need max(rho) near 1
+)
+def _perimeter_of_z_ZernikeRZToroidalSection(
+    params, transforms, profiles, data, **kwargs
+):
+    # NOTE: this is perimeter of a constant zeta XS
+    # as a ZernikeRZToroidalSection is defined at constant zeta, so
+    # doing a perimeter at constant phi would make no sense.
```
unalmis, line 492: Originally, these quantities were designed to be computed at constant phi for the elongation. So although the description says constant zeta surface, the current user expectation has been that the computation be done at constant phi surface.

Frankly, these should either be renamed to `perimter(phi)` (no use case where you compute constant phi stuff on grids that are uniform in zeta not phi), or we should decide that we only care about constant zeta sections rather than constant phi cross sections.

desc/compute/_geometry.py
```diff
@@ -466,6 +466,46 @@ def _perimeter_of_z(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="perimeter(z)",
+    label="P(\\zeta)",
+    units="m",
+    units_long="meters",
+    description="Perimeter of enclosed cross-section (enclosed constant zeta surface), "
```
unalmis, line 474: This computes something different than `eq.compute("perimeter(z)")` and `FourierRZToroidalSurface.compute("perimeter(z)")`. Therefore, it should have a different name.

desc/compute/_core.py
```diff
@@ -2118,6 +2118,7 @@ def _lambda_zzz(params, transforms, profiles, data, **kwargs):
 )
 def _omega(params, transforms, profiles, data, **kwargs):
     data["omega"] = data["0"]
+    # TODO: what does omega !=0 mean for ZernikeRZToroidalXS?
```
unalmis, line 2121: It means a ZernikeRZToroidalXS cross-section for a (rho, theta, zeta) grid is no longer a constant phi cut in the lab frame. A ZernikeRZToroidalXS cross-section for a (rho, theta, phi) grid remains a constant phi cut in the lab frame.

unalmis (changes requested):
Everything is great except I do not like how two different quantities are named the same thing for `perimeter(z)` maybe make the old one `perimter(phi)` if we want it at constant phi.

unalmis (dismissed):

f0uriest (approved):

PR #1725, opened 2025-05-14
Change default constraints for eq.optimize to just be ForceBalance

dpanici:
currently they are fixing every DOF and then adding `ForceBalance`, which can be confusing.

Also adds a warning to users so they know this is not a recommended way to approach the equilibrium optimization problem

### DIFF
--- a/desc/equilibrium/equilibrium.py
+++ b/desc/equilibrium/equilibrium.py
@@ -2290,7 +2290,8 @@ class Equilibrium(IOAble, Optimizable):
         objective : ObjectiveFunction
             Objective function to optimize.
         constraints : Objective or tuple of Objective
-            Objective function to satisfy. Default = fixed-boundary force balance.
+            Objective function to satisfy. By default, force-balance is placed
+            as a constraint, but every other DOF (boundary, profiles, Psi) is free.
         optimizer : str or Optimizer (optional)
             optimizer to use
         ftol, xtol, gtol, ctol : float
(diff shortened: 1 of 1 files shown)


dpanici:
- no default
- no ForceBalance or equilibrium (check is eq objective flag) objective in constraints in eq.optimize should yield a warning

tests/test_equilibrium.py
```diff
@@ -458,3 +458,12 @@ def test_assigning_profile_iota_current():
...
     assert eq.current is None
+
+
+@pytest.mark.unit
+def test_eq_optimize_default_constraints_warning(DummyStellarator):
+    """Tests default constraints warning for eq.optimize."""
+    eq = load(load_from=str(DummyStellarator["output_path"]), file_format="hdf5")
+    eq.change_resolution(M=1, N=0, M_grid=2, N_grid=0)
+    with pytest.warns(UserWarning, match="DOF is free"):
+        eq.optimize(ObjectiveFunction(ForceBalance(eq)), maxiter=0)
```
YigitElma, line 469: This will probably fail since you changed the message, and also, it will throw an error due to missing constraint, so you will get `eq.optimize expects 2 positional arguments, 1 given ...`

YigitElma (changes requested):
Test will probably fail. Otherwise, looks fine.

YigitElma (dismissed):

f0uriest (dismissed):

YigitElma:
You need to add one more warning suppression to `test_scipy_constrained_solve `

YigitElma (approved):

unalmis (approved):

PR #1730, opened 2025-05-16
Add new initialization method based on `map2disc` package to ensure nested surfaces

### LINKED ISSUE #1729: map2disc initialisation
This is broken for the moment, but I added the "babin_init" branch anyway, two problems remain:

1. This uses map2disc which relies on packages that use /p instead of //p in comment strings, which result in syntaxerror's instead of syntaxwarnings (tested in python 3.13 and 3.12). This seems unnecessary because the breaking part is in comment strings.

2. It adds some functions to `initial_guess.py` which are all used by `def babin_initial_guess(eq):` resulting in an axis initialized with map2disc for N=2 modes `(n_babin_init = 2)`. I am not sure what is the best way to re-initialise the axis information for an equilibrium because I had some problem setting the axis via `FourierRZCurve` and using `eq.set_new_initial_guess`.

https://github.com/PlasmaControl/DESC/tree/babin_init

dpanici:
- [ ] If we merge this, do we want this to just be the default? Some of the heuristics can succeed but yield very poor initial surfaces (high grad(rho) on one side, etc), so the map2disc might be better even in cases when the heuristics do technically work.
- [x] add tests
- [x] simplify the added script (map2disc already gives back the mapping in zernike, we don't need to go through the VMEC form of R,Z)
- [x] Figure out solution to the syntax errors from the `"/p"` strings (add an ignore for this specific error at the top of our `initial_guesses.py` script? (I don't see these as of 3-5-26 using Python 3.13 environment and installing with `pip install -r devtools/dev-requirements.txt` then `pip install --editable .` )
- [x]  Install currently fails because some dependency packages seem to not work with newer (3.12+) python versions (likely works now perhaps bc the numpy is allowed to be >2 now for map2disc)
Resolves #1729 

Update 5-3-26:
- Not sure what changed, but able to install now on my Mac with Python 3.13. Still wary of a couple of the [package dependencies ](https://gitlab.mpcdf.mpg.de/gvec-group/map2disc/-/blob/main/pyproject.toml?ref_type=heads)which are severely unmaintained (https://github.com/dbstein/pybie2d/issues/2 for example)
- Install issues notwithstanding, is super fast and works well.

Currently just evaluating R,Z in real space then fitting with 3D FourierZernike. Seems to work fine, but better way would be to not go through real space and just fit the set of 2D Zernike polynomial coefficients obtained from map2disc directly, but this requires knowing the mode orderings which I did not wish to dig into at the moment, as the real space approach works fine. Placing the other method here for now
```python
    # TODO: smarter way would be to take the existing Zernike coeffs as fxn of phi and
    # fit those coefficients to immediately get the 3D FourierZernike coefficients.
    # Need to know what the mode orderings are though for cx and cy to use this
    # approach, which I've not dug into yet as the other method works fine.
   grid = LinearGrid(zeta=zeta_cut, NFP=eq.NFP)
    Rtransform = Transform(
        grid=grid,
        basis=FourierSeries(
            N, sym={True: "cos", False: False}[surface.sym], NFP=eq.NFP
        ),
    )
    Rtransform.build_pinv()
    R_lmn2 = vmap(Rtransform.fit)(all_cx)

    Ztransform = Transform(
        grid=grid,
        basis=FourierSeries(
            N, sym={True: "sin", False: False}[surface.sym], NFP=eq.NFP
        ),
    )
    Ztransform.build_pinv()
    Z_lmn2 = vmap(Ztransform.fit)(all_cy)
```


Mandelbr0t (dismissed):
Looks good from my side and tests pass. Im not sure what changed regarding DESC internals but dpanici knows more about that than I do

desc/equilibrium/initial_guess.py
```diff
@@ -228,21 +232,54 @@ def set_initial_guess(eq, *args, ensure_nested=True):  # noqa: C901
...
-        eq.solve(
-            objective=obj,
-            constraints=constraints,
-            ftol=0,
-            xtol=0,
-            gtol=1e-8,
-            verbose=0,
-            optimizer="fmintr-bfgs",
-        )
+        try:
```
dpanici: put import for map2disc here as well

desc/equilibrium/initial_guess.py
```diff
@@ -4,18 +4,22 @@
 import warnings
 
 import numpy as np
+from map2disc import BCM
```
YigitElma: Need to add try/except

daniel-dudt:
Look at how involved it would be to rewrite `map2disc` as part of DESC with JAX?

dpanici:
> add "auto", "scaling", "opt", "map2disc"

Not sure how to do this without changing completely the way that the set_initial_guess logic works...
I would like a way to force it to use map2disc if the user specifies it, but then it just would bypass the whole thing. I guess I coudl instead just add a "use_map2disc" option which will use map2disc if available and if not fall back to the usual logic, that probably is easiest as map2disc is always going to give a nested guess

Mandelbr0t:
I updated this PR so that the github checks pass.
The map2disc package was updated to a jax version (https://pypi.org/project/map2disc-jax/) which includes a test for differentiability.

desc/equilibrium/initial_guess.py
```diff
@@ -228,21 +242,71 @@ def set_initial_guess(eq, *args, ensure_nested=True):  # noqa: C901
...
+        if ensure_nested_method not in ("opt", "map2disc"):
+            raise ValueError(
+                "ensure_nested_method should be one of 'opt', 'map2disc', got "
+                + f"{ensure_nested_method}."
+            )
+        if ensure_nested_method == "map2disc":
+            try:
+                import map2disc_jax  # noqa: F401
+            except ImportError:
+                raise ImportError(
```
dario-panici, line 259: Instead of this can we just fallback to good_coordinates and throw a warning saying map2disc requested but not installed?

desc/equilibrium/initial_guess.py
```diff
@@ -373,3 +437,115 @@ def _initial_guess_points(nodes, x, x_basis):
...
+        bcm.cy = all_cy[:, z]
+        r_rt, z_rt = bcm.eval_rt(
+            rhos_grid[:, :, z].squeeze(), thetas_grid[:, :, z].squeeze()
+        )
+        Rout[:, :, z] = np.asarray(r_rt)
+        Zout[:, :, z] = np.asarray(z_rt)
+    R_lmn = np.array(Rtransform_3d.fit(grid.meshgrid_flatten(Rout, "rtz")), copy=True)
+    Z_lmn = np.array(Ztransform_3d.fit(grid.meshgrid_flatten(Zout, "rtz")), copy=True)
+
+    # TODO: smarter way would be to take the existing Zernike coeffs as fxn of phi and
```
dario-panici, line 533: @Mandelbr0t now that you rewrote it, do you know the mode orderings etc? Want to change this to do the more efficient method of doing a vmap over 1D fits of the Zernike polynomial coefficients as functions of toroidal angle phi to get the Fourier-Zernike coefficients? 

The memory needed to make the R Z lambda matrices can get large for higher resolutions so would prefer to not need to make it if we don't have to.

This works fine as is though, this is not a blocking comment just a nice-to-have thought

dario-panici (changes requested):
Thanks @Mandelbr0t for doing the map2disc JAX rewrite, this looks great! 

Left a small comment on falling back to goodcoordintres when map2disc requested but not installed, and a nice-to-have of improving the Rlmn Zlmn coordinate fits to be 1D toroidal fits of the zernike coeffs instead of the 3D brute force I did (bc I did not want to dig into map2disc to find the mode orderings for cx and cy)

Also since we now have it in jax and thus easily to be installed, do we want to now use map2disc as the default initial guess? it is waaay faster than Goodcoordinates (last time I checked) and more robust too, so my vote is in favor of having it as default. 

Also would vote to add map2disc_jax to requirements, since the idea of the rewrite to avoid the pybie2d thing fixes the issue of relying on that package (do the dev requirements get installed if one just pip installs desc-opt? idt so right?)

desc/equilibrium/initial_guess.py
Mandelbr0t: I agree, this is the better way if map2disc-jax stays a non-dep of DESC

PR #1736, opened 2025-05-23
Add exponential spectral scaling for optimization, an alternative to Fourier continuation

byoungj:
Add an exponential decaying x_scale as a preconditioner for stellarator optimizations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ New Features
         - ``VacuumGuidingCenterTrajectory`` : Integrates the particle motion by vacuum guiding center ODEs, conserving energy and mu.
     - Particle trajectories can be plotted with ``desc.plotting.plot_particle_trajectories`` function.
 - Adds new option for ``loss_function``, ``"sum"``, which changes an objective to target the sum of the values computed.
+- Adds utility functions ``desc.external.export_surface_to_paraview``, ``desc.external.export_volume_to_paraview`` and ``desc.external.export_coils_to_paraview`` to export Paraview files for surfaces, volume and coils. These functions use an optional dependency ``pyvista`` which is not automatically installed.
 - The `x_scale` option for `eq.optimize` and related functions can now be given as a dictionary mapping individual parameter names to their relevant scales,
 or if multiple things are being optimized, `x_scale` can be a list of dict, one for each optimizable thing.
 - Adds new option `x_scale='ess'` to use exponential spectral scaling from (Jang 2025) which has been shown to improve performance and robustness as an
@@ -23,7 +24,6 @@ alternative to fourier continuation methods.
 Bug Fixes
 
 - No longer uses the full Hessian to compute the scale when ``x_scale="auto"`` and using a scipy optimizer that approximates the hessian (e.g. if using ``"scipy-bfgs"``, no longer attempts the Hessian computation to get the x_scale)
-
 
 Performance Improvements
 
--- a/desc/external/README.rst
+++ b/desc/external/README.rst
@@ -1,13 +1,16 @@
 **Warning:**
 
-These objectives depend on other codes external to DESC. They are not routinely tested
+These objectives/utilities depend on other codes external to DESC. They are not routinely tested
 and may not be compatible with all versions of the codes. The DESC team is not
 responsible for maintaining or documenting those external codes, and we do not guarantee
-to regularly maintain these objectives. Use these at your own risk!
+to regularly maintain these objectives/utilities. Use these at your own risk!
 
 The external codes are not included with DESC, and you may need to obtain access to them
-to use these objectives. Those codes may not be publicly available and may require a
+to use these objectives/utilities. Those codes may not be publicly available and may require a
 license to install them.
 
 `_terpsichore.py` was last tested on September 8, 2025 with DESC v0.15.0 and a version
 of TERPSICHORE compiled on Perlmutter in January of 2025.
+
+`paraview.py` was last tested on November 24, 2025 with DESC v0.16.0, Paraview v5.13.3
+and pyvista v0.46.4.
--- a/desc/external/__init__.py
+++ b/desc/external/__init__.py
@@ -1,3 +1,8 @@
 """Classes defining objectives that wrap external codes."""
 
 from ._terpsichore import TERPSICHORE
+from .paraview import (
+    export_coils_to_paraview,
+    export_surface_to_paraview,
+    export_volume_to_paraview,
+)
--- a/desc/external/paraview.py
+++ b/desc/external/paraview.py
@@ -1,0 +1,218 @@
+"""Utility functions to export DESC objects to Paraview."""
+
+import warnings
+
+import numpy as np
+
+try:
+    import pyvista as pv
+except ImportError:
+    warnings.warn(
+        "DESC objects are exported to Paraview using `pyvista`"
+        "package which is an optional dependency. Please pip install "
+        "`pyvista` to your environment."
+    )
+
+from desc.coils import CoilSet, _Coil
+from desc.equilibrium import Equilibrium
+from desc.geometry import FourierRZToroidalSurface
+from desc.grid import LinearGrid
+
+
+def export_surface_to_paraview(
+    obj, filename, res=(100, 100), keys=[], rho=1.0, return_mesh=False
+):
+    """Export a constant rho surface data for Paraview as VTK file.
+
+    Parameters
+    ----------
+    obj : Equilibrium or FourierRZToroidalSurface
+        Object to be exported. Different flux surfaces of the Equilibrium
+        can be exported by supplying `rho`.
+    filename : str
+        Name for the saved file. The file extension will be `.vtk`.
+    res : tuple, list
+        The resolution used to create the mesh. Must be in the form (Np, Nt),
+        number of points in poloidal and toroidal direction, respectively.
+        Default is (100, 100).
+    keys : list, optional
+        The names of the quantities to be computed on the grid points. Defaults to
+        empty list.
+    rho : float
+        The flux surface to be exported if `obj` is an Equilibrium. Defaults to 1.
+    return_mesh : bool
+        If True, return the created pyvista StructuredGrid object. Defaults to False.
+
+    Returns
+    -------
+    mesh : pyvista.StructuredGrid
+        Created structured grid object. With this object one can compute more
+        quantities on `LinearGrid(rho=rho, theta=Np, zeta=Nt, NFP=1, endpoint=True)`
+        and add it to the mesh by `mesh['name'] = value`. Once the mesh is changed,
+        the user has to save it again `mesh.save(filename)`.
+    """
+    if not isinstance(obj, (Equilibrium, FourierRZToroidalSurface)):
+        raise ValueError(
+            "This function only support Equilibrium or FourierRZToroidalSurface "
+            f"objects but {type(obj)} is given."
+        )
+
+    Np, Nt = res
+    grid = LinearGrid(rho=rho, theta=Np, zeta=Nt, NFP=1, endpoint=True)
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="Unequal number of field")
+        data = obj.compute(["x"] + keys, grid=grid, basis="xyz")
+    nodes = np.asarray(data.pop("x"))
+
+    # Create the structured mesh
+    mesh = pv.StructuredGrid()
+    mesh.points = nodes
+    mesh.dimensions = (Np, Nt, 1)
+
+    # Additional data
+    for key in keys:
+        value = data[key]
+        if value.size == Nt * Np or value.shape == (Nt * Np, 3):
+            # this will handle both 1D and 3D
+            mesh[key] = value
+        else:
+            raise ValueError(f"Data {key} has an unimplemented shape ({value.shape})")
+
+    mesh.save(filename + ".vtk")
+    print(f"File is saved as {filename}.vtk")
+    if return_mesh:
+        return mesh
+
+
+def export_coils_to_paraview(coils, filename, res=100, keys=[]):
+    """Export coils data for Paraview as VTP files.
+
+    Parameters
+    ----------
+    coils : Coil, CoilSet, MixedCoilSet
+        Object to be exported. Can be a single coil or multiple coils given
+        as CoilSet or MixedCoilSet.
+    filename : str
+        Name for the saved file. The file extension will be `.vtp`.
+        If there are multiple coils, an index will be appended to the given name.
+    res : int
+        The resolution to use for all coils. Defaults to 100 points around a
+        single coil.
+    keys : list of str
+        Additional data to store in the file. By default, only the current
+        corresponding to each coil is stored.
+    """
+    if not isinstance(coils, _Coil):
+        raise ValueError(
+            "This function only support classes inherited from _Coil "
+            f"but {type(coils)} is given."
+        )
+
+    def flatten_coils(coilset):
+        if hasattr(coilset, "__len__"):
+            if hasattr(coilset, "_NFP") and hasattr(coilset, "_sym"):
+                if coilset.NFP > 1 or coilset.sym:
+                    # plot all coils for symmetric coil sets
+                    coilset = CoilSet.from_symmetry(
+                        coilset,
+                        NFP=coilset.NFP,
+                        sym=coilset.sym,
+                        check_intersection=False,
+                    )
+            return [a for i in coilset for a in flatten_coils(i)]
+        else:
+            return [coilset]
+
+    coils_list = flatten_coils(coils)
+    grid = LinearGrid(zeta=res, endpoint=True)
+
+    for i, coil in enumerate(coils_list):
+        data = coil.compute(["x"] + keys, grid=grid, basis="xyz")
+        points = np.asarray(data["x"])
+        current = getattr(coil, "current", np.nan)
+
+        # Create PolyData object
+        poly = pv.PolyData()
+        poly.points = points
+        # Connectivity of the points
+        poly.lines = np.roll(np.arange(len(points) + 1), shift=1)
+
+        # Add current as a scalar field
+        current_array = np.full(len(points), current)
+        poly["current"] = current_array
+
+        # Additional data
+        for key in keys:
+            value = data[key]
+            if value.size == res or value.shape == (res, 3):
+                # this will handle both 1D and 3D
+                poly[key] = value
+            else:
+                raise ValueError(
+                    f"Data {key} has an unimplemented shape ({value.shape})"
+                )
+
+        poly.save(f"{filename}_{i}.vtp")
+
+    print(f"Saved {len(coils_list)} coils with name format `{filename}_i.vtp`")
+
+
+def export_volume_to_paraview(
+    eq, filename, res=(20, 100, 100), keys=[], return_mesh=False
+):
+    """Export equilibrium volume data for Paraview as VTK file.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium to be exported.
+    filename : str
+        Name for the saved file. The file extension will be `.vtk`.
+    res : tuple of ints, (Nr, Np, Nt)
+        The resolutions to use. Defaults to (20, 100, 100) points in radial, poloidal
+        and toroidal directions, respectively.
+    keys : list of str
+        Additional data to store in the file. Defaults to empty list.
+    return_mesh : bool
+        If True, return the created pyvista StructuredGrid object. Defaults to False.
+
+    Returns
+    -------
+    mesh : pyvista.StructuredGrid
+        Created structured grid object. With this object one can compute more
+        quantities on `LinearGrid(rho=Nr, theta=Np, zeta=Nt, NFP=1, endpoint=True)`
+        and add it to the mesh by `mesh['name'] = value`. Once the mesh
+        is changed, the user has to save it again `mesh.save(filename)`.
+    """
+    if not isinstance(eq, Equilibrium):
+        raise ValueError(
+            "This function only support classes of type Equilibrium "
+            f"but {type(eq)} is given."
+        )
+    Nr, Np, Nt = res
+    grid = LinearGrid(rho=Nr, theta=Np, zeta=Nt, NFP=1, endpoint=True)
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="Unequal number of field")
+        data = eq.compute(["x"] + keys, grid=grid, basis="xyz")
+    nodes = np.asarray(data.pop("x"))
+
+    # Create the structured mesh
+    mesh = pv.StructuredGrid()
+    mesh.points = nodes
+    mesh.dimensions = (Np, Nr, Nt)
+
+    # Additional data
+    for key in keys:
+        value = data[key]
+        if value.size == Nr * Nt * Np or value.shape == (Nr * Nt * Np, 3):
+            # this will handle both 1D and 3D
+            mesh[key] = value
+        else:
+            raise ValueError(f"Data {key} has an unimplemented shape ({value.shape})")
+
+    mesh.save(filename + ".vtk")
+    print(f"File is saved as {filename}.vtk")
+    if return_mesh:
+        return mesh
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -96,6 +96,21 @@ Examples
     desc.examples.get
     desc.examples.listall
 
+External
+********
+Note that these objectives and utility functions are not regularly tested and maybe incompatible with newer version
+of the external code. Please see the README of the `desc.external` for more details.
+
+.. autosummary::
+    :toctree: _api/external
+    :recursive:
+
+    desc.external.export_coils_to_paraview
+    desc.external.export_surface_to_paraview
+    desc.external.export_volume_to_paraview
+    desc.external.TERPSICHORE
+
+
 Geometry
 ********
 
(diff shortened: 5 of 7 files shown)


desc/optimize/optimizer.py
```diff
@@ -211,6 +226,19 @@ def optimize(  # noqa: C901
...
+                ess_min_value = options.pop("ess_min_value", 1e-7)
+                x_scale = _create_exponential_spectral_scale(
+                    eq=eq, alpha=ess_alpha, scale_type=ess_type, min_value=ess_min_value
+                )
+                # This assumes that Equilibrium is first in things
+                for i, t in enumerate(things):
+                    if isinstance(t, OmnigenousField):
+                        x_scale = np.concatenate([x_scale, np.ones(t.dim_x)])
+
         options = {} if options is None else options
```
YigitElma, line 242: You can delete lines 242 and 244, then move line 230 to line 213. I think these are left from some older update

desc/optimize/optimizer.py
```diff
@@ -211,6 +226,19 @@ def optimize(  # noqa: C901
...
+                options = {} if options is None else options
+                ess_alpha = options.pop("ess_alpha", 1.2)
+                ess_type = options.pop("ess_type", "Linf")
+                ess_min_value = options.pop("ess_min_value", 1e-7)
+                x_scale = _create_exponential_spectral_scale(
+                    eq=eq, alpha=ess_alpha, scale_type=ess_type, min_value=ess_min_value
+                )
+                # This assumes that Equilibrium is first in things
+                for i, t in enumerate(things):
+                    if isinstance(t, OmnigenousField):
```
YigitElma, line 239: I don't think this should be specific to `OmnigeousField`. All other `Optimizable` things that are not `Equilibrium` should add ones.

desc/optimize/optimizer.py
```diff
@@ -211,6 +226,19 @@ def optimize(  # noqa: C901
...
 
+            if x_scale == "ess":
+                options = {} if options is None else options
+                ess_alpha = options.pop("ess_alpha", 1.2)
+                ess_type = options.pop("ess_type", "Linf")
+                ess_min_value = options.pop("ess_min_value", 1e-7)
+                x_scale = _create_exponential_spectral_scale(
+                    eq=eq, alpha=ess_alpha, scale_type=ess_type, min_value=ess_min_value
+                )
+                # This assumes that Equilibrium is first in things
```
YigitElma, line 237: instead of this assumption, you can call `_create_exponential_spectral_scale` inside the for loop `if isinstance(t, Equilibrium)`, `else` create an array of ones. At the end, concatenate all of them. This will have the proper ordering

desc/optimize/optimizer.py
```diff
@@ -614,6 +642,62 @@ def _get_default_tols(
     return stoptol
 
 
+##TODO: Add Diagnostics for x_scale to track what modes are being scaled
```
YigitElma, line 645: I am pretty sure that `x_scale` should be the size of full state vector not just `Rb_lmn, Zb_lmn` etc. We have a logic [here](https://github.com/PlasmaControl/DESC/blob/5a5ac103a56fa84e1c815e6b0b4cda526c8a2d07/desc/optimize/optimizer.py#L550)  that accounts for which variables end up in the reduced state vector.

desc/optimize/optimizer.py
```diff
@@ -614,6 +642,62 @@ def _get_default_tols(
...
+    Returns
+    -------
+    ndarray
+        Array of scale values for each parameter
+    """
+    # Initialize x_scale array
+    total_size = (
+        len(eq.p_l) + len(eq.c_l) + len(eq.i_l) + 1 + len(eq.Rb_lmn) + len(eq.Zb_lmn)
+    )
+    x_scale = np.zeros(total_size)
```
YigitElma, line 675: This will set the effect of pressure, current, iota etc to be 0. Do you mean `np.ones`?

desc/optimize/optimizer.py
```diff
@@ -614,6 +642,62 @@ def _get_default_tols(
...
+            mode_level = np.sqrt(np.sum(modes**2, axis=1))  # sqrt(m² + n²)
+        elif scale_type == "Linf":
+            mode_level = np.max(modes, axis=1)  # max(|m|,|n|)
+        return np.maximum(np.exp(-alpha * mode_level) / np.exp(-alpha), min_value)
+
+    R_scales = get_mode_scales(R_modes)
+    Z_scales = get_mode_scales(Z_modes)
+
+    start_idx = len(eq.p_l) + len(eq.c_l) + len(eq.i_l) + 1
+    x_scale[start_idx : start_idx + len(eq.Rb_lmn)] = R_scales
```
YigitElma, line 695: We have this information in cleaner way.
`x_scale[eq.x_idx["Rb_lmn"]] = R_scales`

desc/optimize/optimizer.py
```diff
@@ -614,6 +642,62 @@ def _get_default_tols(
...
+    x_scale[start_idx + len(eq.Rb_lmn) :] = Z_scales
+
+    return x_scale
```
YigitElma, line 698: Currently, if I try to run `eq.solve(x_scale='ess')`, it fails because of the dimensional issues. I think this function should find the scale for `R_lmn`, `Z_lmn` and maybe `L_lmn`, not just the optimizable parameter of fixed boundary problem. There are couple reasons for this.
1. Your logic is specific to fixed boundary problem, but all optimizations can call this (i.e. optimizing for QA where some of the Rb, Zb is fixed)
2. In practice, yes, we want to scale Rb, Zb but the optimizer might end up choosing some other variables as optimizable (the ones that are in the null space), their linear combination again will be equivalent to Rb, Zb but I am not 100% sure that this will work.

https://github.com/PlasmaControl/DESC/blob/5a5ac103a56fa84e1c815e6b0b4cda526c8a2d07/desc/optimize/optimizer.py#L550 must solve these problems, if you change the function to give an `x_scale` that is the same size as `eq.dim_x`.

YigitElma: This might work for example,
```python
    ...
    # Initialize x_scale array
    x_scale = np.ones(eq.dim_x)

    # Get R and Z modes (m,n pairs)
    R_modes = np.abs(eq.R_basis.modes[:, 1:])
    Z_modes = np.abs(eq.Z_basis.modes[:, 1:])
    # you might not want to scale lambda I am not sure
    L_modes = np.abs(eq.L_basis.modes[:, 1:])

    # Calculate mode scales based on scale_type
    def get_mode_scales(modes):
        if scale_type == "L1":
            mode_level = np.sum(modes, axis=1)  # |m| + |n|
        elif scale_type == "L2":
            mode_level = np.sqrt(np.sum(modes**2, axis=1))  # sqrt(m² + n²)
        elif scale_type == "Linf":
            mode_level = np.max(modes, axis=1)  # max(|m|,|n|)
        return np.maximum(np.exp(-alpha * mode_level) / np.exp(-alpha), min_value)

    R_scales = get_mode_scales(R_modes)
    Z_scales = get_mode_scales(Z_modes)
    L_scales = get_mode_scales(L_modes)

    x_scale[eq.x_idx["R_lmn"]] = R_scales
    x_scale[eq.x_idx["Z_lmn"]] = Z_scales
    x_scale[eq.x_idx["L_lmn"]] = L_scales

    return x_scale
```

YigitElma (changes requested):
Thank you very much for adding this!

There are a couple of changes that need to be introduced to make it compatible with DESC internals. I know this is still draft, but I hope my suggestions help you to save some time :)

desc/optimize/optimizer.py
byoungj: Thank you for all the comments :) This is a very clean way to handle this!  I like it. I will have to add a logic handling if the objective is a ProximalProjection class as proximal projections adjust the eq optimizable variables to only include the Rb, Zb and profile coefficients. This is probably why you were seeing the dimensional issues.

desc/optimize/optimizer.py
byoungj: I believe if the objective is a ProximalProjection class, the logic won't be enough as objective._objective.dim_x doesn't equal to eq.dim_x.

desc/optimize/optimizer.py
```diff
@@ -579,6 +584,22 @@ def get_combined_constraint_objectives(
...
+    if is_prox and not isinstance(x_scale, str):
+        # If we have ProximalProjection, get original dimension from equilibrium
+        # and project x_scale through both stages: (eq.dim_x) ->
+        # (dim size post proximal projection)
+        original_dim = objective._objective._eq.dim_x
+        x_scale = np.broadcast_to(x_scale, original_dim)
+
+        # Project from (eq.dim_x) -> (dim size post proximal projection)
+        # (remove excluded parameters from ProximalProjection)
+        excluded_params = ["R_lmn", "Z_lmn", "L_lmn", "Ra_n", "Za_n"]
```
byoungj: I considered using ProximalProjection._feasible_tangents here as an analogue to LinearConstraintProjection._Z, but it didn’t seem to serve the same purpose.

desc/equilibrium/equilibrium.py
```diff
@@ -2305,6 +2305,12 @@ def optimize(
             along any of the scaled variables has a similar effect on the cost
             function. If set to ``'auto'``, the scale is iteratively updated using the
             inverse norms of the columns of the Jacobian or Hessian matrix.
+            If set to ``'ess'``, the scale is set using Exponential Spectral Scaling,
+            this scaling is set with two parameters, ``ess_alpha`` and ``ess_type``.
```
YigitElma:
```suggestion
            this scaling is set with two parameters, ``ess_alpha`` and ``ess_type`` which are passed through ``options``.
```

desc/equilibrium/equilibrium.py
```diff
@@ -2305,6 +2305,12 @@ def optimize(
...
+            ``ess_alpha`` is the decay rate of the scaling, and ``ess_type`` is the type
+            of scaling, which can be ``'L1'``, ``'L2'``, or ``'Linf'``.
+            Default is ``'Linf'`` and ``ess_alpha`` is 1.2. ``ess_min_value`` is the
```
YigitElma: This may be a bit confusing. Maybe mention that if not passed in by `options`, the defaults will be set as these. The default for the `x_scale` is `auto`.

desc/optimize/optimizer.py
```diff
@@ -122,15 +122,26 @@ def optimize(  # noqa: C901
...
+            sum of ``t.dim_x`` for each ``t`` in ``things``). For Equilibrium objects,
+            the ordering must match ``eq.x_idx`` parameter indices. The array should
+            be ordered by concatenating parameters from each object in ``things`` in
+            the same order. If set to ``'auto'``, the scale is iteratively updated
+            using the inverse norms of the columns of the Jacobian or Hessian matrix.
+            If set to ``'ess'``, the scale is set using Exponential Spectral Scaling,
+            this scaling is set with two parameters, ``ess_alpha`` and ``ess_type``.
+            ``ess_alpha`` is the decay rate of the scaling, and ``ess_type`` is the type
+            of scaling, which can be ``'L1'``, ``'L2'``, or ``'Linf'``.
+            Default is ``'Linf'`` and ``ess_alpha`` is 1.2. ``ess_min_value`` is the
```
YigitElma: Same as above

desc/optimize/optimizer.py
```diff
@@ -211,6 +230,24 @@ def optimize(  # noqa: C901
...
+                            eq=t,
+                            alpha=ess_alpha,
+                            scale_type=ess_type,
+                            min_value=ess_min_value,
+                        )
+                    else:
+                        t_scale = np.ones(t.dim_x)
+                    x_scales.append(t_scale)
+                x_scale = np.concatenate(x_scales)
         options = {} if options is None else options
```
YigitElma: Move this line before the `if isinstance(x_scale, str) and x_scale == "ess":` line and delete the first line in the if (since it is same as this one).

desc/optimize/optimizer.py
```diff
@@ -211,6 +230,24 @@ def optimize(  # noqa: C901
...
         timer = Timer()
         options = {} if options is None else options
```
YigitElma: This one is also unnecessary.

desc/optimize/optimizer.py
YigitElma: Yes, they are very similar in concept, but have different shapes. In a nutshell, `LinearConstrainProjection._Z` doens't have 0 rows for fixed variables but `LinearConstrainProjection._feasible_tangents` adds some 0 rows. For `ProximalProjection` the story is the same. What you need might be `self._eq_Z * self._eq_D[self._eq_unfixed_idx, None]` or `self._dxdc` or something like that, I am not very sure.

desc/optimize/optimizer.py
```diff
@@ -547,6 +584,22 @@ def get_combined_constraint_objectives(
...
+        original_dim = objective._objective._eq.dim_x
+        x_scale = np.broadcast_to(x_scale, original_dim)
+
+        # Project from (eq.dim_x) -> (dim size post proximal projection)
+        # (remove excluded parameters from ProximalProjection)
+        excluded_params = ["R_lmn", "Z_lmn", "L_lmn", "Ra_n", "Za_n"]
+        included_idx = []
+        for arg in objective._objective._eq.optimizable_params:
+            if arg not in excluded_params:
+                included_idx.extend(objective._objective._eq.x_idx[arg])
```
YigitElma: Same as above

desc/optimize/optimizer.py
```diff
@@ -547,6 +584,22 @@ def get_combined_constraint_objectives(
...
+        x_scale = x_scale[included_idx]
```
YigitElma: To account for other `things`, you can maybe do something like this that we have in other places in the code.
```python
x_scale = jnp.split(
            x_scale, np.cumsum(self._dimx_per_thing), axis=-1
        )
x_scale[self._eq_idx] = x_scale[self._eq_idx][included_idx]
```

desc/optimize/optimizer.py
```diff
@@ -547,6 +584,22 @@ def get_combined_constraint_objectives(
             )
             nonlinear_constraint.build(verbose=verbose)
 
+    if is_prox and not isinstance(x_scale, str):
+        # If we have ProximalProjection, get original dimension from equilibrium
+        # and project x_scale through both stages: (eq.dim_x) ->
+        # (dim size post proximal projection)
+        original_dim = objective._objective._eq.dim_x
```
YigitElma: If I am not mistaken, `objective` is `ProximalProjection`, so you can just call `objective._eq.dim_x`. `objective._objective` is an `ObjectiveFunction` and doesn't have `_eq`.

desc/optimize/optimizer.py
YigitElma: Wouldn't just removing the unnecessary stuff be enough? Multiplication by null-space happens at somewhere as for x_scale?

desc/optimize/optimizer.py
YigitElma: Also, noting that this is only intended for equilibrium optimization, and not for equilibrium solve might be helpful as it is only getting scales for Rb and Zb

desc/equilibrium/equilibrium.py
```diff
@@ -2296,7 +2296,7 @@ def optimize(
             stopping tolerances. `None` will use defaults for given optimizer.
         maxiter : int
             Maximum number of solver steps.
-        x_scale : array_like or ``'auto'``, optional
+        x_scale : array_like or ``'auto'`` or ``'ess'``, optional
```
YigitElma: May ask you to add a check in `eq.solve` like,
```python
if isinstance(x_scale, str):
    assert x_scale in ["auto"]

Comment (dpanici) [reply desc/equilibrium/equilibrium.py]: this could techincally be changed in the future though, so maybe add something like `assert x_scale in ["auto"], "ess scaling currently not supported for equilibrium solves"`

File: desc/optimize/optimizer.py (hunk 11 above)
Comment (dpanici) [reply desc/optimize/optimizer.py]: question I should probably know: is x_scale supposed to be an array of every DOF BEFORE applying the linear constraints and projecting down to the feasible subspace? or after?

File: desc/optimize/optimizer.py [hunk 18]
```diff
@@ -614,6 +667,59 @@ def _get_default_tols(
     return stoptol
 
 
+##TODO: Add Diagnostics for x_scale to track what modes are being scaled
+def _create_exponential_spectral_scale(
```
dpanici: Future extension of this would be to also allow other objects, or to make this more user-accessible/usable so they can specify themselves, I am thinking like fourier coefficients for coilsets or just FourierRZToroidalSurface objects

desc/optimize/optimizer.py
```diff
@@ -614,6 +667,59 @@ def _get_default_tols(
...
+    eq, alpha=1.2, scale_type="Linf", min_value=1e-7
+):
+    """Create x_scale using exponential spectral scaling.
+
+    Parameters
+    ----------
+    eq : Equilibrium
```
dpanici: Since right now this is only the surface fourier modes, this should also work if `FourierRZToroidalSurface` is passed in, would just need some extra logic (`R_lmn` for the surface instead of `Rb_lmn` for the eq)

desc/optimize/optimizer.py
f0uriest: before (otherwise the dofs get jumbled and any physical intuition gets weird)

f0uriest:
One thought that might make future versions of this easier, is if we allow the `x_scale` the user passes in to be a pytree with the same structure as `things.params_dict` that we then parse into the correct vector to hand off to the optimizer. 
eg instead of 
```
x_scale = np.array([1,2,3,4,5,6])
```
we do
```
x_scale = {"R_lmn": [1,2,3], "Z_lmn": [4,5,6]}
```

dpanici:
> Thanks for checking in @dpanici! I will get back to this today/tomorrow so we can push to main :)

I'm also happy to help with any of the dev work!

YigitElma:
I would also be happy to help!

desc/optimize/optimizer.py
byoungj: My intention was to future proof if the logic within ProximalProjection is changed down the line, you are right how it is implemented now should suffice.

desc/optimize/optimizer.py
byoungj: I believe here the objective._objective is ProximalProjection if there is LinearConstraint which seems to wrap on top of the ProximalProjection. I can add a logic to check this.

dpanici:
@f0uriest  Add to `Optimizable` a (private) method to compute ess scaling per object (takes in alpha), we can then use this in optimizer.py when `"ess"` is used

f0uriest:
I played around with ESS for regular fixed boundary solve and it seems to work quite well, surfaces look identical by eye and force error is more or less the same. These are 3 highly shaped equilibria where we usually needed to use our continuation/perturbation method to get good solutions

## HSX
image
image

## W7-X
image
image

## NCSX
image
image

desc/equilibrium/equilibrium.py
```diff
@@ -2490,6 +2515,46 @@ def perturb(
...
+        # we use ESS for the following:
+        modes = {
+            "R_lmn": self.R_basis.modes,
+            "Z_lmn": self.Z_basis.modes,
+            "L_lmn": self.L_basis.modes,
+            "Rb_lmn": self.surface.R_basis.modes,
+            "Zb_lmn": self.surface.Z_basis.modes,
+        }
+        if hasattr(self.surface, "Phi_mn"):
+            modes["Phi_mn"] = self.surface.Phi_basis.modes
```
f0uriest: question for yall: Do we also want to include a factor of mu0 or something here?
ESS takes care of the disparity due to the spectral basis, but not due to units. So we could also scale this and coil currents etc to make it more reasonable?

dpanici:
Something I noticed: right now using `x_scale="ess"` and doing an augmented lagrangian optimization with inequality constraints, you will get an error due to shape mismatch bc the ess x_scale used is just for x, but when inequality constraints exist we have additional slack variables  due to the constraints. The usual `"auto"` jac scale takes care of this bc it uses `lagjac` to compute the jacobian, but we will either have to add ones for the slack variables or add some heuristic for them (or use jac scaling for just them?)

desc/equilibrium/equilibrium.py
dpanici: I think a factor of mu0 would be good, maybe we could also add to the current for the coils, other codes I think already do this sort of scaling internally

desc/equilibrium/equilibrium.py
```diff
@@ -2242,18 +2243,27 @@ def solve(
...
-            have the same keys and dimensions as thing.params_dict.
+            function. Default is ``'auto'``, which iteratively updates the scale using
+            the inverse norms of the columns of the Jacobian or Hessian matrix.
+            If set to ``'ess'``, the scale is set using Exponential Spectral Scaling,
+            this scaling is set with two parameters, ``ess_alpha`` and ``ess_type``
+            which are passed through ``options``. ``ess_alpha`` is the decay rate of
+            the scaling, and ``ess_type`` is the type of scaling, which can be
+            ``1``, ``2``, or ``np.inf``. If not provided in ``options``, the
+            defaults are: ``ess_alpha=1.2`` (decay rate), ``ess_type=np.inf'`` (scaling
+            type, can be ``1``, ``2``, or ``np.inf``), and ``ess_min_value=1e-7``
```
YigitElma: `(scaling type, can be ``1``, ``2``, or ``np.inf``)` part is repeated

YigitElma: Also, maybe mention that those specify the order of the norm.

desc/equilibrium/equilibrium.py
```diff
@@ -2351,18 +2361,27 @@ def optimize(
...
```
YigitElma: Same as above

desc/optimize/optimizer.py
```diff
@@ -124,18 +124,27 @@ def optimize(  # noqa: C901
...
-            of dict, the list should have 1 element for each thing, and each dict should
-            have the same keys and dimensions as thing.params_dict.
+            function. Default is ``'auto'``, which iteratively updates the scale using
+            the inverse norms of the columns of the Jacobian or Hessian matrix.
+            If set to ``'ess'``, the scale is set using Exponential Spectral Scaling,
+            this scaling is set with two parameters, ``ess_alpha`` and ``ess_type``
+            which are passed through ``options``. ``ess_alpha`` is the decay rate of
+            the scaling, and ``ess_type`` is the type of scaling, which can be
+            ``1``, ``2``, or ``np.inf``. If not provided in ``options``, the
+            defaults are: ``ess_alpha=1.2`` (decay rate), ``ess_type=np.inf'`` (scaling
```
YigitElma: Same above

desc/equilibrium/equilibrium.py
dpanici: @dpanici try for 

- [ ] coil current at least

- [ ] and do for free bdry (with ess) I,G,Phi_mn

desc/optimize/optimizer.py
```diff
@@ -371,44 +393,85 @@ def _parse_x_scale(x_scale, things, options):
...
+        if x_scale == "auto":
+            return x_scale
+        if x_scale == "ess":
+            x_scale = ["ess"] * len(things)
+        else:
+            raise ValueError(
+                "only 'auto' and 'ess' are allowed string values for x_scale"
+            )
     if isinstance(x_scale, (jnp.ndarray, np.ndarray)) or (
         np.isscalar(x_scale) and not isinstance(x_scale, str)
```
YigitElma, line 406: I guess `and not isinstance(x_scale, str)` part is unnecessary here

desc/optimize/optimizer.py
```diff
@@ -371,44 +393,85 @@ def _parse_x_scale(x_scale, things, options):
...
+
     all_scales = []
+    ess_alpha = options.pop("ess_alpha", 1.2)
+    ess_type = options.pop("ess_type", np.inf)
+    ess_min_value = options.pop("ess_min_value", 1e-7)
+
     for xsc, tng in zip(x_scale, things):
-        if isinstance(xsc, (jnp.ndarray, np.ndarray)) or np.isscalar(xsc):
+        if isinstance(xsc, (jnp.ndarray, np.ndarray)) or (
+            np.isscalar(xsc) and not isinstance(xsc, str)
```
YigitElma, line 424: Again, is `and not isinstance(xsc, str)` really necessary? Not that it is causing any problems but just for clarity.

desc/optimize/optimizer.py
```diff
@@ -371,44 +393,85 @@ def _parse_x_scale(x_scale, things, options):
...
 
 def _project_x_scale(x_scale, objective):
     """Project x_scale vector to remove fixed DoFs etc."""
-    if isinstance(objective, LinearConstraintProjection) and not isinstance(
-        x_scale, str
-    ):
+    if isinstance(x_scale, str):
+        return x_scale
+
+    is_prox = isinstance(objective, ProximalProjection) or (
```
YigitElma, line 446: Is `isinstance(objective, ProximalProjection)` possible? I don't think it can be unless if the user calls `optimizer.optimize` with their own objective.

desc/optimize/optimizer.py
f0uriest: this is needed bc numpy treats strings as scalars, so eg `np.isscalar("ess") == True`

desc/optimize/optimizer.py
f0uriest: I think if you were to only give an equilibrium constraint, without any other linear constraints? Unlikely, but possible

desc/optimize/optimizer.py
YigitElma: oh ok!

PR #1744, opened 2025-05-29
Ability to `safenormalize` multiple vectors

### LINKED ISSUE #1743: Error in `safenormalize` when used with multiple vectors
```python
import numpy as np
from desc.utils import safenormalize

a = np.array([[1,2,3],[4,5,6],[7,8,9]])
a_norm = a / np.linalg.norm(a,axis=-1)[:,None]
a_safenorm = safenormalize(a,axis=1)

print(np.linalg.norm(a_norm,axis=1))
print(np.allclose(np.linalg.norm(a_norm,axis=1), 1))
print(np.linalg.norm(a_safenorm,axis=1))
print(np.allclose(np.linalg.norm(a_safenorm,axis=1), 1))
```

```
[1. 1. 1.]
True
[0.41202958 1.28572916 2.17914997]
False
```

I think the culprit is [this line](https://github.com/PlasmaControl/DESC/blob/ed506fa5b3a78e6fcfdc1d24a88dc4f5f0266c4b/desc/utils.py#L939)

```python
n = safenorm(x, ord, axis, fill, threshold) * jnp.ones_like(x)
```

Which is the transpose of what we actually want (we want an array of the same size as `x` but with the norm of the vector in each place, so that n[0,:] is the norm of x[0,:]) i.e. currentlt for the above `n` is

```
array([[ 3.74165739,  8.77496439, 13.92838828],
       [ 3.74165739,  8.77496439, 13.92838828],
       [ 3.74165739,  8.77496439, 13.92838828]])
```

when it really should be 

```
array([[ 3.74165739,  3.74165739,  3.74165739],
       [ 8.77496439,  8.77496439,  8.77496439],
       [13.92838828, 13.92838828, 13.92838828]])
```

dpanici:
Resolves #1743

### DIFF
--- a/desc/utils.py
+++ b/desc/utils.py
@@ -919,6 +919,7 @@ def safenorm(x, ord=None, axis=None, fill=0, threshold=0, keepdims=False):
     n = jnp.linalg.norm(y, ord=ord, axis=axis)
     n = jnp.where(is_zero.squeeze(), fill, n)  # replace norm with zero if is_zero
     if keepdims:
+        axis = 0 if axis is None else axis
         n = jnp.expand_dims(n, axis)
     return n
 
--- a/desc/utils.py
+++ b/desc/utils.py
@@ -893,7 +893,7 @@ def cross(a, b, axis=-1):
     return jnp.cross(a, b, axis=axis)
 
 
-def safenorm(x, ord=None, axis=None, fill=0, threshold=0):
+def safenorm(x, ord=None, axis=None, fill=0, threshold=0, keepdims=False):
     """Like jnp.linalg.norm, but without nan gradient at x=0.
 
     Parameters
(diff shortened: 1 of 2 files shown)


desc/utils.py
```diff
@@ -936,9 +936,13 @@ def safenormalize(x, ord=None, axis=None, fill=0, threshold=0):
     """
     is_zero = (jnp.abs(x) <= threshold).all(axis=axis, keepdims=True)
     y = jnp.where(is_zero, jnp.ones_like(x), x)  # replace x with ones if is_zero
-    n = safenorm(x, ord, axis, fill, threshold) * jnp.ones_like(x)
+    n = jnp.atleast_1d(safenorm(x, ord, axis, fill, threshold))[
+        :, None
```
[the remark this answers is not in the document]
dpanici: Good catch, I think the current version is correct now

YigitElma (approved):

desc/utils.py
unalmis, line 1: this was already done since early this year I think on 1360 for other reasons btw

rahulgaur104 (approved):

desc/utils.py
ddudt: This solution adds documentation and tests, so @unalmis remember to accept these changes in #1360 over what you had once it gets merged in.

tests/test_utils.py
```diff
@@ -268,3 +275,29 @@ def test_take_mask():
...
+@pytest.mark.unit
+def test_safenormalize():
+    """Test safenormalize on single and multiple vectors."""
+    a = np.array([1, 2, 3])
+    a_norm = a / np.linalg.norm(a, axis=-1)
+    a_safenorm = safenormalize(a, axis=-1)
+
+    np.testing.assert_allclose(a_norm, a_safenorm)
+    np.testing.assert_allclose(np.linalg.norm(a_safenorm, axis=-1), 1)
+    # 2d array of vectors
```
ddudt, line 289: my OCD wants to have a line break here

ddudt (approved):

tests/test_utils.py
dpanici: overruled

PR #1747, opened 2025-05-31
Ability to get eigenfunctions and top k eigenvalues from the ideal ballooning solver.

### LINKED ISSUE #1595: Fix period-m cycling between the ideal ballooning eigenvalues
For the ideal ballooning objective we only use the maximum eigenvalue for a given field line. However, after becoming stable, the max eigenvalue can cycle between the most unstable and the second most unstable eigenvalue causing the optimizer to get stuck. The eigenvalues can also cycle between the first m most unstable eigenvalues.

Therefore, a static argument that allows the user to set the number of m most unstable eigenvalues to use during optimization would be useful.

rahulgaur104:
Resolves #1595 

The user can specify an integer Neigvals to the compute function and get the larges Neigvals eigenvalues + and the corresponding eigenfunctions.

Neigvals can also be specified to the BallooningStability objective.


desc/compute/_stability.py
```diff
@@ -265,7 +265,10 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
     source_grid_requirement={"coordinates": "raz", "is_meshgrid": True},
     zeta0="array: points of vanishing integrated local shear to scan over. "
     "Default 15 points linearly spaced in [-π/2,π/2]",
+    # --no-verify Neigvals="int: Number of top eigenvalues to retain. "
```
dpanici: either include this as an actual kwarg or remove these lines

desc/compute/_stability.py
```diff
@@ -265,7 +265,10 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
...
+    # --no-verify "Default is 1",
 )
+# Neigvals above works for compute but not optimization
```
dpanici: what do you mean? if it does not work for some reason with optimization then just remove the kwargs etc entirely

desc/compute/_stability.py
```diff
@@ -398,11 +401,26 @@ def _ideal_ballooning_gamma2(params, transforms, profiles, data, **kwargs):
 
     A_redo = B_inv @ A @ jnp.transpose(B_inv, axes=(0, 1, 3, 2))
 
-    w, _ = jnp.linalg.eigh(A_redo)
-    # max over "zeta" axis, still a function of rho, alpha, zeta0
-    gamma = jnp.real(jnp.max(w, axis=(2,)))
+    # TODO: Try jax.scipy.eigh_tridiagonal for increased performance
```
dpanici: add an issue for this

desc/compute/_stability.py
rahulgaur104: Done

desc/compute/_stability.py
rahulgaur104: I get a "static attribute" error if I use constants["Neigvals"] instead of self._Neigvals.

rahulgaur104: It's complicated. For now, I have removed the comment.
Basically, it's not read as a "static attribute".

desc/compute/_stability.py
rahulgaur104: Done!

rahulgaur104:
Unit test 6, 3.12 failed due to OOM or something. Rerunning tests

desc/compute/_stability.py
```diff
@@ -398,11 +398,27 @@ def _ideal_ballooning_gamma2(params, transforms, profiles, data, **kwargs):
 
     A_redo = B_inv @ A @ jnp.transpose(B_inv, axes=(0, 1, 3, 2))
 
-    w, _ = jnp.linalg.eigh(A_redo)
-    # max over "zeta" axis, still a function of rho, alpha, zeta0
-    gamma = jnp.real(jnp.max(w, axis=(2,)))
+    # TODO: Issue #1750
+    # Try jax.scipy.eigh_tridiagonal for increased performance
```
f0uriest: is A tridiagonal? I thought it was dense?

Also FYI I've used eigh_tridiagonal for some other stuff, and on CPU it can be faster but its purely sequential so its pretty bad on GPU. also back when I tried it, it could only give eigenvalues, not eigenvectors.

desc/compute/_stability.py
```diff
@@ -398,11 +398,27 @@ def _ideal_ballooning_gamma2(params, transforms, profiles, data, **kwargs):
...
+    w, v = jnp.linalg.eigh(A_redo)
 
-    data["ideal ballooning lambda"] = gamma.flatten()
+    flat_w = w.flatten()
+
+    top_eigvals, top_flat_indices = jax.lax.top_k(flat_w, k=N_zeta - 2)
```
f0uriest: can you summarize the difference here?

it used to take max over axis=2 which iirc is the "zeta" axis, so it gives one value per fieldline?

And now it gives N_zeta-2 values, independent of the number of fieldlines?

From the description in the PR/changelog I would think this topk stuff would be moved to the objective, and have this always return the full spectrum?

desc/compute/_stability.py
```diff
@@ -398,11 +398,27 @@ def _ideal_ballooning_gamma2(params, transforms, profiles, data, **kwargs):
...
+    alpha_idx, zeta0_idx, theta_idx = jnp.unravel_index(top_flat_indices, w.shape)
+
+    # v becomes less than the precision at some theta points which gives NaNs
+    # stop_gradient prevents that. Not sure how it will affect an objective that
+    # requires both the eigenvalue and eigenfunction
+    top_eigenfuns = jnp.transpose(jax.lax.stop_gradient(v), axes=(0, 1, 3, 2))[
+        alpha_idx, zeta0_idx, theta_idx
+    ]
+    # Combine eigenvalue and eigenfunction data (Neigvals, Ntheta-1)
+    # Eigenvalues are on the top row
```
f0uriest: from an API standpoint I think it makes more sense to treat the eigenvalues and eigenfunctions as seperate qtys with different keys (can use the same compute fn, we have some other ones that compute 2 qtys at the same time). This also can help the compiler reduce calculations if the eigenvectors aren't used.

desc/objectives/_stability.py
```diff
@@ -603,6 +610,12 @@ def compute(self, params, constants=None):
 
         lambda0, w0, w1 = constants["lambda0"], constants["w0"], constants["w1"]
 
+        # If I used constants["Neigvals"], I get an error about static argnums
```
f0uriest: this is poor naming on our part. `constants` means constant during optimization, not the same as "constant" aka "static" in jax calculations. using `self._Neigvals` is fine, I think in the future we might get rid of `constants` since it was sort of a hack to get around some jax pytree stuff that has since been resolved.

desc/compute/_stability.py
```diff
@@ -398,11 +398,27 @@ def _ideal_ballooning_gamma2(params, transforms, profiles, data, **kwargs):
...
+    top_eigvals_and_eigfuns = jnp.concat((top_eigvals[:, None], top_eigenfuns), axis=1)
+    data["ideal ballooning lambda"] = top_eigvals_and_eigfuns.flatten()
```
f0uriest: would be useful to include a comment here to remind us what the "natural" shape of this is since its not the usual (nrho, ntheta, nzeta). AKA, if we want to reshape it to separate eigenvalues/eigenvectors by fieldline or by zeta etc what is the correct `reshape`

desc/compute/_stability.py
rahulgaur104: For a given alpha and zeta_0, the eigenvalue matrix should be symmetric, real and tridiagonal.

A is a 4D array, comprising 2D traidiagonal matrices stacked in the alpha and zeta_0 dimensions.

desc/compute/_stability.py
f0uriest: Also I think in theory we may not need to flatten at all. Since the calculation is always done on a meshgrid, we could leave the shape as (rho, alpha, zeta) or something? The usual flattening we do is mostly because not all grids have a tensor product shape, and we need things to broadcast correctly, but we don't really need that in this case.

desc/compute/_stability.py
rahulgaur104: Yeah, I see why this is confusing. I wanted k to be a fixed number > Neigvals so I chose N_zeta - 2. But the total number of eigenvalues should be (N_zeta-2) * N_alpha * N_zeta0. But, I don't want to store all the eigenvectors corresponding to these eigenvalues in memory.

For N_zeta ~ 2048, N_alpha = 8, N_zeta0 ~ 16, the matrix A and v occupy 4 GBs each for a single flux surface. So someone running this on their CPU might end up getting OOMs even when they only needed the top eigenvalues.
I tried passing Neigvals as a static keyword argument but that fails during optimization because Neigvals is not considered a static argument.

I don't know what the best solution is. For now, we can manually fix k to be a number << N_zeta and throw error of someone sets a large Neigvals in the BallooningStability objective.

desc/compute/_stability.py
rahulgaur104: Is there an existing compute function that has two different keys in data and returns both? We may need to use both eigenfunctions and eigenvalues in an objective in the future.

f0uriest: This is the one I was thinking of: https://github.com/PlasmaControl/DESC/blob/master/desc/compute/_geometry.py#L598-L599 

I think there are others.

In the case above we added a second dummy compute fn for the 2nd qty, I'm not sure if that's really necessary. I vaguely recall @ddudt did some stuff to make this cleaner?

desc/compute/_stability.py
f0uriest: was the static `Neigvals` still trying to pass it as part of `constants`? (see other comment). If it's an attribute of the objective it should actually be static (and we can ensure it is with the changes in #1662 ).

That would probably be the ideal, where you can pass `Neigvals` to the compute qty. Then the default can be to return everything (ie if you just to `eq.compute(...)` for analysis) but when optimizing the objective will call it with `Neigvals=small` to reduce memory usage.

desc/objectives/_stability.py
```diff
@@ -373,6 +373,9 @@ class BallooningStability(_Objective):
     zeta0 : array-like
         Points of vanishing integrated local shear to scan over.
         Default 15 points in [-π/2,π/2]
+    Neigvals : int
+        Number of top eigenvalues to select.
```
dpanici, line 377: define top better: this is the top sorted eigvals? so if I had eigvales [-1, 2 -2] and `Neigvals=2` it chooses the `2, -1` values?

Maybe just give a short example like 
```suggestion
        Number of top eigenvalues to select i.e. if `Neigvals=2` and the eigenvalues are `[-1, 1, -0.5]`, the eigenvalues considered in this objective will be `[1, -0.5]`
```

desc/compute/_stability.py
rahulgaur104: This does not work for fieldline quantities. I get data dependency error when I ask for the eigenfunction.

rahulgaur104: My bad. I was fudging up!

desc/objectives/_stability.py
rahulgaur104: Done.

desc/compute/_stability.py
rahulgaur104: Not concatenating anymore. Eigenvalues and eigenfunctions providede separately.,

desc/compute/_stability.py
rahulgaur104: Done!

desc/compute/_stability.py
rahulgaur104: It's working now. You can directly pass Neigvals as a kwarg to the compute function to request the number of largest eigenvalues and their eigenfunctions separately. Same with objective. By default it will only give the largest eigenvalue not the whole spectrum. We want to minimize the memory usage as much as possible.

desc/objectives/_stability.py
rahulgaur104: Maybe create an issue or warning of some kind?

desc/compute/_stability.py
```diff
@@ -398,12 +402,46 @@ def _ideal_ballooning_gamma2(params, transforms, profiles, data, **kwargs):
...
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["ideal ballooning lambda"],
+    parameterization=["desc.equilibrium.equilibrium.Equilibrium"],
+    source_grid_requirement={"coordinates": "raz", "is_meshgrid": True},
+)
+def _ideal_ballooning_eigenfunction(params, transforms, profiles, data, **kwargs):
     return data
```
YigitElma: this compute function is weird. You have a dependency for `ideal ballooning lambda` but it is not used. This is why that test fails. Also, you don't set `data[""ideal ballooning eigenfunction"]` so the output data doesn't have this key. Maybe do
```python
data["ideal ballooning eigenfunction"] = data["ideal ballooning lambda"]
return data
```

rahulgaur104: But the eigenfunction and lambda are two different things! 
I am trying to do something similar to what was done[ here](https://github.com/PlasmaControl/DESC/blob/master/desc/compute/_geometry.py#L598-L599)

YigitElma: Oh I see, the above compute function registers 2 quantities at once. I don't know if we can do that. There was a discussion before but I don't remember the final verdict.

rahulgaur104: I think we can do that for most usual cases but not special cases like fieldline quantities, bounce-average etc.
Rory suggested to do this experiment in this simple PR.

rahulgaur104: Btw, only this test fails. If I calculate ideal ballooning eigenfuntion or ideal ballooning lambda, it works perfectly.
Also, the optimization runs normally so it seems like a pointless test.

unalmis:
```suggestion
    return data  # noqa: unused dependency
```
@rahulgaur104 This will fix the failing test.

rahulgaur104: Yes, exactly what I was looking for!

rahulgaur104: Thanks!

rahulgaur104: Thanks!

desc/compute/_stability.py
```diff
@@ -398,15 +402,49 @@ def _ideal_ballooning_gamma2(params, transforms, profiles, data, **kwargs):
...
+    # Find the top_k eigenvalues.
+    top_eigvals, top_flat_indices = jax.lax.top_k(flat_w, k=Neigvals)
 
-    data["ideal ballooning lambda"] = gamma.flatten()
+    # convert back to original shape
+    alpha_idx, zeta0_idx, theta_idx = jnp.unravel_index(top_flat_indices, w.shape)
+
+    # v becomes less than the machine precision at some theta points which gives NaNs
+    # stop_gradient prevents that. Not sure how it will affect an objective that
+    # requires both the eigenvalue and eigenfunction
```
unalmis, line 423: Would using https://docs.jax.dev/en/latest/_autosummary/jax.lax.linalg.eigh.html#jax.lax.linalg.eigh be simpler and solve this issue?

desc/compute/_stability.py
f0uriest: OK cool. My other question was whether we want `Neigvals` to be per-fieldline or global. I would think per fieldline, so that `Neigvals=1` is equivalent to the old behavior, and would likely also be better at preventing wack-a-mole type situations where you stabilize one fieldline but destabilize another.

rahulgaur104: Global is what matters for a stability optimization. People don't use per fieldline. They only do per fieldline analysis for linear gyrokinetic  flux tube simulations.

dpanici:
make sure compute returns eigvals as fxn of rho, alpha, zeta0 like in master

desc/compute/_stability.py
rahulgaur104: If you try it, this is what you'll get

│NotImplementedError: subset_by_index not supported on CPU and GPU

rahulgaur104: It's funny because there is no warning in their docs.

dpanici (approved):

unalmis (approved):

f0uriest (dismissed):

rahulgaur104:
The coverage suddenlydropped from 90+% to less than 60% for no apparent reason. Bizzare!

unalmis (commented):
would prefer if eigenfunction array shape matched numpy api

rahulgaur104:
> The coverage suddenlydropped from 90+% to less than 60% for no apparent reason. Bizzare!

Rerunning unit tests increased coverage.

> would prefer if eigenfunction array shape matched numpy api
What does that mean?

unalmis (approved):

desc/objectives/_stability.py
unalmis: yigit is handling the deprecation so marking as resolved

desc/compute/_stability.py
unalmis: > remind us "natural" shape of this is since its not the usual (nrho, ntheta, nzeta)

done over at `ku/ballooning_parallel` branch

PR #1755, opened 2025-06-04
Use svd for first iteration of the optimisation

YigitElma:
Use SVD for the first iteration of the optimisation since it requires 1 decomposition and the root finding is faster, QR is better for later steps where root finding converges in a couple of steps.

See the [docs](https://desc-docs.readthedocs.io/en/latest/dev_guide/notebooks/getting-started-eq-solve.html#Solving-Trust-Region-Problem-with-QR) for the math, but in a nutshell, trust region subproblem when solved by SVD requires only 1 matrix decomposition. Then the root finding in alpha uses USV. However, with QR method, initial Q and R matrices cannot be used multiple times. Even though QR is significantly faster than SVD, this difference causes the first iteration to be slower in practice (due to multiple big linear system solves) since the initial values of alpha and trust radius are far from optimal.

This change attempts to combine the best of both worlds (hopefully). I created the PR mainly for the benchmarks.


unalmis (approved):

dpanici:
change initial trust radius to a different heuristic

PR #1756, opened 2025-06-05
Add fix to jax cublas issue to docs

YigitElma:


unalmis (approved):

f0uriest (approved):

docs/installation.rst
```diff
@@ -138,6 +138,15 @@ On computing clusters you must ensure to `module load anaconda` in order to use
         .. attention::
             Note that DESC does not always test on or guarantee support of the latest version of JAX (which does not have a stable 1.0 release yet), and thus older versions of GPU-accelerated versions of JAX may need to be installed, which may in turn require lower versions of JaxLib, as well as CUDA and CuDNN.
 
+        .. attention::
+            JAX version 0.6.1 may cause silent installation failures on GPU where the installation appears to succeed, but when running DESC, you will get an error like ``XlaRuntimeError: INTERNAL: cuSolver internal error``. To solve this problem, you need to run
```
ddudt, line 142: Is it just version 0.6.1? I think it might be 0.6.0+ but I'm not certain.

ddudt (commented):
This is really a JAX installation issue, not a DESC installation concern. We are simply reiterating something they already have already documented. I would rather add a general comment like "If you are having errors related to JAX, please look at their documentation and search their GitHub Issues for help." (and link to their docs). Then we don't have to keep updating our documentation to copy theirs.

docs/installation.rst
YigitElma: For some reason, 0.6.0 installs the correct version of cublas, but 0.6.1 install an older version...

PR #1762, opened 2025-06-06
Correct the description of the coordinate triplet

unalmis:
The quantity `x` does not transform as a vector unless the basis is cartesian. Since it is not an element of a position vector space it should not be named a position vector.


desc/compute/geom_utils.py
```diff
@@ -94,19 +94,19 @@ def rpz2xyz(pts):
 
 
 def xyz2rpz_vec(vec, x=None, y=None, phi=None):
-    """Transform vectors from cartesian (X,Y,Z) to polar (R,phi,Z) form.
+    """Transform from cartesian (X,Y,Z) to polar (R,phi,Z) form.
```
dpanici: This was still correct, no need to change. same for rpz2xyz_vec

desc/compute/_core.py
```diff
@@ -55,10 +55,11 @@ def _1(params, transforms, profiles, data, **kwargs):
 
 @register_compute_fun(
     name="x",
-    label="\\mathbf{r}",
+    label="\\mathbf{x}",
     units="m",
     units_long="meters",
```
unalmis: This is not meters either.

desc/compute/geom_utils.py
unalmis: This is more precise.

dpanici: I think it loses the information that it is meant to transform vectors specifically

desc/compute/_core.py
```diff
@@ -55,10 +55,11 @@ def _1(params, transforms, profiles, data, **kwargs):
 
 @register_compute_fun(
     name="x",
-    label="\\mathbf{r}",
-    units="m",
-    units_long="meters",
-    description="Position vector",
+    label="\\mathbf{x}",
+    units="~",
```
f0uriest, line 59: I'd prefer to keep this the way it was even though its technically not correct for cylindrical coordinates. I think it's still more correct to say it has units of meters than to say it doesn't have units.

desc/compute/geom_utils.py
f0uriest: yeah I agree with dario. The way it's worded now sounds like it would be valid to use this to transform `x` but that's not the case. This is for elements of a vector field defined locally at each point.

desc/compute/_core.py
unalmis: I am not saying it is dimensionless; I am saying that this quantity has no dimensional meaning.

desc/compute/geom_utils.py
unalmis:
```suggestion
    """Transform vectors from Cartesian (X,Y,Z) to polar (R,phi,Z) form.
```

desc/compute/geom_utils.py
```diff
@@ -129,19 +129,19 @@ def inner(vec, phi):
 
 
 def rpz2xyz_vec(vec, x=None, y=None, phi=None):
-    """Transform vectors from polar (R,phi,Z) to cartesian (X,Y,Z) form.
+    """Transform coordinates from polar (R,phi,Z) to cartesian (X,Y,Z) form.
```
unalmis:
```suggestion
    """Transform vectors from polar (R,phi,Z) to Cartesian (X,Y,Z) form.
```

desc/compute/_core.py
```diff
@@ -55,10 +55,11 @@ def _1(params, transforms, profiles, data, **kwargs):
...
+    units_long="not applicable",
```
YigitElma, line 60: Maybe, `meters, radians, meters` for the long version?

desc/compute/_curve.py
```diff
@@ -183,9 +183,10 @@ def _center_PlanarCurve(params, transforms, profiles, data, **kwargs):
 @register_compute_fun(
     name="x",
     label="\\mathbf{x}",
-    units="m",
-    units_long="meters",
-    description="Position vector along curve",
+    units="~",
+    units_long="not applicable",
```
YigitElma, line 187: see above

desc/compute/geom_utils.py
```diff
@@ -94,19 +94,19 @@ def rpz2xyz(pts):
...
 
 def xyz2rpz_vec(vec, x=None, y=None, phi=None):
-    """Transform vectors from cartesian (X,Y,Z) to polar (R,phi,Z) form.
+    """Transform from cartesian (X,Y,Z) to polar (R,phi,Z) form.
 
     Parameters
     ----------
     vec : ndarray, shape(...,3)
-        vectors, in cartesian (X,Y,Z) form
+        coordinate triplet, in cartesian (X,Y,Z) form
```
YigitElma: I think this was true, in cartesian `coordinate triplet`=`position vector`. I would prefer the old version to emphasize that this is a vector not a coordinate like `x, y, phi` below

desc/compute/geom_utils.py
```diff
@@ -94,19 +94,19 @@ def rpz2xyz(pts):
...
     x, y, phi : ndarray, shape(...,)
-        anchor points for vectors. Either x and y, or phi must be supplied
+        anchor points for coordinates. Either x and y, or phi must be supplied
 
     Returns
     -------
     vec : ndarray, shape(...,3)
-        vectors, in polar (R,phi,Z) form
+        coordinates, in polar (R,phi,Z) form
```
YigitElma: This returns vectors that are anchored at given points, I think the old version is more accurate

desc/compute/geom_utils.py
```diff
@@ -129,19 +129,19 @@ def inner(vec, phi):
...
 
 def rpz2xyz_vec(vec, x=None, y=None, phi=None):
-    """Transform vectors from polar (R,phi,Z) to cartesian (X,Y,Z) form.
+    """Transform coordinates from polar (R,phi,Z) to cartesian (X,Y,Z) form.
 
     Parameters
     ----------
     vec : ndarray, shape(n,3)
-        vectors, in polar (R,phi,Z) form
+        coordinates, in polar (R,phi,Z) form
```
YigitElma: same as above. This takes both vectors and the coordinates that the vector is anchored. I prefer the old version

YigitElma (changes requested):

desc/compute/_core.py
unalmis: doesn't work when basis=xyz

YigitElma (dismissed):

desc/compute/_curve.py
```diff
@@ -183,9 +183,10 @@ def _center_PlanarCurve(params, transforms, profiles, data, **kwargs):
...
+    description="Coordinate triplet. "
```
f0uriest, line 188: Might be helpful to at least say that this is in real physical space rather than flux coordinates

unalmis: then the description would no longer be basis agnostic

f0uriest: How so? It's always in physical space, regardless of it being cartesian or polar.

desc/compute/_core.py
f0uriest: I think most physicists would still refer to this as a position vector with units of meters regardless of the coordinate system used. In general "vector" means slightly different things to mathematicians vs physicists. 

I think we should still say somewhere that this is in meters when basis="xyz" or something.

desc/compute/_curve.py
unalmis: If I request basis = e_rho, e_theta, e_zeta, then the coordinates x = [x_1, x_2, x_3] are flux coordinates for position = x_1 e_rho + x_2 e_theta + x_3 e_zeta

YigitElma: Can we request this `basis = e_rho, e_theta, e_zeta`? I thougt options were only `rpz` or `xyz`

desc/compute/_core.py
```diff
@@ -55,10 +55,11 @@ def _1(params, transforms, profiles, data, **kwargs):
...
+    description="Coordinate triplet. "
+    "This is not a position vector unless basis is cartesian.",
```
unalmis:
```suggestion
    "This is not a position vector unless basis is cartesian. "
    "When basis is cartesian, the units are meters.",
```

desc/compute/_curve.py
```diff
@@ -183,9 +183,10 @@ def _center_PlanarCurve(params, transforms, profiles, data, **kwargs):
...
+    "This is not a position vector unless basis is cartesian.",
```
unalmis:
```suggestion
    "This is not a position vector unless basis is cartesian. "
    "When basis is cartesian, the units are meters.",
```

desc/compute/_curve.py
```diff
@@ -382,9 +383,10 @@ def _x_sss_FourierPlanarCurve(params, transforms, profiles, data, **kwargs):
...
```
unalmis:
```suggestion
    "This is not a position vector unless basis is cartesian. "
    "When basis is cartesian, the units are meters.",
```

desc/compute/_curve.py
```diff
@@ -586,9 +588,10 @@ def _center_FourierRZCurve(params, transforms, profiles, data, **kwargs):
...
```
unalmis:
```suggestion
    "This is not a position vector unless basis is cartesian. "
    "When basis is cartesian, the units are meters.",
```

desc/compute/_curve.py
```diff
@@ -736,9 +739,10 @@ def _center_FourierXYZCurve(params, transforms, profiles, data, **kwargs):
...
```
unalmis:
```suggestion
    "This is not a position vector unless basis is cartesian. "
    "When basis is cartesian, the units are meters.",
```

desc/compute/_curve.py
```diff
@@ -863,9 +867,10 @@ def _center_SplineXYZCurve(params, transforms, profiles, data, **kwargs):
...
```
unalmis:
```suggestion
    "This is not a position vector unless basis is cartesian. "
    "When basis is cartesian, the units are meters.",
```

desc/compute/_core.py
unalmis: > I think most physicists would still refer to this as a position vector with units of meters

People will learn to fix their mistakes.

> I think we should still say somewhere that this is in meters when basis="xyz" or something.

Done.

PR #1763, opened 2025-06-09
Update ballooning stability

### LINKED ISSUE #1797: Parallelize ballooning
Parallelize ballooning stability metrics

unalmis:
- [x] Resolves #1797 
- [x] Bypass building quadrature transforms in objective with #1094 .
- [x] Diagonalize tensor products to reduce computation
- [x] Fix issue where compute functions are recompiled each time they are called.
- [x] #1782 
- [x] adds ability to hide quantities from public list of documentation as requested by DESC developers

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ netcdf4 >= 1.5.4, <= 1.7.2
 numpy >= 1.20.0, <= 2.2.6
 nvgpu <= 0.10.0
 orthax <= 0.2.4
-plotly >= 5.16, <= 6.1.2
+plotly >= 5.16, <= 6.2.0
 psutil <= 7.0.0
 pylatexenc >= 2.0, <= 2.10
 quadax >= 0.2.2, <= 0.2.8


desc/compute/_stability.py
```diff
@@ -309,115 +312,111 @@ def _ideal_ballooning_lambda(params, transforms, profiles, data, **kwargs):
...
-
-    top_eigfuns = jnp.take_along_axis(v_T, top_theta_idxs[..., None], axis=2)
-
-    data["ideal ballooning lambda"] = top_eigvals
-    data["ideal ballooning eigenfunction"] = top_eigfuns
+    if eigfuns:
+        B_inv = jnp.reciprocal(jnp.sqrt(f[..., 1:-1]))
+        w, v = jnp.linalg.eigh(B_inv[..., jnp.newaxis] * A * B_inv[..., jnp.newaxis, :])
+    else:
+        # We may seek eigenvalues of B⁻¹ @ A because for diagonal matrices,
```
f0uriest: don't you still need the other factor of Binv to make the matrix symmetric (since eigvalsh only accesses one half of the matrix)

unalmis: Yes I agree

unalmis:
would prefer if eigenfunction array shape matched numpy api

unalmis:
All tests pass.

desc/compute/_stability.py
```diff
@@ -234,199 +235,210 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
...
+    def reshape(f):
+        assert f.shape == (num_zeta0, grid.num_nodes)
+        f = jnp.swapaxes(grid.meshgrid_reshape(f.T, "raz"), -1, -2)
+        assert f.shape == (grid.num_rho, grid.num_alpha, num_zeta0, grid.num_zeta)
+        return f
+
+    c = reshape(data["c balloon"])
+    f = reshape(data["f balloon"])
+    g = reshape(data["g balloon"])
+    # Approximate derivative along field line with second order finite differencing.
```
unalmis, line 392: we could do chain rule like i did for the bounce integral stuff to get spectrally accurate derivatives but its fine

desc/compute/_stability.py
```diff
@@ -234,199 +235,210 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
...
+    g = (g[..., 1:] + g[..., :-1]) / 2
+    diag_inner = c[..., 1:-1] - (g[..., 1:] + g[..., :-1]) / dz**2
+    diag_outer = g[..., 1:-1] / dz**2
+
+    j = np.arange(grid.num_zeta - 2)
```
unalmis: if someone thinks I should use `jnp` let me know why please

unalmis:
I still do not like at all that alpha is set to 2pi periodicity in the coordinate mapping. I mentioned this in the original pr that it worked fine for the tests in that pr but things could change in future in how we compute dependencies etc.

I suggest removing that.

desc/objectives/_stability.py
```diff
@@ -405,7 +405,7 @@ def __init__(
         normalize_target=True,
         loss_function=None,
         deriv_mode="auto",
-        rho=0.5,
+        rho=1,
```
rahulgaur104: Why? The instability usually peaks in the middle.

tests/test_examples.py
```diff
@@ -2377,57 +2324,39 @@ def test_ballooning_stability_opt():
...
+            FixBoundaryR(eq=eq, modes=modes_R),
+            FixBoundaryZ(eq=eq, modes=modes_Z),
+            FixPressure(eq=eq),
+            FixIota(eq=eq),
+            FixPsi(eq=eq),
+        ),
         ftol=1e-4,
         xtol=1e-6,
         gtol=1e-6,
-        maxiter=2,  # increase maxiter to 50 for a better result
```
rahulgaur104, line 2403: Why remove that comment?

desc/compute/_stability.py
```diff
@@ -234,199 +235,210 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="c balloon",
```
rahulgaur104: I would suggest c ballooning, same for g and f.

desc/compute/_stability.py
```diff
@@ -234,199 +235,210 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
...
+        assert f.shape == (num_zeta0, grid.num_nodes)
+        f = jnp.swapaxes(grid.meshgrid_reshape(f.T, "raz"), -1, -2)
+        assert f.shape == (grid.num_rho, grid.num_alpha, num_zeta0, grid.num_zeta)
+        return f
+
+    c = reshape(data["c balloon"])
+    f = reshape(data["f balloon"])
+    g = reshape(data["g balloon"])
+    # Approximate derivative along field line with second order finite differencing.
+    g = (g[..., 1:] + g[..., :-1]) / 2
```
rahulgaur104: I think it's important to preserve names even if it requires a new variable and takes a bit more memory. g_half is the value of g on half/middle points instead of the zeta point. This was done for numerical stability and the variable name was chosen accordingly. 

Please change it back to g_half.

desc/compute/_stability.py
```diff
@@ -437,245 +449,118 @@ def _ideal_ballooning_lambda(params, transforms, profiles, data, **kwargs):
...
-        "B^zeta",
-        "p_r",
-        "iota",
-        "shear",
-        "psi",
-        "psi_r",
-        "rho",
-    ],
+    coordinates="r",
+    data=["c balloon", "g balloon"],
```
rahulgaur104: Same comment as above.

desc/compute/_stability.py
```diff
@@ -437,245 +449,118 @@ def _ideal_ballooning_lambda(params, transforms, profiles, data, **kwargs):
...
+
+    def reshape(f):
+        assert f.shape == (num_zeta0, grid.num_nodes)
+        f = jnp.moveaxis(grid.meshgrid_reshape(f.T, "raz"), -2, 0)
+        assert f.shape == (grid.num_zeta, grid.num_rho, grid.num_alpha, num_zeta0)
+        return f
+
+    c = reshape(data["c balloon"])[:-1]
+    g = reshape(data["g balloon"])
+    g = (g[1:] + g[:-1]) / 2
```
rahulgaur104: Same as above.

desc/compute/_stability.py
unalmis: It is a very memory intensive computation. Doing the computation on the 5 surfaces in parallel in `test_compare_with_cobraVMEC` crashed the CI and went out of memory on my computer which has 40 GB of memory. I have added a comment that explain what the variable is with more verbosity.

unalmis: these are the types micro optimizations that the other teams do at the hackathons, that we don't usually do, but for memory intensive stuff why not

desc/compute/_stability.py
```diff
@@ -234,199 +236,191 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
...
+            2 * data["rho"] ** 2 * data["cvdrift"]
+            - data["shear"] * data["cvdrift0"] * zeta0
+        )
+    )
+    return data
+
+
+@register_compute_fun(
+    name="f ballooning",
+    # f = a  Bₙ³ / b⋅∇ζ (dψ_N/dρ)² |∇α|² / B³
```
unalmis: Since `gds2` is not the same as (dψ_N/dρ)² |∇α|², we should update this comment. What is it supposed to be?

CHANGELOG.md
```diff
@@ -8,11 +8,13 @@ New Features
 - Adds ``PlasmaCoilSetDistanceBound`` objective to allow a simultaneous constraint on minimum and maximum distance between the plasma and coils.
 - Adds ``grid.meshgrid_flatten`` for flattening 3d data to a 1d array in the correct order.
 - Ability to obtain the top eigenvalues and the corresponding eigenfunctions from the ``ideal ballooning lambda`` compute function by specifying the variable ``Neigvals``.
+- Parallelized ideal ballooning stability and Newcomb ballooning metrics and [other improvements](https://github.com/PlasmaControl/DESC/pull/1763).
 
 Bug Fixes
 
 - Fixes issues with the ``desc.geometry.curve.FourierPlanarCurve`` that arose when the curve normal was parallel or anti-parallel to the +Z axis.
 - Fixes an issue in ``desc.objectives.BoundaryError`` where an equilibrium with kinetic profiles could result in undefined Jacobian values.
+- Fixes bugs in ideal ballooning stability and Newcomb ballooning metrics.
```
f0uriest: Can you be more specific here. What bugs exactly? Are people's old calculations incorrect?

desc/backend.py
```diff
@@ -134,8 +147,62 @@ def wrapper(*args, **kwargs):
...
-    # JAX implementation is not differentiable on gpu.
-    eigh_tridiagonal = execute_on_cpu(jax.scipy.linalg.eigh_tridiagonal)
+    _eigh_tridiagonal = jnp.vectorize(
+        jax.scipy.linalg.eigh_tridiagonal,
+        signature="(m),(n)->(m)",
+        excluded={"eigvals_only", "select", "select_range", "tol"},
+    )
+    if desc_config["kind"] == "gpu":
+        # JAX eigh_tridiagonal is not differentiable on gpu.
+        # https://github.com/jax-ml/jax/issues/23650
```
f0uriest, line 157: that issue is still waiting on a MRE, is this still an issue? and in what cases?

desc/backend.py
```diff
@@ -134,8 +147,62 @@ def wrapper(*args, **kwargs):
...
         return wrapper
 
-    # JAX implementation is not differentiable on gpu.
-    eigh_tridiagonal = execute_on_cpu(jax.scipy.linalg.eigh_tridiagonal)
+    _eigh_tridiagonal = jnp.vectorize(
+        jax.scipy.linalg.eigh_tridiagonal,
+        signature="(m),(n)->(m)",
+        excluded={"eigvals_only", "select", "select_range", "tol"},
+    )
+    if desc_config["kind"] == "gpu":
```
f0uriest, line 155: using `jax.lax.platform_dependent` might be a better way to go here, especially once #1495 is working

desc/objectives/_stability.py
```diff
@@ -407,12 +405,12 @@ def __init__(
...
         loss_function=None,
         deriv_mode="auto",
-        rho=0.5,
+        rho=np.array([0.5]),
         alpha=None,
         nturns=3,
         nzetaperturn=200,
-        zeta0=None,
-        Neigvals=None,
+        zeta0=jnp.linspace(-0.5 * np.pi, 0.5 * np.pi, 15),
```
f0uriest: note that making arrays default args sometimes messes with autodoc stuff.

FYI `BallooningStability` is also missing from the docs

desc/objectives/_stability.py
```diff
@@ -445,176 +451,118 @@ def __init__(
...
         """
-        eq = self.things[0]
+        self._iota_keys = ["iota", "iota_r", "shear", "a"]
 
-        # we need a uniform grid to get correct surface averages for iota
+        eq = self.things[0]
         iota_grid = LinearGrid(
-            rho=self._rho,
+            # to compute length scale quantities correctly
+            rho=np.append(self._rho, 1) if self._add_lcfs else self._rho,
```
f0uriest, line 471: is it really that much more expensive to create a dedicated grid on the LCFS to compute length scales?

desc/objectives/_stability.py
```diff
@@ -422,15 +420,23 @@ def __init__(
...
-        self._alpha = alpha
         self._nturns = nturns
         self._nzetaperturn = nzetaperturn
-        self._zeta0 = zeta0
         self._Neigvals = Neigvals
         self._lambda0 = lambda0
         self._w0 = w0
         self._w1 = w1
+        self._rho = rho
+        self._add_lcfs = np.all(self._rho < 0.97)
```
f0uriest, line 431: why 0.97?

desc/objectives/_stability.py
```diff
@@ -422,15 +420,23 @@ def __init__(
...
-        self._rho = rho
-        self._alpha = alpha
         self._nturns = nturns
         self._nzetaperturn = nzetaperturn
-        self._zeta0 = zeta0
         self._Neigvals = Neigvals
         self._lambda0 = lambda0
         self._w0 = w0
         self._w1 = w1
+        self._rho = rho
```
f0uriest: does rho always need to be a 1d array? if so should add `atleast_1d`

desc/compute/_stability.py
```diff
@@ -234,199 +236,191 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
     return data
 
 
+@register_compute_fun(
+    name="c ballooning",
```
f0uriest, line 276: do we want these as part of the public list of variables? Did we ever add the option to register a function but leave it private?

desc/compute/_stability.py
```diff
@@ -234,199 +236,191 @@ def _magnetic_well(params, transforms, profiles, data, **kwargs):
...
-        "psi_r",
-        "rho",
-    ],
+    data=["c ballooning", "f ballooning", "g ballooning"],
     source_grid_requirement={"coordinates": "raz", "is_meshgrid": True},
-    zeta0="array: points of vanishing integrated local shear to scan over. "
-    "Default 15 points linearly spaced in [-π/2,π/2]",
     Neigvals="int: number of largest eigenvalues to return, default value is 1.`"
     "If `Neigvals=2` eigenvalues are `[-1, 0, 1]` we get `[1, 0]`",
+    eigfuns="bool: Whether to return eigenfunctions. Default is true.",
```
f0uriest: I think this probably isn't the best way to handle this, it should be its own compute function

desc/compute/_stability.py
```diff
@@ -437,245 +431,119 @@ def _ideal_ballooning_lambda(params, transforms, profiles, data, **kwargs):
...
+
+    """
+    assert kwargs.get("eigfuns", True)
     return data  # noqa: unused dependency
 
 
 @register_compute_fun(
     name="Newcomb ballooning metric",
-    label="\\mathrm{Newcomb-ballooning-metric}",
+    label="Newcomb ballooning metric",
```
f0uriest: i think we still need the mathrm otherwise this will get formatted weird

tests/test_stability_funs.py
```diff
@@ -429,11 +429,7 @@ def test_ballooning_geometry(tmpdir_factory):
             "g^rz",
         ]
 
-        grid = Grid.create_meshgrid(
-            [rho, alpha, zeta],
-            coordinates="raz",
-            period=(np.inf, 2 * np.pi, np.inf),
```
f0uriest, line 435: why get rid of `period` on all of these calls?

f0uriest (changes requested):

desc/compute/_stability.py
f0uriest: Does this actually save time/memory? The compiler should be able to elide that assignment, and if it can I'd prefer the more verbose but clearer version.

desc/objectives/_stability.py
unalmis: I don't understand the question because that is what we are doing here

desc/objectives/_stability.py
unalmis: The extrapolation should work fine when it is that close

tests/test_stability_funs.py
unalmis: The mapping from flux coords to field line coordinates is secular. If you enforce periodicity the mapping is in general wrong. See #1180. For this application, I remember I showed/convinced myself that it doesn't matter before we merged the original pull request for ballooning stability. Still I don't like it, because its basically a foot-gun waiting to happen and I have yet to see it converge any faster in the root finding.

unalmis: Plus it is less stuff for the user's to type and less stuff they need to learn to get their application working.

CHANGELOG.md
unalmis: the bug is #1772. the computation was not being done along field lines. yes.

desc/backend.py
unalmis: as of today i think forward mode is differentiable on cpu, and reverse mode is not differenitable on either.

desc/backend.py
unalmis: looks like it does not support kwargs

desc/compute/_stability.py
unalmis: We would need to initiate a deprecation cycle. People are already using `ideal ballooning lambda` to compute eigenfunctions and I don't want to make rahul email them again for a change.

currently this pr does not interrupt their workflow.

desc/backend.py
unalmis: ballooning notebook doesn't work for example with eigh_tridiagonal.

desc/compute/_stability.py
rahulgaur104: psi_boundary =  a_N**2 * B_N / 2 and rho = sqrt(psi/psi_boundary).

Also,  psi_N = psi/(a_N^2 * B_N)  = psi/psi_boundary * 1/2 = rho**2/2.
so dpsi_N/drho = rho

gds2 = (dψ_N/dρ)² |∇α|² = rho ^ 2 (g^aa - 2 * zeta_0 * (shear/rho)*g^ra + zeta_0^2 * (shear/rho)^2 * g^rr)  
and 
f ballooning = a_N B_N^3 /b dot grad zeta * (dpsi/drho)^2 |grad alpha|^2 /B^3 = a_N  B_N^3 / b dot grad zeta * gds2 /B^3 
so I don't understand your question. 

gds2 already includes the rho ** 2 coming from (dpsi_N/drho)^2.

Also, since all the coefficients c and g, f are now calculated separately does this mean the equilibrium quantities used by these coefficients are also calculated multiple times?

tests/test_stability_funs.py
rahulgaur104: Removing is fine with me but this is all 1D root finding and not 2D interpolation in alpha, zeta so it seems our results won't change. In practice, we always keep alpha in [0, pi]

desc/compute/_stability.py
rahulgaur104: Yeah, someone may want to plot the ballooning potential which is a combo of these terms.
 I am more concerned about potential double computation of equilibrium quantities when we call c ballooning and f ballooning

desc/objectives/_stability.py
rahulgaur104: I guess this extrapolation is to calculate normalizing factors a_N and B_N. The solver itself is fine for all rhos <= 1.0

desc/backend.py
rahulgaur104: Is there a case where we are doing eigendecomp on a CPU? We should avoid that, even if it means always calculating both eigenvalues and eigenfunctions.

unalmis: Right now everything is hardcoded to call `jax.scipy.linalg.eigh`. I assume that uses gpu when available. If there is not gpu available, I wanted to use eigh_tridiagonal, but we still can't because it doesn't work with reverse mode anymore.

desc/compute/_stability.py
unalmis: that sounds like we want it public. can remove the `public=False` in the compute function register before this PR is merged

desc/backend.py
rahulgaur104: jax.scipy.linalg.eigh calls jax.lax.linalg.eigh which is calculating both eigenvalues and eigenvectors so I guess we are saving some memory by not storing the eigenvector array when we just use compute? So the option of having eigenvectors is only saving memory a few GBs when we compute the eigvals on multiple surfaces.

When we do optimization, I don't think we are saving anything because eigenvectors are needed to calculate the gradient of the eigenvalue.
 
In terms of performance it should be exactly as fast as the previous version.

unalmis: eigenvector matrix is big so if we can save gigabytes with an if statement it's worth it

desc/compute/_stability.py
unalmis: then we'd have to recompute everything which isn't practical

desc/compute/_stability.py
unalmis: the dependencies are only computed once

desc/compute/_stability.py
unalmis: These comments were from before I wrapped the function in JIT.

desc/compute/_stability.py
unalmis: I don't understand the `gds2` equality. `gds2` depends on my choice of `zeta0`, while |∇α|² does not. Below test doesn't pass, so naiively there must be something missing from the statement that gds2 = (dψ_N/dρ)² |∇α|² .

```python
@pytest.mark.unit
def test_gds2():
    """Test computation of  (dψ_N/dρ)² |∇α|²."""
    zeta0 = np.linspace(0, 2 * np.pi, 15)[:, np.newaxis]
    data = eq.compute(["rho", "g^aa", "iota", "shear", "g^ra", "g^rr"], zeta0=zeta0)
    gds2 = (
        data["rho"] ** 2 * data["g^aa"]
        - 2
        * data["rho"]
        * jnp.sign(data["iota"])
        * data["shear"]
        * data["g^ra"]
        * zeta0
        + data["shear"] ** 2 * data["g^rr"] * zeta0**2
    )
    np.testing.assert_allclose(gds2, data["g^aa"] * data["rho"] ** 2)
```

rahulgaur104: alpha = theta - iota * (zeta-zeta0). Then it should work out. Also, once we are done with the math, we rename iota zeta0 to  zeta0.

desc/compute/_stability.py
rahulgaur104: So, the answer's no?

desc/compute/_stability.py
rahulgaur104: So full grad alpha = grad alpha (as defined in DESC) - iota' zeta_0 grad rho.

desc/backend.py
rahulgaur104: Yes, but that saving can be done inside the compute function as well. What if you do 
if eigfuns:
    w, _ = eigh
else:
    w, v = eigh
    
but use the same eigenvalue solver (with full eigenvectors and eigenvalue). 
Anyway, I don't have a problem with how it's done but generally speaking, the more changes you make the more you will be questioned.

desc/compute/_stability.py
rahulgaur104: In your test you have ignored the zeta0 contribution which comes from the modified definition of alpha we use for the ballooning solver (also given in the paper)

desc/compute/_stability.py
unalmis: i didn't check because I believe that jit would elide this. there was another pr with this issue but that code wasn't jitted. it is g_half now

desc/compute/_stability.py
unalmis: I added this test; it does not pass. It passes for the value of zeta0=0, but not the others.
```python
@pytest.mark.unit
@pytest.mark.xfail(reason="trying to understand")
def test_gds2():
    """Test computation of (dψ_N/dρ)² |∇α|²."""
    data = eq.compute(
        ["rho", "g^aa", "iota", "shear", "g^ra", "g^rr", "alpha_r (secular)", "iota_r"]
    )
    zeta0 = np.linspace(0, 2 * np.pi, 15)[:, np.newaxis]
    gds2 = (
        data["rho"] ** 2 * data["g^aa"]
        - 2
        * data["rho"]
        * jnp.sign(data["iota"])
        * data["shear"]
        * data["g^ra"]
        * zeta0
        + data["shear"] ** 2 * data["g^rr"] * zeta0**2
    )
    data = {"alpha_r (secular)": data["alpha_r (secular)"] + data["iota_r"] * zeta0}
    data = eq.compute(["g^aa", "rho"], data=data)
    np.testing.assert_allclose(gds2, data["g^aa"] * data["rho"] ** 2)
```

```
>       np.testing.assert_allclose(gds2, data["g^aa"] * data["rho"] ** 2)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 137128 / 147015 (93.3%)
E       Max absolute difference among violations: 7.4268526
E       Max relative difference among violations: 0.42971927
E        ACTUAL: array([[ 1.103992,  1.607312,  3.045009, ...,  7.785284,  6.383699,
E                5.476742],
E              [ 1.103992,  1.607595,  3.045536, ...,  5.350778,  4.053961,...
E        DESIRED: array([[ 1.103992,  1.607312,  3.045009, ...,  7.785284,  6.383699,
E                5.476742],
E              [ 1.103992,  1.607555,  3.04546 , ...,  5.44027 ,  4.137223,...

```

unalmis: The test passes when `gds2` is redefined in #1782

rahulgaur104: It should fail because alpha in ballonog is redefined to have other contributions at non-zero zeta0 so that PR  is unnecessary. gds2 is also rarely ever used as a separate compute function.
Also, the sign of iota needs to be accounted for because that can potentially change.

rahulgaur104: Also, range of zeta0 is always chosen so that the iota zeta0 term goes from -pi/2, pi/2 and must be smaller than the size of the domain zeta.

rahulgaur104: So it won't matter if I do iota zeta_0 or -iota zeta_0

unalmis: That new PR defines g as in eq. 3.4 of your adjoint ballooning article with  g = (dψ_N/dρ)² |∇α|² where alpha = theta - iota * (zeta-zeta0) and zeta = phi.

If you are saying that the definition of `g` on master and this PR is correct then the g here is such that g $\neq$ (dψ_N/dρ)² |∇α|². So we should fix this comment. Am I missing something?

unalmis: > It should fail because alpha in ballonog is redefined to have other contributions at non-zero zeta0 so that PR is unnecessary.
Also, the sign of iota needs to be accounted for because that can potentially change.

What do these mean; I don't know what the additional contributions are. Is there something more than gds2 = (dψ_N/dρ)² |∇α|²?

rahulgaur104: > If you are saying that the definition of g on master and this PR is correct then the g here is such that g 
≠
 (dψ_N/dρ)² |∇α|². So we should fix this comment. Am I missing something?

Yes, the comment is wrong, not the code. Just make the comment
g = a³ Bₙ * b⋅∇ζ  (dψ_N/dρ)² |∇(α-iota zeta_0)|²/B and the code should then be consistent.

I have assumed alpha = theta - iota(zeta + zeta_0) (I know I said (zeta - zeta_0) earlier) but iota zeta_0 is chosen (-pi/2. pi/2 so it doesn't matter)

rahulgaur104: > What do these mean; I don't know what the additional contributions are. Is there something more than gds2 = (dψ_N/dρ)² |∇α|²?

That (dψ_N/dρ)² * data["g^aa"] in DESC is not equal to gds2 = (dψ_N/dρ)² * |∇(α-iota zeta_0)|² because of the extra iota zeta_0 term. 

Let's take small steps because the longer my response, the more confusing it can be. Is the above statement clear?

desc/compute/_stability.py
rahulgaur104: I like this option. Also, @f0uriest   the eigenfunction is still defined as a separate compute function so this is consistent with what you ask.

desc/objectives/_stability.py
f0uriest: this is adding a new surface to an existing grid which then requires some weird indexing, rather than just using a dedicated grid for the length qtys

desc/compute/_stability.py
f0uriest: doesn't this mean that if i do `eq.compute(["ballooning eigenvalues", "ballooning eigenfunctions"])` it will throw an error because I actually need to do `eq.compute(["ballooning eigenvalues", "ballooning eigenfunctions"], eigfuns=True)` which feels more confusing

unalmis: The default is `eigfuns=True`.

desc/compute/_stability.py
unalmis: Let's move the discussion to the other pull request

desc/objectives/_stability.py
unalmis: Oh. This is what I thought to do first. I did not benchmark

unalmis: But actually yes, making a new grid would require recomputing the eqilibrium quantitites three times total. Right now it is only done twice (once to get iota and a, and the second to get ballooning along field line)

f0uriest: but it wouldn't actually be any extra flops I think. Whether its 1 grid with extra surface or 2 grids, the total number of nodes is the same.

desc/compute/_stability.py
f0uriest: then what does this save? given that `eigh_tridiagonal` doesn't even really work for us, we end up just calling `eigh` or `eigvalsh` which under the hood always computes the eigenvectors.

If we ever do get `eigh_tridiagonal` working, we could then also add calculation of eigenvectors as a separate computation that uses the eigenvalues as inputs, eg https://github.com/tensorflow/tensorflow/blob/5bc9d26649cca274750ad3625bd93422617eed4b/tensorflow/python/ops/linalg/linalg_impl.py#L1463

unalmis: It does not store the eigenvectors in memory when `eigfuns=False`

desc/objectives/_stability.py
unalmis: evaluating transforms another time is a flop

unalmis: and that is different with two grids once partial summation is working

desc/compute/_stability.py
unalmis: Even though it does save memory we decided to remove it as the api may change later.

desc/objectives/_stability.py
unalmis: I'd like to keep this since changing it requires effort

docs/write_variables.py
```diff
@@ -35,6 +35,8 @@ def write_csv(parameterization):
         keys = datidx.keys()
         for key in sorted(keys):
             if key not in data_index[parameterization][key]["aliases"]:
+                if not data_index[parameterization][key]["public"]:
```
rahulgaur104, line 38: A comment explaining this would be nice.

unalmis: I documented here: https://github.com/PlasmaControl/DESC/blob/2fa75d5554bb27d9c2f76af8f74dc0bff306d5cf/desc/compute/data_index.py#L131

desc/backend.py
unalmis: (I can't make a wrapper for a function that does not support keyword arguments, when the original function requires keyword args).

desc/objectives/_stability.py
rahulgaur104: Kind attention

rahulgaur104 (approved):

PR #1769, opened 2025-06-10
Deprecate `constants`

### LINKED ISSUE #1394: Pass `constants` dict into `vjp` for augmented lagrangian
https://github.com/PlasmaControl/DESC/blob/95612fa625bd5ba939569004c767ed02594586f0/desc/optimize/_desc_wrappers.py#L90

YigitElma:
Historically, this was needed for static values with JAX, but with proper pytree registration, we don't need them anymore.

Also, `constants` cause multiple cache misses and recompilation, since we pass it in different ways in different parts of the code.

Resolves #1394


f0uriest (changes requested):
`constants` is part of the public facing API so I don't think we should remove it without going through a proper deprecation cycle where we start by warning that it will be removed but keep it functional etc.

desc/objectives/_bootstrap.py
```diff
@@ -251,25 +251,21 @@ def build(self, use_jit=True, verbose=1):
...
 
         Returns
         -------
         obj : ndarray
             Bootstrap current self-consistency residual on each rho grid point.
 
         """
-        if constants is None:
-            constants = self.constants
+        constants = self.constants
```
f0uriest: I'd recommend leaving it how it was, but throw a deprecation warning `if constants is not None`

f0uriest: and also change this to `constants = self._constants` (see other comment below)

desc/objectives/objective_funs.py
```diff
@@ -947,11 +890,6 @@ def compile(self, mode="auto", verbose=1):
             timer.disp("Total compilation time")
         self._compiled = True
 
-    @property
-    def constants(self):
-        """list: constant parameters for each sub-objective."""
-        return [obj.constants for obj in self.objectives]
```
f0uriest, line 1115: throw a deprecation warning here. Eventually we'll remove `obj`constants` but objectives can still have a private `_constants` attribute if they want

YigitElma:
> `constants` is part of the public facing API so I don't think we should remove it without going through a proper deprecation cycle where we start by warning that it will be removed but keep it functional etc.

Yes, this is a breaking change (I will update the changelog), but a very very easy one to solve. Users just need to remove constants from the compute method and instead define them in the build method. For optimization point of view, there is actually no change (if they do the previous little change) because we pass the constants to compute method, not the user. 

Another problem that they can face is not being able to compute the objective value with different set of constants. But this is a one time usage, and IMO very error prone, i.e. some quantities in the new constants array might not match what they should be, and we don't have a check for that. So, they can just initialize and build the objective with new constants.

My point is the fix for this breaking change is way easier than throwing deprecation warning and removing all of that extra code later. If people vote for that I can do the changes but I prefer to remove it and add emphasize the change in changelog and maybe create a capitalized issue (possibly fix it to the top)

f0uriest:
It's obviously easier for us to just make the change, but the point of a deprecation cycle is to make it easier for users to transition their code to the new API. And we may think it's an easy change for them, but for people who wrote custom objectives and optimizers it may more annoying, and as a policy I would strongly vote in favor of not making breaking changes without a proper deprecation cycle. It's not like this change is needed anytime soon for anything we have in the pipeline, so I don't see any reason to rush it.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ New Features
 - Adds the classes ``FourierXYCurve`` and ``FourierXYCoil`` to represent planar curves/coils with Fourier series for X and Y instead of the radius.
 - Removes default objective and constraints for ``desc.equilibrium.Equilibrium.optimize``, so now user is required to pass in the constraints and objective when using this method.
 - Adds ``PlasmaCoilSetDistanceBound`` objective to allow a simultaneous constraint on minimum and maximum distance between the plasma and coils.
+- `constants` argument is deprecated and will be removed in a future release. This argument was not necessary, and the code will still work if user doesn't pass it. Users should update their custom objectives for this change.
```
f0uriest: should say what it is an argument to

desc/objectives/_bootstrap.py
```diff
@@ -269,7 +266,16 @@ def compute(self, params, constants=None):
 
         """
         if constants is None:
-            constants = self.constants
+            constants = self._constants
```
f0uriest: a possibly cleaner option that avoids adding a bunch of uncovered lines:
```python
warnif(constants is not None, 
       DeprecationWarning,
       "blah blah blah"
)
constants = setdefault(constants, self._constants)
```

should also double check the warning type, I forget if its supposed to be a `DeprecationWarning` or `FutureWarning` or `PendingDeprecationWarning` etc, the main python docs has recommended best practice

desc/objectives/_fast_ion.py
```diff
@@ -189,6 +189,8 @@ def __init__(
             "num_pitch": num_pitch,
             "pitch_batch_size": pitch_batch_size,
             "surf_batch_size": surf_batch_size,
+            "alpha": alpha,
+            "quad_weights": 1.0,
```
f0uriest: why put them here and then pop to constants? why not just put them in constants to begin with?

desc/objectives/_fast_ion.py
```diff
@@ -305,29 +317,37 @@ def _build_spline(self, use_jit=True, verbose=1):
...
+
+        self._constants = {
+            "rho": self._grid.compress(self._grid.nodes[:, 0]),
+            "zeta": jnp.linspace(0, 2 * jnp.pi * num_transit, Y_B * num_transit),
+            "quad": get_quadrature(
+                leggauss(self._hyperparam.pop("num_quad")),
+                (automorphism_sin, grad_automorphism_sin),
+            ),
+            "alpha": self._hyperparam.pop("alpha"),
+            "quad_weights": self._hyperparam.pop("quad_weights"),
```
f0uriest: see above

desc/objectives/_fast_ion.py
```diff
@@ -305,29 +317,37 @@ def _build_spline(self, use_jit=True, verbose=1):
...
-            self._keys_1dr + [self._key], eq, self._grid
-        )
         super().build(use_jit=use_jit, verbose=verbose)
 
     def _compute_spline(self, params, constants=None):
         if constants is None:
-            constants = self.constants
+            constants = self._constants
+        else:
+            warnif(
```
f0uriest: probably dont need this here since its already a private function

desc/objectives/_free_boundary.py
```diff
@@ -257,11 +263,10 @@ def print_value(self, args, args0=None, **kwargs):
         f = self.compute_unscaled(*args, **kwargs)
         f0 = self.compute_unscaled(*args0, **kwargs) if args0 is not None else f
         # try to do weighted mean if possible
-        constants = kwargs.get("constants", self.constants)
-        if constants is None:
+        if not hasattr(self, "_constants"):
```
f0uriest: minor point but its usually clearer if you remove the `not` and reverse the order
```python
if hasattr(self, "constants"):
    w = self.constants[...]
else:
    w = ones_like(...)

File: desc/objectives/_free_boundary.py [hunk 9]
```diff
@@ -777,11 +788,10 @@ def print_value(self, args, args0=None, **kwargs):
...
```
f0uriest: see above

desc/objectives/_neoclassical.py
```diff
@@ -210,24 +212,24 @@ def build(self, use_jit=True, verbose=1):
...
+            "clebsch": FourierChebyshevSeries.nodes(
+                self._X,
+                self._Y,
+                self._grid.compress(self._grid.nodes[:, 0]),
+                domain=(0, 2 * np.pi),
+            ),
+            "fieldline quad": leggauss(self._hyperparam["Y_B"] // 2),
+            "quad": chebgauss2(self._hyperparam.pop("num_quad")),
+            "alpha": self._hyperparam.pop("alpha"),
+            "quad_weights": self._hyperparam.pop("quad_weights"),
```
f0uriest: see above

desc/objectives/objective_funs.py
```diff
@@ -437,15 +438,13 @@ def _set_things(self, things=None):
         self._flatten = _ThingFlattener(len(flat_), treedef_)
 
     @jit
-    def compute_unscaled(self, x, constants=None):
```
f0uriest, line 607: I think we don't want to remove the option until after its deprecated

desc/objectives/_bootstrap.py
```diff
@@ -258,9 +258,6 @@ def compute(self, params, constants=None):
         ----------
         params : dict
             Dictionary of equilibrium degrees of freedom, eg Equilibrium.params_dict
-        constants : dict
```
f0uriest, line 261: I think best practice is to leave it in the docstring but note that it is deprecated

desc/objectives/objective_funs.py
```diff
@@ -454,27 +453,20 @@ def compute_unscaled(self, x, constants=None):
 
         """
         params = self.unpack_state(x)
-        if constants is None:
-            constants = self.constants
-        assert len(params) == len(constants) == len(self.objectives)
+        assert len(params) == len(self.objectives)
         f = jnp.concatenate(
-            [
```
f0uriest: to maintain the API we can still allow constants to be passed through here, but wrap it in `warnings.catch_warnings` so it only warns once

desc/objectives/objective_funs.py
```diff
@@ -947,11 +891,6 @@ def compile(self, mode="auto", verbose=1):
             timer.disp("Total compilation time")
         self._compiled = True
 
-    @property
-    def constants(self):
```
f0uriest, line 1113: don't just delete this, keep it public but add a deprecation

desc/optimize/aug_lagrangian.py
f0uriest, line 1: these changes (and the ones to other low level optimizers) aren't needed. Let's keep these the same, since they can still handle functions with additional args

desc/optimize/utils.py
f0uriest, line 1: these changes also aren't needed. No reason to delete functionality.

f0uriest (changes requested):

desc/objectives/_neoclassical.py
```diff
@@ -252,8 +252,6 @@ def _build_bounce1d(self, use_jit=True, verbose=1):
         super().build(use_jit=use_jit, verbose=verbose)
 
     def _compute_bounce1d(self, params, constants=None):
-        if constants is None:
-            constants = self.constants
```
ddudt, line 256: [todo(blocking)](https://pullpo.io/cc?l=todo&d=blocking) 
Did you forget to add `constants = self._get_deprecated_constants(constants)` to this function?

desc/objectives/objective_funs.py
```diff
@@ -1088,6 +1083,21 @@ def compile(self, mode="auto", verbose=1):
             timer.disp("Total compilation time")
         self._compiled = True
 
+    def _get_deprecated_constants(self, constants=None):
+        """Return constants and thraw deprecation warning."""
```
ddudt: thraw -> throw

desc/objectives/objective_funs.py
```diff
@@ -1730,6 +1740,21 @@ def xs(self, *things):
             )
         return tuple([t.params_dict for t in things])
 
+    def _get_deprecated_constants(self, constants=None):
+        """Return constants and thraw deprecation warning."""
```
ddudt: thraw -> throw

desc/objectives/_neoclassical.py
YigitElma: This private function is called by the main `compute`, where I make the check. Either I can call the check twice or once at the beginning, I chose to have it once. I can have it twice if people want.

ddudt: > `def _compute_bounce1d(self, params, constants=None):`

It feels awkward to have `constants` be an optional argument when it really isn't optional (I don't think it will work properly with the default value of `None`). It isn't an issue because this is a private function and `constants` is always passed when it is called, but maybe we should just make it a required argument?

YigitElma: Make the check twice in both branches

ddudt (dismissed):

rahulgaur104:
Should there be a note somewhere for new users about not using constants?
I guess currently it's deprecation, not removal!

rahulgaur104 (dismissed):

YigitElma:
> Should there be a note somewhere for new users about not using constants?
> I guess currently it's deprecation, not removal!

I updated the adding objectives docs page.

desc/objectives/objective_funs.py
```diff
@@ -1088,6 +1083,21 @@ def compile(self, mode="auto", verbose=1):
...
+        if constants is None:
+            constants = [None] * len(self.constants)
+        else:
+            warnif(
+                not all(c is None for c in constants),
+                FutureWarning,
+                "constants is deprecated and will be removed in a future "
+                "release. Users should not include constants in the arguments "
+                "of their objective compute methods. Instead declare all the "
+                "constants in the build method and use as obj._constants.",
```
f0uriest, line 1118: Is the plan to keep the constants attribute? Or do we want to get rid of that as well and just reference obj.whatever directly?

desc/objectives/objective_funs.py
```diff
@@ -1088,6 +1083,21 @@ def compile(self, mode="auto", verbose=1):
...
+            )
+        return constants
+
     @property
     def constants(self):
```
f0uriest, line 1123: Should there be a deprecation warning here too?

desc/objectives/objective_funs.py
```diff
@@ -1730,6 +1740,21 @@ def xs(self, *things):
...
+                FutureWarning,
+                "constants is deprecated and will be removed in a future "
+                "release. Users should not include constants in the arguments "
+                "of their objective compute methods. Instead declare all the "
+                "constants in the build method and use as self._constants.",
+            )
+        return constants
+
     @property
     def constants(self):
```
f0uriest, line 1863: Should there be a deprecation warning here too

f0uriest (commented):
Not sure what the standard approach is re: documenting deprecated args, is it better to leave it in the docstring with a deprecated marker, or just remove it from the docs entirely?

desc/objectives/objective_funs.py
ddudt: [suggestion](https://pullpo.io/cc?l=suggestion) 
My preference for the future is to get rid of `constants` and reference `obj.whatever` directly. I think this is what we've mostly been doing in the newest objectives, although we haven't been consistent.

YigitElma: Ok, I deprecated the property `constants`. People can still use `self._constants` to pack stuff into a single attribute or use just `self.whatever`. We don't need special property definitions for this.

YigitElma:
> Not sure what the standard approach is re: documenting deprecated args, is it better to leave it in the docstring with a deprecated marker, or just remove it from the docs entirely?

I think `numpy` and some other big libraries keep the deprecated thing in the docstring until they are fully removed. I prefer keeping `constants` in docstring with `(Deprecated)` flag.

ddudt (approved):

f0uriest (approved):

PR #1773, opened 2025-06-11
Parallelization via sharding (multi-device) composed with vmap (across memory in single device)

unalmis:
PR Age 

## Robust sharding for factor of N improvement in compute and memory

This is the continuation of #1869. Broadly, there are two methods of parallelizing computation.

1. The first is to split the computation and parallelize across the memory of a single device. This is ideal if _the problem can fit on a single device and if the device has sufficient processing power to complete the computation in parallel_.

2. The second is to split the computation and parallelize across multiple devices, which is useful when the problem size becomes too large for the first strategy.

In JAX, option 1 is done with `vmap`. When the problem size becomes larger, we mimic the benefits of option 2 with `vmap_chunked`. With `vmap_chunked` we choose the `chunk_size` such that the italic part of option 1 is best satisfied and then compute on each chunk in sequence.

This PR composes multiple device parallelization with `vmap_chunked` to improve compute ability.

With the current JAX sharding API, it is most practical to first split the data `S` across `n` devices, then on each subset of data of size `S/n` we may chunk with size `m` across the memory of a single device with `vmap_chunked`. So each device is computing on a subset of size `S/n/m` simultaneously. These API require S to be divisible by `n` and `m`, but it is just some simple extra logic to compute the remainder of `S/n/m` on _each_ device after they finish the divisible portion, and likewise compute the remainder of `S/n` on a _single_ device after that.

The net result will be reducing the compute time. This also ensures that if there was a memory bottlneck in constructing the Jacobian, then this bottleneck will be pushed away toward storing the Jacobian. Then that can later be avoided by using iterative subspace methods so that only a subset of the Jacobian needs to be stored at a time.

If the use case is not clear, then explicitly, this would enable parallelization without rebuilding an objective for each flux surface/duplcating computation and transforms.


unalmis:
note to self: Maybe check if there's something in here that does the sharding better for pytrees https://github.com/netket/netket/blob/master/netket/jax/sharding.py

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ New Features
 - Ability to obtain the top eigenvalues and the corresponding eigenfunctions from the ``ideal ballooning lambda`` compute function by specifying the variable ``Neigvals``.
 - Parallelized ideal ballooning stability and Newcomb ballooning metrics and [other improvements](https://github.com/PlasmaControl/DESC/pull/1763).
 - Adds ``FourierXYCoil`` to compatible coils for ``CoilSetArclengthVariance`` objective.
+- Support for computing and construction of Jacobian in parallel.
```
YigitElma: I think if the scope of the PR will be limited by the current code, then it should be more general like `utility functions to make batched_vectorize work with pmap and sharding` etc. This sentence can create false impression

desc/batching.py
```diff
@@ -42,32 +33,98 @@
...
+
+    sharding = canonicalize_sharding(core.typeof(leaf).sharding, "lax.map")
+    if sharding is not None and sharding.mesh._any_axis_explicit:
+        return auto_axes(f, out_sharding=sharding, axes=sharding.mesh.explicit_axes)(
+            leaf
+        )
+    return f(leaf)
+
+
+def _batch_and_remainder(x, batch_size: int):
```
YigitElma: Are these changes from most recent JAX source-code since the last update of these functions? Then, maybe we should add a version information too.

desc/batching.py
```diff
@@ -134,15 +193,80 @@ def f_(*args, **kwargs):
     return f_
 
 
+def make_shardable(f, axis=0, num_devices=None):
```
YigitElma: I am not sure how this will behave with `objective.jac_scaled_error()`. Have you tested it?

My concern is the first argument of the function is `self`, so the objective, which has constants etc. When you call this function `make_shardable`, I assume it divides all the arrays to all the devices, but I think this is not what you want. For example, if you shard the transform matrices, other GPUs will need to access the data from an other GPU. If the idea is to shard tangent directions to different devices, and run them in parallel, I think you need to replicate sharding the `self.constants` and just `make_shardable` the tangent directions. So, for instance, if you have 10 tangent directions, 3 devices, each device should compute 3 tangent directions first with each having their own `self.constants` since they will need it, and compute the last one separately.

The remainder in your code will be different for different array shapes in the same pytree, I think this will create array dimension incompatibilities.

I had similar problems when I tried this before.

desc/batching.py
```diff
@@ -290,6 +420,15 @@ def batch_map(
 def batched_vectorize(pyfunc, *, excluded=frozenset(), signature=None, chunk_size=None):
```
YigitElma: Need to be able to pass 'shard_input_data`

desc/batching.py
```diff
@@ -134,15 +193,80 @@ def f_(*args, **kwargs):
...
     vmapped_fun,
     chunk_size,
     argnums,
     reduction=None,
     chunk_reduction=_identity,
+    shard_input_data=False,
     *args,
     **kwargs,
 ):
+    if shard_input_data:
```
YigitElma: Can you `jit/grad` this function? Have you tested it? I think the sharding should happen before any `jit` wrapper. But if it allows creating a custom sharding in the jit function, then it is fine.

tests/test_backend.py
```diff
@@ -140,3 +143,25 @@ def test_lstsq():
     np.testing.assert_allclose(
         _lstsq(A, b), np.linalg.lstsq(A, b, rcond=None)[0], rtol=1e-6
     )
+
+
+@pytest.mark.xfail(reason="The flags need to be set before desc.backend is imported.")
```
YigitElma: Btw, to make this PR work with multiple GPUs, you need the changes from #1495 in `set_device` and couple other places.

desc/batching.py
```diff
@@ -134,15 +193,80 @@ def f_(*args, **kwargs):
...
+            vmapped_fun,
+            chunk_size,
+            argnums,
+            reduction,
+            chunk_reduction,
+            False,
+            args_remainder,
+            **kwargs,
+        )
+        return _concat(out_shardable, out_remainder)
```
YigitElma: Does `_concat` work with inputs of different sharding? If you have actually 2 different devices, worth testing on cluster

tests/test_backend.py
```diff
@@ -140,3 +143,25 @@ def test_lstsq():
...
+    os.environ["XLA_FLAGS"] = flags
+
+    f = np.arange(201)
+    sf, rf = make_shardable(f, num_devices=num_cpu)
+    assert sf.size == 200
+    assert rf.size == 1
+
+    gsf = jnp.sin(sf)
+    grf = jnp.sin(rf)
+    g = np.concatenate([gsf, grf])
```
YigitElma: As a test, you can try to call `ObjectiveFunction._jvp_batched()` with `shard_input_data`, and also call some objectives jac method with it.

desc/batching.py
unalmis: the computation is done now so it should be put back on one device since we haven't implemented parallel svd yet

CHANGELOG.md
unalmis, line 19:
```suggestion
- Adds experimental backend support for computing and construction of Jacobian in parallel
```

unalmis:
@PlasmaControl/desc-dev we can merge this unless someone wants to take over.

desc/batching.py
unalmis: You mentioned a few different applications which have different expectations and requirements. This PR is for sharding and parallelizing computations as discussed in https://docs.jax.dev/en/latest/notebooks/Distributed_arrays_and_automatic_parallelization.html.

This is for chunked vmap of separable computations. This code of course cannot magically parallelize a fundamentally serial algorithm. So users should not be sharding dense 3D pseduospactral matrices across 2D surfaces in your example with the expectation that an array required for each computation will not be duplicated across devices. (In that case this approach only improves the speed of the computation and not memory.)

The use case of this PR is when we have a function we would like to compute on lots of input data. We want to scale up the performance so we split the data across devices and apply the same function. For example this is very useful for the vmap_chunked in the singular integrals. There no transform matrices are precomputed because the computations are done with FFTs, so no data is duplicated across devices. this is another utility of using FFTs rather than precomputed and stored matrices for transforms: sharding across devices doesn't duplicate transform data on each device if FFTs are used.

Note that f is marked as the input data to shard, not the function to compute. The standard requirement for vmap is that the vmapped axis has the same length for all the inputs. That is the same requirement here, so there are no unexpected array shape issues.

desc/batching.py
unalmis: The provided function `vmapped_fun` is already vmapped and jitted. The shard_input_data flag is a general purpose convenience method if the input data is not already sharded.
A function needs to jit compiled to know how to do `automatic sharding` which is the method used here. See https://docs.jax.dev/en/latest/notebooks/Distributed_arrays_and_automatic_parallelization.html#computation-follows-data-sharding-and-is-automatically-parallelized

This approach is tested to work for stuff like the examples given in that link.

desc/batching.py
unalmis: Maybe there's some purred structure that doesn't  work  but in almost all cases i have in mind f is just a big array

desc/batching.py
unalmis: Yes

desc/batching.py
unalmis: i haven't tested with stuff like objective.jac_scaled_error() so I didn't make that change

desc/batching.py
unalmis: https://docs.jax.dev/en/latest/notebooks/Distributed_arrays_and_automatic_parallelization.html#computation-follows-data-sharding-and-is-automatically-parallelized

desc/batching.py
unalmis: Yes

dpanici:
- Vmap changes could still be put into master, make new PR for this

PR #1775, opened 2025-06-14
Make classes hashable

### LINKED ISSUE #1781: Set default values as their final type for less compilation
For example, in trust region subproblem solvers, we pass `alpha=None`, then pass a value of float. This results in double compilation.

```python
@jax.jit
def f(x):
    return x

a_s = [None, 0, 1, 1., 0.]
for a in a_s:
    print(f"========Value: {a} Type: {type(a)}=========")
    with jax.log_compiles():
        f(a)
```
```
========Value: None Type: <class 'NoneType'>=========
WARNING:2025-06-16 21:01:24,192:jax._src.dispatch:195: Finished tracing + transforming f for pjit in 0.000533104 sec
WARNING:2025-06-16 21:01:24,198:jax._src.interpreters.pxla:1921: Compiling f with global shapes and types []. Argument mapping: ().
WARNING:2025-06-16 21:01:24,204:jax._src.dispatch:195: Finished jaxpr to MLIR module conversion jit(f) in 0.003754854 sec
WARNING:2025-06-16 21:01:24,212:jax._src.dispatch:195: Finished XLA compilation of jit(f) in 0.005926609 sec
========Value: 0 Type: <class 'int'>=========
WARNING:2025-06-16 21:01:24,214:jax._src.dispatch:195: Finished tracing + transforming f for pjit in 0.000160217 sec
WARNING:2025-06-16 21:01:24,217:jax._src.interpreters.pxla:1921: Compiling f with global shapes and types [ShapedArray(int64[], weak_type=True)]. Argument mapping: (UnspecifiedValue,).
WARNING:2025-06-16 21:01:24,228:jax._src.dispatch:195: Finished jaxpr to MLIR module conversion jit(f) in 0.008798122 sec
WARNING:2025-06-16 21:01:24,250:jax._src.dispatch:195: Finished XLA compilation of jit(f) in 0.017543077 sec
========Value: 1 Type: <class 'int'>=========
========Value: 1.0 Type: <class 'float'>=========
WARNING:2025-06-16 21:01:24,261:jax._src.dispatch:195: Finished tracing + transforming f for pjit in 0.000915289 sec
WARNING:2025-06-16 21:01:24,264:jax._src.interpreters.pxla:1921: Compiling f with global shapes and types [ShapedArray(float64[], weak_type=True)]. Argument mapping: (UnspecifiedValue,).
WARNING:2025-06-16 21:01:24,273:jax._src.dispatch:195: Finished jaxpr to MLIR module conversion jit(f) in 0.007640839 sec
WARNING:2025-06-16 21:01:24,290:jax._src.dispatch:195: Finished XLA compilation of jit(f) in 0.011921167 sec
========Value: 0.0 Type: <class 'float'>=========
```

Anywhere else might have made this mistake?

YigitElma:
- Make basis class hashable and set it static in profiles and transforms
- Set Levemberg-Marquardt parameter `alpha` as float instead of `None` for preventing extra compilation (Resolves #1781)
- Move constants to device, and make them JAX arrays after build (deleted in #1662)

See #1662 for the later changes to this PR

### DIFF
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -317,6 +317,21 @@ class _Basis(IOAble, ABC):
             )
         )
 
+    def __hash__(self):
+        """Get the hash of the object."""
+        return hash(
+            (
+                self.__class__.__name__,
+                self._L,
+                self._M,
+                self._N,
+                self._NFP,
+                self._sym,
+                self._spectral_indexing,
+                # add device/sharding info for multi-device support
+            )
+        )
+
 
 class PowerSeries(_Basis):
     """1D basis set for flux surface quantities.
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -36,6 +36,17 @@ class _Basis(IOAble, ABC):
         "_spectral_indexing",
     ]
 
+    _static_attrs = [
+        "_fft_poloidal",
+        "_fft_toroidal",
+        "_L",
+        "_M",
+        "_N",
+        "_NFP",
+        "_sym",
+        "_spectral_indexing",
+    ]
+
     def __init__(self):
         self._enforce_symmetry()
         self._sort_modes()
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -328,8 +328,22 @@ class _Basis(IOAble, ABC):
                 self._NFP,
                 self._sym,
                 self._spectral_indexing,
-                # add device/sharding info for multi-device support
+                # add device/sharding info for multi-device support?
             )
+        )
+
+    def __eq__(self, other):
+        """Check if two basis objects are equal."""
+        if not isinstance(other, _Basis):
+            return False
+        return (
+            self.__class__ == other.__class__
+            and self.L == other.L
+            and self.M == other.M
+            and self.N == other.N
+            and self.NFP == other.NFP
+            and self.sym == other.sym
+            and self.spectral_indexing == other.spectral_indexing
         )
 
 
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -339,7 +339,6 @@ class _Basis(IOAble, ABC):
                 self._NFP,
                 self._sym,
                 self._spectral_indexing,
-                # add device/sharding info for multi-device support?
             )
         )
 
--- a/desc/objectives/objective_funs.py
+++ b/desc/objectives/objective_funs.py
@@ -1207,8 +1207,9 @@ class _Objective(IOAble, ABC):
         if not self._use_jit:
             self._unjit()
 
-        # put the constants to device (CPU or GPU) as jax arrays
-        self._constants = jax.device_put(self.constants, jax.devices()[0])
+        # put the constants to device as jax arrays
+        if desc_config["kind"] == "gpu":
+            self._constants = jax.device_put(self.constants, jax.devices()[0])
 
         self._built = True
 
--- a/desc/objectives/objective_funs.py
+++ b/desc/objectives/objective_funs.py
@@ -232,7 +232,23 @@ class ObjectiveFunction(IOAble):
 
     """
 
-    _io_attrs_ = ["_objectives"]
+    _io_attrs_ = [
+        "_deriv_mode",
+        "_jac_chunk_size",
+        "_name",
+        "_objectives",
+        "_use_jit",
+    ]
+    _static_attrs = [
+        "_built",
+        "_compile_mode",
+        "_compiled",
+        "_deriv_mode",
+        "_jac_chunk_size",
+        "_name",
+        "_things_per_objective_idx",
+        "_use_jit",
+    ]
 
     def __init__(
         self,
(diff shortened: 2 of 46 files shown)


tests/test_transform.py
```diff
@@ -326,15 +326,16 @@ def test_direct_fft_equal(self):
...
-        t1f.change_resolution(grid, basis1)
-        t2f.change_resolution(grid, basis2)
-        t3f.change_resolution(grid, basis3)
-        t1d1.change_resolution(grid, basis1)
-        t2d1.change_resolution(grid, basis2)
-        t3d1.change_resolution(grid, basis3)
-        t1d2.change_resolution(grid, basis1)
-        t2d2.change_resolution(grid, basis2)
-        t3d2.change_resolution(grid, basis3)
+        # should pass the methods, otherwise default might change
```
f0uriest, line 329: this is addressed separately in #1508

tests/test_basis.py
```diff
@@ -400,6 +400,42 @@ def test_basis_resolutions_assert_integers(self):
...
+        )
+        all_bases.append(
+            [
+                ChebyshevDoubleFourierBasis(L=3, M=2, N=1),
+                ChebyshevDoubleFourierBasis(L=3, M=2, N=1),
+            ]
+        )
+        all_bases.append([ChebyshevPolynomial(L=3), ChebyshevPolynomial(L=3)])
+        for basis1, basis2 in all_bases:
+            # check that hash is consistent
```
f0uriest, line 423: we should also add checks that different (but similar) bases have different hashes and don't compare equal

desc/objectives/objective_funs.py
```diff
@@ -1243,6 +1244,10 @@ def build(self, use_jit=True, verbose=1):
         if not self._use_jit:
             self._unjit()
 
+        # put the constants to device as jax arrays
+        if desc_config["kind"] == "gpu":
+            self._constants = jax.device_put(self.constants, jax.devices()[0])
```
f0uriest, line 1249: I'd prefer to move this to #1495 

Also note that this can cause problems if for example, `x` is put on the CPU, so I think it's better if we only do it within an optimization loop and not here

f0uriest: besides, if we're using the GPU than all jax arrays should default to being on the GPU already, but uncommitted, so they can be used with CPU arrays. whereas this puts them on the gpu committed, so that things will error if you try to use them in a cpu calculation

tests/test_transform.py
YigitElma: I came up with this problem after I changed `get_matrices` function

desc/objectives/objective_funs.py
YigitElma: This is not related to #1495, in general, every time we call compute methods, there is some host-to-device transfer overhead (for compute functions, it is longer than compute itself, see [the discussion](https://github.com/jax-ml/jax/issues/29470#issuecomment-2978319648)).

For CPU stuff, I am not sure if this is a problem (maybe for `ExternalObjective` etc it is) in general. Do you suggest moving it to `LinearConstraintProjection` or `optimizer.optimize`?

YigitElma: In either case, original `constants` will be moved to GPU, so sub-objective constants will live on GPU, the only difference is that this transfer will occur on a different time

desc/basis.py
```diff
@@ -328,6 +328,35 @@ def __repr__(self):
...
+        return hash(
+            (
+                self.__class__.__name__,
+                self._L,
+                self._M,
+                self._N,
+                self._NFP,
+                self._sym,
+                self._spectral_indexing,
+                # add device/sharding info for multi-device support?
```
dpanici: make a TODO/issue for this

tests/test_basis.py
dpanici: is this test actually doing this? they are comparing equal and have same hash

dpanici (changes requested):
just make a todo for the thing I marked, also is the test doing what Rory had asked for?

desc/basis.py
YigitElma: I was not sure about this in the time of writing but I think this is not necessary. Jax has some internal checks for sharding etc already, so our hash doesn't need to be different for same values on different devices.

tests/test_basis.py
YigitElma: I have added a new one at the bottom. I am checking whether all the hashes are different for different basis types (with same LMN values). If some of them have same hash then `set()` will remove the duplicates and will assert.

desc/basis.py
YigitElma: I think these should have the same hash
```python
with jax.default_device(jax.devices("cpu")[0]):
    basis1 = FourierZernikeBasis(L=3, M=3, N=3)
basis2 = FourierZernikeBasis(L=3, M=3, N=3)
print(hash(basis1) == hash(basis2))
```

dpanici (approved):

desc/objectives/objective_funs.py
YigitElma: You can delete this after merging it with your PR

f0uriest (approved):

PR #1782, opened 2025-06-17
gds2

unalmis:


tests/test_stability_funs.py
```diff
@@ -522,6 +523,40 @@ def test_ballooning_geometry(tmpdir_factory):
...
+    """Test computation of gds2."""
+    eq = desc.examples.get("W7-X")
+    iota_zeta0 = np.linspace(-np.pi / 2, np.pi / 2, 15)[:, np.newaxis]
+    data = eq.compute(
+        ["alpha_r (secular)", "iota_r", "iota", "gds2", "c ballooning"],
+        zeta0=iota_zeta0,
+    )
+    gds2 = data["gds2"]
+    c_ballooning = data["c ballooning"]
+    data = eq.compute(
```
unalmis: @rahulgaur104 This line redefines $\alpha_{DESC}$ to be the same $\alpha$ in your ballooning paper ($\alpha_{DESC} + \iota \zeta_0$). Therefore statements like ``np.testing.assert_allclose(gds2, data["g^aa"] * data["rho"] ** 2)`` must pass if `gds2` is computed correctly.

On this branch, this test passes.

rahulgaur104: Well, this was passing in master in `test_ballooning_geometry` because I calculate the geometry using two different methods. I may have redefined alpha = alpha_DESC -iota zeta_0 in my definition of gds2.

So I don't know why this is failing. I still believe my calculations are correct in master.

unalmis: It will fail on `master` because it tests that `gds2` $=$ ρ² |∇(α + ι ζ₀)|². This relation is not true with the introduction of the sign factors: sign(iota) on the `g^ra` term.

unalmis:
merging now as requested in dev meeting

PR #1784, opened 2025-06-17
Keep the linear constraint projection scaling constant

YigitElma:
Keep the linear constraint projection scaling constant throughtout the optimization

### DIFF
--- a/desc/optimize/_constraint_wrappers.py
+++ b/desc/optimize/_constraint_wrappers.py
@@ -230,23 +230,10 @@ class LinearConstraintProjection(ObjectiveFunction):
         # remove fixed parameters from A and b again by the same loop as in factorize
         # Actually A (unscaled linear constraint matrix without any degenerate rows)
         # does not change here, but still recompute it while updating others
-        A, b, xp, unfixed_idx, fixed_idx = remove_fixed_parameters(A, b, xp)
+        _, b, xp, _, fixed_idx = remove_fixed_parameters(A, b, xp)
 
-        x_scale = self._objective.x(*self._objective.things)
-        self._D = jnp.where(jnp.abs(x_scale) < 1e2, 1, jnp.abs(x_scale))
-
-        # since D has changed, we need to update the ADinv
-        # as mentioned above A does not change, so we can use the same Ainv
-        # pinv(A) = Ainv, ADinv = pinv(A @ D) = Dinv @ Ainv, Dinv = 1 / D
-        self._ADinv = (1 / self._D)[unfixed_idx, None] * self._Ainv
-        # we also need to update the nullspace Z of AD in a similar way
-        # A @ ZA = 0 -> (A @ D) @ ((1 / D) @ ZA) = 0 -> Z = (1 / D) @ ZA
-        # where ZA is the nullspace of A, and Z is the nullspace of AD
-        self._Z = (1 / self._D)[self._unfixed_idx, None] * self._ZA
-        # we also normalize Z to make each column have unit norm
-        self._Z = self._Z / jnp.linalg.norm(self._Z, axis=0)
-
-        xp = put(xp, unfixed_idx, self._ADinv @ b)
+        # When we change some fixed parameter values, A, Z and D doesn't change
+        xp = put(xp, self._unfixed_idx, self._ADinv @ b)
         xp = put(xp, fixed_idx, ((1 / self._D) * xp)[fixed_idx])
         # cast to jnp arrays
         self._xp = jnp.asarray(xp)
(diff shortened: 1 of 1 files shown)


desc/optimize/_constraint_wrappers.py
```diff
@@ -351,7 +338,7 @@ def grad(self, x_reduced, constants=None):
         """
         x = self.recover(x_reduced)
         df = self._objective.grad(x, constants)
-        return df[self._unfixed_idx] @ (self._Z * self._D[self._unfixed_idx, None])
+        return df[self._unfixed_idx] @ self._ZA
```
dpanici: I believe we want this to retain the scaling matrix D (`_ZA` is `self._Z * self._D[self._unfixed_idx, None]` but then its columns are scaled to be unit norm, so it is not a drop-in replacement)

desc/optimize/_constraint_wrappers.py
```diff
@@ -374,7 +361,7 @@ def hess(self, x_reduced, constants=None):
         return (
             (self._Z.T * (1 / self._D)[None, self._unfixed_idx])
             @ df[self._unfixed_idx, :][:, self._unfixed_idx]
-            @ (self._Z * self._D[self._unfixed_idx, None])
+            @ self._ZA
```
dpanici: same here

desc/optimize/_constraint_wrappers.py
```diff
@@ -491,7 +478,7 @@ def jvp_unscaled(self, v, x_reduced, constants=None):
     def _vjp(self, v, x_reduced, constants=None, op="vjp_scaled"):
         x = self.recover(x_reduced)
         df = getattr(self._objective, op)(v, x, constants)
-        return df[self._unfixed_idx] @ (self._Z * self._D[self._unfixed_idx, None])
+        return df[self._unfixed_idx] @ self._ZA
```
dpanici: same here

dpanici (changes requested):
undo the _ZA change, maybe instead make another self attribute for the un-normalized ZA and use that

dpanici (dismissed):

desc/objectives/utils.py
```diff
@@ -120,6 +122,7 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
 
     # compute x_scale if not provided
     # Note: this x_scale is not the same as the x_scale as in solve_options["x_scale"]
+    # but the one given as solve_options["linear_constraint_options"]["x_scale"]
```
rahulgaur104, line 125: Really! Should we add a comment stating this somewhere?

desc/optimize/_constraint_wrappers.py
```diff
@@ -232,19 +236,21 @@ def update_constraint_target(self, eq_new):
...
-        # pinv(A) = Ainv, ADinv = pinv(A @ D) = Dinv @ Ainv, Dinv = 1 / D
-        self._ADinv = (1 / self._D)[unfixed_idx, None] * self._Ainv
-        # we also need to update the nullspace Z of AD in a similar way
-        # A @ ZA = 0 -> (A @ D) @ ((1 / D) @ ZA) = 0 -> Z = (1 / D) @ ZA
-        # where ZA is the nullspace of A, and Z is the nullspace of AD
-        self._Z = (1 / self._D)[self._unfixed_idx, None] * self._ZA
-        # we also normalize Z to make each column have unit norm
-        self._Z = self._Z / jnp.linalg.norm(self._Z, axis=0)
+        # if user specified x_scale, don't dynamically change it
+        if self._x_scale == "auto":
```
rahulgaur104, line 240: From the comment it sounds like when the self._x_scale="auto" means user-specified x_scale?
But from the if condition it looks like self._xscale="auto" means it's automatically chosen.

rahulgaur104 (dismissed):
Some more clarification would be nice.

desc/objectives/utils.py
rahulgaur104: I guess you do add it in the docstring later.

desc/optimize/_constraint_wrappers.py
YigitElma: Yeah, if it is `auto` then we choose it by values of the parameters. The `solve_options["linear_constraint_options"]["x_scale"]` was added recently, we forgot to update this part. The code was using the given x_scale only for 1 optimization iteration (for eq solve this is not a problem), then moving to computing it automatically!

desc/objectives/utils.py
YigitElma: @ddudt added this option in #1697, this is briefly mentioned here https://github.com/PlasmaControl/DESC/blob/1d4806d280d11722839a73aee74f0888d07eeca1/desc/optimize/optimizer.py#L144

And with the change in `LinearConstraintProjection`s docs, it is complemented, but yes, this is a bit hard to find.

dpanici (approved):

ddudt (approved):

PR #1789, opened 2025-06-21
Differentiation matrices

### LINKED ISSUE #1777: Differentiation matrices
# Differentiation Matrices 

Differentiation matrices are an interesting choice to calculate the numerical derivative of a smooth function. Once the matrix *D* is assembled, obtaining the derivative of a function vector **f** reduces to a single operation

$$
f\\, \prime \approx\\; D\\,f
$$

which should be fast on a GPU.

A simple 1D example on a uniform grid would be a tridiagonal finite-difference matrix. Increasing the FD order widens the bandwidth and drops the truncation error to $\mathcal{O}(\Delta x^{p})$.  
This technique can also be implemented with a spectral instead of a finite-difference representation, potentially increasing the accuracy of the derivatives.

For a tensor-product grid $(\rho,\vartheta,\varphi)$ in which the coordinates are **independent**, directional derivatives commute. Multi-D operators can be assembled with Kronecker products:

$$
\begin{aligned}
\nabla\_\varphi   &= I\_\rho     \otimes I\_\vartheta \otimes D\_\varphi
\end{aligned}
$$

Issues: Some spectral differentiation matrices are ill-conditioned and have no nice properties (symmetry, orthogonality). Writing a full matrix in 3D can be expensive in terms of memory consumption even with the current GPUs. But it would still be a nice feature to have for 1D, 2D or low-resolution 3D derivatives.

---

## 5 · References
1. J. A. C. Weideman & S. C. Reddy,  
   “A MATLAB Differentiation Matrix Suite,” *ACM TOMS* **26** (2000) 465-519.  
   <https://doi.org/10.1145/365723.365727>

2. L. N. Trefethen, *Spectral Methods in MATLAB*. SIAM, 2000.

rahulgaur104:
Addresses #1777 

- [x] Add the differentiation matrix code
- [x] Pass the diffmatrices to compute function as a `diffmat` object as a part of `transform`
- [x] Solve the ideal ballooning problem using differentiation matrices and compare with the old finite-difference solver
- [x] Update tests to increase coverage


rahulgaur104:
Compare the fourier diffmat with DESC diffmat calculation. Think about the end point!

desc/diffmat_utils.py
```diff
@@ -0,0 +1,257 @@
...
+        return jnp.array([])
+
+    if domain is None:
+        domain = [-1, 1]
+
+    k = jnp.arange(n)
+    x = jnp.sin(jnp.pi * jnp.flip(2 * k - n + 1) / (2 * (n - 1)))
+
+    # Affine map to [a, b] only when necessary (allows JIT constant folding)
+    if domain[0] != -1 or domain[1] != 1:
```
unalmis: This is not necessary. JIT compilation will recognize addition with static floats that reduces to 0 and multiplications/divisions with static floats that reduce to 1 as identity operations and remove them in the compiled code.[^1] You can check the compiled HLO code to see if curious.

[^1]: I had checked the compiled HLO code is identical for all the domain shifting for the functions in desc/integrals.

desc/diffmat_utils.py
```diff
@@ -0,0 +1,264 @@
...
+from functools import partial
+
+from jax import tree_util
+
+from desc.backend import jit, jnp
+from desc.integrals.quad_utils import leggauss_lob
+
+
+@tree_util.register_pytree_node_class
+class DiffMat:
```
YigitElma: Is there a reason for not using `IOable` here and register the class manually. If you just inherit from `IOable`, it will handle registration, `tree_flatten` and `tree_unflatten` for you.

desc/objectives/_stability.py
```diff
@@ -489,6 +496,7 @@ def build(self, use_jit=True, verbose=1):
             self._iota_keys + ["ideal ballooning lambda"], eq, iota_grid
         )
         self._constants = {
+            "diffmat": self._diffmat,
```
YigitElma: In general, I don't see a point in having a reference to the same thing in the constants, and eventually it will be cleaner to get rid of the constants #1769

tests/test_diffmatrices.py
```diff
@@ -0,0 +1,456 @@
...
+def _eval_3D(f, x, y, z, NFP):
+    """Evaluate function f at grid points (x[i], y[j], z[k])."""
+    return vmap(
+        lambda z_val: vmap(
+            lambda y_val: vmap(lambda x_val: f(x_val, y_val, z_val, NFP))(x)
+        )(y)
+    )(z)
+
+
+# --no-verify a = 0.015  # [0.0, 0.1]
```
YigitElma: Were these for debugging?

tests/test_diffmatrices.py
```diff
@@ -0,0 +1,456 @@
...
+    if dz_order == 0:
+        Dz = jnp.eye(nz)
+    elif dz_order == 1:
+        D, _ = fourier_diffmat(nz)
+        Dz = D * NFP
+    elif dz_order == 2:
+        D, _ = fourier_diffmat(nz)
+        Dz = (D @ D) * NFP**2
+
+    # Create identity matrices for tensor product
```
YigitElma: these are not necessary, right? You can already use the Dx Dy Dz which is identity for the special case

desc/diffmat_utils.py
```diff
@@ -0,0 +1,264 @@
...
+    Returns
+    -------
+    jax.Array
+        Array of shape ``(n,)`` with spacing ``h = (b - a) / n``.
+    """
+    if domain is None:
+        domain = [0, 2 * jnp.pi]
+    a, b = domain
+    h = (b - a) / n
+    return jnp.arange(n) * h + a
```
YigitElma: not important but not just `jnp.linspace(domain[0], domain[1], n, endpoint=False)`?

desc/diffmat_utils.py
YigitElma: You can still add your custom `hash` and `eq` functions.

desc/objectives/_stability.py
```diff
@@ -420,6 +425,7 @@ def __init__(
         nzetaperturn=200,
         zeta0=None,
         Neigvals=1,
+        diffmat=DiffMat(),
```
YigitElma: When you have this default, it is not creating matrices for you, right?

docs/notebooks/tutorials/ideal_ballooning_stability.ipynb
YigitElma, line 1: If I understand correctly, `diffmat = DiffMat()` doesn't change the internal computation and since there is a default, it not not necessary, the code will run and do the same operations either way. Can you add this as a comment or can you show how to properly create these matrices and pass it that way? I would prefer the latter.

desc/objectives/_stability.py
```diff
@@ -563,7 +571,11 @@ def get(key):
             ["ideal ballooning lambda"],
             params,
             transforms=get_transforms(
-                ["ideal ballooning lambda"], eq, grid, jitable=True
+                ["ideal ballooning lambda"],
+                eq,
+                grid,
+                diffmat=constants["diffmat"],
```
YigitElma: I am slightly confused and I have couple questions. Most objectives have a constant grid in rho, theta, zeta, then the transforms stay a constant. Here during the optimization rho, alpha, zeta stays the same and you find the equivalent rho, theta, zeta grid. How does this affect the differentiation matrices? Do they always stay constant? Or how do you know they have proper distribution which they are generated for in the first place?

I would appreciate it if you could explain it to me.

rahulgaur104: Yeah, no problem. Yes, the rho and zeta coordinate stay the same when we go from the fixed (rho, alpha, zeta) -> (rho, theta, zeta) grid. Since the equation is always being solved at the same zeta points, the differentiation matrices are fixed and have the exact same distribution for all fieldlines and flux surfaces.

desc/objectives/_stability.py
rahulgaur104: Yes, no matrices are being created but the transform inside the registered compute function needs "diffmat". I don't know how to avoid passing an empty diffmat because the logic in desc/compute/utils.py will then throw an error. 

I was able to make this work and not the other option (where I don't pass anything).

desc/diffmat_utils.py
rahulgaur104: Yeah, I'll change that Thanks!

tests/test_diffmatrices.py
rahulgaur104: Oh, I think you are right. Let me see!

tests/test_diffmatrices.py
rahulgaur104: Yeah. I'll remove that.

desc/diffmat_utils.py
rahulgaur104: No reason. I commited the code I was able to run.
Ok, great! Can you point me to a template to follow? Like another class that looks like this one?

YigitElma: Any class in DESC inherits from that, but yes it is pretty hidden sometimes. This case, it should look like,
```python
from desc.io import IOAble

class DiffMat(IOAble):
    """Some docs"""
    _static_attrs = ["_token"]

   # your __init__, __eq__ and __hash__
```

desc/objectives/_stability.py
YigitElma: Oh so you are not using `alpha` or `theta` derivatives, right? Ok, then I understand now.

dpanici (commented):
jax ensure compile time eval + stop gradient

potentially could work here

desc/diffmat_utils.py
rahulgaur104: Added!

desc/objectives/_stability.py
rahulgaur104: Removed!

tests/test_diffmatrices.py
rahulgaur104: Removed.

desc/objectives/_stability.py
rahulgaur104: Don't have to create a DiffMat anymore.

docs/notebooks/tutorials/ideal_ballooning_stability.ipynb
rahulgaur104: I have updated the notebook with text explaining how to use diffmat without actually using diffmat in any of the notebook calculations. Other tests demonstrate how to use it.

PR #1791, opened 2025-06-22
A finite-n ideal MHD stability solver

rahulgaur104:
Finite-n stability is important...sometimes. There are no open source codes to do this. So here's something...

TODO:
- [ ] Improve the description of this PR
- [ ] Separate the diffmat

### DIFF
--- a/desc/compute/_stability.py
+++ b/desc/compute/_stability.py
@@ -13,6 +13,7 @@ from scipy.constants import mu_0
 
 from desc.backend import jax, jit, jnp, scan, vmap
 
+from ..diffmat_utils import cheb_D1, fourier_diffmat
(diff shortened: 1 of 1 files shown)


rahulgaur104:
The choice of basis is somehow destroying the self-adjoint nature of the force operator. Specifically, I refer to the Chebyshev differentiation matrix. I didn't know but unlike fourier matrices that satisfy WT F = F W for a weight matrix W (=I for Fourier) Chebyshev-Lobatto matrices don't have a simple diagonal matrix so it doesn't satisfy the weighted symmetry for a diagonal W.

My hunch is that diagonal weighted symmetry is somehow necessary to preserve the self-adjointness of the final matrix. I could be wrong and this approach may have other pathologies.

PR #1798, opened 2025-06-29
Freeobj

unalmis:


unalmis:
@dpanici It looks like this PR is for the old free surface method. That method has different math etc. then the one in the shared paper, so I don't follow the logic in the [23fd1a7](https://github.com/PlasmaControl/DESC/pull/1798/commits/23fd1a716d05549906ff004278d0333bf145e7bf) commit

PR #1802, opened 2025-07-03
Updates to `write_ascii`, along with adding deprecation warning

dpanici:
- fixes some issues with ascii io, but also adds deprecation warning so users know to stay away from it, and we will eventually remove or remove from public API
- `write_ascii` always outputs an iota profile. This makes it get the iota profile if it does not exist on an equilibrium passed in.
- `write_ascii` now handles when the input profiles are even power series, instead of listing the incorrect modenumbers


desc/io/ascii_io.py
```diff
@@ -1,9 +1,11 @@
 """Legacy functions for reading and writing ascii format."""
 
+import warnings
+
 import numpy as np
 
 
-def write_ascii(fname, eq):
+def write_ascii(fname, eqq):
```
f0uriest: would prefer to keep this name the same

desc/io/ascii_io.py
```diff
@@ -14,6 +16,11 @@ def write_ascii(fname, eq):
         dictionary of equilibrium parameters.
 
     """
+    eq = eqq.copy()  # so dont mess up original one
```
f0uriest: you can just do `eq = eq.copy()`

desc/io/ascii_io.py
```diff
@@ -14,17 +24,51 @@ def write_ascii(fname, eq):
         dictionary of equilibrium parameters.
 
     """
+    warnings.warn(
+        "text-based IO is not as well-tested or maintained as the .h5"
+        " format and is therefore not recommended, please use the"
+        " .h5 output format instead. text-based IO will be deprecated"
+        "in a future release",
```
YigitElma:
```suggestion
        " in a future release",
```

desc/io/ascii_io.py
```diff
@@ -125,6 +168,13 @@ def read_ascii(filename):
     from desc.equilibrium import Equilibrium
     from desc.utils import copy_coeffs
 
+    warnings.warn(
+        "text-based IO is not as well-tested or maintained as the .h5"
+        " format and is therefore not recommended, please use the"
+        " .h5 output format instead. text-based IO will be deprecated"
+        "in a future release",
```
YigitElma:
```suggestion
        " in a future release",
```

desc/io/ascii_io.py
```diff
@@ -1,11 +1,21 @@
...
+
 
-def write_ascii(fname, eq):
+def write_ascii(fname, eq):  # noqa: C901
     """Print the equilibrium solution to a text file.
 
+    Note: will save the pressure and iota profile as
+    power series profiles. If not an iota constrained
+    equilibrium, the iota profile will be computed and
+    fit with a fourier series of the equilibrium's L
```
YigitElma:
```suggestion
    fit with a power series of the equilibrium's L
```
I assume this is a typo?

YigitElma (changes requested):

desc/io/ascii_io.py
```diff
@@ -14,17 +24,51 @@ def write_ascii(fname, eq):
...
+        DeprecationWarning,
+    )
+    eq = eq.copy()  # so don't mess up original one
+    if eq.iota is None:
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
+            eq.iota = eq.get_profile("iota", kind="power_series")
+    from desc.profiles import FourierZernikeProfile, PowerSeriesProfile
+
     # can't import other stuff in io due to circular imports, so have to check by name
```
f0uriest: can get rid of this comment

docs/output.rst
```diff
@@ -5,12 +5,13 @@ Outputs
...
-ASCII
+HDF5
 ************
 
-.. include:: ./text_output_format.rst
+.. include:: ./binary_output_format.rst
 
-HDF5
+
+ASCII
```
f0uriest, line 15: could also add a note here recommending HDF5 instead, or possibly add a note to the docstring itself

desc/io/ascii_io.py
```diff
@@ -14,17 +24,51 @@ def write_ascii(fname, eq):
...
+        "cannot save FourierZernikeProfile as ascii",
+    )
 
     pressure = eq.pressure
     iota = eq.iota
+
+    if not isinstance(iota, PowerSeriesProfile):
+        iota = iota.to_powerseries(order=eq.L, sym=True)
+    if not isinstance(pressure, PowerSeriesProfile):
+        pressure = pressure.to_powerseries(order=eq.L, sym=True)
```
f0uriest, line 61: if the equilibrium has kinetic profiles then `pressure` is `None` so I think this will fail, but it should always be safe to do `eq.get_profile("pressure", kind="power_series")`

docs/text_output_format.rst
```diff
@@ -89,6 +88,11 @@ profiles given in the example are:
    p &= 3.4\times10^3 (1-\rho^2)^2 \\
    \iota &= 0.5 + 1.5 \rho^2.\end{aligned}
 
+If the equilibrium was solved with a fixed-current profile, then the iota
+computed from the equilibrium will be fit with a power series, and the coefficients of
+that power series will be saved. Likewise, if the equilibrium does not have a pressure profile
+but rather kinetic profiles, the pressure will be computed, fit with a power series, and saved.
```
f0uriest, line 94: see note above, I don't think this is currently correct

docs/text_output_format.rst
```diff
@@ -114,148 +118,115 @@ where :math:`L = |m|, |m|+2, |m|+4, \ldots, 2 M`.
...
 -------------------
 
 ::
 
-   NFP =   3
+   NFP =  19
    Psi =   1.00000000E+00
-   Nbdry =  7
+   Nbdry =  24
+   m:  -2 n:  -2 bR =   1.75911802E-17 bZ =   0.00000000E+00
```
f0uriest, line 164: we really should have named these `Rb`, `Zb`. `bR`,`bZ` look like components of the `b` vector. Probably not worth changing now though.

dpanici: agreed, hence my desire to just deprecate rather than fix it up, there is a pretty long list of things to improve here

rahulgaur104 (approved):

PR #1803, opened 2025-07-03
Update perlmutter installation instructions

f0uriest:


docs/installation.rst
```diff
@@ -185,6 +185,8 @@ On computing clusters you must ensure to `module load anaconda` in order to use
                 # optionally install developer requirements (if you want to run tests)
                 pip install -r devtools/dev-requirements.txt
 
+            Note that you may also need to execute `unset LD_LIBRARY_PATH` before starting a python process for the JAX/CUDA initalization to work properly.
```
daniel-dudt: Do you have to do this bash command every time before running python? So like you would have to put this line in your SLURM file when submitting a job?

YigitElma: I think yes, you should always put it according to this answer https://github.com/jax-ml/jax/issues/28929#issuecomment-2902258786.

I am not sure when environment variables get reset, but usually for new slurm jobs you need to set them again.

YigitElma (approved):

ddudt (approved):

docs/installation.rst
rahulgaur104: I think you should make it more clear in your PR
Maybe add something like
"if you are submitting a job using the SLURM scheduler please add unset LD_LIBRARY_PATH to your job file"
or just add
unset LD_LIBRARY_PATH`
in the same code black as the one used for loading modules. Then they will just copy the whole thing into their slurm.

YigitElma: @rahulgaur104 I think `before starting python process` means that. Each slurm creates a new process hence needs this line.

rahulgaur104: Most users submit "slurm jobs" on Perlmutter and don't run "python processes". The language does not appeal to me. 
I use Perlmutter more than all of you combined.

This PR already has more than 2 approvals so proceed as you deem fit.

YigitElma: I am fine with both, if someone makes the change, I will approve again.

dpanici (dismissed):
make more clear like @rahulgaur104 said

docs/installation.rst
dpanici:
```suggestion
            Note that you may also need to execute `unset LD_LIBRARY_PATH` before starting a python process (e.g. execute this as part of your slurm script, before calling python to run DESC) for the JAX/CUDA initalization to work properly.
```

PR #1805, opened 2025-07-08
new magnetic field API

unalmis:


desc/compute/_laplace.py
```diff
@@ -0,0 +1,667 @@
...
+    # TODO: https://github.com/patrick-kidger/lineax/pull/86
+    return (
+        jnp.linalg.solve(D, boundary_condition)
+        if (potential_grid.num_nodes == basis.num_modes)
+        else jnp.linalg.lstsq(D, boundary_condition)[0]
+    )
+
+
+@register_compute_fun(
+    name="interpolator",
```
dpanici, line 129: I find it a bit strange to start putting other things other than actual data into the data dictionary. Like this is an FFTInterpolator object right?

unalmis: Yes

unalmis: Was there something else haha?

PR #1814, opened 2025-07-11
Fix circular imports

### LINKED ISSUE #1240: circular imports with compute utils

unalmis:
Resolves #1240


publications/panici2022/make_plots_Panici_2022.ipynb
unalmis, line 1: what is from `desc.compute._core import dot` @dpanici

dpanici: That is from like 3 years ago, I would not touch that or worry about it, or of anything in the publications folder. Anyone trying to use that script should be checking out whatever old version of DESC I reference in the README of that folder

unalmis: ok this file was reformatted by black

dpanici (approved):

PR #1815, opened 2025-07-11
fix error in input_reader for parse_vmec_inputs PMASS_TYPE

### LINKED ISSUE #1813: Simple bug in input_reader.py
https://github.com/PlasmaControl/DESC/blob/a3781a9b1fe83827aca89f2c6ad9bae551ec87e2/desc/input_reader.py#L1178

I think there is a "\" missing: r"bPMASS_TYPE\s*=\s*\w*" -> r"\bPMASS_TYPE\s*=\s*\w*"

dpanici:
Resolve #1813

### DIFF
--- a/desc/input_reader.py
+++ b/desc/input_reader.py
@@ -1175,7 +1175,7 @@ class InputReader:
                     iota_flag = False
 
             # pressure profile
-            match = re.search(r"bPMASS_TYPE\s*=\s*\w*", command, re.IGNORECASE)
+            match = re.search(r"\bPMASS_TYPE\s*=\s*\w*", command, re.IGNORECASE)
             if match:
                 if not re.search(r"\bpower_series\b", match.group(0), re.IGNORECASE):
                     warnings.warn(colored("Pressure is not a power series!", "yellow"))


desc/input_reader.py
```diff
@@ -1175,7 +1175,7 @@ def parse_vmec_inputs(vmec_fname, threshold=0):  # noqa: C901
                     iota_flag = False
 
             # pressure profile
-            match = re.search(r"bPMASS_TYPE\s*=\s*\w*", command, re.IGNORECASE)
+            match = re.search(r"\bPMASS_TYPE.*\b", command, re.IGNORECASE)
```
f0uriest: this also removes `\s` etc, can you remind me what this is doing?

Mandelbr0t: \s* should be there to match one or more spaces. the \b signals start/end of a word made up of characters (i.e. in python regex "characters" are everything that matches \w)

dpanici: I changed it to match what `PCURR_TYPE` does, which is check if PMASS_TYPE is in the line (`command`, then inspects the whole line `command` itself to see if it contains `power_series`. For some reason before it was using a more convoluted way but this way works fine.

YigitElma (approved):

f0uriest (approved):

YigitElma (approved):

Issue #224, opened 2022-06-28
Allow reading in of boundary from a SPEC file


dpanici:
SPEC has same bdry representation and bdry input file syntax as VMEC so it would just be a matter of allowing the input reader to recognize the different namelist lines (`&XXX`, not sure what these are called)

example SPEC input file:

```
&physicslist
 Igeometry   =         3
 Istellsym   =         0
 Lfreebound  =         0
 phiedge     =   -3.500000000000000E+00
 curtor      =   1.038123580200000E-09
 curpol      =   0.000000000000000E+00
 gamma       =   0.000000000000000E+00
 Nfp         =         1
 Nvol        =         1
 Mpol        =         10
 Ntor        =         1
 Lrad        =                       8
 tflux       =   3.033538489388870E-01  1.000000000000000E+00
 pflux       =   0.000000000000000E+00 -2.040878894181875E-01
 helicity    =   2.036396095021530E-02  1.559429589793997E-03
 pscale      =   1.000000000000000E-03
 Ladiabatic  =         0
 pressure    =   1.000000000000000E+00  0.000000000000000E+00
 adiabatic   =   1.000000000000000E+00  0.000000000000000E+00
 mu          =   1.000000000000000E+00 -1.606028675492643E-04
 Lconstraint =         1
 pl          =                       0                      0                      0
 ql          =                       0                      0                      0
 pr          =                       0                      0                      0
 qr          =                       0                      0                      0
 iota        =   0.000000000000000E+00  1.400000000000000E-01  
 lp          =                       0                      0                      0
 lq          =                       0                      0                      0
 rp          =                       0                      0                      0
 rq          =                       0                      0                      0
 oita        =   0.000000000000000E+00  1.400000000000000E-01 
 mupftol     =   1.000000000000000E-12
 mupfits     =       128
 Rac         =   1.650000000000000E+00  0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00
 Zas         =   0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00
 Ras         =   0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00
 Zac         =   0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00  0.000000000000000E+00
[...]

Issue #183, opened 2022-03-29
Handle VMEC Boundary inputs that do not list each boundary mode separately


dpanici:
Apparently in VMEC the boundary modes can be given with the following logic, which we currently do not support converting in DESC. Also, this could be useful to add as an input option in our own input files

   ```
 rbc(-22:22,0) = 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
                    0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
                    0.9752041013880218, -0.693363161259351, 0.3158097406419154,
                    -0.1336173987189171, 0.04250319143001195, -0.0017939240980503425,
                    -0.011073301883956194, 0.010402231681464624, -0.005115714919129545,
                    0.0001348021675985992, 0.002564506738780156, -0.0028903931958533633,
                    0.0018064368511084795, -0.00023521761858271873, -0.0007086499231920249,
                    0.0011111467874450785, -0.0005442993972098523, -1.2982368610783194e-05,
                    0.0007834476816723562, 2.2221180650775065e-06, -4.416136866781143e-05,
                    8.239232451572815e-05, -0.00010136354765774249
```

f0uriest:
A possible generalized solution would be to use a fortran namelist reader like https://f90nml.readthedocs.io/en/latest/ for all the vmec reading/writing stuff (and spec too)

dpanici:
@MothVine the f90 namelist reader we mentioned ^

dpanici:
#431 improves this a bit but we should just use f90nml as it takes a lot of logic to deal with the flexibility that the namelists allow.

improvements that need to be made that could be resolved with f90nml:
- [ ] handling duplicate inputs that are on the same line (i.e.
 RBC(0,0) = 1. , ZBS(1,1) = 1.
 RBC(1,1) = 1. , ZBS(1,1) = 1.
currently DESC will not realize the 2nd line has a duplicate)

dpanici:
Also, make sure that we can handle VMEC input files with `INDATA` and `END` with different cases, since they are not always upper case

dpanici:
@dpanici I did this sometime, dig up the code for it

PR #1816, opened 2025-07-12
Change most of vmec input file parsing to f90nml

### LINKED ISSUE #1902: `write_vmec_input` will create a vmec input file that `Equilibrium.from_input_file` will not correctly handle
We assume the current profile in vmec inputs is current derivative, yet when we write a VMEC input file we save as `power_series_I` which is the current, so DESC cannot correctly read the vmec input file we make.

We probably should just make the input reader support current profiles from VMEC instead of just current deriv profiles  (at least `power_series_I` )

dpanici:
work towards #183 
- [x] convert boundary part to use f90nml, right now I have not bc discerning the modenumbers from just the f90nml-read arrays of RBC and ZBS is not straightforward and I think has  edge cases that could cause it to fail (i.e. if we always assume a [M, N] RBC array is such that the modenumbers are [-M//2, M//2] and [-N//2, N//2], this assumption fails if there are potentially ragged arrays or asymmetric arrays (like if one specifies only the m=1 n=0 m=1 n=1  m=1 n=2 modes, how do we discern this from something that has like m=1 n=-1 m=1 n=0 m=1 n=1? might not be an issue but I did not use much effort to figure it out)
- [x] resolves #1902
- [ ] potentially address #224 (lower priority)
- [ ] read in splines (also lower priority as there are many spline types in VMEC and other profile types, I'd prefer tokeep as just power series)

### DIFF
--- a/desc/input_reader.py
+++ b/desc/input_reader.py
@@ -1101,7 +1101,7 @@ class InputReader:
                 )
             )
         ctype = vmec_indata.get("PCURR_TYPE", "power_series")
-        if ctype.lower() != "power_series":
+        if not ctype.lower() in ["power_series", "power_series_i"]:
             warnings.warn(
                 colored(
                     "current is not a power series! DESC can only read power series"
@@ -1135,7 +1135,11 @@ class InputReader:
         # read current
         curr_tor = vmec_indata.get("CURTOR", None)
         AC = np.atleast_1d(vmec_indata.get("AC", np.array([0.0]))).astype(float)
-        ls = np.arange(0, AC.size) * 2
+        ls = (
+            np.arange(0, AC.size) * 2
+            if ctype.lower == "power_series"
+            else np.arange(1, AC.size + 1) * 2
+        )
         inputs["current"] = np.vstack([ls, AC]).T
 
         # axis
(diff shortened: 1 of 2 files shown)


dpanici:
Example of why idt we can safely use f90nml for reading in boundary coefficients
two simplified vmec inputs
```
&INDATA
  MPOL =    12
  NTOR =    0
  RBC( 0,0) =  3.510    ZBS( 0,0) =  0.000
  RBC( 0,1) =  1.000    ZBS( 0,1) =  1.470
  RBC( 0,2) =  0.106    ZBS( 0,2) = -0.160
```

```python
import f90nml
nml = f90nml.read("input.VMEC1")
print(nml["indata"]["RBC"])
```
yields
```python
[[3.51], [1.0], [0.106]]
```
ok this is fine, so it is a list of lists, the first index corresponds to the toroidal modes (hence why it is only length one, corresponds to the 0th index so we know it had only N=0 modes)

but if we add a single toroidal mode:
```
&INDATA
  MPOL =    12
  NTOR =    0
  RBC( 0,0) =  3.510    ZBS( 0,0) =  0.000
  RBC( 0,1) =  1.000    ZBS( 0,1) =  1.470
  RBC( 0,2) =  0.106    ZBS( 0,2) = -0.160
  RBC( -1,0) =  0.1
```
and run the same python script, we get

```python
[[0.1, 3.51], [None, 1.0], [None, 0.106]]
```

if we add a toroidal mode with the opposite sign modenumber we get

```
&INDATA
  MPOL =    12
  NTOR =    0
  RBC( 0,0) =  3.510    ZBS( 0,0) =  0.000
  RBC( 0,1) =  1.000    ZBS( 0,1) =  1.470
  RBC( 0,2) =  0.106    ZBS( 0,2) = -0.160
  RBC( 1,0) =  0.1
```
and run the same python script, we get

```python
[[3.51, 0.1], [1.0], [0.106]]
```

and the nail in the coffin for me is

```
&INDATA
  MPOL =    12
  NTOR =    0
  RBC( 0,0) =  3.510    ZBS( 0,0) =  0.000
  RBC( 0,1) =  1.000    ZBS( 0,1) =  1.470
  RBC( 0,2) =  0.106    ZBS( 0,2) = -0.160
  RBC( -2,4) =  0.1
```
yields
```python
[[None, None, 3.51], [None, None, 1.0], [None, None, 0.106], [None], [0.1]]
```

I guess we could do something like checking for the longest array size among them and use that to then deduce what the n modes are, and then in the cases where the length is less than the longest we just automatically assign the largest number, but this is just not something I understand fully nor have I worked out all the edge cases, and so I would much prefer to keep our current method which feels less error prone.

f0uriest:
it looks like it might store the start index as a separate variable? 
https://f90nml.readthedocs.io/en/latest/usage.html#f90nml.namelist.Namelist.start_index

dpanici:
> it looks like it might store the start index as a separate variable? https://f90nml.readthedocs.io/en/latest/usage.html#f90nml.namelist.Namelist.start_index

Does not seem to be working when I load in a vmec namelist for some reason

https://github.com/marshallward/f90nml/issues/178

PR #1826, opened 2025-07-30
Ku/partialsum

unalmis:


desc/equilibrium/coords.py
```diff
@@ -503,6 +503,132 @@ def fixup(x, *args):
     return out
 
 
+def _L_partial_sum(L, L_lmn):
+    """Convert FourierZernikeBasis to set of Fourier series.
+
+    TODO(#1243) Do proper partial summation once the DESC
+    DESC basis are improved to store the padded tensor product basis.
+    https://github.com/PlasmaControl/DESC/issues/1243#issuecomment-3131182128.
```
unalmis, line 405: @PlasmaControl/desc-dev in #1508 header it says "figure out how to do FourierZernike". Basically until the basis is padded https://github.com/PlasmaControl/DESC/issues/1243#issuecomment-3131182128 there is no efficient implementation without loops.

unalmis:
As this pull request made no changes to the benchmarks that show speed improvement, we can conclude those benchmarks are useless and misleading.

dpanici:
Do you have a script we can profile on master and here to see the speed improvement you imply?

> As this pull request made no changes to the benchmarks that show speed improvement (see previous benchmark run), we can conclude those benchmarks are useless and misleading.

unalmis:
> Do you have a script we can profile on master and here to see the speed improvement you imply?
> 
> 
> 
> > As this pull request made no changes to the benchmarks that show speed improvement (see previous benchmark run), we can conclude those benchmarks are useless and misleading.
> 
> 
> 
> 

no i mean click to github actions benchmark from the previous run. there will be a bunch of green speed improvement according to DESC benchmarks. however those are  all wrong because this PR doesn't change any of the code those benchmarks run

unalmis:
It's the same issue on the laplace solver PR and your recent https://github.com/PlasmaControl/DESC/pull/1828#issuecomment-3138046359. the benchmarks are just noise

dpanici:
weird, it really should not be making htem green if the change is within the std dev... I agree they are quite noisy but I guess their use isif we completely tank a given function it should come up from the noise. Would be nice if we could run them on specific architectures with dedicated resources though.

desc/equilibrium/equilibrium.py
```diff
@@ -1339,55 +1345,21 @@ def _get_rtz_grid(
...
-            coords=flux_coords,
-            inbasis=("rho", "theta_PEST", "zeta"),
-            outbasis=("rho", "theta", "zeta"),
-            params=self.params_dict if L_lmn is None else {"L_lmn": L_lmn},
-            tol=tol,
-            maxiter=maxiter,
-            full_output=full_output,
-            **kwargs,
+        return _map_clebsch_coordinates(
+            iota, alpha, zeta, L_lmn, L, theta0, period, tol, maxiter, **kwargs
```
unalmis: this is to avoid circular imports

desc/grid.py
```diff
@@ -1334,6 +1334,10 @@ def _create_nodes(  # noqa: C901
             # if custom zeta used usually safe to assume its non-uniform so no fft
             self._fft_toroidal = not endpoint
         elif zeta is not None:
+            errorif(
+                np.any(np.asarray(zeta) > zeta_period),
+                msg="LinearGrid defined on 1 field period.",
```
dpanici:
```suggestion
                msg="LinearGrid should be defined on 1 field period.",
```

desc/objectives/_fast_ion.py
```diff
@@ -305,50 +314,57 @@ def compute(self, params, constants=None):
         return constants["transforms"]["grid"].compress(data[self._key])
 
     def _build_spline(self, use_jit=True, verbose=1):
+        Y_B = self._hyperparam.pop("Y_B")
+        num_transit = self._hyperparam.pop("num_transit")
+        num_quad = self._hyperparam.pop("num_quad")
+        del self._constants["X"]
```
dpanici, line 320: why the deletion?

unalmis: it's just not used so I wanted to delete it.

desc/equilibrium/coords.py
```diff
@@ -123,18 +131,9 @@ def map_coordinates(  # noqa: C901
...
-                L_basis=eq.L_basis,
-                guess=guess[:, 1] if guess is not None else None,
-                period=period[1] if period is not None else np.inf,
-                tol=tol,
-                maxiter=maxiter,
-                full_output=full_output,
-                **kwargs,
-            )
+            rho, alpha, zeta = coords.T
+            coords = jnp.column_stack([rho, alpha + iota * zeta, zeta])
```
f0uriest: is this still correct if `zeta != phi`? i think we usually define `alpha = theta_PEST - iota*phi`

docs/notebooks/tutorials/EffectiveRipple.ipynb
f0uriest, line 1: did anything actually change here? The only difference I see is a warning about unequal NFP

desc/equilibrium/equilibrium.py
```diff
@@ -1339,55 +1345,21 @@ def _get_rtz_grid(
             self, radial, poloidal, toroidal, coordinates, period, jitable, **kwargs
         )
 
-    def compute_theta_coords(
-        self, flux_coords, L_lmn=None, tol=1e-6, maxiter=20, full_output=False, **kwargs
+    @staticmethod
+    def _map_clebsch_coordinates(
```
f0uriest, line 1351: does this need to be a method here? I'd vote for just using the function directly where needed

desc/equilibrium/coords.py
```diff
@@ -376,58 +375,100 @@ def fixup(x, *args):
...
+        grid.M > L.basis.M,
+        ResolutionWarning,
+        msg="Poloidal grid resolution is higher than necessary for coordinate mapping.",
+    )
+    warnif(
+        grid.M < L.basis.M,
+        ResolutionWarning,
+        msg="High frequency lambda modes will be truncated in coordinate mapping.",
+    )
+    lmbda_minus_iota_omega = L.transform(L_lmn)
```
f0uriest: why is this lambda-iota*omega and not just lambda?

desc/equilibrium/coords.py
```diff
@@ -376,58 +375,100 @@ def fixup(x, *args):
...
+        Shape (num zeta, )
     L_lmn : jnp.ndarray
-        Spectral coefficients for lambda.
-    L_basis : Basis
-        Spectral basis for lambda.
-    guess : jnp.ndarray
-        Shape (k, ).
+        Spectral coefficients for λ.
+    L : Transform
+        Transform for λ built on coordinates [ρ, θ, ζ].
```
f0uriest: what is theta here?

also I'd vote to name this `L_transform` or something to be more explicit and not confuse with resolution parameter `L`

desc/equilibrium/coords.py
```diff
@@ -436,71 +477,52 @@ def _map_clebsch_coordinates(
...
-                jac=jacfun,
-                args=p,
-                fixup=fixup,
-                tol=tol,
-                maxiter=maxiter,
-                full_output=full_output,
-                **kwargs,
-            )
+    @partial(jnp.vectorize, signature="(),(),(m)->()")
+    def vecroot(theta0, alpha, c_m):
```
f0uriest: I think you'll still have the issue @YigitElma mentioned about recompilation here. The fix would be to make rootfun/jacfun/vecroot global (private) functions rather than define them locally within this function

desc/equilibrium/coords.py
```diff
@@ -436,71 +477,52 @@ def _map_clebsch_coordinates(
...
     # noqa: D202
 
     # Root finding for θₖ such that r(θₖ) = αₖ(ρ, θₖ, ζ) − α = 0.
-    def rootfun(theta, alpha, rho, zeta, iota):
-        nodes = jnp.array([rho.squeeze(), theta.squeeze(), zeta.squeeze()], ndmin=2)
-        A = L_basis.evaluate(nodes)
-        lmbda = A @ L_lmn
-        alpha_k = theta + lmbda - iota * zeta
-        return _fixup_residual(alpha_k - alpha, period).squeeze()
+    def rootfun(theta, alpha, c_m):
```
f0uriest: is alpha here really theta_PEST?

desc/equilibrium/coords.py
```diff
@@ -436,71 +477,52 @@ def _map_clebsch_coordinates(
...
-    else:
-        theta = vecroot(guess, alpha, rho, zeta, iota)
 
-    out = jnp.column_stack([rho, jnp.atleast_1d(theta.squeeze()), zeta])
-    if full_output:
-        return out, (res, niter)
-    return out
+    c_m, modes = _partial_sum(L, L_lmn, None, None, iota)
+    c_m = c_m[:, jnp.newaxis]
+    alpha = alpha[:, jnp.newaxis] + iota[:, jnp.newaxis, jnp.newaxis] * zeta
```
f0uriest: is this really theta_PEST?

docs/notebooks/tutorials/EffectiveRipple.ipynb
unalmis: Just rerunning notebook to track history /effect on optimization of changes to source code.

desc/equilibrium/equilibrium.py
unalmis: It has to be an object method to avoid circular imports.

desc/equilibrium/coords.py
unalmis: In `ku/nufft` I jitted the outer function as well.

desc/equilibrium/coords.py
unalmis: No I fixed the labels in `ku/nufft`. In terms of returned result the values are correct

desc/equilibrium/coords.py
unalmis: No, it is not supposed to be theta_PeST

desc/equilibrium/coords.py
unalmis: The root-finding from desc coords to clebsch coords involves both lambda and omega

desc/equilibrium/coords.py
unalmis: DESC theta. This is currently a private method, eventually once 1243 is done, the passed in parameter will be changed to just the spectral coefficients (no transform) and we can then do proper partial summation

desc/equilibrium/coords.py
unalmis: This assumes `zeta=Phi`. I have added comment to address this.

desc/equilibrium/coords.py
unalmis: I hope the new labels clarify.

desc/equilibrium/coords.py
unalmis: I'll rename to what you want on `ku/nufft`. It's not worth running GitHub CI again for a name change for private function

desc/equilibrium/equilibrium.py
```diff
@@ -1334,61 +1341,67 @@ def _get_rtz_grid(
...
+            zeta,
+            L_lmn,
+            lmbda,
+            theta0,
+            tol=tol,
+            maxiter=maxiter,
+            **kwargs,
+        )
+
+    def _compute_iota_under_jit(self, rho, params=None, profiles=None, **kwargs):
```
unalmis, line 1374: this is used in `map_coordinates` and `Bounce2D.compute_theta`

unalmis:
I had to pull changes from `ku/nuffts` to address Rory's comments. So the codecov lowered by 1%. It's fine.

desc/equilibrium/coords.py
```diff
@@ -115,33 +124,27 @@ def map_coordinates(  # noqa: C901
...
-                period=period[1] if period is not None else np.inf,
-                tol=tol,
-                maxiter=maxiter,
-                full_output=full_output,
-                **kwargs,
-            )
+                iota = eq._compute_iota_under_jit(coords, params, profiles, **kwargs)
+            rho, alpha, zeta = coords.T
+            omega = 0  # TODO(#568)
+            coords = jnp.column_stack([rho, alpha + iota * (zeta + omega), zeta])
```
rahulgaur104, line 139: If zeta is the generalized toroidal angle, don't we assume zeta = phi + Omega where phi is the cylindrical toroidal angle?
So shouldn't theta_PEST = alpha + iota * (zeta - omega)?

desc/equilibrium/coords.py
```diff
@@ -376,131 +382,147 @@ def fixup(x, *args):
...
+        ResolutionWarning,
+        msg="Poloidal grid resolution is higher than necessary for coordinate mapping.",
+    )
+    warnif(
+        grid.M < lmbda.basis.M,
+        ResolutionWarning,
+        msg="High frequency lambda modes will be truncated in coordinate mapping.",
+    )
+    lmbda_minus_iota_omega = lmbda.transform(L_lmn)
+    lmbda_minus_iota_omega = (
```
rahulgaur104, line 447: So meshgrid reshape is reshaping from rtz to rzt?
What is the division and multiplication by 2 doing? Can you explain?

desc/equilibrium/coords.py
```diff
@@ -376,131 +382,147 @@ def fixup(x, *args):
...
-    coords,
     iota,
+    alpha,
+    zeta,
     L_lmn,
-    L_basis,
-    guess=None,
-    period=np.inf,
+    lmbda,
+    theta0=None,
```
rahulgaur104: theta0 is a very bad variable name. theta0 is dangerously close to theta_0 that people in the gyrokinetics communit use in flux tube codes to define the location of vanishing integrated magnetic shear. 
Why did you change it from guess? I strongly recomment you change it back. guess is great because it immediately tells you what it is. theta0 requires user to figure out what is happening in the rest of the code.

rahulgaur104: As a rule of thumb, please don't make unnecessary changes to variable names.

desc/equilibrium/coords.py
```diff
@@ -376,131 +382,147 @@ def fixup(x, *args):
...
     kwargs : dict, optional
         Additional keyword arguments to pass to ``root_scalar`` such as ``maxiter_ls``,
         ``alpha``.
 
     Returns
     -------
-    out : ndarray
-        Shape (k, 3).
-        DESC computational coordinates [ρ, θ, ζ].
-    info : tuple
```
rahulgaur104, line 451: Why did you remove info? This is not a good coding practice. This is basically passive encryption.

rahulgaur104: What happens if root finding fails for some reason and I want to debug it.

rahulgaur104: If the residual is calculated elsewhere, ignore this comment.

desc/integrals/_bounce_utils.py
```diff
@@ -79,8 +79,8 @@ def bounce_points(pitch_inv, knots, B, dB_dz, num_well=None):
 
     """
     intersect = polyroot_vec(
```
rahulgaur104, line 81: This routine find the intersection of lines of constant pitch with the magnetic field B, i.e., the bounce points.
B is an array with a shape rho, alpha, 1/lambda (inv_pitch), something, num_wells.
Is that right?

desc/integrals/bounce_integral.py
```diff
@@ -145,7 +145,7 @@ class Bounce2D(Bounce):
     Notes
     -----
     Magnetic field line with label α, defined by B = ∇ψ × ∇α, is determined from
-      α : ρ, θ, ζ ↦ θ + λ(ρ,θ,ζ) − ι(ρ) [ζ + ω(ρ,θ,ζ)]
+      α : ρ, θ, ζ ↦ θ + Λ(ρ,θ,ζ) − ι(ρ) [ζ + ω(ρ,θ,ζ)]
```
rahulgaur104, line 148: I am confused again by this equation. Specifically the + sign before omega.
Is the convention zeta = phi + omega or zeta = phi - omega, where phi is the cylindrical angle and zeta is the generalized angle.

rahulgaur104 (commented):
Mostly minor comments. Addressing them will help future development/ developers.

desc/equilibrium/coords.py
unalmis: Phi = zeta + omega, and theta_PEST = theta + lambda. The left hand side of these relations are defined quantities. Phi is the cylindrical toroidal angle and theta_PEST is the angle where the field lines are straight in the (theta_PEST, Phi) plane. When we mention generalizing angles, we refer to changing the meaning of the angles "zeta" and "theta". These relations must still hold and hence the stream functions must negate the change in zeta and theta.

unalmis: alpha = theta_PEST - iota phi

desc/equilibrium/coords.py
unalmis: 1. This is a new function, so nothing is an unnecessary change because nothing is a change; it is new.[^1]
2. `theta0` is the initial guess in the root finding for `theta`. I will change it to `guess` as you requested. Whatever the name is, the public documentation next to the parameter `theta0: jnp.ndarray : Optional initial guess for the computational coordinates.` is there. So no one has to read code to figure out what it should be.

[^1]: The function `_map_clebsch_coordinates` is private and not used anywhere in DESC. If there is external code that was using this function, we are not responsible for breaking API convention.

desc/equilibrium/coords.py
unalmis: That is correct for the reshape.

One way to think about `rfft` is as follows, it will give you the $c_n$ for $f(\theta) = \sum_n^{N} c_n b_n(n \theta)$ where  $b_n$ is some basis function. This basis function is $b_n(n \theta) = \exp(j n \theta)/2$ for $n \neq 0$ and $n \neq N$. For $n = 0$, and $n = N$ you have to be more careful. We multiply etc. so that we have the coefficients of the familiar spectral Fourier basis.

desc/equilibrium/coords.py
unalmis: This is a new function, see https://github.com/PlasmaControl/DESC/pull/1826#discussion_r2275331172, so I disagree with the statement about coding practice. Whatever code users have that uses root finding, nothing has changed. They can still get their `info` tuple.

In this new function, I have not added functionality to return auxiliary information about the root finding because that is impossible --- functions that are decorated with `jnp.vectorize`, such as this one, must return arrays. `info` is not an array.

unalmis: I am not convinced that this root finding can fail by the way.

desc/integrals/bounce_integral.py
unalmis: The latter. One can control +f search desc/compute/_core.py for `name="phi"` and `name="theta_PEST"` to see the definitions.

desc/integrals/_bounce_utils.py
unalmis: I put the shape of all my stuff in the public documentation string (docstring for short) because I find it helpful. Here is the definition of B and its shape in that docstring. 

```python
B : jnp.ndarray
        Shape (..., N - 1, B.shape[-1]).
        Polynomial coefficients of the spline of B in local power basis.
        Last axis enumerates the coefficients of power series. Second to
        last axis enumerates the polynomials that compose a particular spline.
```
When you see `Shape (..., N - 1, B.shape[-1])`, in Python, that means the code is making the following contract with you, the developer:

The last two axes need to have that shape. All the leading axes ``...`` don't matter, the code is agonstic to them. If there are leading axes, the code will simply perform whatever it does to the last two axes in a vectorized manner. That is the same contract that `numpy` makes with you. Numpy calls this contract its "broadcasting conventions".

Less experienced developers will mess up this broadcasting contract, and instead may promise broadcasting on the trailing axes of the form `Shape (N - 1, B.shape[-1], ...)`. You should NEVER do that. If you broadcast on the trailing axes like that you break from numpy convention, and both your code and the user-facing code will have to have a million transposes and reshapes to get their stuff working with what you wrote.

desc/equilibrium/coords.py
unalmis: This function should be fine, but you are welcome to fix `map_coordinates`.

unalmis:
I addressed the comments. I think at developer meeting you all wanted me to add more of my pull request comments to code. Please make a new pull request and add whichever of my comments into the code that you would like to add.

desc/integrals/_bounce_utils.py
rahulgaur104: So what is the shape of B for this problem? 
(..., B.shape[-3], B.shape[-2], B.shape[-1]) is an answer but I meant in terms of coordinates. Docstring does not have the shape here.

unalmis: N is documented [here](https://github.com/PlasmaControl/DESC/blob/fd17cc6fe5b18419cd44ca278cf50019c20e90dc/desc/integrals/_bounce_utils.py#L42) as the number of knots. B.shape[-1] is documented [here](https://github.com/PlasmaControl/DESC/blob/fd17cc6fe5b18419cd44ca278cf50019c20e90dc/desc/integrals/_bounce_utils.py#L47) as the number of coefficients in each polynomial of the spline. If you have a cubic spline that is 4. The last axis of `pitch` is documented [here](https://github.com/PlasmaControl/DESC/blob/fd17cc6fe5b18419cd44ca278cf50019c20e90dc/desc/integrals/_bounce_utils.py#L39) to be the number of pitch angles. The returned shape is documented [here](https://github.com/PlasmaControl/DESC/blob/fd17cc6fe5b18419cd44ca278cf50019c20e90dc/desc/integrals/_bounce_utils.py#L71) to have its last two axes be number of pitch angles, number of wells, respectively.

PR #1828, opened 2025-07-30
Fix bug in FourierRZCurve from_values when numpy array is passed

dpanici:


unalmis (commented):
Approving because the benchmarks are junk :)

unalmis (approved):

unalmis (approved):

YigitElma (approved):

PR #1830, opened 2025-07-31
don't close over field in field_line_integrate

### LINKED ISSUE #1829: Long compile times for field trace with high res `SplineMagneticField`
When using a field from an mgrid file, the resulting `SplineMagneticField` has a lot of large arrays as attributes, which then can cause long compilation times when performing calculations. The most aggregious example I have found is field line tracing, which can take long compile times and often returns [warnings about constant folding](https://github.com/jax-ml/jax/issues/30642).

As an example and a possible way to lower these compile times, using this mgrid file and eq [here](https://drive.google.com/drive/folders/1c0PYURB4Q7YBQaMwWqX752GmbRGfp-TL?usp=sharing):

This script takes 23s (on a della cpu node)
```python
import os
# os.environ["JAX_CAPTURED_CONSTANTS_REPORT_FRAMES"]="5" # this will show where the constants are
# os.environ['XLA_FLAGS'] = "--xla_disable_hlo_passes=constant_folding" # this disables constant folding
from desc.io import load
from desc.plotting import * 
import numpy as np
from desc.magnetic_fields import SplineMagneticField

extcur = [277E+03,-240.00E+03]
field = SplineMagneticField.from_mgrid(
    "mgrid_8.nc", extcur=extcur,method="linear"
)
eq = load("eq_with_surf_qfm_axis_from_field_free_bdry_solved_LMN10_quadgrid_force.h5")
r = eq.compute("R",grid=LinearGrid(rho=np.linspace(0,1,10)))["R"]
z = eq.compute("Z",grid=LinearGrid(rho=np.linspace(0,1,10)))["Z"]

fig,ax=plot_surfaces(eq,theta=0,figsize=(8,8),phi=6)
import time
t0 = time.time()
poincare_plot(field,r,z,ax=ax,phi=6,size=1)#,ode_chunk_size=1)

print(time.time()-t0)
# with constant folding
```

while the following took 15s on the same node. Maybe it is worth using this flag? or at least making it known when field line tracing occurs. 
```python
import os
# os.environ["JAX_CAPTURED_CONSTANTS_REPORT_FRAMES"]="5" # this will show where the constants are
os.environ['XLA_FLAGS'] = "--xla_disable_hlo_passes=constant_folding" # this disables constant folding
from desc.io import load
from desc.plotting import * 
import numpy as np
from desc.magnetic_fields import SplineMagneticField

extcur = [277E+03,-240.00E+03]
field = SplineMagneticField.from_mgrid(
    "mgrid_8.nc", extcur=extcur,method="linear"
)
eq = load("eq_with_surf_qfm_axis_from_field_free_bdry_solved_LMN10_quadgrid_force.h5")
r = eq.compute("R",grid=LinearGrid(rho=np.linspace(0,1,10)))["R"]
z = eq.compute("Z",grid=LinearGrid(rho=np.linspace(0,1,10)))["Z"]

fig,ax=plot_surfaces(eq,theta=0,figsize=(8,8),phi=6)
import time
t0 = time.time()
poincare_plot(field,r,z,ax=ax,phi=6,size=1)#,ode_chunk_size=1)

print(time.time()-t0)
# with constant folding
```
This is something others [have seen as well](https://github.com/jax-ml/jax/issues/21300), there are some ways that try to avoid this when jit itself is called, but they seem a bit hacky. Is there maybe a way we can re-configure the SplineMagneticField class to avoid this?

dpanici:
Don't close over `field` to avoid long compile times due to constant folding when the field has large array attributes.

This helps dramatically reduce compile times when constant folding would be slow, and seems to have no effect on runtimes in cases where there would not be appreciably slow constant folding occuring, so it seems to be a harmless change.

- [x] check effect on GPU (before, running on gpu took a surprisingly large amount of memory, even if it could run easily on the cpu with less memory) - this PR also improves speed for GPU, have not profiled memory.

Resolves #1829

### DIFF
--- a/desc/magnetic_fields/_core.py
+++ b/desc/magnetic_fields/_core.py
@@ -2646,6 +2646,7 @@ def field_line_integrate(
     @jit
     def odefun(s, rpz, args):
         rpz = rpz.reshape((3, -1)).T
+        field = args[0]
         r = rpz[:, 0]
         br, bp, bz = (
             scale
@@ -2684,6 +2685,7 @@ def field_line_integrate(
         saveat=saveat,
         max_steps=maxstep * len(phis),
         dt0=min_step_size,
+        args=(field,),
         **kwargs,
     ).ys
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Bug Fixes
 
 Backend
 - Significant changes to how DESC handles static attributes during JIT compilation. Going forward if any class/object has attributes that should be treated as static by `jax.jit`, these should be declared at the class level like `_static_attrs = ["foo", "bar"]`. Generally, non-arraylike attributes such as functions, strings etc should be marked static, as well as any attributes used for control flow. Previously this was done automatically, but in a way that caused a lot of performance bugs and unnecessary recompilation. These changes have been implemented for all classes in the `desc` repository, but if you have custom objectives or other local objects that subclass from `desc` you may need to add this yourself. JAX error messages usually do a good job of alerting you to things that need to be static, and feel free to open an issue with `desc` if you have any questions.
+- No longer closes over the field in ``desc.magnetic_fields._core.field_line_integrate``, which can dramatically reduce compile times when the field being traced has large size attributes (for example, when using a ``desc.magnetic_fields._core.SplineMagneticField`` object).
 
 v0.14.2
 -------


dpanici:
> I was also thinking about making field line tracing work with equilibrium too. I think this is a small step towards that.

would the next step be having `params` be also passed into `odefun` through `args`?

desc/magnetic_fields/_core.py
```diff
@@ -2684,6 +2685,7 @@ def default_terminating_event_fxn(state, **kwargs):
         saveat=saveat,
         max_steps=maxstep * len(phis),
         dt0=min_step_size,
+        args=(field,),
```
YigitElma, line 2688: Yes, probably adding `params` to here and also making sure that `params` is not None like `params = field.params.dict if params is None else params` should do it

YigitElma (approved):

unalmis (approved):

PR #1833, opened 2025-08-04
Add objective to optimize particle orbits

dpanici:
Waiting for #2012


dpanici:
Add ABC for particle tracing that just implements to the point of the trajectory result

YigitElma:
Notes for future:
- The objective should definitely use #2012. Better to branch from that PR
- For larger number of particles, the line fitting step may become a bottle as it is currently implemented. Jax `polyfit`function uses an svd based algorithm, and even though we fit 2 parameters, it may be an over kill. A faster linear algebra routine can be used.
- Should the objective return a slope for each particle? Can we clump the similar energy particles into packets? Does this makes sense?
- `steps=True` is faster for tracing but it returns an array of size `max_steps`, then this increases the size of the linear fit matrix. Probably having couple hundred points is enough. But what if particle is lost before it stores enough trajectory data?

YigitElma:
I will open a new PR later with a better objective.

Issue #1692, opened 2025-04-16
Clean Up Errors and Warnings


dpanici:
- remove `colored` warnings
- replace warnings with `equinox` warnings
- Make new classes for specific warning/error types

#1638

Issue #1388, opened 2024-11-18
Move some general interpolation functions into `interpax`


dpanici:
- [x] `desc/integrals/_interp_utils.py` https://github.com/PlasmaControl/DESC/blob/ku%2Fbounce/desc%2Fintegrals%2F_interp_utils.py

- [x] `desc/integrals/basis.py` https://github.com/PlasmaControl/DESC/blob/ku%2Fbounce/desc%2Fintegrals%2Fbasis.py

PR #1834, opened 2025-08-04
Ku/nufft

### LINKED ISSUE #1294: Nonuniform fast transforms

unalmis:
Resolves #1294


desc/integrals/_interp_utils.py
```diff
@@ -186,8 +154,8 @@ def ifft_non_uniform(xq, a, domain=(0, 2 * jnp.pi), axis=-1, _modes=None):
         n = a.shape[axis]
         _modes = jnp.fft.fftfreq(n, (domain[1] - domain[0]) / (2 * jnp.pi * n))
     a = jnp.moveaxis(a, axis, -1)
-    vander = jnp.exp(-1j * _modes * (xq - domain[0])[..., jnp.newaxis])
-    return jnp.linalg.vecdot(vander, a)
+    vander = jnp.exp(1j * _modes * (xq - domain[0])[..., jnp.newaxis])
+    return (vander * a).sum(axis=-1)
```
unalmis, line 158: Don't worry these are identical mathematically.

tests/test_integrals.py
```diff
@@ -1620,11 +1618,29 @@ def test_bounce2d_checks(self):
...
         _, _ = bounce.plot_theta(l, m, show=False)
 
-        # make sure tests pass when spline=True
+        # check for consistency with different options
+        num_nufft = bounce.integrate(
+            integrand=TestBounce._example_numerator,
+            pitch_inv=pitch_inv,
+            data=_data,
+            points=points,
+            # ~1% of the integrals differ significantly at lower epsilon.
```
unalmis, line 1630: In a few years time, I suspect finufft will fix some obscure bug (like the one they fixed around June this year in v.2.4.0) and this issue will vanish.

dpanici:
TODO if not done already
- [ ]  Make nuffts opt-in only by default
  - [ ]  basically, make it so if not installed, things work as is
- [ ] add a clear warning when nuffts are enabled in the objective (and in compute) that it requires an optional dependency that is not completely stable yet (and for objective case, that the gradients are wrong, maybe just make it an error when it is enabled for an objective until we know they fix their issues)
- [ ] can we try checking 
  - how jax handles when jax-finufft is only installed for cpu and not gpu
- [ ] make issue on JAX-finufft for the incorrect jacobian computation
- [ ] Can we also put a comment here on the current potential issues (like if it is hard to install on cpu, error in jacobian, etc) and then make that a standing issue so it is documented until resolved?

tests/baseline/test_plot_gammac.png
unalmis, line 1: This change is purely from increasing the default knot density to round up to an integer multiple of NFP instead of rounding down. If you revert that change the old plot on `ku/fix_1529` is recovered: 

Image

CHANGELOG.md
```diff
@@ -18,6 +18,11 @@ New Features
 - Adds ``FourierXYCoil`` to compatible coils for ``CoilSetArclengthVariance`` objective.
 - Separated ``gamma_c`` calculation from ``Gamma_c``. User can also plot ``gamma_c`` using the ``plot_gammac`` function.
 
+Performance Improvements
+
+- [Partial summation in coordinate mapping](https://github.com/PlasmaControl/DESC/pull/1826).
+- [Nuffts in bounce integrals](https://github.com/PlasmaControl/DESC/pull/1834).
```
dpanici: note that they are only for compute currently and not for use in objective functions/optimizations

desc/compute/_neoclassical.py
```diff
@@ -78,12 +78,36 @@
...
+        you must specify ``nufft_eps=0`` if you intend to use automatic differentiation
+        to differentiate the computation. Performance will improve significantly once
+        jax-finufft fixes this.
+        """,
+    "spline": """bool :
+        Whether to use cubic splines to compute bounce points.
+        The recommended setting is ``True``.
+        """,
+    "_vander": """dict[str,jnp.ndarray] :
+        Precomputed transform matrices.
```
dpanici: which ones, specifically?

desc/objectives/_fast_ion.py
```diff
@@ -177,6 +190,7 @@ def __init__(
         num_pitch=64,
         pitch_batch_size=None,
         surf_batch_size=1,
+        nufft_eps=0,  # TODO: Set to 1e-7 once JAX-finufft fixes AD.
```
dpanici: add a warning in the init if nufft_eps > 0 that says it is not allowed for objective optimization, and then set nufft_eps=0

I want to make sure the user cannot access this bug of the AD

desc/objectives/_neoclassical.py
```diff
@@ -161,6 +175,7 @@ def __init__(
         num_pitch=51,
         pitch_batch_size=None,
         surf_batch_size=1,
+        nufft_eps=0,  # TODO: Set to 1e-6 once JAX-finufft fixes AD.
```
dpanici: warning for nufft_eps>0

desc/objectives/_fast_ion.py
dpanici: or at least a warning, if this is sometimes used as a method to just compute the objective value

desc/grid.py
```diff
@@ -710,11 +710,11 @@ def meshgrid_reshape(self, x, order):
             shape += (-1,)
         x = x.reshape(shape, order="F")
         # swap to change shape from trz/arz to rtz/raz etc.
-        x = jnp.swapaxes(x, 1, 0)
+        x = x.swapaxes(1, 0)
         newax = tuple(self.coordinates.index(c) for c in order)
         if vec:
             newax += (3,)
-        x = jnp.transpose(x, newax)
+        x = x.transpose(newax)
```
unalmis, line 717: so i can use this method with numpy arrays

desc/integrals/_bounce_utils.py
unalmis, line 1: Some of these changes were required to win the fight against codecov after adding warnings you guys wanted that can't be hit by github CI.

dpanici:
Install instructions: make current master the default (CPU+GPU), have separate tab for CPU+GPU (with finufft)

tests/test_neoclassical.py
```diff
@@ -16,11 +16,13 @@
 
 @pytest.mark.unit
 @pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_1d)
-def test_effective_ripple_2D():
+@pytest.mark.parametrize("nufft_eps", [0, 1e-6])
```
unalmis, line 19: You can set `num_transit` to 100 with nuffts and it doesn't affect runtime or memory of this test too much. without the nuffts you get OOM and takes forever.

desc/batching.py
```diff
@@ -34,7 +34,7 @@
 )
 
 from desc.backend import jax, jnp, scan, vmap
-from desc.utils import errorif
+from desc.utils import errorif, identity
 
 if jax.__version_info__ >= (0, 4, 16):
     from jax.extend import linear_util as lu
```
YigitElma, line 36: Unrelated but we can remove the if here. Our minimum jax version is already 0.4.24

desc/compute/_neoclassical.py
```diff
@@ -78,7 +78,17 @@
...
         """,
-    "spline": "bool : Whether to use cubic splines to compute bounce points.",
+    "nufft_eps": """float :
+        Precision requested for interpolation with non-uniform fast Fourier
+        transform (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
+        """,
+    "spline": """bool :
+        Whether to use cubic splines to compute bounce points.
+        """,
+    "_vander": """dict[str,jnp.ndarray] :
```
YigitElma, line 88: The underscore name feels a bit weird. If a user is giving it, it is clearly not private?

desc/objectives/_fast_ion.py
```diff
@@ -177,13 +182,31 @@ def __init__(
...
+        use_bounce1d=False,
         Nemov=True,
+        **kwargs,
     ):
+        try:
+            import jax_finufft  # noqa: F401
+        except ImportError:
+            warnif(
+                nufft_eps >= 1e-14,
+                msg=colored(
```
YigitElma: I prefer to keep the warnings/errors as plain text instead of colored. Yes, it is better looking but it causes problems, for example, when you try to suppress a warning based on a message. It fails to check the content of colored text.

desc/objectives/_fast_ion.py
```diff
@@ -114,11 +116,15 @@ class GammaC(_Objective):
         Number of flux surfaces with which to compute simultaneously.
         If given ``None``, then ``surf_batch_size`` is ``grid.num_rho``.
         Default is ``1``. Only consider increasing if ``pitch_batch_size`` is ``None``.
-    spline : bool
-        Set to ``True`` to replace pseudo-spectral methods with local splines.
+    nufft_eps : float
+        Precision requested for interpolation with non-uniform fast Fourier
+        transform (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
+    use_bounce1d : bool
```
YigitElma, line 121: If this is not really necessary, I prefer to keep this as is. Even in the description, you explain it as spline, so why not keep the name.

desc/objectives/_neoclassical.py
```diff
@@ -102,6 +104,8 @@ class EffectiveRipple(_Objective):
         A tighter upper bound than ``num_well=(Aι+B)*num_transit`` is preferable.
         The ``check_points`` or ``plot`` methods in ``desc.integrals.Bounce2D``
         are useful to select a reasonable value.
+
+        This is the most important paramter to specify for performance.
```
YigitElma:
```suggestion
        This is the most important parameter to specify for performance.
```

desc/objectives/_neoclassical.py
```diff
@@ -161,12 +172,30 @@ def __init__(
...
+        nufft_eps=1e-6,
+        use_bounce1d=False,
+        **kwargs,
     ):
+        try:
+            import jax_finufft  # noqa: F401
+        except ImportError:
+            warnif(
+                nufft_eps >= 1e-14,
+                msg=colored(
```
YigitElma: Same as above

docs/installation.rst
```diff
@@ -89,10 +89,11 @@ On Your Local Machine
                 # run jupyter
                 uv run --with jupyter jupyter lab
 
-    .. tab-item:: CPU+GPU
+    .. tab-item:: CPU+GPU (partial support)
 
         For GPU support, you must install the JAX library as discussed in `JAX installation docs <https://github.com/google/jax#installation>`__.
         For example, below are the instructions to install on compatible devices with an NVIDIA GPU.
+        These instructions install GPU support for a subset of the features in DESC.
```
YigitElma, line 96: I think you can just state what features are not covered by this since there are not many of them. Something like `if you want to have better performance for EffectiveRipple or any of the Bounce integral quantities, you might find the GPU installation instruction with FINUFFT useful`. This currently sounds like many things are not supported.

docs/installation.rst
```diff
@@ -89,10 +89,11 @@ On Your Local Machine
                 # run jupyter
                 uv run --with jupyter jupyter lab
 
-    .. tab-item:: CPU+GPU
+    .. tab-item:: CPU+GPU (partial support)
```
YigitElma, line 92: I think we should just keep it as is. For example, in libraries like MPI, they keep the instructions simple, but if you want to have CUDA-awareness, they have a subsection for it as `with CUDA support`. This case is very similar to that. Code runs without FINUFFT (just throws a warning), so you still have the full support. However, with these additional instructions, you can make certain things faster.

docs/installation.rst
```diff
@@ -104,7 +105,32 @@ On Your Local Machine
             pip install --editable .
 
         Note that on BSD systems, the ``sed`` command that replaces ``jax`` with ``jax[cuda12]``
-        in the ``requirements.txt`` file is ``sed -i '' '1 s/^jax/jax[cuda12]/' requirements.txt``
+        in the ``requirements.txt`` file is ``sed -i '' '1 s/^jax/jax[cuda12]/' requirements.txt``.
```
YigitElma, line 108: As I mentioned in previous PR, I still think just `pip install jax[cuda12]` is a way cleaner instruction and doesn't need any of the special considerations.

docs/installation.rst
```diff
@@ -104,7 +105,32 @@ On Your Local Machine
...
+
+        You may optionally install developer requirements if you want to run tests.
+
+        .. code-block:: sh
+
+            pip install -r devtools/dev-requirements.txt
+
+    .. tab-item:: CPU+GPU (with FINUFFT)
+
```
YigitElma, line 117: Should state what type of computations make use of this. Currently there are not that many, we can keep the list updated future.

docs/installation.rst
```diff
@@ -208,31 +234,34 @@ On computing clusters you must call ``module load anaconda`` to use conda (in so
...
             .. code-block:: sh
 
                 module load anaconda3/2024.10
-                conda create --name desc-env python=3.12 -y
-                conda activate desc-env
+                module load cudatoolkit/12.9
+
                 git clone https://github.com/PlasmaControl/DESC.git
                 cd DESC
+                conda create --name desc-env -c conda-forge 'python=3.12' 'fftw' 'gxx<12'
```
YigitElma, line 236: As done above, I prefer having a 3rd tab for FINUFFT support

docs/installation.rst
```diff
@@ -208,31 +234,34 @@ On computing clusters you must call ``module load anaconda`` to use conda (in so
...
+                conda activate desc-env
+
+                export CMAKE_PREFIX_PATH=$CONDA_PREFIX:$CMAKE_PREFIX_PATH
+                sed -i '1 s/^jax/jax[cuda12]/' requirements.txt
+                sed -i '/^jax-finufft/d' requirements.txt
+
                 pip install --editable .
+                pip install -Ccmake.define.JAX_FINUFFT_USE_CUDA=ON --no-binary=jax-finufft jax-finufft
+
+            Note that on BSD systems, the ``sed`` command that replaces ``jax`` with ``jax[cuda12]``
```
YigitElma, line 246: same as above

tests/baseline/test_plot_gammac.png
YigitElma, line 1: This one looks pretty different to me. Especially the lower left corner, the previous one has a lot of yellow, the new one is more purple.

desc/compute/_neoclassical.py
unalmis: This is standard practice for private arguments to functions in Python. Jax does it quite often for example.

desc/objectives/_fast_ion.py
unalmis: I like how it looks. And Daniel wanted all warnings to be yellow.

desc/objectives/_fast_ion.py
unalmis: Because there is an option called spline in bounce2d that has nothing to do with this option, so it is confusing.

docs/installation.rst
unalmis: Thats bad practice because it would need to updated every time a new feature is added. And frankly it's information overload to users to start listing off features they mhaven't even heard of/may not understand.

docs/installation.rst
unalmis: It is just listing the truth. 

This has nothing to do with CUDA awareness. If you install DESC with "CPU + GPU support", then any objectives you use to optimize should support GPU usage. Anything other than that is misleading and confusing to users.

Therefore, this installation is labeled as "partial support"

docs/installation.rst
unalmis: - This PR just follows the example of `master`. Changes to conventions on `master` are outside the scope of this PR.  Changes should be made after this is merged as I am unable to retest installation.

- I personally disagree. This was also discussed in an earlier PR and people were fine as long as both commands for BSD and linux systems are given.

- I worked as a teaching assistant, hired tutor, and my experience is it is more foreign/challenging for people to learn how to use and exit a command line text editor than it is to paste one more command into the terminal. Especially because the key bindings to save and exit on command line terminals are different not only on each system, but sometimes also os versions with windows.

- If this is so unappealing you should just create multiple installation files.

- you cant just pip install without changing the requirements. you got lucky if you did that. just this week a desc user requested i help them debug and they tried pip installing as was done on master prior to #1862 and their build was destroyed because that auto updated jax to an incompatible version. In 1862 I mentioned how the changes I made there are required for correctness, and is not just a cosmetic change.

docs/installation.rst
unalmis: https://github.com/PlasmaControl/DESC/pull/1834#discussion_r2326356023

docs/installation.rst
unalmis: This is tested and works. The only reason for having separate instal instructions for the local machine stuff is because we don't make a contract with the user that we tested the install instructions work on their particular device Y on date X. We do advertise that with della and stellar.

docs/installation.rst
unalmis: Marking as resolved because addressed in different PR and comment.

tests/baseline/test_plot_gammac.png
unalmis: Yes because the resolution is much higher.

unalmis: If you switch to the old resolution, then also switch the spline method to the lower resolution of rounding Y_B down to the nearest multiple of NFP rather than round up (as was donepreviously), you will recover the same plot.

desc/batching.py
unalmis: Out of scope for this PR.

desc/objectives/_fast_ion.py
unalmis: You can change this if you want but you'll need to hold your ground against others if they request to change back to yellow.

f0uriest: I thought a while back we decided to start getting rid of colored text in new code, and eventually removing it entirely? but it was ages ago so im fine if people have different opinions now.

desc/compute/__init__.py
```diff
@@ -33,6 +33,7 @@
     _core,
     _curve,
     _equil,
+    _fast_ion,
```
f0uriest, line 36: wait has this been missing this whole time? i thought we needed to import them here for `register_compute_fun` to work correctly. How has this been working?

desc/integrals/_interp_utils.py
```diff
@@ -396,6 +378,135 @@ def rfft2_modes(n_fft, n_rfft, domain_fft=(0, 2 * jnp.pi), domain_rfft=(0, 2 * j
...
+
+     - ``tests/test_interp_utils.py::TestFastInterp::test_non_uniform_real_FFT``
+     - ``tests/test_interp_utils.py::TestFastInterp::test_nufft2_vec``
+
+    Parameters
+    ----------
+    x : jnp.ndarray
+        Real query points of coordinate in ``domain`` where interpolation is desired.
+        The coordinates stored here must be the same coordinate enumerated across
+        axis ``-1`` of ``f``.
```
f0uriest, line 394: can you explain this bit? This wording makes it sound like it needs `f[...,-1] == x` which doesn't make sense to me

desc/compute/_fast_ion.py
```diff
@@ -147,6 +147,9 @@ def _Gamma_c(params, transforms, profiles, data, **kwargs):
             (automorphism_sin, grad_automorphism_sin),
         )
     )
+    nufft_eps = kwargs.get("nufft_eps", 1e-7)
```
f0uriest, line 150: i think elsewhere you default to 1e-6, is more accuracy needed here?

more generally, how much does the computed results and gradients depend on eps? (we know the error in the fft scales like eps, but how does that effect downstream stuff, im assuming the final gamma_c or eps_effective is still dominated by other sources of error?)

are there cases where lower eps makes a big difference? What about for optimization?

desc/integrals/_interp_utils.py
```diff
@@ -7,34 +7,30 @@
...
 
 import numpy as np
 from interpax import interp1d
-from orthax.chebyshev import chebroots
+from termcolor import colored
+
+try:
+    from jax_finufft import nufft2, options
+except ImportError:
+    warnings.warn(
```
f0uriest, line 19: putting the warning here means i get warned as soon as i import desc, even if i'm not doing anything that ever needs nuffts. I'd rather do something like
```python
try:
    from jax_finufft import nufft2, options
except ImportError:
    nufft2 = options = None


...

def nufft1d2r(....)
    if nufft2 is None:
        warnings.warn("jax-finufft not installed.... falling back to MMT method which is significantly slower")
        return nummt(...)

    return nufft2(...)
```

f0uriest: also do we need to check that we have jax-finufft with gpu support specifically? ie if they only have cpu finufft but desc is running on a gpu

desc/integrals/basis.py
```diff
@@ -313,14 +314,19 @@ def Y(self):
         """Chebyshev spectral resolution."""
         return self.cheb.shape[-1]
 
+    @property
+    def tag(self):
```
f0uriest: what is tag needed for beyond `self.cheb.shape`? my main worry is if there would be cases with the same shape but different tags, which would trigger additional compilation.

requirements.txt
```diff
@@ -3,6 +3,7 @@ colorama <= 0.4.6
 diffrax >= 0.4.1, <= 0.7.0
 h5py >= 3.0.0, <= 3.14.0
 interpax >= 0.3.3, <= 0.3.10
+jax-finufft
```
f0uriest: do we need v1.0.0 or does v0.1.0 also work? Should also add a maximum version <1.1.0 to guard against future breaking changes (and let dependabot update as needed)

desc/objectives/_fast_ion.py
```diff
@@ -102,6 +102,8 @@ class GammaC(_Objective):
         A tighter upper bound than ``num_well=(Aι+B)*num_transit`` is preferable.
         The ``check_points`` or ``plot`` methods in ``desc.integrals.Bounce2D``
         are useful to select a reasonable value.
+
+        This is the most important paramter to specify for performance.
```
f0uriest: *parameter

desc/objectives/_fast_ion.py
```diff
@@ -177,13 +182,31 @@ def __init__(
...
         pitch_batch_size=None,
         surf_batch_size=1,
-        spline=False,
+        nufft_eps=1e-7,
+        use_bounce1d=False,
         Nemov=True,
+        **kwargs,
     ):
+        try:
+            import jax_finufft  # noqa: F401
```
f0uriest, line 190: see previous comment about also checking that jax_finufft has cuda support if desc is using GPU

desc/objectives/_neoclassical.py
```diff
@@ -374,3 +408,25 @@ def _compute_spline(self, params, constants=None):
...
+    return {
+        "dct cfl": _vander_dct_cfl(x, obj._constants["Y"].size),
+        "dft cfl": _vander_dft_cfl(x, obj._grid),
+        "dct spline": _vander_dct_cfl(
+            jnp.linspace(-1, 1, Y_B, endpoint=False), obj._constants["Y"].size
+        ),
+    }
+
+
+def _vander_dft_cfl(x, grid):
```
f0uriest, line 419: what does `cfl` stand for here?

tests/test_integrals.py
```diff
@@ -1479,19 +1474,26 @@ def integrand_grad(data, B, pitch):
...
+                pitch_inv=1 / pitch,
+                data=data,
+                nufft_eps=nufft_eps,
+                check=True,
             ).sum()
 
         pitch = 1.0
-        # can obtain from math or just extrapolate from analytic expression plot
-        analytic_approximation_of_gradient = 650
+        analytic_approximation_of_gradient = 651.8
```
f0uriest, line 1493: why did this change?

tests/test_interp_utils.py
```diff
@@ -29,139 +34,38 @@
 )
 from desc.integrals.basis import FourierChebyshevSeries
 from desc.integrals.quad_utils import bijection_to_disc
+from desc.utils import identity
 
 
-class TestPolyUtils:
```
f0uriest, line 34: are these tests no longer needed?

tests/test_interp_utils.py
f0uriest, line 1: the diff on this file is a mess, can you summarize what was actually removed vs added vs moved/renamed?

docs/installation.rst
f0uriest: I thought you made it so that even without finufft everything still works on GPU? so i don't undestand this comment.

desc/compute/__init__.py
unalmis: think it rebuilt the dependency structure each time compute was called rather than querying the cached structure for those items

desc/integrals/_interp_utils.py
unalmis: coordinate not coordinate value. i'm saying it should be zeta coordinate for the axis that has zeta modes

desc/compute/_fast_ion.py
unalmis: 1e-6 is default for finufft. from this default I  minimized eps until i saw no visual difference in MMT vs nufft for a range of configurations. I also compared against MMT and the differences are quite small.

The reason gamma c is 1e-7 and ripple is >= 1e-6 is because the quadrature is more densely packed for gamma c, and hence requires more precision.

By far the dominant source of error that negatively affects gradients/optimization results and correctness is how well the spline of B along field lines approximates the locations of the bounce points of the Fourier series |B|. Once the _polish_points method is implemented in different PR this will be resolved as the purpose of the spline will simply be reduced to transforming a local root finding algorithm to a global one, and the infinity norm error of the spline would be irrelevant so you can stop it's reconstruction at much lower number of knots

tests/test_integrals.py
unalmis: I replaced my mostly pen and paper taylor expansion approximation with the true numerical finite difference, to the relevant number of sig figs. Made the change while debugging jax finufft.

docs/installation.rst
unalmis: Without finufft on GPU, you don't have support for nuffts on GPU. you only have support for nuffts on cpu in that case. Therefore you do not have full gpu support in DESC if you don't install nuffts for GPU.

tests/test_interp_utils.py
unalmis: New tests were added. The old tests were split into ones that test FFTs and MMTs. This is done for the tests to be consistent with the ones I added to `jax-finnuft`.

tests/test_interp_utils.py
unalmis: It's at the bottom of the file now. This is done for this file to match more consistent with the file I added to tests in `jax-finnuft`.

desc/compute/_fast_ion.py
unalmis: Also of the error subset that is solely due to nufft eps, that subset of error is mostly from the error in constructing the spline that determines the bounce points with the nufft in `fast_cubic_spline`. The error from the nuffts in the quadrature in `bounce.integrate` is less.

desc/integrals/_interp_utils.py
unalmis: No because objective that use GPU gracefully switch to nuffts on CPU instead of nuffts on GPU when GPU support  is not installed.

f0uriest: in that case i feel like we don't need these warnings at all, since some version of finufft will always be available - the default GPU installation would still install a cpu version of finufft. We can still include a warning if GPU support isn't enabled in the objective function, since that's where the extra speed matters most. That also avoids throwing the same warning twice (once in the objective and once again in the underlying internal functions)

desc/integrals/_interp_utils.py
unalmis: frequency space is labeled by zeta and theta as well. it's just the meaning of the coordinate value that changes.

desc/integrals/_interp_utils.py
unalmis: If you don't include this try except guard, then all existing DESC installations will break if the user does not explicitly install finufft installed in their environment. Most developers don't reinstall their environment when DESC updates. People would blame me for the fallout.

f0uriest: actually can you explain this

> No because objective that use GPU gracefully switch to nuffts on CPU instead of nuffts on GPU when GPU support is not installed.

I have CPU version of finufft installed, and this fails:
```python
desc.set_device("gpu")
obj = ObjectiveFunction(EffectiveRipple(eq))
obj.build()
x = obj.x()
obj.compute_scaled_error(x).block_until_ready()
```

```
NotImplementedError: MLIR translation rule for primitive 'nufft2' not found for platform cuda
```

desc/integrals/_interp_utils.py
f0uriest: i think most physicists prefer to use separate labels for real vs frequency space, ie `k` vs `x`, `omega` vs `t` to avoid this sort of confusion

desc/integrals/_interp_utils.py
unalmis: eq.compute ripple worked fine for me after calling set device to gpu in an environment where only jax-finufft cpu was installed on della. 

Also
1. GPU support for ~~`nufft1` ~~ 1D nuffts is not even in the gpu version jax-finufft yet
2. ~~`nufft1`~~ a 1D nufft is called prior to ~~`nufft2`~~ a 2D nufft in the code you posted
3. so it did gracefully switch ~~nufft1~~ the 1D nufft to cpu on your machine but not the 2D ~~nufft2~~ nufft to cpu

f0uriest: `eq.compute` isn't compiled which might explain the different behavior. And where is `nufft1` called? I only see `nufft2` (the primitive referred to above is the `nufft2` from `jax-finufft`, not your wrapper functions `nufft1d2r` and `nufft2d2r`)

unalmis: Sorry I misspoke since `jax-finufft` broke away from the naming conventions from finufft. A 1D version of a type 2 nufft is called in `fast_cubic_spline` method.

unalmis: Ok that means users should set nufft_eps to zero if they get your error then

unalmis: my nufft1d2r etc. functions follow FINUFFT api as there is documentation for FINUFFT. jax-finufft functions aren't documented well so I didn't preserve their API.

unalmis: and eq.compute is compiled for these functions.

desc/integrals/_interp_utils.py
unalmis: i usually see x and x tilde. More often I see X remains the same and the argument changes: X(t) vs X(i omega) . you can change this stuff to whatever you prefer

desc/integrals/basis.py
unalmis: for developers to be less confused during review. it has no utility otherwise.

desc/objectives/_neoclassical.py
unalmis: compute field line

requirements.txt
unalmis: JAX made breaking changes that inhibits use of 0.1.0 on new jax versions. So if you have a new jax version then you need 1.0.

desc/compute/_neoclassical.py
YigitElma: My bad, I thought it was also an argument passed to objective function, but yeah it is created internally

desc/compute/__init__.py
YigitElma: I am not sure but the import of `_fast_ion` in `_old` might have built the dependency tree properly? `@register_compute_fun` probably runs for each function, even if you import 1 thing from that file

desc/integrals/_interp_utils.py
YigitElma: I think the difference is eq.compute is partially compiled (for some compute quantities in data that have jit in their definition) but the objective functions compute compile everything in a single function, so it forces everything to run on the same device. Jax's jit doesn't allow certain portions of the code to run on different devices (unless the data is sharded).

docs/installation.rst
YigitElma: Tbh, I never use the `pip install --editable . ` version, that might be the difference here. But in general, unless you use conda install and pip install for different packages (and you don't have multiple environments activated at the same time, I had this before once, maybe they resolved the issue not sure), normal pip install works fine. Yes, it doesn't check the cross package dependencies, but for `jax[cuda12]`, it installs the same version of jax already installed for cpu, so it is compatible with others.

Anyway, I already know how to install DESC by heart. I just think for new users, keeping them simple is helpful

desc/objectives/_fast_ion.py
YigitElma: https://github.com/PlasmaControl/DESC/issues/1692

docs/installation.rst
unalmis: > Yes, it doesn't check the cross package dependencies, but for jax[cuda12], it installs the same version of jax already installed for cpu, so it is compatible with others.

No it does not maintain compatibility, leading to broken installations. You got lucky if it didn't. As mentioned in the 1862 pill request discussions, pip installing jax[cuda12] can and will change versions of other installed dependencies.  Just as pip installing jax finufft or any other package separately from the requirements file can and will change the jax version.

YigitElma: I have been lucky for the last 2 years and almost every week then

f0uriest:
FYI spent a few hours trying to get jax-finufft GPU working on perlmutter but no luck so far. @rahulgaur104 any luck on your end?

desc/integrals/_interp_utils.py
f0uriest: I think these checks and warnings need to be clarified more. From the wording, it sounds like it's asking the user to just `pip install jax-finufft` avoid this warning (which is what I had initially tried, leading to the error above), but if desc is on GPU that just causes a different error which is more cryptic. if desc is on gpu then we also need to check that finufft is installed with gpu support, and make that part clear in the error message.

Also I'd still like to move this message somewhere farther down, as right  now it warns the user as soon as they import anything from desc, regardless of it it has anything to do with bounce integrals or finufft

f0uriest:
is the 2d nufft we need expressible as a composition of 1d nufft? if we can get away with 1d then it looks like it would be pretty easy to port https://github.com/jakevdp/nufftpy to jax (I can work on it next week) which might make this a lot easier until jax-finufft provides cuda wheels

desc/integrals/_interp_utils.py
unalmis: > From the wording, it sounds like it's asking the user to just pip install jax-finufft avoid this warning (which is what I had initially tried, leading to the error above)

Ok to address that, it is now explicitly written that you should follow the DESC installation instructions.

> Also I'd still like to move this message somewhere farther down, as right now it warns the user as soon as they import anything from desc, regardless of it it has anything to do with bounce integrals or finufft

jax-finufft is now a dependency of DESC. If users don't have this installed they are warned. I have made the warning graceful in that users can continue using DESC. Note that quadax is also a dependency of DESC - if users don't have quadax installed then everything breaks even though quadax is only used to compute the quantity `chi` which is an incredibly narrow application that most people never need. So why are you uncomfortable with a helpful and benign warning when jax-finufft is not installed, but you are fine with a cryptic error message when quadax is not installed? If the answer is because it leads to that GPU issue, then I believe the best resolution is as follows:

> if desc is on gpu then we also need to check that finufft is installed with gpu support

If a user has a GPU, then correctness of installation is checked via the commands in the "checking your installation" section of the documentation. I will add a few commands there that check jax-finufft is properly installed on gpu. DESC doesn't do any runtime checks to see if JAX is installed on GPU and therefore it shouldn't do it for jax-finufft either.

docs/notebooks/tutorials/EffectiveRipple.ipynb
unalmis, line 1: See #1906 if you wonder why optimization result changed.

requirements.txt
```diff
@@ -3,6 +3,7 @@ colorama <= 0.4.6
 diffrax >= 0.4.1, <= 0.7.0
 h5py >= 3.0.0, <= 3.14.0
 interpax >= 0.3.3, <= 0.3.10
+jax-finufft >= 1.1.0
```
YigitElma, line 6: Is `jax-finufft >= 1.1.0, <=1.1.0` a valid syntax? Then, can we add the max version as the other dependencies? This will also remind devs to check the installation stuff regularly through Dependabot PRs.

desc/integrals/_interp_utils.py
YigitElma: > DESC doesn't do any runtime checks to see if JAX is installed on GPU

JAX does that automatically for us and warns like `An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.`.

Something similar to what Rory suggested above, maybe [this](https://github.com/flatironinstitute/jax-finufft/blob/d213861911772bd1b033ddc053ea3009fc1a967f/src/jax_finufft/lowering.py#L11) (or some other variable from the package) can be used to detect the GPU version of `jax-finufft` in combination with `desc_config["kind"]=='gpu'`?

unalmis: I am not going to use undocumented methods in the source code of an external package.

jax-finufft has no obligation or contract to ensure those methods work in a way DESC expects.

unalmis: > JAX does that for us

you guys can make an issue on jax-finufft to request a similar message if you want

requirements.txt
unalmis: there's four other dependencies that we don't top pin. someone can make a new pr after this is merged to top pin everything and retest installation on clusters etc. if they want. this works fine now and there are advantages to not top pinning as well.

unalmis:
> > is the 2d nufft we need expressible as a composition of 1d nufft?
> 
> Yes always, but not with the same performance as a 2D algorithm. It's not like an FFT where the best implementations compose factorized 1D FFT modulo caching details.

If I was not explicit enough, please do not re-implement a NUFFT. Flatiron does it well. I would like to merge this as everything works and is tested and the GPU installation is just two lines. I'm still seeing papers that are using old fortran codes to compute epsilon effective so there's definitely a need for this PR, especially because of the regression #1864 .

f0uriest:
I would like to use finufft, but the "2 extra line" installation is only on Princeton clusters, which most users and developers don't have access to. Installing it on other clusters seems more difficult (still no success, will keep trying this week) so I'm trying to avoid future issues.

Another option would be to make it work the way you originally claimed, moving data back and forth so that CPU finufft works with everything else on GPU. In theory should be doable with a callback+ custom jvp

unalmis:
> moving data back and forth so that CPU finufft works with everything else on GPU. In theory should be doable with a callback+ custom jvp


Yes that can be implemented with a pure jax callback and `jvp`. In fact, that's entirely what the `jax-finufft` library does except an FFI callback is used instead of a pure jax callback.

Hence you would need to duplicate the `jax-finufft` source code for the `jvp`. I disapprove of that for the reason mentioned in https://github.com/PlasmaControl/DESC/pull/1834#discussion_r2338663150 . It could also cause issues with being able to check whether the program is correctly installed on GPU. Right now there is a loud error that indicates to users whether GPU is installed correctly. If they fail to install on GPU, they can either run the objective on CPU or select the option to switch to MMT.

rahulgaur104:
> FYI spent a few hours trying to get jax-finufft GPU working on perlmutter but no luck so far. @rahulgaur104 any luck on your end?

@f0uriest 
No, it doesn't install with the current instructions on Perlmutter. I get the following error:

```
desc-env) rgaur@login12:/pscratch/sd/r/rgaur/DESC> pip install -Ccmake.define.JAX_FINUFFT_USE_CUDA=ON --no-binary=jax-finufft 'jax-finufft >= 1.1.0'
Collecting jax-finufft>=1.1.0
  Using cached jax_finufft-1.1.0.tar.gz (4.4 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: jax>=0.5.0 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax-finufft>=1.1.0) (0.6.0)
Requirement already satisfied: numpy in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax-finufft>=1.1.0) (2.2.6)
Collecting pydantic>=2 (from jax-finufft>=1.1.0)
  Using cached pydantic-2.11.9-py3-none-any.whl.metadata (68 kB)
Requirement already satisfied: jaxlib<=0.6.0,>=0.6.0 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax>=0.5.0->jax-finufft>=1.1.0) (0.6.0)
Requirement already satisfied: ml_dtypes>=0.5.0 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax>=0.5.0->jax-finufft>=1.1.0) (0.5.1)
Requirement already satisfied: opt_einsum in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax>=0.5.0->jax-finufft>=1.1.0) (3.4.0)
Requirement already satisfied: scipy>=1.11.1 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax>=0.5.0->jax-finufft>=1.1.0) (1.15.3)
Collecting annotated-types>=0.6.0 (from pydantic>=2->jax-finufft>=1.1.0)
  Using cached annotated_types-0.7.0-py3-none-any.whl.metadata (15 kB)
Collecting pydantic-core==2.33.2 (from pydantic>=2->jax-finufft>=1.1.0)
  Using cached pydantic_core-2.33.2-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (6.8 kB)
Requirement already satisfied: typing-extensions>=4.12.2 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from pydantic>=2->jax-finufft>=1.1.0) (4.15.0)
Collecting typing-inspection>=0.4.0 (from pydantic>=2->jax-finufft>=1.1.0)
  Using cached typing_inspection-0.4.1-py3-none-any.whl.metadata (2.6 kB)
Using cached pydantic-2.11.9-py3-none-any.whl (444 kB)
Using cached pydantic_core-2.33.2-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (2.0 MB)
Using cached annotated_types-0.7.0-py3-none-any.whl (13 kB)
Using cached typing_inspection-0.4.1-py3-none-any.whl (14 kB)
Building wheels for collected packages: jax-finufft
  Building wheel for jax-finufft (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for jax-finufft (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [32 lines of output]
      *** scikit-build-core 0.11.6 using CMake 3.20.4 (wheel)
      *** Configuring CMake...
      loading initial cache file build/cp313-cp313-linux_x86_64/CMakeInit.txt
      -- The C compiler identification is GNU 11.4.0
      -- The CXX compiler identification is GNU 11.4.0
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /global/homes/r/rgaur/.conda/envs/desc-env/bin/gcc - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /global/homes/r/rgaur/.conda/envs/desc-env/bin/g++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
      -- Using CMake version: 3.20.4
      -- Found OpenMP_C: -fopenmp (found version "4.5")
      -- Found OpenMP_CXX: -fopenmp (found version "4.5")
      -- Found OpenMP: TRUE (found version "4.5")
      -- jax_finufft: OpenMP found
      -- Looking for a CUDA compiler
      -- Looking for a CUDA compiler - NOTFOUND
      CMake Error at CMakeLists.txt:48 (message):
        jax_finufft: No CUDA compiler found! Please ensure the CUDA Toolkit is
        installed, or set JAX_FINUFFT_USE_CUDA=OFF to disable GPU support.
      
      
      -- Configuring incomplete, errors occurred!
      See also "/tmp/pip-install-sllhzrji/jax-finufft_cd1f0b001f914f6fa23592aefee8e12a/build/cp313-cp313-linux_x86_64/CMakeFiles/CMakeOutput.log".
      See also "/tmp/pip-install-sllhzrji/jax-finufft_cd1f0b001f914f6fa23592aefee8e12a/build/cp313-cp313-linux_x86_64/CMakeFiles/CMakeError.log".
      
      *** CMake configuration failed
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for jax-finufft
Failed to build jax-finufft
error: failed-wheel-build-for-install

× Failed to build installable wheels for some pyproject.toml based projects
╰─> jax-finufft
```

I have made sure that without finufft everything runs on CPUs and GPUs. I was really hoping this to work! It would've helped a lot with the database stuff.

unalmis:
> > FYI spent a few hours trying to get jax-finufft GPU working on perlmutter but no luck so far. @rahulgaur104 any luck on your end?
> 
> 
> 
> @f0uriest 
> 
> No, it doesn't install on Perlmutter with the current instructions. I get the following error:
> 
> 
> 
> ```
> 
> desc-env) rgaur@login12:/pscratch/sd/r/rgaur/DESC> pip install -Ccmake.define.JAX_FINUFFT_USE_CUDA=ON --no-binary=jax-finufft 'jax-finufft >= 1.1.0'
> 
> Collecting jax-finufft>=1.1.0
> 
>   Using cached jax_finufft-1.1.0.tar.gz (4.4 MB)
> 
>   Installing build dependencies ... done
> 
>   Getting requirements to build wheel ... done
> 
>   Installing backend dependencies ... done
> 
>   Preparing metadata (pyproject.toml) ... done
> 
> Requirement already satisfied: jax>=0.5.0 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax-finufft>=1.1.0) (0.6.0)
> 
> Requirement already satisfied: numpy in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax-finufft>=1.1.0) (2.2.6)
> 
> Collecting pydantic>=2 (from jax-finufft>=1.1.0)
> 
>   Using cached pydantic-2.11.9-py3-none-any.whl.metadata (68 kB)
> 
> Requirement already satisfied: jaxlib<=0.6.0,>=0.6.0 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax>=0.5.0->jax-finufft>=1.1.0) (0.6.0)
> 
> Requirement already satisfied: ml_dtypes>=0.5.0 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax>=0.5.0->jax-finufft>=1.1.0) (0.5.1)
> 
> Requirement already satisfied: opt_einsum in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax>=0.5.0->jax-finufft>=1.1.0) (3.4.0)
> 
> Requirement already satisfied: scipy>=1.11.1 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from jax>=0.5.0->jax-finufft>=1.1.0) (1.15.3)
> 
> Collecting annotated-types>=0.6.0 (from pydantic>=2->jax-finufft>=1.1.0)
> 
>   Using cached annotated_types-0.7.0-py3-none-any.whl.metadata (15 kB)
> 
> Collecting pydantic-core==2.33.2 (from pydantic>=2->jax-finufft>=1.1.0)
> 
>   Using cached pydantic_core-2.33.2-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (6.8 kB)
> 
> Requirement already satisfied: typing-extensions>=4.12.2 in /global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.13/site-packages (from pydantic>=2->jax-finufft>=1.1.0) (4.15.0)
> 
> Collecting typing-inspection>=0.4.0 (from pydantic>=2->jax-finufft>=1.1.0)
> 
>   Using cached typing_inspection-0.4.1-py3-none-any.whl.metadata (2.6 kB)
> 
> Using cached pydantic-2.11.9-py3-none-any.whl (444 kB)
> 
> Using cached pydantic_core-2.33.2-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (2.0 MB)
> 
> Using cached annotated_types-0.7.0-py3-none-any.whl (13 kB)
> 
> Using cached typing_inspection-0.4.1-py3-none-any.whl (14 kB)
> 
> Building wheels for collected packages: jax-finufft
> 
>   Building wheel for jax-finufft (pyproject.toml) ... error
> 
>   error: subprocess-exited-with-error
> 
>   
> 
>   × Building wheel for jax-finufft (pyproject.toml) did not run successfully.
> 
>   │ exit code: 1
> 
>   ╰─> [32 lines of output]
> 
>       *** scikit-build-core 0.11.6 using CMake 3.20.4 (wheel)
> 
>       *** Configuring CMake...
> 
>       loading initial cache file build/cp313-cp313-linux_x86_64/CMakeInit.txt
> 
>       -- The C compiler identification is GNU 11.4.0
> 
>       -- The CXX compiler identification is GNU 11.4.0
> 
>       -- Detecting C compiler ABI info
> 
>       -- Detecting C compiler ABI info - done
> 
>       -- Check for working C compiler: /global/homes/r/rgaur/.conda/envs/desc-env/bin/gcc - skipped
> 
>       -- Detecting C compile features
> 
>       -- Detecting C compile features - done
> 
>       -- Detecting CXX compiler ABI info
> 
>       -- Detecting CXX compiler ABI info - done
> 
>       -- Check for working CXX compiler: /global/homes/r/rgaur/.conda/envs/desc-env/bin/g++ - skipped
> 
>       -- Detecting CXX compile features
> 
>       -- Detecting CXX compile features - done
> 
>       -- Using CMake version: 3.20.4
> 
>       -- Found OpenMP_C: -fopenmp (found version "4.5")
> 
>       -- Found OpenMP_CXX: -fopenmp (found version "4.5")
> 
>       -- Found OpenMP: TRUE (found version "4.5")
> 
>       -- jax_finufft: OpenMP found
> 
>       -- Looking for a CUDA compiler
> 
>       -- Looking for a CUDA compiler - NOTFOUND
> 
>       CMake Error at CMakeLists.txt:48 (message):
> 
>         jax_finufft: No CUDA compiler found! Please ensure the CUDA Toolkit is
> 
>         installed, or set JAX_FINUFFT_USE_CUDA=OFF to disable GPU support.
> 
>       
> 
>       
> 
>       -- Configuring incomplete, errors occurred!
> 
>       See also "/tmp/pip-install-sllhzrji/jax-finufft_cd1f0b001f914f6fa23592aefee8e12a/build/cp313-cp313-linux_x86_64/CMakeFiles/CMakeOutput.log".
> 
>       See also "/tmp/pip-install-sllhzrji/jax-finufft_cd1f0b001f914f6fa23592aefee8e12a/build/cp313-cp313-linux_x86_64/CMakeFiles/CMakeError.log".
> 
>       
> 
>       *** CMake configuration failed
> 
>       [end of output]
> 
>   
> 
>   note: This error originates from a subprocess, and is likely not a problem with pip.
> 
>   ERROR: Failed building wheel for jax-finufft
> 
> Failed to build jax-finufft
> 
> error: failed-wheel-build-for-install
> 
> 
> 
> × Failed to build installable wheels for some pyproject.toml based projects
> 
> ╰─> jax-finufft
> 
> ```

1. What do you mean the install instructions for finufft fail on perlmutter? I am confused because I do not provide any install instructions for perlmutter. Only della and stellar.

2. Are you actually on a GPU node in the terminal where you run the command to install jax-finufft? The most common mistake that people make is that they try to build the install on a node that doesn't have a gpu. Pip builds the wheels for the gpu associated with your current terminal session. I got that same error until I actually ssh'ed to della gpu for the install. I suspect this issue was the cause of Rory's installation issues but I don't know what installation methods he tried so I didn't speculate.

unalmis:
I suppose there is no utility in adding the test jax label since master fails that

rahulgaur104:
> > > FYI spent a few hours trying to get jax-finufft GPU working on perlmutter but no luck so far. @rahulgaur104 any luck on your end?
> > 
> > 
> > @f0uriest
> > No, it doesn't install on Perlmutter with the current instructions. I get the following error:
> 
> > × Building wheel for jax-finufft (pyproject.toml) did not run successfully.
> > ```
> >   -- Looking for a CUDA compiler - NOTFOUND
> > 
> >   CMake Error at CMakeLists.txt:48 (message):
> > 
> >     jax_finufft: No CUDA compiler found! Please ensure the CUDA Toolkit is
> > 
> >     installed, or set JAX_FINUFFT_USE_CUDA=OFF to disable GPU support.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > × Failed to build installable wheels for some pyproject.toml based projects
> > ╰─> jax-finufft
> 
> 1. What do you mean the install instructions for finufft fail on perlmutter? I am confused because I do not provide any install instructions for perlmutter. Only della and stellar. I can help more if you share what you tried.
> 2. Are you actually on a GPU node in the terminal where you run the command to install jax-finufft? The most common mistake that people make is that they try to build the install on a node that doesn't have a gpu. Pip builds the wheels for the gpu associated with your current terminal session. I got that same error until I actually ssh'ed to della gpu for the install. I suspect this issue might be the cause of Rory's installation issues but he didn't share the installation method he tried nor the error message, so I didn't speculate.

I am aware that the instructions were meant for two Princeton clusters. I was trying to see if they would work on Perlmuter. If they did, I would have been much more inclined to set default bounce calculation to use jax-finufft for both GPUs and CPUs since most DESC users in the US use Perlmutter. 

Perlmutter login node always has one 40G A100 GPU.

unalmis:
> I would like to merge this as everything works and is tested and the GPU installation is just two lines. I'm still seeing papers that are using low order accurate fortran codes to compute epsilon effective so there's definitely a need for this PR, especially because of the regression #1864 .

It'd be really nice to get this and the stream map PR in so that I can cite DESC pypi in my paper instead of some random pull request

unalmis:
> >  We can speculate about non-ability to debug anything. What matters here is whether it works now and whether the installation we use is one that jax-finufft advertises it supports, which it does. If something breaks with the installation method jax-finufft advertises it supports, then it is the obligation of jax-finufft developers to debug so our expertise in debugging such issues is not relevant.
> 
> 
> 
> >    what happens if someone has another compiler/fftw etc
> 
> 
> 
> >  All that is necessary is for the installation to work on the devices and methods that DESC promises it should work on. If someone wants to use a different compiler or a different setup from the installation instructions you advertise support on, then they are welcome too. However, you will not be obligated to debug their issues. This is the same contract DESC and other packages always make with their users. Example: If a user wants to install DESC with a JAX version that DESC does not promise support for, it is not the DESC developers burden to debug that user's code.
> 
> 
> 
> 
> 
> While this is true in an ideal world, in practice we do end up having to debug and fix other people's issues, even if its just to identify the source of the problem and say we won't fix it, it's still a nonzero amount of time for each occurrence. @unalmis if you want to volunteer to be the point person for updates/maintenance/support on this that's fine, but if not then it will be up to the rest of us to maintain this code and these instructions, and to help users with issues that pop up, so we should be sure we have the expertise and bandwidth necessary for that.

The reality is any issues will probably get forwarded to me to resolve regardless of whether I volunteer to absolve the other coauthors of this code base of their responsibility to maintain the code.

It is not clear to me if this request was being stated as a general remark or whether an implication was made that my work's admission to DESC was contingent on my promise to maintain the code in the future. If the latter, it is improper for this to be requested as my current employment does not have relation to the research enabled by this code, while the other developers' employment does. 

I think it is an irrelevant point anyway:  
- If the NUFFTs break due to an external code change and nobody from DESC, jax, jax-finifft or the research computing clusters want to fix it, then all the users need to do is specify to turn NUFFTs off with a single command.
- The old version of DESC will still have that feature working if they need.
- There's plenty of open issues that anyone can contribute to solve. We don't remove features from the code just because in the future an issue came up. We just make an issue until it gets fixed.

desc/plotting.py
```diff
@@ -4295,10 +4295,10 @@ def plot_gammac(
         Default: 0.5
     alphas : array_like, optional
         Fieldline label values (toroidal angle).
-        Default: np.linspace(0, 2π, 32, endpoint=True)
+        Default: np.linspace(0, 2π, 25, endpoint=True)
     num_pitch : int, optional
         Number of pitch angle values for bounce integral calculation.
-        Default: 16
+        Default: 28
```
dpanici, line 4343: Can `nufft_eps` be added as a kwarg to the plotting function for gammaC

docs/installation.rst
dpanici: If the list ever gets out of date we will be reminded and can update. I made an issue at least so it is noted which features are enabled by jax-finufft on GPU.

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@ Backend
 - Minimum JAX version bumped up to ``0.4.29``
 
 
+Performance Improvements
+
+- [Partial summation in coordinate mapping](https://github.com/PlasmaControl/DESC/pull/1826).
+- [NUFFTS](https://github.com/PlasmaControl/DESC/pull/1834) are now used by default for computing bounce integrals.
```
dpanici, line 32: You can list the performance improvement here too

requirements.txt
dpanici: there are also advantages to only allow up to the versions we've tested, and our git dependencies action will automatically up the top-pin as new versions exist. I vote for a top pin here esp. since it is jax-related

desc/plotting.py
unalmis: If I make that change it's just going to be replaced with the compute kwargs PR.

CHANGELOG.md
unalmis: I am fine with how it is. If you want to advertise it more feel free to.

requirements.txt
unalmis: if it ever breaks, then I will top pin. Jax-finufft is the most stable program there is, they don't even merge  bug fixes . Jax is the one that breaks with each point release.

unalmis: When you have difficulty building (as Rory had on non Princeton clusters) you do not want to narrow the version range of supported codes.  You want to give pip more freedom to find versions that can yield compatible builds.

desc/integrals/_interp_utils.py
```diff
@@ -396,6 +375,135 @@ def rfft2_modes(n_fft, n_rfft, domain_fft=(0, 2 * jnp.pi), domain_rfft=(0, 2 * j
...
+    else:
+        s = f.shape[rfft_axis] // 2
+        s = jnp.exp(1j * s * (x1 if rfft_axis == -1 else x0))
+        s = s[..., jnp.newaxis, :] if vec else s
+        f = jnp.fft.ifftshift(f, rfft_axis)
+
+    opts = options.Opts(modeord=1)
+    JF_BUG = True
+    if JF_BUG:
+        # https://github.com/flatironinstitute/jax-finufft/pull/159
```
dpanici, line 500: if this PR on jax-finnuft's side gets merged, will this logical no longer be necessary?

This logical is avoiding the bug in jax-finnuft that causes issues with the derivatives? it seems from the notebook that the derivatives are working as inteded

desc/integrals/basis.py
```diff
@@ -86,21 +86,9 @@ def _in_epigraph_and(is_intersect, df_dy_sign, /):
...
-    errorif(
-        jnp.ndim(arr) > cheb.ndim,
-        NotImplementedError,
-        msg=f"Only one additional axis for batch dimension is allowed. "
-        f"Got {jnp.ndim(arr) - cheb.ndim + 1} additional axes.",
-    )
-    return cheb if jnp.ndim(arr) < cheb.ndim else cheb[jnp.newaxis]
-
-
+# TODO: Move interpolation part to interpax. This is the only
```
dpanici, line 89: Tagging for reference: #1388

desc/integrals/bounce_integral.py
```diff
@@ -439,11 +442,14 @@ def compute_theta(eq, X=16, Y=32, rho=1.0, iota=None, clebsch=None, **kwargs):
...
+            Parameters from the equilibrium, such as R_lmn, Z_lmn, i_l, p_l, etc
+            Defaults to ``eq.params_dict``.
+        profiles
+            Optional profiles.
+        tol : float
+            Stopping tolerance for root finding.
+            Default is ``1e-7``.
         kwargs
             Additional parameters to supply to the coordinate mapping function.
             See ``desc.equilibrium.Equilibrium.map_coordinates``.
```
dpanici, line 455: `lmbda` is not in the kwargs of map_coordinates, should be mentioned here to avoid undocumented kwargs

desc/integrals/_interp_utils.py
unalmis: correct, if my PR is merged in jax-finufft then this logic is only necessary for versions prior to my PR

desc/integrals/bounce_integral.py
unalmis: Lmbda is a temporary crutch that should not be necessary for the proper partial summation. After  #1243 is merged lmbda will be removed. This is why it is not supported in the public API. (It's typical for private parameters in Python to be specified as a hidden kwarg to avoid polluting the signature of the function).

dpanici: Got it, did not realize it was meant to be private

dpanici:
Generally I am fine with this going in, I reviewed as deep as I'm willing to. The tests and the before/after eff ripple and gammaC being identical makes me trust this just fine.

I will hold off on the final approval till the jax tests run and don't all fail like master was doing, just to make sure we don't introduce anything unexpected.

One thing to note, which tbh I am ok, is upon cursory glance at the jax tests, with is this pins our upper jax version to at least 0.5.0 (we just upped the lower to 0.4.29 which is 15 months old, while 0.5.0 is 9 months old. I am not sure what exact criterion we have for what we want the lower bounds to be, though the wider the more likely a random cluster will be able to install)
```
The conflict is caused by:
    The user requested jax==0.4.38
    jax[cpu] 0.4.38 depends on jax 0.4.38 (from https://files.pythonhosted.org/packages/22/49/b4418a7a892c0dd64442bbbeef54e1cdfe722dfc5a7bf0d611d3f5f90e99/jax-0.4.38-py3-none-any.whl (from https://pypi.org/simple/jax/) (requires-python:>=3.10))
    diffrax 0.4.1 depends on jax>=0.4.13
    interpax 0.3.3 depends on jax<0.5.0 and >=0.3.2
    orthax 0.1.0 depends on jax<=0.5.0 and >=0.3.2
    quadax 0.2.2 depends on jax<=0.5.0 and >=0.3.2
    equinox 0.11.0 depends on jax>=0.4.13
    jax-finufft 1.1.0 depends on jax>=0.5.0
```



For the perlmutter stuff, I don't think it is a dealbreaker:
- This still is supported and available on CPU there
- on GPU the old method is still available upon passing in `nufft_eps=0`, with the relevant error explaining this explicitly
- I would be pretty surprised if it ends up being impossible to install on perlmutter, though I do not have access myself to test this. @rahulgaur104 have you maybe opened a support ticket with their team, since they've been pretty helpful in the past?
- I do think the instructions should mention if the install is with/without jax-finnuft gpu support though just to be clear to users that the tested installation we have right now is without the GPU jax-finnuft

unalmis:
Yes the minimum jax version for jax-finufft is 0.5.0.

dpanici:
> > I do think the instructions should mention if the install is with/without jax-finnuft gpu support though just to be clear to users that the tested installation we have right now is without the GPU jax-finnuft
> 
> What should I do to address this request? I thought I already have separate installs. Do you mean Della? On della and stellar instsll the GPU nuffts are tested to work.

I mean Perlmutter specifically, to mention that the "tested install" we have there does not include GPU-supported nuffts right now (assuming Rahul/Rory/Yigit have not figured out the install there already)

dpanici (approved):
Approving but would still like a 2nd approver before merge since this is a larger PR

desc/integrals/_interp_utils.py
f0uriest: This warning still doesn't cover the cases it needs to. 

As a concrete example, someone following our installation instructions for GPU on perlmutter (or any other non-princeton cluster) will get a cpu only version of jax-finufft, meaning this warning won't trigger and they'll instead get an error about missing translation rules.

A simple way to test that jax-finufft works on gpu would be something like
```python
if desc.backend.desc_config['kind'] == 'gpu':
    try:
        jax.jit(nufft)(jax.device_put(x, jax.devices("gpu")[0])
```

I also reiterate my point that this check/warning should be moved lower so we don't bother users with warnings about code they may never execute.

requirements.txt
f0uriest: >  Jax-finufft is the most stable program there is, they don't even merge bug fixes . 

i wouldn't call that "stable" i'd call it "unmaintained". This is another issue that worries me.

desc/plotting.py
f0uriest: #1915 only adds them to the "generic" plotting functions that accept an arbitrary compute key (eg `plot_{1d,2d,3d,section}`). For this case we should add `nufft_eps` ourselves (we already have explicit arguments for all the other kwargs used by gamma_c)

unalmis: That PR should also change this   plotting function which exists in desc to accept compute kwargs  There are other parameters which are excluded besides nufft eps right now as well.

unalmis:
> - Perlmutter is a GPU cluster, so having it work only on CPU is a major limitation.

> So I think we should make sure we support perlmutter before merging this.

1. I do not have access to Perlmitter and therefore cannot solve your issues in #1924 or test installation of this new feature on perlmutter GPU.

2. I am obligated to emphasize that  whether DESC works  on Perlmutter has no association to this PR. This PR does nothing to increase or decrease support on perlmutter.

> - ie, as is this PR is basically only an improvement for the few people using princeton clusters. For everyone else, its at minimum some annoying warnings, at worse breaking code.

The implementation on master is broken due to a memory regression in JAX.  This PR solves that problem. Users do not get annoyed at this.

Maybe if one has constant access to 80+gb GPUs then they are wealthy enough to not  notice the issue, but common folk need this for this work to be usable.

> at worse breaking code.

When DESC increased JAX support past 0.5.0, end user code broke due to the memory regression. For those users, NOT merging this PR would therefore cause their code to break.

Right now (at least for a fact a couple weeks ago) researchers are spending development time using the inferior Bounce1D methods for new features as well as analysis due to this issue. 

Researchers have also decided to not use DESC at all for this. It is safe to presume that a decision was made because of this memory issue.

> - The existing check/warning doesn't work if a CPU only version of jax-finufft is installed (which is what our instructions will do), the code will instead throw a cryptic error in that case.

This is the only complaint that I consider actionable and valid.

Note, this is no different than what DESC currently does with JAX - We officially support version 0.6.1 enable users to install yet when DESC is run with GPUs with that version they get the cryptic errors. I am sure Rahul spent time with that issue  on #1924.

The resolution that the DESC developers implemented for that was to just add a section to the installation instructions. 

If DESC has changed its philosophy that such resolutions are no longer sufficient,  then I can move the 5 lines that test the installation of finifft on GPU from the installation directions to a mandatory block of code that executes when DESC is installed. Then if I get the cryptic error, I'll catch it and tell users they need set to nufft_eps=0 through this medium as well. 


> At the bare minimum we should make clear what is supported and what isn't

>  "1 line install on any system" I'm not a fan of adding significant additional complications to that.

Right now, in parts of the code that uses this, I documented what is expected as well as how users type the 2 words to resolve the issue if they don't satisfy this expectation.

I think the installation is 5 lines of code. It's not possible to make it one line install.

I don't understand the significant complications part. Is there something  besides printing a different warning that you would like to be changed?

unalmis:
Ok its good to go. #1944 was made to address more warnings as it has not been pointed out what "significant additional complications"  means.

unalmis:
And it's impossible for the user to end up with an installation where they have finufft on CPU and not on GPU on a GPU device unless they attempt to reinstall DESC after this Prnis merged. So the point about breaking user code with the cryptic message is not possible.

desc/integrals/_interp_utils.py
unalmis: The point about moving the warning lower was addressed above^. See comparison to quadax.

unalmis:
I keep spending significant time debugging all the failing tests and fixing merges etc  from other prs to maintain this Pr. Please do not continue to block PR for cosmetic reasons as these delays also handicap my time/ability to submit the publication.

unalmis:
FYI @f0uriest shared that he is neutral for this PR rather than against and @dpanici is fine with it and has already approved.

PR #1820, opened 2025-07-21
Guiding Center Particle Tracing 

YigitElma:
This PR is a combination of #734 and #1410.

The main structure is under discussion in #1657.

- Adds particle tracing capabilities in ``desc.particles`` module.
    - Particle tracing is done via ``desc.particles.trace_particles`` function.
    - Particles can be initialized in couple different ways:
        - ``ManualParticleInitializerLab`` : Initializes particles at given positions in lab coordinates.
        - ``ManualParticleInitializerFlux`` : Initializes particles at given positions in flux coordinates.
        - ``CurveParticleInitializer`` : Initializes N particles on a given curve.
        - ``SurfaceParticleInitializer`` : Initializes N particles on a given surface.
    - Implemented particle trajectory models are:
        - ``VacuumGuidingCenterTrajectory`` : Integrates the particle motion by vacuum guiding center ODEs, conserving energy and mu.
    - Particle trajectories can be plotted with ``desc.plotting.plot_particle_trajectories`` function.

We can delete #734 and #1410 once this is merged.

PR #1839, opened 2025-08-10
Updates to  `field_line_integrate`

### LINKED ISSUE #1759: `diffrax.diffeqsolve(throw=False)`
By default, `diffrax.diffeqsolve` uses the argument `throw=True`. In some cases this may cause `field_line_integrate` to terminate prematurely with an error. It might be better to set `throw=False`, and then check what data can still be safely returned. This would allow the user to plot partial field lines trajectories up until the integration failed.

YigitElma:
Includes some of the changes realized during #1820 

- The lambda function definition in the `field_line_integrate` can cause recompilation if one uses it in an objective. This PR moves the lambda function out, passing all the arguments in `args`.
- Uses `vmap_chunked` instead of `jnp.vectorize`
- Adds more arguments of diffrax to docs (will be useful for objectives)
- Renames `maxsteps` to `max_steps` and updates the docs
- Bumps the minimum `diffrax` version to 0.6.0 (`Event` is added in that release)

Resolves #1759


desc/magnetic_fields/_core.py
```diff
@@ -2642,60 +2642,45 @@ def field_line_integrate(
...
-        y0=x,
-        t0=phis[0],
-        t1=phis[-1],
-        saveat=saveat,
-        max_steps=maxstep * len(phis),
-        dt0=min_step_size,
-        args=(field,),
-        **kwargs,
-    ).ys
+    event = event if event is not None else Event(default_terminating_event)
```
YigitElma: We might need to bump the minimum `diffrax` to version 0.6.0 for this

dpanici:
Should I just point #1855 to this PR?

desc/magnetic_fields/_core.py
```diff
@@ -2579,13 +2580,17 @@ def field_line_integrate(
...
     atol=1e-8,
-    maxstep=1000,
+    max_steps=None,
     min_step_size=1e-8,
     solver=Tsit5(),
     bounds_R=(0, np.inf),
     bounds_Z=(-np.inf, np.inf),
+    stepsize_controller=None,
+    saveat=None,
+    event=None,
```
f0uriest: im not sure how I feel about exposing these publicly, since they're mutually exclusive with other input parameters. Is there a common use case where they need to be provided? Otherwise I'd vote to keep them private/keyword only and not for public use.

desc/magnetic_fields/_core.py
```diff
@@ -2644,60 +2666,48 @@ def field_line_integrate(
...
-                rpz, params, basis="rpz", source_grid=source_grid, chunk_size=chunk_size
-            ).T
-        )
-        return jnp.array(
-            [r * br / bp * jnp.sign(bp), jnp.sign(bp), r * bz / bp * jnp.sign(bp)]
-        ).squeeze()
+    max_steps = (
+        max_steps
+        if max_steps is not None
+        else int((phis[-1] - phis[0]) / min_step_size * 1000)
```
f0uriest: this would make the default max_steps ~ 1e11 which is likely way too huge. Also note that the `RecursiveCheckpointAdjoint` means that the integration loop is actually done as a for loop + conditional in the body, so it will always do the maximum number of steps (but with a no-op for steps after convergence) but that still takes time, especially for ~1e11 steps. 

It seems like in most cases it takes ~<100 steps per transit, so I think we could still be pretty safe if we just did max_steps=~500-1000 steps per transit

desc/magnetic_fields/_core.py
```diff
@@ -2644,60 +2666,48 @@ def field_line_integrate(
...
+    )
 
     # diffrax parameters
-
-    def default_terminating_event_fxn(state, **kwargs):
-        R_out = jnp.logical_or(state.y[0] < bounds_R[0], state.y[0] > bounds_R[1])
-        Z_out = jnp.logical_or(state.y[2] < bounds_Z[0], state.y[2] > bounds_Z[1])
+    def default_terminating_event(t, y, args, **kwargs):
```
f0uriest, line 2658: this is another locally defined function that could trigger recompilation, so should probably be moved outside this scope

desc/magnetic_fields/_core.py
```diff
@@ -2644,60 +2666,48 @@ def field_line_integrate(
...
-        y0=x,
-        t0=phis[0],
-        t1=phis[-1],
-        saveat=saveat,
-        max_steps=maxstep * len(phis),
-        dt0=min_step_size,
-        args=(field,),
-        **kwargs,
-    ).ys
+    event = event if event is not None else Event(default_terminating_event)
```
f0uriest: `desc.utils.setdefault` is a slightly shorter way of doing this kind of thing

tests/test_magnetic_fields.py
```diff
@@ -1201,25 +1201,35 @@ def test_field_line_integrate(self):
...
         r0 = [10.001]
         z0 = [0.0]
         phis = [0, 2 * np.pi]
+
+        def fun0(r0, z0, field):
+            return field_line_integrate(r0, z0, phis, field, max_steps=1000)
+
         # check if it is jittable
-        r, z = jit(field_line_integrate)(r0, z0, phis, field)
+        r, z = jit(fun0)(r0, z0, field)
```
f0uriest: i think you don't want to jit fun0 here if you're trying to check recompilation. This should pass on master I think? since its only caching based on r0, z0 and field.

desc/magnetic_fields/_core.py
YigitElma: Hmm I was not aware that it was looping for max steps, I chose the max steps based on worst case scenario, like you use min step for whole integration. But yeah, smaller default sounds better.

desc/magnetic_fields/_core.py
YigitElma: Similar to your later comment, we create the default event if not given and it might cause recompilation. So, I wanted to expose them for people who want to use it for optimization. 

Another option could be to pass these in `options` dict. Btw, the reason I changed kwargs to options dict, is to know and hard code the number of parameters that is passed to vmapped function. If there is a cleaner way, I can also switch to that.

desc/magnetic_fields/_core.py
dpanici: use less steps here, let particle tracing have different default

desc/magnetic_fields/_core.py
dpanici: add private, jit-compatible only expecting necessary diffrax objects and arguments (more similar to calling diffrax directly, just without specifying the odefun), and let public have simpler API

desc/magnetic_fields/_core.py
YigitElma: With the private version, it is not a problem.

desc/magnetic_fields/_core.py
YigitElma: Removed the division by min_step_size

tests/test_magnetic_fields.py
YigitElma: I updated the test, but basically I need to close over some of the variables that are not hashable (declaring static doesn't work). One option could be writing a dummy objective, and calling compute and grad, because basically what I care is not recompiling the objective function.

desc/plotting.py
```diff
@@ -2042,14 +2044,30 @@ def poincare_plot(
...
+        options={"throw": False},
+        return_aux=True,
         **fli_kwargs,
     )
+    # result._value is 0 if integration completed successfully
+    # for a field line, and >0 if it failed or hit bounds.
+    if any(result._value > 0):
+        err0_idx = np.where(result._value > 0)[0][0]
+        # derived from https://github.com/patrick-kidger/equinox/pull/1102/files
+        err0_msg = np.vectorize(lambda val: RESULTS._index_to_message[val])(
```
YigitElma, line 2066: This looks annoying, but it seemed to be the only way to access the error message without limiting ourselves to latest `equinox` version

desc/magnetic_fields/_core.py
```diff
@@ -2614,96 +2618,205 @@ def field_line_integrate(
...
+    saveat : diffrax.SaveAt, optional
+        SaveAt object to specify at which points to save the results of the
+        integration. Defaults to saving at all phis.
+    event : diffrax.Event, optional
+        Event object to specify when to stop the integration. Defaults to stopping
+        when the trajectory leaves the bounds_R and bounds_Z bounding box.
+    adjoint : diffrax.AbstractAdjoint, optional
+        How to take derivatives of the trajectories. ``RecursiveCheckpointAdjoint``
+        supports reverse mode AD and tends to be the most efficient. For forward mode AD
+        use ``diffrax.ForwardMode()``.
```
ddudt: Remove the "optional" tag and also remove the default value descriptions. These arguments are not optional and do not have default values.

ddudt: Could we give default values to things like `chuk_size`, `options`, and `return_aux`?

YigitElma: `options` is a dict (not hashable), Rory mentioned a couple times before to not give default dict values or empty dict. I can set `chunk_size=None` and `return_aux=False`, I am indifferent, but the point of this function to be intimidating if you don't know what exactly you are doing.

ddudt (approved):

desc/magnetic_fields/_core.py
```diff
@@ -2579,13 +2580,14 @@ def field_line_integrate(
     source_grid=None,
     rtol=1e-8,
     atol=1e-8,
-    maxstep=1000,
+    max_steps=None,
     min_step_size=1e-8,
     solver=Tsit5(),
     bounds_R=(0, np.inf),
     bounds_Z=(-np.inf, np.inf),
     chunk_size=None,
```
dpanici, line 2588: This changes a functionality, chunk_size before was for the BS-law computation, now it is over the field lines. I still think one should be able to specify the bs_chunk_size

CHANGELOG.md
```diff
@@ -2,14 +2,20 @@ Changelog
...
 - Adds keyword argument `normalize` to plot_1d, plot_3d. `normalize` is a string to use for normalization.
+- Changes related to ``field_line_integrate``:
+    - `field_line_integrate` now returns `diffrax.diffeqsolve.stats` and `diffrax.diffeqsolve.result` if the flag `return_aux` is set to True.
+    - Renames `maxsteps` argument of `field_line_integrate` to `max_steps`. Now the argument has a consistent meaning with the `diffrax` package and specifies the maximum number of steps allowed for whole integration. Previously, it was used as maximum number of iterations between integration steps.
+    - `field_line_integrate` function doesn't accept additional keyword-arguments related to `diffrax`, if it is necessary, they must be given through `options` dictionary.
+    - ``poincare_plot`` and ``plot_field_lines`` functions can now plot partial results if the integration failed. Previously, user had to pass ``throw=False`` or change the integration parameters. Users can ignore the warnings that are caused by hitting the bounds (i.e. `Terminating differential equation solve because an event occurred.`).
```
dpanici, line 12: right now re-purposing chunk_size is a potentially breaking change. Update this to say it adds new chunk_size or renames it or whatevr you decide to do, once you address my other request for change to add the functionality back.

desc/magnetic_fields/_core.py
```diff
@@ -2614,96 +2619,206 @@ def field_line_integrate(
...
     chunk_size : int or None
-        Size to split computation into chunks of evaluation points.
-        If no chunking should be done or the chunk size is the full input
-        then supply ``None``. Default is ``None``.
-    kwargs: dict
-        keyword arguments to be passed into the ``diffrax.diffeqsolve``
+        Chunk of field lines to trace at once. If None, traces all at once.
+        Defaults to None.
+    bs_chunk_size: int or None
+        Chunk size to use when evaluating the magnetic field. If None, evaluates
```
dpanici: specify it is over src grid

desc/magnetic_fields/_core.py
```diff
@@ -2614,96 +2619,206 @@ def field_line_integrate(
...
-        keyword arguments to be passed into the ``diffrax.diffeqsolve``
+        Chunk of field lines to trace at once. If None, traces all at once.
+        Defaults to None.
+    bs_chunk_size: int or None
+        Chunk size to use when evaluating Biot-Savart for the magnetic field. If None,
+        evaluates all the source grid points at once. Defaults to None.
+    options : dict, optional
+        Additional arguments to pass to the diffrax diffeqsolve. If user wants to
+        provide ``stepsize_controller``, ``saveat``, ``event``, or ``adjoint``, they
+        should use ``_field_line_integrate``.
```
f0uriest: but these are still accepted and handled as options?

desc/plotting.py
f0uriest: I'm fine with this for now, but add a note/todo with the version of equinox this was fixed in, so that we know to remove this once our requirements catch up.

f0uriest (approved):

desc/magnetic_fields/_core.py
YigitElma: Oh sorry, at some point they were not allowed but I forgot to update docs

f0uriest (approved):

dpanici (approved):

PR #1841, opened 2025-08-11
Reuse Z transform for L in from_near_axis

dpanici:
Reuse the Z transform for lambda in `equilibrium.from_near_axis`, is more efficient and requires building fewer large matrices.

### DIFF
--- a/desc/equilibrium/equilibrium.py
+++ b/desc/equilibrium/equilibrium.py
@@ -2079,26 +2079,16 @@ class Equilibrium(IOAble, Optimizable):
             sym="sin" if not na_eq.lasym else False,
             spectral_indexing=spectral_indexing,
         )
-        basis_L = FourierZernikeBasis(
-            L=L,
-            M=M,
-            N=N,
-            NFP=na_eq.nfp,
-            sym="sin" if not na_eq.lasym else False,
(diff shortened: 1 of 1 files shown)


desc/equilibrium/equilibrium.py
```diff
@@ -2079,26 +2079,16 @@ def from_near_axis(
...
-        transform_L = Transform(grid, basis_L, method="direct1")
         A_R = transform_R.matrices["direct1"][0][0][0]
         A_Z = transform_Z.matrices["direct1"][0][0][0]
-        A_L = transform_L.matrices["direct1"][0][0][0]
 
         W = 1 / grid.nodes[:, 0].flatten() ** w
         A_Rw = A_R * W[:, None]
         A_Zw = A_Z * W[:, None]
-        A_Lw = A_L * W[:, None]
+        A_Lw = A_Zw  # can just reuse Z for L, works for both sym and asym cases
```
YigitElma: if there is a memory bottleneck for this function, you can apply the same trick we did for Jacobian, and override A_Z with A_Zw since A_Z is not used anymore

rahulgaur104 (approved):

f0uriest (approved):

PR #1842, opened 2025-08-11
Always re-solve eq in ProximalProjection before first computation of value and jacobian

### LINKED ISSUE #1831: Ensure or recommend that equilibrium is solved (`eq.solve`) before beginning an optimization (`eq.optimize`)
Not sure if this is something we want to automate, but there can be cases where if one does not solve the equilibrium before doing `.optimize`, the results can greatly differ. 

An example of this is if you want to say increase a free-boundary equilibrium's pressure a little bit, then re-solve the free boundary, I've found that because the initial equilibrium is not completely in force balance when you do the .optimize call, it will fail due to a bad approximation error because any subsequent steps it takes then solve the fixed bdry equilibrium, which is then much different than the starting equilibrium.

A schematic script (using the two files [here](https://drive.google.com/drive/folders/1UrlWqeOk2ssGKFlSbHkXnitdp9JbzW-n?usp=sharing)) where the first script here actually works (that is, the free bdry does not immediately stall, and the resulting equilibrium is a slight outwardly shifted equilibrium as expected from the shafranov shift due to increased pressure)

```python
from desc.io import load
from desc.grid import QuadratureGrid
eq_master = load("free_bdry_GS_lowres_master.h5")
field = load("surf_curr_field_GS_lowres_master.h5")

eq = eq_master.copy()
from desc.profiles import ScaledProfile
eq.pressure = ScaledProfile(1.005, eq.pressure)
obj_force = ObjectiveFunction(ForceBalance(eq,grid=QuadratureGrid(eq.L,eq.M_grid,eq.N)))
eq.solve(ftol=1e-6, verbose=3, xtol=1e-8, gtol=1e-10, objective=obj_force)

from desc.objectives import *
from desc.grid import *

freeb_obj = ObjectiveFunction(BoundaryError(eq,field=field,field_fixed=True, field_grid = LinearGrid(M=60,N=0,NFP=64)))
cons = (FixPressure(eq), FixCurrent(eq),FixPsi(eq),ForceBalance(eq,grid=QuadratureGrid(eq.L,eq.M_grid,eq.N)))
eq.optimize(objective=freeb_obj, constraints=cons,verbose=3, ftol=1e-6, options={"solve_options":{"ftol":1e-6,"xtol":1e-8,"gtol":1e-10}} );
```

[log1.txt](https://github.com/user-attachments/files/21540986/log1.txt)

The second script here with the eq.solve taken out instead stalls the free bdry immediately. In this specific case it is due to the objective for free boundary (BoundaryError) not actually being substantially affected by the increased pressure, so the old free bdry equilibrium with the small pressure increase (but unsolved) still has a very low cost. This could, I suppose, be just abuse of the code, really the equilibrium needs to be solved before doing the new free boundary run, but in almost all cases this is true, and so it may be useful to always do an initial solve (say in the build maybe or by not including the initial equilibrium x in the allx list of `ProximalProjection`) to avoid this issue.

```python
from desc.io import load
from desc.grid import QuadratureGrid
eq_master = load("free_bdry_GS_lowres_master.h5")
field = load("surf_curr_field_GS_lowres_master.h5")

eq = eq_master.copy()
from desc.profiles import ScaledProfile
eq.pressure = ScaledProfile(1.005, eq.pressure)
# obj_force = ObjectiveFunction(ForceBalance(eq,grid=QuadratureGrid(eq.L,eq.M_grid,eq.N)))
# eq.solve(ftol=1e-6, verbose=3, xtol=1e-8, gtol=1e-10, objective=obj_force)

from desc.objectives import *
from desc.grid import *

freeb_obj = ObjectiveFunction(BoundaryError(eq,field=field,field_fixed=True, field_grid = LinearGrid(M=60,N=0,NFP=64)))
cons = (FixPressure(eq), FixCurrent(eq),FixPsi(eq),ForceBalance(eq,grid=QuadratureGrid(eq.L,eq.M_grid,eq.N)))
eq.optimize(objective=freeb_obj, constraints=cons,verbose=3, ftol=1e-6, options={"solve_options":{"ftol":1e-6,"xtol":1e-8,"gtol":1e-10}} );
```

[log2.txt](https://github.com/user-attachments/files/21540988/log2.txt)

dpanici:
Resolves #1831


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Bug Fixes
...
 - No longer closes over the field in ``desc.magnetic_fields._core.field_line_integrate``, which can dramatically reduce compile times when the field being traced has large size attributes (for example, when using a ``desc.magnetic_fields._core.SplineMagneticField`` object).
+- When using any of the ``"proximal-"`` optimization methods, the equilbrium is now always solved before beginning optimization to the specified tolerance (as determined, for example, by ``options={"solve_options":{"ftol"...}}`` passed to the ``desc.optimize.Optimizer.optimize`` call). This ensure the assumptions of the proximal projection method are enforced starting from the first step of the optimization.
```
dpanici:
```suggestion
- When using any of the ``"proximal-"`` optimization methods, the equilbrium is now always solved before beginning optimization to the specified tolerance (as determined, for example, by ``options={"solve_options":{"ftol"...}}`` passed to the ``desc.optimize.Optimizer.optimize`` call). This ensures the assumptions of the proximal projection method are enforced starting from the first step of the optimization.
```

tests/test_optimizer.py
```diff
@@ -447,7 +447,13 @@ def compute(self, params, constants=None):
         options={
             "initial_trust_radius": 0.5,
             "perturb_options": {"verbose": 0, "order": 1},
-            "solve_options": {"verbose": 0, "maxiter": 2},
+            "solve_options": {
```
YigitElma, line 451: In line 412 of this test, we use `objective._jac_scaled_error()` but we don't have a method like that, how that works? I have a problem with this function in multigpu PR, I facedd it there

dpanici: huh that would explain why the test does not behave like we expect when it has noisy gradients...

YigitElma: I found that this solves the issues for my case,
```python
    objective = ObjectiveFunction(DummyObjective(things=eq), use_jit=False)
    # we need to build before we declare a new method to properly unjit
    # the objective methods, so that _static_attrs is set correctly
    objective.build()
    # make gradient super noisy so it stalls
    objective.jac_scaled_error = lambda x, *args: objective.jac_scaled_error(
        x
    ) + 1e2 * (np.random.random((objective._dim_f, x.size)) - 0.5)
```

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ Bug Fixes
...
 - No longer closes over the field in ``desc.magnetic_fields._core.field_line_integrate``, which can dramatically reduce compile times when the field being traced has large size attributes (for example, when using a ``desc.magnetic_fields._core.SplineMagneticField`` object).
+- When using any of the ``"proximal-"`` optimization methods, the equilbrium is now always solved before beginning optimization to the specified tolerance (as determined, for example, by ``options={"solve_options":{"ftol"...}}`` passed to the ``desc.optimize.Optimizer.optimize`` call). This ensures the assumptions of the proximal projection method are enforced starting from the first step of the optimization.
```
YigitElma: move this to top

YigitElma (changes requested):
Hmm, I don't see how we can prevent the failing benchmarks since the kwarg is recently added. If you can merge it by overriding somehow, I am fine with it.

Should update the changelog though

YigitElma (approved):

f0uriest (approved):

Issue #891, opened 2024-02-20
generalized radial coordinate


ddudt:
We assume $\psi = \rho^2$, so that $\rho$ is proportional to the minor radius. This is necessary for the Zernike polynomials to properly resolve the magnetic axis. However, it is not necessary far from the axis and might come at the cost of accuracy near the LCFS. 

We should generalize the radial coordinate. One option could be $\psi = \rho^2 + \rho^3 - \rho^4$, so that it still has $\rho^2$ scaling near the axis but also $\psi \propto \rho$ near the boundary for better accuracy there.

ddudt:
I had tried simply changing the definition of $\psi(\rho)$ in the [compute funtion](https://github.com/PlasmaControl/DESC/blob/master/desc/compute/_profiles.py#L34), but that didn't seem to work well for basic tokamak equilibrium solves. It's difficult to compare because changing this definition also changes the initial guess for the flux surfaces.

f0uriest:
We probably don't have to find the "perfect" scaling (and there likely isnt one for all cases). We could just allow Psi(rho) to be a profile like the others, defaulting to rho**2 but in theory allowing users to override that (maybe with some checks to make sure it scales correctly at the axis and is monotonic etc)

dpanici:
Biggest issue is we need to find everywhere that we may assume psi derivs are zero or constant (i.e. assuming the quadratic scaling)

dpanici:
#1015 could remove need for this issue

dpanici:
Use SOLOVEV analytic to test this

dpanici:
Make generic Psi profile

PR #1855, opened 2025-08-14
Add ability to compute iota from field line integration given a magnetic axis

### LINKED ISSUE #491: Add iota calclulation from field line integration of a coilset/magnetic field
When doing field line integration from a magnetic field, would be useful to also get out the iota values. Multiple ways to do this but the easiest to implement initially is probbaly to just do the literal definition for iota,

$$ \iota = \lim_{n\rightarrow \infty} \frac{1}{n} \sum_1^n \Delta \theta_n$$

where $\Delta \theta_n$ is the change in poloidal angle in the $\phi=0$ plane from the $n-1$ toroidal transit to the $n$ th toroidal transit, which could be calculated as $arctan(\frac{Z-Z_{axis}}{R-R_{axis}})$

Where iota is calculated in FOCUS, as an example

https://github.com/PrincetonUniversity/FOCUS/blob/be28c104bebc1b4ea281547b03fc1e9fab34e8e2/sources/poinplot.f90#L113

dpanici:
- Adds iota computation from external fields as an optional output from `field_line_integrate`

Resolve #491 

Math
-----

Adds an extra variable to the state, which is the polar angle $\theta$ of the field line w.r.t. some input magnetic axis $(R_0(s), Z_0(s)$ represented by a DESC ``Curve`` object defined as 

$$
\begin{equation}
\theta = tan^{-1} \left(\frac{Z-Z_0}{R - R_0}\right) =:  tan^{-1} \left(\frac{\Delta Z}{\Delta R}\right)
\end{equation}
$$

We want to essentially keep a running sum of how theta has changed across the field line's trajectory w.r.t., so we choose to follow how FOCUS does it (see #491 ) and find the differential $d\theta/ds$ ($s$ being arclength along field line), then just integrate that along with our field line trajectories (which we need to integrate anyways to compute theta and thus iota).

The differential becomes

$$
\begin{align}
d\theta/ds =& \frac{1}{1+(\Delta Z)^2 / (\Delta R)^2} d/ds (\frac{\Delta Z}{\Delta R})\\
&= \frac{(\Delta R)^2}{(\Delta R)^2+(\Delta Z)^2 } \left( d \Delta Z/ds \frac{1}{\Delta R} - \Delta Z d\Delta R / ds \frac{1}{(\Delta R)^2} \right)\\
&= \frac{1}{(\Delta R)^2+(\Delta Z)^2 } \left(\Delta R \frac{d\Delta Z}{ds} - \Delta Z \frac{d\Delta R}{ds}\right)
\end{align}
$$

Where $d\Delta X/ds = \frac{dX}{ds} - \frac{dX_0}{ds}$ ($X \in \{R,Z\}$), and $dR/ds$ and $dZ/ds$ are just the RHS of the magnetic field line trajectory ODE that we already integrate. To account for the fact that this expression came from a polar angle definition which increases in the opposite direction of DESC's poloidal angle, we must add a negative sign. Then, to account for when $B_{\phi}<0$, we add $sign(B_{\phi})$ (this is needed I believe because at its most basic, iota is defined as $d\chi/d\psi$, so if the geometry all stays the same (i.e. same dR_ds etc) and so does the poloidal field, but the toroidal flux changes sign, so must iota, my tests for the various scenarios seem to confirm this):

$$
\begin{align}
d\theta/ds &= \frac{-sign(B_{\phi})}{(\Delta R)^2+(\Delta Z)^2 }\left(\Delta R \frac{d\Delta Z}{ds} - \Delta Z \frac{d\Delta R}{ds}\right)
\end{align}
$$

At the end, iota is very simply found as $\theta_f/\phi_f$, the final poloidal and toroidal angles, where no further sums or anything are needed as the integration itself took care of the $\sum_N d\theta / \sum_N d\phi$


desc/magnetic_fields/_core.py
```diff
@@ -2621,17 +2623,29 @@ def field_line_integrate(
         Size to split computation into chunks of evaluation points.
         If no chunking should be done or the chunk size is the full input
         then supply ``None``. Default is ``None``.
+    iota: bool
+        Whether or not to also find the rotational transform of each field line. If
+        True, requires axis to be passed in. iota is computed assuming the poloidal
+        angle is increasing counter-clockwise about the axis as the field line
```
dpanici, line 2649: Not completely sure if this description actually describes how we are defining iota. If the field line is pointing opposite the direction of increasing phi, then the toroidal flux is negative, and if the B_poloidal in DESC coords is positive, this results in a negative iota since d\chi/d\psi < 0 (chi increases as psi decreases).

desc/magnetic_fields/_core.py
```diff
@@ -2621,17 +2623,29 @@ def field_line_integrate(
...
     -------
-    r, z : ndarray
+    r, z : ndarray of shape [n_phis, n_initial_points]
         arrays of r and z coordinates of the field line, corresponding to the
         input phis
+    iotas : 1-D ndarray of shape [n_initial_points]
+        if iota=True, the rotational transform computed for each field line traced.
     """
-    r0, z0, phis = map(jnp.asarray, (r0, z0, phis))
+    r0, z0, phis = map(jnp.atleast_1d, (r0, z0, phis))
```
dpanici: Had to change this to always be 1D, otherwise the grabbing of the last theta value later will throw an index error when r0 and z0 are passed in as only 0-D arrays

YigitElma (commented):
Tbh, I would rather have this as a separate function like "iota_from_field_lines"

desc/magnetic_fields/_core.py
unalmis: forget map and just do jnp.atleast_1d(r0, z0, phis)

f0uriest: I think this should be a seperate thing, after `rshape = r0.shape`. That way we still know the original input shape so that we return things in that shape, and adding an extra dimension is only done internally

f0uriest: ie the whole block would be
```python 
r0, z0, phis = map(jnp.asarray, (r0, z0, phis))
assert r0.shape == z0.shape, "r0 and z0 must have the same shape"
errorif(iota and axis is None, ValueError, "Must pass in an axis if iota=True")
rshape = r0.shape
r0 = jnp.atleast_1d(r0).flatten()
z0 = jnp.atleast_1d(z0).flatten()

File: desc/magnetic_fields/_core.py [hunk 3]
```diff
@@ -2658,6 +2671,58 @@ def odefun(s, rpz, args):
             [r * br / bp * jnp.sign(bp), jnp.sign(bp), r * bz / bp * jnp.sign(bp)]
         ).squeeze()
 
+    @jit
+    def odefun_iota(s, rpzt, args):
```
f0uriest: per discussion last week this should probably be moved to a global scope to avoid recompiling each time. Same for the original `odefun`

desc/magnetic_fields/_core.py
```diff
@@ -2658,6 +2671,58 @@ def odefun(s, rpz, args):
...
+        field = args[0]
+        axis = args[1]  # magnetic axis
+        r = rpzt[:, 0]
+        phi = rpzt[:, 1]
+        z = rpzt[:, 2]
+        trans = get_transforms(
+            ["x", "x_s"],
+            axis,
+            grid=Grid(
+                jnp.array([jnp.zeros_like(phi), jnp.zeros_like(phi), phi]).squeeze(),
```
f0uriest: should add a note that we'll need to revisit this after #891

docs/notebooks/tutorials/coil_stage_two_optimization.ipynb
f0uriest, line 1: warning about unequal NFP?

f0uriest, line 1: also the new plot of iota doesn't look that great, do we expect better agreement? Do we need to integrate the fieldlines for more transits?

Maybe its just the axis limits make it look worse than it is?

f0uriest:
OK I think I'm a bit confused about the math here. Suppose we're in PEST coordinates $(\vartheta, \phi)$. Then $\iota = \frac{d\vartheta}{d\phi}$ along a field line. But you're using polar $\theta = \vartheta - \lambda(\theta, \phi)$ for some unknown stream function $\lambda$ (which generally isnt the same as our usual one because our usual theta isn't polar).

You're computing $\frac{d\theta}{d\phi} = \frac{d\vartheta}{d\phi} + \frac{d\lambda}{d\phi} + \frac{d\lambda}{d\vartheta} \frac{d\vartheta}{d\phi} = \iota + \frac{d\lambda}{d\phi} + \iota \frac{d\lambda}{d\vartheta}$

So it seems you're getting some extra stuff on top of iota. I can see that when averaged over a surface, those other bits cancel out, but that may require a loooot more transits than we usually do?

dpanici:
> OK I think I'm a bit confused about the math here. Suppose we're in PEST coordinates ( ϑ , ϕ ) . Then ι = d ϑ d ϕ along a field line. But you're using polar θ = ϑ − λ ( θ , ϕ ) for some unknown stream function λ (which generally isnt the same as our usual one because our usual theta isn't polar).
> 
> You're computing d θ d ϕ = d ϑ d ϕ + d λ d ϕ + d λ d ϑ d ϑ d ϕ = ι + d λ d ϕ + ι d λ d ϑ
> 
> So it seems you're getting some extra stuff on top of iota. I can see that when averaged over a surface, those other bits cancel out, but that may require a loooot more transits than we usually do?

That's right, it is essentially taking the basic definition of iota. Definitely we need enough transits to accurately compute it, though I've not done any studies of convergence

image

unalmis:
> > OK I think I'm a bit confused about the math here. Suppose we're in PEST coordinates ( ϑ , ϕ ) . Then ι = d ϑ d ϕ along a field line. But you're using polar θ = ϑ − λ ( θ , ϕ ) for some unknown stream function λ (which generally isnt the same as our usual one because our usual theta isn't polar).
> 
> > 
> 
> > You're computing d θ d ϕ = d ϑ d ϕ + d λ d ϕ + d λ d ϑ d ϑ d ϕ = ι + d λ d ϕ + ι d λ d ϑ
> 
> > 
> 
> > So it seems you're getting some extra stuff on top of iota. I can see that when averaged over a surface, those other bits cancel out, but that may require a loooot more transits than we usually do?
> 
> 
> 
> That's right, it is essentially taking the basic definition of iota. Definitely we need enough transits to accurately compute it, though I've not done any studies of convergence
> 
> 
> 
> image
> 
> 

Do a convergence study by calling `Bounce2D.plot_theta`. You will see theta(alpha,zeta). can check how long until mean slope matches what you expect

dpanici:
perhaps could be improved by WBA along the orbit: https://arxiv.org/pdf/2403.19003

dpanici:
> > OK I think I'm a bit confused about the math here. Suppose we're in PEST coordinates ( ϑ , ϕ ) . Then ι = d ϑ d ϕ along a field line. But you're using polar θ = ϑ − λ ( θ , ϕ ) for some unknown stream function λ (which generally isnt the same as our usual one because our usual theta isn't polar).
> > You're computing d θ d ϕ = d ϑ d ϕ + d λ d ϕ + d λ d ϑ d ϑ d ϕ = ι + d λ d ϕ + ι d λ d ϑ
> > So it seems you're getting some extra stuff on top of iota. I can see that when averaged over a surface, those other bits cancel out, but that may require a loooot more transits than we usually do?
> 
> That's right, it is essentially taking the basic definition of iota. Definitely we need enough transits to accurately compute it, though I've not done any studies of convergence
> 
> image

image

at least for the tutorial notebook, seems that by ntransit=50 most points are converged but yes other methods may work better than this and converge faster

dpanici:
Try other weighting and EMA, specifically for very irrational iota

PR #1856, opened 2025-08-15
Fixing documentation from recent pull request

unalmis:


desc/compute/_fast_ion.py
```diff
@@ -227,18 +227,15 @@ def _poloidal_drift(data, B, pitch):
 
 @register_compute_fun(
     name="gamma_c",
```
unalmis, line 229: I don't like that this name was chosen since it's really pi/2 gamma_c

rahulgaur104: I'll be more than happy if you can remove the prefactor.

desc/compute/_neoclassical.py
```diff
@@ -126,8 +138,10 @@ def _compute(
         simp=simp,
     )
     out = batch_map(fun, fun_data, surf_batch_size)
-
-    return grid.expand(out)
+    if expand_out:
+        assert out.ndim == 1, "Are you sure you want to expand to full grid?"
```
dpanici, line 142: why this change?

unalmis: We only expand quantities if that expansion were to enable them to broadcast with others on the grid.

rahulgaur104: I need to check this. Because full grid expansion was needed in some places which is why I removed that assert statement.

unalmis: For example, we don't expand the ballooning stability quantities either. It is confusing for the user if the shape `(num rho * num theta * num zeta, num alpha, num pitch)`. The user always wants `(num rho, num alpha, num pitch)`.

rahulgaur104: I think there is case where the output shape is (num_rho, num_alpha, num_wells, num_pitch) which looks like (num_rho, num_alpha, num_zeta, num_pitch) to grid compress and it compresses num_wells. Perhaps the right solution is to add an extra dimension but let me find an example where out.ndim != 1 is needed. 
I guess for this PR we can allow this change.

rahulgaur104:
I wanted to add gamma_c Velasco but I can do that later.

desc/compute/_fast_ion.py
```diff
@@ -227,18 +227,15 @@ def _poloidal_drift(data, B, pitch):
...
-    label=(
-        # 2⁄π · arctan(<v_ds · ∇ψ> ⁄ <v_ds · ∇α>)
-        "\\gamma_c"
-    ),
+    label="\\sum_{w} \\gamma_c(\\rho, \\alpha, \\lambda, w)",
     units="~",
     units_long="None",
     description="Fast ion confinement proxy",
-    dim=1,
```
rahulgaur104, line 237: I guess this is not spatial dimensions anymore.

desc/compute/_fast_ion.py
```diff
@@ -227,18 +227,15 @@ def _poloidal_drift(data, B, pitch):
...
+    dim=3,
     params=[],
     transforms={"grid": []},
     profiles=[],
-    coordinates="r",
```
rahulgaur104, line 241: Not sure this is what we want because the "coordinates" are probably not spatial coordinates.
Can be streamlined later.

desc/compute/_stability.py
```diff
@@ -462,7 +462,7 @@ def reshape(f):
     units="~",
     units_long="None",
     description="Ideal ballooning eigenfunction",
-    dim=4,
```
rahulgaur104, line 465: We should probably create an issue because it is potentially confusing.

rahulgaur104 (dismissed):

desc/compute/_stability.py
f0uriest: What is the 5th dimension?

rahulgaur104: The array of all the eigenfunctions is (rho, alpha, zeta_0, N_zeta-2, N_zeta-2).

rahulgaur104: Because there will be N_zeta - 2 eigenvalue and for each eigenvalue the eigenfunction will be N_zeta-2 points (+ 2 end points for the Dirichlet boundary condition)

desc/compute/_fast_ion.py
unalmis: Yea it's not ideal, but right now it seems "r" means constant over theta and zeta.

desc/compute/_neoclassical.py
unalmis: `grid.compress` and `grid.expand` will always compress and expand the first axis only. It is agnostic to the dimension of the input.

desc/compute/_stability.py
unalmis: I mean technically you can embed this mapping in $R^5$ and not $R^4$

unalmis: so in this case `dim` is consistent with `rank`

dpanici:
I'm a bit confused, we have tests for big Gamma_c right? if the gamma_c was wrong, did those tests get affected? or did the gamma_c that got added only used for plotting and not as dependency for Gamma_c?

rahulgaur104:
> I'm a bit confused, we have tests for big Gamma_c right? if the gamma_c was wrong, did those tests get affected? or did the gamma_c that got added only used for plotting and not as dependency for Gamma_c?

I think we have discussed it before during a triage and a group meeting. Currently, gamma_c is only used for plotting. Using gamma_c in Gamma_c, d, a will require the transforms to be built twice, increasing the optimization complexity(both time and memory).

rahulgaur104:
> > > I'm a bit confused, we have tests for big Gamma_c right? if the gamma_c was wrong, did those tests get affected? or did the gamma_c that got added only used for plotting and not as dependency for Gamma_c?
> > 
> > 
> > I think we have discussed it before during a triage and a group meeting. Currently, gamma_c is only used for plotting. Using gamma_c in Gamma_c, d, a will require the transforms to be built twice, increasing the optimization complexity(both time and memory).
> 
> @rahulgaur104 The bigger issue is that math prevents its reuse. Gamma_c squares gamma_c before summing over wells. gamma_c sums over wells without squaring. These are very different quantities.

If what you are saying is the only limitation, we don't sum over the wells, until after the gamma_c is available to plot_gammac or Gamma_c.
We can sum it in the compute or plotting functions and that will easily fix the problem.

unalmis:
>> we can avoild summing over the wells until after the full gamma_c is available to plot_gammac or Gamma_c. We can sum it in the plotting or and square + sum in compute functions separately and that will easily fix the problem.

Performing any reduction over the wells at a later time will increase the memory required to compute ``gamma_c``.

desc/plotting.py
```diff
@@ -4315,21 +4311,18 @@ def plot_gammac(
...
 
-    # Compute bounce integral
-    theta = Bounce2D.compute_theta(eq, X, Y, rho)
     grid = LinearGrid(rho=rho, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False)
     data0 = eq.compute(
         "gamma_c",
         grid=grid,
-        theta=theta,
+        theta=Bounce2D.compute_theta(eq, X, Y, rho),
         Y_B=Y_B,
```
unalmis, line 4320: The default in this plotting function of `Y_B=24` is too small. `Y_B` should be `~100` or at least `>=2*Y` to be meaningful.

unalmis: I increased resolution at `ku/nufft` thanks to the nuffts..

unalmis: they are like 500x more memory efficient.

dpanici: increasing defaults that rely on other PR should really happen in the other PR, not in this one

dpanici: unless this one requires the other one to be in first now

unalmis: > increasing defaults that rely on other PR should really happen in the other PR, not in this one

That is not what was done. The defaults in this PR are chosen according to this PR and nothing else.

However, I would highly recommend merging the nuffts PR before the next pip release because that JAX memory bug I found destroys the ability the compute this quantity at reasonable resolution.

desc/compute/_fast_ion.py
```diff
@@ -227,18 +227,15 @@ def _poloidal_drift(data, B, pitch):
...
-    label=(
-        # 2⁄π · arctan(<v_ds · ∇ψ> ⁄ <v_ds · ∇α>)
-        "\\gamma_c"
-    ),
+    label="\\sum_{w} \\gamma_c(\\rho, \\alpha, \\lambda, w)",
     units="~",
     units_long="None",
     description="Fast ion confinement proxy",
-    dim=1,
+    dim=3,
```
unalmis:
```suggestion
    dim=2,
```

desc/compute/_neoclassical.py
```diff
@@ -235,7 +249,7 @@ def fun(pitch_inv):
                 bounce.points(pitch_inv, num_well),
                 is_fourier=True,
             )
-            return safediv(H**2, I).sum(axis=-1).mean(axis=-2)
+            return safediv(H**2, I).sum(-1).mean(-2)
```
dpanici, line 252: for future, I like the verbosity of keeping axis there... fine with this but just saying this is not really a needed change

dpanici (approved):

PR #1862, opened 2025-08-19
Fix pip & pytest bug and improve installation

### LINKED ISSUE #1859: Pytest problems
After creating a new DESC environment following the install instructions, attempts to run tests with `pytest` yield the following error.

```
ImportError while loading conftest '/home/kaya/Documents/project/DESC/tests/conftest.py'.
tests/conftest.py:5: in <module>
    import h5py
E   ModuleNotFoundError: No module named 'h5py'
```

Running `pip install h5py` yields

```
Requirement already satisfied: h5py in /home/kaya/miniconda3/envs/desc-env/lib/python3.13/site-packages (3.14.0)
Requirement already satisfied: numpy>=1.19.3 in /home/kaya/miniconda3/envs/desc-env/lib/python3.13/site-packages (from h5py) (2.2.6)
```

Still the first error persists.

Note that when launching a clean python session in this environment, the following imports are successful.
```python
import h5py
import jax
```

Some debugging shows that the issue has nothing to do with `h5py`; indeed in `tests/conftest.py` the following libraries are imported at the top of the file.
```python
import os

import h5py
import jax
import numpy as np
import pytest
from netCDF4 import Dataset
```
If I modify the import order such that `jax` is before `h5py`, then the following error is returned.

```
ImportError while loading conftest '/home/kaya/Documents/project/DESC/tests/conftest.py'.
tests/conftest.py:6: in <module>
    import jax
E   ModuleNotFoundError: No module named 'jax'
```

If I move any other library to be imported before `h5py` or `jax`, they import successfully until one of the import errors with `h5py` or `jax` is hit. I have seen similar issues before, and it is usually related to some import order in `desc.backend`.

unalmis:
- [x] Resolves #1859 . The solution is to add `pytest` to the user installation file, but you guys pushed back on that, so people will have to stumble on to the troubleshooting section if they have the issue.
- [x] Improves installation documentation


requirements.txt
```diff
@@ -12,6 +12,7 @@ orthax <= 0.2.5
 plotly >= 5.16, <= 6.2.0
 psutil <= 7.0.0
 pylatexenc >= 2.0, <= 2.10
+pytest ~= 8.4
```
YigitElma: I don't think this should go here. It is just required for devs

docs/installation.rst
```diff
@@ -3,91 +3,111 @@ Installation
 ============
 
 Follow these instructions to install DESC and its dependencies.
-Note that most of the installation options here do not install JAX with GPU support.
-We do include installation instructions to install JAX with GPU support on some computing clusters that we have tested.
```
YigitElma, line 7: I would prefer to keep these.

docs/installation.rst
```diff
@@ -3,91 +3,111 @@ Installation
...
+        For GPU support, you must install the JAX library as discussed in `JAX installation docs <https://github.com/google/jax#installation>`__.
+        For example, below are the instructions to install on compatible devices with an NVIDIA GPU.
+
+        .. code-block:: sh
+
+            git clone https://github.com/PlasmaControl/DESC.git
+            cd DESC
+            conda create --name desc-env 'python>=3.10, <=3.13'
+            conda activate desc-env
+            sed -i '1 s/^jax/jax[cuda12]/' requirements.txt
```
YigitElma, line 103: You can just `pip install jax[cuda12]` after installation, like Della instructions, or refer to Della instructions, since it is literally the same. I always use that for my laptop with GPU

requirements.txt
unalmis: it is intentional

docs/installation.rst
unalmis: it is moved to gpu install. it is very confusing otherwise

docs/installation.rst
unalmis: 1. pip installing after installs an incompatible version. This is reason we have a requirements file to install compatible versions.
2. Strongly disagree with the expectation that the user should just know that to use GPU on their computer they should follow the installation instructions for some random particular super cluster called della.

If you want people to install your code correctly you have to be explicit .

dpanici: I agree with Kaya for 2. here, I am ok with another thing that says if you have GPU on laptop you can follow this, as long as we say on which type of architecture it was tested and when.

1. I thought that we tested this and as long as -U is not used, it should just install the same version you have already, which should then work with the other requirements?

docs/installation.rst
dpanici: Maybe a different note saying "- Note that unless explicitly specified in the installation instructions, DESC will not be installed with GPU support " or something like that

requirements.txt
dpanici: why?

docs/installation.rst
unalmis: it is clear: https://desc-docs--1862.org.readthedocs.build/en/1862/installation.html

docs/installation.rst
unalmis: not including -U might prevent upgrading jax. however it can and will still change versions of other packages to install dependencies required for jax[cuda12].

unalmis: and going back to 2. the della instructions are not the same anyway.

requirements.txt
unalmis: the purpose of the pr, see the last section of installation . there is new dicussion in docs there

unalmis: fine i'll move it back because i don't want to test cluster installs. but to compromise i am moving pytest warning up

docs/installation.rst
YigitElma: I don't understand why you always focus on one "or" statement in my comments, referring to della instructions was just a suggestion, take it or not, the point I want to make is not that. In our installation instructions (mostly earlier versions) used very different syntax for dealing jax gpu instructions. Like, writing bash commands to change txt files but in different ways. It would be nice to unify them throughout. Dario previously spent his time to see that della stuff works fine (there is nothing della related there).  

Technically, the biggest difference between a cluster and other laptops in terms of gpu installation is the available cuda libraries. If recent the cuda required by jax is not compatible with their setup, we give downgraded instruction (scipy stuff is to prevent jax internal import errors). For laptops, if there is no existing cuda installations (like clusters), usually there is no problem.

unalmis: If there is something actionable in this comment, please make that clear.

unalmis: @YigitElma I value your input, so I focus on all parts of your statement. This necessarily includes the "or" part.

YigitElma:
> i don't understand why benchmarks, tests, and codecov has to pass for PRs that are marked as only docs/comments.

I couldn't find a way to add label-based exceptions to codecov, so these types of PR's are always expected to be merged by override access.

For memory, I forgot to add this logic to the workflow, if you want you can add it in this PR. Basically add,
```
- name: Filter changes
        id: changes
        uses: dorny/paths-filter@v3
        with:
          filters: |
            has_changes:
              - 'desc/**'
              - 'requirements.txt'
              - 'devtools/dev-requirements.txt'
              - '.github/workflows/memory_benchmark.yml'

        id: check_changes
        run: echo "has_changes=${{ !contains(github.event.pull_request.labels.*.name, 'only-docs-comments') && steps.changes.outputs.has_changes }}" >> $GITHUB_ENV
```

Before setup python step. Then add, 
```
if: env.has_changes == 'true'
```
to other steps. You can look for the workflow yml's of unit/regression tests for implementation.


Once benchmarks start, it will skip actual benchmarks, it will take couple seconds.

docs/installation.rst
YigitElma: Sorry, maybe I didn't get the tone right then. Usually my "or" s have a silent 'whatever find something that works" at the end, so I didn't expect it to be taken very seriously :)

rahulgaur104: I don't like Linux terminal commands as a part of the installation instructions because that assumes that the user is comfortable with linux terminal. 

Also, macOS uses a different type of sed (BSD) which requires a slightly different syntax from GNU sed
`sed -i '' '1 s/^jax/jax[cuda12]/' requirements.txt
`
Have you tested it on a macOS system? Your instructions might fail.

The easiest way would be to "dumb" it down so the instructions are words explaining what to change instead of Linux commands to execute. We assume that the user can code python and is barely familiar with pip and conda and nothing else.

Also, linking to the della (or another cluster) instructions can also be useful if the instructions are exactly the same but that's your choice.

unalmis: Then the logical conclusion of that thought, is that for local computer instructions, we tell users to edit the file via gui, and for clusters (these are always linux anyway) instructions we give the `sed` command. On a cluster it is easy to run a command in the already open terminal. However, many users do not know how to launch a terminal text editor like `nano`, `vim` etc. to edit a file, save it, and exit.

unalmis: I have updated to resolve Rahul's request.

rahulgaur104: I guess if you can find a sed command that works for linux and macOS, it would cover a big chunk of DESC users. 

Most younger people use VScode anyway. In VSCode, after connecting to your favorite cluster, opening and editing a file is trivial and does not require learning vim or nano. Ctrl + Shift + click opens the file and you can edit it as you do it on Windows.

I don't know which option is better. I would say learning and editing using VScode is easier than vim/nano which is easier than checking if your sed works.

rahulgaur104: Having instructions for both macOS and Linux is ok.

rahulgaur104 (approved):
Should cover most users.

docs/installation.rst
```diff
@@ -103,6 +103,9 @@ On Your Local Machine
             sed -i '1 s/^jax/jax[cuda12]/' requirements.txt
             pip install --editable .
 
+        Note that on BSD systems, the ``sed`` command that replaces ``jax`` with ``jax[cuda12]``
+        in the ``requirements.txt`` file is ``sed -i '' '1 s/^jax/jax[cuda12]/' requirements.txt``
```
unalmis, line 107:
```suggestion
        in the ``requirements.txt`` file is ``sed -i '' '1 s/^jax/jax[cuda12]/' requirements.txt``.
```

docs/installation.rst
YigitElma: Comment for future, if we go back to alphabetical ordering for requirements.txt, we should update this instruction to search for a different line.

rahulgaur104: Ah, actually there is a command that should search all the lines instead of a single one. We should probably use that.
This also underlines the point about this process being prone to failure if you haven't accounted for all the possibilities.

`sed -i '' 's/^jax/jax[cuda12]/' requirements.txt
`
This should search all instances of jax.

unalmis: No, I explicitly only search the first line because i do not want to replace `jax-finufft` with `jax[cuda12]-finufft`.

You should always use the more conservative find and replace command, which is what I have done.

unalmis: Likewise I do not want to replace jaxlib etc. It is always best practice to use the more conservative command.

rahulgaur104: I think we have different ideas of robustness. As Yigit said, the moment someone adds a new package to requirements or reorders the package list, delibrately or otherwise, the thing will break.

A better case would be to replace exactly what we want to replace and specify what to ignore. Something like this will work even after reordering while avoiding jax-finufft or any package of the form jax-package

`sed -i '' 's/^jax\([>=<!=~ ]\)/jax[cuda12]\1/g' requirements.txt`

rahulgaur104: But I have already approved it so let's just keep this point in mind.
If it can fail, it definitely will. Matter of time

rahulgaur104: Btw, as far as I know the huge Microsoft + Crowdstrike bug last year was actually a regex bug. So, beware!

dpanici (approved):

Issue #1389, opened 2024-11-18
Make surface integral functionality object-oriented


dpanici:
https://github.com/PlasmaControl/DESC/blob/95612fa625bd5ba939569004c767ed02594586f0/desc/integrals/surface_integral.py#L11

either their own classes, or as methods of grid classes

PR #1866, opened 2025-08-19
Add some missing functions to api

dpanici:
Let me know if any should be private but I think these are useful for users

### DIFF
--- a/desc/integrals/singularities.py
+++ b/desc/integrals/singularities.py
@@ -1029,7 +1029,7 @@ def compute_B_plasma(
     The magnetic field due to the plasma current can be written as a Biot-Savart
     integral over the plasma volume:
 
-    𝐁ᵥ(𝐫) = μ₀/4π ∫ 𝐉(𝐫') × (𝐫 − 𝐫')/|𝐫 − 𝐫'|³ d³𝐫'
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -137,6 +137,12 @@ Integrals
 
     desc.integrals.Bounce2D
     desc.integrals.Bounce1D
+    desc.integrals.compute_B_plasma
+    desc.integrals.surface_averages
+    desc.integrals.surface_integrals
```
unalmis: https://github.com/PlasmaControl/DESC/issues/1389

dpanici: Thanks Kaya, I would still like `compute_B_plasma` documented though

docs/api.rst
```diff
@@ -137,6 +137,7 @@ Integrals
 
     desc.integrals.Bounce2D
     desc.integrals.Bounce1D
+    desc.integrals.compute_B_plasma
```
f0uriest, line 140: we might want to hold off on this for a bit, since I think there may be some overlap or potential confusion with #1847. This method only calculates B_plasma on the plasma boundary, while the stuff in #1847 is more general and would likely make a better public method.

unalmis: I also committed significant effort to add a maintainable API for this stuff in #1807 . So I recommend using that.

YigitElma:
@dpanici If there is no new function (according to comments, they are better off staying private), should we close this? the number of draft PRs is getting out of hand...

PR #1868, opened 2025-08-19
Block jax versions 0.5.1 and 0.5.2 which are broken

unalmis:


unalmis:
i've spammed the usual stickers to pass the tests and it still fails

PR #1869, opened 2025-08-20
Work toward parallelization via sharding (multi-device) composed with vmap (across memory in single device)

unalmis:
Pulls just the code from ku/shard on #1773 that other developers said they were okay with merging on that day.


YigitElma (approved):

unalmis:
Yigit can you override codecov for merge? Thank you

rahulgaur104:
I suspect this to be the reason behind #1925

dpanici:
This is on us, the jax tests all failed. I did not notice that

PR #1871, opened 2025-08-20
Use ScaledProfile in rescale

### LINKED ISSUE #1870: Incorrect scaling with `ScaledProfile` in `desc.compat.rescale`
In `desc.compat.rescale` the profile scalings are generally applied as some factor $x$ multiplying each element of the `eq.pl` array, however this will incorrectly multiply the `scale` parameter of a `ScaledProfile` as well as each parameter of the underlying profile, resulting in a new profile which has been scaled by $x^2$ instead of just $x$

https://github.com/PlasmaControl/DESC/blob/f4891cf779ea95b0a320ff63f8432fd2f01c24da/desc/compat.py#L250

I am not sure of a good solution to this given that the profiles can be arbitrarily nested, but perhaps the `params_dict` of the profile can be inspected, and we can exclude `scale`, or even we can just return a new `ScaledProfile` of the original input profiles to easily deal with all cases correctly.

dpanici:
Resolves #1870 

I chose to just use `ScaledProfile` because we have other profile types where the scaling could be incorrect as well, such as in the case of `PowerProfile` (where the power it is raised to would be scaled, and the sub-profile coefficients would also be scaled, resulting in a completely incorrect rescaling). Using `ScaledProfile` is the only way to get this to work correctly for all profile types.

I could maybe see the case for adding an instance check to see if the profiles are `PowerSeriesProfile` instances, and in those cases doing the multiplication, in order to preserve the current behavior of it not modifying power series profiles, but I would rather just use the `ScaledProfile`.


dpanici:
just multiply profile itself by the scalar

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ New Features
 - Parallelized ideal ballooning stability and Newcomb ballooning metrics and [other improvements](https://github.com/PlasmaControl/DESC/pull/1763).
 - Adds ``FourierXYCoil`` to compatible coils for ``CoilSetArclengthVariance`` objective.
 - Separated ``gamma_c`` calculation from ``Gamma_c``. User can also plot ``gamma_c`` using the ``plot_gammac`` function.
+- ``desc.compat.rescale`` will now return ``ScaledProfile`` instances for most of its profiles, to fix a bug where improper scaling could occur for certain profile types.
```
YigitElma: Move this up

YigitElma (approved):

PR #1880, opened 2025-08-26
Use an orthogonal basis for the straight field line equilibrium

### LINKED ISSUE #1879: Straight field line equilibrium conversion issues
In general, a spectral basis basis is (discretely) orthogonal on a different set of nodes for each coordinate system. In particular, this statement holds for FourierZernike series in DESC and straight field line coordinates.

However, this is overlooked in the routine which converts a DESC equilibrium to straight field line coordinates (function by the name of `to_sfl`). This will handicap the convergence of the basis in PEST coordinates. It may also lead to numerical issues with ill-conditioned inversions. If this has not caused obvious issues so far, then the stream function magnitude must have been small enough that convergence wasn't dramatically affected. (It is often small, but unlikely to always be "small enough".)

The fix is simple.

unalmis:
Resolves #1879 

Previously the basis used to convert to straight field line coordinates was not orthogonal. This handicaps convergence. This is now resolved.

I have marked this as a bug fix because the previous method did not converge well. Oversampling cannot overcome an arbitrarily poor choice for an interpolation grid, especially in 3D. The previous method would solve the equation in part b to O(1). Clearly, even O(1e-7) is bad).

image

### DIFF
--- a/desc/equilibrium/coords.py
+++ b/desc/equilibrium/coords.py
@@ -612,7 +612,7 @@ def to_sfl(
     rcond=None,
     copy=False,
 ):
-    """Transform this equilibrium to use straight field line coordinates.
+    """Transform this equilibrium to use straight field line PEST coordinates.
 
     Uses a least squares fit to find FourierZernike coefficients of R, Z, Rb, Zb
     with respect to the straight field line coordinates, rather than the boundary
--- a/desc/equilibrium/equilibrium.py
+++ b/desc/equilibrium/equilibrium.py
@@ -1444,7 +1444,7 @@ class Equilibrium(IOAble, Optimizable):
         rcond=None,
         copy=False,
     ):
-        """Transform this equilibrium to use straight field line coordinates.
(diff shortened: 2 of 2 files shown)


desc/equilibrium/coords.py
```diff
@@ -651,67 +652,56 @@ def to_sfl(
...
-    del R_sfl_transform  # these can take up a lot of memory so delete when done.
+    eq_PEST.Z_lmn = Transform(
+        grid_PEST,
+        eq_PEST.Z_basis,
+        build=False,
+        build_pinv=True,
+        rcond=rcond,
+    ).fit(data["Z"])
+
+    eq_PEST.L_lmn = np.zeros_like(eq_PEST.L_lmn)
```
unalmis, line 680: I retained this logic from `master` for line 680, but I do not understand. Is `eq_PEST.L_lmn` initialized to something other than zero?

dpanici (dismissed):

PR #1881, opened 2025-08-27
Ku/finufft gpu

### LINKED ISSUE #1851: Add tutorial for docker image of mpi and nufft
At the NVIDIA Princeton hackathon, Scot made a docker with `jax-finufft` and `mpi4jax`. We need to add his tutorial to the DESC documentation website so that users can easily set up that docker.

unalmis:
Resolves #1851 . I don't have any stuff from the hackathon, but I spent the afternoon on figuring this out and got this working without docker.


unalmis:
This time the benchmark is real.

PR #1884, opened 2025-08-28
Throw error in `VMECIO.load` if `lrfp=T`

dpanici:
We should not be trying to fit VMEC wouts where poloidal flux was used as the flux label. I at least do not know what is going on there, and these typesof equilibria may have weird behavior as that flag is meant for RFP's where things can happen like toroidal field reversign direction in the plasma


dpanici:
remove the .nc and test

PR #1893, opened 2025-09-02
A variational principle based ideal MHD stability solver and optimizer

### LINKED ISSUE #1790: A finite-n stability solver in DESC?
Relaible and publicly available finite-n plasma stability solvers are not available the magnetic confinement fusion community.
 
I think after #1771 and #1777 are merged, we will have all the tools required to write the Linearized ideal MHD force operator (equation 8.20 in Freidberg's ideal MHD book). If we can do that, we can write our own eigenvalue solver in DESC.

This would essentially make DESC the first open-source ideal stability analysis and optimization tool.

rahulgaur104:
Finite-n stability is important for tokamaks and non-poloidally omnigenous stellarators. Currently, there are no open source codes to perform ideal MHD stability optimization. 

After a preliminary attempt discretizing and working with the ideal MHD force operator in #1791 , I realized that it would be easier to discretize the energy principle instead. Therefore, this PR creates a variational principle based eigenvalue solver and optimizer.

After discretization, the generalized eigenvalue problem becomes

$$A x = \lambda B x $$

where B is symmetric positive definite and A is symmetric. With some efficient linear algebra we cholesky decompose B and convert the problem above into a standard eigenvalue problem.

$$\tilde{A} \tilde{x} = \lambda \tilde{x}$$

Addresses #1790 

# TODOs for the current PR:

- [x] Test different basis for representing radial derivatives. Fourier is fine for toroidal and poloidal derivatives. Finite-difference, or Legendre can be used for radial. Now, we have Chebyshev, Legendre, finite-difference, Bsplines
- [x] Decide the API to use for creating differentiation matrices and ensure that the matrices are not recalculated at each iteration. They only have to be built once and the optimizer should not track their gradients. Tests needed to ensure that.
Since the fundamental matrices themselves are pretty small (< 100 x 100 ~ few MBs), they can be passed via transforms.
- [x] Ensure the mass matrix is symmetric positive definite and the stiffness matrix without the instability term is close to symmetric positive definite (no or very small negative eigenvalues). This can then be corrected using Gershgorin circle theorem by adding a small positive value to the diagonal before adding the instability drive and eigendecomposing.
- [x] Effcient decomposition of the mass matrix. Mass matrix is N x N, so vanilla Cholesky will be O(N^3) which is prohibitive for large N ~ 100 k. However, the structure of the mass matrix allows us to rearrange it so that the cost of doing Cholesky becomes O(N). Section 4.1 of the preprint.
- [x] Enforce incompressibility and make sure it works without polluting the eigenspectrum. Tests needed($\mathbf{\gamma \rightarrow \infty}$ should be equivalent to imposing incompressibility). Observation: Eigenfunctions are smooth when incompressibility is not enforced so perhaps we should optimize without imposing incompressibility (and compare with the incompressible modes post-optimization) 
- [x] New plotting routine needed to plot the eigenfunction and the physical quantities associated with it. Probably needs to be merged in `plot_section`.
- [x] Write a custom, matrix-free eigenvalue solver to allow high-resolution stability calculations. This will save a lot of memory but will take longer. Future PR.
- [x] Generalize the code to study open devices such as magnetic mirrors
- [ ] Compare and benchmark with another finite-n stability solver, possibly TERPSICHORE.
- [x] Compare with NIMSTELL Landreman-Buller-Drevlak QH interchange example from Patil et al. (Figure 4 in the attached preprint below)
- [x] Test the eigenvalue gradient calculation and test the resolution limits that can be fit on a single A100 GPU.
- [x] Study the mode familyN (mod NFP) = 0, we can solve the stability problem in a single-field period, reducing the matrix size $$1/NFP^2$$. Add a separate condition to study  this mode family.

# Preprint with benchmark details
[AGNI: A differentiable MHD stability solver & optimizer for magnetic confinement fusion devices
](https://arxiv.org/html/2608.01750v1)

# Future PRs

- External mode (vacuum) stability calculation. Future work.

# Updated driver scripts
[run_lambda_GJ.py](https://github.com/user-attachments/files/31678163/run_lambda_GJ.py)

## Checkpoints (only for me):
commit 7166c2d9bd1d923e83864e6412b6fde101661b16
commit 28e41c883f6126e6506a52fe07ece8f933f7ce13
commit d3d9039
commit 704adcb22af4252c18bbb75a0c4c8607ff2625ef (before pruning)
commit cacf77de712846d24601fb155deec000add267b8 (before removing adaptive shift-invert)


rahulgaur104:
Note that the PR is not actually 6k lines. I have multiple slightly different copies of the same solver. The final PR will be ~ 2k lines.

unalmis:
> Finite-difference, chebyshev, or Legendre can be used for radial.

one way to avoid the clustering of Chebyshev nodes at the origin of the polar disk is to do Chebyshev interpolation in the variable `y(x)` instead of the Chebyhsev points in `x`. The function `y` is defined below.
https://github.com/PlasmaControl/DESC/blob/65f8c3549cf90e0dbccf89e7a51f35fbdae1379a/desc/integrals/quad_utils.py#L38

desc/diffmat_utils.py
```diff
@@ -0,0 +1,471 @@
...
+    return D
+
+
+########################################################################
+# ---------------------- LEGENDRE MATRICES --------------------------- #
+########################################################################
+
+
+@partial(jit, static_argnums=0)
+def legendre_lobatto_nodes(N: int) -> jnp.ndarray:
```
unalmis: The function `leggauss_lob` in `desc.integrals.quad_utils` already exists and is tested.

desc/diffmat_utils.py
```diff
@@ -0,0 +1,471 @@
...
+    k = jnp.arange(1, N - 2 + 1, dtype=jnp.float64)  # 1 … N−3  (empty if N≤3)
+    beta = jnp.sqrt(k * (k + 2) / ((2 * k + 1) * (2 * k + 3)))  # new formula
+    T = jnp.diag(beta, 1) + jnp.diag(beta, -1)  # (N−2)×(N−2)
+    eigs = jnp.linalg.eigh(T)[0] if (N > 3) else jnp.array([])  # no eigs if <4
+
+    return jnp.concatenate((-jnp.ones((1,)), jnp.sort(eigs), jnp.ones((1,))))
+
+
+@partial(jit, static_argnums=0)
+def legendre_lobatto_weights(N: int) -> jnp.ndarray:
```
unalmis: The function `leggauss_lob` in `desc.integrals.quad_utils` already exists and is tested.

PR #1894, opened 2025-09-02
Patch to work around discretization error impeding optimizer

unalmis:
A computationally inferior approach for the alternative style free surface optimization in `dp/laplace` is implemented in this PR to appease the optimizer. The fact that this approach works, while the original in `dp/laplace` does not, along with other support from debugging that is not detailed here, implies that the discretization error of the quadrature causes the optimizer to fail in `dp/laplace`. 

The only thing that is changed is that |B_out| is computed from equation 4.26 of the shared paper instead of the equation sandwiched between 4.23 and 4.24. That is, on `dp/laplace` once $\Phi$ is known, |B_out| is computed directly via Fourier differentiation of the $C^{\infty}$ potential. On this branch, |B_out| is instead backed out from another singular integral transform of $\Phi$ because using that integral transform suppresses the discretization error of the original. This approach currently seems necessary because otherwise the discretization error for the high order polar quadrature is larger than the change in the potential that arises from the optimizer's perturbation of the boundary. I don't expect this to have ever been an issue to debug for forward-style free surface optimization because there the discretization error of $B \cdot n$ is a subset of the objective which is minimized by the optimizer.

This is a patch until a quadrature whose error is less strongly dependent on the boundary surface and integrand is added which should help resolve the issue. I also believe #1208 will be useful to resolve.

Vacuum free surface solutions can be obtained with this patch that could not be obtained on `dp/laplace`.
vacplot

[code for above optimization](https://github.com/user-attachments/files/22105729/free_surf_discretization_patch.py)


unalmis:
@dpanici can you share what you have tried for finite beta

dpanici:
[run_laplace_finite_beta_maxiter.nbconvert.ipynb](https://github.com/user-attachments/files/22124650/run_laplace_finite_beta_maxiter.nbconvert.ipynb)
this notebook shows that even starting from the free-boundary solution from master, using this patch and the neumaan method results in the boundary moving far from the nominal solution

I've not done a lot of tweaking of code yet to try to debug why this is happening but I've confirmed this behavior also on current-constrained tokamak equilibria with coils (as opposed to this notebook which is iota constrained and external field from mgrid)

dpanici:
TODO
- [ ] vacuum free bdry only works for SourceFreeField right now (bugs in obj logic with FreeSurfaceOuterField)
- [ ] only works for same eval and source grid (bugs in logic when the two grids are different)

dpanici:
the gradients if I use double the eval and source grid resolution (also maybe to be clearer, I set maxiter=0 in each case shown here to avoid any conflation with the iterative fixed point method)

The higher grid resolution seems to make things agree closer between the two, especially for the stalled where this actually matters (the correctly solved eq I am not surprised gradient increased, it is now looking at eval points it did not see in the actual optimization)

However running an optimization with higher grid resolutions with the obj from #1360 does not actually fix the stalling issue

image

unalmis:
> (also maybe to be clearer, I set maxiter=0 in each case shown here to avoid any conflation with the iterative fixed point method)

Both the original and patched version use iterative method to compute Phi, so that should not be necessary.

dpanici:
> What I gather from these plots is more confirmation of the issue described in the PR header. I still think switching the quadrature to the method in the overleaf will help solve this and will be more performant.
> 
> However it may be worth checking again whether `I` is still identically zero at all the points of these plots? I wonder if the issue for the finite beta case is that somehow by passing `I` as a `float` into the compute function and multiplying by the numpy array `theta` to compute `I * data["theta"]` that JAX auto diff looses track of an important gradient. If so we should make a PR that casts the theta and zeta compute quantities on master to jax arrays to avoid such issues.

If that was the case I would think that it would also be problematic for optimizations with `FourierCurrentPotentialField` objects, but I have done that just fine in the past. `I` is zero for the equilibria plotted here (or at least very close to zero since it is computed as the integral of B_theta)

unalmis:
Yes just |B_out| dot n alongside Bout^2

unalmis:
> > Yes with B_out dot n as another residual alongside pressure balance
> 
> 
> 
> Isn't this basically like what the patch is doing? Instead of using Phi we use grad(Phi) and a singular integral to evaluate grad(Phi) on the boundary?

No, both methods use n x grad phi to compute B_out. On `dp/laplace` the potential cannot have B dot n error regardless of discretization error. The patched method computes grad phi via an integral expression which will always have some B dot n error

dpanici:
> Perhaps additional evidence that `I` is the likely culprit for the finite beta not working well: this is a zero net tor current (so I~0), finite beta ellipse which exhibits strong shafranov shift despite the modest beta of 0.25%. The patch objective works fine for this (once a quadgrid is used for the forcebalance subobj at least, which I think may be a nuance of this eq and IC as even the usual free bdry obj needs a quadgrid for force balance to be used to converge correctly as well)
> 
> image
> Shown against some of the vacuum flux surfaces to show that this actually was a finite beta free bdry solve that was non-trivial:
> 
> image
> This was ran on a stellar GPU on `dp/debug-laplace` but I think I dont use anything that is not on this patch PR
> 
> [laplace_finite_beta_zero_current.zip](https://github.com/user-attachments/files/22176470/laplace_finite_beta_zero_current.zip)

Note that this also seems to work (or at least gets very close) with the `dp/laplace` implementation (i.e. not using an additional biot-savart integral to compute |B_out|^2)

image

dpanici:
> > Perhaps additional evidence that `I` is the likely culprit for the finite beta not working well: this is a zero net tor current (so I~0), finite beta ellipse which exhibits strong shafranov shift despite the modest beta of 0.25%. The patch objective works fine for this (once a quadgrid is used for the forcebalance subobj at least, which I think may be a nuance of this eq and IC as even the usual free bdry obj needs a quadgrid for force balance to be used to converge correctly as well)
> > image
> > Shown against some of the vacuum flux surfaces to show that this actually was a finite beta free bdry solve that was non-trivial:
> > image
> > This was ran on a stellar GPU on `dp/debug-laplace` but I think I dont use anything that is not on this patch PR
> > [laplace_finite_beta_zero_current.zip](https://github.com/user-attachments/files/22176470/laplace_finite_beta_zero_current.zip)
> 
> Note that this also seems to work (or at least gets very close) with the `dp/laplace` implementation (i.e. not using an additional biot-savart integral to compute |B_out|^2)
> 
> image

some weird behavior tho, using M=N=20 for both grids seems to give better results than using M=N=30 or 40, for the same Phi resolution of M=N=10. I would naively expect that the solution should stay similar if the grid resolution increases but the basis resolution stayed the same. I will think about this more

unalmis:
I suggest checking the automatic differentiation derivatives of the objective against finite differences with https://docs.jax.dev/en/latest/jax.test_util.html.

PR #1900, opened 2025-09-08
Add Publication List

### LINKED ISSUE #1838: Make list of publications that use the code
Could be interesting to have a place in the docs where we cite all the works that use the code, as a sort of showcase of what it is being used for.

singh-jaydeep:
- Creates file pubs_list.rst listing publications which use DESC (up to 9/8/25)
- Adds pubs_list.rst to the table of contents in index.rst

Resolves #1838

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -56,7 +56,7 @@ jobs:
         run: echo "has_changes=${{ !contains(github.event.pull_request.labels.*.name, 'only-docs-comments') && steps.changes.outputs.has_changes}}" >> $GITHUB_ENV
 
       - name: Set up Python
-        uses: actions/setup-python@v5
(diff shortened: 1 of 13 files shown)


rahulgaur104:
Did the Type One paper actually use DESC?

Also, I think it would be even better if we can put a link to the publication list in README.rst and/or the docs.
Otherwise, it would be left to be explored by the users.

f0uriest:
Agree about adding the list to the readme, maybe just as a direct `..include`? or maybe formatted as a table like the one shorter list there now?

Also wondering if we want to split the list into "papers about desc" ie ones by us that describe new features being added, vs "papers that use desc" to get other results but don't really add anything new to the code.

papers about desc, numbered from the existing list: 1, 3-10, 12, 13, 15, 19

Issue #1800, opened 2025-06-30
Add `normalize` to `plot_1d` and `plot_3d`


YigitElma:
Same as `plot_2d` and `plot_section`. 

https://github.com/PlasmaControl/DESC/pull/1683#pullrequestreview-2972444588

PR #1905, opened 2025-09-09
Add ```normalize``` option in ```plot_1d```, ```plot_3d```

singh-jaydeep:
Add Normalize Option in 1D, 3D Plots
- Adds "normalize" parameter to plot_1d, plot_3d with
  default None
- If "normalize" is passed to function, rescales plotting data
  by grid average of the "normalize" quantity. Adjusts y-axis
  label in plot_1d, and colorbar label in plot_3d
- ~~Copies code from plot_2d into plot_1d, plot_3d which
  warns users that "norm_F" keyword is deprecated~~
- Adds #noqa C901 to plot_1d, plot_3d function signature
  to avoid Flake8 warning
- Adds unit test for plot_1d and plot_3d
- Updates changelog

Resolves https://github.com/PlasmaControl/DESC/issues/1800

Uncertainties:
- should there be any restrictions on the quantity passed in for "normalize"
- ~~Is the deprecation warning for "norm_F" needed in plot_1d, plot_3d~~
- should units be checked for 3d colorbar (e.g. if data/normalize 
   isn't dimensionless)

### DIFF
--- a/desc/transform.py
+++ b/desc/transform.py
@@ -482,7 +482,7 @@ class Transform(IOAble):
                 "Transform must be built with transform.build_pinv() before being used"
             )
 
-        if self.method == "direct1":
+        if self.method in ["direct1", "jitable"]:
             Ainv = self.matrices["pinv"]
             c = jnp.matmul(Ainv, x)
         elif self.method == "direct2":
@@ -532,7 +532,7 @@ class Transform(IOAble):
                 )
             )
 
-        if self.method == "direct1":
+        if self.method in ["direct1", "jitable"]:
             A = self.matrices["direct1"][0][0][0]
             return jnp.matmul(A.T, y)
 
@@ -719,7 +719,10 @@ class Transform(IOAble):
 
     @property
     def method(self):
-        """{``'direct1'``, ``'direct2'``, ``'fft'``}: method of computing transform."""
+        """{``'direct1'``, ``'direct2'``, ``'fft'``, ``'jitable'``}.
+
+        Transform compute method.
+        """
         return self.__dict__.setdefault("_method", "direct1")
 
     @method.setter
--- a/tests/test_transform.py
+++ b/tests/test_transform.py
@@ -491,13 +491,18 @@ class TestTransform:
         assert t.method == "direct1"
 
     @pytest.mark.unit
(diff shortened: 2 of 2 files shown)


desc/plotting.py
```diff
@@ -556,6 +560,28 @@ def plot_1d(eq, name, grid=None, log=False, ax=None, return_data=False, **kwargs
         plot_1d(eq, 'p')
 
     """
+    normalize = parse_argname_change(normalize, kwargs, "norm_name", "normalize")
```
YigitElma: You can delete these depreciation checks since this function never had `normalize` or `norm_F` before

desc/plotting.py
```diff
@@ -996,6 +1044,28 @@ def plot_3d(
         fig = plot_3d(eq, "|F|", log=True, grid=grid)
 
     """
+    normalize = parse_argname_change(normalize, kwargs, "norm_name", "normalize")
```
YigitElma: Again, we can delete these since they were not an option before

desc/plotting.py
```diff
@@ -556,6 +560,28 @@ def plot_1d(eq, name, grid=None, log=False, ax=None, return_data=False, **kwargs
...
+                + "`name=|F|_normalized` instead. If you want to normalize by "
+                + "another quantity, use the `normalize` keyword argument."
+            )
+        )
+        if normalize is None and norm_F:
+            # replicate old behavior before #1683
+            normalize = "<|grad(|B|^2)|/2mu0>_vol"
+        elif normalize is not None and norm_F:
+            raise ValueError("Cannot use both norm_F and normalize keyword arguments.")
+    errorif(
```
YigitElma, line 563: Just this error is necessary

YigitElma (commented):
Thanks for the contribution!

As you mentioned in the PR description, we don't need depreciation warnings as they were not an option before.

Couple other points;
- Need to add `normalize` to Changelog
- Need to add tests

For the test of `plot_1d`, you can take a look at `test_fsa_F_normalized`. The test case could have name="<|B|>" (flux surface averaged B magnitude) and the normalization can be normalize="<|B|>_vol" (volume averaged B magnitude).

For `plot_3d`, you can take a look at `test_3d_tz`

PR #1906, opened 2025-09-10
Ku/nufft notebook

unalmis:
Reruns the optimization to see changes due to #1842.


docs/notebooks/tutorials/EffectiveRipple.ipynb
unalmis, line 1: Attached is link to web diff with the notebook ran on CPU vs a GPU [diff.html](https://github.com/user-attachments/files/22250253/diff.html). The diff you see in this PR is that GPU vs CPU from before the prox projection.

dpanici: what do you mean by before the prox projection? you did not run the cells before prox projection?

unalmis: you merged a pr that changed optimizer. the ci doesn't update the notebooks on each merge to master. the changes you see are a result of that prox projection pr.

this is separate pr to avoiding misleading people into thinking changes are from using nuffts

dpanici: Got it, right

Issue #1899, opened 2025-09-07
issues computing/plotting eps_effective and Gamma_c


f0uriest:
I was under the impression that this should work with some sensible defaults for unspecified parameters:

```python
eq = desc.examples.get("SOLOVEV")
grid = Grid.create_meshgrid([np.array([0.5]), np.linspace(0, 2*np.pi, 20), np.linspace(0, 2*np.pi, 20)], coordinates="raz")
eq.compute("effective ripple 3/2", grid=grid)
```

but I get an error that `theta` needs to be passed as a kwarg. Can't that be determined automatically from the given grid?

Similarly, plotting it also fails

```python
plot_1d(eq, "effective ripple 3/2")
```
this errors because the grid can't fft2

or 
```python
grid = Grid.create_meshgrid([np.array([0.5]), np.linspace(0, 2*np.pi, 20), np.linspace(0, 2*np.pi, 20)], coordinates="raz")
plot_1d(eq, "effective ripple 3/2", grid=grid)
```
this errors because the grid poloidal coordinate isn't theta

The same issue occurs for `Gamma_c`

unalmis:
The quantities you requested to compute are computed in DESC coordinates, but you have given a grid in field line coordinates.

Regarding the `theta` kwarg, I believe we had chatted about this a long time ago before introducing these quantities to master. If I recall correctly, we decided ir needs to be passed as a kwarg because computing it required a coordinate mapping, and there is no public infrastructure in DESC to compute a coordinate mapping without an equilibrium object.

unalmis:
If you prepend "old" to those quantity names it will work.

rahulgaur104:
> I'd like to figure out what we need to fix so that calling eq.compute with any grid "just works". That should also fix the issues with plotting since all the plot routines under the hood just call eq.compute with some custom grid

This will require setting a bunch of defaults that specifying the pitch angle resolution, number of wells to include, Fourier-Chebyshev quadrature required to fit |B| in alpha, zeta space, and quadrature resolution for each bounce integral. These quantities are needed to calculate the bounce integral.
Not sure if choosing these defaults is a great idea because there aren't any standard values and hiding this process behind a command like eq.compute may encourage users to use to ignore many of the important details of the calculation.

 test_effective_ripple_1D calculates ripple using the old grid so it should with with an "raz" grid. Similarly, look test_Gamma_c_Nemov_1D and test_Gamma_c_Velasco_1D.

unalmis:
[...]

PR #1915, opened 2025-09-16
Add pass through kwargs from plotting to compute

### LINKED ISSUE #1352: Allow plotting utitilities to accept compute quantity hyperparameters

f0uriest:
- Only adds `compute_kwargs` option to "generic" plotting functions `plot_{1d,2d,3d,section,fsa}`, the other ones generally already accept kwargs relevant to their function (eg `plot_boozer_modes`)
- No tests yet, as I can't think of any that would be meaningful until #1899 is resolved.

Resolves #1352


dpanici:
> #1899

I mean if you did `theta=...` in that call, does it work?

unalmis (approved):

dpanici:
> > #1899
> 
> I mean if you did `theta=...` in that call, does it work?

FWIW this works (added after the notebook cell for effective ripple where precise QH eps is computed)
```python
from desc.plotting import *
fig,ax,data=plot_1d(eq0,"effective ripple",grid=grid,
    compute_kwargs=dict(
    theta=theta,
    Y_B=Y_B,
    num_transit=num_transit,
    num_well=num_well,
    num_quad=num_quad,
    num_pitch=num_pitch),return_data=True)
ax.scatter(rho,eps)
```

dpanici (dismissed):

desc/plotting.py
```diff
@@ -614,7 +637,12 @@ def plot_1d(  # noqa : C901
...
-        eq, name, grid, kwargs.pop("component", None), reshape=False
+        eq,
+        name,
+        grid,
+        kwargs.pop("component", None),
+        reshape=False,
+        compute_kwargs=compute_kwargs,
     )
 
     if normalize:
```
YigitElma, line 648: Any reason for not including it here too?

rahulgaur104 (dismissed):

desc/plotting.py
```diff
@@ -1413,7 +1485,13 @@ def plot_fsa(  # noqa: C901
         plot_data_ylabel_key = f"<{name}>_fsa"
 
     if normalize:
-        norm_data = _compute(eq, normalize, grid, reshape=False)[0]
+        norm_data = _compute(
```
YigitElma, line 1490: I meant this and the next one, they have the kwargs

YigitElma (approved):

PR #1916, opened 2025-09-16
Magnetic Diagnostics API

### LINKED ISSUE #1783: API for `MagneticDiagnostic` Class
We may want to think of having an overrarching "MagneticDiagnostic" objective that tries to minimize the re-computation of the magnetic field (think, concatenate all of the eval coords for both point B measurements for B probes, as well as the points along the curves for flux/mirnov and partial/full rogowski loops, to better take advantage of vectorization. 

Something I would sort of like is to have some sort of wrapper `MagneticDiagnostic` around the sub-diagnostic objectives, whose only job is to
- take in the source resolutions for the external field and the plasma-generated field
- receive from the sub-objectives what evaluation coordinates in real space they require 
- concatenate these coordinates together into a single `(n, 3)` array (while keeping track of which parts are needed by which sub-objective)
- compute the plasma-generated magnetic field and the external magnetic field at these points
- send the necessary info for each sub-diagnostic so it can compute its own error
 - maybe with the `._compute(params, field)` method outlined below
- concatenate these errors and return as one single array 
  - (may be though annoying if 

I think that maybe one potential way could be for reconstruction objectives to have two compute functions, 
- one which is the usual `objective.compute` and takes in params only (and constants until after #1758 )
- another which takes in the params needed for this as well as the total magnetic field evaluated at the physical coordinates it needs, and it then uses this already-computed field to do whatever integrations it needs. something like just  `._compute(params, field)`
  - This may be need some extra logic with the current idea I have for `FluxLoop` , which takes in an entire coilset of the individual mirnov/flux loop diagnostics, as we need to take the coords of each coil and flatten them to send to the `MagneticDiagnostic` class, then when it receives the coords it will have to reshape them and re-split them up according to the individual loops.
 
 (may be kind of annoying though as if say you included a rogowski loop and a point B measurement, if we allowed diagnostic positions (like rog coil shape) to change, we would have to track within this objective which params go to where. Not a big issue if diagnostics are assumed fixed though)

Something that should be checked with all this is if it is worth it, that is, it is not like these are singular integrals, the only thing this would do is ensure that the maximum vectorization possible is being carried out (i.e. the largest single array eval_coords is being computed). However I think for Bplasma computations, the sources will either be 2D surface currents (in case of using virtual casing principle) or 3D volume currents (in case of #996 ), both of which quickly require large amounts of memory for even modest numbers of eval points.

dpanici:
Adds new class `AbstractDiagnostic` which is a new base class for synthetic diagnostic objects.

- `AbstractDiagnostic` realizations must
  - define the points at which they may require the total magnetic field (as determined by Bplasma + Bcoil if outside of the plasma
    - TODO: Might need a separate class for objects which need the INTERNAL magnetic field, but I actually think the current API would work as the internal magnetic field can be computed as like auxiliary data with the `_compute_data` method already in this
  - as well as implement a compute method for computing the signal they represent
  - implement a `_compute` method which takes in the magnetic field (or vector potential) at the points they require (for e.g. integration over a rogowski coil), as well as a dict of aux data
  - a `_compute_data` method for compiting this aux data from `eq_params`, `field_params` as well as its own`diag_params` (here for example a `RogowskiCoil` computes its geometric information. A ThomsonScattering may here compute the intersection of the chord and the plasma, as another example)
  - may have DOFs to be optimized
Current realizations: `RogowskiCoilXXX` for each coil type, `PointBMeasurements` for ptwise B measurements like magnetic probes.

- `DiagnosticSet` objects are collections of `AbstractDiagnostic` realizations
  - mainly operating like `MixedCoilSet` for now, its job is to collect the diagnostics in one object,
  -  collect and concatenate the points at which B is needed for them so that in `MeasurementError` it can be computed efficiently

 TODO

- [ ]  How to easiest pass in target/bounds?
- [ ] Maybe implement a DiagnosticSet that is like a CoilSet too?
- [ ] have diagnostics have two eval_x attrs required, one for eval_B and other for eval_A and update MeasurementError to compute both A and B at these points
- [ ] maybe move the computation of magnetic field to the DiagnosticSet, since some diagnostics may not need the coil field, for instance, if they are internal measurements
- [ ] currently need `_diagnostics` to be a static attr to avoid JAX issues with `MeasurementError` but not sure why, maybe because we are using its method? but that should be fine I thought... so I dont think that is the issue
- [ ] Add method of specifying correlated uncertainties (i.e. generalize W weight matrix from diagonally as it implicitly is now to arbitrary W) within a single diagnostic (like here within a single diagnostic objective)
  - [ ] for `MeasurementError`, allow weights to be a matrix and assume it to be the inverse of the covariance, and Cholesky factor it and make _scale method for this class using that factorization
- [ ] merge this into #1436 so tests can run
- [ ] Add flux loops from branch which has those implemented
  - make them and Rogowski from same parent class 

Resolves #1783

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
   - package-ecosystem: "pip"
     directory: "/" # Location of package manifests
     schedule:
-      interval: "weekly"
+      interval: "daily"
     labels: ["dependencies", "skip_changelog", "run_benchmarks"]
     groups:
       # group similar dependencies together into fewer PRs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,14 @@ New Features
         - ``VacuumGuidingCenterTrajectory`` : Integrates the particle motion by vacuum guiding center ODEs, conserving energy and mu.
     - Particle trajectories can be plotted with ``desc.plotting.plot_particle_trajectories`` function.
 - Adds new option for ``loss_function``, ``"sum"``, which changes an objective to target the sum of the values computed.
-- Adds utility functions ``desc.external.export_surface_to_paraview``, ``desc.external.export_volume_to_paraview`` and ``desc.external.export_coils_to_paraview`` to export Paraview files for surfaces, volume and coils. These functions use an optional dependency ``pyvista`` which is not automatically installed.
+- Adds utility functions ``desc.external.paraview.export_surface_to_paraview``, ``desc.external.paraview.export_volume_to_paraview`` and ``desc.external.paraview.export_coils_to_paraview`` to export Paraview files for surfaces, volume and coils. These functions use an optional dependency ``pyvista`` which is not automatically installed.
 - The `x_scale` option for `eq.optimize` and related functions can now be given as a dictionary mapping individual parameter names to their relevant scales,
 or if multiple things are being optimized, `x_scale` can be a list of dict, one for each optimizable thing.
 - Adds new option `x_scale='ess'` to use exponential spectral scaling from (Jang 2025) which has been shown to improve performance and robustness as an
 alternative to fourier continuation methods.
 - Adds ``"scipy-l-bfgs-b"`` optimizer option as a wrapper to scipy's ``"l-bfgs-b"`` method.
+- Adds ``check_intersection`` flag to ``desc.magnetic_fields.FourierCurrentPotentialField.to_Coilset``, to allow the choice of checking the resulting coilset for intersections or not.
+- Changes the import paths for ``desc.external`` to require reference to the sub-modules.
 
 Bug Fixes
 
@@ -46,6 +48,10 @@ New Features
     - `field_line_integrate` function doesn't accept additional keyword-arguments related to `diffrax`, if it is necessary, they must be given through `options` dictionary.
     - ``poincare_plot`` and ``plot_field_lines`` functions can now plot partial results if the integration failed. Previously, user had to pass ``throw=False`` or change the integration parameters. Users can ignore the warnings that are caused by hitting the bounds (i.e. `Terminating differential equation solve because an event occurred.`).
     - `chunk_size` argument is now used for chunking the number of field lines. For the chunking of Biot-Savart integration for the magnetic field, users can use `bs_chunk_size` instead.
+- Adds support for optimization targeting individual coils in a coilset.
+  - Coil objectives accept pytree inputs for `target`, `bounds`, and `weight`.
+  - Able to set weights to zero, excluding certain coils from the objective.
+
 
 
 Bug Fixes
--- a/desc/backend.py
+++ b/desc/backend.py
@@ -88,6 +88,7 @@ if use_jax:  # noqa: C901
     from jax.scipy.special import gammaln
     from jax.tree_util import (
         register_pytree_node,
+        tree_broadcast,
         tree_flatten,
         tree_leaves,
         tree_map,
@@ -620,6 +621,10 @@ else:  # pragma: no cover
 
     def treedef_is_leaf(*args, **kwargs):
         """Check is leaf of pytree for numpy backend."""
+        raise NotImplementedError
+
+    def tree_broadcast(*args, **kwargs):
+        """Broadcast pytree for numpy backend."""
         raise NotImplementedError
 
     def register_pytree_node(foo, *args):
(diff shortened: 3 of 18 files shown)


desc/objectives/_reconstruction.py
```diff
@@ -1,18 +1,56 @@
...
 from desc.grid import LinearGrid
-from desc.utils import Timer, dot, errorif
+from desc.utils import Timer, dot, errorif, warnif
 
 from .normalization import compute_scaling_factors
 from .objective_funs import _Objective, collect_docs
 
 
-class PointBMeasurement(_Objective):
+class _MagneticMeasurement(_Objective, ABC):
```
f0uriest: For new stuff I've been pushing to name things like `AbstractMagneticMeasurement` rather than an underscore to make clear that its an ABC, and that it maybe semi-public (if users want to add their own subclasses)

desc/objectives/_reconstruction.py
```diff
@@ -1,18 +1,56 @@
...
+        Parameters
+        ----------
+        use_jit : bool, optional
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        super().build(use_jit=use_jit, verbose=verbose)
+        assert hasattr(self, "_all_eval_x_rpz"), (
```
f0uriest: `_all_eval_x_rpz` might be better as an abstract property/attribute itself?

Also, the variable name implies it but should document somewhere that `x` here is in R,phi,Z. And that the compute function should expect B in R,phi,Z basis as well

desc/objectives/_reconstruction.py
```diff
@@ -1,18 +1,56 @@
...
+            "Subclasses of _MagneticMeasurement must define self._all_eval_x_rpz in"
+            "  their build method!"
+        )
+
+    @abstractmethod
+    def _compute(self, B):
```
f0uriest: is the assumption that these objectives will only ever need B and not any other info about the equilibrium? What about vector potential or current or position vector etc?

desc/objectives/_reconstruction.py
```diff
@@ -344,3 +387,365 @@ def compute(self, eq_params=None, field_params=None, constants=None):
...
+            Whether to just-in-time compile the objective and derivatives.
+        verbose : int, optional
+            Level of output.
+
+        """
+        from desc.magnetic_fields._current_potential import (
+            _compute_A_or_B_from_CurrentPotentialField,
+        )
+
+        self._compute_A_or_B_from_CurrentPotentialField = (
```
f0uriest, line 181: can't this just be a class method?

f0uriest (commented):
Is there still a plan to have classes for different diagnostics? I didn't realize those would each be objectives. I thought it would be something like
```python
class AbstractDiagonstic(ABC, Optimizeable?):
    def compute_measurement(self, eq, field):
        ...
    
    def _compute_from_B(self, B):
        ...
class RogowskiCoil(AbstractDiagnostic):
    ...

class MeasurementError(_Objective):

    def __init__(self, eq, field, diagnostic, target, weight):
        ...

    def compute(self, params):
        B = self.field.compute_magnetic_field(...)
        return self.diagnostic._compute_from_B(B)
```

I like the idea of separate classes for different diagnostics, independent from the `Objective` API for doing analysis etc. The diagnostic only knows about how to compute a synthetic measurement, but nothing about the expected value from experiment. The objective then couples the diagonstic class with the target/bounds/uncertainty from experiment to compute a residual/z score.

Another thing to consider is uncertainties. We can handle simple uncorrelated uncertainty with `weight=1/uncertainty` but often there are off diagonal terms in the covariance matrix that are important. This might require the `Measurement` objective to have special logic in `compute_scaled_error` etc.


At some point we may also want to consider optimizing the placement of diagnostics, this is something that a lot of experimentalists think about and could be a good use of DESC. So then `AbstractDiagnostic` would also inherit from `Optimizeable`

desc/objectives/_reconstruction.py
dpanici: True, vector potential would make sense. Also I guess position vector for other stuff like MSE

desc/objectives/_reconstruction.py
dpanici: I can do that, but it seems that that way does not explicitly require that the attributes themselves be set, only that the properties exist?

Like one still could instantiate a subclass with the property _all_eval_x_rpz but never fill it?

desc/objectives/_reconstruction.py
dpanici: I can't import it without getting a circular import so I have to import it locally. I could do that as a class method I guess

PR #1898, opened 2025-09-05
Add comments about recommeded resolutions to docstrings

f0uriest:
- [ ] @dpanici can you check that the recommendations for coils seem reasonable? Also if there are other `MagneticField` classes that we can give recommendations for
- [ ] @ddudt @rahulgaur104 can you add some recommendations/comments for the `OmnigenousField` and `Omnigenity` objectives
- [x] @unalmis @rahulgaur104 can you add some recommendations/comments for the bounce integral and stability objectives

Resolves #1885

PR #1919, opened 2025-09-18
New inverse stream map to accelerate convergence

### LINKED ISSUE #1303: Differentiable code with dynamic shapes

### LINKED ISSUE #1347: Checkpointing to reduce reverse mode AD memory usage
Right now reverse-mode AD on objectives in DESC that compute a surface average of bounce integrals consumes significant memory.
JAX checkpointing should help resolve this.

### LINKED ISSUE #1388: Move some general interpolation functions into `interpax`

### LINKED ISSUE #1389: Make surface integral functionality object-oriented

### LINKED ISSUE #1864: Memory regression in bounce integrals
Issue documented [here](https://github.com/jax-ml/jax/issues/30627). A workaround is to use a different algorithm to compute bounce integrals which is memory efficient enough such that this issue is less relevant. This is completed in #1834 .

### LINKED ISSUE #1922: Choose stream s.t. the map to the moving mesh of field lines is trivial
Define $\delta = \lambda - \iota \omega$. If you choose $\omega = \lambda / \iota$ then the map from the usual Clebsch coordinates to DESC coordinates is trivial: $\theta(\alpha, \zeta) = \alpha + \iota \zeta - \delta(\alpha, \zeta)$. It would become simple to trace data along field lines without expensive off-grid interpolation bottlenecks.

Just enforcing this constraint (even just approximately) for the highest order toroidal modes of $\omega$ such that $\omega_{n, \text{high}} = \lambda_{n, \text{high}} / \iota$ will make such computations cheaper.

Right now in vanilla DESC, since $\omega = 0$, the frequency transform of $\delta$ is nontrivial. Below is the discrete Hermitian Fourier-Chebyshev transform computed on a grid of size $32 \times 55$ nodes with tolerance of $10^{-7}$ in the coordinate mapping. The width of the toroidal spectrum unfortunately increases with NFP.

Image

If I parameterize $\delta$ under $(\vartheta, \zeta)$ to ensure that it is $2\pi, 2\pi/\text{NFP}$ periodic, then I can condense the Fourier transform significantly. Below is the Hermitian Fourier transform computed on a grid of size $32 \times 31$ nodes. However, evaluating this map on points along the field line now requires computing the full 2D transform, whereas with $\delta(\alpha, \zeta)$ it required a 1D transform. Thus, there was nothing gained by the spectral condensation under $(\vartheta, \zeta)$.

Image

For problem sizes that are large enough, the penalty of the dimension increase with the NUFFTs introduced in #1834 doesn't bite as hard as it does with an MMT. However, a 2D NUFFT is still significantly more expensive than 1D.

You can see #1919 for more plots on different flux surfaces.

### LINKED ISSUE #1928: Increase `Y` parameter for Heliotron (NFP=19) in effective ripple tutorial
Need more resolution to capture locations of bounce points on high frequency fields of Heliotron (NFP=19) on the flux surfaces near the boundary.

Now that we have NUFFTs we can do this on the GitHub CI without going out of memory.

### LINKED ISSUE #1945: Use #1919 for effective ripple and Gamma_c etc.
Users should use #1919 for these quantities. Here is example [documentation](https://desc-docs--1919.org.readthedocs.build/en/1919/notebooks/tutorials/EffectiveRipple.html) and [installation](https://desc-docs--1919.org.readthedocs.build/en/1919/installation.html) directions.

This branch improves convergence significantly and fixes bugs.

unalmis:
PR Age

## Inverse stream maps
- [x] Significantly improves convergence of inverse stream maps. Resolves #1922 to my satisfaction.
- [x] Truncation rule to remove aliasing error at the shortest wavelengths where the signal to noise ratio is lowest.
- [x] Resolves #1928 and increases default objective resolution for `Y_B`.
- [x] Method to plot frequency spectrum of inverse stream map.

## Improvements
- [x] Check-pointing to increase speed and reduce memory consumption of reverse mode differentiation #1347.
- [x] Adds ``low_ram`` mode which is same speed and less memory for `objective.compute`, but slower for `objective.grad` since JAX is poor at iterative algorithms.
- [x] Fully resolves #1864 by avoiding materialization of a large tensor in memory. Previously, we had closed the issue by adding nuffts as a workaround. The improvement here actually solves the JAX regression.
- [x] Reuses some computations in identifying bounce points to make more efficient.
- [x] Increase cache hits, fusing, and reduce floating point error in computing bounce points (very important for accurate integrals).
- [x] Transforms an improper field line integral to one on a compact domain where the integrand is periodic to achieve faster convergence.
- [x] Improves performance complexity of `interp_to_argmin` for `Bounce2D` from fourth order to spectral as required for Alpert quadrature.
- [x] Resolves #1389.

## Usability
- [x] Updates plotting utilities to accept `kwargs` as was needed for my article.
- [x] Method to compute integrals over plasma volume and velocity space in batches is now added to the public API. People kept asking how to use this private function and also used it for their work, so better to make it public.
- [x] Method to make adding new bounce integral objectives simpler. Private for developers only.
- [x] Unifies default parameter setup to resolve #1898 for bounce integrals.
- [x] Resolves #1388.

## Bugs
- [x] Fixes bug in ``test_compute_everything``.
- [x] Fixes inverse stream map convergence.

## Benchmarks
Just go to #2026 and run `effective_ripple_profile.py`. You will see the large performance improvement from `master`. The CI benchmarks do not reveal this because those benchmarks are essentially just noise. Note that, using the same parameter inputs, the resolution of this branch is also higher than `master` due to the faster convergence.

- If you set `use_bounce1d=True` on that script, you will run out of memory as expected since it is an inferior approach (as expected, you get the OOM in the jacobian before you compute a single bounce integral).
- If you set ``nufft_eps=0``, you need 175 GB to run that script on `master` (you'll get an OOM and JAX will tell you it needs 175GB), but only 34 GB on this branch.
- Using nuffts, the script requires only 6.5 GB on this branch. 

# Examples

### `Master` branch
test_theta_chebyshev_HELIOTRON

### This branch
test_delta_fourier_chebyshev_HELIOTRON

### `Master` branch
test_theta_chebyshev_W7-X

### This branch
test_delta_fourier_chebyshev_W7-X

### `Master` branch
test_theta_chebyshev_NCSX

### This branch
test_delta_fourier_chebyshev_NCSX

## Removal of spectral aliasing

Figure_1

# Dynamic shapes

- [x] resolves #1303 .
- [x] Activates newton step to find bounce points.
  - It can be shown there exists O(sqrt(epsilon)) error for bounce integrals with 1/v_ll where epsilon is error of bounce point. For v_|| integrals error is conveniently O(epsilon times sqrt(epsilon)). Hence, the `spline` method would require thousands of knots per transit for just a couple digits of accuracy, and it would stop convergence at epsilon<=1e-5 (so sqrt(epsilon) <=3 digits) error due to condition number. Of course, the `spline=False` method has always computed the points with spectral accuracy and has very fast convergence after #1919 ; that method converges to epsilon = machine precision without the newton step. With the newton step, fast convergence is achieved with the spline method as well.
  - I suspect fast ion confinement optimization will be easier now.

## Benchmarks
Here is a timing benchmark on my CPU with `nufft_eps=1e-6`. Prior to this PR, every adjoint call to nufft1 took `>= 1 second` and the full computation was `34 seconds`. Now every adjoint call to nufft1 is `250 milliseconds`, and the full computation is `14 seconds`. These `improvements become larger` as the `sparsity` grows and error tolerance parameter for the nuffts epsilon tends to 0. Likewise, the _improvement_ grows linearly with the `problem size`. As this is called within an optimization loop where time and memory are tight, the improvement is significant.

### Before
Screenshot From 2026-03-29 15-22-56
Screenshot From 2026-03-29 15-22-36

### After
Screenshot From 2026-03-29 15-12-35
Screenshot From 2026-03-29 15-15-59


desc/compute/_fast_ion.py
```diff
@@ -168,7 +170,7 @@ def Gamma_c(data):
 
         def fun(pitch_inv):
             points = bounce.points(pitch_inv, num_well)
-            v_tau, drift1, drift2 = bounce.integrate(
+            vτ, drift1, drift2 = bounce.integrate(
```
f0uriest: when naming variables and function arguments please stick to symbols that can be written with a standard english keyboard.

f0uriest (dismissed):

desc/compute/_fast_ion.py
unalmis: None of these variables are callable by users or developers. So no one can ever type them anyway. In the bounce integral source code at least, using unicode has decreased my development time because I can confirm correctness much easier.

desc/batching.py
```diff
@@ -272,9 +272,6 @@ def vmap_chunked(
 
     """
     in_axes, argnums = _parse_in_axes(in_axes)
-    if isinstance(argnums, int):
```
unalmis, line 218: duplicate code

desc/compute/_equil.py
```diff
@@ -144,7 +144,7 @@ def _J_sup_theta_PEST_v_PEST(params, transforms, profiles, data, **kwargs):
 # TODO: Generalize for a general phi before #568
 @register_compute_fun(
     name="(J^theta_PEST_p)|PEST",
-    label="\\partial_{\\phi}|_{\\rho, \\vartheta} J^{\\theta_{PEST}}",
+    label="\\partial_{\\phi}|_{\\rho, \\vartheta} J^{\\vartheta}",
```
unalmis, line 147: the latex label used to have both vartheta and theta pest which is confusing in plots.

desc/compute/_equil.py
```diff
@@ -107,7 +107,7 @@ def _J_sup_theta_PEST(params, transforms, profiles, data, **kwargs):
 # TODO: Generalize for a general zeta before #568
 @register_compute_fun(
     name="(J^theta_PEST_v)|PEST",
-    label="\\partial_{\\vartheta} J^{\\theta_{PEST}}",
+    label="\\partial_{\\vartheta}|_{\\rho, \\phi} J^{\\vartheta}",
```
unalmis, line 110: see above

desc/equilibrium/coords.py
```diff
@@ -453,43 +451,57 @@ def _partial_sum(lmbda, L_lmn, omega, W_lmn, iota):
     return lmbda_minus_iota_omega, jnp.fft.rfftfreq(grid.num_theta, 1 / grid.num_theta)
 
 
-@partial(jit, static_argnames=["tol", "maxiter"])
-def _map_clebsch_coordinates(
+@partial(jit, static_argnames=["inbasis", "outbasis", "tol", "maxiter"])
+def _map_poloidal_coordinates(
```
unalmis, line 452: this is a private method so we are free to make changes as we please. slowly pushing it to be more like eq.map_coordinates but it's not quite the same because it takes advantage of the 1D mesh grid nature of problem

desc/compute/_fast_ion.py
f0uriest: my point is that sometime in the future, someone will have to edit/fix/modify/refactor this code. I don't know of any major python libaries that use unicode in their codebase outside of comments, and I think there's a good reason for that.

Maybe it's because I'm old but I find stuff like this much harder to read, because in my head I don't know if thats a `p` or `rho` or `v` or `nu` etc, not to mention that there are 10 different kinds of each letter in unicode which all look the same. The mental burden for reading this is much higher.

dpanici: I'm with Rory on this, no unicode variables

unalmis: thiis was removed weeks ago

unalmis: and the wistell code uses unicode fwi if interested

unalmis: https://gitlab.com/wistell/StellaratorOptimizationMetrics.jl/-/blob/master/src/GammaC/Functions.jl?ref_type=heads

desc/compute/_neoclassical.py
```diff
@@ -89,14 +88,15 @@
         Precomputed transform matrices "dct spline", "dct cfl", "dft cfl".
         This parameter is intended to be used by objectives only.
         """,
+    "theta": "",
```
YigitElma, line 91: This is deprecated, right?

desc/equilibrium/equilibrium.py
```diff
@@ -1352,23 +1352,30 @@ def _get_rtz_grid(
         )
 
     @staticmethod
-    def _map_clebsch_coordinates(
+    def _map_poloidal_coordinates(
```
YigitElma, line 1363: I agree with your previous comment that this is private function and we can change the name, but recently I have seen couple groups that make use of these functions to create meshes for gyrokinetic simulations using DESC. I would vote to mention this name change at least in the changelog.

desc/equilibrium/coords.py
```diff
@@ -453,43 +451,57 @@ def _partial_sum(lmbda, L_lmn, omega, W_lmn, iota):
...
+        Values for coordinate specified by ``inbasis[1]``.
+    zeta : jnp.ndarray
         Shape (num zeta, ).
-        DESC toroidal angle.
+        DESC toroidal angle ζ.
     L_lmn : jnp.ndarray
         Spectral coefficients for λ.
     lmbda : Transform
-        Transform for λ built on DESC coordinates [ρ, θ, ζ].
+        Transform for λ built on DESC coordinates (ρ, θ, ζ) uniformly spaced in θ.
```
YigitElma: Just to clarify, the coordinate system doesn't matter here, right? You create a transform object with some `Grid` with uniform spacing in the poloidal direction, depending on the inbasis outbasis values whatever poloidal angle this transform correcponds to will change. Just noting this since `built on DESC coordinates (ρ, θ, ζ) uniformly spaced in θ.` might create confusion (say if you have a different inbasis)

desc/integrals/_interp_utils.py
```diff
@@ -66,9 +66,9 @@ def cheb_pts(n, domain=(-1, 1), lobatto=False):
     return bijection_from_disc(y, domain[0], domain[-1])
 
 
-def fourier_pts(n):
+def fourier_pts(n, domain=(0, 2 * jnp.pi)):
     """Get ``n`` Fourier points in [0, 2π)."""
```
YigitElma:
```suggestion
    """Get ``n`` Fourier points in domain."""
```
?

desc/integrals/basis.py
```diff
@@ -256,14 +251,167 @@ def compute_cheb(self, x):
...
-        # Add axis to broadcast against Chebyshev coefficients.
+        # Add axis to broadcast against Chebyshev series in y.
         x = jnp.atleast_1d(x)[..., None]
         # Add axis to broadcast against multiple x values.
         cheb = cheb_from_dct(irfft_mmt(x, self._c[..., None, :, :], self.X, axis=-2))
         assert cheb.shape[-2:] == (x.shape[-2], self.Y)
         return cheb
 
 
+class ChebyshevSeries(IOAble):
```
YigitElma: `DoubleChebyshevSeries` might be more appropriate for this.

desc/integrals/basis.py
```diff
@@ -256,14 +251,167 @@ def compute_cheb(self, x):
...
+    f : jnp.ndarray
+        Shape (..., X, Y).
+        Samples of real function on the ``ChebyshevSeries.nodes`` grid.
+    domain_x : tuple[float]
+        Domain for x coordinates. Default is [-1, 1].
+    domain_y : tuple[float]
+        Domain for y coordinates. Default is [-1, 1].
+    lobatto : bool
+        Whether ``f`` was sampled on the Gauss-Lobatto (extrema-plus-endpoint)
+        instead of the interior roots grid for Chebyshev points.
```
YigitElma: In addition to the error below, the docstring can also mention that jax doesn't have dct with lobatto points.

desc/integrals/basis.py
```diff
@@ -256,14 +251,167 @@ def compute_cheb(self, x):
...
+        L : int or jnp.ndarray
+            Optional, resolution in radial direction of domain [0, 1].
+            May also be an array of coordinates values. If given, then the
+            returned ``coords`` is a 3D tensor-product with shape (L * X * Y, 3).
+        domain_x : tuple[float]
+            Domain for x coordinates. Default is [-1, 1].
+        domain_y : tuple[float]
+            Domain for y coordinates. Default is [-1, 1].
+        lobatto : bool
+            Whether to use the Gauss-Lobatto (Extrema-plus-Endpoint)
```
YigitElma: Same as above

desc/integrals/_bounce_utils.py
```diff
@@ -467,145 +463,170 @@ def argmin(z1, z2, f, ext, g_ext):
...
+        αᵢ - αᵢ₊₁ = -ι 2π in δ. This is consistent with our interpolation of δ on the
+        branch (α, ζ) ∈ [0, 2π)². It is necessary to consider branch cuts in α where
+        αᵢ₊₁ ∉ [0, 2π) ∋ αᵢ as well. Observe the solutions θ* modulo 2π in [1] are
+        invariant against rotations in α by integer multiples of 2π. The periodicity
+        of δ in α ensures the θ = α + δ we compute may be discontinuous at branch cuts
+        by integer multiplies of 2π only. We intend to evaluate maps which are 2π
+        periodic in θ. Hence, θ mod 2π is simple to recover from the interpolation of
+        δ on a single branch (α, ζ) ∈ [0, 2π)².
+
+    The field line label α changes discontinuously along a magnetic field line.
```
YigitElma: After one toroidal transit you mean?

YigitElma (dismissed):

desc/compute/_neoclassical.py
unalmis: yes

desc/equilibrium/coords.py
unalmis: The docstring is correct. You need rho, theta, zeta regardelss of inbasis.

desc/integrals/_bounce_utils.py
unalmis: it is discontinuous across any branch cut. Toroidal or poloidal

desc/equilibrium/equilibrium.py
unalmis: If people want functionality to use one of our hidden private methods, they should make a feature request issue. These methods were made private for the sole purpose of preventing people from using them.

desc/equilibrium/coords.py
```diff
@@ -500,40 +512,70 @@ def _map_clebsch_coordinates(
...
+        f"outbasis={outbasis} is not implemented.",
+    )
 
-    def rootfun(theta, target, c_m):
-        c = (jnp.exp(1j * modes * theta) * c_m).real.sum()
-        target_k = theta + c
-        return target_k - target
+    # Root finding for θₖ such that θₖ + (λ−ιω)(ρ,θₖ,ζ) - ε = 0.
+    def rootfun(θ, varepsilon, q_m):
+        q = (jnp.exp(1j * modes * θ) * q_m).real.sum()
```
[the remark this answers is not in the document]
unalmis: i can change this variable here, but I am not changing the bounce integral code

desc/integrals/basis.py
unalmis: I try to follow the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) code standard to decrease maintaince burden. If I add it to the docstring, that's just more places that need to be edited in the future.

desc/equilibrium/equilibrium.py
unalmis: can make an issue or publish the notification in the pip release notes if you want.

desc/integrals/basis.py
unalmis: it's not public and all this interpolation stuff will be moved to a new library in the future #1388

desc/equilibrium/coords.py
f0uriest: and does the rho of this transform need to be the same rho as `iota(rho)`?

desc/equilibrium/coords.py
```diff
@@ -453,43 +448,57 @@ def _partial_sum(lmbda, L_lmn, omega, W_lmn, iota):
...
+        DESC toroidal angle ζ.
     L_lmn : jnp.ndarray
         Spectral coefficients for λ.
     lmbda : Transform
-        Transform for λ built on DESC coordinates [ρ, θ, ζ].
+        Transform for λ built on DESC coordinates (ρ, θ, ζ) uniformly spaced in θ.
+    varepsilon : jnp.ndarray
+        Shape should broadcast with shape (num ρ, num poloidal, num ζ).
+        Optional meshgrid of values ε = α + ι ζ.
+        This meshgrid need not be a tensor-product grid.
```
f0uriest, line 494: what exactly does this mean? in most places we've been using meshgrid to mean a tensor product grid

desc/equilibrium/coords.py
```diff
@@ -500,40 +509,70 @@ def _map_clebsch_coordinates(
...
+    def rootfun(t, varepsilon, q_m):
+        q = (jnp.exp(1j * modes * t) * q_m).real.sum()
+        return t + q - varepsilon
 
-    def jacfun(theta, target, c_m):
-        dc_dt = ((1j * jnp.exp(1j * modes * theta) * c_m).real * modes).sum()
-        return 1 + dc_dt
+    def jacfun(t, varepsilon, q_m):
+        dq_dt = ((1j * jnp.exp(1j * modes * t) * q_m).real * modes).sum()
+        return 1 + dq_dt
```
f0uriest: is this missing a `d_varepsilon/dt` term?

desc/equilibrium/equilibrium.py
```diff
@@ -1134,7 +1134,13 @@ def need_src(name):
                 ),
             )
             data1dr_seed = {
-                key: grid1dr.copy_data_from_other(data[key], grid, surface_label="rho")
+                key: (
+                    data[key]
+                    if (key == "V_r/psi_r")
```
f0uriest: why can't this be copied like the others?

desc/external/neo.py
f0uriest, line 1: should add this file to the coverage exclude list in `setup.cfg`

desc/integrals/quad_utils.py
```diff
@@ -1,17 +1,4 @@
-"""Utilities for quadratures.
-
-Notes
```
f0uriest, line 3: i think this note is useful to keep if its still accurate

desc/objectives/_fast_ion.py
```diff
@@ -91,22 +95,22 @@ class GammaC(_Objective):
         irrational magnetic surface better, with diminishing returns.
     num_well : int
         Maximum number of wells to detect for each pitch and field line.
-        Giving ``None`` will detect all wells but due to current limitations in
+        Giving ``-1`` will detect all wells but due to current limitations in
```
f0uriest, line 98: is the default of `None` the same as `-1` or does it use the `A+iota*C` heuristic?

desc/objectives/_neoclassical.py
```diff
@@ -67,17 +67,18 @@ class EffectiveRipple(_Objective):
         Determines the flux surfaces to compute on and resolution of FFTs.
         Default grid samples the boundary surface at ρ=1.
     X : int
-        Poloidal Fourier grid resolution to interpolate the poloidal coordinate.
+        Poloidal Fourier grid resolution to interpolate the angle.
```
f0uriest, line 69: "angle" here means delta = theta - alpha?

desc/equilibrium/coords.py
```diff
@@ -453,43 +448,57 @@ def _partial_sum(lmbda, L_lmn, omega, W_lmn, iota):
...
+def _map_poloidal_coordinates(
     iota,
-    alpha,
+    poloidal,
     zeta,
     L_lmn,
     lmbda,
+    varepsilon=None,
+    inbasis=("rho", "alpha", "zeta"),
+    outbasis=("rho", "theta", "zeta"),
```
f0uriest: im a bit confused by inbasis and outbasis here. isn't the input radial coordinate always iota and input toroidal coordinate always zeta? and the output is always only the poloidal angle? I think it would be clearer if inbasis and outbasis were just single coordinates, not triplets (basically to label what `poloidal` is and what the output should be)

tests/baseline/test_effective_ripple_1D.png
f0uriest, line 1: any idea why this changed so much?

tests/test_interp_utils.py
```diff
@@ -396,6 +421,99 @@ def test_fourier_chebyshev(self, func, m, n):
...
+        Y = TestStreams.Y
+        angle = Bounce2D.angle(eq, X, Y, TestStreams.rho, tol=TestStreams.tol)
+        return Bounce2D.plot_angle_spectrum(angle, 0, norm=TestStreams.norm)
+
+    @pytest.mark.unit
+    @pytest.mark.parametrize("name", ["W7-X", "NCSX", "HELIOTRON"])
+    @pytest.mark.mpl_image_compare(remove_text=True, tolerance=tol_2d)
+    @staticmethod
+    def test_lambda_fourier_vartheta_zeta(name):
+        """Plot Fourier spectrum of Λ(ϑ, ζ)."""
```
f0uriest, line 213: is this lowercase lambda? (ie our usual poloidal stream function?)

desc/compute/_core.py
f0uriest, line 1: kind of a nit, but I'd vote for coming up with more descriptive names for these qtys, or adding a modifier like `delta_bounce` or something. These don't already have a standard label in the literature like `alpha`, and greek letters are a finite resource in physics.

desc/integrals/_bounce_utils.py
```diff
@@ -340,15 +347,19 @@ def plot_ppoly(
...
+        Whether to plot the legend. Default is true.
 
     Returns
     -------
     fig, ax
         Matplotlib (fig, ax) tuple.
 
     """
-    fig, ax = plt.subplots()
+    if "figsize" in kwargs:
```
f0uriest: I don't think you need the if/else here, you can always use `figsize=kwargs.pop("figsize", None)`

desc/integrals/_bounce_utils.py
```diff
@@ -467,148 +481,160 @@ def argmin(z1, z2, f, ext, g_ext):
...
+    [1] θ* - (δ−ιζ)(θ*, ζ) = α + ιζ.
+
+    This feature avoids expensive off-grid re-interpolation in optimization.
+
+    [2] δ is (2π, ∞) periodic in (α, ζ).
+        A Fourier series efficiently interpolates δ(α, ζ) at fixed ζ.
+        Likewise, a Chebyshev series interpolates δ(α, ζ) at fixed α.
+
+    [3] The angle θ must be continuous at the branch cuts.
+        θ = αᵢ + δ(α=αᵢ, ζ=k 2π/NFP) = αᵢ₊₁ + δ(α=αᵢ₊₁, ζ=0) for k ∈ ℤ.
```
f0uriest: `for k ∈ ℤ` -> `for i ∈ ℤ`?

desc/integrals/_bounce_utils.py
```diff
@@ -467,148 +481,160 @@ def argmin(z1, z2, f, ext, g_ext):
...
+    domain = (0, 2 * jnp.pi / NFP)
+    Y = truncate_rule(angle.shape[-1])
+    delta = (
+        FourierChebyshevSeries(angle, domain, truncate=Y)
+        .compute_cheb(alpha)
+        .swapaxes(0, -3)
     )
-    return T
+    alpha = alpha.swapaxes(0, -2)
+    delta = delta.at[..., 0].add(alpha)
```
f0uriest, line 773: the delta on the lhs of this expression is really theta right, but just doing the modification in place?

desc/integrals/_bounce_utils.py
```diff
@@ -617,84 +638,84 @@ def fast_chebyshev(T, f, Y, num_theta, m_modes, n_modes, NFP=1, *, vander=None):
...
 
 def fast_cubic_spline(
-    T,
+    theta,
     f,
     Y,
-    num_theta,
-    m_modes,
-    n_modes,
+    num_θ,
```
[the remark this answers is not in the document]
f0uriest: still not resolved

desc/integrals/_bounce_utils.py
```diff
@@ -705,62 +731,72 @@ def fast_cubic_spline(
...
         else:
-            theta = theta.transpose(3, 0, 1, 2).reshape(num_zeta, -1)
-        f = nufft1d2r(theta, f, eps=nufft_eps).mT
+            θ = θ.transpose(3, 0, 1, 2).reshape(num_ζ, -1)
+        f = nufft1d2r(θ, f, eps=nufft_eps).mT
     f = f.reshape(*lines, -1)
 
-    zeta = jnp.ravel(zeta + (T.domain[1] - T.domain[0]) * jnp.arange(T.X)[:, None])
-    f = CubicSpline(x=zeta, y=f, axis=-1, check=check).c
+    ζ = jnp.ravel(
```
f0uriest: unicode

desc/integrals/_bounce_utils.py
```diff
@@ -769,46 +805,123 @@ def _move(f, out=True):
...
+    Y : int
+        Resolution for Gauss-Legendre quadrature over one field period.
+
+    """
+    return max(Y, 8)
+
+
+def Y_B_rule(Y, NFP, spline=True):
+    """Guess Y_B from resolution of Chebyshev spectrum of angle."""
+    return (2 * Y * int(np.sqrt(NFP))) if spline else Y
```
f0uriest: where does the `int(sqrt(NFP))` come from?

desc/integrals/_bounce_utils.py
```diff
@@ -769,46 +805,123 @@ def _move(f, out=True):
...
+
+
+def num_well_rule(num_transit, NFP, Y_B=None):
```
f0uriest, line 1069: shouldn't this depend on the M,N resolution of the equilibrium in some way?

desc/integrals/_bounce_utils.py
```diff
@@ -769,46 +805,123 @@ def _move(f, out=True):
...
 
     """
     if jnp.ndim(pitch_inv) == 2:
         pitch_inv = pitch_inv[:, None]
     return pitch_inv
+
+
+def truncate_rule(Y):
+    """Truncation of Chebyshev series to reduce spectral aliasing."""
+    return max(1, 7 * Y // 8)
```
f0uriest, line 1023: is there any particular reason for 7/8? shouldn't this depend in some way on the resolution/bandwidth of the underlying function being approximated relative to the resolution used for the approximation?

desc/integrals/basis.py
```diff
@@ -256,14 +257,187 @@ def compute_cheb(self, x):
...
+        )
+        warnif(
+            Y < self.Y,
+            msg="Frequency spectrum of DCT interpolation will be truncated because "
+            "the grid resolution is less than the Chebyshev resolution.\n"
+            f"Got Y = {Y} < {self.Y} = self.Y.",
+        )
+        BUG = jax.__version_info__ <= (0, 7, 2)
+        errorif(
+            self._c.ndim > 2 and BUG,
```
f0uriest: this would only be an issue if we're approximating a vector valued qty correct? ie `self._c.shape == (..., X, Y)`?

f0uriest: mainly trying to see if we expect this to occur in practice, in which case we should either bump the jax version or figure out a workaround

desc/integrals/bounce_integral.py
```diff
@@ -240,60 +252,19 @@ class Bounce2D(Bounce):
...
         Whether to use cubic splines to compute bounce points instead of
-        Chebyshev series. Note the algorithm for efficient root-finding on
-        Chebyshev series has not been implemented.
+        Chebyshev series. Default is true.
     check : bool
         Flag for debugging. Must be false for JAX transformations.
 
     """
 
-    # For applications which reduce to computing a nonlinear function of distance
```
f0uriest, line 250: I'd be fine keeping this note for future developers/students

desc/integrals/bounce_integral.py
```diff
@@ -308,62 +279,291 @@ def __init__(
...
+        --------
+          * ``desc/objectives/_fast_ion.py::GammaC``
+          * ``desc/objectives/_neoclassical.py::EffectiveRipple``
+
+        Parameters
+        ----------
+        obj : _Objective
+            The objective instance.
+        get_profiles, get_transforms, parse_callable_target_bounds
+            Functions that must be passed in to avoid circular import errors.
```
f0uriest: could also use a local import?

desc/integrals/bounce_integral.py
```diff
@@ -308,62 +279,291 @@ def __init__(
...
+        if obj._grid is None:
+            obj._grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False)
+        assert obj._grid.can_fft2
+
+        X = obj._constants["X"]
+        Y = obj._constants["Y"]
+        x_pts = fourier_pts(X)
+        y_pts = cheb_pts(Y, (0, 2 * jnp.pi / eq.NFP))[::-1]
+        obj._constants["X"] = x_pts
+        obj._constants["Y"] = y_pts
```
f0uriest: i'd prefer to use lowercase x,y for these since everywhere else capital X,Y are integers denoting the resolution not arrays of coordinates

desc/integrals/bounce_integral.py
```diff
@@ -308,62 +279,291 @@ def __init__(
...
+          * ``desc/compute/_neoclassical.py::_epsilon_32``
+
+        Parameters
+        ----------
+        fun : callable
+            A function  which takes a single argument ``fun_data`` and computes
+            bounce integrals assuming ``fun_data`` holds all required quantities
+            to construct a ``Bounce2D`` operator as well as call its methods with
+            the flag ``is_fourier=True``.
+        fun_data : dict[str, jnp.ndarray]
```
f0uriest, line 533: why do we need `fun_data` and `desc_data`? it looks like we just populate `fun_data` from the given `desc_data`?

desc/integrals/bounce_integral.py
```diff
@@ -422,121 +622,154 @@ def compute_theta(
...
+        rho=jnp.array([1.0]),
+        iota=None,
+        params=None,
+        profiles=None,
+        tol=1e-7,
+        maxiter=30,
         **kwargs,
     ):
-        """Return DESC coordinates θ of (α,ζ) Fourier Chebyshev basis nodes.
+        """Return the angle for mapping boundary coordinates to field line coordinates.
```
f0uriest, line 661: what does "boundary coordinates" refer to here?

desc/integrals/bounce_integral.py
```diff
@@ -760,11 +996,12 @@ def _nufft(self, zeta, theta, data, eps):
         c = c.swapaxes(0, -2).reshape(len(data) + 2, *shape)
         return dict(zip([*data.keys(), "B^zeta", "|B|"], c))
 
-    def _nummt(self, zeta, theta, data):
-        v = rfft2_vander(zeta, theta, self._n_modes, self._m_modes)
-        data = {name: _mmt_for_bounce(v, c) for name, c in data.items()}
-        data["B^zeta"] = _mmt_for_bounce(v, self._c["B^zeta"])
-        data["|B|"] = _mmt_for_bounce(v, self._c["|B|"])
+    def _nummt(self, ζ, data):
```
f0uriest: unicode

desc/integrals/bounce_integral.py
```diff
@@ -812,99 +1049,98 @@ def interp_to_argmin(self, f, points, *, nufft_eps=1e-6, is_fourier=False):
             polyder_vec(self._c["B(z)"]),
             sentinel=0.0,
         )
-        theta = self._c["T(z)"].eval1d(ext)
+        θ = self._theta.eval1d(ext)
```
f0uriest: unicode

desc/integrals/bounce_integral.py
```diff
@@ -812,99 +1049,98 @@ def interp_to_argmin(self, f, points, *, nufft_eps=1e-6, is_fourier=False):
...
+    def compute_fieldline_length(self, quad=None):
         """Compute the (mean) proper length of the field line ∫ dℓ / B.
 
-        Computes mean_A ∫ dℓ / B where A is the set of field line labels
-        given when making this object.
+        Notes
+        -----
+        This result will converge to
+        (num transit / 2π) * ∬_Ω abs(𝐁⋅∇ζ)⁻¹ dα dζ where (α,ζ) ∈ Ω = [0, 2π)².
+        In new versions of DESC, this can be computed more efficiently as
```
f0uriest: should we deprecate this function then?

f0uriest: it looks like its not currently tested?

desc/integrals/bounce_integral.py
```diff
@@ -984,36 +1220,119 @@ def plot_theta(self, l, m, **kwargs):
...
+            See https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html.
+            Default is logarithmic scale with cutoff at ``1e-7``.
+        h_ax_numticks : int
+            If given, labels at most ``h_ax_numticks`` marks on the horizontal axis.
+        v_ax_numticks : int
+            If given, labels at most ``v_ax_numticks`` marks on the vertical axis.
+        kwargs
+            Keyword arguments to pass to ``matplotlib``.
 
-    The bounce integral is defined as ∫ f(ρ,α,λ,ℓ) dℓ where
```
f0uriest: see note above, we may want to keep this for RTD

desc/integrals/bounce_integral.py
```diff
@@ -1332,24 +1723,22 @@ def _integrate(
     ):
         shape = (*z1.shape, x.size)  # (..., num pitch, num well, num quad)
 
-        zeta = flatten_mat(
-            bijection_from_disc(x, z1[..., None], z2[..., None]), 2 + batch
-        )
+        ζ = flatten_mat(bijection_from_disc(x, z1[..., None], z2[..., None]), 2 + batch)
```
f0uriest: unicode

desc/integrals/bounce_integral.py
```diff
@@ -308,62 +279,291 @@ def __init__(
...
+                )
+            )
+            num_pitch = kwargs.get("num_pitch", 65)
+            nufft_eps = kwargs.get("nufft_eps", 1e-7)
+
+        pitch_batch_size = kwargs.get("pitch_batch_size", None)
+        surf_batch_size = kwargs.get("surf_batch_size", 1)
+        assert (
+            surf_batch_size == 1 or pitch_batch_size is None
+        ), f"Expected pitch_batch_size to be None, got {pitch_batch_size}."
```
f0uriest, line 474: remind me why we can't batch both surfaces and pitch?

desc/compute/_neoclassical.py
```diff
@@ -86,70 +76,41 @@
...
+    transforms={"grid": []},
+    profiles=[],
+    coordinates="r",
+    resolution_requirement="tz",
+    data=["psi_r/sqrt(g)"],
+)
+def _field_line_weight(params, transforms, profiles, data, **kwargs):
+    """∬_Ω abs(𝐁⋅∇ζ)⁻¹ dα dζ where (α,ζ) ∈ Ω = [0, 2π)².
 
+    The returned quantity has shape (num rho, ).
```
f0uriest: I'd prefer if this was expanded out to match with the other qtys, I don't see any strong reason to make an exception just for this.

desc/compute/_neoclassical.py
```diff
@@ -86,70 +76,41 @@
...
-    for name in fun_data:
-        fun_data[name] = Bounce2D.fourier(Bounce2D.reshape(grid, fun_data[name]))
-    fun_data["iota"] = grid.compress(data["iota"])
-    fun_data["theta"] = theta
-    fun_data["pitch_inv"], fun_data["pitch_inv weight"] = Bounce2D.get_pitch_inv_quad(
-        grid.compress(data["min_tz |B|"]),
-        grid.compress(data["max_tz |B|"]),
-        num_pitch,
-        simp=simp,
+    data["V_r/psi_r"] = surface_integrals(
```
f0uriest: We already have compute qtys for `V_r(r)` and `psi_r`, isn't this just the same as `data['V_r(r)'] / data['psi_r'])`

tests/test_axis_limits.py
```diff
@@ -252,7 +252,9 @@ def assert_is_continuous(
             # can't check radial continuity of scalar or function of toroidal angle
             continue
         # make single variable function of rho
-        if data_index[p][name]["coordinates"] == "r":
+        if name == "V_r/psi_r":
```
f0uriest: why the special case here? can't we just add to not implemented limits?

desc/equilibrium/coords.py
unalmis: Meshgrids are more general than tensor-product grids. For example, if you iterate along the the second axis of the quantity ``varepsilon.reshape(num rho, num poloidal, num zeta)``, then the values of ``varepsilon[:, i, :]`` can be different for each ``i``. In contrast, on a tensor-product grid, varepsilon has to be the same for each `i`.

The purpose of the comment is basically to inform the user that we support tensor product grids in (rho, alpha, iota zeta) in addition to tensor product grids in (rho, alpha, zeta). 

If you want to perform a coordinate mapping for a single poloidal transit on every flux surface, then giving ``varepsilon`` as a meshgrid in (rho, alpha, zeta) works but it can't be a tensor-product grid in (rho, alpha, zeta) unless iota is constant wrt to rho.

desc/equilibrium/coords.py
unalmis: The left hand side in this image is a function of theta. The right hand side is varepsilon, which is always a constant. We solve f(theta) = constant.

image

varepsilon is also always continuous along field lines, regardless of where you define the start and end of the field line, so that's why i chose it as the constant rather than solving something else like: f(theta) - iota zeta = varepsilon - iota zeta.

desc/objectives/_fast_ion.py
unalmis: In this PR, I have changed the default `None` to be very safe heuristic that we set. If the user has unlimited compute, and does not trust our heurstic, they can set it to `-1` to ensure all wells are found.

desc/objectives/_neoclassical.py
unalmis: Yes

desc/equilibrium/coords.py
unalmis: I was trying to match the eq.map_coordinates api. i can change as you have requested to just one variable.

tests/baseline/test_effective_ripple_1D.png
unalmis: None of our computations have any visible changes.[1]

It is the NEO output that has changed: Master has NEO ran with 3x booozer resolution. This branch has a NEO file ran with 5x boozer resolution which converged to be closer to ours:
image



[1]: Besides the heliotron example in the linked github issue which was not using sufficient resolution.

tests/test_interp_utils.py
unalmis: yes Λ is DESC stream. little lambda is pitch angle

desc/integrals/_bounce_utils.py
unalmis: I believe k is right. I wrote a simpler explanation in the paper, so I intend to removed this comment in the code.

image
Define the left hand side as f. We are solving f(theta) = constant.

where `constant` = alpha + iota zeta.

Now I solve this equation on a single field period, and want to know what the solution to the equation is at some zeta outside this field period. If I mod zeta by 2pi/NFP, then f(theta, zeta) doesn't change since it's NFP periodic. So you just have to shift alpha by some amount to ensure that `constant` stays the same.

image

desc/integrals/_bounce_utils.py
unalmis: correct, once i add alpha it becomes theta.

desc/integrals/_bounce_utils.py
unalmis: `Y_B` is a resolution over a toroidal transit, whereas `Y` is resolution over field period. Ideally, they'd both be over a field period, but I don't want to break user code. `2 * sqrt NFP` seemed to work better than `NFP`.

desc/integrals/_bounce_utils.py
unalmis: It depends on the dominant frequency along field lines. There's a strong dependence on NFP obviously, but increasing M, N on an equilibrium doesn't really change the dominant frequency since the spectral coefficients decay rapidly.

desc/integrals/_bounce_utils.py
unalmis: Just to ensure the readers are on the same page, note that sometimes one should consider an interpolating approximator as closer to the truth than a projection approximator. That is not the case here; here the truth is better modeled by the projection.

1. I did a parameter scan across various resolutions. For the range of resolutions that make any sense to use, e.g. from Y = 8 to 64, the index of the mode at which the error of the interpolant (green) approached the magnitude of its spectral coefficients (orange) was typically around `7/8 Y`. At such a point, those spectral coefficients are junk.


image

This plot was without the NFP compression, but the same philosophy applies.

2. In this application we greatly prefer the galerkin approximator to the pseudospectral one because we are evaluating the function at so many points. Since it's really expensive to keep more modes, we prefer if all the error is due to truncation (galerkin) instead of both error from interpolation + truncation. It would actually be better to have a more aggressive rule since the evaluation grid is much denser, but the NFP compression made the convergence fast enough that I think it's fine.

3. i'd argue you always need some form of truncation with the chebyshev interpolation because the points cluster, and so the higher order modes you have, the more likely they are sensitive to noise from errors in root finding. removing some of the highest order ones guards against this.

unalmis: > shouldn't this depend in some way on the resolution/bandwidth of the underlying function being approximated relative to the resolution used for the approximation?

You choose that dependence when you select `Y`. The assumption is you made a good choice for `Y`. Given that good choice, a truncation is preferred for reasons discussed above.

desc/integrals/basis.py
unalmis: yes. i was going to move this class to external package since we don't use it in code anyway. It was to mainly to see what `master` should have been, had i chosen the correct basis for theta.

desc/integrals/bounce_integral.py
unalmis: We interpolate a Fourier series to everything in fun_data. That's typically about 3 smooth quantities that we use to compose a singular integrand with primitive functions.  In contrast, DESC data has a thousand dependencies stored in it.

desc/integrals/bounce_integral.py
unalmis: Boundary coordinates are DESC theta and zeta always. I figured boundary is easier to swallow than understanding our symbols.

desc/integrals/bounce_integral.py
unalmis: It's tested in test_bounce2d_checks. I am fine with deprecating. Note that, the new method is only more efficient if you have already computed all the dependencies for V_psi. compute_fieldline_length is still faster than tranform.build().

desc/integrals/bounce_integral.py
unalmis: We compute integrals between each bounce point here.
image

 Therefore, there are two costs.
1. All the quadratures. Since fast multipole methods do not exist for these integrals, it is important to choose a quadrature that converges as fast as possible to reduce this cost.
2. All the interpolation to get data at the quadrature points. If we make the best choice for the first point, then the quadrature points will be different for each integral and hence require different interpolation data. Nuffts solve this exact problem. With NUFFTs, the more interpolation data you have computed, the easier it becomes to compute new interpolation data. That's why we want to compute all the interpolation data for every pitch angle in one shot. Then the cost of evaluating a fourier series with resolution M N at P points is linear in P. $O(M N log (MN) + P)$.

3. If you do half the pitch angles and multiple surfaces, you don't take advantage of 2, and you have to recompute all the dependencies (bounce points, stream map, splines... etc.).

unalmis: You can batch both. Just don't bother batching surfaces if you're not doing all pitches in one shot.

desc/compute/_neoclassical.py
unalmis: anything that differs from the implementation i have written makes it a total pain to do the axis limit

desc/integrals/bounce_integral.py
unalmis: I removed because some parts were wrong.

> Function approximation in (α, ζ) coordinates demands particular interpolation
> points in that coordinate system...

Not really, could compute the projection via integrals in your preferred coordinate system. Though you can't use FFTs, and it converges slower since it introduces another Jacobian factor to the integrand among other things.

> For applications which reduce to computing a nonlinear function of distance
> along field lines between bounce points, it is required to identify these
> points with field-line-following coordinates. (In the special case of a linear
> function summing integrals between bounce points over a flux surface, arbitrary
> coordinate systems may be used as that task reduces to a surface integral,
> which is invariant to the order of summation).

This is also false. After all the physics is done, I can write any surface integral of some nonlinear map of bounce integrals completely independent of field lines. For example, I can compute 3.24
image
purely in DESC coordinates. The reason I don't is because finding the boundaries of the relevant integrals becomes painful. The fact that field lines ergodically cover the surface means that  field line coordinates work as a preconditioner for root finding (of bounce points). You are not guaranteed to find roots of |B| - 1/lambda  moving along zeta at constant theta for example (just consider the axisymmetric case).

The other statements are correct.

desc/integrals/quad_utils.py
unalmis: i'd prefer they just read my paper

desc/compute/_core.py
unalmis: i set public to false to resolve.

desc/integrals/bounce_integral.py
unalmis: yes

desc/integrals/bounce_integral.py
unalmis: see effective ripple 3/2 compute function for example

desc/integrals/bounce_integral.py
unalmis: ok changed to lowercase in next commit

desc/equilibrium/coords.py
unalmis: yes

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
...
+v0.16.1
+-------
+
 New Features
 
+- [Significantly improves convergence of inverse stream maps](https://github.com/PlasmaControl/DESC/pull/1919).
+    - Truncation rule to remove aliasing error at the shortest wavelengths where the signal to noise ratio is lowest.
+    - Method to plot frequency spectrum of inverse stream map in field line coordinates ``Bounce2D.plot_angle_spectrum``.
+- Method to compute integrals over plasma volume and velocity space in batches is now added to the public API ``Bounce2D.batch``.
+- Method for developers to make adding new bounce integral objectives simpler ``Bounce2D._objective_build``.
```
YigitElma: Private method doesn't need to be here

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
 Changelog
 =========
 
+v0.16.1
```
YigitElma: Next release will probably be a big one, we might change the minor version

desc/external/neo.py
```diff
@@ -0,0 +1,117 @@
...
+    def write(self, **kwargs):
+        """Write neo input file."""
+        print(f"Writing VMEC wout to {self.vmec_file}")
+        VMECIO.save(self.eq, self.vmec_file, surfs=self.ns, verbose=0)
+        self._write_booz()
+        self._write_neo(**kwargs)
+
+    def _write_booz(self):
+        print(f"Writing boozer output file to {self.booz_file}")
+        import booz_xform as bx
```
YigitElma, line 51: another try except could be nice

desc/external/paraview.py
```diff
@@ -7,11 +7,13 @@
 try:
     import pyvista as pv
 except ImportError:
-    warnings.warn(
-        "DESC objects are exported to Paraview using `pyvista`"
-        "package which is an optional dependency. Please pip install "
-        "`pyvista` to your environment."
-    )
+    with warnings.catch_warnings():
```
YigitElma, line 10: if we are going to use any of the `external` code in tests, maybe we shouldn't import all functions in `__init__.py`? What do others think?

desc/external/README.rst
```diff
@@ -9,8 +9,7 @@ The external codes are not included with DESC, and you may need to obtain access
 to use these objectives/utilities. Those codes may not be publicly available and may require a
 license to install them.
 
-`_terpsichore.py` was last tested on September 8, 2025 with DESC v0.15.0 and a version
-of TERPSICHORE compiled on Perlmutter in January of 2025.
-
-`paraview.py` was last tested on November 24, 2025 with DESC v0.16.0, Paraview v5.13.3
+* ``_terpsichore.py`` was last tested on September 8, 2025 with DESC v0.15.0 and a version of TERPSICHORE compiled on Perlmutter in January of 2025.
+* ``_neo.py`` was last tested on October 23, 2025 with the version of NEO on Stellar.
```
YigitElma: Stellar doesn't mean much to anyone except Princeton people, and it will be decommissioned in 8 months. What is the official version number?

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
...
+-------
+
 New Features
 
+- [Significantly improves convergence of inverse stream maps](https://github.com/PlasmaControl/DESC/pull/1919).
+    - Truncation rule to remove aliasing error at the shortest wavelengths where the signal to noise ratio is lowest.
+    - Method to plot frequency spectrum of inverse stream map in field line coordinates ``Bounce2D.plot_angle_spectrum``.
+- Method to compute integrals over plasma volume and velocity space in batches is now added to the public API ``Bounce2D.batch``.
+- Method for developers to make adding new bounce integral objectives simpler ``Bounce2D._objective_build``.
+- The resolution parameter for the field line quadrature now corresponds to the resolution for integration over a single field period.
```
YigitElma: Add where this `field line quadrature` change happened (like EffectiveRipple, Bounce etc)

desc/integrals/bounce_integral.py
YigitElma: Just to clarify, theta is not just defined by boundary, I mean it is not the same on the straight line from axis to some point on the boundary. See https://github.com/PlasmaControl/DESC/issues/1973#issuecomment-3408917056. So, this sounds like an even more confusing explanation of what this function is doing to me

YigitElma (commented):
I assume there is no math error, I couldn't check it very well because it is really hard to follow the code with all the different `angle` calls and some of function names are ambigious.

Are you planning to maintain the `interpax_fft` package?

desc/integrals/bounce_integral.py
unalmis: There are many choices one can make for theta. The defining characteristic of theta is that it matches the angle which the user chose to parameterize the boundary. This function returns some angle (that is, $\theta - \alpha$, but that's an implementation detail) that assists in mapping $\theta$ to a field line coordinate system.

desc/external/neo.py
unalmis: what would do you intend for the except block to be? we would just raise the import error that would have been raised anyway

desc/external/README.rst
unalmis: neo output logs don't have that info

desc/external/paraview.py
unalmis: all catch warnings does is prevent the CI from failing. it still prints the warning in exactly same way as before. nothing changes from user perspective

YigitElma: I am not criticizing your change. Currently, if you import `NeoIO` you get the warning for `pyvista`, this is weird for a user. If we don't have anything in init, and only allow imports like `from desc.external.neo import NeoIO` or `from desc.external.paraview import export_surface_to_paraview` etc, we won't have this problem. In future, we can have multiple optional dependencies just for external codes to work, and all of them will throw a warning

CHANGELOG.md
```diff
@@ -26,10 +32,14 @@ Bug Fixes
 
 - No longer uses the full Hessian to compute the scale when ``x_scale="auto"`` and using a scipy optimizer that approximates the hessian (e.g. if using ``"scipy-bfgs"``, no longer attempts the Hessian computation to get the x_scale).
 - ``SplineMagneticField.from_field()`` correctly uses the ``NFP`` input when given. Also adds this as a similar input option to ``MagneticField.save_mgrid()``.
+- Significantly improves convergence of inverse stream maps in bounce integrals.
```
YigitElma, line 47: duplicate

unalmis: I put it in bug fix too so that if a user notices their computation has changed they will know it is due to this

desc/external/paraview.py
unalmis: made an issue as it is outside the scope of this pr. master would get the same pyvista warning when terpsichore stuff is imported

YigitElma:
> * To now state that you are blocking the work because of that dependency is unreasonable

I am not blocking it. If others approve it, you can merge it; that is fine. I don't have to approve all PRs. I've reviewed it multiple times and spent a couple of days on it.

Tal-Shpigel (approved):

desc/integrals/_interp_utils.py
```diff
@@ -819,20 +355,47 @@ def polyroot_vec(
...
+    dc_dr = poly_val(x=r, c=c[..., None, :], der=True)
+    dc_dr = jnp.where(
+        jnp.abs(dc_dr) > eps,
+        dc_dr,
+        dc_dr + jnp.copysign(eps, dc_dr.real),
+    )
+    dr = jnp.where(
+        r == sentinel,
+        0.0,
+        (jnp.expand_dims(dk, -1) - poly_val(x=r, c=dc[..., None, :])) / dc_dr,
```
unalmis, line 381: Same as https://github.com/PlasmaControl/DESC/pull/2145#discussion_r3018605864.

dpanici (approved):

PR #1920, opened 2025-09-18
add a to I normalization scale

dpanici:
I scale is not correct units, to me makes more sense to have it be "B" * a * 2pi / mu0 since I can be computed as line integral of B_theta divided by mu_0 around a toroidal XS. 

This seems to give better scaling for a tokamak case I had where the scale was an order of magnitude higher than the current. Does not differ much from master for most of our finite beta eqs.

this wont affect any tests or notebooks as none use ToroidalCurrent which is the only objective using this particular scale

```python
####################
DSHAPE
####################
net current: 2.2516e+05
master d[I]: 1.3534e+06
d[B]*d[a]*2*np.pi/mu_0 1.6409e+06
####################
W7-X
####################
net current: 1.3875e+04
master d[I]: 1.3888e+07
d[B]*d[a]*2*np.pi/mu_0 7.6768e+06
####################
SOLOVEV
####################
net current: 4.3745e+05
master d[I]: 1.2272e+06
d[B]*d[a]*2*np.pi/mu_0 1.5625e+06
####################
NCSX
####################
net current: -1.7859e+05
master d[I]: 7.5910e+06
d[B]*d[a]*2*np.pi/mu_0 2.7398e+06
####################
reactor_QA
####################
net current: 2.2960e+06
master d[I]: 2.7987e+07
d[B]*d[a]*2*np.pi/mu_0 5.2811e+07
####################
ATF
####################
net current: -1.9183e+06
master d[I]: 2.3656e+07
d[B]*d[a]*2*np.pi/mu_0 6.8601e+06
####################
HELIOTRON
####################
net current: 7.7807e+05
master d[I]: 1.9894e+06
d[B]*d[a]*2*np.pi/mu_0 1.9894e+06
```

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -9,34 +9,44 @@ type: software
 authors:
   - given-names: Daniel
     family-names: Dudt
-    email: ddudt@princeton.edu
-    affiliation: Princeton University
+    email: daniel@thea.energy
+    affiliation: Thea Energy
     orcid: 'https://orcid.org/0000-0002-4557-3529'
   - given-names: Rory
     family-names: Conlin
-    email: wconlin@princeton.edu
-    affiliation: Princeton University
+    email: roryconlin@gmail.com
+    affiliation: University of Maryland, College Park
     orcid: 'https://orcid.org/0000-0001-8366-2111'
   - given-names: Dario
     family-names: Panici
+    email: dpanici@princeton.edu
     affiliation: Princeton University
     orcid: 'https://orcid.org/0000-0003-0736-4360'
-    email: dpanici@princeton.edu
-  - given-names: Egemen
-    family-names: Kolemen
-    email: ekolemen@princeton.edu
-    affiliation: Princeton University
-    orcid: 'https://orcid.org/0000-0003-4212-3247'
   - given-names: Kaya
     family-names: Unalmis
     email: kunalmis@princeton.edu
     affiliation: Princeton University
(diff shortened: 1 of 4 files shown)


dpanici:
> > net current: 2.2516e+05
> > master d[I]: 1.3534e+06
> > d[B]*d[a]_2_np.pi/mu_0 1.6409e+06
> 
> just to confirm: `net current` is the actual current? `master d[I]` is our current scale factor? `d[B]*d[a]*2*np.pi/mu_0` is the proposed new scale factor?

Correct

> I agree that including `a` makes physical sense, but
> 
> * I'm surprised it would make a 2 order of magnitude difference, since `a` is usually ~1m?

I was mistaken, it is usually off by a factor of 2 which can at least drop the value into the correct order of magnitude range.

> * The numbers you posted makes it look like in several cases (dshape, solovev, reactor_QA) the new scaling factor is worse?

Yes it is worse in these cases, but I would rather us be off but with a physically motivated choice than the opposite.

desc/objectives/normalization.py
```diff
@@ -43,7 +43,7 @@ def get_lowest_mode(basis, coeffs):
         scales["V"] = 2 * np.pi * scales["R0"] * scales["A"]
         scales["B"] = scales["Psi"] / scales["A"] * 1.25
```
ddudt, line 44: We could remove this 1.25 fudge factor (assuming it doesn't mess up all the notebooks and tests)

PR #1877, opened 2025-08-25
Increase number of concentric grid nodes near axis

dpanici:
- Makes the innermost ring and the 2nd-innermost ring of points in `ConcentriGrid` have the same points (i.e. make the inner ring poloidal points match those of the 2nd most inner ring)  so that flux surface averages are more correct there.
Tests seem to show that this improves equilibria and free boundary in most cases

Resolves #1876 
Resolves #1985

PR #1921, opened 2025-09-18
Allow PyTree inputs for coil objectives

### LINKED ISSUE #1152: Allow for Pytree input for targets and bounds
Input should allow for different bounds/targets for different coils. I don't think this is possible yet

singh-jaydeep:
Adds new input formats for objectives inheriting from ```_CoilObjective```. In particular, 
 
- ```bounds```, ```target```, and ```weight``` can accept pytrees with values for individual coils, provided the input is a prefix of the input ```CoilSet```. weights can be zero, in which case the corresponding coil is masked.
- ~~A new option ```indices``` accepts pytrees with the same structure, indicating whether the bound/target should be applied to given coils.~~ 
- The attribute ```_broadcast_input``` is required for coil objectives, stating if they are computed per-coil or per-grid node.

Notes
- Resolves #1152. 
- This may not be the cleanest/most general approach, and am open to any suggestions! 
- When the user gives an input which is not just a simple scalar we broadcast the input to be of size dim_f and store it. This isn't very efficient, but saves us from adding new logic to the compute function in ```_Objective```.
- The setter functions in ```_Objective``` will have issues with pytree inputs. Would it make sense to add specific setters in ```_CoilObjective```? [resolved]
- Would require tests, etc. [done]


YigitElma:
@singh-jaydeep is this ready for review?

singh-jaydeep:
I'll have it ready by Monday. There's one thing related to memory usage I wanted to check.

desc/objectives/_coils.py
```diff
@@ -137,6 +161,31 @@ def _prune_coilset_tree(coilset):
             "Only use toroidal resolution for coil grids.",
         )
 
+        def broadcast_input(arr):
```
dpanici: It might make more sense to have this as a (maybe private) method of this objective, then when you make the target setter, you can use this as well.

dpanici:
> * The setter functions in `_Objective` will have issues with pytree inputs. Would it make sense to add specific setters in `_CoilObjective`?

Actually question here: if we always in the `_CoilObjective` clean up the target, bounds, etc before we set these attributes, we don't even need to worry about a target setter, right? If that is the case, I'd prefer to then not need the target setter. maybe I am missing something key here though on why we need the extra target setter. I guess it would mainly be to handle the case where an objective is created, and then its target is updated post-hoc?

desc/objectives/_coils.py
```diff
@@ -220,6 +268,7 @@ class CoilLength(_CoilObjective):
     _scalar = False  # Not always a scalar, if a coilset is passed in
     _units = "(m)"
     _print_value_fmt = "Coil length: "
+    _broadcast_input = "Coil"
```
dpanici, line 363: document that subclasses of `_CoilObjective` shouuld define this attribute and explain the different possible modes (`"Node"`,"`Coil"` etc)

desc/objectives/_coils.py
```diff
@@ -38,7 +40,27 @@ class _CoilObjective(_Objective):
...
+        each entry a float or list satisfying requirements
+        of "target". Cannot be used with "target."
+    weight: float, list, optional
+        Weight for the coil objective during optimization.
+        If a list, must have the same structure as coil.
+    indices: bool, list, optional
+        Boolean indicating whether particular coils in a CoilSet
+        should be optimized according to the objective. Default
+        is to optimize all coils. If a list, must have the same structure
+        as coil.
```
dpanici: this is probably worth some additional description as this is I think the first time `indices` is being used in this specific way (usually it is with a FixedParameter subclass and so indices are the specific parts of params being fixed by the constraint).

Maybe something additional like "Boolean indicating coils in the CoilSet to be targeted by the objective. If True, includes every coil."

singh-jaydeep: Sorry I missed this! Will do.

tests/test_examples.py
```diff
@@ -2294,6 +2294,37 @@ def test_coil_arclength_optimization():
...
+        verbose=3,
+        ftol=1e-4,
+        copy=False,
+    )
+    length_fin = obj.compute(None)
+    length_ref = np.array(
+        [obj._target[i] if obj._mask[i] else length_init[i] for i in range(0, 6)]
+    )
+
+    np.testing.assert_allclose(length_fin, length_ref, rtol=1e-4)
```
dpanici: So in this test, the first coil of the second element of the MixedCoilSet is not included in the optimization, that just means that the gradient of the objective w.r.t the params associated with that coil is zero right?

I guess what is unclear to me is why we would expect no change in its length? It is not like we are not including it at all as an optimizable DOF, it is instead a nullspace direction of our jacobian (it has no effect on f) and thus nothing really is keeping it from changing a lot vs not at all

dpanici (changes requested):
last thing I think I want for this PR: adding an example use of this to the [coil tutorial notebook](https://github.com/PlasmaControl/DESC/blob/master/docs/notebooks/tutorials/coil_stage_two_optimization.ipynb). In that notebook as one of the last cells, we have an optimization of a MixedCoilSet of modular coils and saddle coils. It would be nice to say have the length penalty only target the modular coils (as it effectively does now) and then maybe add some other penalty for the saddle coils, and have this take advantage of the new API as well as show the use of `indices`

And also while we are changing that notebook, adding to the constraints for that last finite beta coil optimization this constraint
`LinkingCurrentConsistency(eq,coilset,eq_fixed=True)` is something I've been meaning to add, and this would be a good spot to change that

@YigitElma do not change the coil notebook on #1877 until this PR is in to avoid annoying conflicts

tests/test_examples.py
singh-jaydeep: I see, yes this just zeros out the gradient associated to that coil's parameters. Would the behavior (whether the coil length still changes, or none of its params change) then depend on the particular optimizer? It passed locally, but maybe it's not a robust test.

Maybe I just test that the explicitly targeted coils do have the length we want, and check the gradients of the objective wrt. the non-optimized coils are zero? Or is there something else which makes sense

singh-jaydeep: I updated the test to separately check the gradients of unoptimized coils and the lengths of optimized coils.

desc/objectives/objective_funs.py
```diff
@@ -1266,6 +1267,8 @@ def compute_unscaled(self, *args, **kwargs):
         f = self.compute(*args, **kwargs)
         if self._loss_function is not None:
             f = self._loss_function(f)
+        if hasattr(self, "_mask"):
```
dpanici: perhaps would be good to leave a comment here saying this is due to the coils objectives, or to add to the _Objective docstring that if _mask is an attribute, it will multiply f here

dpanici (changes requested):
my last change request I promise... I think either leave a comment where mask is used in objective funs to describe why it is there, or add a line about mask to the docstring of _Objective

singh-jaydeep:
The "mask" option is proving to be a bit clunky. In addition to getting the normalization somewhat wrong, it doesn't work well for scalar valued loss functions unless the mask gets passed in (e.g. the mean by default includes the masked values). Fixing these things would further clutter up the objective compute functions, which I don't want to do unless masking is useful for other objectives as well. Is there a preferred way of handling this? 

(Another option is if "indices" gets passed into a coilobjective, when building it changes self.things to only include the coils which are being optimized. This is cleaner in that it keeps the masking within the coil objective only. But the optimizer objects when passing in a coilset and objective with "indices", since the set of optimizable things doesn't match.)

desc/objectives/_coils.py
```diff
@@ -38,10 +40,30 @@ class _CoilObjective(_Objective):
     grid : Grid, list, optional
         Collocation grid containing the nodes to evaluate at.
         If a list, must have the same structure as coil.
-
+    target : float, list, optional
```
YigitElma, line 43: This change won't affect other docstrings, because we don't inherit docstrings from base classes. I think we should update the logic in `desc.objective_funs.collect_docs` for `coil=True` case. You can override the default for coils instead of adding a string to the end as it is currently the case. Something like,
```python

doc_target_coil = """
    target : float, list, optional
        Target values for the coil objective.
        If a float, target is applied to all coils.
        If a list, must have the same structure as coil.
"""
doc_weight_coil = """
...
"""

def collect_docs(
    overwrite=None,
    target_default="",
    bounds_default="",
    normalize_detail=None,
    normalize_target_detail=None,
    loss_detail=None,
    coil=False,
):
    """Collect default parameters for the docstring of Objective.

    Parameters
    ----------
    ...
    coil : bool, optional
        Whether the objective is a coil objective. If ``True``, updates docs
        of ``target``, ``weight``, ``bounds`` and ``loss_function``.
    ...

    """
    doc_params = ""
    for key in docs.keys():
        if overwrite is not None and key in overwrite.keys():
            doc_params += overwrite[key].rstrip()
        else:
            if key == "target":
                target = ""
                if coil:
                    target += doc_target_coil
                if target_default != "":
                    target = target + " Defaults to " + target_default
                if not coil:
                    doc_params += docs[key].rstrip() + target
                else:
                    doc_params += target
            ....
    return doc_params
```
Getting the formatting might be tricky (tabs vs spaces cause trouble sometimes), you can check the doc build locally, running `make html` in your local repo and check the `docs/_build/html` folder. Sometimes updates doesn't show up, then you need to run `make clean`

YigitElma: Once you update that function all the coil objective docs will be updated automatically.

desc/objectives/_coils.py
```diff
@@ -38,10 +40,30 @@ class _CoilObjective(_Objective):
...
+
+    Subclasses must define a static attribute "_broadcast_input." Equals
+    "Coil" if the objective returns a single scalar per coil, and "Node"
+    if it returns a scalar at every grid point. To be compatible with
+    masking, compute function should apply the mask
+    self._coilset_tree["coilset_mask"] before returning data.
     """
 
     __doc__ = __doc__.rstrip() + collect_docs(coil=True)
+    _static_attrs = _Objective._static_attrs + ["_coilset_tree"]
```
YigitElma: Am I missing where you declare `_broadcast_input` as a static attribute? How can it work without that?

desc/objectives/_coils.py
```diff
@@ -747,7 +889,10 @@ class CoilSetMinDistance(_Objective):
         coil=True,
     )
 
-    _static_attrs = _Objective._static_attrs + ["_dist_chunk_size", "_use_softmin"]
+    _static_attrs = _Objective._static_attrs + [
```
YigitElma: if this fits, can you revert the change?

desc/objectives/objective_funs.py
```diff
@@ -1287,7 +1288,8 @@ def compute_scaled_error(self, *args, **kwargs):
         f = self.compute(*args, **kwargs)
         if self._loss_function is not None:
             f = self._loss_function(f)
-        return jnp.atleast_1d(self._scale(self._shift(f), **kwargs))
+        f = jnp.atleast_1d(self._scale(self._shift(f), **kwargs))
```
YigitElma: For consistency with other `compute_` methods, I prefer the old version

YigitElma (commented):
Besides couple places that need change, it looks good to me. Again, thanks for the contribution!

desc/objectives/_coils.py
singh-jaydeep: Added coil objective specific docstrings and changed the logic of collect_docs() a bit.

desc/objectives/_coils.py
singh-jaydeep: Perhaps because it's only used in the init/build parts of the objective? Not sure, but I added it as a static attribute to be safe.

desc/objectives/_coils.py
singh-jaydeep: sounds good

desc/objectives/objective_funs.py
singh-jaydeep: Done

desc/objectives/_coils.py
YigitElma: Thanks! We can technically delete the docstring for target, bounds and weight from here since `__doc__ = __doc__.rstrip() + collect_docs(coil=True)` below will add these. But this class is not shown in official website, so I leave it up to you if you wanna leave it here to make it more visible for devs, I am fine either way

YigitElma (dismissed):
It looks alright to me. I don't think the notebook result is related to this PR, it has been a while since we last ran that notebook, it might be due to some optimization or other code change.

desc/objectives/_coils.py
YigitElma: Hmm, maybe. But usually Jax complains if you have a string or boolean or those kinds of stuff in the pytree of the class, I was expecting to see an error regarding that.

singh-jaydeep:
I installed a fresh venv, and (even on master) can't reproduce the values. Still not sure whether it's architecture differences. I'm on a mac with ARM64. Using JAX 0.6.2 and numpy 2.2.6.

YigitElma (approved):

desc/objectives/_coils.py
```diff
@@ -38,10 +40,30 @@ class _CoilObjective(_Objective):
...
+        Weight for the coil objective during optimization.
+        Default is a uniform weight. If a list, must have
+        the same structure as coil, and consist of non-negative
+        floats. Set weight to zero to exclude given coils from optimization.
+
+    Subclasses must define a static attribute "_broadcast_input." Equals
+    "Coil" if the objective returns a single scalar per coil, and "Node"
+    if it returns a scalar at every grid point. To be compatible with
+    masking, compute function should apply the mask
+    self._coilset_tree["coilset_mask"] before returning data.
```
dpanici, line 62: Just to remind me, this masking is necessary to avoid returning values in the objective where the weights were set to zero?

And just as a hypothetical, if we instead let the weight=0 for some coils, would this render the masking unnecessary?

singh-jaydeep: We got rid of an explicit mask input by allowing weight=0 for some coils, but we still build a mask so that the coilobjective.compute() knows to only return the objective values for unmasked coils. This was to handle loss functions that are min/max/mean. Might be other ways of doing it (e.g. changing logic in Objective class instead).

dpanici: Right right, ok thanks for refreshing my memory

dpanici (approved):

PR #1924, opened 2025-09-19
Updating Perlmutter installation instructions

### LINKED ISSUE #1324: Update the installation instructions for RAVEN and NERSC/Perlmutter
@rahulgaur104 will update the instructions for Perlmutter and RAVEN in a later PR.

_Originally posted by @YigitElma in https://github.com/PlasmaControl/DESC/issues/1293#issuecomment-2436078532_
            

Since they use an older JAX version 0.4.23, current instructions require to change scipy version from requirements file for proper installation. If newer versions of JAX are supported for these clusters, the scipy version change is redundant. JAX should be greater than 0.4.24 ( also this is the new requirement for DESC).

### LINKED ISSUE #1923: DESC installation on Perlmutter doesn't work
I followed the installation instruction verbatim and I get this on Perlmutter.

```
module load cudatoolkit/12.4
module load cudnn/9.5.0
module load conda

conda create -n desc-env python=3.12
conda activate desc-env
pip install --upgrade "jax[cuda12]"

cd DESC
pip install --editable .
```
To test my installation, I import Equilibrium from my login node

```
(desc-env) rgaur@login09:/pscratch/sd/r/rgaur/DESC> python3
Python 3.12.11 | packaged by Anaconda, Inc. | (main, Jun  5 2025, 13:09:17) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
Ctrl click to launch VS Code Native REPL
>>> import desc
>>> from desc.equilibrium import Equilibrium
ERROR:2025-09-19 06:12:41,224:jax._src.xla_bridge:444: Jax plugin configuration error: Exception when calling jax_plugins.xla_cuda12.initialize()
  File "/global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.12/site-packages/jax_plugins/xla_cuda12/__init__.py", line 201, in _version_check
    version = get_version()
              ^^^^^^^^^^^^^
RuntimeError: jaxlib/cuda/versions_helpers.cc:81: operation cusparseGetProperty(MAJOR_VERSION, &major) failed: The cuSPARSE library was not found.

The above exception was the direct cause of the following exception:

  File "/global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.12/site-packages/jax/_src/xla_bridge.py", line 442, in discover_pjrt_plugins
    plugin_module.initialize()
  File "/global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.12/site-packages/jax_plugins/xla_cuda12/__init__.py", line 328, in initialize
    _check_cuda_versions(raise_on_first_error=True)
  File "/global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.12/site-packages/jax_plugins/xla_cuda12/__init__.py", line 266, in _check_cuda_versions
    _version_check("cuSPARSE", cuda_versions.cusparse_get_version,
  File "/global/homes/r/rgaur/.conda/envs/desc-env/lib/python3.12/site-packages/jax_plugins/xla_cuda12/__init__.py", line 205, in _version_check
    raise RuntimeError(err_msg) from e
RuntimeError: Unable to load cuSPARSE. Is it installed?
```
Adding set_device("gpu") before importing Equilibrium doesn't help either.

Also, by default, this process installated jax = jaxlib = 0.6.2 while jax-cuda12-pjrt = jax-cuda12-plugin = 0.7.2. 
Regardless of the version, DESC installation doesn't work on Perlmutter.

rahulgaur104:
Resolves #1923 
Resolves #1324


unalmis:
Does this resolve #1324 too?

rahulgaur104:
> Does this resolve #1324 too?

Yes

unalmis (dismissed):
Merge at will

docs/installation.rst
```diff
@@ -163,30 +163,15 @@ On computing clusters you must call ``module load anaconda`` to use conda (in so
...
             .. code-block:: sh
 
-                module list
-
-            Create a conda environment for DESC following `these instructions <https://docs.nersc.gov/development/languages/python/using-python-perlmutter/#jax>`__.
-
-            .. code-block:: sh
-
-                conda create -n desc-env python=3.12
+                CONDA_OVERRIDE_CUDA="12.4" conda create --name desc-env "jax==0.6.0" "jaxlib==0.6.0=cuda12*" -c conda-forge
```
f0uriest: The pip method yigit and I posted also seems to work fine, and I prefer that to conda-forge as it's what's more supported by the jax devs, and the conda-forge packages seem to be out of date (latest is 0.7.0, but 0.7.2 was released last week).

rahulgaur104: There are subtleties that we realized an hour ago while fixing the installation. I'm updating the PR now.

YigitElma (approved):

rahulgaur104:
@f0uriest could you test this?

docs/installation.rst
```diff
@@ -163,46 +163,33 @@ On computing clusters you must call ``module load anaconda`` to use conda (in so
...
-                pip install --editable .
-
-            You may optionally install developer requirements if you want to run tests.
 
-            .. code-block:: sh
-
-                pip install -r devtools/dev-requirements.txt
+                pip install --no-cache-dir -r devtools/dev-requirements.txt
+                pip install --no-cache-dir --editable .
+                pip install --no-cache-dir "jax[cuda12]"
```
f0uriest: i think we might still need a version pin here for future proofing to avoid accidentally installing an incompatible version? (or possibly on that doesn't play well with nersc)

docs/installation.rst
```diff
@@ -163,46 +163,33 @@ On computing clusters you must call ``module load anaconda`` to use conda (in so
...
 
-            Note that you may also need to execute `unset LD_LIBRARY_PATH` before starting a python process (e.g. execute this as part of your slurm script, before calling python to run DESC) for the JAX/CUDA initialization to work properly.
+        The `--no-cache-dir` avoids conflicts with existing DESC environments or other software that use CUDA on your system.
 
+        Before running a DESC script, you MUST also execute `unset LD_LIBRARY_PATH` either in your interactive node (for interactive jobs)
```
f0uriest: I tried these instructions on both a login node and jupyter worker and didn't need this. Are you both sure it's necessary, and if so do we know why?

rahulgaur104: I already had one DESC environment and I was trying to install another. To make sure the new environment works, I needed this command.

YigitElma: Maybe the `MUST` part is too strict. I tried the instructions on my fresh Perlmutter account, so my pip cache was completely empty. I didn't need this when I was using compute node. But for Rahul, the older environments seemed to cause problems, and the only way to fix it was `--no-cache-dir` and `unset LD_LIBRARY_PATH`. We can make the `LD_LIBRARY_PATH` step optional (although it doesn't hurt to type).

docs/installation.rst
YigitElma: I used this command in many different systems, and it seems to work consistently, even if there is a newer version of jax (for example, 0.7.2 is out but I use 0.6.2) it grabs the one installed for CPU before on the environment. If people change the order, like call `pip install --no-cache-dir -r devtools/dev-requirements.txt` after jax for GPU, it will definitely break and install back CPU (we experienced this with Rahul yesterday). 

This is a guess but I think what is happening is that the difference for CPU GPU jax is not the `jax` itself but the `jaxlib`. So, if you already have `jax` (installed during CPU phase), the corresponding `jaxlib` checks the version of that, and they are in sync. Of course, people can include other packages or other steps which can mess it up. If there is a clean way to do this, I am fine (but let me say, I really don't like the `sed` command)

rahulgaur104:
Actually, I was doing optimization and I see other anomalies. I modify the function errorif in desc/utils.py

```python3
   7 def errorif(cond, err=ValueError, msg=""):
   6     """Raise an error if condition is met.
   5 
   4     Similar to assert but allows wider range of Error types, rather than
   3     just AssertionError.
   2     """
   1     if cond:
   1         try:
   2             jax.debug.print(msg)
   3         except:
   4             pass
   5         raise err(colored(msg, "red"))
   
```
   and it prints 
   
   `Grid does not have unique indices assigned. It is not possible to do this automatically on grids made under JIT.`
   
   hundreds of times before and during optimization. Eventually, I get the following error
   
```
    5   File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/objectives/_omnigenity.py", line 679, in build
   4     errorif(
   3   File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/utils.py", line 564, in errorif
   2     raise err(colored(msg, "red"))
   1           ^^^^^^^^^^^^^^^^^^^^^^^^
   0  TypeError: 'str' object is not callable
```

But this error only occurs sometimes.

dpanici:
> Actually, I was doing optimization and I see other anomalies. I modify the function errorif in desc/utils.py
> 
> ```python
>    7 def errorif(cond, err=ValueError, msg=""):
>    6     """Raise an error if condition is met.
>    5 
>    4     Similar to assert but allows wider range of Error types, rather than
>    3     just AssertionError.
>    2     """
>    1     if cond:
>    1         try:
>    2             jax.debug.print(msg)
>    3         except:
>    4             pass
>    5         raise err(colored(msg, "red"))
>    
> ```
> 
> and it prints
> 
> `Grid does not have unique indices assigned. It is not possible to do this automatically on grids made under JIT.`
> 
> hundreds of times before and during optimization. Eventually, I get the following error
> 
> ```
>     5   File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/objectives/_omnigenity.py", line 679, in build
>    4     errorif(
>    3   File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/utils.py", line 564, in errorif
>    2     raise err(colored(msg, "red"))
>    1           ^^^^^^^^^^^^^^^^^^^^^^^^
>    0  TypeError: 'str' object is not callable
> ```
> 
> But this error only occurs sometimes.

This seems weird, if you run the same script on your local laptop do you get similar errors? i.e. is it a jax thing or just a thing with custom grids?

rahulgaur104:
The script that I am running on a GPU is memory intensive. Let me reduce it to a low-memory minimally failing case and then run it on my local system

YigitElma:
> Grid does not have unique indices assigned. It is not possible to do this automatically on grids made under JIT.

This doesn't look like an installation issue to me. Are you sure this only happens on Perlmutter?

docs/installation.rst
f0uriest: I tried this both in a totally clean conda setup (no previous environments) and with other environments and in both cases I don't need it, with any recent versions of jax. It might be something specific to your situations? IE do you have any stuff in your bashrc or other codes installed? 

I think we should move this to a "if you see this issue here's a possible fix" but I definitely think we shouldn't say this has to be done. Note that `LD_LIBRARY_PATH` exists outside of conda, and is used by lots of codes and libraries (iirc i had to mess with it when using sfincs, stride, dcon etc), so unsetting it may break other things people have installed, even if they create a fresh conda environment for desc, and we really don't want that.

docs/installation.rst
rahulgaur104: Have to tried another version using this process?

Here's what I did. I set the max version of jax to <=0.6.1 in requirements.txt and repeated the exact same process, even making sure to use
`pip install --no-cache-dir "jax[cuda12]==0.6.1" "jaxlib[cuda12]==0.6.1"
`
 but now I get the following error:

```
E0925 15:05:46.285344  817786 pjrt_stream_executor_client.cc:2917] Execution of replica 0 failed: INTERNAL: jaxlib/gpu/solver_handle_pool.cc:37: operation gpusolverDnCreate(&handle) failed: cuSolver internal error
  File "<stdin>", line 1, in <module>
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/equilibrium/__init__.py", line 3, in <module>
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/equilibrium/equilibrium.py", line 16, in <module>
    from desc.compute import compute as compute_fun
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/compute/__init__.py", line 30, in <module>
    from . import (
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/compute/_bootstrap.py", line 16, in <module>
    from ..integrals.surface_integral import surface_averages_map
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/integrals/__init__.py", line 3, in <module>
    from .bounce_integral import Bounce1D, Bounce2D
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/integrals/bounce_integral.py", line 122, in <module>
    leggauss(32),
    ^^^^^^^^^^^^
  File "/global/homes/r/rgaur/.conda/envs/desc-env2/lib/python3.12/site-packages/orthax/legendre.py", line 1509, in leggauss
    x = jnp.linalg.eigvalsh(m)
        ^^^^^^^^^^^^^^^^^^^^^^
jaxlib._jax.XlaRuntimeError: INTERNAL: jaxlib/gpu/solver_handle_pool.cc:37: operation gpusolverDnCreate(&handle) failed: cuSolver internal error
--------------------
```

rahulgaur104:
Here's what I did. I set the max version of jax to <=0.6.1 in requirements.txt and repeated the exact same process, even making sure to use
`pip install --no-cache-dir "jax[cuda12]==0.6.1" "jaxlib[cuda12]==0.6.1"
`
 but now I get the following error:

```
E0925 15:05:46.285344  817786 pjrt_stream_executor_client.cc:2917] Execution of replica 0 failed: INTERNAL: jaxlib/gpu/solver_handle_pool.cc:37: operation gpusolverDnCreate(&handle) failed: cuSolver internal error
  File "<stdin>", line 1, in <module>
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/equilibrium/__init__.py", line 3, in <module>
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/equilibrium/equilibrium.py", line 16, in <module>
    from desc.compute import compute as compute_fun
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/compute/__init__.py", line 30, in <module>
    from . import (
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/compute/_bootstrap.py", line 16, in <module>
    from ..integrals.surface_integral import surface_averages_map
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/integrals/__init__.py", line 3, in <module>
    from .bounce_integral import Bounce1D, Bounce2D
  File "/pscratch/sd/r/rgaur/DESC2/DESC/desc/integrals/bounce_integral.py", line 122, in <module>
    leggauss(32),
    ^^^^^^^^^^^^
  File "/global/homes/r/rgaur/.conda/envs/desc-env2/lib/python3.12/site-packages/orthax/legendre.py", line 1509, in leggauss
    x = jnp.linalg.eigvalsh(m)
        ^^^^^^^^^^^^^^^^^^^^^^
jaxlib._jax.XlaRuntimeError: INTERNAL: jaxlib/gpu/solver_handle_pool.cc:37: operation gpusolverDnCreate(&handle) failed: cuSolver internal error
--------------------
```

If I repeat the same process with jax==0.6.2 I get not errors, atleast while loading things.

docs/installation.rst
rahulgaur104: Nothing in bashrc. No recent codes but I may have gx from an old installation.

Also, when you say you tested, what exactly are you doing? import desc and exit? 
You should explain your test because I worry it's something trivial.

rahulgaur104: I am happy to change the wording but my focus is to make it robust or atleast understand the problem so next time something like this happens we can create a stable solution quickly!

rahulgaur104: I'll update the wording I promise but there are other problems either with Perlmutter and/or jax.

docs/installation.rst
YigitElma: We already have the solution for that in docs.
image

YigitElma:
> Here's what I did. I set the max version of jax to <=0.6.1 in requirements.txt and repeated the exact same process, even making sure to use
> `pip install --no-cache-dir "jax[cuda12]==0.6.1" "jaxlib[cuda12]==0.6.1" `
> but now I get the following error:

See response above https://github.com/PlasmaControl/DESC/pull/1924#discussion_r2380555752

docs/installation.rst
rahulgaur104: What about 0.6.0? Are you saying it should work if I repeat the same process with 0.6.0?

YigitElma: As far as I remember this was specific to 0.6.1. So, yes, it should work without issue.

docs/installation.rst
f0uriest: my usual test is 3 fold:
```
import jax; jax.devices("gpu")
```
do this before anything with desc, if it fails i know its a jax issue not a desc issue
```
from desc import set_device; set_device("gpu")
```
to see if desc introduces any further issues
```
python -m desc desc/examples/DSHAPE -vv -g
```
runs the dshape example on gpu, to make sure everything actually works

you can then verify with `nvidia-smi` that jax is running on gpu by looking at memory and usage

YigitElma (approved):

f0uriest (approved):

dpanici (changes requested):
wait for CI before merging to check #1946 worked

dpanici (approved):

Issue #1599, opened 2025-02-20
note to make minimal working example issues upstream


unalmis:
In #1440 and #1360 it was observed that the matrix multiply and linear algebra solve respectively gave wrong results when the tensor's size became large. This is a note to self to simplify the examples and report to upstream repositories.

unalmis:
Here is one example. The `test_harmonic_general` test fails unless `chunk_size` is reduced. https://github.com/PlasmaControl/DESC/blob/869c81ba35827320a02e15c58ffbfab2afc7232d/tests/test_integrals.py#L762

The reason it fails is because the matrix multiplication in the line linked below is performed incorrectly when `chunk_size` is large.
https://github.com/PlasmaControl/DESC/blob/869c81ba35827320a02e15c58ffbfab2afc7232d/desc/integrals/singularities.py#L504

unalmis:
hunch: Sharp transition in error when you increase batch size past a cutoff, and slower transition from good to bad in matrix error. Check this by scanning over different options

f0uriest:
Possibly related: https://github.com/jax-ml/jax/issues/27591

unalmis:
Posting parts of previous conversation here.
 
- For example, compare the following computations. In particular the only change in the W7-X one is changing the free boundary chunk size computation from 1000 to 100. For the elliptical cross section one, the chunk size is 1000 but the surface parameterization is simpler, so the tensor products for the integrations are smaller.
- The reason the error in grad(Phi) is supposed to be higher for W7-X is because computing it requires a principal value integral, whereas Phi is an integrable one.
- I never see these issues with the bounce integrals, which are also large computations, because I manually vectorized the bounce integrals (no vmap).

[torsion_error.pdf](https://github.com/user-attachments/files/19969822/torsion_error.pdf)
[w7x_error.pdf](https://github.com/user-attachments/files/19969823/w7x_error.pdf)
[w7x_error2.pdf](https://github.com/user-attachments/files/19969821/w7x_error2.pdf)

PR #1926, opened 2025-09-20
Hotfix for batching in older jax versions

dpanici:
hotfix, add compatibility layer to sharding.update call as it is different api before jax 0.6.2

### DIFF
--- a/desc/batching.py
+++ b/desc/batching.py
@@ -41,6 +41,21 @@ if jax.__version_info__ >= (0, 4, 16):
 else:
     from jax import linear_util as lu
 
+if jax.__version_info__ < (0, 6, 2):
+    # older versions dont have sharding.update, it was changed in JAX 0.6.2
+    # in https://github.com/jax-ml/jax/commit/7904b86bbacdb72f7769a14b5e5d07bba05bd9e8
+    def _sharding_update(aval):
+        return aval.sharding.with_spec(
+            PartitionSpec(None, None, *aval.sharding.spec[1:])
+        )
+
+else:
+
+    def _sharding_update(aval):
+        return aval.sharding.update(
+            spec=PartitionSpec(None, None, *aval.sharding.spec[1:])
+        )
+
 
 def _scan_leaf(leaf, batch_elems, num_batches, batch_size):
     """https://github.com/jax-ml/jax/blob/main/jax/_src/lax/control_flow/loops.py.
@@ -62,9 +77,7 @@ def _scan_leaf(leaf, batch_elems, num_batches, batch_size):
             f" Got {aval.str_short(True, True)}"
         )
 
-    out_s = aval.sharding.update(
-        spec=PartitionSpec(None, None, *aval.sharding.spec[1:])
-    )
+    out_s = _sharding_update(aval)
     out_s = canonicalize_sharding(out_s, "lax.map")
     if out_s is not None and out_s.mesh._any_axis_explicit:
         return auto_axes(f, out_sharding=out_s, axes=out_s.mesh.explicit_axes)(leaf)


unalmis:
some comments
1. If a full reversion is being done, should make a new PR that also includes the reversion of this reversion.
1. It sounds like this reversion would prevent use of DESC on JAX versions beyond 0.6.2. ?
2. The changes we backported from JAX (I think also include some JAX versions before 0.6.2) include JAX's bugfixes for various batching issues. As this PR intends to revert those fixes, this leads to a danger in shipping a version frankenstein of code. If the JAX version is one which assumes those are fixed and they aren't we would get subtle issues.
3. It is hard enough to convince developers of other dependencies to fix their bugs. Anything short of a creating a GitHub issue with the actual bugfix, or even more difficult: with a minimial working example, always gets ignored. That is my experience with some jax dependencies at least. I doubt anyone will take #1599 seriously if we mix jax versions for example. I am especially hesitant because JAX is still fixing batching issues that caused subtle bugs in DESC #1522 and maybe still affects #1894 in some way which took a very long time to isolate and delayed work.

so basically be careful

dpanici:
> some comments
> 
> 1. If a full reversion is being done, should make a new PR that also includes the reversion of this reversion.

can you explain this part? I dont follow

unalmis (changes requested):
> > so basically be careful
> 
> 
> 
> Fair, I need to think more. I welcome any suggestions
> we either need to
> 
> - revert the changes that caused these specific import errors or
> 
> - change our requirements to jax ==0.6.2

Here is how I would have done the original PR if the jax tests had failed there it.
There are 3 functions on master that we took  from the newest jax version.

If you want to preserve DESC ability to run old jax versions (even though these old jax versions have bugs that effect our stuff), and there is no simple fix, then it's safer to make new legacy batching functions for those 3 functions.

Call those legacy functions only when jax has an old version.
see the top of the desc.batching file to see how to check the jax version. 

So update the citation of those 3 functions to include the version you are taking the code from. 

```
if jax < compatible version:
    from jax import batch_abd_renainder
else:
    def batch_and....
        """..... jax version xxxx"""
         stuff
```

.github/workflows/jax_tests.yml
```diff
@@ -14,15 +14,16 @@ jobs:
       matrix:
         jax-version: [0.4.24, 0.4.25, 0.4.26, 0.4.27, 0.4.28, 0.4.29,
                       0.4.30, 0.4.31, 0.4.33, 0.4.34, 0.4.35, 0.4.37,
-                      0.4.38, 0.5.0, 0.5.1, 0.5.2, 0.5.3, 0.6.0]
+                      0.4.38, 0.5.0, 0.5.1, 0.5.2, 0.5.3, 0.6.0, 0.6.1,
```
unalmis: some of 0.5. is broken (we block from install)

desc/batching.py
```diff
@@ -41,90 +38,124 @@
 else:
     from jax import linear_util as lu
 
+try:  # noqa : C901
```
unalmis: why not just try to import the batch and remainder function from jax and if it is not there then you use the one from your favorite jax version that passes all the github CI tests?

.github/workflows/jax_tests.yml
YigitElma: Can you remind me what the problem was exactly? Numerics or installation failure?

unalmis: the latter, see commit message of #1868

desc/batching.py
YigitElma: @f0uriest and you were not ok with that previously (conversation is on slack)

unalmis: In that slack conversation

1. I said the point of having these function instead of importing from JAX is because the old versions of JAX that DESC claims to support don't have the function. Therefore, I had to rewrite an implementation of it in DESC. I chose the most recent JAX implementation because that had the least bugs. If you are already providing a fallback method to use when JAX lacks the feature, then you can just import the proper method from jax in the try block.

2. Rory's point was that you shouldn't import private stuff from JAX. I don't follow how that applies because to reimplement this in DESC you have to import even more private functions from JAX. See your import from jax._src.util for example.

YigitElma: Whoever convinced him before can deal with it; I just resolved the issue.

unalmis: Ok, then if the fallback method you chose passes all the jax tests, then I will make the update the use the jax method if the import is succesful.

unalmis: Looks like the fallback didn't work. I think it will be more likely to work with the method I suggested in the requested changes (which is same as the suggestion I made in this thread).

YigitElma: The failing ones fail due to something else (`rtol` argument to `jnp.pinv` which is added in 0.4.29 apparently), let's see the rest

unalmis: The ones above 0.4.29 fail as well, so the fallback should be taken from an older jax version.

.github/workflows/jax_tests.yml
```diff
@@ -14,15 +14,16 @@ jobs:
       matrix:
         jax-version: [0.4.24, 0.4.25, 0.4.26, 0.4.27, 0.4.28, 0.4.29,
                       0.4.30, 0.4.31, 0.4.33, 0.4.34, 0.4.35, 0.4.37,
-                      0.4.38, 0.5.0, 0.5.1, 0.5.2, 0.5.3, 0.6.0]
+                      0.4.38, 0.5.0, 0.5.3, 0.6.0, 0.6.1, 0.6.2]
```
f0uriest: while we're at it we might as well also test against newer versions (through 0.7.2) and bump the max version accordingly

unalmis: - are you going to block the 0.7.0 due to the performance regressions in those versions? think it was with diffrax
- should test after nuffts are added as dependency. If the nuffts don't work with 0.7.2 yet i'm just going to revert and block 0.7.2.


0.7.2 came out this week and JAX advertises breaking changes

unalmis:
Thank you for removing 0.7.2 as i requested. As you can see, the bug reports for batching in 0.7.2 are now beginning. Look up issue 32028

unalmis (approved):

unalmis (approved):

desc/batching.py
unalmis: done

requirements.txt
```diff
@@ -1,4 +1,4 @@
-jax >= 0.4.24, != 0.4.36, != 0.5.1, != 0.5.2, <= 0.6.2
+jax >= 0.4.29, != 0.4.36, != 0.5.1, != 0.5.2, <= 0.6.2
```
f0uriest, line 1: Are we bumping the max version here and excluding 0.7.0?

unalmis:
I try not to add workarounds because it would pollute our code base. From my memory, there's a series of bugs with FFT accuracy before 0.6.2 and AD with indexing multi-dim arrays.

unalmis:
yes

unalmis (dismissed):

tests/test_compute_funs.py
```diff
@@ -1546,13 +1546,13 @@ def test_contravariant_basis_vectors_PEST(eq):
...
 @pytest.mark.slow
-@pytest.mark.parametrize("eq", [get("W7-X")])
+@pytest.mark.parametrize("eq", [get("precise_QA")])
 def test_PEST_derivative_math(eq):
     """Verify math to write PEST derivative quantities by redefining θ to θ_PEST."""
     from desc.compute import data_index
 
     tol = 1e-10
-    eq_PEST = eq.to_sfl(4 * eq.L, 5 * eq.M, 4 * eq.N, copy=True, tol=tol)
+    eq_PEST = eq.to_sfl(3 * eq.L, 3 * eq.M, 3 * eq.N, copy=True, tol=tol)
```
rahulgaur104, line 1556: Maybe add a comment here explaining that the rtol can be reduced if the resolution is increased. Maybe add a TODO link to Kaya's PR and/or comment.

dpanici, line 1556:
```suggestion
    #TODO: can reduce rtol of test if resolution is increased. See DESC git #1919
    eq_PEST = eq.to_sfl(3 * eq.L, 3 * eq.M, 3 * eq.N, copy=True, tol=tol)
```

dpanici:
Ok no more changes, once tests pass and we confirm they all are passing we can merge

YigitElma (approved):

Issue #1937, opened 2025-09-26
GPU Features Enabled by JAX-FINUFFT Installation


dpanici:
> Should state what type of computations make use of this. Currently there are not that many, we can keep the list updated future. 

 _Originally posted by @YigitElma in [#1834](https://github.com/PlasmaControl/DESC/pull/1834/files#r2326114088)_

Current features which require jax-finufft GPU installation in order to be used
- using nuffts for bounce integral-related, generally these can still be ran on GPU by setting `nufft_eps=0` to use the older, less efficient implementation. This includes
  - `EffectiveRipple` and `GammaC` with `nufft_eps>0`
  - `plot_gammac`

rahulgaur104:
> Current features which require jax-finufft GPU installation in order to be used
Is this already in master?

dpanici:
No, this is to document things once #1834 is in, I'd prefer that we list it somewhere

unalmis:
I looked through the `jax-finufft` repository and the minimum supported `jax` version there is 0.5.0.

I'd like to bump our minimum `jax` version to that in the nuffts PR as well, but we could also just remove versions less than that from the `test_jax` actions script.

PR #1944, opened 2025-09-29
Remove FINUFFT warning

unalmis:
Replaces jax-finufft's warning with the one desc developers want.


dpanici (approved):

desc/integrals/_interp_utils.py
```diff
@@ -972,3 +972,27 @@ def trig_vander(x, n, domain=(0, 2 * jnp.pi)):
     return jnp.concatenate(
         [jnp.sin(nx[..., n_rfft.size - is_even - 1 : 0 : -1]), jnp.cos(nx)], axis=-1
     )
+
+
+def _test_gpu_jax_finufft():
+    """Replacing jax-finufft's warning with ours."""
+    from tests.test_interp_utils import TestFastInterp, _test_inputs_1D
```
f0uriest: The test folder isn't always included, ie if you install from pip. I don't think we need to run the full test anyways, we just need to test that it jits ok on gpu

desc/integrals/_interp_utils.py
```diff
@@ -972,3 +972,27 @@ def trig_vander(x, n, domain=(0, 2 * jnp.pi)):
...
+    try:
+        from jax_finufft import nufft2  # noqa: F401
+    except ImportError:
+        return
+
+    try:
+        TestFastInterp.test_non_uniform_real_FFT(*_test_inputs_1D[0])
+    except NotImplementedError:
+        warnings.warn(
+            "\njax-finufft is not installed.\n"
```
f0uriest: At this point we know that finufft is installed in some capacity from the previous lines, this is mainly checking GPU support so the message here should make that clear

desc/integrals/_interp_utils.py
unalmis: I prefer to do better than "if it compiles, ship it"

desc/integrals/_interp_utils.py
```diff
@@ -972,3 +965,8 @@ def trig_vander(x, n, domain=(0, 2 * jnp.pi)):
     return jnp.concatenate(
         [jnp.sin(nx[..., n_rfft.size - is_even - 1 : 0 : -1]), jnp.cos(nx)], axis=-1
     )
+
+
+from desc._checks import check_jax_finufft  # noqa: E402
+
+check_jax_finufft()
```
YigitElma: Any basic import from desc is already taking a lot of time, adding this to the file will always run the test which will add a finite amount of time to the start-up. This is especially annoying for `ExternalObjective` kind of thing that use multiprocessing (i.e. each process needs to wait 6 7 seconds just to import stuff). I will repeat myself but the number of places that depends on finufft is less than 5, I prefer having this overhead for only those use cases.

YigitElma (changes requested):
I still think these kind of checks that include computation must be application specific. For example, on my laptop, I use GPU but the installation instruction doesn't work, so I just installed jax-finufft cpu to shut the warning down, but with this, I will always get this warning. The additional import time is also not necessary.

Also, we can either use this PR or some other one, but the installation instructions are not complete for jax-finufft. 
- For example, the local GPU one doesn't work. You need to install CudaToolkit, we should at least mention that (I haven't tested it with Cuda Toolkit because it might cause my other environments to fail with some CUDA related errors, I don't want to take that risk). 
- Della instructions only work on `della-gpu` login node, this again needs explanation. 
- Another thing is that if you try to use the environment with a H100, you cannot, you need to compile it again in a different environment with `CMAKE_CUDA_ARCHITECTURES=90` or something like that. Because on Della and Perlmutter, login nodes have A100 with 8.0 architecture.
- I shared the Perlmutter instructions in #1937. They are very very limiting but work. I tried making it more general, but looks like if you change one line of the instructions, everything fails.

desc/integrals/_interp_utils.py
dpanici: master takes 6s, this PR is like almost 7s, so we already have a long startup.

dpanici:
Ideally actually, can this be done within the objectives/compute functions which use jax-finufft? instead of being upon import or in installation?

YigitElma:
> I like this idea of making the user test their installation instead of checking it every time the functions are imported. But we should add a note in the [Verifying your Installation](https://desc-docs.readthedocs.io/en/latest/installation.html#verifying-your-installation) section with example code of how to test it. Otherwise this looks good.

I think we already have that
```python
from desc import set_device
set_device("gpu")

from desc.examples import get
from desc.objectives import ObjectiveFunction, GammaC

obj = ObjectiveFunction(GammaC(get("W7-X"), num_transit=1, num_pitch=1))
obj.build()
x = obj.x()
obj.compute_scaled_error(x).block_until_ready()
```
This is under the `CPU+GPU` tab of verify your installation. CPU version should work out of the box.

dpanici:
https://github.com/flatironinstitute/jax-finufft/issues/206
FYI

docs/installation.rst
unalmis, line 1: all the changes here need double ticks for code not single ticks

PR #1946, opened 2025-09-30
CI related fixes

### LINKED ISSUE #1852: Add disclaimer to GitHub benchmarks script
Should add disclaimer to benchmarks saying that they do not work well for measuring performance. For example in #1842 it says objective.compute performance decreases by 20%. In reality nothing has changed, and this noise is due to well known issues with github benchmarks . Developers can be  misled unless this warning is printed with the benchmarks.

### LINKED ISSUE #1913: Bypass memory benchmarks for documentation changes
I couldn't find a way to add label-based exceptions to codecov, so these types of PR's are always expected to be merged by override access.

For memory, I forgot to add this logic to the workflow, if you want you can add it in this PR. Basically add,

```python
- name: Filter changes
        id: changes
        uses: dorny/paths-filter@v3
        with:
          filters: |
            has_changes:
              - 'desc/**'
              - 'requirements.txt'
              - 'devtools/dev-requirements.txt'
              - '.github/workflows/memory_benchmark.yml'

        id: check_changes
        run: echo "has_changes=${{ !contains(github.event.pull_request.labels.*.name, 'only-docs-comments') && steps.changes.outputs.has_changes }}" >> $GITHUB_ENV
```

Before setup python step. Then add, 
```
if: env.has_changes == 'true'
```
to other steps. You can look for the workflow yml's of unit/regression tests for implementation. Note: if you do it in this PR, it will probably still run for this one, but once it is merged, it will be resolved.

Once benchmarks start, it will skip actual benchmarks, it will take couple seconds.

_Originally posted by @YigitElma in https://github.com/PlasmaControl/DESC/pull/1862#issuecomment-3201699472_

### LINKED ISSUE #1929: Codecov should not be run on PR's marked to be only documentation changes
Also override-codecov doesn't work

YigitElma:
- Skip memory benchmarks for `only-docs-comments` labeled PRs
- Fail memory benchmark if the underlying test fails
- Skip codecov if no report is uploaded for any of the workflows
- Add comment on the possible noise on benchmarks

Resolves #1913 
Resolves #1929 
Resolves #1852

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,16 +3,18 @@ comment:                  # this is a top-level key
   behavior: default
   require_changes: false  # if true: only post the comment if coverage changes
   require_base: true        # [true :: must have a base report to post]
-  require_head: true       # [true :: must have a head report to post]
-  after_n_builds: 14
+  require_head: true        # [true :: must have a head report to post]
+  after_n_builds: 14   # wait for 8 unit and 6 regression tests to complete
 coverage:
   status:
     patch:
       default:
         informational: false
+        if_not_found: success # for 'only-docs-comments' labeled PRs
     project:
       default:
         informational: false
         threshold: 2.0
+        if_not_found: success # for 'only-docs-comments' labeled PRs
 github_checks:
   annotations: true
--- a/.github/workflows/memory_benchmark.yml
+++ b/.github/workflows/memory_benchmark.yml
@@ -14,12 +14,29 @@ jobs:
       GH_TOKEN: ${{ github.token }}
     strategy:
       matrix:
-        python-version: ['3.12']
+        python-version: ["3.12"]
 
     steps:
       - uses: actions/checkout@v5
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Filter changes
+        id: changes
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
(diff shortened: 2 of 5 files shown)


codecov.yml
```diff
@@ -3,16 +3,18 @@ comment:                  # this is a top-level key
...
-  require_head: true       # [true :: must have a head report to post]
-  after_n_builds: 14
+  require_head: true        # [true :: must have a head report to post]
+  after_n_builds: 14   # wait for 8 unit and 6 regression tests to complete
 coverage:
   status:
     patch:
       default:
         informational: false
+        if_not_found: success # for 'only-docs-comments' labeled PRs
```
f0uriest, line 13: just so i understand this:
- if no reports are uploaded (ie, no tests were run bc it just doc updates) then this will succeed
- If eg, 13/14 reports are uploaded (ie bc one of the unit tests fails) then this will still fail? (I think we want it to fail in this case)

YigitElma: Yes, this is the way I understand it from their documentation.

Our unit and regression tests upload a report even if they fail (upload step runs even for pytest failure, see `always()` at the condition). However, if 14 tests ran and still the status is pending/expecting, then this change should make the codecov pass.

We can use Perlmutter PR as a test for this. I am also not 100% sure if it is going to work.

PR #1949, opened 2025-10-02
Fix failing benchmarks

unalmis:
This PR fixes failing benchmarks as required from changes made in #1946.

#1834 PR will rename the variable `spline` to `use_bounce1d`, and without this change, both memory and performance benchmarks fail.


YigitElma (approved):
This should also solve the other benchmark issue, right?

unalmis:
It will solve the failing benchmark on my open prs but I never made an issue for that.

Issue #1584, opened 2025-02-14
`plot_surfaces` vartheta contours can be discontinuous at boundary, likely due to rootfind fails


dpanici:
For the following input file (attached), the resulting asymmetric equilibrium, when `plot_surfaces` is called, has the vartheta contours make a sharp change at the boundary, when if you just do `plot_section(eq,"theta_PEST")`, you see none of this behavior, indicating something is happening with the rootfind for `theta_PEST` contours during the `plot_surfaces` call. 

The equilibrium is perfectly nested so it is not an equilibrium issue. Maybe we can just do a grid calculation of `theta_PEST` and make contours based off of that instead of doing the rootfind for the `plot_surfaces`?

[input.asym_tokamak.txt](https://github.com/user-attachments/files/18804111/input.asym_tokamak.txt)

![Image](https://github.com/user-attachments/assets/a1348be2-af24-46b1-9e92-b378d4bf41d7)

unalmis:
Check if removing period argument from fixes it
1. https://github.com/PlasmaControl/DESC/blob/89cedb4fe06cdecaed0f4ef66fe89e16246b897f/desc/plotting.py#L1710
2.https://github.com/PlasmaControl/DESC/blob/89cedb4fe06cdecaed0f4ef66fe89e16246b897f/desc/plotting.py#L1688

dpanici:
check lambda derivative wrt theta, if it is >1 or not

dpanici:
![Image](https://github.com/user-attachments/assets/672b3df8-bb60-480f-9e76-d2492db03b74)

d lambda d theta > 1 at some points in the plasma, probably explains this issue

dpanici:
Spectral condensation of the boundary helps with this issue immensely and lowers the value of $d\lambda/d\theta$ we see. Closing this as resolved but this is motivation to make a spectral condensation routine for surface fitting available in the code

dpanici:
I see this even in less strongly shaped cases, this can also happen due to poor initial guesses for the vartheta contour. In master currently the general rootfinding `map_coordinates` is used with `inbasis=["rho","theta_PEST","phi"]` which does NOT use the `_map_PEST...` method inside `map_coordinates` (as it is allowing for `omega` nonzero and so performing the rootfind in all the coordinates). 

<img width="492" height="703" alt="Image" src="https://github.com/user-attachments/assets/5902f8e8-36d6-4c8b-a98e-a307a7d6cf18" />

[...]

PR #1951, opened 2025-10-02
Change plot_surfaces to use more robust _map_PEST_coordinates, until generalized toroidal angle is in

dpanici:


desc/plotting.py
```diff
@@ -1840,10 +1840,13 @@ def plot_surfaces(eq, rho=8, theta=8, phi=None, ax=None, return_data=False, **kw
             map_coordinates(
                 eq,
                 t_grid.nodes,
-                ["rho", "theta_PEST", "phi"],
-                ["rho", "theta", "zeta"],
+                # TODO: once generalized toroidal angle is used, change
+                # inbasis to ["rho", "theta_PEST", "phi"],
```
dpanici, line 1845: Change this back once #568 is in

unalmis:
https://github.com/PlasmaControl/DESC/issues/1584#issuecomment-3363356650

desc/plotting.py
f0uriest: can you include the PR number in the comment? I think we did that for the rest of the places (eg `TODO(#568)` and it makes it easier to grep for things that need to be changed

YigitElma (approved):

PR #1963, opened 2025-10-10
Prevent weak type promotion to reduce recompilation

### LINKED ISSUE #1581: Extra compilation in `perturb`
When I was playing around with the benchmarks, I noticed that `perturb` takes 2 runs to compile fully. Normally, when I call `eq.perturb(order=2)` 3 times, I would expect the first one to take long time due to compilation of first and second derivatives, but then significantly faster second and third iterations. However, it seems like it takes 2 runs to fully compile the perturbation. Here are some results.

```python
eq = desc.examples.get("ATF")
eq.change_resolution(12, 12, 4, 24, 24, 8)
tr_ratio = [0.01, 0.25, 0.25]
dp = np.zeros_like(eq.p_l)
dp[np.array([0, 2])] = 8e3 * np.array([1, -1])
deltas = {"p_l": dp}
obj = get_equilibrium_objective(eq)
obj.build(eq)
con = maybe_add_self_consistency(eq, constraints)
eq.perturb(objective=obj, constraints=con, deltas=deltas, tr_ratio=tr_ratio, verbose=3)
```
```
Building objective: lcfs R
Building objective: lcfs Z
Building objective: fixed Psi
Building objective: fixed pressure
Building objective: fixed iota
Building objective: fixed sheet current
Building objective: self_consistency R
Building objective: self_consistency Z
Building objective: lambda gauge
Building objective: axis R self consistency
Building objective: axis Z self consistency
Timer: Objective build = 858 ms
Perturbing p_l
Factorizing linear constraints
Timer: linear constraint factorize = 4.59 sec
Computing df
Timer: df computation = 7.35 sec
Factoring df
Timer: df/dx factorization = 148 ms
Computing d^2f
Timer: d^2f computation = 7.86 sec
Timer: Objective build = 1.86 ms
||dx||/||x|| =  1.746e-03
Timer: Total perturbation = 27.3 sec
```
```python
eq.perturb(objective=obj, constraints=con, deltas=deltas, tr_ratio=tr_ratio, verbose=3)
```
```
Timer: Objective build = 13.6 ms
Perturbing p_l
Factorizing linear constraints
Timer: linear constraint factorize = 2.69 sec
Computing df
Timer: df computation = 8.78 sec
Factoring df
Timer: df/dx factorization = 167 us
Computing d^2f
Timer: d^2f computation = 6.39 sec
Timer: Objective build = 1.89 ms
||dx||/||x|| =  1.792e-03
Timer: Total perturbation = 21.8 sec
```
```python
eq.perturb(objective=obj, constraints=con, deltas=deltas, tr_ratio=tr_ratio, verbose=3)
```
```
Timer: Objective build = 5.09 ms
Perturbing p_l
Factorizing linear constraints
Timer: linear constraint factorize = 604 ms
Computing df
Timer: df computation = 50.9 ms
Factoring df
Timer: df/dx factorization = 291 us
Computing d^2f
Timer: d^2f computation = 39.9 ms
Timer: Objective build = 2.01 ms
||dx||/||x|| =  2.429e-03
Timer: Total perturbation = 10.7 sec
```
For example, the first derivative takes 7.35 seconds, 8.78 seconds and 50.9 milliseconds!
Similarly, the second derivative takes 7.86 seconds, 6.39 seconds and 39.9 milliseconds.

Am I missing something? I don't think this is expected. Also, this behavior is consistent. Start a new session and run perturb 3 times, you will see this.

YigitElma:
Resolves #1581 

Some of the attributes in optimizable classes are cast to float by `float()` function which returns a Python value, but internally JAX converts them to `jnp.float64` before jit, this causes recompilation in some cases. Technically, the number of compilations shouldn't be more than 2, but this can be the cause of some recompilations.

I know there has been some discussion on not using `jnp.float64` before to leave some room for #1048, but I don't see that working in the near term.

```python
jax.config.update('jax_explain_cache_misses', True)
```
This was very useful to pinpoint the exact cause of recompilation in #1581

### DIFF
--- a/desc/optimize/least_squares.py
+++ b/desc/optimize/least_squares.py
@@ -258,6 +258,22 @@ def lsqtr(  # noqa: C901
     actual_reduction = jnp.inf
     reduction_ratio = 0
 
+    if verbose > 2:
+        print("Solver options:")
+        print("-" * 60)
+        print(f"{'Maximum Function Evaluations':<35}: {max_nfev}")
+        print(f"{'Maximum Allowed Total Δx Norm':<35}: {max_dx:.3e}")
+        print(f"{'Scaled Termination':<35}: {scaled_termination}")
+        print(f"{'Trust Region Method':<35}: {tr_method}")
+        print(f"{'Initial Trust Radius':<35}: {trust_radius:.3e}")
+        print(f"{'Maximum Trust Radius':<35}: {max_trust_radius:.3e}")
+        print(f"{'Minimum Trust Radius':<35}: {min_trust_radius:.3e}")
+        print(f"{'Trust Radius Increase Ratio':<35}: {tr_increase_ratio:.3e}")
+        print(f"{'Trust Radius Decrease Ratio':<35}: {tr_decrease_ratio:.3e}")
+        print(f"{'Trust Radius Increase Threshold':<35}: {tr_increase_threshold:.3e}")
+        print(f"{'Trust Radius Decrease Threshold':<35}: {tr_decrease_threshold:.3e}")
+        print("-" * 60, "\n")
+
     if verbose > 1:
         print_header_nonlinear()
         print_iteration_nonlinear(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Bug Fixes
 - ``desc.compat.rescale`` will now return ``ScaledProfile`` instances for most of its profiles, to fix a bug where improper scaling could occur for certain profile types.
 - Now always use ``sym=False`` in the default grid for ``plot_fsa`` to ensure correct averages
 - Fixes bug that could lead extra compilation of jit-compiled functions that include `field_line_integrate`.
+- Fixes inaccurate normalizations scales that could be computed for certain equilibria which had m=1 n=0 R and m=-1 n=0 Z components much smaller than their actual average minor radius, see [GH issue](https://github.com/PlasmaControl/DESC/issues/1954)
+
 
 Backend
 -------
--- a/desc/backend.py
+++ b/desc/backend.py
@@ -373,8 +373,7 @@ if use_jax:  # noqa: C901
                 return state[0]
 
         def tangent_solve(g, y):
-            A = jax.jacfwd(g)(y)
-            return y / A
+            return y / g(1.0)
 
         if full_output:
             x, (res, niter) = jax.lax.custom_root(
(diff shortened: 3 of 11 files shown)


desc/coils.py
```diff
@@ -274,7 +274,7 @@ class _Coil(_MagneticField, Optimizable, ABC):
     _static_attrs = _MagneticField._static_attrs + Optimizable._static_attrs
 
     def __init__(self, current, *args, **kwargs):
-        self._current = float(np.squeeze(current))
+        self._current = jnp.float64(float(np.squeeze(current)))
```
YigitElma, line 277: in some tests, we depend on the `TypeError` behavior of `float()` which `jnp.float64()` doesn't have, that is why I nested the calls.

desc/equilibrium/equilibrium.py
```diff
@@ -215,11 +215,11 @@ def __init__(
         **kwargs,
     ):
         errorif(
-            not isinstance(Psi, numbers.Real),
+            not isinstance(float(Psi), numbers.Real),
             ValueError,
             f"Psi should be a real integer or float, got {type(Psi)}",
         )
-        self._Psi = float(Psi)
+        self._Psi = jnp.float64(float(Psi))
```
f0uriest, line 222: I think the `float` here may be unecessary?
ie just use `jnp.float64(Psi)`

f0uriest: Actually I guess `float` also enforces that its a length 1 array

desc/optimize/least_squares.py
```diff
@@ -298,7 +298,7 @@ def lsqtr(  # noqa: C901
         actual_reduction = -1
 
         # theta controls step back step ratio from the bounds.
-        theta = max(0.995, 1 - g_norm)
+        theta = jnp.float64(max(0.995, 1 - g_norm))
```
f0uriest, line 317: would be cleaner to just use `jnp.maximum` here

desc/equilibrium/equilibrium.py
YigitElma: Yes, technically with the above check we make sure that psi is scalar, but I kept it there just in case

desc/optimize/least_squares.py
YigitElma: That can also return `weak_type=True` depending on `g_norm` (but I guess `g_norm` is always jax float64)
```python
g_norm = jnp.float64(0.2)
jnp.maximum(0.995, 1 - g_norm)
# Array(0.995, dtype=float64)

g_norm = 0.2
jnp.maximum(0.995, 1 - g_norm)
# Array(0.995, dtype=float64, weak_type=True)
```

PR #1982, opened 2025-10-20
Always ensure that equilibrium profiles are consistent with eq.L

### LINKED ISSUE #1974: Resolution bug in `solve_continuation_automatic`
```python
import desc
eq = desc.equilibrium.Equilibrium(L=12, M=12)
eq.pressure = desc.profiles.PowerSeriesProfile(np.array([1e4, 0, -1e4]))
eq.current = None
eq.iota = desc.profiles.PowerSeriesProfile(0.42)
solve_continuation_automatic(eq)
```

Fails due to a mismatch in array shapes when trying to add pressure:
```
Cell In[34], line 1
----> 1 eq1 = solve_continuation_automatic(eq)

File [~/Work/DESC/desc/continuation.py:586](http://localhost:8888/lab/tree/Work/DESC/local/nikita_pressure_bug/Work/DESC/desc/continuation.py#line=585), in solve_continuation_automatic(eq, objective, optimizer, pert_order, ftol, xtol, gtol, maxiter, verbose, checkpoint_path, jac_chunk_size, **kwargs)
    567     eqfam = _add_pressure(
    568         eq,
    569         eqfam,
   (...)    580         jac_chunk_size=jac_chunk_size,
    581     )
    583 # for other cases such as fixed iota or nonzero current we do pressure first
    584 # since its cheaper to do it without the 3d modes
    585 else:
--> 586     eqfam = _add_pressure(
    587         eq,
    588         eqfam,
    589         pres_step,
    590         objective,
    591         optimizer,
    592         pert_order,
    593         ftol,
    594         xtol,
    595         gtol,
    596         maxiter,
    597         verbose,
    598         checkpoint_path,
    599         jac_chunk_size=jac_chunk_size,
    600     )
    602     eqfam = _add_shaping(
    603         eq,
    604         eqfam,
   (...)    615         jac_chunk_size=jac_chunk_size,
    616     )
    617 eq.params_dict = eqfam[-1].params_dict

File [~/Work/DESC/desc/continuation.py:254](http://localhost:8888/lab/tree/Work/DESC/local/nikita_pressure_bug/Work/DESC/desc/continuation.py#line=253), in _add_pressure(eq, eqfam, pres_step, objective, optimizer, pert_order, ftol, xtol, gtol, maxiter, verbose, checkpoint_path, jac_chunk_size)
    252     if verbose > 0:
    253         print("Perturbing equilibrium")
--> 254     eqi.perturb(
    255         objective=objective_i,
    256         constraints=constraints_i,
    257         deltas=deltas,
    258         order=pert_order,
    259         verbose=verbose,
    260         copy=False,
    261     )
    262     deltas = {}
    264 stop = not eqi.is_nested()

File [~/Work/DESC/desc/equilibrium/equilibrium.py:2452](http://localhost:8888/lab/tree/Work/DESC/local/nikita_pressure_bug/Work/DESC/desc/equilibrium/equilibrium.py#line=2451), in Equilibrium.perturb(self, deltas, objective, constraints, order, tr_ratio, weight, include_f, verbose, copy)
   2449     else:
   2450         constraints = get_fixed_boundary_constraints(eq=self)
-> 2452 eq = perturb(
   2453     self,
   2454     objective=objective,
   2455     constraints=constraints,
   2456     deltas=deltas,
   2457     order=order,
   2458     tr_ratio=tr_ratio,
   2459     weight=weight,
   2460     include_f=include_f,
   2461     verbose=verbose,
   2462     copy=copy,
   2463 )
   2465 return eq

File [~/Work/DESC/desc/perturbations.py:282](http://localhost:8888/lab/tree/Work/DESC/local/nikita_pressure_bug/Work/DESC/desc/perturbations.py#line=281), in perturb(eq, objective, constraints, deltas, order, tr_ratio, weight, include_f, verbose, copy)
    268     dc = jnp.concatenate(
    269         [
    270             deltas[arg]
   (...)    273         ]
    274     )
    275     x_idx = jnp.concatenate(
    276         [
    277             eq.x_idx[arg]
   (...)    280         ]
    281     )
--> 282     tangents += jnp.eye(eq.dim_x)[:, x_idx] @ dc
    284 # 1st order
    285 if order > 0:

File [~/miniconda3/envs/desc/lib/python3.12/site-packages/jax/_src/numpy/array_methods.py:583](http://localhost:8888/lab/tree/Work/DESC/local/nikita_pressure_bug/miniconda3/envs/desc/lib/python3.12/site-packages/jax/_src/numpy/array_methods.py#line=582), in _defer_to_unrecognized_arg.<locals>.deferring_binary_op(self, other)
    581 args = (other, self) if swap else (self, other)
    582 if isinstance(other, _accepted_binop_types):
--> 583   return binary_op(*args)
    584 # Note: don't use isinstance here, because we don't want to raise for
    585 # subclasses, e.g. NamedTuple objects that may override operators.
    586 if type(other) in _rejected_binop_types:

    [... skipping hidden 15 frame]

File [~/miniconda3/envs/desc/lib/python3.12/site-packages/jax/_src/numpy/tensor_contractions.py:245](http://localhost:8888/lab/tree/Work/DESC/local/nikita_pressure_bug/miniconda3/envs/desc/lib/python3.12/site-packages/jax/_src/numpy/tensor_contractions.py#line=244), in matmul(a, b, precision, preferred_element_type)
    243 a = lax.squeeze(a, tuple(a_squeeze))
    244 b = lax.squeeze(b, tuple(b_squeeze))
--> 245 out = lax.dot_general(
    246   a, b, (((np.ndim(a) - 1,), (np.ndim(b) - 1 - b_is_mat,)), (a_batch, b_batch)),
    247   precision=precision, preferred_element_type=preferred_element_type)
    248 result = lax.transpose(out, perm)
    249 return lax_internal._convert_element_type(result, preferred_element_type, output_weak_type)

    [... skipping hidden 9 frame]

File [~/miniconda3/envs/desc/lib/python3.12/site-packages/jax/_src/lax/lax.py:5216](http://localhost:8888/lab/tree/Work/DESC/local/nikita_pressure_bug/miniconda3/envs/desc/lib/python3.12/site-packages/jax/_src/lax/lax.py#line=5215), in _dot_general_shape_rule(lhs, rhs, dimension_numbers, precision, preferred_element_type, out_sharding)
   5213 if not core.definitely_equal_shape(lhs_contracting_shape, rhs_contracting_shape):
   5214   msg = ("dot_general requires contracting dimensions to have the same "
   5215          "shape, got {} and {}.")
-> 5216   raise TypeError(msg.format(lhs_contracting_shape, rhs_contracting_shape))
   5218 return _dot_general_shape_computation(lhs.shape, rhs.shape, dimension_numbers)

TypeError: dot_general requires contracting dimensions to have the same shape, got (7,) and (2,).
```

I think it's because we're implicitly assuming that the profile resolution is always equal or higher than `eq.L`? It seems to go away if I explicitly change the profile resolution (`eq.pressure.change_resolution(eq.L)`), so maybe we should always do that if the user assigns new profiles manually?

dpanici:
This is one way to resolve the issue in #1974, by just ensuring eq profile resolution is at least or higher than the eq radial resolutions always. This also should work fine in the cases of non-power-series profiles, as those would be the same as far as number of params between different eq resolutions.

Resolves #1974

### DIFF
--- a/devtools/dev-requirements.txt
+++ b/devtools/dev-requirements.txt
@@ -29,7 +29,7 @@ pre-commit <= 4.3.0
 # testing and benchmarking
 nbmake <= 1.5.5
 pytest ~= 8.4
-pytest-benchmark <= 5.1.0
+pytest-benchmark <= 5.2.0
 pytest-cov >= 2.6.0, <= 7.0.0
 pytest-monitor <= 1.6.6
 pytest-mpl == 0.16.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,10 +11,10 @@ numpy >= 1.20.0, <= 2.2.6
 nvgpu <= 0.10.0
 orthax <= 0.2.5
 plotly >= 5.16, <= 6.3.1
-psutil <= 7.1.1
+psutil <= 7.1.3
 pylatexenc >= 2.0, <= 2.10
 quadax >= 0.2.2, <= 0.2.9
 scikit-image <= 0.25.2
 scipy >= 1.7.0, <= 1.15.3
 sympy
-termcolor <= 3.1.0
+termcolor <= 3.2.0


desc/equilibrium/equilibrium.py
```diff
@@ -377,12 +390,7 @@ def __init__(
             "anisotropy",
         ]:
             p = getattr(self, profile)
-            if hasattr(p, "change_resolution"):
-                p.change_resolution(max(p.basis.L, self.L))
-            warnif(
-                isinstance(p, PowerSeriesProfile) and p.sym != "even",
-                msg=f"{profile} profile is not an even power series.",
-            )
+            ensure_consistent_profile_eq_resolution(p, self, name=profile, warn=False)
```
dpanici: Part of me wants `warn=True` here as if the user passes in a certain profile say quadratic power series, it may be important for them to know that we've changed it to be higher resolution, in case they are trying to say only allow certain modes of it to vary during an optimiztiaon.

But I dont wanna add some complex logic to only do this when the profile passed is not None as otherwise it warns even on just using the default Equilibrium profiles (since they are initially zero but not of the right resolution as the eq)

dpanici:
I anticipate some test failures due to warnings

desc/equilibrium/utils.py
```diff
@@ -59,6 +60,46 @@ def parse_profile(prof, name="", **kwargs):
...
+    -------
+    profile : Profile
+        Profile with updated resolution if necessary.
+    """
+    if profile is None:
+        return None
+    changed_res = False
+    if hasattr(profile, "change_resolution"):
+        profile.change_resolution(max(profile.basis.L, eq.L))
+        changed_res = True
```
YigitElma: I think changed_res must be based on
```python
changed_res = (profile.basis.L < eq.L)
```
Otherwise, you will always warn (given `warn=True`), but the underlying call actually doesn't change the res.

dpanici: good point

desc/perturbations.py
```diff
@@ -76,7 +76,9 @@ def get_deltas(things1, things2):  # noqa: C901
                 t1 = t1.copy()
                 t2 = t2.copy()
                 if hasattr(t1, "change_resolution") and hasattr(t2, "basis"):
-                    t1.change_resolution(t2.basis.L)
+                    max_res = max(t1.basis.L, t2.basis.L)
```
dpanici: remove these changes

YigitElma (approved):
Approving for me and Dario

f0uriest (approved):

PR #1993, opened 2025-10-27
Make error message for AssertionError when set(objective.things) != set(things) more informative

### LINKED ISSUE #1984: Make `AssertionError` when `set(objective.things) != set(things)` more informative
This error occurs when the objectives expect `things` different than the `things` passed into the optimize call. Would be good to add a little message suggesting to check that they missed a "XXX_fixed" thing or not

This is because we tie each built objective to a given Python object instance and internally use its Python object ID to track which objective needs which things. So if you pass in a different python obejct in `things` as was used to make the objective, this assertion error will trigger.

dpanici:
Resolves #1984

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -173,7 +173,7 @@ jobs:
 
       - name: Upload benchmark data
         if: always() && env.has_changes == 'true'
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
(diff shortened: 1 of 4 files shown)


YigitElma (dismissed):

f0uriest (commented):
I think it would be good to also print the sets themselves to make clear what the mismatch is, similar to how we do in other places like `expected {set(things1)}, got {set(things2)}`

We could also consider replacing the `assert` with `errorif` to give a specific error type rather than a generic assertion error.

PR #2014, opened 2025-11-13
Allow DESC to read wout files that lack lrfp logical (like VMEC++ files)

dpanici:


desc/vmec.py
```diff
@@ -95,11 +95,13 @@ def load(
                 + " while DESC is only designed for compatibility with VMEC version"
                 + " 9. Some data may not be loaded correctly."
             )
-
-        assert float(file.variables["lrfp__logical__"][0]) == 0, (
-            "DESC currently does not support poloidal flux label, "
-            "and so cannot load this VMEC wout, which has LRFP=T"
-        )
+        if "lrfp__logical__" in file.variables:
+            # guard against the wout not having it, as VMEC++ does not save
```
YigitElma, line 99: Does VMEC++ always output without poloidal flux label?

dpanici: yes for now at least https://github.com/proximafusion/vmecpp/blob/4f41149a01e8a43affcdb83bebb82c47b2a9a8a7/tests/test_init.py#L165

ddudt (approved):

f0uriest (approved):

desc/vmec.py
jonathanschilling: Ahh yeah, sorry for the inconsistency. I hope it is not too inconvenient for you guys to have to work around that. We probably should add these wout variables to VMEC++, even if they would just be static values, as in the case of `lrfp__logical__`. Or maybe we'll even see a PR adding these some time :crossed_fingers: ?

YigitElma: Thanks for the reply @jonathanschilling ! I think the current workaround is fine (as long as a user gives an output with poloidal flux label that doesn't have this attribute).

PR #2015, opened 2025-11-13
bug fix for using `legend` kwarg in `plot_surfaces`

ddudt:
This fixes a bug in `plot_surfaces` that prevented the user from adding a legend, because it didn't account for that kwarg when checking for unexpected keyword arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ New Features
 - Adds ``desc.external.TERPSICHORE`` objective for computing linear ideal MHD stability growth rates. This objective subclasses from ``ExternalObjective`` and requires access to the TERPSICHORE code, which is not included with DESC or its dependencies.
 - Adds ``docs/dev_guids/external_objectives.rst`` as a tutorial for how to use external objectives, with TERPSICHORE as an example using parallel processing.
 - Adds keyword argument `normalize` to plot_1d, plot_3d. `normalize` is a string to use for normalization.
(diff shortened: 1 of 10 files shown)


desc/plotting.py
```diff
@@ -1778,10 +1778,11 @@ def plot_surfaces(eq, rho=8, theta=8, phi=None, ax=None, return_data=False, **kw
     axis_alpha = kwargs.pop("axis_alpha", 1)
     axis_marker = kwargs.pop("axis_marker", "o")
     axis_size = kwargs.pop("axis_size", 36)
-    label = kwargs.pop("label", "")
     title_fontsize = kwargs.pop("title_fontsize", None)
     xlabel_fontsize = kwargs.pop("xlabel_fontsize", None)
     ylabel_fontsize = kwargs.pop("ylabel_fontsize", None)
+    label = kwargs.pop("label", "")
+    legend = kwargs.pop("legend", False)
```
f0uriest: remind me why this defaults to False? shouldn't it default to True if label is given?

ddudt: That's how it was before, but I could change it. I think we want:
```
legend = kwargs.pop("legend", False if label == "" else True)
```
?

YigitElma (approved):

YigitElma (changes requested):
Inside `plot_comparison` can you manually pass in `legend=False`. With the current setup it adds 2 legends to both ax and fig

dpanici (approved):

Issue #1904, opened 2025-09-09
Make jitable transform methods for direct2 and fft


dpanici:
direct2 and fft require specific grids (equispaced) and thus have checks for that, which are not jit-compatible. 

I suppose one could make jitable-fft and jitable-direct2 where the user is trusted to have entered in the correct grids and bases, but I wanna keep this PR scope to just fixing this one bug

_Originally posted by @dpanici in https://github.com/PlasmaControl/DESC/pull/1901#discussion_r2333616658_

f0uriest:
this would require moving some of the logic from `Transform._check_inputs_{fft,direct2}` to the grid and basis classes. I think it's possible to do it statically for our existing basis classes and structured grids but not sure how easy it is.

PR #2016, opened 2025-11-15
Make a differentiable version of `FourierRZToroidalSurface.constant_offset_surface`

dpanici:
Adds differentiable version of offset surface algorithm. Potentially useful if one desires an objective that concerns a conformal vessel to a changing eq during optimization.

Also fixes what was likely a bug affecting robustness before, which was that I was using `arctan` instead of a corrected `arctan2` when computing the zeta of the offset point for the rootfind. Arctan has a range of [-pi/2, pi/2] while arctan2 will choose the correct quadrant btwn [-pi,pi] (and which can be made to go to [0,2pi] by adding 2pi to the negative part), which more closely adheres to our grid zeta domain which is always positive by convention

notebook showing example use within an objective: 
[test_differentiable_offset.ipynb](https://github.com/user-attachments/files/25052274/test_differentiable_offset.ipynb)


desc/geometry/surface.py
```diff
@@ -1146,3 +1146,114 @@ def get_axis(self):
...
+
+    Implementation of algorithm described in Appendix B of
+    "An improved current potential method for fast computation of
+    stellarator coil shapes", Landreman (2017)
+    https://iopscience.iop.org/article/10.1088/1741-4326/aa57d4
+
+    NOTE: Must have the toroidal angle as the cylindrical toroidal angle
+    in order for this algorithm to work properly
+
+    NOTE: this function lacks the checks of the constant_offset_surface
```
dpanici, line 1198: only is differentiable though if params is passed, otherwise the end will not be differentiatedcorrectlt

desc/geometry/surface.py
```diff
@@ -1146,3 +1146,114 @@ def get_axis(self):
...
+        to create the resulting constant offset surface, by default equal
+        to base_surface.N
+    full_output : bool, optional
+        If True, also return a dict of useful data about the surfaces and a
+        tuple where the first element is the residual from
+        the root finding and the second is the number of iterations.
+
+    Returns
+    -------
+    offset_surface : FourierRZToroidalSurface
```
dpanici: todo: update this (does not return surface, only the points that would then be fit)

desc/objectives/_geometry.py
```diff
@@ -452,6 +453,192 @@ def compute(self, params, constants=None):
...
+                offset_zetas,
+            ]
+        ).T
+        # make transform to fit
+        t = get_transforms(
+            obj=surf,
+            keys=["R", "Z"],
+            grid=Grid(offset_rtz_nodes, jitable=True),
+            jitable=True,
+            build_pinv=True,
```
dpanici: dont know if this build_pinv does anything

desc/objectives/_geometry.py
```diff
@@ -452,6 +453,192 @@ def compute(self, params, constants=None):
         return data["V"]
 
 
+class VolumeOffset(_Objective):
```
dpanici: here just to test, will remove for final version

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,151 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+    info : tuple
+        2 element tuple containing residuals and number of iterations
+        for each point.
+
+    """
+    if params is None:
+        params = base_surface.params_dict
+
+    def n_and_r_jax(nodes):
+        data = base_surface.compute(
```
YigitElma, line 1250: Are we sure this is differentiable?

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,151 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+    M : int, optional
+        Poloidal resolution of the basis used to fit the offset points
+        to create the resulting constant offset surface, by default equal
+        to base_surface.M. If basis is given, this is ignored.
+    N : int, optional
+        Toroidal resolution of the basis used to fit the offset points
+        to create the resulting constant offset surface, by default equal
+        to base_surface.N.  If basis is given, this is ignored.
+    R_basis, Z_basis: DoubleFourierSeries, optional
+        Basis to use to fit the offset surface's R and Z, respectivelu. If None,
```
YigitElma:
```suggestion
        Basis to use to fit the offset surface's R and Z, respectively. If None,
```

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,151 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+        new bases will be created using the given M and N.
+    params : dict, optional
+        dictionary of parameters to use when computing data from the base_surface.
+        If None, uses base_surface.params_dict, however the resulting computation
+        will not be differentiable with respect to the base_surface parameters
+        (since the JAX AD inside of an objective traces the params dictionaries
+        that are passedto their compute methods)
```
YigitElma:
```suggestion
        that are passed to their compute methods)
```

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,151 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+    zetas = jnp.asarray(zetas)
+    nodes = jnp.vstack((jnp.ones_like(grid.nodes[:, 1]), grid.nodes[:, 1], zetas)).T
+    n, x, x_offsets = n_and_r_jax(nodes)
+
+    data = {}
+    data["n"] = xyz2rpz_vec(n, phi=nodes[:, 1])
+    data["x"] = xyz2rpz(x)
+    data["x_offset_surface"] = xyz2rpz(x_offsets)
+
+    offset_zetas = data["x_offset_surface"][:, 1]
```
YigitElma: I don't think I follow the code. I know this was added in a previous PR but I wanted to ask to see if this code is correct or not. So, the above rootfind basically tries to find some zeta where if you start from (theta, zeta) and go in the normal direction by offset amount, your zeta is equal to the original grid's zeta. In short, find some points on the original surface with uniform spacing in theta but non-uniform spacing in zeta that give points in offset surface that are uniformly spaced in zeta (and assign a different theta definition that is uniform).

Why do you find the offset zetas here? They should be already equal to `grid.nodes[:, 2]`?

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,151 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+        Z_basis = DoubleFourierSeries(
+            M=M, N=N, NFP=base_surface.NFP, sym=base_surface.Z_basis.sym
+        )
+
+    zetas = jnp.asarray(zetas)
+    nodes = jnp.vstack((jnp.ones_like(grid.nodes[:, 1]), grid.nodes[:, 1], zetas)).T
+    n, x, x_offsets = n_and_r_jax(nodes)
+
+    data = {}
+    data["n"] = xyz2rpz_vec(n, phi=nodes[:, 1])
```
YigitElma: `nodes[:, 1]` is theta, isn't it?

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,151 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+        base_surface.M and base_surface.N and base_surface.NFP
+    M : int, optional
+        Poloidal resolution of the basis used to fit the offset points
+        to create the resulting constant offset surface, by default equal
+        to base_surface.M. If basis is given, this is ignored.
+    N : int, optional
+        Toroidal resolution of the basis used to fit the offset points
+        to create the resulting constant offset surface, by default equal
+        to base_surface.N.  If basis is given, this is ignored.
+    R_basis, Z_basis: DoubleFourierSeries, optional
```
YigitElma: Giving the basis doesn't seem efficient, why not give the transforms directly? Instead of rebuilding them everytime, they will be reused, and they store the information on basis too if you need to access that.

YigitElma (changes requested):
Need to add test to make sure this is not compiled again. I had a similar check for field line intergration in `test_field_line_integrate_jax_transforms`

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,151 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+    nodes = jnp.vstack((jnp.ones_like(grid.nodes[:, 1]), grid.nodes[:, 1], zetas)).T
+    n, x, x_offsets = n_and_r_jax(nodes)
+
+    data = {}
+    data["n"] = xyz2rpz_vec(n, phi=nodes[:, 1])
+    data["x"] = xyz2rpz(x)
+    data["x_offset_surface"] = xyz2rpz(x_offsets)
+
+    offset_zetas = data["x_offset_surface"][:, 1]
+    offset_rtz_nodes = jnp.vstack(
```
YigitElma: I guess since `offset_zeta==grid.nodes[:,2]` and you choose some new theta definition for the offset surface that matches `grid.nodes[:,1]`, you can just use the original `grid` here. And since that will stay constant, you can pass the pre-built transforms directly

desc/geometry/surface.py
```diff
@@ -717,6 +721,8 @@ def constant_offset_surface(
                     coordinates on the offset surface, corresponding to the
                     ``x`` points on the base_surface (i.e. the points to which the
                     offset surface was fit)
+            as well as the DoubleFourierSeries bases used to fit R and Z.
+            Only returned if ``full_output`` is True
```
YigitElma: Docs of the grid can also be updated. The default grid resolution is twice the M,N

desc/geometry/surface.py
dpanici: They *should* be equal, assuming the rootfind succeeded. If the rootfind did not succeed they may be different. The rootfind should succeed in most cases tho so I made the change to use the grid.nodes[:,2]

Something weird I found is in some cases, the "fft" transform method is more fragile when doing the fit then the "direct1" method. Not sure why

desc/geometry/surface.py
dpanici: true, fixed

tests/test_surfaces.py
```diff
@@ -212,6 +215,116 @@ def test_constant_offset_surface_circle(self):
...
+            N=s.N,
+            NFP=s.NFP,
+            sym=s.sym,
+            modes_R=data["transforms"]["R"].basis.modes[:, 1:],
+            modes_Z=data["transforms"]["Z"].basis.modes[:, 1:],
+        )
+
+        r_offset_surf = data["x_offset_surface"]
+        r_surf = data["x"]
+        dists = np.linalg.norm(r_surf - r_offset_surf, axis=1)
```
YigitElma: just to note, if the surface `s` wasn't an axisymmetric surface, then this wouldn't be true because the rootfind would find a different zeta.

tests/test_surfaces.py
```diff
@@ -212,6 +215,116 @@ def test_constant_offset_surface_circle(self):
...
+                    np.zeros_like(grid_compute.nodes[:, 1]),
+                )
+            ).T,
+        }
+        for key in ["e_theta", "e_zeta"]:
+            np.testing.assert_allclose(
+                correct_data_offset[key],
+                data_offset[key],
+                atol=1e-4,
+                err_msg=f"Failed test at comparison of {key}",
```
YigitElma: Lines 234-302 seem to be duplicate of `test_constant_offset_surface_circle`, you may just keep the `jit` and `grad` stuff.

YigitElma (commented):
Seems fine to me for now. I will test it later myself

tests/test_surfaces.py
dpanici: Yea I know

desc/geometry/surface.py
```diff
@@ -717,6 +722,8 @@ def constant_offset_surface(
                     coordinates on the offset surface, corresponding to the
                     ``x`` points on the base_surface (i.e. the points to which the
                     offset surface was fit)
+            as well as the transforms bases used to fit R and Z.
```
YigitElma: Building the docs seems to fail because of this line. Maybe just add `transforms` as the other ones?

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,136 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+        n, r, r_offset = n_and_r_jax(nodes)
+        # add 2pi to the arctan2<0 so it matches our convention of
+        # zeta being btwn 0 and 2pi
+        zeta_offset = jnp.arctan2(r_offset[0, 1], r_offset[0, 0])
+        return jnp.where(zeta_offset < 0, zeta_offset + 2 * np.pi, zeta_offset) - zeta
+
+    vecroot = jit(
+        vmap(
+            lambda x0, *p: root_scalar(
+                fun_jax, x0, jac=None, args=p, full_output=True, tol=1e-12
```
YigitElma: We might need `maxiter=100` or some higher number here. The default is 20 and it might not be enough for `tol=1e-12`

tests/test_surfaces.py
```diff
@@ -212,6 +215,45 @@ def test_constant_offset_surface_circle(self):
...
+        out = capsys.readouterr()
+        assert out.out == ""
+
+        # check gradient is correct
+        # R00 of offset should change with R00 of original surface
+        grad_R00 = jax.grad(lambda params: fun(params)[0][s.R_basis.get_idx(M=0, N=0)])(
+            s.params_dict
+        )
+        # check that the gradient is nonzero for the R00 component
+        assert np.any(np.abs(grad_R00["R_lmn"][s.R_basis.get_idx(M=0, N=0)]) > 1e-10)
```
YigitElma: Is this equal to 1? I think if the surface is axisymetric, it should be 1?

YigitElma (dismissed):

tests/test_surfaces.py
dpanici: yea I check it later, kind of redundant this one is

YigitElma (dismissed):

desc/geometry/surface.py
```diff
@@ -738,57 +745,23 @@ def constant_offset_surface(
...
-        vecroot = jit(
-            vmap(
-                lambda x0, *p: root_scalar(
-                    fun_jax, x0, jac=None, args=p, full_output=full_output
-                )
-            )
+        R_lmn, Z_lmn, data, (res, niter) = _constant_offset_surface(
+            base_surface,
+            offset,
+            grid=grid,
```
ddudt: could fit this all on fewer lines if you remove the trailing comma :)

desc/geometry/surface.py
```diff
@@ -1205,3 +1178,136 @@ def _get_ess_scale(self, alpha=1.2, order=np.inf, min_value=1e-7):
...
+        # NOTE: we are assuming here that the rootfind was successful for every point,
+        # so that the zeta=arctan(y/x) of the offset surface point are the same as
+        # the grid nodes' zeta values. If this is not the case, the fitting
+        # will be incorrect.
+        transforms = get_transforms(
+            obj=base_surface,
+            keys=["R", "Z"],
+            grid=grid,
+            # this is more robust than letting method become fft if
+            # jitable=False, and will also work within jitted functions
```
dpanici, line 1300: Once #1904 is in we can allow this to be fft, though I did notice robustness issues that I did not fully understand when I used fft instead of direct here.

tests/test_surfaces.py
ddudt: I vote to remove this line since it is a redundant check

ddudt (dismissed):

YigitElma (dismissed):

YigitElma (dismissed):

YigitElma (approved):

PR #2020, opened 2025-11-23
Adds wrapper for scipy l-bfgs-b method, also does not call hessian if using a scipy-wrapped method which approximates hessian

dpanici:
- Adds wrapper for scipy l-bfgs-b
- No longer computes hessian for setting `x_scale` if `x_scale="auto"` if the method is a hessian-free method (previously we were doing this, so say you were calling `scipy-bfgs` bc you did not want to compute the hessian, the default behavior was still trying to compute the hessian to provide the x_scale)
  - We already had checks for this in the desc methods, only the scipy-wrapped optimization methods had this issue

### DIFF
--- a/devtools/dev-requirements.txt
+++ b/devtools/dev-requirements.txt
@@ -9,7 +9,7 @@
 # which will need to be updated regularly, but we don't want to do so without testing.
 
 # building the docs
-nbsphinx >= 0.8.12, <= 0.9.7
+nbsphinx >= 0.8.12, <= 0.9.8
 sphinx >= 5.0, <= 8.2.3
 sphinx-argparse >= 0.4.0, != 0.5.0, <= 0.5.2
 sphinx_copybutton <= 0.5.2


docs/notebooks/tutorials/coil_stage_two_optimization.ipynb
YigitElma, line 1: I assume this is some metadata change?

dpanici: oh oops yea I did not intend to hcange anythinghere

YigitElma (approved):

desc/optimize/_scipy_wrappers.py
```diff
@@ -30,6 +31,8 @@
         "scipy-trust-krylov",
     ],
     description=[
+        "L-BFGS-B quasi-newton method with line search. "
```
f0uriest: I think this is in the wrong order wrt the list of names above

PR #2025, opened 2025-11-24
Add Paraview functions to `desc.external`

### LINKED ISSUE #1299: Exporting coils, surfaces
Previously I had used paraview to better visualize coils and the plasma boundary. Is there an equivalent software that is recommended for .h5 files? Or is there a way to save coils/surfaces to vtk format?

YigitElma:
Adds utility functions to be able to export coils, equilibrium volume and surfaces to Paraview. Uses an optional dependency `pyvista`. I had these scripts for a while, these are the cleaned-up versions. It is possible to do very cool stuff with volume data in Paraview.

These functions won't be tested on CI as other `desc.external` tools.

example-image

Resolves #1299

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,17 @@ New Features
     - Particle trajectories can be plotted with ``desc.plotting.plot_particle_trajectories`` function.
 - Adds new option for ``loss_function``, ``"sum"``, which changes an objective to target the sum of the values computed.
 - Adds utility functions ``desc.external.export_surface_to_paraview``, ``desc.external.export_volume_to_paraview`` and ``desc.external.export_coils_to_paraview`` to export Paraview files for surfaces, volume and coils. These functions use an optional dependency ``pyvista`` which is not automatically installed.
+- The `x_scale` option for `eq.optimize` and related functions can now be given as a dictionary mapping individual parameter names to their relevant scales,
+or if multiple things are being optimized, `x_scale` can be a list of dict, one for each optimizable thing.
 - Adds ``"scipy-l-bfgs-b"`` optimizer option as a wrapper to scipy's ``"l-bfgs-b"`` method.
 
 Bug Fixes
 
 - No longer uses the full Hessian to compute the scale when ``x_scale="auto"`` and using a scipy optimizer that approximates the hessian (e.g. if using ``"scipy-bfgs"``, no longer attempts the Hessian computation to get the x_scale)
+
+Performance Improvements
+
+- `ProximalProjection.grad` uses a single VJP on the objective instead of multiple JVP followed by a manual VJP. This should be more efficient for expensive objectives.
 
 v0.16.0
 -------
--- a/desc/equilibrium/equilibrium.py
+++ b/desc/equilibrium/equilibrium.py
@@ -2242,7 +2242,7 @@ class Equilibrium(IOAble, Optimizable):
             stopping tolerances. `None` will use defaults for given optimizer.
         maxiter : int
             Maximum number of solver steps.
-        x_scale : array_like or ``'auto'``, optional
+        x_scale : array, list[dict] or ``'auto'``, optional
             Characteristic scale of each variable. Setting ``x_scale`` is equivalent
             to reformulating the problem in scaled variables ``xs = x / x_scale``.
             An alternative view is that the size of a trust region along jth
@@ -2250,7 +2250,10 @@ class Equilibrium(IOAble, Optimizable):
             be achieved by setting ``x_scale`` such that a step of a given size
             along any of the scaled variables has a similar effect on the cost
             function. If set to ``'auto'``, the scale is iteratively updated using the
-            inverse norms of the columns of the Jacobian or Hessian matrix.
+            inverse norms of the columns of the Jacobian or Hessian matrix. If an array,
+            should be the same size as sum(thing.dim_x for thing in things). If a list
+            of dict, the list should have 1 element for each thing, and each dict should
+            have the same keys and dimensions as thing.params_dict.
         options : dict
             Dictionary of additional options to pass to optimizer.
         verbose : int
@@ -2348,7 +2351,7 @@ class Equilibrium(IOAble, Optimizable):
             stopping tolerances. `None` will use defaults for given optimizer.
         maxiter : int
             Maximum number of solver steps.
-        x_scale : array_like or ``'auto'``, optional
+        x_scale : array, list[dict] or ``'auto'``, optional
             Characteristic scale of each variable. Setting ``x_scale`` is equivalent
             to reformulating the problem in scaled variables ``xs = x / x_scale``.
             An alternative view is that the size of a trust region along jth
@@ -2356,7 +2359,10 @@ class Equilibrium(IOAble, Optimizable):
             be achieved by setting ``x_scale`` such that a step of a given size
             along any of the scaled variables has a similar effect on the cost
             function. If set to ``'auto'``, the scale is iteratively updated using the
-            inverse norms of the columns of the Jacobian or Hessian matrix.
+            inverse norms of the columns of the Jacobian or Hessian matrix. If an array,
+            should be the same size as sum(thing.dim_x for thing in things). If a list
+            of dict, the list should have 1 element for each thing, and each dict should
+            have the same keys and dimensions as thing.params_dict.
         options : dict
             Dictionary of additional options to pass to optimizer.
         verbose : int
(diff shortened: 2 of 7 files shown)


desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+        The flux surface to be exported if `obj` is an Equilibrium. Defaults to 1.
+    filename : str
+        Name for the saved file. The file extension will be `.vtu`. Default name will be
+        `surface.vtu`
+
+    Returns
+    -------
+    mesh : pyvista.UnstructuredGrid
+        Created unstructured grid object. With this object one can compute more
+        complicated quantities on `LinearGrid(rho=rho, theta=Np, zeta=Nt, NFP=1)`
```
dpanici: either use `N_toroidal` and `N_poloidal` here or change the above mentions to `Np` and `Nt`

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+
+def export_coils_to_paraview(coils, res=100, keys=[], filename="coil"):
+    """Export coils as VTP files.
+
+    Parameters
+    ----------
+    coils : Coil, CoilSet, MixedCoilSet
+        Object to be exported. Can be a single coil or multiple coils given
+        as CoilSet or MixedCoilSet.
+    res : int
```
dpanici, line 99: be clear that this is e.g. the resolution used to discretize each individual coil (and that the same res is used for every coil)

Also this res is the N for LinearGrid so this is not clear that the number of points used will be 2*res+1

either change docstring or just allow grid to be passed in with a default value set by N=100 if grid is None

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def export_volume_to_paraview(eq, res=(20, 100, 100), keys=["B"], filename="volume"):
+    """Export equilibrium volume as VTU files.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium to be exported.
+    res : tuple of ints, (Nr, Np, Nt)
```
dpanici, line 172: same res comment, explain these more (number of points in each direction)

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+    """Export a constant rho surface as VTU file.
+
+    Parameters
+    ----------
+    obj : Equilibrium or FourierRZToroidalSurface
+        Object to be exported. Different flux surfaces of the Equilibrium
+        can be exported by supplying `rho`.
+    res : tuple, list
+        The resolution used to create the mesh. Must be in the form
+        (N_toroidal, N_poloidal). Default is (100, 100).
```
f0uriest: I'd use `M_poloidal` here, and maybe switch the order to make it match other places in the code

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+    keys : list, optional
+        The names of the quantities to be computed on the grid points. Defaults to
+        empty list.
+    rho : float
+        The flux surface to be exported if `obj` is an Equilibrium. Defaults to 1.
+    filename : str
+        Name for the saved file. The file extension will be `.vtu`. Default name will be
+        `surface.vtu`
```
f0uriest: I think in most other places we don't add the file extension automatically, but expect the user to name it how they want. (ie, `eq.save("foo.h5")` rather than `eq.save("foo")`)

I'd also be in favor of not giving this a default, it feels like it might lead to accidentally overwriting things.

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+        and add it to the mesh by `mesh['complicated name'] = value`. Once the mesh
+        is changed, the user has to save it again `mesh.save(filename + ".vtu")`.
+    """
+    if not isinstance(obj, (Equilibrium, FourierRZToroidalSurface)):
+        raise ValueError(
+            "This function only support Equilibrium or FourierRZToroidalSurface "
+            f"objects but {type(obj)} is given."
+        )
+
+    Nt, Np = res
```
f0uriest: do we want to multiply Nt by NFP so that Nt is the number of points per period? I think we do that in our other 3d plotting?

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+            # Append one quad: format = [4, pt0, pt1, pt2, pt3]
+            cells.extend([4, p0, p1, p2, p3])
+            celltypes.append(CellType.QUAD)
+
+    # Convert to numpy arrays
+    cells = np.array(cells)
+    celltypes = np.array(celltypes, dtype=np.uint8)
+
+    # Create the unstructured mesh
+    mesh = pv.UnstructuredGrid(cells, celltypes, nodes)
```
f0uriest: isn't a LinearGrid a structured mesh? is it possible to take advantage of that?

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+
+    # Additional data
+    for key in keys:
+        value = data[key]
+        if len(value) == Nt * Np or value.shape == (Nt * Np, 3):
```
f0uriest: FYI `len(value) == Nt*Np` would also match things with shape(Nt*Nz, 10) for example bc the leading dimension is Nt*Nz

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+            # this will handle both 1D and 3D
+            mesh[key] = value
+        else:
+            raise ValueError(f"Data {key} has an unimplemented shape ({value.shape})")
+
+    mesh.save(filename + ".vtu")
+    print(f"File is saved as {filename}.vtu")
+    return mesh
```
f0uriest: Is this function supposed to save the mesh or return it? I'd prefer if it was just one or the other, not both. (it seems like returning it is the easier option?)

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+    coils : Coil, CoilSet, MixedCoilSet
+        Object to be exported. Can be a single coil or multiple coils given
+        as CoilSet or MixedCoilSet.
+    res : int
+        The resolution to use. Defaults to 100.
+    keys : list of str
+        Additional data to store in the file. By default, only the current
+        corresponding to each coil is stored.
+    filename : str
+        Name for the saved file. The file extension will be `.vtp`. If there
```
f0uriest: is there a difference between `vtu` and `vtp`?

desc/external/paraview.py
```diff
@@ -0,0 +1,260 @@
...
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium to be exported.
+    res : tuple of ints, (Nr, Np, Nt)
+        The resolutions to use. Defaults to (20, 100, 100) in radial, poloidal
+        and toroidal directions, respectively.
+    keys : list of str
+        Additional data to store in the file. By default, only has the magnetic field
+        B components.
```
f0uriest: The other ones don't include any other data by default, why include B here?

docs/api.rst
```diff
@@ -96,6 +96,21 @@ Examples
     desc.examples.get
     desc.examples.listall
 
+External
```
f0uriest, line 99: Not sure if the stuff in external is meant to be publicly documented?

desc/external/paraview.py
YigitElma: `plot_3d` etc takes in a grid, if none given, yes, we use NFP times some default res, I guess this is easier to explain since I create the grid internally here. But I am indifferent

desc/external/paraview.py
YigitElma: I think 99% of the users will not use the mesh (so if we will do one thing, saving is probably more convenient), but in certain cases it might be easier to have the mesh object to store different types of data which doesn't have compute function directly. I can add a flag like `return_mesh=False` but omitted for now. 

Btw, one can combine different data in Paraview with custom equations but Python is probably a lot easier.

docs/api.rst
YigitElma: I am in favor of documenting them with enough warnings. Otherwise, these are very hidden functionalities.

desc/external/paraview.py
YigitElma: `pyvista.UnstructuredGrid` object cannot be saved as `vtp` and `pyvista.PolyData` object cannot be saved as `vtu`. The options are limited so I chose a default for them.

desc/external/paraview.py
YigitElma: The file types have limited options that users might not be familiar with. If you give a wrong file extension pyvista throws an error but I feel like adding the default file type is more convenient.

desc/external/paraview.py
YigitElma: I changed to structured mesh. The file size is slightly smaller and the code is cleaner. There are couple features in Paraview which were not available for unstructured grids (extract certain cells etc), now it should be ready.

f0uriest (approved):

PR #2028, opened 2025-12-03
Update NAE tests

YigitElma:
NAE tests, `test_NAE_QSC_solve_near_axis_asym` and `test_NAE_QSC_solve_near_axis_based_off_eq_asym`, started to fail randomly on unrelated PRs (probably due to different CI processors). This updates the tolerances and resolution for the tests.

Also, `test_NAE_QSC_solve_near_axis_asym` takes around 35 minutes on CI, hopefully, with these changes it will take around 20 mins. The most time consuming part is solving the 2nd order constraint equilibrium, it takes at least 50 iterations to get to the new tolerances (proiblem size is pretty big for CPU,  9546 x 2276)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Changelog
 New Features
 
 - Adds new option for ``loss_function``, ``"sum"``, which changes an objective to target the sum of the values computed.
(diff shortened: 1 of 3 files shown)


YigitElma:
`test_NAE_QSC_solve_near_axis_asym` still takes 1315 seconds which is better than before but still very long for a single test.

PR #2029, opened 2025-12-03
Add logic for handling pytree x_scale

f0uriest:
This should make #1736 a bit easier, as well as making it easier to do other custom scalings for different objects/variables

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ Bug Fixes
 - No longer uses the full Hessian to compute the scale when ``x_scale="auto"`` and using a scipy optimizer that approximates the hessian (e.g. if using ``"scipy-bfgs"``, no longer attempts the Hessian computation to get the x_scale)
 
 
+Performance Improvements
+
+- `ProximalProjection.grad` uses a single VJP on the objective instead of multiple JVP followed by a manual VJP. This should be more efficient for expensive objectives.
+
 v0.16.0
 -------
 
--- a/desc/optimize/_constraint_wrappers.py
+++ b/desc/optimize/_constraint_wrappers.py
@@ -1055,10 +1055,27 @@ class ProximalProjection(ObjectiveFunction):
             gradient vector.
 
         """
-        # TODO (#1393): figure out projected vjp to make this better
-        f = jnp.atleast_1d(self.compute_scaled_error(x, constants))
-        J = self.jac_scaled_error(x, constants)
-        return f.T @ J
+        # We are looking for the gradient of L = 0.5 * G.T @ G
+        # Then, the gradient is ∇L = G.T @ J_of_G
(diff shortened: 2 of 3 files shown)


desc/optimize/optimizer.py
```diff
@@ -341,6 +347,34 @@ def optimize(  # noqa: C901
         return things, result
 
 
+def _parse_x_scale(x_scale, things, options):
```
YigitElma, line 366: Is the `options` used at all? Or do you have plans for the `ess` PR?

desc/optimize/optimizer.py
```diff
@@ -341,6 +347,34 @@ def optimize(  # noqa: C901
...
+        dimx_all = sum([t.dim_x for t in things])
+        return jnp.broadcast_to(x_scale, (dimx_all,))
+    elif len(things) == 1 and not isinstance(x_scale, (list, tuple)):
+        x_scale = [x_scale]
+    assert len(x_scale) == len(
+        things
+    ), f"expected {len(things)} x_scales for {len(things)} things but "
+    f"only got {len(x_scale)}"
+    all_scales = []
+    for xsc, tng in zip(x_scale, things):
```
YigitElma, line 388: In `optimizer.optimize`, we have a convoluted logic to sort `things` for the state vector `x`. Do we need to consider that here or things work out fine without it?

desc/optimize/optimizer.py
```diff
@@ -341,6 +347,34 @@ def optimize(  # noqa: C901
...
+        if isinstance(xsc, (jnp.ndarray, np.ndarray)) or np.isscalar(xsc):
+            all_scales.append(jnp.broadcast_to(xsc, (tng.dim_x,)))
+        elif isinstance(xsc, dict) or (
+            isinstance(xsc, list) and isinstance(tng, OptimizableCollection)
+        ):
+            all_scales.append(tng.pack_params(xsc))
+        else:
+            raise TypeError(
+                f"all x_scales should be either array or dict, got {type(xsc)}"
```
YigitElma, line 397: In `OptimizableCollection` case, it can be list of dicts too, right?

desc/optimize/optimizer.py
```diff
@@ -341,6 +347,34 @@ def optimize(  # noqa: C901
...
+            )
+    return np.concatenate(all_scales)
```
YigitElma: Why not form `jnp` array directly

YigitElma (commented):
Either in this one or in `ess` PR, can we also add the `x_scale` to the auxilary data returned with optimization? It would help testing `ess` a lot, without doing optimization.

desc/optimize/optimizer.py
f0uriest: its used in ess

YigitElma (dismissed):

YigitElma (approved):

dpanici (approved):

PR #2030, opened 2025-12-04
Use VJP for Proximal's `grad`

### LINKED ISSUE #1393: Implement `vjp` for `ProximalProjection` instead of using forward mode Jacobian
https://github.com/PlasmaControl/DESC/blob/95612fa625bd5ba939569004c767ed02594586f0/desc/optimize/_constraint_wrappers.py#L875

YigitElma:
Resolves #1393


desc/optimize/_constraint_wrappers.py
```diff
@@ -1055,10 +1055,27 @@ def grad(self, x, constants=None):
...
+        constants = setdefault(constants, self.constants)
+        xg, xf = self._update_equilibrium(x, store=True)
+        jvpfun = lambda u: self._get_tangent(u, xf, constants, op="scaled_error")
+        tangents = batched_vectorize(
+            jvpfun,
+            signature="(n)->(k)",
+            chunk_size=self._constraint._jac_chunk_size,
+        )(v)
+        g = jnp.atleast_1d(self.compute_scaled_error(x, constants))
+        g_vjp = self._objective.vjp_scaled_error(g, xg, constants[0])
```
YigitElma, line 1077: This might not be the most optimal but if the objective is very expensive, it reduces the N JVPs to a single VJP and a product. If anyone has any idea on how to use VJPs for the `tangents` part, I appreciate any help.

f0uriest: I don't think we can really get around computing the full tangents, since it needs the inverse of `Fx`

f0uriest (approved):

f0uriest (approved):

desc/optimize/_constraint_wrappers.py
```diff
@@ -1055,10 +1055,27 @@ def grad(self, x, constants=None):
             gradient vector.
 
         """
-        # TODO (#1393): figure out projected vjp to make this better
-        f = jnp.atleast_1d(self.compute_scaled_error(x, constants))
-        J = self.jac_scaled_error(x, constants)
-        return f.T @ J
+        # We are looking for the gradient of L = 0.5 * G.T @ G
+        # Then, the gradient is ∇L = G.T @ J_of_G
```
unalmis, line 1059: Why are you assuming symmetry?

YigitElma: ∇L is a vector, for convenience, writing it in this form looks better suited to describe the vjp (G.T @ ∇G). The final column vector can be obtained by tranposing the thing. Probably, ∇L.T = G.T @ J_of_G is better notation but I think it is clear.

Regarding general array dimensions and definitions here:
tangents = [dc_tangents - (∇F @ dx_tangents) ^ -1 @ (∇F @ dc_tangents)].T
tangents.shape = (dim_c, dim_full)

g_vjp = ( G.T @ ∇G ).T
g_vjp.shape = (dim_full, )

So, the final thing is actually computing [dc_tangents - (∇F @ dx_tangents) ^ -1 @ (∇F @ dc_tangents)].T @ ( G.T @ ∇G ).T which is the transpose.

I didn't care much about the transposes etc in the math because our functions don't return the same shape as the math suggests. For example, you would expect to get a row vector from vjp but our vjp functions return a column vector (at least it is treated as column vector due to missing dimension).

PR #2031, opened 2025-12-05
Add option to choose Proximal inversion method

### LINKED ISSUE #1569: Use QR in Proximal
I want to believe there is a way to do this but may need some additional changes to some tests. Benchmarks for this,
```diff
|             benchmark_name             |         dt(%)          |         dt(s)          |        t_new(s)        |        t_old(s)        | 
| -------------------------------------- | ---------------------- | ---------------------- | ---------------------- | ---------------------- |
 test_build_transform_fft_lowres         |     +0.15 +/- 9.10     | +8.40e-04 +/- 4.95e-02 |  5.45e-01 +/- 3.3e-02  |  5.45e-01 +/- 3.7e-02  |
 test_equilibrium_init_medres            |     -2.49 +/- 3.23     | -1.16e-01 +/- 1.51e-01 |  4.54e+00 +/- 5.8e-02  |  4.66e+00 +/- 1.4e-01  |
 test_equilibrium_init_highres           |     -0.12 +/- 1.95     | -7.13e-03 +/- 1.15e-01 |  5.87e+00 +/- 8.5e-02  |  5.87e+00 +/- 7.7e-02  |
 test_objective_compile_dshape_current   |     -0.88 +/- 3.60     | -3.70e-02 +/- 1.51e-01 |  4.17e+00 +/- 1.2e-01  |  4.21e+00 +/- 9.1e-02  |
 test_objective_compute_dshape_current   |     +3.03 +/- 4.06     | +1.57e-04 +/- 2.10e-04 |  5.33e-03 +/- 2.0e-04  |  5.17e-03 +/- 6.3e-05  |
 test_objective_jac_dshape_current       |     +1.88 +/- 7.02     | +8.16e-04 +/- 3.05e-03 |  4.43e-02 +/- 1.7e-03  |  4.35e-02 +/- 2.5e-03  |
 test_perturb_2                          |     -0.86 +/- 1.87     | -1.74e-01 +/- 3.79e-01 |  2.01e+01 +/- 3.0e-01  |  2.02e+01 +/- 2.3e-01  |
+test_proximal_jac_atf_with_eq_update    |    -11.98 +/- 1.23     | -5.31e+00 +/- 5.46e-01 |  3.90e+01 +/- 3.9e-01  |  4.44e+01 +/- 3.8e-01  |
 test_proximal_freeb_jac                 |     -1.28 +/- 1.26     | -9.58e-02 +/- 9.47e-02 |  7.40e+00 +/- 4.2e-02  |  7.50e+00 +/- 8.5e-02  |
 test_solve_fixed_iter_compiled          |     +0.53 +/- 2.79     | +1.10e-01 +/- 5.85e-01 |  2.11e+01 +/- 5.1e-01  |  2.10e+01 +/- 2.9e-01  |
 test_LinearConstraintProjection_build   |     +1.59 +/- 2.14     | +1.73e-01 +/- 2.32e-01 |  1.11e+01 +/- 2.3e-01  |  1.09e+01 +/- 3.6e-02  |
 test_build_transform_fft_midres         |     +0.31 +/- 2.92     | +1.84e-03 +/- 1.75e-02 |  6.02e-01 +/- 1.5e-02  |  6.00e-01 +/- 9.8e-03  |
 test_build_transform_fft_highres        |     -0.33 +/- 2.30     | -3.20e-03 +/- 2.20e-02 |  9.52e-01 +/- 2.0e-02  |  9.56e-01 +/- 8.1e-03  |
 test_equilibrium_init_lowres            |     +0.12 +/- 0.92     | +4.87e-03 +/- 3.73e-02 |  4.07e+00 +/- 3.2e-02  |  4.06e+00 +/- 2.0e-02  |
 test_objective_compile_atf              |     -0.16 +/- 1.34     | -1.30e-02 +/- 1.11e-01 |  8.27e+00 +/- 8.4e-02  |  8.28e+00 +/- 7.3e-02  |
 test_objective_compute_atf              |     -0.19 +/- 1.37     | -3.01e-05 +/- 2.15e-04 |  1.57e-02 +/- 1.5e-04  |  1.58e-02 +/- 1.5e-04  |
 test_objective_jac_atf                  |     +1.68 +/- 2.13     | +3.27e-02 +/- 4.14e-02 |  1.98e+00 +/- 3.2e-02  |  1.95e+00 +/- 2.7e-02  |
 test_perturb_1                          |     +0.10 +/- 1.02     | +1.41e-02 +/- 1.48e-01 |  1.45e+01 +/- 1.3e-01  |  1.45e+01 +/- 7.6e-02  |
+test_proximal_jac_atf                   |    -32.47 +/- 0.61     | -2.69e+00 +/- 5.09e-02 |  5.59e+00 +/- 3.1e-02  |  8.28e+00 +/- 4.1e-02  |
+test_proximal_freeb_compute             |    -46.95 +/- 0.56     | -9.54e-02 +/- 1.13e-03 |  1.08e-01 +/- 6.5e-04  |  2.03e-01 +/- 9.2e-04  |
 test_solve_fixed_iter                   |     +0.34 +/- 1.77     | +1.07e-01 +/- 5.65e-01 |  3.20e+01 +/- 2.9e-01  |  3.19e+01 +/- 4.9e-01  |
```

_Originally posted by @YigitElma in https://github.com/PlasmaControl/DESC/pull/1409#discussion_r1942020613_

### LINKED ISSUE #2000: Necessity of choice of regularization in proximal jvp step
In `ProximalProjection` we use a regularization in this line

https://github.com/PlasmaControl/DESC/blob/548a83b30bf6ec319539a896fa6c5a560e052911/desc/optimize/_constraint_wrappers.py#L1314

This specifically is the part that is doing the `(dF/dx^{-1}) @ dF/dc` in the proximal projection (i.e. in the term on LHS of eq 2.9 in [dudt QS paper](https://arxiv.org/pdf/2204.00078) )

Image

while I understand that we are typically oversampling force error, and so this matrix may not be very well-conditioned, it seems like in practice having a blanket regularization of the smallest singular value may not always be the best choice. If we take our vacuum stellarator free boundary tutorial and run the free boundary problem without using any continuation stepping, we get two very different results with vs without this regularization: (left with the regularization, right without)

Image

of course, with Fourier continuation we get good results in both cases, but this is showing that perhaps there are cases where the regularization is not helping us.

Maybe a different approach could be to not include this regularization and instead rely on the [cutoff](https://github.com/PlasmaControl/DESC/blob/548a83b30bf6ec319539a896fa6c5a560e052911/desc/optimize/_constraint_wrappers.py#L1315) to remove the problematically small singular values, or adding a threshold based approach based on the condition number like `sf += jnp.where(sf[-1]/sf[0] < 1e-8, sf[-1], 1e-8)` or something?

notebooks with the runs (also has runs with the ess scaling as that seems to help free boundary too)

[free_boundary_equilibrium_vac_no_regularization.ipynb](https://github.com/user-attachments/files/23346520/free_boundary_equilibrium_vac_no_regularization.ipynb)
[free_boundary_equilibrium_vac.ipynb](https://github.com/user-attachments/files/23346521/free_boundary_equilibrium_vac.ipynb)

YigitElma:
Adds new options to use during the computation of projected tangents. The operation is used to solve:

$$dx =\left(\frac{\partial F}{\partial x}\right)^{-1}\frac{\partial F}{\partial c}$$

- `svd` : to reproduce the old behavior, adds the smallest eigenvalue to all eigenvalues as regularization
- `qr` : uses QR to solve the system with Tikhonov regularization by estimating the smallest singular value. This is very similar to the previous regularization but applies damping to a smaller range of eigenvectors (default)

Resolves #2000
Resolves #1569 

image

Todo:
- [ ] Check different scenarios to understand the effect


tests/test_examples.py
```diff
@@ -265,30 +265,13 @@ def test_qh_optimization():
     eq = Equilibrium(M=5, N=5, Psi=0.04, surface=surf)
     eq.solve(verbose=3)
 
-    eq1 = run_qh_step(0, eq)
+    eq1 = run_qh_step(2, eq)
```
YigitElma: I should check the notebooks to see if anything is changed with this

YigitElma: @dpanici also reported this kind of effect in #2000 but I think here doing continuation with the current settings overcommits the lower order modes to a local minimum and cause issues in later steps. Since without continuation it converges immediately, I don't think this is a problem.

desc/optimize/_constraint_wrappers.py
```diff
@@ -583,14 +586,20 @@ def __init__(
         eq,
         perturb_options=None,
         solve_options=None,
+        inv_method="svd",
```
```suggestion
        inv_method="qr",
```

desc/optimize/optimizer.py
```diff
@@ -467,12 +467,14 @@ def _maybe_wrap_nonlinear_constraints(
     if wrapper is not None and wrapper.lower() in ["prox", "proximal"]:
         perturb_options = options.pop("perturb_options", {})
         solve_options = options.pop("solve_options", {})
+        inv_method = options.pop("prox_inv_method", "svd")
```
YigitElma:
```suggestion
        inv_method = options.pop("prox_inv_method", "qr")
```

desc/optimize/optimizer.py
```diff
@@ -467,12 +467,14 @@ def _maybe_wrap_nonlinear_constraints(
     if wrapper is not None and wrapper.lower() in ["prox", "proximal"]:
         perturb_options = options.pop("perturb_options", {})
         solve_options = options.pop("solve_options", {})
+        inv_method = options.pop("prox_inv_method", "qr")
```
f0uriest, line 604: I'd like to see more testing on a wide class of problems before we change the default.

tests/test_examples.py
f0uriest: If this test doesn't work with the new option that suggests that there may be something wrong. We should figure out what the actual issue is rather than changing the test.

(note that this is such low resolution that it would already converge without the fourier continuation, but we want to test that part as well)

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ New Features
 
 - Adds new option for ``loss_function``, ``"sum"``, which changes an objective to target the sum of the values computed.
 - Adds ``"scipy-l-bfgs-b"`` optimizer option as a wrapper to scipy's ``"l-bfgs-b"`` method.
+- Adds ``prox_inv_method`` option to be used in equilibrium constrained optimization problems. ``prox_inv_method`` can be ``qr``, ``svd`` or ``svd-reg``. ``svd-reg`` is the legacy behaviour which had an unnecessary regularization term. The new default is ``qr`` which is faster than ``svd`` without loss of significant accuracy. Note that this may change the results of the optimization scripts. For number of example cases, it is been seen that the new default facilitates the optimization without continuation.
```
f0uriest: these example cases are super low res, so I wouldn't take that as representative of the true behavior

f0uriest (changes requested):

desc/optimize/_constraint_wrappers.py
```diff
@@ -1309,11 +1322,20 @@ def _proximal_jvp_f_pure(constraint, xf, constants, dc, eq_feasible_tangents, dx
...
-    uf, sf, vtf = jnp.linalg.svd(Fxh, full_matrices=False)
-    sf += sf[-1]  # add a tiny bit of regularization
-    sfi = jnp.where(sf < cutoff * sf[0], 0, 1 / sf)
-    return vtf.T @ (sfi * (uf.T @ Fc))
+    if inv_method == "svd":
+        cutoff = jnp.finfo(Fxh.dtype).eps * max(Fxh.shape)
+        uf, sf, vtf = jnp.linalg.svd(Fxh, full_matrices=False)
+        sfi = jnp.where(sf < cutoff * sf[0], 0, 1 / sf)
+        return vtf.T @ (sfi * (uf.T @ Fc))
+    elif inv_method == "svd-reg":  # this option will be deleted
```
dpanici: why the comment on deleting the option?

desc/optimize/optimizer.py
dpanici: agreed, I will try it on some trickier free boundary problems I have

desc/optimize/_constraint_wrappers.py
YigitElma: I was planning to remove it, but I will keep it at least for foreseable future

YigitElma:
```python
eq = get("precise_QA")
cons = get_fixed_boundary_constraints(eq)
cons = maybe_add_self_consistency(eq, cons)
con = ObjectiveFunction(cons)
lcp = LinearConstraintProjection(obj, con)
lcp.build()
J = obj.jvp_scaled_error(lcp._feasible_tangents.T, obj.x(eq), obj.constants).T
uf, sf, vtf = jnp.linalg.svd(J, full_matrices=True)

smallest_v_dir = put(jnp.zeros(eq.dim_x), lcp._unfixed_idx, lcp._Z @ vtf[-1, :])
biggest_v_dir = put(jnp.zeros(eq.dim_x), lcp._unfixed_idx, lcp._Z @ vtf[0, :])
plt.plot(smallest_v_dir)
size_total = 0
# Add vertical lines to divide up the horizontal axis into parameters
for key, value in eq.params_dict.items():
    size = value.size
    size_total += size
    plt.axvline(x=size_total, color="r", linestyle="--")
plt.xlim([0, len(biggest_v_dir)])

plt.figure()
params = obj.unpack_state(smallest_v_dir)[0][0]
plt.plot(eq.R_basis.modes[:, 0]/eq.L)
plt.plot(eq.R_basis.modes[:, 1]/eq.M)
plt.plot(eq.R_basis.modes[:, 2]/eq.N)
plt.plot(params["R_lmn"], "k")
```
image
image

dpanici:
Let's just do this: run each of below cases and compare the methods (svd, qr, svd reg)
- each of our stage-one tutorials (basic, advanced, free bdry, omnigenity, stability, transport)
- precise QA script
- precise QH script
- reactor QA bootstrap consistency script

desc/optimize/_constraint_wrappers.py
```diff
@@ -1326,11 +1339,20 @@ def _proximal_jvp_f_pure(constraint, xf, constants, dc, eq_feasible_tangents, dx
...
+        cutoff = jnp.finfo(Fxh.dtype).eps * max(Fxh.shape)
+        uf, sf, vtf = jnp.linalg.svd(Fxh, full_matrices=False)
+        sf += sf[-1]
+        sfi = jnp.where(sf < cutoff * sf[0], 0, 1 / sf)
+        return vtf.T @ (sfi * (uf.T @ Fc))
+    elif inv_method == "qr":
+        Q, R = qr(Fxh, mode="economic")
```
YigitElma: Haiger-Hagem decompositon to get smallest singular value then use that as Levenberg-Marquardt regularization.

desc/optimize/_constraint_wrappers.py
```diff
@@ -1350,11 +1370,26 @@ def _proximal_jvp_f_pure(constraint, xf, constants, dc, eq_feasible_tangents, dx
...
+        uf, sf, vtf = jnp.linalg.svd(Fxh, full_matrices=False)
+        sf += sf[-1]
+        sfi = jnp.where(sf < cutoff * sf[0], 0, 1 / sf)
+        return vtf.T @ (sfi * (uf.T @ Fc))
+    elif inv_method == "qr":
+        Qt_fc, R = qr_multiply(Fxh, Fc, mode="right")
+        # Levenberg-Marquardt regularization with alpha=smin is equivalent to
+        # above regularization. Use power iteration to estimate smin/smax
+        smin = estimate_singular_value(R, mode="min")
+        smax = estimate_singular_value(R, mode="max")
```
YigitElma, line 1384: Maybe just use the matrix norm instead, should upper bound the max singular value

desc/optimize/_constraint_wrappers.py
YigitElma: I used power iteration to estimate the largest/smallest singular values.

desc/optimize/_constraint_wrappers.py
YigitElma: These are very cheap, so not worth it probably

desc/optimize/optimizer.py
YigitElma: I reran the tutorials that use proximal projection. Most of them have very minor numerical changes that don't affect the result visually. Effective ripple optimization was almost the same, with a slightly lower objective value at the end. The ballooning tutorial had the biggest visual difference, even though the objective values were similar. The biggest difference was the initial trust radius, which resulted in bigger steps. When I increased the `maxiter`, the new `qr` method achieved a smaller objective value by running until iteration 20, while the old `svd` gave a bad approximation at iteration 10.

I will share the results for `precise_XX` scripts too.

PR #2032, opened 2025-12-11
Allow combination of ESS and automatic scaling

f0uriest:
This allows you to use both ESS and automatic jacobian scaling in the same problem, by using different scales for different objects/variables. Note this still does not allow using automatic scaling "on top of" ESS, since in that case I think the automatic scaling would undo the effect of ESS.

Basically, any element of `x_scale` set to 0 means use the automatic jacobian scaling, and now `x_scale=="auto"` basically just sets `x_scale=np.zeros()`

A few questions/concerns:
- I'm not totally sure how it works with LinearConstraintProjection if you have a constraint that couples variables with different scaling types. IE, if you used "ess" for `R_lmn` but "auto" for `Rb_lmn` and have a `BoundaryRSelfConsistency` constraint, the results may be somewhat undefined
- Right now for ESS, we use a default scale of 1 for any extra variables, but we could maybe use a scale of 0 to use auto scaling for non-ess variables?


desc/equilibrium/equilibrium.py
```diff
@@ -2361,7 +2362,7 @@ def optimize(
             stopping tolerances. `None` will use defaults for given optimizer.
         maxiter : int
             Maximum number of solver steps.
-        x_scale : array, list[dict | ``'ess'``], ``'ess'`` or ``'auto'``, optional
+        x_scale : array, list[dict | ``'ess'``, ``'auto'``], ``'ess'`` or ``'auto'``
```
dpanici, line 2439: example use to use auto for just lambda:

x_scale = get_ess_scale(eq)

x_scale["L_lmn"] = np.zeros_like(x_scale["L_lmn"])

dpanici:
- Have LinearConstraintProjection apply the ess scale (or whatever fixed, pre-computable scale is passed in)
  - in which case, only have one xscale, not two
- How to mix ess / auto? if ess for just some variables make sense, what to choose for the others?  
  - e.g. equilibrium, I,G should be what? I am in favor of `"auto"` for these, using jac scaling
- rename "auto" to "jac" for optimizer xscale option? to make not degenerate with linear constraint projections xscale

Can we write down the possible branches of how the xscale can be applied right now in master code before this PR is applied? (and with/without constraints on the optimization, if no linear constraints we get no D matrix bc we never project/recover, an option here would be to always make the LinearConstraintProjection but have it be identity / D for project if there are no linear constraints passed in)

- optimizer xscale = "auto" for ALL variables
  - linear constraint projection xscale = "auto" - Fine, is how we already do it, this works fine
  - linear constraint projection xscale = some passed-in array -this also I think works fine, we have to trust user is passing in something sensible
 - optimizer xscale = "ess" for ALL variables
  - linear constraint projection xscale = "auto"  - this does nothing in almost all cases, unless coefficient is >1e2
    - this may matter in coil optimization with shape change and current change. what happens in this case rn is that current is not really "scaled" other than making x_reduced order unity assuming initial currents were correctly scaled to expected magnitudes
  - linear constraint projection xscale = some passed-in array
 
Different problems if auto/ess are mixed for the optimizer xscale (some for auto, some for ess). Need to figure out good solution for this case, and figure out if linear constraint projections are still working correctly in this case

How do we want this to change with this PR?

desc/optimize/_desc_wrappers.py
```diff
@@ -260,10 +254,7 @@ def _optimize_desc_least_squares(
     """
     assert constraint is None, f"method {method} doesn't support constraints"
     options = {} if options is None else options
-    if not isinstance(x_scale, str) and jnp.allclose(x_scale, 1):
-        options.setdefault("initial_trust_radius", 1e-3)
-        options.setdefault("max_trust_radius", 1.0)
-    elif options.get("initial_trust_radius", "scipy") == "scipy":
+    if options.get("initial_trust_radius", "scipy") == "scipy":
```
YigitElma, line 292: Why do we have these here and not in the optimization function itself? I remember getting confused because this default is different than the one set inside the function. If there is no specific reason, I would vote for having these defaults set in the same place.

desc/optimize/optimizer.py
```diff
@@ -386,62 +385,80 @@ def optimize(  # noqa: C901
...
-            all_scales.append(tng.pack_params(xsc))
+            # we pack then unpack as a check for the correct size/structure
+            all_scales.append(tng.unpack_params(tng.pack_params(xsc)))
         elif isinstance(xsc, str) and xsc == "ess":
-            scl = tng._get_ess_scale(ess_alpha, ess_order, ess_min_value)
-            all_scales.append(tng.pack_params(scl))
+            scl = tng._get_ess_scale(ess_alpha, ess_order, ess_min_value, ess_default)
+            all_scales.append(scl)
+        elif isinstance(xsc, str) and xsc == "auto":
+            scl = tree_map(jnp.zeros_like, tng.params_dict)
```
YigitElma, line 447: Whether #2041 is merged before or after this, we need to add special logic to SGD type optimizers to deal with these 0 values. SGD only looks at the gradient and norm scaling doesn't work there.

desc/optimize/utils.py
```diff
@@ -484,11 +484,15 @@ def compute_jac_scale(A, prev_scale_inv=None):
 
     if prev_scale_inv is not None:
         scale_inv = jnp.maximum(scale_inv, prev_scale_inv)
-    return 1 / scale_inv, scale_inv
+    scale = 1 / scale_inv
+    if user_scale is not None:
+        scale = jnp.where(user_scale == 0, scale, user_scale)
```
YigitElma, line 489: I am not sure if this will work as intended. The reason Jacobian column scaling works (to some extent) is when you look at all the norms, their relative relation gives a sense of scaling. If we mix the column norms with the random scaling given by the user, we lose this relationship. For example, let's say norms are [1000, 2000, 1000] and user given x_scale is [1, 0, 1], this will result in [1, 2000, 1]. I don't think is will good results in most cases due to this inconsistent scaling.

Maybe we can first normalize the norms based on the maximum of the user given x_scale, then use that? Something like:
```python
if user_scale is not None:
        user_scale_max = jnp.max(user_scale)
        user_scale_max_id = jnp.where(user_scale == user_scale_max)[0][0]
        scale_max_at_id = scale[user_scale_max_id]
        scale = scale * user_scale_max / scale_max_at_id
        scale = jnp.where(user_scale == 0, scale, user_scale)
```

This first normalizes the scales such that the norm of the column corresponding to maximum user scale, is equal to maximum user scale, this removes the order of magnitude difference of the user scale and Jacobian column norm. We can find a better solution probably.

desc/optimize/utils.py
```diff
@@ -497,7 +501,11 @@ def compute_hess_scale(H, prev_scale_inv=None):
...
```
YigitElma, line 506: Same as above

desc/objectives/utils.py
```diff
@@ -121,17 +121,18 @@ def factorize_linear_constraints(objective, constraint, x_scale="auto"):  # noqa
...
-        x_scale = objective.x(*objective.things)
+        x_scale = auto_x_scale
     errorif(
         x_scale.shape != xp.shape,
         ValueError,
         "x_scale must be the same size as the full state vector. "
         + f"Got size {x_scale.size} for state vector of size {xp.size}.",
     )
-    D = np.where(np.abs(x_scale) < 1e2, 1, np.abs(x_scale))
+    # x_scale==0 means use auto scale, otherwise use user scale
```
dpanici, line 134: sort of confused here, which auto is being referred to here? Are we overloading auto to mean both use inverse jac norm scales iteratively updated during optimization, but also as the overall scale of each variable used in factorize_linear_constraints to make sure the particular solution does not carry the entire magnitude of the variable?

desc/optimize/_constraint_wrappers.py
```diff
@@ -236,21 +236,24 @@ def update_constraint_target(self, eq_new):
...
-            # we also normalize Z to make each column have unit norm
-            self._Z = self._Z / jnp.linalg.norm(self._Z, axis=0)
+        x_scale = self._x_scale
+        x0 = self._objective.x(*self._objective.things)
+        auto_x_scale = np.where(np.abs(x0) < 1e2, 1, np.abs(x0))
+
+        if isinstance(x_scale, str) and x_scale == "auto":
+            x_scale = auto_x_scale
+
+        self._D = jnp.where(x_scale == 0, auto_x_scale, x_scale)
```
dpanici, line 246: this is not necessary if the x_scale is not auto right? but I guess does not hurt to have? or is this going to be changing if x_scale is passed in or "ess" now?

desc/optimize/_scipy_wrappers.py
```diff
@@ -360,7 +362,14 @@ def _optimize_scipy_least_squares(  # noqa: C901
     """
     assert constraint is None, f"method {method} doesn't support constraints"
     options = {} if options is None else options
-    x_scale = "jac" if x_scale == "auto" else x_scale
+    if np.all(x_scale == 0):
+        x_scale = "jac"  # automatic scaling
+    else:
+        # we can't combine adaptive scaling with user specified scale, but we
```
dpanici, line 372: which user specified scale is this here? like if one used `"ess"` then what is happening here? is ess being used still?

PR #2034, opened 2025-12-12
NFP bug fix in `MagneticField.save_mgrid()` and `SplineMagneticField.from_field()`

ddudt:
This fixes two small bugs related to NFP in the magnetic field classes:

1. `SplineMagneticField.from_field()` took an input argument `NFP` but never actually used it. It would always use `field.NFP` or default to 1 instead. This PR allows the input argument to override the field's attribute. 
2. `MagneticField.save_mgrid()` would always use `self.NFP` or default to 1 if the attribute didn't exist. This PR adds `NFP` as an input argument to override this logic (same as above). 

Here is an example where overriding the default `field.NFP` attribute is useful: 
```
coil0 = FourierPlanarCoil(center=[2, 0, 0], normal=[0, 1, 0], r_n=[1])
coil1 = FourierPlanarCoil(center=[2, 0, 0], normal=[0, 1, 0], r_n=[2])
coilset0 = CoilSet(coil0, NFP=4)
coilset1 = CoilSet(coil1, NFP=4)
mixed_coilset = MixedCoilSet((coilset0, coilset1))
```
A `MixedCoilSet` always has `NFP = 1` by construction, even though all of its coils in this example have NFP=4 symmetry. With the changes in this PR, you can then take advantage of this field period symmetry by manually passing in the NFP argument: 
```
spline_field = SplineMagneticField.from_field(mixed_coilset, NFP=4)
mixed_coilset.save_mgrid("mgrid.nc", NFP=4)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ alternative to fourier continuation methods.
 
 Bug Fixes
 
-- No longer uses the full Hessian to compute the scale when ``x_scale="auto"`` and using a scipy optimizer that approximates the hessian (e.g. if using ``"scipy-bfgs"``, no longer attempts the Hessian computation to get the x_scale)
+- No longer uses the full Hessian to compute the scale when ``x_scale="auto"`` and using a scipy optimizer that approximates the hessian (e.g. if using ``"scipy-bfgs"``, no longer attempts the Hessian computation to get the x_scale).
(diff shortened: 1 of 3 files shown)


ddudt:
> Does the user need to give proper R, Z, phi for this to work as expected or the period given to `interpax` handles that?

If you are specifying the NFP then you only need to give the phi coordinates in a single field period. For example: 
```
spline_field = SplineMagneticField.from_field(mixed_coilset, phi=np.linspace(0, np.pi / 2), NFP=4)
```
I don't think it will break anything otherwise, it will just be a wasteful use of resolution.

PR #2041, opened 2025-12-16
Add a lot of optimizers including `optax` support to `sgd`

YigitElma:
- Unifies all the SGD type optimizers, developers only need to implement the update rule
- `x_scale` is now used with SGD methods too
- Adds wrappers for `optax` optimizers, and they can be called by `optax-name`
- Any custom optax optimizer can be used via
```python
        import optax
        from desc.examples import get

        eq = get("DSHAPE")

        # Optimizer
        opt = optax.chain(
            optax.sgd(learning_rate=1.0),
            optax.scale_by_zoom_linesearch(max_linesearch_steps=15),
        )
        optimizer = Optimizer("optax-custom")
        eq.solve(optimizer=optimizer, options={"optax-options": {"update_rule": opt}})
```


desc/optimize/stochastic.py
```diff
@@ -27,12 +28,25 @@ def sgd(
...
+    Update rule for 'adam':
+        m_t = beta * m_{t-1} + (1 - beta) * grad(x_{t-1})
+        v_t = beta2 * v_{t-1} + (1 - beta2) * grad(x_{t-1})^2
+        m_hat = m_t / (1 - beta^t)
+        v_hat = v_t / (1 - beta2^t)
+        x_t = x_{t-1} - alpha * m_hat / (sqrt(v_hat) + epsilon)
+
+    Update rule for 'rmsprop':
+        v_{k}   = beta*v_{k-1} + (1-beta)*grad(x_{k})^2
+        x_{k+1} = x_{k} - alpha * grad(x_{k}) / (sqrt(v_{k}) + epsilon)
```
YigitElma: Some literature uses different Greek letters for alpha and beta, but I used these to make api slightly more compact.

desc/particles.py
YigitElma, line 1: Sorry for the unrelated change, noticed this while fixing the doc errors

desc/optimize/stochastic.py
```diff
@@ -88,24 +135,47 @@ def sgd(
...
+        method not in ["sgd", "adam", "rmsprop"],
+        f"Available options for method are 'sgd', 'adam' and 'rmsprop', but {method} "
+        "is given.",
+    )
+    errorif(
+        isinstance(x_scale, str) and x_scale not in ["auto"],
+        ValueError,
+        "x_scale should be one of 'auto' or array-like, got {}".format(x_scale),
+    )
+    if isinstance(x_scale, str):
```
dpanici, line 162: should say in docstring that 'x_scale="auto"' is essentially just means no scaling as we just set xscale to 1 in that case.

dpanici (changes requested):
just the small docstring fix, should be explitict that x_scale='"auto"` does no scaling here

dpanici (dismissed):

desc/optimize/stochastic.py
```diff
@@ -13,12 +13,13 @@
 )
 
 
-def sgd(
+def generic_sgd(
```
f0uriest: `sgd` is technically public (https://desc-docs.readthedocs.io/en/stable/_api/optimize/desc.optimize.sgd.html#desc.optimize.sgd) so if we want to change the name we should keep an alias to the old one with a deprecation warning. That said, I'm not sure we really need to change the name. "SGD" is already used pretty generically in the ML community for a bunch of first order stochastic methods like ADAM, ADAGRAD, RMSPROP, etc

desc/optimize/stochastic.py
```diff
@@ -74,11 +108,24 @@ def sgd(
         are the same arguments passed to fun and grad. If callback returns True
         the algorithm execution is terminated.
     options : dict, optional
-        dictionary of optional keyword arguments to override default solver settings.
+        dictionary of optional keyword arguments to override default solver settings
+        for the update rule chosen.
+
+        - ``"alpha"`` : (float > 0) Learning rate. Defaults to
+          1e-1 * ||x_scaled|| / ||g_scaled||.
```
f0uriest: this seems pretty large (steps would be 10% of x), have you checked how robust this is?

desc/optimize/stochastic.py
```diff
@@ -88,24 +135,47 @@ def sgd(
...
     ngev = 0
     iteration = 0
 
     N = x0.size
     x = x0.copy()
-    v = jnp.zeros_like(x)
     f = fun(x, *args)
     nfev += 1
-    g = grad(x, *args)
+    g = grad(x, *args) / x_scale
```
f0uriest: shouldnt this be *x_scale here?

```
xs = x/x_scale
gs = df/dxs = df/dx*dx/dxs = df/dx*x_scale

File: desc/optimize/stochastic.py [hunk 6]
```diff
@@ -88,24 +135,47 @@ def sgd(
...
     ngev += 1
-
-    maxiter = setdefault(maxiter, N * 100)
-    g_norm = jnp.linalg.norm(g, ord=2)
+    # scaled and unscaled norms
+    g_norm = jnp.linalg.norm(g * x_scale, ord=2)
```
f0uriest: see above

desc/optimize/stochastic.py
```diff
@@ -27,12 +28,40 @@ def sgd(
...
 
-    Update rule is x_{k+1} = x_{k} - alpha*v_{k}
-                   v_{k}   = beta*v_{k-1} + (1-beta)*grad(x)
+    This is the generic function. The update method is chosen based on the `method`
+    argument.
 
-    Where alpha is the step size and beta is the momentum parameter.
+    Update rule for ``'sgd'``:
+
+    .. math::
```
f0uriest: personally I prefer unicode for stuff like this. TeX looks nice in the rendered html docs, but is much harder to read as code.

desc/particles.py
f0uriest: see previous comment about TeX vs unicode

f0uriest (commented):
I'd double check that the `x_scale` logic is correct

Also, did you look at whether we could just wrap stuff from optax?

From the examples eg https://optax.readthedocs.io/en/latest/api/optimizers.html#optax.adam

it looks like the user could just pass in an `optax.solver` and then we can just do

```python
opt_state = solver.init(x0)
...
g = grad(x)*x_scale
updates, opt_state = solver.update(g, opt_state, x)
x = optax.apply_updates(x, x_scale*updates)
```

or something similar. That would give users access to a much wider array of first order optimizers, and save us having to do it all ourselves

f0uriest:
`optimistix` wraps stuff in a jax loop which jits things, and uses `jax.jacfwd` etc internally, so it won't work with our `Proximal` stuff. But `optax` doesn't care how you compute the gradients so its more flexible (and could for example be used with SPSA as well)

desc/optimize/stochastic.py
YigitElma: Yeah, SGD is in fact the general name. I can revert to the old name, and just emphasize that "sgd" option is with nesterov momentum. I was trying to make a distinction I guess

desc/optimize/stochastic.py
YigitElma: I mostly agree and don't have a strong stance either way. My general preference is to use LaTeX for complex equations or public-facing objectives that users will first encounter in the documentation (like guiding center equations, optimization algorithms). For internal development notes or specific compute functions that aren't usually viewed on the web, I’m fine with Unicode since it keeps the source code more readable.

desc/optimize/stochastic.py
YigitElma: I was trying to solve eq with these and even though none of them converged 10% was better for a variety of equilibrium. I haven't checked other optimization problems. Reverted the change and added a backguard against 0 and NaNs.

dpanici (dismissed):

desc/optimize/_desc_wrappers.py
```diff
@@ -376,9 +413,18 @@ def _optimize_desc_fmin_scalar(
 
 
 @register_optimizer(
-    name="sgd",
-    description="Stochastic gradient descent with Nesterov momentum"
-    + "See https://desc-docs.readthedocs.io/en/stable/_api/optimize/desc.optimize.sgd.html",  # noqa: E501
+    name=["sgd", "optax-custom"] + ["optax-" + opt for opt in _all_optax_optimizers],
```
f0uriest, line 414: are we planning on deprecating `sgd` in favor of `optax-sgd`?

desc/optimize/stochastic.py
```diff
@@ -194,3 +317,10 @@ def sgd(
         print("         Gradient evaluations: {:d}".format(result["ngev"]))
 
     return result
+
+
+def _sgd(g, v, alpha, beta):
```
f0uriest: is this the same as some version of `optax-sgd`? if so I'd vote to remove this and just do something like `if method == "sgd": method = "optax-sgd"`. Then we can simplify a lot of the code here and just always assume we're using optax stuff

tests/test_optimizer.py
```diff
@@ -267,6 +271,121 @@ def test_sgd_convex(self):
...
+            # extra hyperparameters
+            eq.solve(
+                optimizer=optimizer,
+                options={"optax-options": {"update_rule": opt, "learning_rate": 0.1}},
+                verbose=3,
+                maxiter=2,
+            )
+
+    @pytest.mark.unit
+    def test_available_optax_optimizers(self):
```
f0uriest: I'm not sure if we want this test. I get the idea is to let us know if optax adds new optimizers, but that assumes we need to support everything that optax does, and it means this test could start failing at any point because of backwards compatible changes in another library. I'd vote to remove it and just periodically check whats new in optax and update accordingly

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ or if multiple things are being optimized, `x_scale` can be a list of dict, one
 - Adds new option `x_scale='ess'` to use exponential spectral scaling from (Jang 2025) which has been shown to improve performance and robustness as an
 alternative to fourier continuation methods.
 - Adds ``"scipy-l-bfgs-b"`` optimizer option as a wrapper to scipy's ``"l-bfgs-b"`` method.
+- The ``x_scale`` parameter can now be used with stochastic gradient descent type optimizers.
+- Adds wrappers for ``optax`` optimizers. They can be used by prepending ``'optax-'`` to the name of the optimizer. Additional arguments to the optimizer such as `learning_rate` can be pass via ``options = {'optax-options': {'learning_rate': 0.01}}``. Even a custom ``optax`` optimizer can be used by specifying the method as ``'optax-custom'`` and passing the ``optax`` optimizer via the ``'update-rule'`` key of `optax-options` in the `options` dictionary.
```
f0uriest: > They can be used by prepending ``'optax-'`` to the name of the optimizer, 

include example syntax `eg ``'optax-adam'```

desc/optimize/stochastic.py
YigitElma: I am not 100% sure but it can be equivalent to optax-sgd with momentum=beta, learning_rate=alpha and nesterov=True. The amount of code dedicated to that is not much and it also handles future implementations (I don't know if anyone wants to add their sgd optimizers but anyway). If people want to I can add depreciation but it doesn't seem urgent to me.

tests/test_optimizer.py
YigitElma: I referenced the below comment in `_all_optax_optimizers`.

dpanici (dismissed):

desc/optimize/stochastic.py
```diff
@@ -87,25 +119,110 @@ def sgd(
...
+        "is given.",
+    )
+    errorif(
+        isinstance(x_scale, str) and x_scale not in ["auto"],
+        ValueError,
+        "x_scale should be one of 'auto' or array-like, got {}".format(x_scale),
+    )
+    deprecated_sgd = False
+    if method == "sgd":
+        # warn the user but do not fail the pytest
```
f0uriest: why is this here? shouldn't this be in the test?

CHANGELOG.md
```diff
@@ -32,6 +34,11 @@ Performance Improvements
 
 - `ProximalProjection.grad` uses a single VJP on the objective instead of multiple JVP followed by a manual VJP. This should be more efficient for expensive objectives.
 
+Deprecations
+
+- ``sgd`` optimizer is deprecated in favor of ``optax-sgd``, and will be removed in a future release. To achieve the same behavior with `optimizer = Optimizer['sgd']` and `options={'alpha': ..., 'beta': ...}` when the optimizer is removed, one can use `optimizer = Optimizer['optax-sgd']` and `options={'optax-options': {'learning_rate': alpha, 'momentum': beta, 'nesterov': True}}`.
```
f0uriest:
```suggestion
- ``sgd`` optimizer is deprecated in favor of ``optax-sgd``, and will be removed in a future release. To achieve the same behavior with `optimizer = Optimizer('sgd')` and `options={'alpha': ..., 'beta': ...}` when the optimizer is removed, one can use `optimizer = Optimizer('optax-sgd')` and `options={'optax-options': {'learning_rate': alpha, 'momentum': beta, 'nesterov': True}}`.
```

desc/optimize/stochastic.py
YigitElma: There are multiple tests and some are parameterized, this seemed easier

f0uriest: But doesn't this mean the warning isn't actually emitted?

YigitElma: It is not treated as a warning but the message is still printed as a warning. So, this is equivalent to something like
```
print("DepreciationWarning: SGD is deprecated.")
```
with proper syntax for the warnings. It will be more annoying to add conditionals to parameterized tests; I would rather not do that

YigitElma: And tests look like 
image
but still pass.

dpanici (dismissed):

dpanici (dismissed):

ddudt (dismissed):

dpanici (approved):

ddudt (approved):

Issue #18, opened 2020-09-27
plotting to specified axes


f0uriest:
All the plotting routines should have an option to plot to a preallocated figure/axis, and should also return the fig/axis that was used, whether an existing one was used or a new one created.

Issue #14, opened 2020-09-24
2nd order perturbations


f0uriest:
Autodiff hessian and solve quadratic tensor equation for dx
ways to format it as matrix equation?

ddudt:
Now implemented in `dev` but might not be stable.  Boundary perturbations for both orders seem to shift the major radius outward, even with low `errr_ratio`.  Need testing to determine if there is a bug in the BC or perturbation functions.  (square system?)  
![bdry_pert_test](https://user-images.githubusercontent.com/33005725/95415626-c6812580-08fe-11eb-8216-13e8adee1090.png)

f0uriest:
how low did you go for `errr_ratio`? I've noticed that even with pretty low values the jacobian is still dominated by a few large values in the force error, and the bc error terms are pretty small comparatively. 
We could also look into doing a weighted least squares thing (effectively using a larger error ratio just for the perturbation stuff) or regularized as well by penalizing large changes in the low mode numbers

ddudt:
The issue was a bug in the boundary condition function.  This line: 
```
# ratio of non-axisymmetric boundary modes to use
ratio = jnp.clip(bdry_ratio+(bdryN == 0), 0, 1)
```
needed to be replaced with:
```
ratio = jnp.where(bdryN != 0, bdry_ratio, 1)
```
The performance might still be affected by the choice of `errr_ratio` weight.  
![bdry_pert_good](https://user-images.githubusercontent.com/33005725/96026483-8cdf7b80-0e24-11eb-8f7f-40369966d4a2.png)

PR #2044, opened 2025-12-22
Add Deflation methods

dpanici:
Deflation method motivation: find multiple solutions to non-convex optimization problems (which can include certain equilibirum solves)

This PR adds ways to apply deflation methods in stellarator optimization and equilibrium solving through the new `DeflationOperator` objective

- adds a new generic objective `DeflationOperator` whose cost is simply M(x;y) = 1/(x-y)^p + sigma (to add as constraints to an optimization like in [Tarek 2022 work](https://arxiv.org/abs/2201.11926)). This can be used as a standalone metric, or another `_Objective` can be passed to it to wrap it and return as the cost M(x;y)f(x) where f(x) is that `_Objective`'s compute value, like is done in [usual deflation ](https://arxiv.org/abs/1410.5620)
- Adds tutorial covering these

References:
- [Riley 2024](https://arxiv.org/abs/2409.14438) - for the `"exp"` deflation type
- [ Farrell 2015](https://arxiv.org/abs/1410.5620) - for the addition of the shift parameter and the general form of deflation used

TODO
- [x] add tests
- [x] update changelog
- [x] update `ForceBalanceDeflated` to use pytree inputs for `params_to_deflate_with`
- [x] add option for using single shift, like discussed in [Riley 2024](https://arxiv.org/abs/2409.14438)
- [x] Add a wrapper objective so that one can multiply the deflation operator with any arbitrary objective, instead of only able to add it as an extra cost when doing stage one/two optimization.
- [x] figure out how to avoid recompilation

Future work for another PR:
- Implement algorithms from [Riley 2024](https://arxiv.org/abs/2409.14438)
  - Implement deflated line-search Gauss-Newton algorithm like they use
  - Adapt their algorithm for our usual trust-region approach
- add `_equilibrium` as attribute of DeflationOperator and test using it in proximal-lsq-exact


f0uriest (commented):
So as I see it there are two main ways of applying deflation
1. As a multiplicative factor on the objective eg `f(x) -> M(x,x*)f(x)`
2. As an additional inequality constraint `M(x,x*)<r`

The new `DeflationOperator` objective seems to cover the 2nd case, but for the first case `ForceBalanceDeflated` only works for equilibrium problems. I think it would be better as a sort of "wrapper objective" that can be applied to any objective to multiply the deflation operator.

We could possibly combine the two and make it a single objective like
```python
class DeflationOperator:
    """Multiplicative or constraint type deflation"""

def __init__(self, objective=None, ...):
    self.objective = objective

def compute(self, x):
    if self.objective is not None:
        f = self.objective.compute(x)
    else:
        f = 1
    return M(x,x*)*f
```

This would cover both cases, either treating the deflation as an extra constraint (with `objective=None`) or applying multiplicative deflation to an arbitrary objective (eg by passing `objective=ForceBalance()`)

f0uriest:
Also, we should decide if we want to implement some sort of optimizer wrapper to handle the delfation stuff, or at least should give an example notebook that uses the new classes in a loop to show how to actually do things

dpanici:
> Check if recompiling on subsequent deflation iterations, and if not, see if can do dummy values for the full amount you know you will deflate against. Dummy can be inf or nan or something like that. Either automated or show manually in notebook

Two sources of potential recompilation:

- if the `objective` passed in is not the same exact objective as was passed in and built/compiled in a prior iteration, then it will be recompiled. I think this is maybe because the objective is a `_static_attr` of the `DeflationOperator`? need to think about it more
- if the length of `things_to_deflate` changes, then a recompilation is triggered, even if the above point is addressed by re-using the same `objective` object.

desc/objectives/_generic.py
```diff
@@ -668,3 +682,246 @@ def compute(self, params, constants=None):
...
+        indices to deflate with for each corresponding parameter (int array).
+        Use True (False) instead of an int array to deflate all (none) of the indices
+        for that parameter.
+        Must have the same pytree structure as thing.params_dict.
+        The default is to deflate all indices of all parameters.
+    objective: _Objective, optional
+        Objective to wrap with the DeflationWrapper. If not None, the cost will
+        be M(x;xₖ)f(x) where f(x) is the Objective's cost. If None, then the cost
+        returned will be M(x;xₖ). The objective must accept only one optimizable
+        thing, and it must be the same as the thing passed to the DeflationWrapper
```
YigitElma:
```suggestion
        thing, and it must be the same as the thing passed to the DeflationOperator
```

desc/objectives/_generic.py
```diff
@@ -668,3 +682,246 @@ def compute(self, params, constants=None):
...
+    """
+
+    __doc__ = __doc__.rstrip() + collect_docs(
+        target_default="``target=0``.", bounds_default="``target=0``."
+    )
+    _static_attrs = _Objective._static_attrs + [
+        "_deflation_type",
+        "_multiple_deflation_type",
+        "_single_shift",
+        "_objective",
```
YigitElma: I am not sure if this is necessary or not, `ObjectiveFunction` also has `self._objectives` and it is not static.

YigitElma: This might cause recompilation since `objective.thing` is changing?

desc/objectives/_generic.py
```diff
@@ -668,3 +682,246 @@ def compute(self, params, constants=None):
...
+        self._power = power
+        self._params_to_deflate_with = params_to_deflate_with
+        assert deflation_type in ["power", "exp"]
+        self._deflation_type = deflation_type
+        assert multiple_deflation_type in ["prod", "sum"]
+        self._multiple_deflation_type = multiple_deflation_type
+        self._single_shift = single_shift
+        self._objective = objective
+        if self._objective is not None:
+            assert not isinstance(self._objective, ObjectiveFunction), (
```
YigitElma: why not check if this is `_Objective`, it is a more general check

desc/objectives/_generic.py
```diff
@@ -668,3 +682,246 @@ def compute(self, params, constants=None):
...
+                "the objective in an ObjectiveFunction, simply pass the"
+                " sub-objective"
+            )
+            assert len(objective.things) == 1
+            assert isinstance(objective.things[0], type(thing))
+            assert objective.things[0] == thing
+            name = "Deflated " + self._objective._name
+            self._units = self._objective._units
+            self._scalar = self._objective._scalar
+            self._equilibrium = self._objective._equilibrium
```
YigitElma: I don't think this is necessary since deflated equilibrium objective is not actually equilibrium objective (like an objective that can be used with proximal)

desc/objectives/_generic.py
```diff
@@ -668,3 +682,246 @@ def compute(self, params, constants=None):
...
+        self._params_to_deflate_with = broadcast_tree(
+            self._params_to_deflate_with, default_params
+        )
+        self._indices = tree_leaves(self._params_to_deflate_with)
+        assert tree_structure(self._params_to_deflate_with) == tree_structure(
+            default_params
+        )
+
+        if self._objective is not None:
+            self._objective.build()
```
YigitElma:
```python
if not self._objective.built:
    self._objective.build()
```

tests/test_objective_funs.py
```diff
@@ -3553,6 +3555,53 @@ def test_compute_scalar_resolution_mercier(self):
...
         np.testing.assert_allclose(f, f[-1], rtol=2e-2)
 
+    @pytest.mark.regression
+    def test_compute_scalar_resolution_ForceBalanceDeflated(self):
+        """Deflated Force Balance."""
+        eq0 = self.eq.copy()
+        eq0.set_initial_guess()  # make eq0 different than self.eq
+        f = np.zeros_like(self.res_array, dtype=float)
+        for i, res in enumerate(self.res_array):
+            self.eq.change_resolution(
```
YigitElma: does this affect other tests that use `self.eq`?

tests/test_objective_funs.py
```diff
@@ -3553,6 +3555,53 @@ def test_compute_scalar_resolution_mercier(self):
...
+                DeflationOperator(
+                    thing=self.eq,
+                    things_to_deflate=[eq0],
+                    objective=ForceBalance(self.eq),
+                ),
+                use_jit=False,
+            )
+            obj.build(verbose=0)
+            f[i] = obj.compute_scalar(obj.x())
+        np.testing.assert_allclose(f, f[-1], rtol=2e-2)
```
YigitElma: I don't understand what this test is doing. Why would this need to be true?

tests/test_objective_funs.py
```diff
@@ -3553,6 +3555,53 @@ def test_compute_scalar_resolution_mercier(self):
...
+            obj = ObjectiveFunction(
+                DeflationOperator(
+                    thing=self.eq,
+                    things_to_deflate=[eq0],
+                ),
+                use_jit=False,
+            )
+            obj.build(verbose=0)
+            f[i] = obj.compute_scalar(obj.x())
+        np.testing.assert_allclose(f, f[-1], rtol=2e-2)
```
YigitElma: This makes more sense, since you only change the grid resolution, spectral coefficients don't change and all of them must be the same, but I don't get the top one..

tests/test_objective_funs.py
```diff
@@ -3718,6 +3768,31 @@ def test_objective_no_nangrad_plasma_vessel(self):
         g = obj.grad(obj.x(eq, surf))
         assert not np.any(np.isnan(g)), "plasma vessel distance"
 
+    @pytest.mark.unit
+    def test_objective_no_nangrad_ForceBalanceDeflated(self):
+        """Deflation operator on force balance."""
+        eq = Equilibrium(L=2, M=2, N=2)
+        eq2 = eq.copy()
+        eq.R_n = eq.R_n * 1.1
```
YigitElma:
```suggestion
        eq.R_lmn = eq.R_lmn * 1.1
```
?

desc/objectives/_generic.py
dpanici: Right but we check `if self._objective is None` which I think would necessitate this?

desc/objectives/_generic.py
dpanici: Technically it could be used with Proximal, though I am not sure if I want that

tests/test_objective_funs.py
dpanici: we want the compute_scalar value to be independent of the grid resoltion (once we are confident we have enough grid resolution to adequately compute the sum of squares value). That is what this test checks, that the highest grid res and the intermediate ones are relatively close (as opposed to the highest one being larger simply due to it being the sum over more grid points or something like that).

tests/test_objective_funs.py
dpanici: not sure but to be certain I changed it to revert the grid res back

desc/objectives/_generic.py
```diff
@@ -668,3 +682,285 @@ def compute(self, params, constants=None):
...
+            self._constants = self._objective._constants
+        else:
+            self._dim_f = 1
+
+        self._is_none_mask = []
+        self._is_not_none_mask = []
+        self._not_all_things_to_deflate_are_None = not np.all(
+            [t is None for t in self._things_to_deflate]
+        )
+        self._not_all_things_to_deflate_are_None = float(
```
YigitElma: shouldn't this case actually throw an error during build? If there is nothing to deflate with, what is the point of deflating :)

YigitElma (changes requested):
The notebook seems to fail due to env name, I guess that was caused by jupyter lab (or cluster), if you just open it with VS code, it should fix it.

docs/notebooks/tutorials/deflation.ipynb
```diff
@@ -0,0 +1,1347 @@
+{
```
YigitElma, line 1: <ul><li>Solving the deflated system encourages**</li><li>In** DESC, it can be applied ...</li><li>... the optimizer not to find minima**</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2684827250'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='3'/>

YigitElma, line 1: <p>To be explicit, it would be nice to omit import alls</p><p> </p><ul><li>in order to run this notebook**</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2684827251'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='6'/>

YigitElma, line 1: <p>we'd like to get ** any</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2684827252'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='10'/>

YigitElma, line 1: <p>First line doesn't fit to the screen horizontally in the docs</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2684827253'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='11'/>

YigitElma, line 1: <p>from desc.io import load was already loaded above</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2684827256'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='12'/>

YigitElma, line 1: <p>There are 3 equilibria</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2684827258'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

YigitElma, line 1: <p>commented code</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2684827260'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='17'/>

YigitElma, line 1: <p>First line is again too long</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2684827263'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='22'/>

desc/objectives/_generic.py
dpanici: Potentially, to compile the function the first time so it does not recompile on subsequent iterations, and to simplify the script (like the QH script in the tutorial notebook always has deflation but it just does nothing when all are None)

YigitElma: ok, I still think it is slightly unnecessary but doesn't hurt to have

YigitElma (dismissed):

YigitElma (dismissed):

YigitElma (dismissed):

desc/objectives/_generic.py
```diff
@@ -669,3 +683,285 @@ def compute(self, params, constants=None):
...
+
+    """
+
+    __doc__ = __doc__.rstrip() + collect_docs(
+        target_default="``target=0``.", bounds_default="``target=0``."
+    )
+    _static_attrs = _Objective._static_attrs + [
+        "_deflation_type",
+        "_multiple_deflation_type",
+        "_single_shift",
```
f0uriest, line 777: does `params_to_deflate_with` need to be static as well?

desc/objectives/_generic.py
```diff
@@ -669,3 +683,285 @@ def compute(self, params, constants=None):
...
+    ]
+
+    _coordinates = "rtz"
```
f0uriest, line 781: not sure if this matters but this should also probably be inherited from the wrapped objective

desc/objectives/_generic.py
```diff
@@ -669,3 +683,285 @@ def compute(self, params, constants=None):
...
+        assert multiple_deflation_type in ["prod", "sum"]
+        self._multiple_deflation_type = multiple_deflation_type
+        self._single_shift = single_shift
+        self._objective = objective
+        if self._objective is not None:
+            assert isinstance(
+                self._objective, _Objective
+            ), "objective passed in must be an _Objective!"
+            assert len(objective.things) == 1
+            assert isinstance(objective.things[0], type(thing))
```
f0uriest: i think you can remove this check, as the one on the following line is stricter

desc/objectives/_generic.py
```diff
@@ -669,3 +683,285 @@ def compute(self, params, constants=None):
...
+            if not self._objective.built:
+                self._objective.build()
+            self._dim_f = self._objective._dim_f
+            self._normalization = self._objective._normalization
+            self._constants = self._objective._constants
+        else:
+            self._dim_f = 1
+
+        self._is_none_mask = []
+        self._is_not_none_mask = []
```
f0uriest, line 915: Why do we need two different masks here? Isn't one just the opposite of the other?

desc/objectives/_generic.py
```diff
@@ -669,3 +683,285 @@ def compute(self, params, constants=None):
...
+                not self._single_shift
+            )
+
+        if self._multiple_deflation_type == "prod":
+            # self._is_none_mask is 1.0 if a specific element was None originally,
+            # and 0.0 if a specific element was not None
+            # we want to ignore those elements in our final product deflation metric so
+            # we will make their values M_i=1.0 by subtracting M_i and adding 1.0
+            M_i = M_i + self._is_none_mask * (1 - M_i)
+            deflation_parameter = jnp.prod(M_i) * (
```
f0uriest: this bit might be cleaner doing `jnp.prod(M_i, where=self._is_not_None_mask, initial=1.0)`

desc/objectives/_generic.py
```diff
@@ -669,3 +683,285 @@ def compute(self, params, constants=None):
...
+                self._not_all_things_to_deflate_are_None
```
f0uriest: so if single shift is false and all the things to deflate are none (ie on the first step of a deflation), this will be zero right? But then it gets multiplied by the objective value later, so the objective is identically zero and the optimizer stops. I think we want a different "default" value depending on whether self.objective is None. If we're adding this as a penalty/constraint term than a default value of zero makes sense, but if its wrapping another objective then we want the default to be 1 i think

desc/objectives/_generic.py
```diff
@@ -669,3 +683,285 @@ def compute(self, params, constants=None):
...
+        loss_function=None,
+        deriv_mode="auto",
+        name="Deflation",
+        jac_chunk_size=None,
+        deflation_type="power",
+        multiple_deflation_type="prod",
+        single_shift=False,
+    ):
+        if target is None and bounds is None:
+            target = 0
```
f0uriest, line 807: we might want to do some extra checking on bounds here. Supposing the deflation operator is being used as a constraint then the minimum value it can take is `sigma`, so if that is greater than the bound then something is wrong. Similarly if there are multiple shifts etc then the bounds would have to increase accordingly.

desc/objectives/_generic.py
```diff
@@ -669,3 +683,285 @@ def compute(self, params, constants=None):
...
+                    jnp.atleast_1d(param[idx])
+                    for param, idx in zip(tree_leaves(t.params_dict), self._indices)
+                ]
+            )
+            for i, t in enumerate(self._things_to_deflate)
+        ]
+
+        diffs = jnp.vstack(diffs)
+        if self._deflation_type == "power":
+            M_i = 1 / jnp.linalg.norm(diffs, axis=1) ** self._power + self._sigma * (
```
f0uriest: do we want to include some small epsilon before dividing here? so its `1/(|x-xi| + eps) + sigma`

When playing around with deflation stuff I often made the mistake of starting from a previously found solution, (as you would in a regular continuation method), which then gives inf/nan for the objective and gradient. Adding some small regularization would avoid this without affecting the desired behavior too much I think.

Alternatively we should at least mention that somewhere to remind people, or possibly add a quick check during init/build to make sure `self.thing != things_to_deflate`

docs/notebooks/tutorials/deflation.ipynb
```diff
@@ -0,0 +1,1338 @@
...
```
f0uriest, line 1: <p>can you add a comment explaining to readers what this block is for and when to uncomment</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2723344597'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='7'/>

f0uriest, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    ntheta = 12  # only using M=4 for our equilibrium, so we don't need a large ntheta
</pre><p>m=5 used below</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2723344601'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='8'/>

f0uriest, line 1: <pre class="ql-syntax" spellcheck="false">Line #6.    eqname = f"QA__O_rho_r_{r}_fit.h5"
</pre><p>looks like this is unused</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2723344603'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='8'/>

f0uriest, line 1: <pre class="ql-syntax" spellcheck="false">Line #14.    )  # 5 5 6 actually ok...
</pre><p>can remove this comment</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2723344605'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='8'/>

f0uriest, line 1: <pre class="ql-syntax" spellcheck="false">Line #18.        if eqs_to_deflate is not None:
</pre><p>from the compute method of the deflation operator it looked like the intention was to allow you to basically just run it without this logic, since if things_to_deflate is all None then the deflated objective just acts like a the undeflated version?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2044/discussion/#comment-2723344606'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='9'/>

desc/objectives/_generic.py
dpanici: because we cannot have a conditional on the mask without it being static, and the mask changes if None or an object is passed in, so if the single mask is static and I pass [None, eq] then [eq, eq], it will recompile on the second time its called even though we went through all this pain to avoid it

desc/objectives/_generic.py
dpanici: It does not seem like it? at least Ive encountered no issues. I can probably make it static just to be safe though

desc/objectives/_generic.py
dpanici: ah yes you are right

dpanici: did not relize that was an option of prod

desc/objectives/_generic.py
dpanici: Yea the two different methods make it unclear what the best default value should be.

I would want it to be zero when it is not wrapping anything but 1 if it is wrapping something.

This logic is getting so confusing...

desc/objectives/_generic.py
dpanici: Hm yea I also tend to make this mistake. Probably can assert to make sure it is not the exact same thing, then  also use epsilon for stability of the division

docs/notebooks/tutorials/deflation.ipynb
dpanici: right that was an old version. removing

desc/objectives/_generic.py
f0uriest: i don't think you ever have conditionals on the mask in the compute function? it looks like its just used as the `where` arg to `prod`/`sum`?

desc/objectives/_generic.py
```diff
@@ -930,37 +951,39 @@ def compute(self, params, constants=None):
...
+        deflation_parameter = (
+            deflation_parameter * self._not_all_things_to_deflate_are_None
+        )
+
         if self._objective is not None:
             f = self._objective.compute(params)
+            # if wrapping an objective, but all things are None, make deflation do
+            # nothing when multiplying f, so here we add 1 to it as it is 0 right now
+            # if all things are None
+            deflation_parameter += 1.0
```
f0uriest: I think you might want this to be `+= jnp.invert(self._not_all_things_to_deflate_are_None)` that way it only adds 1 if all the deflated things are None

dpanici: yes you are right. the logic was getting confusing

desc/objectives/_generic.py
dpanici: Sorry to be clearer

I thought I required both because we can't have conditionals/we cannot convert bool to int etc during the jitted compute function. Though it seems with jnp.invert I can do float(jnp.invert...) just fine, I get no errors thrown.

The way I had originally tried was using conditionals but I changed it to just multiplying with masks to avoid the conditionals, which iirc could not be used to avoid recompilation.

YigitElma (dismissed):

desc/objectives/_generic.py
```diff
@@ -669,3 +683,308 @@ def compute(self, params, constants=None):
...
+        assert deflation_type in ["power", "exp"]
+        self._deflation_type = deflation_type
+        assert multiple_deflation_type in ["prod", "sum"]
+        self._multiple_deflation_type = multiple_deflation_type
+        self._single_shift = single_shift
+        self._objective = objective
+        if self._objective is not None:
+            assert isinstance(
+                self._objective, _Objective
+            ), "objective passed in must be an _Objective!"
```
ddudt: I prefer using our `errorif` util function for this

desc/objectives/_generic.py
```diff
@@ -669,3 +683,308 @@ def compute(self, params, constants=None):
...
+            assert len(objective.things) == 1
+            assert objective.things[0] == thing
+            name = "Deflated " + self._objective._name
+            self._units = self._objective._units
+            self._scalar = self._objective._scalar
+            self._coordinates = self._objective._coordinates
```
ddudt, line 861: Does this not cause problems when `self._objective = None`?

desc/objectives/_generic.py
ddudt: @dpanici I don't see `self._is_none_mask` get used anywhere in the `compute` function. Can we remove that array? Then this would be resolved.

desc/objectives/_generic.py
```diff
@@ -669,3 +683,308 @@ def compute(self, params, constants=None):
...
+            lower_bound_min = (
+                self._sigma
+                if self._single_shift
+                else self._sigma * np.sum(self._is_not_none_mask)
+            )
+            assert np.all(self._bounds[0] <= lower_bound_min), (
+                "Provided lower bound for deflation operator is too high compared "
+                f"to the minimum value of {lower_bound_min} it can take based off "
+                "of sigma, use a smaller lower bound"
+            )
```
ddudt: I prefer using `errorif` for consistent formatting throughout the code.

desc/objectives/_generic.py
```diff
@@ -669,3 +683,308 @@ def compute(self, params, constants=None):
...
+
+        # we use the where= to only count the non-None things in things_to_deflate
+        if self._multiple_deflation_type == "prod":
+            deflation_parameter = jnp.prod(
+                M_i, initial=1.0, where=self._is_not_none_mask
+            ) + self._sigma * (self._single_shift)
+        else:
+            deflation_parameter = jnp.sum(
+                M_i, where=self._is_not_none_mask, initial=0.0
+            ) + self._sigma * (self._single_shift)
```
ddudt, line 1003: Suggested change to be fancy and make the code more readable:
```
fun = jnp.prod if self._multiple_deflation_type == "prod" else jnp.sum
deflation_parameter = fun(
    M_i, initial=1.0, where=self._is_not_none_mask
) + self._sigma * (self._single_shift)
```

desc/objectives/_generic.py
```diff
@@ -669,3 +683,308 @@ def compute(self, params, constants=None):
...
+
+        # enforce deflation paremeter 0 here if every thing_to_deflate is None
+        deflation_parameter = (
+            deflation_parameter * self._not_all_things_to_deflate_are_None
+        )
```
ddudt: Suggested change:
```
deflation_parameter *= self._not_all_things_to_deflate_are_None
```

desc/objectives/_generic.py
```diff
@@ -669,3 +683,308 @@ def compute(self, params, constants=None):
...
+        this objective in a loop with a pre-determined number of iterations and adding
+        each result of the loop iterate to the things_to_deflate, it may trigger
+        recompilation of the objective's compute and jac/grad functions each time,
+        which is wasteful. You can instead pass in a list containing None elements
+        padding the list out to the max length it will attain. In this way, no
+        recompilations will be triggered, and the entire loop will be completed
+        much more quickly.
+        If all things_to_deflate are None, this objective has zero cost (if not
+        wrapping another objective) or simply returns the wrapped objective's
+        cost (if wrapping another objective)
```
ddudt, line 738: How common of a use case is the list of Nones? Because this is a lot of clunky logic to prevent recompilation for something that is happening outside the scope of this objective (in a higher level loop). If it is not a common use case and/or the recompilation time is not significant, then I vote to not worry about it. The code could be simplified a lot.

tests/test_objective_funs.py
```diff
@@ -3553,6 +3556,37 @@ def test_compute_scalar_resolution_mercier(self):
...
+            )
+            obj.build(verbose=0)
+            f[i] = obj.compute_scalar(obj.x())
+        # put res of self.eq back to original
+        res = self.res_array[0]
+        self.eq.change_resolution(
+            L_grid=int(self.eq.L * res),
+            M_grid=int(self.eq.M * res),
+            N_grid=int(self.eq.N * res),
+        )
```
ddudt: Instead of modifying the equilibrium and then reverting the changes back, it would be safer and cleaner to make a copy of `eq` at the beginning of this function and only modify that copy.

tests/test_objective_funs.py
```diff
@@ -3718,6 +3753,44 @@ def test_objective_no_nangrad_plasma_vessel(self):
         g = obj.grad(obj.x(eq, surf))
         assert not np.any(np.isnan(g)), "plasma vessel distance"
 
+    @pytest.mark.unit
+    def test_objective_no_nangrad_ForceBalanceDeflated(self):
```
ddudt: `ForceBalanceDeflated` isn't actually its own objective, right? I assume this was from an old version of the code

ddudt (changes requested):

desc/objectives/_generic.py
f0uriest: `initial` needs to be different for `sum` vs `prod`

desc/objectives/_generic.py
f0uriest: I think it's worth it. The whole point of deflation is that you call it in a loop multiple times, and in common cases the compilation time is a significant fraction of the overall solve.

dpanici:
@ddudt In the tutorial deflation notebook, for the QA NAE constrained. the savings on the deflation optimization is almost half. that one is only taken for 15 iterations so the cost is dominated by compilation, which is ~3s long, when the overall optimization once compiled takes only 3s. 

Even the QH optimization in that notebook, which is using auglag and thus has many iterations, has a 5-10s savings per optimization when recompilation is avoided, bringing the times down from 25-30s per optimization to 20s per optimization. 

While the notebook only runs like 3 outer iterations (i.e. only gets 3 solutions), in practice I've used these to get 20-30 solutions, so the savings would be even higher for real applications.

So while it is problem dependent, it can be a huge timesaver to smartly avoid recompilation using the existing functionality I have. For the user API it is pretty straightforward, so I am fine with how it is. I'll push back on the desire to remove this feature as it seems helpful and effective.

desc/objectives/_generic.py
dpanici: I could do something like that and also change initial but at that point it is only saving a couple lines, and I argue that having if else blocks are more readbale than a fun that changes. I will keep it as is

desc/objectives/_generic.py
dpanici: [See my comment on the PR](https://github.com/PlasmaControl/DESC/pull/2044#issuecomment-4026819826)

desc/objectives/_generic.py
dpanici: it's within the if block that ensure it is not none

YigitElma (dismissed):

YigitElma (approved):

desc/objectives/_generic.py
```diff
@@ -669,3 +683,337 @@ def compute(self, params, constants=None):
...
+            ) ** self._power + self._sigma * (not self._single_shift)
+        else:
+            M_i = jnp.exp(1 / (jnp.linalg.norm(diffs, axis=1) + eps)) + self._sigma * (
+                not self._single_shift
+            )
+
+        # we use the where= to only count the non-None things in things_to_deflate
+        if self._multiple_deflation_type == "prod":
+            deflation_parameter = jnp.prod(
+                M_i, initial=1.0, where=self._is_not_none_mask
```
YigitElma, line 998: Looks like this usage with not bool arrays is deprecated in jax 0.8.0. Some tests are failing due to that. Probably easier fix is to just convert `self._is_not_none_mask` to boolean

desc/objectives/_generic.py
```diff
@@ -669,3 +683,337 @@ def compute(self, params, constants=None):
...
+            ) + self._sigma * (self._single_shift)
+        else:
+            deflation_parameter = jnp.sum(
+                M_i, where=self._is_not_none_mask, initial=0.0
```
YigitElma, line 1002: This is also deprecated with new jax versions

YigitElma (changes requested):

YigitElma (dismissed):

ddudt (dismissed):

YigitElma (approved):

YigitElma:
@ddudt @dpanici This was ready before with 2 approvals, merge conflict caused the stale review

PR #2046, opened 2025-12-22
Fix bug in Kinetic normalization when pressure is present, add warning for this case to profile setters

### LINKED ISSUE #1995: Setting kinetic profiles does not remove pressure profile, and this leads to errors
If I say assign kinetic profiles to an eq that already has a pressure profile, we don't remove the pressure profile currently. This creates errors later if used in optimizations involving quantities reliant on the pressure normalization bc (I think)
- the normalization scales for the FixKinetic, in its build call to compute_scale_factors, see it has a pressure profile and so does not assign the kinetic normalizations
- then immediately after it attempts to grab the kinetic normalization scale, however it does not exist and so an error is thrown

We could either
- always compute both p scale and kinetic scale factors if those profiles are present
- or remove the pressure profile when kinetic profiles are assigned. I vaguely recall that we said we wanted to allow this for whatever reason, but I don't remember why.

example code:

```python
from desc.examples import get

eq = get("SOLOVEV")
eq.electron_density=PowerSeriesProfile()
eq.electron_temperature=PowerSeriesProfile()
eq.ion_temperature=PowerSeriesProfile()
eq.atomic_number=PowerSeriesProfile()
# eq.pressure=None # no error if we uncomment this line

eq.solve()

```

yields error
```python
KeyError                                  Traceback (most recent call last)
Cell In[2], line 10
      7 eq.ion_temperature=PowerSeriesProfile()
      8 eq.atomic_number=PowerSeriesProfile()
---> 10 eq.solve()

File [~/Research/DESC/desc/equilibrium/equilibrium.py:2280](http://localhost:8888/lab/tree/~/Research/DESC/desc/equilibrium/equilibrium.py#line=2279), in Equilibrium.solve(self, objective, constraints, optimizer, ftol, xtol, gtol, maxiter, x_scale, options, verbose, copy)
   2265 warnif(
   2266     self.N > self.N_grid or self.M > self.M_grid or self.L > self.L_grid,
   2267     msg="Equilibrium has one or more spectral resolutions "
   (...)   2271     + "to avoid this warning.",
   2272 )
   2273 errorif(
   2274     self.bdry_mode == "poincare",
   2275     NotImplementedError,
   2276     "Solving equilibrium with poincare XS as BC is not supported yet "
   2277     + "on master branch.",
   2278 )
-> 2280 things, result = optimizer.optimize(
   2281     self,
   2282     objective,
   2283     constraints,
   2284     ftol=ftol,
   2285     xtol=xtol,
   2286     gtol=gtol,
   2287     x_scale=x_scale,
   2288     verbose=verbose,
   2289     maxiter=maxiter,
   2290     options=options,
   2291     copy=copy,
   2292 )
   2294 return things[0], result

File [~/Research/DESC/desc/optimize/optimizer.py:222](http://localhost:8888/lab/tree/~/Research/DESC/desc/optimize/optimizer.py#line=221), in Optimizer.optimize(self, things, objective, constraints, ftol, xtol, gtol, ctol, x_scale, verbose, maxiter, options, copy)
    219 timer.start("Initializing the optimization")
    220 if not is_linear_proj:
    221     objective, nonlinear_constraint, x_scale = (
--> 222         get_combined_constraint_objectives(
    223             eq,
    224             constraints,
    225             objective,
    226             things,
    227             x_scale,
    228             self.method,
    229             method,
    230             verbose,
    231             options,
    232         )
    233     )
    234 else:
    235     nonlinear_constraint = None

File [~/Research/DESC/desc/optimize/optimizer.py:518](http://localhost:8888/lab/tree/~/Research/DESC/desc/optimize/optimizer.py#line=517), in get_combined_constraint_objectives(eq, constraints, objective, things, x_scale, opt_method, method, verbose, options)
    516     objective.build(verbose=verbose)
    517 if linear_constraint is not None and not linear_constraint.built:
--> 518     linear_constraint.build(verbose=verbose)
    519 if nonlinear_constraint is not None and not nonlinear_constraint.built:
    520     nonlinear_constraint.build(verbose=verbose)

File [~/Research/DESC/desc/backend.py:146](http://localhost:8888/lab/tree/~/Research/DESC/desc/backend.py#line=145), in execute_on_cpu.<locals>.wrapper(*args, **kwargs)
    143 @functools.wraps(func)
    144 def wrapper(*args, **kwargs):
    145     with jax.default_device(jax.devices("cpu")[0]):
--> 146         return func(*args, **kwargs)

File [~/Research/DESC/desc/objectives/objective_funs.py:339](http://localhost:8888/lab/tree/~/Research/DESC/desc/objectives/objective_funs.py#line=338), in ObjectiveFunction.build(self, use_jit, verbose)
    337         if verbose > 0:
    338             print("Building objective: " + objective.name)
--> 339         objective.build(use_jit=self.use_jit, verbose=verbose)
    340     self._dim_f += objective.dim_f
    341 if self._dim_f == 1:

File [~/Research/DESC/desc/objectives/linear_objectives.py:2233](http://localhost:8888/lab/tree/~/Research/DESC/desc/objectives/linear_objectives.py#line=2232), in FixElectronDensity.build(self, use_jit, verbose)
   2231 if self._normalize:
   2232     scales = compute_scaling_factors(eq)
-> 2233     self._normalization = scales["n"]
   2234 super().build(use_jit=use_jit, verbose=verbose)

KeyError: 'n'
```

dpanici:
Resolved #1995


desc/equilibrium/equilibrium.py
```diff
@@ -1719,6 +1726,16 @@ def pressure(self, new):
         self._pressure = ensure_consistent_profile_eq_resolution(
             self._pressure, self, name="pressure"
         )
+        has_kinetic = any([hasattr(self, name) for name in _kinetic_profile_names])
```
f0uriest: I think the checks for this should be `if self.electron_temperature is not None` etc, rather than just checking if the attribute exists (we check `is not None` in the compute method so should match that here)

desc/equilibrium/equilibrium.py
```diff
@@ -1771,6 +1788,16 @@ def electron_temperature(self, new):
             self._electron_temperature, self, name="electron temperature"
         )
 
+        warnif(
+            hasattr(self, "_pressure"),
```
f0uriest: same as above, and below

desc/objectives/normalization.py
```diff
@@ -45,9 +45,14 @@ def get_max_mode(basis, coeffs):
...
-            p0 = float(thing.pressure(0)[0])
-        else:
+        p0 = float(thing.pressure(0)[0])
+        if (
+            thing.electron_density is not None
+            and thing.electron_temperature is not None
+            and thing.atomic_number is not None
+            and thing.ion_temperature is not None
+        ):
+            # if kinetic profiles exist, then use them for "p" scale
```
f0uriest: did we decide if this was the right thing to do? in compute pressure is computed using the pressure profile if it exists, even if kinetic profiles are also there, so I could see wanting the normalization to match that (somewhat ill defined) behavior, and rely on the new warnings to make clear to the user that this behavior may lead to unexpected results

tests/test_equilibrium.py
```diff
@@ -462,6 +463,23 @@ def test_assigning_profile_iota_current():
...
 
 
+@pytest.mark.unit
+def test_assigning_profile_pressure_kinetic():
+    """Test assigning pressure to kinetic-constrained eq and vice-versa."""
+    eq = get("HELIOTRON")  # pressure-constrained
+    with pytest.warns(UserWarning, match="existing pressure"):
+        eq.electron_density = PowerSeriesProfile()
+
+    eq = get("reactor_QA")  # pressure-constrained
```
f0uriest: isnt this kinetic constrained?

desc/objectives/normalization.py
dpanici: change this to use pressure for pressure scale if both pressure/kinetic profiles exist, and throw a warning in this case

f0uriest (dismissed):

desc/vmec_utils.py
```diff
@@ -695,7 +695,8 @@ def make_boozmn_output(  # noqa: C901
         np.array(
             [" " * (100 - len(version_str))],
             "S" + str(file.dimensions["dim_00100"].size),
-        )
+        ),
+        encoding="ascii",
```
YigitElma: was this related to #2085 ?

YigitElma (dismissed):

desc/vmec_utils.py
dpanici: Ah yea I likely had accidentally added that here

YigitElma (dismissed):

desc/objectives/normalization.py
```diff
@@ -45,16 +46,32 @@ def get_max_mode(basis, coeffs):
...
-        if thing.pressure is not None:
-            p0 = float(thing.pressure(0)[0])
-        else:
+        has_kinetic = False
+        if (
+            thing.electron_density is not None
+            and thing.electron_temperature is not None
+            and thing.atomic_number is not None
+            and thing.ion_temperature is not None
+        ):
```
ddudt: We should split this up like:

- if thing has either temperature profile, add the temperature scaling
- if thing has a density profile, add the density scaling
- if thing has a pressure profile, add the pressure scaling

And have them all be independent of each other. It might not matter now, but in #2090 some but not all of the kinetic profiles can be `None` which would break this check.

desc/objectives/normalization.py
```diff
@@ -45,16 +46,32 @@ def get_max_mode(basis, coeffs):
...
+        if thing.pressure is not None:
+            p0 = float(thing.pressure(0)[0])
+            warnif(
+                has_kinetic,
+                UserWarning,
+                "Equilibrium has both kinetic and pressure profiles assigned to it."
+                " By default, the pressure profile will be used for the normalization "
+                "and the computation of pressure, which may lead to unexpected "
+                "results when kinetic profiles are also present.",
+            )
```
ddudt, line 79: I think this warning is a little unnecessary since a similar warning will have already been thrown when the equilibrium is initialized or the profiles were assigned.

desc/vmec_utils.py
```diff
@@ -695,7 +695,7 @@ def make_boozmn_output(  # noqa: C901
         np.array(
             [" " * (100 - len(version_str))],
             "S" + str(file.dimensions["dim_00100"].size),
-        )
```
ddudt, line 698: Is this a necessary change?

tests/test_bootstrap.py
```diff
@@ -775,13 +775,19 @@ def test_Redl_sfincs_tokamak_benchmark(self):
...
+            )
+        with pytest.warns(UserWarning, match="existing pressure"):
+            eq.electron_temperature = PowerSeriesProfile(
+                8e3 * np.array([1, -1]), modes=[0, 2]
+            )
+        with pytest.warns(UserWarning, match="existing pressure"):
+            eq.ion_temperature = PowerSeriesProfile(
+                8e3 * np.array([1, -1]), modes=[0, 2]
+            )
+        eq.pressure = None
```
ddudt, line 778: Instead of catching all these warnings, you can set `eq.pressure = None` before assigning the kinetic profiles.

ddudt: Also we don't need the test coverage for these warnings here because they are already covered in `test_configuration.py`

tests/test_objective_funs.py
```diff
@@ -3247,11 +3247,15 @@ def test_compute_scalar_resolution_plasma_vessel(self):
...
-
+        with pytest.warns(UserWarning, match="is being assigned"):
+            eq.electron_density = PowerSeriesProfile([1e19, 0, -1e19])
+        with pytest.warns(UserWarning, match="is being assigned"):
+            eq.electron_temperature = PowerSeriesProfile([1e3, 0, -1e3])
+        with pytest.warns(UserWarning, match="is being assigned"):
+            eq.ion_temperature = PowerSeriesProfile([1e3, 0, -1e3])
+        with pytest.warns(UserWarning, match="is being assigned"):
+            eq.atomic_number = 1.0
+        eq.pressure = None
```
ddudt, line 3271: See my previous comment. You can set `eq.pressure = None` before assigning the kinetic profiles to avoid catching all these warnings. Same for the other changes in this file.

ddudt (changes requested):

desc/objectives/normalization.py
dpanici: Have we decided that #2090 is the way we will be doing this? I mean right now, the profile behavior is not well-defined when there is a subset. It makes little sense for these to be independent. Even in that PR change, I think that there should still be some sort of warning, or you can make the warning specifically consider the case you mention (which I think is that ion density profile may exist but no Zeff or vice versa?)

desc/vmec_utils.py
dpanici: this was just my removing an unneeded change before. I can remove this

desc/vmec_utils.py
```diff
@@ -695,7 +695,7 @@ def make_boozmn_output(  # noqa: C901
...
+        ),
```
dpanici:
```suggestion
        )
```

desc/objectives/normalization.py
dpanici: I think it is necessary as it may be unclear which is used for normalization, like if you are using pressure and kinetic profiles but one is inconsistent with the other. I want to make it very clear what is happening in the case of both pressure and kinetic assigned

desc/objectives/normalization.py
dpanici: I think this is the most I can split up without making the logic more complex, let me know what you think

YigitElma (dismissed):

ddudt (dismissed):

YigitElma (approved):

PR #2060, opened 2026-01-13
Fix docstring example for ExternalObjective

f0uriest:
user function needs to return a 1d array of size dim_f, not a scalar


ddudt (approved):

f0uriest:
To ensure backwards compatibility we could also temporarily add a custom getattr method to external/__init__

YigitElma (approved):

PR #2061, opened 2026-01-14
Fix error in unmarked test script

dpanici:
the `head -n -2` was incorrect (`-2` should just be `2`), the script was silently failing locally for me (on macOS zsh) and causing it to not actually detect any unmarked tests even if they were present.


f0uriest:
Are there any tests we forgot to mark that weren't caught?

YigitElma:
On my system, both versions of the script can catch the missing tests (if I delete some existing ones), this might be OS dependent maybe?

PR #2067, opened 2026-01-26
Fix `np` bug in `LinearConstraintProjection`

YigitElma:
- Fixes the bug in `factorize_linear_constraints` for numpy array `x_scale`
- Removes any change applied to the user-supplied `x_scale`. For example, if I want to give a scale to lambda (to make it have a similar magnitude as R and Z) based on the major radius etc. currently I cannot because the check x<100 will override that scale.


dpanici (changes requested):
Test

dpanici (approved):

PR #2072, opened 2026-01-30
Fix autodiff at bifurcation

unalmis:
[spectral_reverse_bounce_supplement.pdf](https://github.com/user-attachments/files/25376486/spectral_reverse_bounce_supplement.pdf)


desc/integrals/_interp_utils.py
```diff
@@ -244,6 +251,10 @@ def _filter_distinct(r, sentinel, eps):
 # TODO (#1388): Move to interpax.
 
 
+@partial(
+    jax.custom_vjp,
```
f0uriest: custom_vjp can't be forward mode differentiated, is it possible to use custom_jvp instead? (which works in both forward and reverse mode)

PR #2076, opened 2026-02-03
Add NeutronWallLoading objective

tokamaster:
This PR adds a `NeutronWallLoading` objective that computes DT neutron wall loading on a constant-offset surface derived from the equilibrium boundary, enabling it as a target for equilibrium/wall optimization.

Reference:
https://iopscience.iop.org/article/10.1088/1741-4326/ac6a67/


f0uriest:
Why was this closed?

YigitElma:
> Why was this closed?

Oh I think when I merged #2016 and deleted the branch, it automatically closed this one since this was pointing to that. Should we change the target branch to `master` now? I restored the other branch just in case.

desc/objectives/_geometry.py
```diff
@@ -827,6 +830,360 @@ def fun(plasma_pts_at_zeta_plane, surface_pts_at_zeta_plane):
...
+        "_fuel",
+        "_offset",
+        "_surface_data_keys",
+    ]
+
+    _coordinates = "rtz"
+    _units = "(MW/m^2)"
+    _print_value_fmt = "Neutron wall loading: "
+
+    def __init__(
```
dpanici, line 901: We should make this the most general one (where either surface or eq may be free), can you update it to the one with the more general logic?

PR #2078, opened 2026-02-04
Try to prevent tk window from popping up

### LINKED ISSUE #2077: Importing desc.plotting opens a blank GUI window/canvas in interactive sessions
### Bug description
Importing `desc.plotting` in an interactive Python session opens a blank GUI window/canvas immediately. This import-time side effect causes problems for downstream code that imports DESC for non-plotting functionality (we currently work around it with lazy imports).

### Steps to reproduce
1. Start a fresh interactive session (e.g. `python` or `ipython`).
2. Run:
   ```python
   from desc import plotting as desc_plotting
   # (also reproducible with: import desc.plotting)

### Expected behavior
Importing `desc.plotting` should not create any GUI windows. Plot windows should only appear when a plotting function is called.

### Suspected cause (import-time GUI creation)

`desc/plotting.py` appears to create a Tk root window at import time in order to compute DPI:

```python
import tkinter
...
try:
    dpi = tkinter.Tk().winfo_fpixels("1i")
except tkinter._tkinter.TclError:
    dpi = 72
```

This likely results in a visible empty Tk / Matplotlib window, since the Tk root is created on import and not withdrawn/destroyed. (There are also other global matplotlib rcParam changes executed at import time.)

dpanici:
Keep tk window from opening when `desc.plotting` is imported.

Resolves #2077

### DIFF
--- a/desc/plotting.py
+++ b/desc/plotting.py
@@ -2,7 +2,6 @@
 
 import inspect
 import numbers
-import tkinter
 import warnings
 
 import matplotlib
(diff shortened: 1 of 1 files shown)


desc/plotting.py
```diff
@@ -99,7 +99,10 @@
 rcParams["figure.figsize"] = (6, 4)
 
 try:
```
YigitElma, line 101: This fixes the problem on my laptop, but can we just remove this whole together. Looks like `matplotlib.rcParams` has a default value of 100 for the dpi. Do we really need to set it?

dpanici: Probably not. let's see how many tests fail if we unset it...

YigitElma (dismissed):

YigitElma (approved):
It resolves the problem on WSL

Issue #1196, opened 2024-08-16
Turbulence metrics for ITG/TEM


f0uriest:
ITG metrics:

- [ ] Goodman 2024, pretty simple just based on `grad(rho)` in bad curvature regions 
[goodman2024qi.pdf](https://github.com/user-attachments/files/16632448/goodman2024qi.pdf)

- [ ] Mynick et al, several papers, a bit more fancy than Goodman's, has some fudge factors that were obtained by fits to GENE 
[mynick2006transportoptimization.pdf](https://github.com/user-attachments/files/16632456/mynick2006transportoptimization.pdf)
[mynick2010optimizedstellaratortransport.pdf](https://github.com/user-attachments/files/16632457/mynick2010optimizedstellaratortransport.pdf)
[mynick2011reducingturbulence.pdf](https://github.com/user-attachments/files/16632458/mynick2011reducingturbulence.pdf)
[mynick2014turbulentoptimization.pdf](https://github.com/user-attachments/files/16632459/mynick2014turbulentoptimization.pdf)

TEM metrics:

- [ ] Proll 2015 - might need the bounce integral stuff first, but otherwise seems doable 
[proll2015temoptimization.pdf](https://github.com/user-attachments/files/16632471/proll2015temoptimization.pdf)

- [ ] Mackenbach 2022,2023 - also needs bounce integrals but otherwise not too bad, i think its been implemented in simsopt already 
[mackenbach2022availableenergy.pdf](https://github.com/user-attachments/files/16632486/mackenbach2022availableenergy.pdf)
[mackenbach2023availableenergy.pdf](https://github.com/user-attachments/files/16632487/mackenbach2023availableenergy.pdf)

dpanici:
https://arxiv.org/abs/2506.22166
new paper

PR #2080, opened 2026-02-06
Add ITG turbulence proxy and neural network objectives

byoungj:
Related to #1196

Adds two new objective functions for ITG turbulence optimization: an analytical proxy (`ITGProxy`) and a neural network ensemble proxy (`NNITGProxy`), based on Landreman et al. 2025 ("How does ITG turbulence depend on magnetic geometry?").

### GX Geometric Features
- Reference quantities (`gx_B_reference`, `gx_L_reference`), magnetic field strength (`gx_bmag`), gradient features (`gx_gds2`, `gx_gds21_over_shat`, `gx_gds22_over_shat_squared`), drift features (`gx_gbdrift`, `gx_cvdrift`, `gx_gbdrift0_over_shat`), and parallel gradient (`gx_gradpar`)
- Sign convention corrections for toroidal flux direction and pressure gradient terms

### ITGProxy Objective
- Analytical proxy: `f_Q = mean([Theta(cvdrift) + 0.2] * |grad(x)|^3 / B)`
- Uses smooth sigmoid approximation for differentiability during optimization

### Flux Tube Geometry Utilities
- Arclength computation via `gradpar` with uniform resampling to `z in [-pi, pi)`
- Brent's method solver for poloidal turns matching a target flux tube length (I tried making it differentiable via desc.backend.root_scalar within .compute())
- Dynamic `poloidal_turns`(fixed toroidal turns) that adapts as iota changes during optimization

### NNITGProxy Objective
- Pure-JAX re-implementation of the CyclicInvariantNet CNN architecture (circular convolutions, BatchNorm inference, max/global-avg pooling)
- Loads pre-trained PyTorch `.pth` weights and converts to JAX arrays (TODO: will have to make this not Torch dependent.
- Ensemble inference with configurable `top_k` model selection from DeepHyper results
- JIT-compiled forward pass
- Batched grid processing: single `map_coordinates` + `compute` call for all `(rho, alpha, zeta)` points
- Optional features: `return_per_alpha` for per-field-line analysis, `return_std` for ensemble uncertainty, `return_signals` for intermediate GX feature debugging

### Known Limitations/TODO
- The quantities calculated here are the same as in _stability with normalization / sign factors, so this can be consolidated. But for debugging purposes, I have created a separate compute() quantities for now.
- PyTorch dependency for weight loading: Currently requires `torch` to deserialize `.pth` files. Plan is to serialize weights to a torch-free format (e.g. `.npz`) in a separate conversion repo, so DESC only needs JAX at runtime.
- TEM metrics (Proll 2015, Mackenbach 2022/2023) are not included in this PR — they depend on bounce integral infrastructure and will be addressed separately under #1196.

### DIFF
--- a/desc/compute/_turbulence.py
+++ b/desc/compute/_turbulence.py
@@ -1,0 +1,88 @@
+"""Compute functions for ITG turbulence proxies and related quantities.
+
+This module provides GX geometric coefficients and ITG turbulence proxies
+for gyrokinetic turbulence prediction. The implementation follows the
+conventions established in Landreman et al. 2025 (arXiv:2502.11657).
+
+Notes
+-----
+Reference quantities use the following normalizations:
+- B_reference = 2|ψ_b|/a² where ψ_b is the boundary toroidal flux
+- L_reference = a (minor radius)
+
+Some quantities may have singularities at the magnetic axis (ρ=0).
+Objectives using these quantities should use ρ > 0.
+"""
+
+from desc.backend import jnp
+
+from .data_index import register_compute_fun
+
+
+# =============================================================================
+# Reference Quantities
+# =============================================================================
+
+
+@register_compute_fun(
+    name="gx_B_reference",
+    label="B_{\\mathrm{ref}}",
+    units="T",
+    units_long="Tesla",
+    description="GX reference magnetic field: B_ref = 2|ψ_b|/a²",
+    dim=0,
+    params=["Psi"],
+    transforms={},
+    profiles=[],
+    coordinates="",
+    data=["a"],
+)
+def _gx_B_reference(params, transforms, profiles, data, **kwargs):
+    # ψ_b = Psi/(2π) is the boundary toroidal flux
+    psi_b = jnp.abs(params["Psi"]) / (2 * jnp.pi)
+    data["gx_B_reference"] = 2 * psi_b / data["a"] ** 2
+    return data
+
+
+@register_compute_fun(
+    name="gx_L_reference",
+    label="L_{\\mathrm{ref}}",
+    units="m",
+    units_long="meters",
+    description="GX reference length: L_ref = a (minor radius)",
+    dim=0,
+    params=[],
+    transforms={},
+    profiles=[],
+    coordinates="",
+    data=["a"],
+)
+def _gx_L_reference(params, transforms, profiles, data, **kwargs):
+    data["gx_L_reference"] = data["a"]
+    return data
+
+
+# =============================================================================
+# Normalized Magnetic Field
+# =============================================================================
+
+
+@register_compute_fun(
+    name="gx_bmag",
+    label="|\\mathbf{B}|/B_{\\mathrm{ref}}",
+    units="~",
+    units_long="dimensionless",
+    description="Normalized magnetic field magnitude |B|/B_ref for GX",
+    dim=1,
+    params=["Psi"],
+    transforms={},
+    profiles=[],
+    coordinates="rtz",
+    data=["|B|", "a"],
+)
+def _gx_bmag(params, transforms, profiles, data, **kwargs):
+    # B_ref = 2|ψ_b|/a² where ψ_b = Psi/(2π)
+    psi_b = jnp.abs(params["Psi"]) / (2 * jnp.pi)
+    B_ref = 2 * psi_b / data["a"] ** 2
+    data["gx_bmag"] = data["|B|"] / B_ref
+    return data
--- a/desc/compute/_turbulence.py
+++ b/desc/compute/_turbulence.py
@@ -415,3 +415,95 @@ def compute_arclength_via_gradpar(gradpar, theta_pest):
         )
 
     return arclength
+
+
+def resample_to_uniform_arclength(arclength, data, npoints_out):
+    """Resample field line data from non-uniform to uniform arclength spacing.
+
+    This function takes data computed on a theta_PEST grid (with non-uniform
+    arclength spacing) and resamples it to a uniform arclength grid in [-pi, pi).
+    This matches the GX geometry conventions used for CNN training data.
+
+    Parameters
+    ----------
+    arclength : ndarray, shape (npoints_in,) or (npoints_in, num_alpha)
+        Cumulative arclength at each input point, as computed by
+        ``compute_arclength_via_gradpar``.
+    data : ndarray, shape (nfeatures, npoints_in) or (nfeatures, npoints_in, num_alpha)
+        Data to resample. First axis is features (e.g., 7 GX coefficients).
+    npoints_out : int
+        Number of output points. Output z will be uniformly spaced in [-pi, pi)
+        with the periodic endpoint excluded.
+
+    Returns
+    -------
+    z_uniform : ndarray, shape (npoints_out,)
+        Uniform arclength coordinates in [-pi, pi).
+    data_uniform : ndarray, same shape as data with npoints_in -> npoints_out
+        Resampled data on uniform arclength grid.
+
+    Notes
+    -----
+    The transformation maps the cumulative arclength [0, L] to z in [-pi, pi).
+    This normalization ensures the CNN sees features at evenly-spaced physical
+    distances along the field line, regardless of the underlying magnetic
+    geometry.
+
+    Requires ``interpax`` package for cubic interpolation.
+
+    See Also
+    --------
+    compute_arclength_via_gradpar : Compute cumulative arclength from gradpar.
+
+    Examples
+    --------
+    >>> # Compute arclength from gradpar
+    >>> arclength = compute_arclength_via_gradpar(gradpar, theta_pest)
+    >>> # Stack 7 GX features
+    >>> signals = jnp.stack([bmag, gbdrift, cvdrift, ...])  # shape (7, npoints)
+    >>> # Resample to 96 points (CNN input resolution)
+    >>> z, signals_uniform = resample_to_uniform_arclength(arclength, signals, 96)
+    """
+    from interpax import interp1d
+
+    # Handle 1D and 2D cases
+    if arclength.ndim == 1:
+        # Single field line: arclength shape (npoints,), data shape (nfeatures, npoints)
+        L = arclength[-1]
+        z_orig = arclength * (2 * jnp.pi / L) - jnp.pi
+        z_uniform = jnp.linspace(-jnp.pi, jnp.pi, npoints_out + 1)[:-1]
+
+        nfeatures = data.shape[0]
+        data_uniform = jnp.stack(
+            [
+                interp1d(z_uniform, z_orig, data[i], method="cubic")
+                for i in range(nfeatures)
+            ]
+        )
+    else:
+        # Multiple field lines: arclength shape (npoints, num_alpha)
+        # data shape (nfeatures, npoints, num_alpha)
+        num_alpha = arclength.shape[1]
+        z_uniform = jnp.linspace(-jnp.pi, jnp.pi, npoints_out + 1)[:-1]
+
+        def resample_one_alpha(arc_1d, data_2d):
+            """Resample a single field line."""
+            L = arc_1d[-1]
+            z_orig = arc_1d * (2 * jnp.pi / L) - jnp.pi
+            return jnp.stack(
+                [
+                    interp1d(z_uniform, z_orig, data_2d[i], method="cubic")
+                    for i in range(data_2d.shape[0])
+                ]
+            )
+
+        # Process each alpha
+        data_uniform = jnp.stack(
+            [
+                resample_one_alpha(arclength[:, a], data[:, :, a])
+                for a in range(num_alpha)
+            ],
+            axis=2,
+        )
+
+    return z_uniform, data_uniform
--- a/desc/compute/_turbulence.py
+++ b/desc/compute/_turbulence.py
@@ -753,6 +753,41 @@ def _global_avg_pool_1d(x):
     return jnp.mean(x, axis=-1)
 
 
+def _batch_norm_1d(x, gamma, beta, running_mean, running_var, eps=1e-5):
+    """Apply BatchNorm1d in inference mode (using running statistics).
+
+    Uses pre-computed running mean and variance rather than batch statistics,
+    making the output deterministic and independent of other samples in the batch.
+
+    Parameters
+    ----------
+    x : jax.Array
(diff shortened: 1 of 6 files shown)


desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+        solve_length_at_compute=False,
+        name="NN ITG Proxy",
+        jac_chunk_size=None,
+    ):
+        if target is None and bounds is None:
+            target = 0
+
+        # Check for torch dependency (required for loading .pth model weights)
+        try:
+            import torch  # noqa: F401
```
YigitElma, line 354: I couldn't check the whole code yet, but as I understand it, user needs to give the model weights as an extra file to be able to use this objective, right? If the objective requires this extra information which DESC doesn't offer, I would vote to move this objective to `desc.external`. Currently, this extra dependency causes some generic tests to fail, but it is easier to prevent it for things in `desc.external`

YigitElma: Alternatively, it looks like the model parameters are publicly available on https://zenodo.org/records/14867777, is it feasible to have them in DESC in a way that JAX can import directly? The data `.tar` file is pretty large, but how much the model parameters take in total?

YigitElma (commented):
Some questions before reviewing the rest

desc/objectives/_turbulence.py
byoungj: Totally! my plan is to remove this before merge to avoid adding more dependencies within DESC.

desc/compute/_turbulence.py
```diff
@@ -0,0 +1,1126 @@
...
+    profiles=[],
+    coordinates="",
+    data=["ITG proxy integrand"],
+)
+def _itg_proxy(params, transforms, profiles, data, **kwargs):
+    """Compute ITG proxy from Landreman et al. 2025.
+
+    f_Q = mean((sigmoid(cvdrift) + 0.2) * |grad_x|^3 / B)
+    """
+    data["ITG proxy"] = jnp.mean(data["ITG proxy integrand"])
```
f0uriest, line 346: is a simple mean the correct thing here? IE, should it really be a flux surface average? and should this return a value per surface, or per fieldline or 1 for the whole volume etc?

f0uriest: I think in matts paper its the mean over a fieldline right? but in theory it could also be computed on a surface, which would eliminate the need to map to field aligned coordinates. I think that's what's usually done for the other "grad(r) in bad curvature" type objectives

desc/compute/_turbulence.py
```diff
@@ -0,0 +1,1126 @@
...
+    # dl/dθ_PEST = 1 / |gradpar| (for normalized arclength)
+    dl_dtheta = 1.0 / jnp.abs(gradpar)
+
+    # Trapezoidal integration
+    dtheta = theta_pest[1] - theta_pest[0]
+
+    # Handle both 1D and 2D cases
+    if dl_dtheta.ndim == 1:
+        integrand_half = 0.5 * (dl_dtheta[1:] + dl_dtheta[:-1])
+        arclength = dtheta * jnp.concatenate(
```
f0uriest, line 402: can use https://quadax.readthedocs.io/en/latest/_api/quadax.cumulative_trapezoid.html#quadax.cumulative_trapezoid for this, or cumulative simpson if you want a bit more accuracy

f0uriest: can also vmap to handle the multiple fieldline case

desc/compute/_turbulence.py
```diff
@@ -0,0 +1,1126 @@
...
+    if arclength.ndim == 1:
+        # Single field line: arclength shape (npoints,), data shape (nfeatures, npoints)
+        L = arclength[-1]
+        z_orig = arclength * (2 * jnp.pi / L) - jnp.pi
+        z_uniform = jnp.linspace(-jnp.pi, jnp.pi, npoints_out + 1)[:-1]
+
+        nfeatures = data.shape[0]
+        data_uniform = jnp.stack(
+            [
+                interp1d(z_uniform, z_orig, data[i], method="cubic")
```
f0uriest, line 475: `interp1d` can handle extra trailing dimensions the data, so you don't need this loop here

desc/compute/_turbulence.py
```diff
@@ -0,0 +1,1126 @@
...
+                for i in range(nfeatures)
+            ]
+        )
+    else:
```
f0uriest, line 479: I think it should also be possible to use vmap here to get rid of these loops and conditionals

desc/compute/_turbulence.py
```diff
@@ -0,0 +1,1126 @@
...
+    Parameters
+    ----------
+    length_fn : callable
+        Function that takes poloidal_turns (float) and returns the flux tube
+        length. This typically involves evaluating gradpar along a field line
+        and integrating.
+    target_length : float
+        Target flux tube length in units of L_ref.
+    x0_guess : float, optional
+        Initial guess for poloidal turns. Default 1.0. The solver uses a
```
f0uriest, line 523: wasn't most of matts original data using ~2 turns? would that be a better default?

desc/compute/_turbulence.py
```diff
@@ -0,0 +1,1126 @@
...
+    >>> def length_fn(poloidal_turns):
+    ...     # Compute gradpar for this many poloidal turns
+    ...     theta_pest = np.linspace(
+    ...         -np.pi * poloidal_turns, np.pi * poloidal_turns, 1001
+    ...     )
+    ...     gradpar = compute_gradpar_for_field_line(eq, rho, alpha, theta_pest)
+    ...     return np.abs(np.trapezoid(1.0 / gradpar, theta_pest))
+    >>> poloidal_turns = solve_poloidal_turns_for_length(length_fn, target_length=75.4)
+    """
+    from scipy.optimize import brentq
```
f0uriest, line 562: does this function ever need to be jit compiled or differentiated? ie, is it used every time the objective is called? if so can use `desc.backend.root_scalar`

desc/compute/_turbulence.py
```diff
@@ -0,0 +1,1126 @@
...
+            f"Could not find poloidal_turns for target_length={target_length}. "
+            f"Bracket [{bracket_lo:.2f}, {bracket_hi:.2f}] gives lengths "
+            f"[{length_lo:.2f}, {length_hi:.2f}]. Try adjusting x0_guess."
+        ) from e
+
+    return poloidal_turns
+
+
+# =============================================================================
+# CNN Layer Primitives for NNITGProxy
```
f0uriest, line 587: instead of implementing all of these layers ourselves can we use an existing jax library like equinox?

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+
+    Lower values indicate configurations with reduced ITG turbulence.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium to be optimized.
+    rho : float or array-like
+        Flux surface(s) to evaluate on.
+    alpha : float or array-like, optional
```
f0uriest, line 37: see previous comment about whether this can be done over a surface rather than a fieldline

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+        )
+
+        # Average over field line to get per-flux-surface values
+        integrand = data["ITG proxy integrand"]
+        num_rho = len(constants["rho"])
+        num_alpha = len(constants["alpha"])
+        num_zeta = len(constants["zeta"])
+
+        integrand_reshaped = integrand.reshape(num_zeta, num_rho, num_alpha)
+        f_Q = jnp.mean(integrand_reshaped, axis=(0, 2))
```
f0uriest, line 235: might make more sense to not reduce this here so the optimizer sees each surface/fieldline?

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+    solve_poloidal_turns_for_length,
+)
+from desc.compute.utils import _compute as compute_fun
+from desc.grid import Grid, LinearGrid
+from desc.utils import Timer, setdefault
+
+from .objective_funs import _Objective, collect_docs
+
+
+class ITGProxy(_Objective):
```
f0uriest, line 21: Probably want a more descriptive name for this, to differentiate from other proxies people have come up with. Maybe `LandremanITGProxyAnalytical` or something? 

Alternatively, we could have this be a generic class for other simple analytical formula and have a `method` argument for eg `Landreman`, `Goodman`, `Mynick`...

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    rho : float or array-like
+        Flux surface(s) to evaluate on. Must be specified (no default).
+    alpha : float or array-like, optional
+        Field line labels to evaluate. Values should be in [0, 2pi).
+        Default is alpha=0.
+    npoints : int, optional
```
f0uriest, line 269: isn't this fixed by the model?

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+        Number of points along the field line for CNN input. Grid is periodic
+        with endpoint excluded (i.e., uniform in [-π, π)). Default is 96.
+    nz_internal : int, optional
+        Number of internal grid points for accurate arclength computation.
+        Default is 1001.
+    target_flux_tube_length : float, optional
+        Target flux tube length. Default is 75.4 (= 4*pi*6).
+    a_over_LT : float, optional
```
f0uriest, line 277: might want to also have the option to use the actual a/LT from the equilibrium if it has kinetic profiles (and they could be changing during optimization etc)

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+        num_alpha = len(constants["alpha"])
+        num_zeta = len(constants["zeta"])
+
+        integrand_reshaped = integrand.reshape(num_zeta, num_rho, num_alpha)
+        f_Q = jnp.mean(integrand_reshaped, axis=(0, 2))
+
+        return f_Q
+
+
+class NNITGProxy(_Objective):
```
f0uriest, line 240: see previous comment about a more specific name, I'm guessing this won't be the only NN turbulence proxy we add.

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+
+        if len(result) == 1:
+            return result[0]
+        return tuple(result)
+
+    def compute(
+        self,
+        params,
+        constants=None,
+        return_signals=False,
```
f0uriest, line 815: in order to keep the API consistent with other objectives I'd prefer to move these args to another helper function and wrap that here, eg:

```python
def _compute_with_extras(self, params, return_signals,...):
    # blah

def compute(self, params, constants):
    return self._compute_with_extras(params, constants, False, ...)

File: desc/objectives/_turbulence.py [hunk 16]
```diff
@@ -0,0 +1,1179 @@
...
+            If True, also return ensemble std of predictions. Requires ensemble
+            model (not single model). The std represents model uncertainty.
+            Default False.
+
+        Returns
+        -------
+        Q : ndarray
+            Predicted heat flux for each flux surface.
+            Shape (num_rho,) if return_per_alpha=False (default).
+            Shape (num_rho, num_alpha) if return_per_alpha=True.
```
f0uriest, line 844: This should always return a 1d array for it to play nice with other desc objectives and optimizers, so this may need to be flattened

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+        # Using zeta (toroidal angle) parameterization for dynamic iota adaptation
+        # =====================================================================
+
+        # Uniform zeta offset (same for all rho/alpha)
+        # This keeps toroidal extent fixed; poloidal extent adapts with iota
+        zeta_offset = jnp.linspace(
+            -jnp.pi * toroidal_turns, jnp.pi * toroidal_turns, nz
+        )
+
+        # Create meshgrid: shapes will be (nz, num_rho, num_alpha)
```
f0uriest, line 936: recommend just using `jnp.meshgrid` for this

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+        # =====================================================================
+        # RESAMPLING AND CNN INFERENCE
+        # Loop over rho (arclength varies per rho due to different iota)
+        # =====================================================================
+
+        Q_all_per_alpha = []
+        Q_std_all_per_alpha = [] if return_std else None
+        all_signals = [] if return_signals else None
+
+        for i_rho in range(num_rho):
```
f0uriest, line 994: can this be vmapped?

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+            coords_pest,
+            inbasis=("rho", "theta_PEST", "zeta"),
+            outbasis=("rho", "theta", "zeta"),
+            params=params,
+            tol=1e-10,
+            maxiter=50,
+        )
+        theta_desc_flat = desc_coords[:, 1]
+        # Fix theta wrapping for continuity
+        theta_desc_flat = theta_desc_flat + 2 * jnp.pi * jnp.round(
```
f0uriest, line 637: I think you can just do `theta_desc_flat = theta_desc_flat % (2*jnp.pi)`?

desc/objectives/_turbulence.py
```diff
@@ -0,0 +1,1179 @@
...
+        a_over_Ln,
+        models,
+        gx_keys,
+        feature_names,
+        constants,
+        return_signals,
+        return_per_alpha,
+        return_std=False,
+    ):
+        """Compute with exact flux tube length solving per rho.
```
f0uriest, line 1055: I think there might be a more clever way of doing this. 

We can probably get a reasonable upper bound for the number of turns for a given length just based on eg `a`, `R0` etc, or by doing this root finding once during build and then taking 2x that or something. We compute all the data along this longer fieldline, then when we resample to a uniform grid we only use up to the length we need. So no extra rootfinding necessary every time the objective gets called. The only downside would be, for a fixed resolution, slightly worse spacing between points, but assuming the bound is reasonably tight that shouldn't be too bad, plus extra points are relatively cheap compared to iterative root finding.

PR #2081, opened 2026-02-07
Adding support for umbilic curves

singh-jaydeep:
This PR is meant to update the [umbilic curve PR](https://github.com/PlasmaControl/DESC/pull/819) and address some of the TODOs listed there. The changes are based on my understanding of the paper/original PR, so if anything isn't quite right I can revert it.

Overview:
- Cosmetic changes (update docstrings, make variables uniform, removing some unused functions)
- Rename base `UmbilicCurve` class to `FluxSurfaceCurve`, allowing for future representations of curves which by construction lie in a flux surface. Subclasses include `FourierUmbilicCurve`, which is the particular parameterization of umbilic curves used in the original paper.
- Add a parameter `m_umbilic` to `FourierUmbilicCurve`, so that the class can represent all curves in the original paper.
- Rename `NFP_umbilic_factor` to `n_umbilic` when used internally in `FourierUmbilicCurve`, and to `N_scaling` when used externally in basis and grid classes. Wanted a name which doesn't require knowing about umbilic curves when just working with bases or grids. 
- Separate compute functions for some of the operations happening in the `UmbilicHighCurvature` objective.
- Add tests for the curve class and objective.

Notes:
- when we update an i/o attribute, is there a way to regenerate the test examples so it doesn't throw warnings when I run the tests? To avoid those warnings I didn't include `N_scaling` in i/o attributes for grid and basis, and left if for the curve itself to update these in its `_set_up()`.

### DIFF
--- a/desc/basis.py
+++ b/desc/basis.py
@@ -479,8 +479,9 @@ class FourierSeries(_Basis):
     NFP : int
         number of field periods
     n_umbilic : int
-        Prefactor of the form 1/NFP_umbilic_fac.
+        Prefactor of the form 1/n_umbilic.
         This is needed for the umbilic torus design.
+    sym:
         * ``'cos'`` for cos(n*z) symmetry
         * ``'sin'`` for sin(n*z) symmetry
         * ``'no n=0'`` for no n=0 mode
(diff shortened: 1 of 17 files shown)


singh-jaydeep:
Superseded by #2285

PR #2088, opened 2026-02-16
Add .python-version file

YigitElma:
Add .python-version file to prevent dependabot from using the minimum version specified in setup.py which prevents proper dependency updates. As suggested in https://github.com/dependabot/dependabot-core/issues/1455


dpanici:
I thought that was what we intended from the meeting? I dont care either way but that other PR is ready to go anyways I thought

PR #2090, opened 2026-02-17
ion density profile

### LINKED ISSUE #2069: atomic number vs ion density
Presently, we do not allow the user to specify the ion density profile directly. Instead, we input the `atomic_number` $Z_{eff}$ and then compute the ion density as: 

$n_i = \frac{n_e}{Z_{eff}}$

However, the effective atomic number is commonly computed as: 

$Z_{eff} = \frac{\Sigma n_i Z_i^2}{n_e}$

This will result in an incorrect ion density profile. What we really mean is the "average" atomic number defined as: 

$Z_{avg} = \frac{\Sigma n_i Z_i}{n_e}$

I see two solutions to resolve this issue: 

1. Update the documentation for the existing `atomic_number` profile and `"Zeff"` compute quantity to clarify that it is really the "average" atomic number. 
2. Add a new equilibrium profile for `ion_density`. This would be an optional profile that can be used instead of `atomic_number`, similar to how we handle the option for either a `current` or `iota` profile. 

I suggest that we implement both solutions.

ddudt:
Resolves #2069 by adding an optional `ion_density` profile to the `Equilibrium` class. If the profile is not supplied, it defaults to the previous assumption of $n_i = n_e / Z_{eff}$ for backwards compatibility. 

Note that the bootstrap current calculation likely needs further adjustments to work properly for $Z_{eff} > 1$, but this can be addressed in a future PR. 

All the example files were updated to have the new attribute `eq.ion_density = None`.


desc/compute/_bootstrap.py
ddudt, line 259: I'm confused why there is no `d_ni_d_s` term related to $\partial_\rho n_i$ in our bootstrap computations. When comparing this `dnds_term` to equation A1 from Landreman et al., it looks like we are assuming that $\frac{d \ln n_e}{d \psi_t} = \frac{d \ln n_i}{d \psi_t}$?

dpanici:
> I ran a test case comparing the old assumption to using the "correct" profiles. The difference is small, but I still think this PR is beneficial to give users more flexibility in prescribing accurate profiles.
> 
> Details: I have profile data for each ion and impurity species (which obeys quasi-neutrality). In the old code, I had to assume Z e f f ≈ 1 to get the correct ion density. In the new code, I can set the ion density directly and use the correct atomic number profile of Z e f f ≈ 1.2 . The only difference between the two cases is the `Zeff` profile. This plot shows the final bootstrap current profile, after optimizing for self-consistency. The differences in other equilibrium quantities are also negligible. image

How would these compare to say SFINCS for the "correct" profiles?

desc/compute/_bootstrap.py
```diff
@@ -368,15 +372,14 @@ def _J_dot_B_Redl(params, transforms, profiles, data, **kwargs):
...
+        "ni_r": grid.compress(data["ni_r"]),
         "Te": grid.compress(data["Te"]),
         "Te_r": grid.compress(data["Te_r"]),
         "Ti": grid.compress(data["Ti"]),
         "Ti_r": grid.compress(data["Ti_r"]),
     }
-    if profiles["atomic_number"] is None:
-        profile_data["Zeff"] = jnp.ones(grid.num_rho)
-    else:
-        profile_data["Zeff"] = grid.compress(data["Zeff"])
```
ddudt, line 379: [thought](https://pullpo.io/cc?l=thought) 
Generally our practice is to not assume a value if a profile is missing. When an `Equilibrium` is initialized, the `atomic_number` profile defaults to $Z_{eff}=1$. But if the profile is missing, then in the compute functions it returns `NaN`. If you try to compute the bootstrap current without setting the kinetic profiles correctly, you are going to get `NaN` values.

desc/compute/_bootstrap.py
f0uriest: this may be a place where its implicitly assumed that Zeff=1, so ni=ne

desc/equilibrium/equilibrium.py
```diff
@@ -126,9 +127,12 @@ class Equilibrium(IOAble, Optimizable):
     ion_temperature : Profile or ndarray shape(k,2) (optional)
         Ion temperature (eV) profile or array of mode numbers and spectral coefficients.
         Default is to assume electrons and ions have the same temperature.
+    ion_density : Profile or ndarray shape(k,2) (optional)
+        Electron density (m^-3) profile or array of mode numbers and spectral
```
f0uriest: *ion

desc/objectives/linear_objectives.py
```diff
@@ -2559,6 +2559,90 @@ def build(self, use_jit=False, verbose=1):
...
+    weight : {float, ndarray}, optional
+        Weighting to apply to the Objective, relative to other Objectives.
+        Must be broadcastable to ``Objective.dim_f``
+    normalize : bool, optional
+        Whether to compute the error in physical units or non-dimensionalize.
+    normalize_target : bool, optional
+        Whether target and bounds should be normalized before comparing to computed
+        values. If `normalize` is `True` and the target is in physical units,
+        this should also be set to True.
+    profile : Profile, optional
```
YigitElma: What is this profile input? I don't see it getting used later.

desc/objectives/normalization.py
```diff
@@ -53,14 +53,16 @@ def get_max_mode(basis, coeffs):
...
         )
-        if thing.electron_density is not None and thing.atomic_number is not None:
-            scales["n"] = float(
-                ((thing.atomic_number(0) + 1) / 2 * thing.electron_density(0))[0]
-            )
+        if thing.electron_density is not None:
+            if thing.ion_density is not None:
+                scales["n"] = np.mean([thing.electron_density(0), thing.ion_density(0)])
+            else:
+                scales["n"] = thing.electron_density(0)
```
YigitElma, line 70: Is there a specific reason why the old normalization has changed? I think they are same only if `thing.atomic_number(0)=1`

YigitElma:
This is not about this PR, but since this PR will trigger this warning a lot, maybe we can consider making the change here.

The warning [here](https://github.com/PlasmaControl/DESC/blob/624442e47fa8a52edd59e72fb4d359e681b61690/desc/io/hdf5_io.py#L127), creates the impression that developers should make a change to the code to make this work, but if the object has `_set_up` method defined, there is nothing to do other than saving the object again. I suggest changing it slightly,
```python
            if attr not in loc.keys():
                msg = (
                    "\n"
                    f"The object attribute '{attr}' was not loaded from the file.\n"
                    "This is likely because the file containing "
                    f"'{obj.__class__.__name__}' was created before '{attr}' "
                    f"became an attribute of objects of class '{obj.__class__}'.\n"
                    "The user may verify that a default value has been set.\n"
                    "This warning will persist until the file is saved with the "
                    "new object.\n"
                )
                if not hasattr(obj, "_set_up"):
                    msg += (
                        "\n"
                        "Note to developers: Add 'def _set_up(self)' as a method to "
                        f"class '{obj.__class__}'\n"
                        "(or the superclass where this new attribute is assigned) that "
                        f"assigns a value to '{attr}'.\n"
                        "This method is called automatically when a file is loaded.\n"
                        "Recall that the testing suite will fail on warnings."
                    )
                warnings.warn(colored(msg, "yellow"), RuntimeWarning)
                continue
```

desc/objectives/normalization.py
ddudt: The new normalization will always use the average density on axis: $(n_e + n_i) / 2$, unless the ion density can't be calculated and then only the electron density is used. There was a bug in the old normalization where it wasn't computing the average density properly for $Z_{eff} > 1$.

YigitElma (approved):

f0uriest (approved):

PR #2094, opened 2026-02-18
Move FiniteDiffDerivative to test utils, remove from public API

dpanici:


desc/derivatives.py
```diff
@@ -2,17 +2,13 @@
 
 from abc import ABC, abstractmethod
 
+import jax
```
YigitElma: I think during the dev meeting, we decided to keep the backend flexible for other autodiff libraries, so I think the conditional `if use_jax` is still useful. To be honest, I think that is a very distant future (so not very important for now) but everywhere in the code, we omit importing jax directly, so for consistency, having the conditional is better imo.

dpanici: Ah right sorry, I will revert that

ddudt (approved):
I still think we could remove the `FiniteDiffDerivative` class completely and replace it with something simpler, since it is only used in the testing functions. But this is fine for now.

YigitElma (approved):

PR #2098, opened 2026-02-19
Add errors, to MagneticWell and MercierStability objectives if grid contains axis.

### LINKED ISSUE #2097: ``MagneticWell`` objective could accept on-axis grid point, which does not make sense as it is always zero there
By default, we don't use the axis point for magnetic well since it is more expensive to compute. However it seems the reverse mode gradient is also NaN currently if there is a point on-axis 

```python
from desc.examples import get
from desc.objectives import MagneticWell
import numpy as np
eq = get("precise_QA")
obj = MagneticWell(eq,grid=LinearGrid(L=2,M=10,N=10,NFP=eq.NFP),deriv_mode="fwd")
obj.build()
grad_fwd = obj.jac_scaled_error(eq.params_dict)
print(np.any(np.isnan(grad_fwd[0]["R_lmn"]))) # False

obj = MagneticWell(eq,grid=LinearGrid(L=2,M=10,N=10,NFP=eq.NFP),deriv_mode="rev")
obj.build()
grad_rev = obj.jac_scaled_error(eq.params_dict)
print(np.any(np.isnan(grad_rev[0]["R_lmn"]))) # True
```

(this works fine if `axis=False` is passed to the grid)

```python
from desc.examples import get
from desc.objectives import *
import numpy as np
eq = get("precise_QA")
obj = MagneticWell(eq,grid=LinearGrid(L=2,M=10,N=10,NFP=eq.NFP,axis=False),deriv_mode="fwd")
obj.build()
grad_fwd = obj.jac_scaled_error(eq.params_dict)
print(np.any(np.isnan(grad_fwd[0]["R_lmn"]))) # False

obj = MagneticWell(eq,grid=LinearGrid(L=2,M=10,N=10,NFP=eq.NFP,axis=False),deriv_mode="rev")
obj.build()
grad_rev = obj.jac_scaled_error(eq.params_dict)
print(np.any(np.isnan(grad_rev[0]["R_lmn"]))) # False
```

I thought the issue might be due to the `replace_on_axis` but other quantities which have axis limits implemented (e.g. `"psi_r/sqrt(g)"`) can compute the reverse mode derivative just fine when an on-axis grid point is present, so it must be something else

For now making this an error if an on-axis grid point is present in `MagneticWell` (and doing so for `MercierStability` as that is actually not defined on-axis)

dpanici:
Resolves #2097 by ensuring the situation (which is a bit nonsensical anyways) is not reachable by the user.

Enforces what we already had assumed which was that the user should never pass an on-axis grid point to the MagneticWell or MercierStability objectives (the latter bc it is undefined and should be NaN, the former because it is just zero on-axis anyways by definition of its limit, and also we get NaN in the reverse mode gradient of the on-axis point)


desc/objectives/_stability.py
```diff
@@ -283,6 +289,14 @@ def build(self, use_jit=True, verbose=1):
...
             "resolution for surface averages",
         )
+        errorif(
+            grid.axis.size,
+            ValueError,
+            "MagneticWell objective grid cannot contain an axis point, "
+            "as its on-axis limit is always zero."
+            " Instead, pass axis=False when making the grid, which will "
+            " automatically place a grid point close but not at rho=0.",
+        )
```
ddudt: I would prefer if we are able to correctly implement the axis limit, since it should be well defined and always 0.

YigitElma:
- replace the division by 0 with safediv

YigitElma (approved):

ddudt (approved):

PR #2110, opened 2026-03-03
Adding QUADCOIL integration to DESC as an objective

lankef:
This pull request adds two files under /objective, `_quadcoil.py` and `_quadcoil_utils.py`. These allows one to call QUADCOIL using a DESC equilibrium and filament coil set as input, and perform quasi-single-stage optimization on the equilibrium and/or the coilset. There are some additions in coils and curves for conversion from simsopt objects. (since quadcoil is written in the simsopt convention).  Winding surface can be either fixed or auto-generated, but optimizing the geometry of the winding surface is not yet implemented. The unit test compares quadcoil to desc's built-in REGCOIL.


desc/geometry/curve.py
```diff
@@ -631,6 +631,49 @@ def from_values(cls, coords, N=10, s=None, basis="xyz", name=""):
             X_n=X_n, Y_n=Y_n, Z_n=Z_n, modes=basis.modes[:, 2], name=name
         )
 
+    def from_simsopt(curve_simsopt, name=""):
```
f0uriest: I think this should have a `@classmethod` decorator

desc/coils.py
```diff
@@ -949,6 +949,32 @@ def from_values(cls, current, coords, N=10, s=None, basis="xyz", name=""):
             name=name,
         )
 
+    def from_simsopt(coil_simsopt, name=""):
```
f0uriest: this also should have a `@classmethod` decorator

desc/compute/data_index.py
```diff
@@ -281,6 +281,12 @@ def _decorator(func):
         "desc.geometry.core.Surface",
         "desc.magnetic_fields._core.MagneticField",
     ],
+    "desc.magnetic_fields._quadcoil_field.QuadcoilField": [
```
f0uriest: is this class missing?

desc/external/_quadcoil_utils.py
```diff
@@ -0,0 +1,452 @@
...
+        ["G"],
+        params=params_eq,
+        transforms=constants["net_poloidal_current_transforms"],
+        profiles=constants["net_poloidal_current_profiles"],
+    )
+    net_poloidal_current_amperes = -G_data["G"][0] / mu_0 * 2 * jnp.pi
+    return net_poloidal_current_amperes
+
+
+def _ptolemy_identity_rev_precompute(m_1, n_1):
```
f0uriest, line 104: I think you can use `desc.vmec_utils.ptolemy_linear_transform` instead of this. It returns a matrix `A` such that `A@x_desc = x_vmec`. You can create the matrix in the build method. See the `QuasisymmetryBoozer` objective for an example

desc/objectives/_quadcoil_utils.py
```diff
@@ -0,0 +1,452 @@
...
+    except ModuleNotFoundError:
+        raise ModuleNotFoundError("QuadcoilProxy requires a QUADCOIL installation.")
+    # Converts a kwargs for quadcoil into a kwargs for QuadcoilField or
+    # FourierCurrentPotentialField
+    filtered = {}
+    source_kwargs = quadcoil_kwargs.copy()
+    if "winding_stellsym" in source_kwargs.keys():
+        winding_stellsym = source_kwargs["winding_stellsym"]
+    else:
+        winding_stellsym = sym_default
```
f0uriest: could be `winding_stellsym = source_kwargs.get("winding_stellsym", sym_default)`

desc/objectives/_quadcoil_utils.py
```diff
@@ -0,0 +1,452 @@
...
+    else:
+        filtered["sym_Phi"] = False
+    if "smoothing" not in source_kwargs.keys():
+        filtered["smoothing"] = "approx"
+        filtered["smoothing_params"] = {"lse_epsilon": 1e-3}
+    elif source_kwargs["smoothing"] == "slack":
+        warnings.warn(
+            "It is not advised to perform single-stage "
+            "optimization using the 'slack' smoothing mode, because "
+            "DESC may hang when the constraint has large dimensions (~500), "
```
f0uriest: can you explain this a bit? is this a large linear constraint causing DESC to hang?

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+        sure that the grid points in DESC B calculations line up exactly
+        with the QUADCOIL grids.
+        Values lower than eq.M_grid will trigger interpolation truncation
+        warnings.
+    plasma_N_phi : int, optional, default=eq.N_grid
+        The plasma toroidal quadrature resolution.
+    target : scalar or ndarray, optional, default=None
+    bounds : scalar or ndarray, optional, default=None
+    weight : scalar or ndarray, optional, default=None
+        The original targets, bounds and weight available in every DESC Objective class.
```
f0uriest: should make clear that the default is to use target=0, see some of the other objective docstrings for example.

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+    metric_name : str or tuple of str, default=None
```
f0uriest: what does the default value of None do?

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+    normalize : bool, optional, default=False
+    normalize_target : bool, optional, default=False
+        Normalization options for a typical DESC Objective. Disabled by default.
+        When enabled, overrides the ``<quantity>_unit`` in
+        ``quadcoil_kwargs`` with scaling constants calculated from the parameteres
+        of the DESC equilibrium. Note that QUADCOIL usually works the best
+        when ``<quantity>_unit`` are the same quantities measured from another
+        winding surface solution (either the solution of the same problem with
+        ``<quantity>_unit=1``, or the solution of a REGCOIL problem). Setting
+        this to ``True`` may impact QUADCOIL's accuracy.
```
f0uriest: why would this affect accuracy? isnt it just an overall scale factor?

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+    field_grid : Grid, optional, default=None:
+        The grid for ``field``.
+    enable_net_current_plasma : bool, optional, default=True,
+        Whether to enable a net poloidal current in the winding surface. ``True``
+        by default.
+    eq_fixed : bool, optional, default=False
+        Whether to fix ``eq``, or make it optimizable degrees of freedom. ``False``
+        by default for quasi-single-stage optimization.
+    field_fixed : bool, optional, default=True
+        Whether to fix ``eq``, or make it optimizable degrees of freedom. ``True``
```
f0uriest: whether to fix ``field``

f0uriest: I think also we usually default to assuming its not fixed

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+        by default for quasi-single-stage dipole/PM optimization with known
+        filament coils.
+    Bnormal_plasma_chunk_size : int, optional, default=None
+    jac_chunk_size : int, optional, default=None
+    bs_chunk_size : int, optional, default=None
+        Chunk sizes.
```
f0uriest: should copy the full docstring for these from another objective

desc/external/quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+    # r=rho, t=theta, z=zeta? i.e. if only a profile, it is "r" , while if all
+    # 3 coordinates it is "rtz"
+    _units = "N/A"  # units of the output
+    # string with python string formatting for printing the value
+    _print_value_fmt = "QUADCOIL subproblem: "
+
+    def __init__(  # noqa: C901
+        self,
+        eq,
+        quadcoil_kwargs,
```
f0uriest, line 204: is there some sensible default we can give for quadcoil kwargs?

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+        metric_weight=1.0,
+        metric_target=0.0,
+        vacuum: bool = False,
+        normalize=False,
+        normalize_target=False,
+        verbose=0,
+        name="QUADCOIL Proxy",
+        source_grid=None,
+        # External coils - no external coils by default
+        field=[],
```
f0uriest: lists are mutable so they generally shouldn't be used as default args since it can lead to some weird behavior. Better to use `None` or an empty tuple

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+                "This is very uncommon (windowpane filaments + "
+                "winding surface with net current)."
+            )
+
+        quadcoil_kwargs = quadcoil_kwargs.copy()
+        if target is None and bounds is None:
+            target = 0  # default target value
+        # Uses LSE to smooth non-smooth problems rather than slack variables
+        # by default.
+        if "smoothing" not in quadcoil_kwargs.keys():
```
f0uriest: can use `quadcoil_kwargs.setdefault("smoothing", "approx")`

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+        # through quadcoil.io.quadcoil_for_diff, which cannot see their default
+        # values in quadcoil.quadcoil.
+        if "net_toroidal_current_amperes" in quadcoil_kwargs.keys():
+            self.net_toroidal_current_amperes = quadcoil_kwargs.pop(
+                "net_toroidal_current_amperes"
+            )
+        else:
+            self.net_toroidal_current_amperes = 0.0
+        if "plasma_coil_distance" in quadcoil_kwargs.keys():
+            # A sign flip is necessary here!
```
f0uriest: why sign flip? shouldn't distance always be positive?

desc/external/quadcoil.py
```diff
@@ -0,0 +1,921 @@
...
+                eval_grid=self._constants["eval_grid"],
+            )
+            self._constants["source_profiles"] = source_profiles
+            self._constants["source_transforms"] = source_transforms
+            self._constants["interpolator"] = interpolator
+
+        if self._field:
+            from desc.magnetic_fields import SumMagneticField
+
+            self._constants["sum_field"] = SumMagneticField(self._field)
```
f0uriest, line 554: I think it's better to assume that the user will already give a sum field, rather than a possible list of fields to sum. The general format we've been working with is that `field` is always assumed to be a single `MagneticField` object

f0uriest (commented):
Just some initial comments after a first pass, I'll have to understand quadcoil a bit better to give a more detailed review.

Also, you should add quadcoil to `devtools/dev-requirements.txt` so it will be installed for testing. (and maybe also `requirements.txt`? Or we can leave it as an optional dependency?)

desc/external/quadcoil.py
dpanici: We silently allow this in `BoundaryError`, `VacuumBoundaryError` and `QuadraticFlux` as a stopgap before we actually make things work correctly with `SumMagneticField` being passed in directly... this is just further impetus to actually get that working

desc/geometry/curve.py
lankef: Fixed

desc/coils.py
lankef: Fixed

desc/compute/data_index.py
lankef: Thanks for catching this. It's a WIP class and I've decided to not include it here

lankef: removed now

desc/external/_quadcoil_utils.py
lankef: I'm already using `ptolemy_linear_transform` inside this function. It's just a wrapper to reduce some code reusing.

desc/objectives/_quadcoil_utils.py
lankef: Fixed. Also replaced a few other occurrences.

desc/objectives/_quadcoil.py
lankef: Fixed to None.

desc/objectives/_quadcoil.py
lankef: Fixed

desc/objectives/_quadcoil.py
lankef: Because simsopt surface and DESC surface have different handedness. QUADCOIL uses the simsopt convention so a sign flip is necessary. I've added more comments here.

desc/external/quadcoil.py
lankef: I can make it solve the NESCOIL problem by default but it will require a default value for `surface_plasma_distance`, too. I feel maybe it's sensible to have the user think about it themselves? I've uploaded example ones in the tutorial notebook too.

desc/objectives/_quadcoil_utils.py
lankef: This is a leftover from the WIP QuadcoilField class. I've removed it.
QuadcoilField is for optimizing the current potential directly in desc, like you can do with FourierCurrentPotentialField. There is a mode in quadcoil that smoothes non-smooth functions like absolute values with slack variables. These slack variables can have O(ngrid) elements because many non-smooth functions are pointwise quantities (L1 norm of dipole density), and having this many dofs in a DESC problem will cause it to freeze.

desc/objectives/_quadcoil.py
lankef: Fixed

desc/objectives/_quadcoil.py
lankef: The default value uses the QUADCOIL objective specified in quadcoil_kwargs. Should be fixed now.

desc/objectives/_quadcoil.py
lankef: Yeah. The scale of a quantity (like f_B, f_K) on a QUADCOIL solution can differ by orders of magnitude from the DESC estimates based on equilibrium properties. The scale of these quantity can acctually differ by orders of magnitude across different quadcoil problems. Quadcoil's auglag's convergence rate test is based on this scale factor, so picking this scale factor from a similar quadcoil problem is the most reliable. I added this warning just so that the user is aware of this.

desc/objectives/_quadcoil.py
lankef: Fixed.

desc/objectives/_quadcoil.py
lankef: Gotcha. I've modified my `collect_docs` call for this. Please help me check if it's correct.

desc/objectives/objective_funs.py
```diff
@@ -270,6 +270,7 @@ class ObjectiveFunction(IOAble):
         "_use_jit",
     ]
     _static_attrs = [
+        "_static_attrs",  # A bug as of Oct 10 2025
```
dpanici: specifically, for some reason one needs static_attrs included in the static_attrs or one will get JAX errors. This seems unintuitive but is empirically what we've found

docs/index.rst
```diff
@@ -44,6 +44,7 @@
    notebooks/tutorials/basic_optimization.ipynb
    notebooks/tutorials/advanced_optimization.ipynb
    notebooks/tutorials/coil_optimization_REGCOIL.ipynb
+   notebooks/tutorials/coil_optimization_QUADCOIL.ipynb
    notebooks/tutorials/omnigenity.ipynb
    notebooks/tutorials/bootstrap_current.ipynb
    notebooks/tutorials/coil_stage_two_optimization.ipynb
```
dpanici, line 50: Also need to add QUADCOIL stuff to the API, I think the file is https://github.com/PlasmaControl/DESC/blob/master/docs/api_objectives.rst?plain=1

dpanici: add the objectives, I mean

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,917 @@
...
+        Uses the normalized objective by default.
+        We strongly advise using the default value to ensure accurate adjoint
+        differentiation.
+    metric_target : scalar or ndarray, default=None
+        In addition to target, bounds and weight,
+        The QUADCOIL proxy objective allows the user to set weights and
+        targets for each objective terms individually besides using ``target``
+        and ``bounds`` that comes with other DESC objectives.
+        Targets of each property. 0 by default.
+    metric_weight : scalar or ndarray, default=None
```
dpanici: 1.0 not None

desc/external/_quadcoil_utils.py
```diff
@@ -0,0 +1,437 @@
...
+            Bnormal += constants["Bnormal_plasma"].reshape(Bnormal.shape)
+        else:
+            Bnormal += _compute_Bnormal_plasma(
+                constants, params_eq, bplasma_chunk_size
+            ).reshape(Bnormal.shape)
+
+    return Bnormal
+
+
+# Flipping the current potential of a quadcoil
```
dpanici, line 274: what are the different coordinate directions/representations/normalizations used? it woudl be good to list these all somewhere, even if just as a comment or in this PR, of the differences in convention btwn DESC and QUADCOIL. Just so we have it clearly down somewhere. Seems phi is flipped?

desc/external/_quadcoil_utils.py
```diff
@@ -0,0 +1,437 @@
...
+        if mi != 0:
+            j = index_map.get((mi, -ni))
+            if j is not None:
+                phi_swapped[i] = phi[j]
+        else:
+            phi_swapped[i] = -phi[i]
+    return phi_swapped
+
+
+def _quadcoil_phi_to_desc_phi(phi_mn_quadcoil, stellsym, mpol, ntor):
```
dpanici, line 294: this is converting just phi_mn quadcoil to desc?

desc/external/_quadcoil_utils.py
```diff
@@ -0,0 +1,437 @@
...
+        phic = phi_mn_quadcoil[:-len_sin]
+
+    phis = jnp.insert(phis, 0, 0.0)
+    mc, _, nc, _ = make_rzfourier_mc_ms_nc_ns(mpol, ntor)
+    modes_M, modes_N, Phi_mn = ptolemy_identity_fwd(mc, nc, phis, phic)
+    Phi_mn = Phi_mn.flatten()
+    return Phi_mn, modes_M, modes_N
+
+
+def _create_source(eq, source_grid, eval_grid):
```
dpanici, line 319: this could be placed perhaps into the `integrals/singularities.py` or in a util file there. Also could be used again inside of the other places like `BoundaryError`. I think it is a reasonable util function, with perhaps some additional documentation attached to it

desc/external/_quadcoil_utils.py
```diff
@@ -0,0 +1,437 @@
...
+            "\nReason: " + str(e),
+        )
+        interpolator = DFTInterpolator(eval_grid, source_grid, st, sz, q)
+    # Creating source grids
+    source_profiles = get_profiles(_BPLASMA_DATA_KEYS, obj=eq, grid=source_grid)
+    source_transforms = get_transforms(_BPLASMA_DATA_KEYS, obj=eq, grid=source_grid)
+    return (source_profiles, source_transforms, interpolator)
+
+
+def _quadcoil_kwargs_to_field_kwargs(  # noqa: C901
```
dpanici, line 340: would be nice to document what is being done here exactly?

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,917 @@
...
+        rs = rs_raw.flatten()[1:]
+        zc = zc_raw.flatten()
+        zs = zs_raw.flatten()[1:]
+        if self.stellsym:
+            plasma_dofs = jnp.concatenate([rc, zs])
+        else:
+            plasma_dofs = jnp.concatenate([rc, rs, zc, zs])
+        return plasma_dofs
+
+    def compute_surface_current_field(self, *all_params):
```
dpanici: instead of `compute_surface_current_field` maybe something more like `run_regcoil` or `solve_quadcoil_surface_current_field` or something is better?

desc/objectives/objective_funs.py
lankef: Should I keep this? I don't need to do anything about this right?

docs/index.rst
lankef: Added!
https://github.com/PlasmaControl/DESC/commit/9faf614291c4b146c86324cb2c97f2d1e69a8978

desc/objectives/_quadcoil.py
lankef: Should be fixed now.

desc/objectives/_quadcoil.py
lankef: I renamed this to solve_quadcoil_surface_current_field, I also renamed compute_full to solve_quadcoil.

lankef (commented):
Addressed the comments

rahulgaur104:
qp_nescoil.winding_surface.plot() requires simsopt. Please replace with DESC surface plotter.

desc/objectives/objective_funs.py
YigitElma: `_static_attrs` is already included couple lines below. You probably had to add it before we added in master.

desc/objectives/objective_funs.py
```diff
@@ -1254,11 +1255,12 @@ def build(self, use_jit=True, verbose=1):
 
         # set quadrature weights if they haven't been
         if hasattr(self, "_constants") and ("quad_weights" not in self._constants):
-            grid = self._constants["transforms"]["grid"]
             if self._coordinates == "rtz":
+                grid = self._constants["transforms"]["grid"]
```
YigitElma: can you revert these changes?

desc/plotting.py
```diff
@@ -974,6 +974,7 @@ def plot_2d(  # noqa : C901
     else:
         im = ax.contourf(xx, yy, data, **contourf_kwargs)
     cax = divider.append_axes("right", **cax_kwargs)
+    _set_tight_layout(fig)
```
YigitElma: This change is also unnecessary

desc/io/optimizable_io.py
```diff
@@ -92,7 +92,7 @@ def _make_hashable(x):
 
 def _unmake_hashable(x):
     # turn tuple of ints and shape to ndarray
-    if isinstance(x, tuple) and x[0] == "ndarray":
+    if isinstance(x, tuple) and len(x) and x[0] == "ndarray":
```
YigitElma, line 95: I don't think it hurts to have it, but why did you need this change?

devtools/dev-requirements.txt
```diff
@@ -37,6 +37,7 @@ pytest-split >= 0.8.2, <= 0.11.0
 qicna @ git+https://github.com/rogeriojorge/pyQIC/
 qsc <= 0.1.3
 shapely >= 1.8.2, <= 2.1.2
+quadcoil >= 0.1.0
```
YigitElma, line 40: Sorry if I am missing but is `quadcoil` a pip package? I coulnd't find it on PyPI and our CI cannot install it and the tests don't run because of that.

YigitElma (commented):
Thanks fore adding this!
I will review rest of the code later, but it would be nice if you can resolve the problem with CI tests first. It looks like CI and pip in general cannot install `quadcoil`, I couldn't install it locally either.

desc/geometry/curve.py
```diff
@@ -631,6 +631,50 @@ def from_values(cls, coords, N=10, s=None, basis="xyz", name=""):
...
+        Returns
+        -------
+        coil : FourierXYZCurve
+            A FourierXYZCurve.
+        """
+        try:
+            from simsopt.geo import CurveXYZFourier
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                "desc.geometry.from_simsopt requires a " "simsopt installation"
```
YigitElma:
```suggestion
                "desc.geometry.from_simsopt requires simsopt package."
```

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
+import warnings
+
+from jax import jit
```
YigitElma:
```suggestion
from desc.backend import jit
```

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
...
+            stellsym,
+            plasma_mpol, plasma_ntor,
+            plasma_quadpoints_phi, plasma_quadpoints_theta,
+            plasma_dofs,
+            net_poloidal_current_amperes,
+            Bnormal_plasma,
+            metric_name,
+            value_only,
+            verbose,
+    plasma_M_theta : int, optional, default=eq.M_grid
```
YigitElma: We usually give defaults at the end of the description

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
...
+        by default for quasi-single-stage dipole/PM optimization with known
+        filament coils.
+    B_plasma_chunk_size : int or None
+        Size to split singular integral computation for B_plasma into chunks.
+        If no chunking should be done or the chunk size is the full input
+        then supply ``None``. Default is ``bs_chunk_size``.
+    bs_chunk_size : int, optional, default=None
+        Size to split Biot-Savart computation into chunks of evaluation points.
+
+    Attributes
```
YigitElma: These don't seem necessary

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
...
+                "will be discarded."
+            )
+
+        # ----- Storing equilibrium-independent, differentiable variables -----
+        # These are differentiable quantities that are not equilibrium-dependent.
+        # They can be user-provided, but they also all have default values, so
+        # we set them here. This is necessary because we are calling quadcoil
+        # through quadcoil.io.quadcoil_for_diff, which cannot see their default
+        # values in quadcoil.quadcoil.
+        if "net_toroidal_current_amperes" in quadcoil_kwargs.keys():
```
YigitElma: You can replace all `if/else` statements by `something = dict.pop("something", default)`

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
...
+        self.metric_target = metric_target
+        self.metric_weight = metric_weight
+        self._verbose = verbose
+        self._bplasma_chunk_size = B_plasma_chunk_size
+        self._bs_chunk_size = bs_chunk_size
+        self._vacuum = vacuum
+        if not plasma_M_theta:
+            plasma_M_theta = eq.M_grid
+        elif plasma_M_theta <= eq.M_grid:
+            warnings.warn(
```
YigitElma, line 352: These warnings can be more readable and compact if you use f-strings. Like
```python
warnings.warn(
    f"plasma_M_theta={plasma_M_theta} <= eq.M_grid={eq.M_grid}."
    "An interpolation truncation warning may appear."
)
```

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
...
+                + " <= "
+                + "eq.M_grid = "
+                + str(eq.M_grid)
+                + ". "
+                + "An interpolation truncation warning may appear."
+            )
+        if not plasma_N_phi:
+            plasma_N_phi = eq.N_grid
+        elif plasma_N_phi <= eq.N_grid:
+            warnings.warn(
```
YigitElma, line 359: Same as above

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
...
+        # here we treat all of them like staic using
+        # partial(quadcoil, **quadcoil_kwargs), implemented in gen_quadcoil_for_diff.
+        # The function also generates the custom_jvp rule based on the static arguments.
+        # We store the resulting function as a static attribute.
+        _quadcoil_values, _quadcoil_for_diff = gen_quadcoil_for_diff(**quadcoil_kwargs)
+        # Used later for Bnormal_plasma also
+        self._quadcoil_for_diff = jit(_quadcoil_for_diff)
+        self._quadcoil_values = jit(_quadcoil_values)
+        # ----- Setting and registering keyword arguments -----
+        self._static_attrs = _Objective._static_attrs + [
```
YigitElma: You can move this to line 165, right after docs. We usually define it there

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
...
+            # External-coils related
+            "_enable_net_current_plasma",
+            "_eq_fixed",
+            "_field_fixed",
+            "_bs_chunk_size",
+            # Basics
+            "_static_attrs",
```
YigitElma: `_Objective._static_attrs` have the `_static_attrs` and most of these other keys in it. You can remove them safely

desc/objectives/_quadcoil_utils.py
```diff
@@ -0,0 +1,437 @@
+import inspect
+from functools import partial
+
+import numpy as np
+from jax import jit
```
YigitElma:
```suggestion
from desc.backend import jit
```

desc/objectives/_quadcoil.py
```diff
@@ -0,0 +1,916 @@
...
+        if verbose > 1:
+            timer.disp("Precomputing transforms")
+        # finally, call ``super.build()``
+        super().build(use_jit=use_jit, verbose=verbose)
+
+    def compute(self, *all_params, constants=None):
+        """Computes the scalar value of the QUADCOIL proxy.
+
+        Computes the scalar value of the QUADCOIL proxy. A wrapper for
+        ``compute_full``.
```
YigitElma: Why do you have a wrapper here? And what is `compute_full`? You actually pass `full_mode=False`.

desc/io/optimizable_io.py
lankef: The default values for some QUADCOIL arguments are empty list/tuples (I had to do it because the adjoint derivative is wrapped in a vjp rule before handed to DESC and this way some flows are simpler). Without this check it breaks DESC Optimizable.

devtools/dev-requirements.txt
lankef: It wasn't but now it is!

desc/objectives/_quadcoil.py
lankef: Moved.

YigitElma:
@lankef https://github.com/PlasmaControl/DESC/actions/runs/25258105132/job/74060745121?pr=2110 this error happens if you commit a notebook from cluster or after running notebook in Jupyter Lab session (tbh it is very annoying and we have a long standing issue about this #1215, but it is actually not our problem but `nbmake` or jupyterlab's). You can check this related comment https://github.com/PlasmaControl/DESC/pull/1498#issuecomment-2578440156 for fix. Basically change `kernel name` of the notebook metadata or just open/save (without running it) on VS code extension.

YigitElma:
I made some small changes to fix the failing test and fixed a couple of bugs (for example, `classmethod`s expect `cls` as the first argument). In the notebook, I added a helper function for plotting, I think you can use it instead of duplicated code (I just added the helper, but didn't change the code to use it yet). Also, it looks like the notebook is currently trying to use 127GB of memory which is not possible to run on CI. I added `jac_chunk_size=1` to `QuadcoilProxy` objective, but we may need to reduce the resolution or something else to keep it short. For CI purposes, we want notebooks to run in around 10 minutes.

One additional question for other devs: do we want this in the `desc/objectives` module or `desc/external` module? @dpanici @f0uriest @ddudt

YigitElma:
We decided to put this to `desc/external`

lankef:
So this will be merged into desc/external? or should I open a new pull request or something?

YigitElma:
> So this will be merged into desc/external? or should I open a new pull request or something? 
> 

desc/external is just a submodule of desc where we keep external code couplings. Since this is mostly a wrapper around Quadcoil which we cannot test well, we think that module is a better fit for this code.

All we need to do is to move this code to that folder. We can handle it once it is finalized or you can move it there now. I can help with it.

lankef:
> > So this will be merged into desc/external? or should I open a new pull request or something?
> 
> desc/external is just a submodule of desc where we keep external code couplings. Since this is mostly a wrapper around Quadcoil which we cannot test well, we think that module is a better fit for this code.
> 
> All we need to do is to move this code to that folder. We can handle it once it is finalized or you can move it there now. I can help with it.

That's perfect.  Do you think it is in a good enough state for the move now? I can take a stab at it.

YigitElma:
> Do you think it is in a good enough state for the move now?

Yeah, I think we can move it now and then continue making changes there. I will try to find some time to properly review this; sorry for the delay.

YigitElma:
@lankef I moved the code to `desc.external`. I also tried to solve the memory issue. On my laptop GPU, the first step is able to run, but it is very slow. Either the memory issue happens in later steps, or the CPU XLA compiler behaves weirdly. In any case, I think it's worth understanding why it is very slow in the first place. Just for reference, 6 iterations of the first optimization loop took around 30 mins on GPU. I am pretty sure all this time is spent in `QuadcoilProxy` since other parts shouldn't take more than couple seconds for this resolution (LMN=6)

lankef:
> @lankef I moved the code to `desc.external`. I also tried to solve the memory issue. On my laptop GPU, the first step is able to run, but it is very slow. Either the memory issue happens in later steps, or the CPU XLA compiler behaves weirdly. In any case, I think it's worth understanding why it is very slow in the first place. Just for reference, 6 iterations of the first optimization loop took around 30 mins on GPU. I am pretty sure all this time is spent in `QuadcoilProxy` since other parts shouldn't take more than couple seconds for this resolution (LMN=6)

Interesting. I haven't ran this for a while so let me check (I've also changed some internals of quadcoil since then). Were you running the example file? Are you using the latest quadcoil from the repo? It should take about 10 secs to 1 min per steps (on my pc with a 4060). Something must be off.

YigitElma:
> Are you using the latest quadcoil from the repo?

I am using the pip release, not the repo. 

>  Were you running the example file? 

I used the same notebook.

lankef:
> > Are you using the latest quadcoil from the repo?
> 
> I am using the pip release, not the repo.
> 
> > Were you running the example file?
> 
> I used the same notebook.

Lemme try and get back to you. The pip release is a bit out of date and I let me update it too.

PR #2128, opened 2026-03-17
Update version constraints for interpax, orthax, quadax, jax-finufft

f0uriest:
New versions support jax 0.9

Also loosens the versioning a bit to only peg to minor releases, not patch, so we don't have to update as often.


requirements.txt
```diff
@@ -3,19 +3,19 @@ colorama <= 0.4.6
 diffrax >= 0.6.0, <= 0.7.2
 equinox >=0.11.10, <=0.13.6
 h5py >= 3.0.0, <= 3.16.0
-interpax >= 0.3.3, <= 0.3.12
+interpax >= 0.3.3, < 0.4
 jax-finufft >= 1.1.0
```
YigitElma: Can we also put an upper bound on `jax-finufft` in this PR? Or should I add it in a new one?

YigitElma (dismissed):

YigitElma (approved):

dpanici (approved):

PR #2133, opened 2026-03-18
Add shared family coil parameters objective

lkadz:
This PR introduces a new objective `ShareFamilyCoilParameters` that enforces equality constraints between selected parameters across groups of coils. The objective allows users to tie parameters (e.g. currents) between coils belonging to the same family or across multiple families, optionally restricted to specific indices.

### Motivation
In coil optimization problems, it is desirable to:
- enforce symmetry between coils (e.g. TF, PF, CS families)
- reduce the number of independent design variables
- impose engineering constraints linking coil parameters

### Description of the approach
The objective:
- groups coils using:
  -   "family" (single family)
  -   "families" (multiple families)
  -   "coils" (explicit indices)
- selects parameters via "param"
- optionally restricts to subsets via "indices"
- constructs linear residuals of the form: first_coil_param − other_coil_param = 0

### Example usage
A usage example is provided in [stage_two_ShareFamilyCoilParameters.ipynb](https://github.com/user-attachments/files/26099953/stage_two_ShareFamilyCoilParameters.ipynb) demonstrating sharing currents between coils.

### DIFF
--- a/desc/objectives/linear_objectives.py
+++ b/desc/objectives/linear_objectives.py
@@ -333,7 +333,7 @@ class ShareParameters(_Objective):
         name="shared parameters",
     ):
         self._params = params
-        assert len(things) > 1, "only makes sense for >1 thing"
+        assert len(things) >= 1, "must have at least 1 thing"
         assert np.all(
             [isinstance(things[0], type(t)) for t in things[1:]]
         ), f"expected same type for all things, got types {[type(t) for t in things]}"
@@ -347,6 +347,55 @@ class ShareParameters(_Objective):
             normalize_target=False,
             name=name,
         )
+
+    def _expand_param_shortcuts(self, params, thing):
+        """Expand shortcut syntax like {"current": [...]} into full pytree."""
+        if not isinstance(params, dict):
+            return params
+
+        dims = thing.dimensions
+
+        def expand_node(mask, dim_node):
+            if isinstance(dim_node, dict):
+                out = {}
+                for pname, dim in dim_node.items():
+                    if pname not in params:
+                        continue
+
+                    if isinstance(mask, dict):
+                        val = mask.get(pname, False)
+                    else:
+                        val = mask
+
+                    if val is True:
+                        out[pname] = np.arange(dim)
+                    elif val is False:
+                        out[pname] = np.array([], dtype=int)
+                    else:
+                        out[pname] = np.atleast_1d(np.asarray(val, dtype=int))
+
+                return out
+
+            if isinstance(dim_node, list):
+                if isinstance(mask, dict):
+                    mask = [
+                        {
+                            pname: (
+                                mask[pname][i]
+                                if isinstance(mask[pname], list)
+                                else mask[pname]
+                            )
+                            for pname in mask
+                        }
+                        for i in range(len(dim_node))
+                    ]
+
+                if not isinstance(mask, list):
+                    mask = [mask] * len(dim_node)
+
+                return [expand_node(m, d) for m, d in zip(mask, dim_node)]
+
+        return expand_node(params, dims)
 
     def build(self, use_jit=False, verbose=1):
         """Build constant arrays.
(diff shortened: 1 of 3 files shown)


desc/objectives/linear_objectives.py
```diff
@@ -3763,3 +3763,253 @@ def compute(self, params, constants=None):
         """
         f = jnp.dot(self._A, params["L_lmn"]).squeeze()
         return f
+
+
+class ShareFamilyCoilParameters(_Objective):
+    """
+    Objective enforcing shared parameter values across groups of coils.
```
YigitElma:
```suggestion
    Fix specific degrees of freedom to be the same between coils.
```

desc/objectives/linear_objectives.py
```diff
@@ -3763,3 +3763,253 @@ def compute(self, params, constants=None):
...
+
+    This objective penalizes differences in specified parameters between
+    coils belonging to the same group (e.g., same family).
+
+    Parameters
+    ----------
+    thing : CoilSet or Coil
```
YigitElma:
```suggestion
    thing : CoilSet or MixedCoilSet
```

desc/objectives/linear_objectives.py
```diff
@@ -3763,3 +3763,253 @@ def compute(self, params, constants=None):
         """
         f = jnp.dot(self._A, params["L_lmn"]).squeeze()
         return f
+
+
+class ShareFamilyCoilParameters(_Objective):
```
YigitElma: I think `ShareCoilParameters` is a better name for this. We use `family` only for `EquilibriaFamily`, not coilsets.

desc/objectives/linear_objectives.py
```diff
@@ -3763,3 +3763,253 @@ def compute(self, params, constants=None):
...
+    Objective enforcing shared parameter values across groups of coils.
+
+    This objective penalizes differences in specified parameters between
+    coils belonging to the same group (e.g., same family).
+
+    Parameters
+    ----------
+    thing : CoilSet or Coil
+        Coil container object.
+    groups : list of dict
```
YigitElma: I think if we use the same API as `ShareParameters`, it would be easier for users. Because I think people don't give names to their coils in general. I was thinking something like,
```python
"""
params : dict
    Dict keys are the names of parameters to fix (str), and dict values are the
    indices to fix for each corresponding parameter (int array).
    Use True (False) instead of an int array to fix all (none) of the indices
    for that parameter.

Example
-------
"""
# Let's say we have a MixedCoilSet with
# CoilSet with 3 unique coils and 2 extra single coils
# Something like this will fix the currents of the 
# first 2 coils of the coilset as well as second single coil
params = {
   "current": {{True, True, False}, False, True},
}

File: desc/objectives/linear_objectives.py [hunk 5]
```diff
@@ -3763,3 +3763,253 @@ def compute(self, params, constants=None):
...
+        family_map = {}
+        start = 0
+
+        if hasattr(thing, "coils"):
+            for i, child in enumerate(thing.coils):
+                leaves = self._flatten_leaf_coils(child)
+                n = len(leaves)
+                name = getattr(child, "name", "") or f"family_{i}"
+
+                errorif(
```
YigitElma: I think instead of using the name, we can use the pytree structure to map indices.

desc/objectives/linear_objectives.py
YigitElma: If the parameter that will be shared is an array, the example usage can be
```python
# For above MixedCoilSet, this will make sure that index 0 of the 3
# coils of the coilset are the same (also index 1, but it can be different than 
# index 0)
params = {
   "R_n": {{[0, 1], [0, 1], [0, 1]}, False, False},
}
```

lkadz:
Here is a notebook, with an example of a single `MixedCoilSet`, that go through the updated `ShareParameters` class:
[ShareParameters.ipynb](https://github.com/user-attachments/files/26612243/ShareParameters.ipynb)

PR #2140, opened 2026-03-27
`num_neighbors` pruning for `CoilSetMinDistance`

TElder-thea:
Add optional `num_neighbors` parameter to the `CoilSetMinDistance` objective. This limits pairwise distance computation to only the `num_neighbors` nearest neighbors per coil, selected by centroid distance. Neighbor indices use jax.lax.stop_gradient so AD only traces through the fine-grained distances to nearby coils. 

Depending on coilset used, we found:
- 16.5x speedup on CoilSetMinDistance Jacobian alone
- 5.8x full auglag optimizer speedup (468s -> 80s)
- 3.5x full lsq-exact optimizer speedup (464s -> 134s)
- Enables jac_chunk_size=32 (baseline OOMs at chunk=6)
- Machine-precision correctness (K=5 matches full at alpha=100)
For more generic modular coil optimizations, pruning may decrease performance due to overhead.

Usage: CoilSetMinDistance(..., n_neighbors=20)
Default n_neighbors=None preserves existing behavior.


desc/objectives/_coils.py
```diff
@@ -1,6 +1,7 @@
 import numbers
 import warnings
 
+import jax
```
ddudt:
```suggestion
from desc.backend import jax
```

desc/objectives/_coils.py
```diff
@@ -911,6 +916,7 @@ def __init__(
         use_softmin=False,
         softmin_alpha=1.0,
         dist_chunk_size=None,
+        n_neighbors=None,
```
ddudt: add documentation for this new input parameter to the class docstring

desc/objectives/_coils.py
```diff
@@ -984,18 +991,36 @@ def compute(self, params, constants=None):
         pts = constants["coilset"]._compute_position(
             params=params, grid=constants["grid"], basis="xyz"
         )
+        n_coils = pts.shape[0]
```
ddudt: `self._dim_f = coilset.num_coils`

If `n_coils` is the number of coils, you can reference `self.dim_f` instead.

ddudt (dismissed):
Remember to also update `CHANGELOG.md`

ddudt:
Note for reviewers: Some of the original code was written by Claude, but I re-wrote almost everything as a human.

f0uriest (approved):

tests/test_objective_funs.py
```diff
@@ -1084,20 +1084,55 @@ def test_coil_type_error(self):
...
+            assert f4.size == coils.num_coils
+            np.testing.assert_allclose(f4, mindist, rtol=5e-2, atol=1e-3)
+            # test derivatives
+            obj1 = ObjectiveFunction(obj1)
+            obj3 = ObjectiveFunction(obj3)
+            obj1.build()
+            obj3.build()
+            g1 = obj1.jac_unscaled(obj1.x(coils))
+            g3 = obj3.jac_unscaled(obj3.x(coils))
+            np.testing.assert_allclose(g1, g3)
```
YigitElma, line 1135: Should this be true if one of the neighboring coils is about to change?

YigitElma (approved):
I am not sure about the behavior when the neighboring coils change, but I guess it is similar to how `jnp.min` or softmin behaves under similar conditions.

It looks good to me.

tests/test_objective_funs.py
ddudt: Yes I believe so. Both the original method and this new pruning should have the same behavior for these derivatives, which doesn't fully capture what would happen if the closest coil changes.

YigitElma:
If I get confused in the future again, here is a small example,
image
image

<details>

```python
from desc.objectives.utils import softmin
def fun(x):
    f = x**2
    g = x
    return softmin(jnp.array([f, g]), alpha=8)

x = jnp.linspace(-0.5, 1.5, 300)
y = jax.vmap(fun)(x)
dy = jax.vmap(jax.grad(fun))(x)
plt.title("softmin with alpha=8")
plt.plot(x, x, "--", label="g", lw=2)
plt.plot(x, x**2, "--", label="f", lw=2)
plt.plot(x, y, label="softmin(f, g)", lw=3)
plt.plot(x, dy, label="dy/dx")
plt.legend()

def fun(x):
    f = x**2
    g = x
    return jnp.min(jnp.array([f, g]))

x = jnp.linspace(-0.5, 1.5, 300)
y = jax.vmap(fun)(x)
dy = jax.vmap(jax.grad(fun))(x)
plt.title("min(f, g)")
plt.plot(x, x, "--", label="g", lw=2)
plt.plot(x, x**2, "--", label="f", lw=2)
plt.plot(x, y, label="min(f, g)", lw=3)
plt.plot(x, dy, label="dy/dx")
plt.legend()
```

<\details

PR #2141, opened 2026-03-27
Add utility function for general order connection btwn zernike and monomials of rho*cos/sin, fix bug when L>M for NAE constraints

dpanici:
- Fixes bug for NAE constraints when L>M where the full eq behavior for the given order rho would not be constrained (i.e. it would not include some of the modes, leadig to the behavior being incompletely constrained)
- Adds utility function for arbitrary order zernike -> given order rho behavior, which reduces redundant code.


dpanici:
Maybe just invert `zernike_radial_coeffs`?

PR #2143, opened 2026-03-29
[codex] Fix GX eik input handling and subprocess errors

rmchurch:
## What changed
- force generated GX inputs to use `geo_option = "eik"` so DESC's plain-text geometry files match the GX input mode
- wrap GX subprocess failures and timeouts in readable runtime errors that include `stdout.gx` / `stderr.gx` context
- add targeted tests for input rewriting and subprocess error reporting

## Why
The GX wrapper was generating a plain-text geometry file while allowing templates to keep `geo_option = "nc"`. That made GX fail with an opaque nonzero exit, and the callback path surfaced it as a raw subprocess failure instead of a useful GX-specific error.

## Impact
This makes the GX integration robust against NC-based templates and preserves actionable failure context when GX exits nonzero.

## Validation
- `/global/cfs/projectdirs/m4505/rchurchi/conda-envs/desc-gx/bin/python -m pytest tests/test_external_gx.py -q`

### DIFF
--- a/desc/external/gx.py
+++ b/desc/external/gx.py
@@ -5,6 +5,7 @@ import os
 import re
 import shutil
 import subprocess
+from collections import deque
 
 import numpy as np
 from interpax import interp1d
@@ -467,7 +468,9 @@ def gx(
     gx_input_file : str
         Path to a template GX TOML input file that specifies physics parameters
         (species, domain, time stepping, etc.). The ``geo_file`` path in this
-        template will be replaced with the generated geometry file.
+        template will be replaced with the generated geometry file, and
+        ``geo_option`` will be set to ``"eik"`` to match DESC's plain-text
+        geometry output.
     launch_cmd : list of str, optional
         Command prefix for launching GX, e.g.
         ``["srun", "-N", "1", "--gpus-per-task=1"]`` for SLURM GPU allocation or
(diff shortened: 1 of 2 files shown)


tests/test_external_gx.py
YigitElma, line 1: We actually don't require writing tests for the `desc.external` module. But these look good to me. We can decide to keep them or not in the actual PR for the `yge/gx` branch.

YigitElma (commented):
Thanks for this @rmchurch! I was naively expecting the user to put `eik`, but I guess people use the same template file for multiple cases.

If this is ready, I can approve/merge it to `yge/gx`, and create a GX wrapper PR later. Does this sound good to you?

desc/external/gx.py
```diff
@@ -603,6 +606,32 @@ def _write_gx_input(template_path, output_path, geo_path):
...
+    # generated input even if the template was copied from an NC-based workflow.
+    data, geo_option_subs = re.subn(
+        r"(^\s*geo_option\s*=\s*)(['\"])[^'\"]*\2",
+        r'\1"eik"',
+        data,
+        count=1,
+        flags=re.MULTILINE,
+    )
+    if geo_option_subs == 0:
+        data, geo_option_subs = re.subn(
```
YigitElma, line 619: In the other branch, I switched to using a loop to update the file content. This is mainly because regex operations are very hard to read, and some other developers wanted that.

@rmchurch Can you switch to the other method when resolving the merge conflicts?

YigitElma (changes requested):
Hi @rmchurch, if you can resolve the merge conflicts, I can merge this to the other branch and finish it up on the other branch.

PR #2147, opened 2026-03-30
Superficial changes for clarity in Bounce functions

dpanici:
- Just trying to add some comments for my own understanding of the bounce things, as well as notes for potential API simplifications in the future (not planning on doing any of these simplifications before we merge in #1919). Putting in separate PR for now, but will merge this comments/clarification soon and then will approve the bounce PR.
- No functionality changes here, just is my way of trying to increase my understanding of the code.

Comments for future consideration/improved usability:
- angle is a little clunky, it would be nice if it were  auto-computed and stored within the Bounce2D functional or within the objective,
	- maybe it is fine for the objective part, but when I compute a quantitiy outside of an objective it is unintuitive to compute angle first, separately. if angle is nonexistent it should be computed automatically, perhaps inside the Bounce2D creation...
- integrate should probably not require the user to reshape the data, it should take the data as is computed normally and reshape it internally. If someone wanted to use it within a script to compute some bounce integrals, it just adds one more step to remember for the process
- integrand fxns for ``Bounce2D.integrate`` cannot contain a vector quantity (at least not according to my one attempt), so if there ever were a use case where one wanted say a basis vector or vector B or something, it would not be possible. This is just a point for future improvement to note.


desc/integrals/_bounce_utils.py
```diff
@@ -729,6 +735,8 @@ def theta_on_fieldlines(angle, iota, alpha, num_transit, NFP):
         .swapaxes(0, -3)
     )
     alpha = alpha.swapaxes(0, -2)
+    # now the variable delta is actually referring to theta
+    # TODO: why not just rename? and why the axes?
```
unalmis: allocates more memory if i rename. keep the same name makes it inplace

desc/compute/_neoclassical.py
```diff
@@ -230,13 +230,15 @@ def fun(pitch_inv):
                 nufft_eps=nufft_eps,
                 is_fourier=True,
             )
+            # sum over wells, average over pitch
```
unalmis: average over alpha

desc/compute/_neoclassical.py
```diff
@@ -230,13 +230,15 @@ def fun(pitch_inv):
...
+            # output here is then (num_rho, num_alpha)
             return safediv(I_1**2, I_2).sum(-1).mean(-2)
 
         return jnp.sum(
             batch_map(fun, data["pitch_inv"], pitch_batch_size)
             * data["pitch_inv weight"]
             / data["pitch_inv"] ** 3,
-            axis=-1,
+            axis=-1,  # summed over alpha, so is only fxn of rho after
```
unalmis: summed over pitch

desc/integrals/bounce_integral.py
```diff
@@ -570,12 +711,17 @@ def batch(
     def reshape(grid, f):
         """Reshape arrays for acceptable input to ``integrate``.
 
+        Bounce2D.integrate requires that data be in the "rzt" shape
+        corresponding to shape (num ρ, num ζ, num θ). This method
+        is just a shortcut for ''grid.meshgrid_reshape(f, "rzt")''
```
unalmis: probably don't want to say this in case internal implementation changes in future. api abstaction

desc/integrals/_bounce_utils.py
```diff
@@ -314,6 +314,9 @@ def check_bounce_points(z1, z2, pitch_inv, knots, B, plot=True, **kwargs):
         "Second to last axis does not enumerate polynomials of spline. "
         f"Spline shape {B.shape}. Knots shape {knots.shape}."
     )
+    # TODO: should eventually be a passable argument, we should not expect users to
+    # potentially modify source code to use the bounce integrals if this assert ever
+    # gets triggered
     assert knots[0] > _sentinel, "Reduce sentinel in desc/integrals/_bounce_utils.py."
```
unalmis, line 320: they would run out of memory way b4 this could be reached, we don't want to add more parameters, the feedback was that makes using it difficult. 

The sentinel thing is documented in `Bounce1D`.

desc/integrals/bounce_integral.py
```diff
@@ -83,9 +148,10 @@ def get_pitch_inv_quad(min_B, max_B, num_pitch, simp=True):
         Parameters
         ----------
         min_B : jnp.ndarray
-            Minimum B value.
+            Minimum B value. May be a 1-D array, in which case the inverse pitch (1/λ)
+            values and weights will be returned for each pair of min_B and max_B values.
```
unalmis, line 157: this change makes the function confusing. the shape is specified in the output already. existing style was numpy style - e.g. numpy docs don't point out numpy style vectorization since thats assumed by every function

desc/integrals/bounce_integral.py
```diff
@@ -207,7 +276,12 @@ class Bounce2D(Bounce):
         Must include names in ``Bounce2D.required_names``.
     angle : jnp.ndarray
         Shape (num ρ, X, Y).
-        Angle returned by ``Bounce2D.angle``.
+        Stream-map angle returned by ``Bounce2D.angle``.
+        In the default pathway this is ``delta = theta - alpha`` in a
+        Fourier-Chebyshev representation used by ``theta_on_fieldlines`` to
+        reconstruct ``theta(zeta)`` on followed field lines.
+        This is required input to ``Bounce2D`` and is not computed inside
+        ``__init__`` so callers can precompute/reuse it across batched calls.
```
unalmis, line 289: These are implementation details that should not be placed in a public API.

desc/integrals/bounce_integral.py
```diff
@@ -660,6 +811,15 @@ def angle(
...
         """Return the angle for mapping boundary coordinates to field line coordinates.
 
+        Specifically, this angle by default is delta = theta - alpha,
+        where theta is the DESC poloidal angle and alpha is the field-line label.
+        This is returned on a grid in (rho,alpha,zeta). This is later used to
+        interpolate to any arbitrary (rho,alpha,zeta) through Fourier-Chebyshev
+        interpolation, effectively obtaining the map delta(rho,alpha,zeta) so one may
+        compute theta on a given field line (given by alpha) at a given point along
+        that field line (given by zeta) by simply calculating
+        theta = alpha + delta(rho,alpha,zeta).
```
unalmis, line 826: these are implementation details that should not be in the public API.

desc/integrals/bounce_integral.py
```diff
@@ -698,10 +858,20 @@ def angle(
...
-            errorif(not kwargs.pop("ignore_lambda_guard", False))
+            # the lambda stuff is deprecated and only used in tests
+            # TODO: just make a different function for the test, or just remove the
+            # lambda part entirely, which would simplify this. Can do so in a later
+            # version of DESC, leave for now.
+            errorif(
+                not kwargs.pop("ignore_lambda_guard", False),
+                "The 'lambda' option for the 'name' kwarg in Bounce2D.angle "
+                "is deprecated and should not be used If you are using this,"
+                " please update the test to use 'delta' instead.",
```
unalmis, line 878: The lambda option was never made publically available, so this is not necessary.

desc/integrals/bounce_integral.py
```diff
@@ -727,6 +897,10 @@ def angle(
             tol=tol,
             maxiter=maxiter,
         )
+
+        # [... is here if there is more than one flux surface,
+        # i.e. will either be [:,::-1] if only one flux surface or
+        # [:,:,::-1] if more than one flux surface.
```
unalmis, line 908: this is not true. size 1 axes are not squeezed out.

desc/integrals/bounce_integral.py
```diff
@@ -789,6 +963,14 @@ def points(self, pitch_inv, num_well=None):
             then the last axis, which enumerates bounce points for a particular field
             line and pitch, is padded with zero.
 
+        Dev Notes
+        -----
```
unalmis, line 972: This is implementation detail that should not be in public API. And this is python code, so the code itself is already basically just

if PiecwiseChebyshev:
    compute piecewise_checbyshev.intersects
else:
    compute spline intersects

desc/integrals/bounce_integral.py
```diff
@@ -814,6 +996,13 @@ def points(self, pitch_inv, num_well=None):
     def check_points(self, points, pitch_inv, *, plot=True, **kwargs):
         """Check that bounce points are computed correctly.
 
+            The checks are ensuring: that the bounce points are ordered
```
unalmis: This is implementation detail that should not be in public API.

desc/integrals/bounce_integral.py
```diff
@@ -885,12 +1074,28 @@ def integrate(
...
+        Axis convention (common broadcasted layout during quadrature):
+        - ``rho``: flux surface index.
+        - ``alpha``: field-line label index.
+        - ``pitch``: ``1/lambda`` sample index.
+        - ``well``: bounce-well index on a given field line and pitch.
+        - ``quad``: quadrature index on the mapped interval.
+
+        Typical intermediate tensor layout is ``(..., pitch, well, quad)``.
+        The final dot-product with quadrature weights contracts ``quad`` and
+        returns ``(..., pitch, well)``.
```
unalmis, line 1095: This is implementation detail that should not be in public API.

desc/integrals/bounce_integral.py
```diff
@@ -947,8 +1154,12 @@ def integrate(
         exclude = ("|B|", "B^zeta", "|e_zeta|r,a|", "zeta", "theta")
         data = setdefault(data, {})
         if is_fourier:
+            # remove the keys in exclude from the
+            # data dictionary
             data = apply(data, subset=names, exclude=exclude)
         else:
+            # remove the keys in exclude from the
+            # data dictionary and transform the rest to Fourier domain
```
unalmis, line 1171: is this not in the docstring of the called function

desc/integrals/bounce_integral.py
```diff
@@ -1296,7 +1509,7 @@ def plot_angle_spectrum(
             Shape (num ρ, X, Y).
             Angle returned by ``Bounce2D.angle``.
         l : int
-            Index into first axis of ``angle``.
+            Index into first axis of ``angle``, which typically is the rho-axis.
```
unalmis, line 1521: is there a time it is not

desc/integrals/bounce_integral.py
```diff
@@ -112,8 +183,11 @@ def get_pitch_inv_quad(min_B, max_B, num_pitch, simp=True):
         # Important to do an open quadrature since the bounce integrals at the
         # global maxima of |B| are not computable even ignoring precision issues.
         x, w = simpson2(num_pitch) if simp else uniform(num_pitch)
+        # change simp x from [−1, 1] ∋ x ↦ y ∈ [a, b] where a is min_B and b is max_B
```
unalmis, line 186: this is already in the docstring of bijection_from_disc

desc/integrals/bounce_integral.py
```diff
@@ -533,14 +670,21 @@ def batch(
...
             Data dictionary with the same structure as the data returned by the
-            functions in ``desc.compute``.
+            functions in ``desc.compute``. This dictionary should at a minimum contain
+            the data keys in ``Bounce2D.required_names``.
+            This dictionary is not modified by this method.
         angle : jnp.ndarray
             Shape (num rho, X, Y).
-            Angle returned by ``Bounce2D.angle``.
+            Stream-map angle delta used to find theta(zeta),
+            returned by ``Bounce2D.angle``.
```
unalmis, line 687: this is implementation detail

desc/integrals/bounce_integral.py
```diff
@@ -587,7 +731,8 @@ def reshape(grid, f):
         grid : Grid
             Tensor-product grid in (ρ, θ, ζ).
         f : jnp.ndarray
-            Data evaluated on grid.
+            Data evaluated on the grid in the canonical shape e.g.
+            (grid.num_nodes,) if 1-D and (grid.num_nodes, 3) if 2-D.
```
unalmis, line 735: this change suggests that vector valued inputs is supported, which is inconsistent with the output shape

desc/integrals/bounce_integral.py
```diff
@@ -624,6 +769,9 @@ def fourier(f):
         # Likewise to perform partial summation in this application, the real transform
         # must be done in the poloidal angle and the complex transform in the toroidal.
         f = rfft2(f, norm="forward").at[..., i].divide(2) * 2
+        # Here we get the correct coefficients by multiplying all by 2 and dividing
+        # the first and (potentially also the last,, if num θ is even)
+        # coefficients by 2.
```
unalmis, line 774: is this helpful? i guess the comment confirms we didn't compute the "incorrect" coefficients

desc/integrals/bounce_integral.py
```diff
@@ -814,6 +1004,14 @@ def points(self, pitch_inv, num_well=None):
     def check_points(self, points, pitch_inv, *, plot=True, **kwargs):
         """Check that bounce points are computed correctly.
 
+            The checks ensure: that the bounce points are ordered
+            correctly (z1 < z2 always where (z1, z2) is a bounce point pair),
+            that for each sequential pair of wells, z1_{i+1} >= z2_i (the start
+            of the next well is >= the end of the current well), and that
+            the straight-line path between two bounce points (z1, z2)
+            actually is in the epigraph of |B| (i.e. that |B(z)| <= 1/λ
+            for all z between z1 and z2).
```
unalmis, line 1013: implementation details should not be in public API

desc/integrals/bounce_integral.py
```diff
@@ -901,9 +1115,11 @@ def integrate(
             Real scalar-valued periodic functions in (θ, ζ) ∈ [0, 2π) × [0, 2π/NFP)
             evaluated on the ``grid`` supplied to construct this object.
             Use the method ``Bounce2D.reshape`` to reshape the data into the
-            expected shape.
+            expected shape for this function.
         names : str or list[str]
             Names in ``data`` to interpolate. Default is all keys in ``data``.
+            Names must correspond only to scalar-values quantities.
+            Vector-valued quantity interpolation is not currently supported
```
unalmis, line 1122: better to just add the assert statement to bounce.reshape rather than overload this information here.

desc/integrals/bounce_integral.py
```diff
@@ -975,6 +1195,8 @@ def integrate(
                 mask=flatten_mat(jnp.broadcast_to((z1 < z2)[..., None], z.shape), 4),
                 sentinel=0.5 * jnp.min(pitch_inv),
             )
+        # shapes of data are (num_rho, num_alpha, num_pitch, num_well, num_quad),
```
unalmis, line 1198: technically its (..., num alpha, num pitch, num well, num quad).

desc/integrals/bounce_integral.py
```diff
@@ -533,14 +670,21 @@ def batch(
         fun_data : dict[str, jnp.ndarray]
             Data to reshape, interpolate, and pass to ``fun``.
             The structure of the data should match the structure
-            returned by the registered compute functions in ``desc.compute``.
-            Note this dictionary will be modified.
+            returned by the registered compute functions in ``desc.compute`` e.g.
+            pointwise quantities are 1-D arrays of length grid.num_nodes and
+            vector quantities are 2-D arrays of shape (grid.num_nodes, 3).
```
unalmis, line 675: we don't support that input shape

desc/integrals/bounce_integral.py
```diff
@@ -479,6 +607,15 @@ def _defaults(eta, grid, **kwargs):
         angle = parse_argname_change(
             kwargs.get("angle", kwargs.get("theta", None)), kwargs, "theta", "angle"
         )
+        errorif(
+            angle is None,
+            ValueError,
+            "Got None for angle. Must pass in angle "
+            "as a kwarg when computing bounce-averaged quantities, with angle being "
+            "the output of Bounce2D.angle, which is the stream-map angle used to "
+            "reconstruct theta(zeta) on field lines. ",
```
unalmis, line 616: this is implementation detail that need not be publicized to users

desc/integrals/bounce_integral.py
```diff
@@ -1,4 +1,74 @@
...
+    ``angle`` (typically ``delta = theta - alpha`` from ``Bounce2D.angle``).
+4. Construct a 1D representation of ``B(zeta)`` along each followed field line
+    (piecewise Chebyshev or cubic spline, depending on configuration).
+5. Find bounce points ``(z1, z2)`` by solving ``B(zeta) = 1/lambda``.
+6. Map quadrature nodes to each bounce interval, interpolate required quantities,
+    apply geometry factors, and contract over quadrature weights.
+7. Return per-well bounce integrals, then reduce/average in callers as needed.
+
+See "effective ripple 3/2" in ``desc/compute/_neoclassical.py`` for an example
+of how the Bounce2D method is used in a compute function, and
```
unalmis, line 68: i already say this in the Bounce1D2D docstrings

desc/integrals/bounce_integral.py
```diff
@@ -1,4 +1,74 @@
...
+    Field-line-following toroidal coordinate; the independent variable that
+    parameterizes position along a field line.
+δ (delta)
+    Stream-map angle, δ = θ − α.  Because δ varies smoothly along a field
+    line, it is used instead of θ to build the field-line parameterization
+    θ(ζ) via ``Bounce2D.angle``.
+ι (iota)
+    Rotational transform (normalized by 2π); relates toroidal and poloidal
+    winding of a field line.  Used in ``theta_on_fieldlines`` and in the
+    upper bound on the number of magnetic wells per toroidal transit.
```
unalmis, line 25: the above is defined in the desc/compute/data index, with all the proper units etc. so redefining all that here just increases the maintenance surface.

desc/integrals/bounce_integral.py
```diff
@@ -1,4 +1,74 @@
...
+B^zeta
+    Contravariant toroidal component of the magnetic field vector in
+    (ρ, θ, ζ) coordinates; used as the Jacobian-like factor when converting
+    the line-element dℓ to an integral over ζ.
+Bref
+    Reference magnetic field strength used for non-dimensionalization.
+Lref
+    Reference length scale used for non-dimensionalization.
+xₖ, wₖ
+    Quadrature nodes and weights for approximating ∫₋₁¹ g(x) dx ≈ ∑ₖ wₖ g(xₖ).
```
unalmis, line 48: Aren't these all defined in the Bounce1d2d docs

unalmis (changes requested):
so it seems that the changes are mostly adding an internal implementation detail describing the `angle` to the public docstrings. That goes against API abstraction, and makes it more difficult to parse out the things I actually need to read to know how to use the functions

desc/integrals/bounce_integral.py
unalmis: a stream map is something which is constant along field lines. theta, zeta \mapsto alpha(theta, zeta) is a stream map. theta(alpha, zeta) is the corresponding inverse stream map. users probably don't need to know this.

unalmis:
> Comments for future consideration/improved usability:
> 
>     angle is a little clunky, it would be nice if it were auto-computed and stored within the Bounce2D functional or within the objective,
>         maybe it is fine for the objective part, but when I compute a quantitiy outside of an objective it is unintuitive to compute angle first, separately. if angle is nonexistent it should be computed automatically, perhaps inside the Bounce2D creation...

ideally, it would be computable quantity in data_index and then simply a required_names just like "|B|". But i think i got push back when I wanted to make map coordinates not require an equilibrium

>     integrate should probably not require the user to reshape the data, it should take the data as is computed normally and reshape it internally. If someone wanted to use it within a script to compute some bounce integrals, it just adds one more step to remember for the process

These functions do not assume you have a block of data on every surface simaltenously therefore, the input data must be reshaped.

>     integrand fxns for Bounce2D.integrate cannot contain a vector quantity (at least not according to my one attempt), so if there ever were a use case where one wanted say a basis vector or vector B or something, it would not be possible. This is just a point for future improvement to note.

Interpolating a vector requires more care to ensure it retains vector invariances such as div free, curl free, rotation etc. Therefore, it is less accurate to interpolate a vector-valued quantity and we don't do it.

unalmis:
closing since i addressed everything in the sparse pr aexcept for anle which has its own issie

PR #2148, opened 2026-03-31
Write auto generated input file to `tmp`

YigitElma:
Resolves #2139 

`InputReader.parse_inputs` method now uses an in-memory buffer if it has to convert a vmec to a desc input, resulting in no files being written.


desc/input_reader.py
```diff
@@ -215,7 +216,9 @@ def parse_inputs(self, fname=None):  # noqa: C901
             isVMEC = re.search(r"&INDATA", line, re.IGNORECASE)
             if isVMEC:
                 print("Converting VMEC input to DESC input")
-                path = self.input_path + "_desc"
+                basename = os.path.basename(self.input_path) + "_desc"
+                tmpdir = tempfile.mkdtemp()
```
mishapadidar, line 220: When does the delete operation occur on tmpdir? Does it do so automatically when the function ends?
Another option is to use `with`, i.e.
```
with tempfile.TemporaryDirectory() as tmpdir:
    path = os.path.join(tmpdir, basename = os.path.basename(self.input_path) + "_desc")
```
though it could be essentially the same.

mishapadidar:
When I ran the test, the temp directory still existed after the code ran.

desc/input_reader.py
YigitElma: I think it should be cleared after the python process is terminated or the session on the cluster is closed or the system is rebooted on personal laptop, not immediately after the function call. These are the options listed on the web for "when the `tmp` is cleared?".

YigitElma:
Add a flag for tmp file option

YigitElma:
Try to pass a file-like object

dpanici:
> Try to pass a file-like object

Did we want to try this?

YigitElma:
> Did we want to try this?

In a previous dev meeting, we said it doesn't worth the effort.

dpanici:
> > Did we want to try this?
> 
> In a previous dev meeting, we said it doesn't worth the effort.

I was bored and did this. If we decide we don't like it we can revert but was surprisingly easy and clean.

dpanici (dismissed):

YigitElma:
Dario's approval is on behalf of me.

ddudt (dismissed):
[question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
Now if I want to convert a VMEC input file to a DESC input file, do I have to explicitly call `InputReader.vmec_to_desc_input()` myself? If so, that might be good to document in the Changelog, since I would often use the command-line interface to do that conversion implicitly.

CHANGELOG.md
```diff
@@ -56,6 +56,17 @@ v0.17.1
 
 Bug Fixes
 
+- Fixes the possible permission error in `from_input_file` method of `Equilibrium`, `FourierRZToroidalSurface` and `FourierRZCurve` classes when used with a VMEC input file. Now the automatically generated DESC input file is written to a temporary file under `./tmp/`.
```
YigitElma: We should move this above.

ddudt (dismissed):

CHANGELOG.md
```diff
@@ -8,8 +8,8 @@ Performance Improvements
 Bug Fixes
 
 - Fixes bug that was always setting NFP=1 in ``to_FourierRZ`` methods.
-- Fixes ``VMECIO.save`` metadata for current-density variables and corrects the
-  asymmetric ``currvmns`` magnetic-axis extrapolation.
+- Fixes ``VMECIO.save`` metadata for current-density variables and corrects the asymmetric ``currvmns`` magnetic-axis extrapolation.
+- Fixes the possible permission error in `from_input_file` method of `Equilibrium`, `FourierRZToroidalSurface` and `FourierRZCurve` classes when used with a VMEC input file. Now the automatically generated DESC input file is written to a temporary file under `./tmp/`.
```
YigitElma: Update this

YigitElma:
Make this work so that the CLI run creates DESC input file for VMEC stuff

dpanici:
> Make this work so that the CLI run creates DESC input file for VMEC stuff

Done. Now if doing ``python -m desc input.vmec`` the ``input.vmec_desc`` converted file will be created, as it does on master still. But if one is using from a script, input reader or from_input_file wont write the files.

YigitElma:
This looks good to me. @dpanici You can approve on my behalf.

ddudt (approved):

dpanici (approved):

PR #2153, opened 2026-04-09
Minor changes from MPI branch

### LINKED ISSUE #2165: test_proximal_scalar fails the CI half the time

YigitElma:
Some changes that were required to make MPI work, and I think are useful apart from MPI.

- Adds a helper function for all different compute methods (similar to jvp, jac wrappers we have)
- Adds `use_jit_wrapper` logic, in case any of the sub-objectives shouldn't be compiled or they shouldn't be compiled together (as in the case of MPI)
- Moves `constants` assignment to proper place (this was causing a bug in MPI PR, don't remember why exactly though)
- Adds jit decorator to `ObjectiveFunction.unpack_params()` (this makes the difference in benchmarks, specifically [this line](https://github.com/PlasmaControl/DESC/blob/7238beae389ea906fac5ff90bc3a607af051dcc7/desc/objectives/utils.py#L156)) and to `ObjectiveFunction.x()`. The substantial speed-up (~1.5s) comes from the compilation overhead, but even after that, each call is 3-4 times faster (but these take couple milliseconds)

Resolves #2165


YigitElma:
`use_jit_wrapper` allows the following:
```python
eq = get("precise_QA")

class DummyObj(_Objective):
    def __init__(self, eq, grid=None):
        if grid is None:
            grid = LinearGrid(M=eq.M, N=eq.N, NFP=eq.NFP)
        self._grid = grid
        super().__init__(things=eq)

    def build(self, use_jit, verbose=3):
        print("build called")
        self._dim_f = 1
        self.transforms = get_transforms(["|B|"], self.things[0], self._grid)
        self.profiles = get_profiles(["|B|"], self.things[0], self._grid)

        print("fun called")
        data = compute_fun(
            self.things[0],
            ["|B|"],
            params=params,
            transforms=self.transforms,
            profiles=self.profiles,
        )
        f = data["|B|"].max()
        if f > 0:
            f += data["|B|"].min()
            print(f"{f}")
        return f


obj1 = DummyObj(eq)
obj1.build(use_jit=False)
obj = ObjectiveFunction((ForceBalance(eq), obj1))
obj.build(use_jit=True)
obj.compute_scalar(obj.x(eq))
obj.grad(obj.x(eq))
```
```
build called
Building objective: force
Precomputing transforms
fun called
2.0040131258026577
fun called
LinearizeTracer<float64[]>
...
```

desc/objectives/objective_funs.py
```diff
@@ -657,6 +651,7 @@ def print_value(self, x, x0=None, constants=None):
                     out[obj._print_value_fmt] = [outi]
         return out
 
+    @functools.partial(jit, static_argnames="per_objective")
```
YigitElma, line 660: Looks like this is making all the speed difference. Kind of unexpected because this function is already jitted by outer functions.

YigitElma:
Might need to unjit helpers of `ProximalProjection` to work properly `use_jit_wrapper=False` case

YigitElma:
Add a warning for `use_jit_wrapper`

desc/objectives/objective_funs.py
YigitElma, line 383: There is no easy way to allow individual objectives to be jitted differently other than building them before calling this function. So, something like
```python
obj1 = ForceBalance(eq)
obj2 = AspectRatio(eq)
obj1.build(use_jit=True)
obj2.build(use_jit=False)
obj = ObjectiveFunction([obj1, obj2])
# skips individual builds, doesn't jit obj methods
# but obj1 is still jitted
obj.build() 
```

CHANGELOG.md
```diff
@@ -1,11 +1,21 @@
 Changelog
 =========
 
+
+New Features
+
+- Sub-objectives of an `ObjectiveFunction` can now have different `use_jit` values than the `ObjectiveFunction`. These objectives has to be built before building the `ObjectiveFunction`.
```
ddudt: typo: has -> have

desc/objectives/objective_funs.py
```diff
@@ -359,6 +359,19 @@ def build(self, use_jit=None, verbose=1):
...
+        use_jits = [obj._use_jit for obj in self.objectives]
+        if not all(use_jits):
+            if self._use_jit:
+                warnif(
+                    True,
+                    UserWarning,
+                    "At least 1 sub-objective has use_jit=False. Setting "
+                    "use_jit=False for the whole ObjectiveFunction. "
+                    f"Sub-objective use_jit values: {use_jits}",
+                )
```
ddudt:
```suggestion
            warnif(
                self._use_jit,
                UserWarning,
                "At least 1 sub-objective has use_jit=False. Setting "
                "use_jit=False for the whole ObjectiveFunction. "
                f"Sub-objective use_jit values: {use_jits}",
            )
```

tests/test_objective_funs.py
```diff
@@ -4296,3 +4296,38 @@ def test_coil_objective_setter(DummyMixedCoilSet):
...
+
+    rs = [-2, 4]
+    for r in rs:
+        eq.R_lmn[0] = r
+        f = obj.compute_scalar(obj.x(eq))
+        g = obj.grad(obj.x(eq))
+
+        assert f == 2 * r**2 if r > 0 else 9 * r**2 / 2
+        assert g[0] == 4 * r if r > 0 else 9 * r
+        assert np.all(g[1:] == 0)
```
ddudt, line 4378: Is this just checking that the objective values and derivatives are correct when it is not JITed? Doesn't feel necessary but doesn't hurt.

ddudt (commented):
Minor comments/suggestions, but otherwise this looks good.

tests/test_objective_funs.py
YigitElma: Yes, just a sanity check that jittability is not necessary for AD

ddudt (dismissed):

desc/optimize/_constraint_wrappers.py
```diff
@@ -1310,7 +1310,25 @@ def __getattr__(self, name):
...
+def jit_if_possible(func):
+    """Jit a function if use_jit."""
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        # this has to be ObjectiveFunction
+        obj = args[0]
+        if not getattr(obj, "_use_jit", False):
+            # Apply jit if we should
+            jitted_func = functools.partial(jit, static_argnames=["op"])(func)
```
f0uriest: Won't this cause recompilation every time?

YigitElma: I tested this with following,
```python
eq = desc.examples.get("ATF")
grid = LinearGrid(M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, rho=np.linspace(0.1, 1, 10))
objective = ObjectiveFunction(QuasisymmetryTwoTerm(eq, grid=grid))
constraint.build(use_jit=True)
prox = ProximalProjection(
    objective, constraint, eq, solve_options={"solve_during_proximal_build": False}
)
prox.build()
x = prox.x(eq)
with jax.log_compiles():
    _ = prox.jac_scaled_error(x, prox.constants).block_until_ready()
with jax.log_compiles():
    _ = prox.jac_scaled_error(x, prox.constants).block_until_ready()
```

The first one prints
```
...
WARNING:2026-04-16 16:20:36,269:jax._src.dispatch:206: Finished jaxpr to MLIR module conversion jit(_proximal_jvp_f_pure) in 0.525352478 sec
WARNING:2026-04-16 16:20:41,081:jax._src.dispatch:206: Finished XLA compilation of jit(_proximal_jvp_f_pure) in 4.811319828 sec
```

Second one doesn't print anything. This might be some `functools` trick, but I am not sure.

I will push a safer version, and I also realized that the if condition is reverse.

YigitElma:
Btw due to the previous bug in if statement (not jitting the proximal objectives), some of the benchmarks were misrepresenting. But it is suprising that unjitted bounce stuff is faster and more memory efficient than jitted counterparts.

For example, benchmarks runs from [this commit](https://github.com/PlasmaControl/DESC/actions/runs/24378570779/job/71325836772)
image

YigitElma:
I ran a couple more benchmarks, and confirmed that all the speed up comes from this line https://github.com/PlasmaControl/DESC/blob/7238beae389ea906fac5ff90bc3a607af051dcc7/desc/objectives/utils.py#L156 

If I add the `jit` inline, I get the same speed up. Weirdly, this line was taking ~0.8s on master on CPU (~1.4s on GPU), with `jit` it takes ~0.06s on CPU (~0.09s on GPU). Given that this function only does basic splitting operations, it is weird.

dpanici:
> I ran a couple more benchmarks, and confirmed that all the speed up comes from this line
> 
> https://github.com/PlasmaControl/DESC/blob/7238beae389ea906fac5ff90bc3a607af051dcc7/desc/objectives/utils.py#L156
> 
> If I add the `jit` inline, I get the same speed up. Weirdly, this line was taking ~0.8s on master on CPU (~1.4s on GPU), with `jit` it takes ~0.06s on CPU (~0.09s on GPU). Given that this function only does basic splitting operations, it is weird.

this explains the bounce things too?

desc/objectives/objective_funs.py
```diff
@@ -359,6 +359,18 @@ def build(self, use_jit=None, verbose=1):
...
             self._use_jit = use_jit
+
+        use_jits = [obj._use_jit for obj in self.objectives]
+        if not all(use_jits):
+            warnif(
+                self._use_jit,
+                UserWarning,
+                "At least 1 sub-objective has use_jit=False. Setting "
+                "use_jit=False for the whole ObjectiveFunction. "
+                f"Sub-objective use_jit values: {use_jits}",
```
dpanici: we should be able to print out which objective is lacking the jit right?

dpanici:
```suggestion
                f"Sub-objectives with use_jit=False: {[o for o in self.objectives if o._use_jit]}",
```

(maybe check my list comprehension logic there)

dpanici (commented):
Looks fine but where do things fail if the sub obejctives with use_jit=False are not built beforehand? Would prefer to have a warning when that occurs saying to build them first.

YigitElma:
> this explains the bounce things too?

Bounce memory improvement was just an artifact of not jitting the Jacobian computation (it was caused by a bug in `jit_if_possible`, it was actually jitting if not possible :) ). Why the compilation causes higher memory usage is a different question which I don't want to dig into.

> where do things fail if the sub obejctives with use_jit=False are not built beforehand?

With the current setup, objectives don't have the `use_jit` attribute until they are built (we have [this](https://github.com/PlasmaControl/DESC/blob/4b1c43b348e2e6ca8ba750948ffccdb3714f4f7f/desc/objectives/objective_funs.py#L1174) but almost never gets used), so user needs to build them separately if they shouldn't be jitted. `ObjectiveFuntion.build()` builds all sub-objectives with the same `use_jit` value if they are not built, and the default is true. It doesn't use the default of the sub-objective (for example, linear constraints usually have `use_jit=False` as default, but that doesn't apply on the `ObjectiveFuntion.build()` path).

dpanici (approved):

ddudt (approved):

PR #2156, opened 2026-04-09
Improve import time

### LINKED ISSUE #2154: Import time is too long
Importing something from DESC takes too much time. Specifically the 
```python
import desc.compute
```

Script to see individual contributions:
```
python -X importtime -c "import desc.compute"
```
This takes more than 5.5 seconds, and 3.5 seconds of it is just `_build_data_index`. `desc.integrals` is also taking around 2 seconds.

[import-time.txt](https://github.com/user-attachments/files/26609765/import-time.txt)

YigitElma:
- Replaces the recursive tree construction by a depth-first search algorithm in `desc.compute`
- Removes the `default_quad` computation at import time.

Importing `desc.compute` takes a very long time, and on my system, only `_build_data_index` itself takes 3.5 seconds. This change makes it run in 0.15 seconds!

The previous function was slow because of recursion, but it also had an extreme amount of redundant calls. For example, `get_params`, `get_profiles`, and `get_derivs` call `get_deps` again internally, and since the result of the first call is not stored yet, it rebuilds the whole tree again, almost 8 more times.

New `_build_data_index` was originally written by Claude Code (Opus 4.6), but I wrote most of the comments and I understand the algorithm behind.

Resolves #2154

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ uv.lock
 # Environments
 .env
 .venv
+
+.vscode/
--- a/desc/backend.py
+++ b/desc/backend.py
@@ -54,10 +54,10 @@ def print_backend_info():
     if use_jax:
         print(
             f"Using JAX backend: jax version={jax.__version__}, "
-            + f"jaxlib version={jaxlib.__version__}, dtype={x.dtype}."
+            + f"jaxlib version={jaxlib.__version__}, dtype={y.dtype}."
         )
     else:
-        print(f"Using NumPy backend: version={np.__version__}, dtype={x.dtype}.")
+        print(f"Using NumPy backend: version={np.__version__}, dtype={y.dtype}.")
     print(
         "Using device: {}, with {:.2f} GB available memory.".format(
             desc_config.get("device"), desc_config.get("avail_mem")
--- a/desc/backend.py
+++ b/desc/backend.py
@@ -36,13 +36,11 @@ else:
                     + "installed JAX with GPU support?"
                 )
                 set_device("cpu")
-            x = jnp.linspace(0, 5)
-            y = jnp.exp(x)
+            x = jnp.arange(2)
         use_jax = True
     except ModuleNotFoundError:
         jnp = np
-        x = jnp.linspace(0, 5)
-        y = jnp.exp(x)
+        x = jnp.arange(2)
         use_jax = False
         set_device(kind="cpu")
         warnings.warn(colored("Failed to load JAX", "red"))
@@ -54,10 +52,10 @@ def print_backend_info():
     if use_jax:
         print(
             f"Using JAX backend: jax version={jax.__version__}, "
-            + f"jaxlib version={jaxlib.__version__}, dtype={y.dtype}."
+            + f"jaxlib version={jaxlib.__version__}, dtype={x.dtype}."
         )
     else:
-        print(f"Using NumPy backend: version={np.__version__}, dtype={y.dtype}.")
+        print(f"Using NumPy backend: version={np.__version__}, dtype={x.dtype}.")
     print(
         "Using device: {}, with {:.2f} GB available memory.".format(
             desc_config.get("device"), desc_config.get("avail_mem")
--- a/desc/backend.py
+++ b/desc/backend.py
@@ -36,11 +36,13 @@ else:
                     + "installed JAX with GPU support?"
                 )
                 set_device("cpu")
-            x = jnp.arange(2)
+            x = jnp.linspace(0, 5, 2)
+            y = jnp.exp(x)
         use_jax = True
     except ModuleNotFoundError:
         jnp = np
-        x = jnp.arange(2)
+        x = jnp.linspace(0, 5, 2)
+        y = jnp.exp(x)
         use_jax = False
         set_device(kind="cpu")
         warnings.warn(colored("Failed to load JAX", "red"))
--- a/desc/compute/__init__.py
+++ b/desc/compute/__init__.py
@@ -139,6 +139,10 @@ def _build_data_index():  # noqa: C901
         # transitive data deps of key K is just:
         #   union of (each direct dep D) + (D's already-cached full data deps)
         # No recursion needed — O(number of direct deps) per key.
+
+        # The deps are stored in topological order (not alphabetical) so that
+        # iterating over them computes quantities in valid dependency order.
+        topo_index = {key: i for i, key in enumerate(order)}
         for key in order:
             d = data_index[p][key]
             deps_info = d["dependencies"]
@@ -150,7 +154,7 @@ def _build_data_index():  # noqa: C901
             for dep in direct_data:
                 full_data_set.add(dep)
                 full_data_set.update(data_index[p][dep]["full_dependencies"]["data"])
-            deps_no_axis = sorted(full_data_set)
+            deps_no_axis = sorted(full_data_set, key=topo_index.__getitem__)
 
             transforms, params, profiles = _collect_deps(p, [key] + deps_no_axis)
             full = {
@@ -177,7 +181,7 @@ def _build_data_index():  # noqa: C901
                 full_data_axis_set.update(
                     data_index[p][dep]["full_with_axis_dependencies"]["data"]
                 )
-            deps_with_axis = sorted(full_data_axis_set)
+            deps_with_axis = sorted(full_data_axis_set, key=topo_index.__getitem__)
 
             if len(deps_no_axis) >= len(deps_with_axis):
                 # This quantity and all its dependencies do not need anything
--- a/desc/compute/__init__.py
+++ b/desc/compute/__init__.py
@@ -23,8 +23,7 @@ data : dict of ndarray
 
 """
 
-# just need to import all the submodules here to register everything in the
-# data_index
+import numpy as np
 
 from ..utils import rpz2xyz, rpz2xyz_vec, xyz2rpz, xyz2rpz_vec
 from . import (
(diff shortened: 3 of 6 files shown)


desc/integrals/bounce_integral.py
```diff
@@ -301,7 +295,16 @@ def __init__(
...
-        self._quad = get_quadrature(setdefault(quad, default_quad), automorphism)
+        self._quad = get_quadrature(
+            setdefault(
+                quad,
+                get_quadrature(
+                    leggauss(32),
+                    (automorphism_sin, grad_automorphism_sin),
+                ),
+            ),
+            automorphism,
```
unalmis: - This will now solve an eigenvalue problem every time `Bounce*D` is called. 
- Autodiff will also trace it into the graph as well now, and JIT is imperfect and does not work well inside scan loops, which is where this stuff is called.
For error analysis, it is vital that the quadrature is not differentiated, and leaving that up to JIT under scan is not something I want given how often bugs are found in it. thus, this change makes it harder to confirm correctness now.
- On the other hand, what I suggested in the original issue of just changing the default_quad defined in this file globally avoids an eigenvalue solve on import.

desc/integrals/bounce_integral.py
```diff
@@ -1468,7 +1471,13 @@ def __init__(
     ):
         """Returns an object to compute bounce integrals."""
         assert grid.is_meshgrid
-        quad = setdefault(quad, default_quad)
+        quad = setdefault(
+            quad,
+            get_quadrature(
+                leggauss(32),
+                (automorphism_sin, grad_automorphism_sin),
+            ),
```
unalmis: https://github.com/PlasmaControl/DESC/pull/2156/changes#r3062003714

desc/backend.py
```diff
@@ -36,13 +36,11 @@
                     + "installed JAX with GPU support?"
                 )
                 set_device("cpu")
-            x = jnp.linspace(0, 5)
-            y = jnp.exp(x)
+            x = jnp.arange(2)
```
unalmis: i don't know what the purpose of this change is, but generally you want to do math to test this stuff is working. arange and the like can often execute on host on old jax (< 0.9) versions

unalmis (dismissed):
my suggestion on the issue avoids an eigenvalue solve on import. current changes force an eigenvalue solve every time an object is made, which is worse for optimization and debugging with jit in general. i didn't check the claude stuff

desc/integrals/bounce_integral.py
YigitElma: https://github.com/PlasmaControl/DESC/issues/2154#issuecomment-4216548470
This was my way of making the default_quad required(ish) argument; if someone wants to avoid this computation, they can pass a precomputed value. If you want to make it actually required, then you can make the change. My point is, there shouldn't be any computation on the import; not everyone uses this function.

desc/backend.py
YigitElma: I think to debug any problem with `jax`, any computation that triggers XLA is enough, at least until now I've never seen otherwise. Pure imports don't show actual problems, but any `jnp` function does the job. I just made the change to get that check minimally. I don't have a strong opinion about it, but I prefer something that is minimal.

desc/integrals/bounce_integral.py
unalmis: >  if someone wants to avoid this computation, they can pass a precomputed value.

This is false. Your changes force an eigenvalue solve every time Bounce*D is called

unalmis: - Trivial computations like `jnp.cos(jnp.arange(32))` on import are done by the packages we import like numpy etc. For example they load up things like powers of two into cache just to name one. unlike avoiding an eigenvalue solve, avoiding calling a cosine function is probably an unnecessary micro-optimization.
- jnp.cos(jnp.arange(32)) probably consumes comparable time to defining a python function in some file. So it doesn't make sense to me to single that out as a computation. everything is machine code; everything is a computation.

- Previously, users could import `default_quad`, a public variable, into their scripts. They can no longer do that with your change. User code will thus break.
- very simple solution is to just use a the default quad that avoids eigenvalue solve as mentioned in original issue.

f0uriest: The main cost of `leggauss(32)` is the jit compiler overhead. The actual calculation takes ~1ms. So one option (if we want to keep using legendre for default quad) could be to just use the original `np.polynomial.legendre.leggauss` for `default_quad`, which avoids the jit overhead at init (and since its just a constant it doesn't really need to be jax anyways). switching to chebyshev quadrature avoids the eigenvalue solve but may still have some jit overhead leading to similar issues. (although in either case a pure np version may be faster). I'll defer to kaya on the accuracy impact of legendre vs chebyshev

desc/backend.py
f0uriest: I'd prefer to keep the old version. As kaya mentioned, older versions of jax will always execute `arange` on the host meaning GPU issues won't necessarily show up.

desc/compute/__init__.py
```diff
@@ -55,42 +54,160 @@
...
+        return transforms, params, profiles
 
     for p in data_index:
-        for key in data_index[p]:
+        # --- Step 1: Topological sort via iterative Depth-First Search ---
+        # We need to process quantities with no dependencies before the
+        # quantities that depend on them. This way, when we process key K,
+        # all of K's dependencies already have their full_dependencies and
+        # full_with_axis_dependencies cached, and we can build K's full
+        # dependency set with a simple set union instead of deep recursion.
```
f0uriest, line 110: This sounds like it's basically doing what `set_tier` is doing below, so ideally we could do them both at the same time.

YigitElma: Yeah, I was actually gonna suggest that in looped compute PR. The order obtained at the end can be used directly. That is why I have special sort instead of normal sort.

desc/integrals/bounce_integral.py
unalmis: accuracy is negligible. does jnp.asarray(numpy computation) have jit overhead? don't want issues like this where jax bakes in the numpy constant: https://github.com/PlasmaControl/DESC/blob/2d9d35a29fa9c1e3e9158ced059ffc8932fa3970/desc/integrals/surface_integral.py#L285

ddudt: The only place that `default_quad` is ever used is to set the default value in `Bounce1D` and `Bounce2D` if `quad=None`. This rarely happens in practice, since the compute functions always pass a value for `quad` when creating instances of these classes. 

My preference is to remove it from the import overhead and create it in the `Bounce*D.__init__` methods only when needed. Something like this:
```
if quad is None:
    quad = leggauss(32)
if automorphism is None:
    automorphism = (automorphism_sin, grad_automorphism_sin)
self._quad = get_quadrature(quad, automorphism)
```
Or document those as the default values and then we only need:
```
self._quad = get_quadrature(quad, automorphism)
```
That should only add the extra compute in the rare instances when the user doesn't specify `quad` themselves. 

I'm also fine with switching to `chebgauss1` or the numpy version if the accuracy is negligible. But I still think it should happen inside the `Bounce*D` classes instead of on import, since it mostly likely never gets used. I'm not worried about removing `default_quad` from the "public API" since it is not publicly documented anyways.

unalmis: Daniel code will break convergence by applying automorphism to all input quadrature.

I will be fine with this:
```python
if quad is None:
    quad = jax.lax.stop_gradient(get_quadrature(leggauss(32), (automorphism_sin, grad_automorphism_sin)))
elif automorphism is not None:
    quad = get_quadrature(quad, automorphism))
self._quad = quad # or put stop grad here
```

unalmis: yea its probably better to always call stop grad to also prevent loss of numerical precision since differentiating the quadrature makes these integrals non-integrable, and we don't want to rely on floating point math to cancel non-integrable singularity

```python
if quad is None:
    quad = get_quadrature(leggauss(32), (automorphism_sin, grad_automorphism_sin))
elif automorphism is not None:
    quad = get_quadrature(quad, automorphism)
self._quad = jax.lax.stop_gradient(quad)
```

we had to call stop grad to avoid floating point issues in stability stuff too

unalmis: leggaus works on nonsingular stuff too so its more dummy proof for users so keep that

desc/integrals/bounce_integral.py
```diff
@@ -301,7 +295,14 @@ def __init__(
         is_reshaped = is_reshaped or is_fourier
         vander = setdefault(vander, {})
 
-        self._quad = get_quadrature(setdefault(quad, default_quad), automorphism)
+        if quad is None:
+            quad = get_quadrature(
+                leggauss(32), (automorphism_sin, grad_automorphism_sin)
+            )
+        elif automorphism is not None:
```
unalmis:
```suggestion
        else:
````

File: desc/integrals/bounce_integral.py [hunk 6]
```diff
@@ -1468,9 +1469,13 @@ def __init__(
...
         """Returns an object to compute bounce integrals."""
         assert grid.is_meshgrid
-        quad = setdefault(quad, default_quad)
-
-        self._quad = get_quadrature(quad, automorphism)
+        if quad is None:
+            quad = get_quadrature(
+                leggauss(32), (automorphism_sin, grad_automorphism_sin)
+            )
+        elif automorphism is not None:
```
unalmis:
```suggestion
        else:
````

**Review by unalmis — APPROVED** [review]

**Comment by unalmis** [comment]
ididn't check calude stuff

File: .gitignore [hunk 7]
```diff
@@ -30,3 +30,5 @@ uv.lock
 # Environments
 .env
 .venv
+
```
ddudt, line 33: accidental line break?

ddudt (approved):

.gitignore
YigitElma: I didn't want to put it directly under the environment tab

desc/compute/__init__.py
YigitElma: https://github.com/PlasmaControl/DESC/pull/1557#discussion_r3068395627

dpanici (approved):

YigitElma:
I will merge this after the dev meeting, in case @f0uriest has some additional comments

YigitElma:
As a final check, I used this. Save both `data_index` objects in master and this PR (I have to remove the `fun` since it cannot be pickled.
```python
import pickle

data_index_new = data_index.copy()

for p in data_index_new:
    for key in data_index_new[p]:
        data_index_new[p][key].pop("fun")
output = open("data_index_pr.pkl", "wb")

# Pickle dictionary using protocol 0.
pickle.dump(data_index, output)
output.close()
```
Then compare them.
```python
pkl1 = open("data_index_master.pkl", "rb")
pkl2 = open("data_index_pr.pkl", "rb")

data1 = pickle.load(pkl1)
data2 = pickle.load(pkl2)

def norm(full):
    # sort alphabetically (since PR sorts in topo order)
    return {
        "data": sorted(full["data"]),
        "params": sorted(full["params"]),
        "profiles": sorted(full["profiles"]),
        "transforms": {k: sorted(v) for k, v in full["transforms"].items()},
    }

for p in data1:
    if p not in data2:
        print(f"{p} not in pr data_index")
        continue
    for k in data1[p]:
        if k not in data2[p]:
            print(f"{p}-{k} not in pr data_index")
            continue
        fd1 = norm(data1[p][k]["full_dependencies"])
        fd2 = norm(data2[p][k]["full_dependencies"])
        assert fd1 == fd2, f"full_dependencies differ for {p}-{k}"
        fa1 = norm(data1[p][k]["full_with_axis_dependencies"])
        fa2 = norm(data2[p][k]["full_with_axis_dependencies"])
        assert fa1 == fa2, f"full_with_axis_dependencies differ for {p}-{k}"

for p in data2:
    if p not in data1:
        print(f"{p} not in master data_index")
        continue
    for k in data2[p]:
        if k not in data1[p]:
            print(f"{p}-{k} not in master data_index")
            continue
        fd1 = norm(data1[p][k]["full_dependencies"])
        fd2 = norm(data2[p][k]["full_dependencies"])
        assert fd1 == fd2, f"full_dependencies differ for {p}-{k}"
        fa1 = norm(data1[p][k]["full_with_axis_dependencies"])
        fa2 = norm(data2[p][k]["full_with_axis_dependencies"])
        assert fa1 == fa2, f"full_with_axis_dependencies differ for {p}-{k}"

pkl1.close()
pkl2.close()
```

And this passes.

YigitElma:
> I will merge this after the dev meeting, in case @f0uriest has some additional comments

Rory said he is fine, so I am merging now.

PR #2053, opened 2026-01-07
New Grid API

ddudt:
Resolves #1840 

Restructures and generalizes the Grid classes. The new ABC `AbstractGrid` that all other grids inherit from does not assume a particular coordinate system. The new ABC `AbstractGridFlux` is specific to ($\rho,\theta,\zeta$) coordinates and is the parent class of all existing grid classes in flux coordinates. Most of the changes involved moving existing code into one of these new ABCs. 

The goal is to allow for new grid classes in the future that represent other coordinate systems, without changing the existing public API. 

To-Do:
- [x] Ensure tests pass
- [x] Generalize `Basis` dimensions (no references to flux coords)
- [x] Include `isinstance(grid, AbstractGridCurve)` checks wherever appropriate
- [x] Add 1D grids for `Curve` objects
- [x] Add 2D grids for `FourierRZToroidalSurface` objects
- [x] Alias flux grid class names like `Grid` -> `CustomGridFlux`, etc. (do not deprecate)
- [x] Update `docs/dev_guide/notebooks/grid.ipynb`
- [x] Utility function for `LinearGrid._create_nodes`?

PR #2157, opened 2026-04-10
Add fft grid and raz grid to test against master

### LINKED ISSUE #2162: nufft Error
When trying to plot effective ripple with `nufft_eps=0`, I still receive errors:

```
File ~/DESC/Nadav/eff_ripple.py:21, in plot_eff_ripple(eq)
     19 fig, ax = plt.subplots()
     20 theta = Bounce2D.compute_theta(eq, rho=np.unique(grid.nodes[: ,0]))
---> 21 res = eq.compute("effective ripple", grid=grid, theta=theta, nufft_eps=0)
     22 ax.plot(res["rho"], res["effective ripple"], '.')
     23 ax.set_yscale("log")

File ~/DESC/desc/equilibrium/equilibrium.py:1229, in Equilibrium.compute(self, names, grid, params, transforms, profiles, data, override_grid, **kwargs)
   1220     data1dz = {
   1221         key: grid.copy_data_from_other(
   1222             data1dz[key], grid1dz, surface_label="zeta"
   (...)   1225         if key in dep1dz and key not in data
   1226     }
   1227     data.update(data1dz)
-> 1229 data = compute_fun(
   1230     self,
   1231     names,
   1232     params=params,
   1233     transforms=transforms,
   1234     profiles=profiles,
   1235     data=data,
   1236     **kwargs,
   1237 )
   1238 return data

File ~/DESC/desc/compute/utils.py:137, in compute(parameterization, names, params, transforms, profiles, data, **kwargs)
    134 if data is None:
    135     data = {}
--> 137 data = _compute(
    138     p,
    139     names,
    140     params=params,
    141     transforms=transforms,
    142     profiles=profiles,
    143     data=data,
    144     **kwargs,
    145 )
    147 # convert data from default 'rpz' basis to 'xyz' basis, if requested by the user
    148 if basis == "xyz":

File ~/DESC/desc/compute/utils.py:195, in _compute(parameterization, names, params, transforms, profiles, data, **kwargs)
    190     continue
    191 if not has_data_dependencies(
    192     parameterization, name, data, transforms["grid"].axis.size
    193 ):
    194     # then compute the missing dependencies
--> 195     data = _compute(
    196         parameterization,
    197         data_index[parameterization][name]["dependencies"]["data"],
    198         params=params,
    199         transforms=transforms,
    200         profiles=profiles,
    201         data=data,
    202         **kwargs,
    203     )
    204     if transforms["grid"].axis.size:
    205         data = _compute(
    206             parameterization,
    207             data_index[parameterization][name]["dependencies"][
   (...)    214             **kwargs,
    215         )

File ~/DESC/desc/compute/utils.py:217, in _compute(parameterization, names, params, transforms, profiles, data, **kwargs)
    205             data = _compute(
    206                 parameterization,
    207                 data_index[parameterization][name]["dependencies"][
   (...)    214                 **kwargs,
    215             )
    216     # now compute the quantity
--> 217     data = data_index[parameterization][name]["fun"](
    218         params=params, transforms=transforms, profiles=profiles, data=data, **kwargs
    219     )
    220 return data

    [... skipping hidden 47 frame]

File ~/DESC/desc-env/lib/python3.14/site-packages/jax/_src/interpreters/mlir.py:2391, in lower_per_platform(ctx, description, platform_rules, default_rule, effects, *rule_args, **rule_kwargs)
   2389   rule = platform_rules.get(platforms[0], default_rule)
   2390   if rule is None:
-> 2391     raise NotImplementedError(
   2392       f"MLIR translation rule for primitive '{description}' not "
   2393       f"found for platform {platforms[0]}")
   2395 # Multi-platform lowering
   2396 kept_rules: list[LoweringRule] = []  # Only the rules for the platforms of interest

NotImplementedError: MLIR translation rule for primitive 'nufft2' not found for platform cuda
```

Seems like the code still tries to use nufft even with a zero eps.

unalmis:
PR Age

- Adds FFT grids and field aligned grids to `test_compute_everything`.  Should be merged before stuff that changes grids #2053 or stuff that changes how the dependencies are ordered/selected for computation #1557 to ensure they do not cause bugs in the computations.
- Deletes a bunch of files that are now not necessary and removes bounce objective docstring boilerplate
- uses `nondiff_argnums` instead of `nondiff_argnames` so that the code works on old jax versions
- Adds comments to explain per bounce_comments PR
- Resolves #2162 .


tests/test_compute_funs.py
```diff
@@ -2049,6 +2049,45 @@ def test_parallel_grad_fd(DummyStellarator):
     )
 
 
+@pytest.mark.unit
+@pytest.mark.slow
+def test_fieldline_average():
```
unalmis, line 2054: moved here from test_neoclassical. not new

tests/test_compute_everything.py
```diff
@@ -297,3 +303,60 @@ def need_special(name):
...
+    # ci and my laptop differ a bunch at rho = 0, so skip that
+    rho = np.linspace(1e-2, 1, 10)
+    grid = LinearGrid(rho=rho, M=eq.M_grid, N=eq.N_grid, NFP=eq.NFP, sym=False)
+
+    kwargs = dict(
+        angle=Bounce2D.angle(eq, X=32, Y=32, rho=rho, tol=1e-10),
+        Y_B=grid.num_zeta * grid.NFP,
+        num_transit=5,
+        num_well=20 * 5,
+        nufft_eps=1e-10,
```
f0uriest, line 326: I think we should also keep testing the nufft_eps=0 path

f0uriest: this could be done below along with the vectorization test

unalmis:
someone else will need to cllick merge when ready

desc/integrals/bounce_integral.py
```diff
@@ -383,14 +391,20 @@ def _objective_build(obj, names, eta):
...
+            Y_B = Y_B_rule(obj._grid, obj._hyperparam["spline"])
             obj._hyperparam["Y_B"] = Y_B
         if obj._hyperparam["num_well"] is None:
             obj._hyperparam["num_well"] = num_well_rule(
-                obj._hyperparam["num_transit"], eq.NFP, Y_B
+                obj._hyperparam["num_transit"],
+                eq.NFP,
+                # Due to legacy reasons Y_B is resolution over full transit
+                # if spline is true.
+                Y_B if obj._hyperparam["spline"] else Y_B * eq.NFP,
```
unalmis:
```suggestion
                Y_B if obj._hyperparam["spline"] else (Y_B * eq.NFP),
```

desc/integrals/bounce_integral.py
```diff
@@ -483,7 +497,12 @@ def _defaults(eta, grid, **kwargs):
         num_transit = kwargs.get("num_transit", 20)
 
         Y_B = kwargs.get("Y_B", Y_B_rule(grid, spline))
-        num_well = kwargs.get("num_well", num_well_rule(num_transit, grid.NFP, Y_B))
+        num_well = kwargs.get(
+            "num_well",
+            # Due to legacy reasons Y_B is resolution over full transit
+            # if spline is true.
+            num_well_rule(num_transit, grid.NFP, Y_B if spline else Y_B * grid.NFP),
```
unalmis:
```suggestion
            num_well_rule(num_transit, grid.NFP, Y_B if spline else (Y_B * grid.NFP)),
```

desc/integrals/_bounce_utils.py
```diff
@@ -1066,21 +1067,27 @@ def Y_B_rule(grid, spline):
...
-    )
-    return {"dct spline": chebvander(x, Y_trunc - 1)}
+    if spline:
+        Y_trunc = truncate_rule(Y)
+        Y_B, num_z = round_up_rule(Y_B, NFP, grid.num_zeta == 1)
+        x = jnp.linspace(
+            -1, 1, (Y_B // NFP) if (grid.num_zeta == 1) else num_z, endpoint=False
+        )
+        return {"dct spline": chebvander(x, Y_trunc - 1)}
+    return {}
```
f0uriest: rather than making this a do-nothing function if spline=False wouldn't it be better to just avoid calling it in the first place?

desc/integrals/_bounce_utils.py
```diff
@@ -762,16 +759,20 @@ def theta_on_fieldlines(angle, iota, alpha, num_transit, NFP):
...
         .compute_cheb(alpha)
         .swapaxes(0, -3)
     )
     alpha = alpha.swapaxes(0, -2)
-    delta = delta.at[..., 0].add(alpha)
+    delta = delta.at[..., 0].add(alpha)  # This is now θ = α + δ.
     assert delta.shape == (*angle.shape[:-2], num_alpha, num_transit * NFP, Y)
+
+    if X < 24:
+        # Short circuit convergence to enforce continuity. This is needed as
```
f0uriest: what do you mean by "short circuit convergence"?

desc/integrals/_bounce_utils.py
```diff
@@ -189,7 +189,7 @@ def _newton(o, pitch_inv, z1, z2, mask, nufft_eps=1e-10):
     return z[..., 0, :, :], z[..., 1, :, :]
 
 
-@partial(jax.custom_jvp, nondiff_argnames=("num_well", "nufft_eps"))
+@partial(jax.custom_jvp, nondiff_argnums=(2, 3))
```
f0uriest: is there a performance difference between argnames/argnums? if not I prefer argnames as I find it a bit clearer.

desc/objectives/_fast_ion.py
```diff
@@ -186,6 +111,7 @@ def __init__(
         pitch_batch_size=None,
         surf_batch_size=1,
         nufft_eps=1e-7,
+        spline=True,
```
f0uriest, line 114: I thought "spline" and "bounce1d" were the same option? What are the differences / new default here?

desc/objectives/objective_funs.py
```diff
@@ -131,6 +131,97 @@
     "jac_chunk_size": doc_jac_chunk_size,
 }
 
+doc_bounce = """
+    Notes
+    -----
+    Performance will improve significantly by resolving these GitHub issues.
+      * ``1206`` Upsample data above midplane to full grid assuming stellarator symmetry
```
f0uriest, line 141: I've never been a fan of putting this stuff in the public docstring. It's an implementation detail that user's can't do anything about / shouldn't need to know about.

desc/integrals/_bounce_utils.py
unalmis: If you want to keep argnames, you need to bump min jax version to 0.7 ish or 0.8 I forget

desc/objectives/_fast_ion.py
unalmis: We need to solve nonlinear equation for non-unique roots of 2D fourier series to find bounce points.

Spline=True finds the initial guess for the bounce point with splines, then does Newton step on the Fourier series to find the true bounce point.

Spline=False just solves the eigenvalue problem on 1D Chebyshev series to get the true bounce point. Due to algorithm improvements and custom jvp, this is now quite performant.


bounce1d is memory intensive junk that I don't recommend optimization.

desc/objectives/objective_funs.py
unalmis: It is to motivate users who want more performance to make a new PR to solve them as opposed to writing a new code. (e.g some people rewrote neo in Jax; I'm guessing, I don't know if neo jax is fast.)

(Note neo algorithm has algebraic order of accuracy 1.5 in the runge kutta step size. < 0.5 for gamma c. With large constants since they don't find bounce points.).

desc/integrals/_bounce_utils.py
unalmis: https://github.com/unalmis/interpax_fft/blob/a9ac58cc58a61a37353b7f191c1bfb3c432814bd/interpax_fft/_series.py#L596

You need this to compute bounce points correctly; running test_bounce2d_checks without it for 10 transits for example will trip the error checking in the check_bounce_points method.

desc/integrals/_bounce_utils.py
unalmis: Sure, just make sure to change objective.compute to constants.get(_vander,{})

desc/integrals/bounce_integral.py
```diff
@@ -1029,12 +1050,30 @@ def _nufft(self, z, data, eps, low_ram, mask, sentinel):
...
+        z = flatten_mat(z, 3)
+        t = self._theta.eval1d(z, loop=low_ram).reshape(*shape, 1)
+        if isinstance(self._c["B(z)"], PiecewiseChebyshevSeries):
+            # Using the same |B| that gave bounce points increases correlation
+            # in discretization error in an open neighboorhood around the
+            # current point in the optimization landscape, and hence removes
+            # noise from optimization derivatives. For example, the difference
+            # between the auto derivative and a 4 point finite difference
+            # stencil at the optimal step size is reduced from 10³ to 10⁻⁶ for
+            # Γ_c (which has the singular weight 1/v_∥).
```
unalmis, line 1067: if you care, you can see this with the script on branch `fast_ion` https://github.com/PlasmaControl/DESC/blob/de0623ed18cd4bf9007fd25401aebee13c3f029b/publications/unalmis2025/effective_ripple_profile.py#L81

but it's not relevant to this pr. the comments there refer to nuffts. i am happy with the gradients

desc/objectives/_fast_ion.py
unalmis: default did not change

unalmis: Since bounce1d is now always less performant, I removed _use_bounce1d from the docs, although the option is still supported to not break user code.

desc/integrals/bounce_integral.py
```diff
@@ -1000,41 +1010,72 @@ def integrate(
 
         return result[0] if len(result) == 1 else result
 
-    def _nufft(self, z, data, eps, low_ram, mask, sentinel):
-        shape = z.shape
-        z = flatten_mat(z, 3)
+    def _nufft(self, x, z1, z2, data, low_ram, eps, pitch_inv):
```
unalmis, line 1017: adds comments per the bounce cosmetic comments pr

YigitElma:
@unalmis, if you make a new PR for the fix of #2162, we can merge it faster. People will be travelling/attending to ISHW, so this bigger PR can take a bit longer.

unalmis:
i'll wait, feel free to click merge after its approved if i don't respond

unalmis:
its mostly docstring deletion and addition btw

desc/external/neo.py
```diff
@@ -26,8 +26,8 @@ def __init__(self, name, eq, ns=256, M_booz=None, N_booz=None):
 
         self.eq = eq
         self.ns = ns  # number of surfaces
-        self.M_booz = setdefault(M_booz, 5 * eq.M + 1)
-        self.N_booz = setdefault(N_booz, 5 * eq.N)
+        self.M_booz = setdefault(M_booz, 6 * eq.M + 1)
+        self.N_booz = setdefault(N_booz, 6 * eq.N)
```
ddudt, line 30: Do we need the default resolution to be so high? In `QuasisymmetryBoozer` the default is only `2 * eq.N`.

desc/external/neo.py
```diff
@@ -39,12 +39,36 @@ def read(name):
...
+        """Write neo input file.
+
+        Parameters
+        ----------
+        theta_n, phi_n : int
+            Spline resolution.
+        num_pitch : int
+            Resolution for quadrature over velocity coordinate.
+            Default is 150.
+        nstep_per : int
```
ddudt, line 52: `n_step_per`

desc/external/neo.py
```diff
@@ -39,12 +39,36 @@ def read(name):
...
+        """
         print(f"Writing VMEC wout to {self.vmec_file}")
         VMECIO.save(self.eq, self.vmec_file, surfs=self.ns, verbose=0)
         self._write_booz()
-        self._write_neo(**kwargs)
+        self._write_neo(
+            theta_n=theta_n,
+            phi_n=phi_n,
+            num_pitch=num_pitch,
+            nstep_per=n_step_per,
```
ddudt: make these arg names the same

desc/external/neo.py
ddudt, line 1: I don't see how the changes in this file are related to the rest of this PR

desc/integrals/_bounce_utils.py
```diff
@@ -762,16 +778,20 @@ def theta_on_fieldlines(angle, iota, alpha, num_transit, NFP):
...
         FourierChebyshevSeries(angle, domain, truncate=Y)
         .compute_cheb(alpha)
         .swapaxes(0, -3)
     )
     alpha = alpha.swapaxes(0, -2)
-    delta = delta.at[..., 0].add(alpha)
+    delta = delta.at[..., 0].add(alpha)  # This is now θ = α + δ.
     assert delta.shape == (*angle.shape[:-2], num_alpha, num_transit * NFP, Y)
+
+    if X < 24:
```
ddudt: Why the magic number 24?

desc/integrals/_bounce_utils.py
```diff
@@ -1066,21 +1086,27 @@ def Y_B_rule(grid, spline):
     return (Y_B * grid.NFP) if spline else Y_B
 
 
-def num_well_rule(num_transit, NFP, Y_B=None):
+def num_well_rule(num_transit, NFP, extrema_per_transit=None):
```
ddudt: Does Y_B represent "extrema per transit"? If so, should we change this name everywhere else in the code?

desc/integrals/_bounce_utils.py
ddudt: I think Rory means don't call the function `get_vander()` if `spline=False`

ddudt: As an aside, I recommend letting the reviewer who originally asks a question mark it as resolved if no changes were made to the code to address it.

desc/integrals/bounce_integral.py
```diff
@@ -383,14 +395,20 @@ def _objective_build(obj, names, eta):
...
+                eq.NFP,
+                # Due to legacy reasons Y_B is resolution over full transit
+                # if spline is true.
+                Y_B if obj._hyperparam["spline"] else (Y_B * eq.NFP),
             )
 
-        obj._constants["_vander"] = get_vander(obj._grid, Y, Y_B, eq.NFP)
+        obj._constants["_vander"] = get_vander(
+            obj._grid, Y, Y_B, eq.NFP, obj._hyperparam["spline"]
+        )
```
ddudt: See previous comment. I think we can change this to:
```
obj._constants["_vander"] = get_vander(obj._grid, Y, Y_B, eq.NFP) if obj._hyperparam["spline"] else {}
```

desc/integrals/_interp_utils.py
```diff
@@ -35,10 +33,17 @@
             f"will not be available: {e}",
             UserWarning,
         )
-    can_use_nufft = False
 
 from desc.backend import jax, jnp
 
+_JF_BUG = True
```
ddudt, line 39: I don't love this hack since it doesn't seem like `jax-finufft` is going to merge your PR anytime soon. Is there any other option?

tests/test_fast_ion.py
ddudt, line 1: Why are we deleting tests? Are these same tests now covered by `test_compute_everything`?

tests/test_neoclassical.py
ddudt, line 1: Why are we deleting tests? Are these same tests now covered by `test_compute_everything`?

desc/objectives/objective_funs.py
ddudt: Unresolving until we decide if we want to do this or not

desc/integrals/bounce_integral.py
```diff
@@ -1068,24 +1109,19 @@ def interp_to_argmin(self, f, points, *, nufft_eps=1e-6, is_fourier=False):
...
         K_z = self._num_z // 2 * self._NFP
-        mins, B_mins = get_mins(
-            self._c["knots"],
-            self._c["B(z)"],
-            num_mins=num_transit * max(K_z, 5),  # liberal heuristic
-            # We set fill value to 0 since we chose our coordinates
-            # such that all bounce points are at ζ >= 0; and therefore,
-            # junk values in B_mins cannot be selected in argmin.
-            fill_value=0.0,
+        num_mins = num_transit * max(K_z, 5)  # liberal heuristic
```
ddudt: I trust this is a fine default, but is this something we should let the user specify if they desire? Trying to avoid hard-coded numbers.

desc/integrals/_bounce_utils.py
unalmis: I ran a scan of tests and 24 was the number required to pass all of them. The hardest ones being high NFP, high poloidal spectrum, and with bounce points hitting near the toroidal cut that is the end of field period, e.g. heliotron.

A small number like 24 is typically the number after which aliasing error is damped exponentially.

desc/integrals/_bounce_utils.py
unalmis: `Y_B` is an upper bound on the number of minima of |B| `per toroidal transit` (and hence an upper bound on the number of wells for any single pitch angle) when `spline=True`. When ``spline=False``, ``Y_B//2`` is an upper bound on the number of minima of |B| `per field period`.

This `per toroidal transit` vs `per field period` is an annoying discrepancy that would ideally be  `per field period` for both, but I did not change due to backwards compatibility. i have yet to author a pr with breaking changes to the bounce integrals in desc, so if we want to make that change it should be its own pr .

desc/integrals/_bounce_utils.py
unalmis: objective.constants["_vander"] needs a dict in it to avoid recompilation - recall there is a jit barrier on the registered compute function. how you choose to add that dict to it is a cosmetic choice. you'll need to do it here or somewhere else. i do not see any reason why rory's suggestion would be better, perhaps he could explain. in any case it's cosmetic

desc/external/neo.py
unalmis: you need at leaset 5* for it to converge to what desc computes for bounce integrals. i bumped the resolution and added documentation to neo's order of convergence so people can stop wondering and asking why desc vs neo differ

desc/integrals/_interp_utils.py
unalmis: The question is open ended; i don't understand it. option to do what?

desc/integrals/bounce_integral.py
unalmis: ok sure

desc/integrals/bounce_integral.py
unalmis: I think there's too many user-specifiable parameters. We should try to reduce them not increase. This is a liberal choice, it won't ever give issues.

tests/test_fast_ion.py
unalmis: Yes. the deleted ones detect visual correctness past 10% visual change only. the new ones test numerical correctness to $10^{-8}$ relative tolerance.

tests/test_neoclassical.py
unalmis: https://github.com/PlasmaControl/DESC/pull/2157/changes#r3089893505

desc/objectives/objective_funs.py
unalmis: i don't get github notifications, i just check if there are unresolved comments to see if i need to reply.

unalmis: it was already there before so if you care you can just make a commit to change

desc/external/neo.py
unalmis: `nstep_per` to match neo

desc/objectives/objective_funs.py
unalmis: resolved via developer notes.

unalmis:
can't do further requests, make new pr and merge that after if you want

desc/objectives/objective_funs.py
```diff
@@ -131,6 +131,94 @@
...
+        Default is ``num_pitch``.
+    surf_batch_size : int
+        Number of flux surfaces with which to compute simultaneously.
+        If given ``None``, then ``surf_batch_size`` is ``grid.num_rho``.
+        Default is ``1``. Only consider increasing if ``pitch_batch_size`` is ``None``.
+    nufft_eps : float
+        Precision requested for interpolation with non-uniform fast Fourier
+        transform (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
+    spline : bool
+        Whether to use cubic splines to compute bounce points instead of
```
unalmis:
```suggestion
        Whether to use cubic splines to compute initial guess for bounce points
```

desc/objectives/objective_funs.py
```diff
@@ -131,6 +131,94 @@
...
+        Chebyshev series. Default is ``True``.
```
unalmis:
```suggestion
        instead of Chebyshev series. Default is ``True``.
```

unalmis (commented):
clarify docstring

desc/integrals/_bounce_utils.py
ddudt: Magic numbers are a bad coding practice. We should either make it a user-specified parameter that can default to 24, or define it as a constant and clearly document this explanation.

desc/integrals/bounce_integral.py
ddudt: See previous comment about avoiding magic numbers. If we don't want to add another user parameter, we should at least better document this constant.

desc/integrals/_bounce_utils.py
ddudt: Thanks for the explanation, that's an annoying discrepancy. I would be in favor of changing this in a future PR to make `Y_B` always mean "per field period" for both cases. If I understand correctly, this breaking change would only make old code compute at higher resolution. 

@unalmis Can you make a new issue about this if there isn't one already? I tried searching for it but didn't see anything.

desc/integrals/_interp_utils.py
ddudt: It looks like this is a temporary solution until [your PR](https://github.com/flatironinstitute/jax-finufft/pull/216) gets merged into `jax-finufft`. I'm skeptical that PR will ever get merged by the `jax-finufft` team. If we can't fix this issue on their end, what would be the best solution on our end? Maybe the answer is still this `_JF_BUG` flag, but I'm trying to understand what our options are and brainstorm other solutions.

ddudt:
@unalmis Please do not mark comments as resolved unless they are addressed in the code or the discussion has concluded. Responding is not the same as resolving. Comments are helpful for everyone to read, but when they are resolved it makes them less visible for other reviewers to see and join the conversation. 

I can show you how to enable GitHub notifications to see when there are new comments.

desc/integrals/_interp_utils.py
unalmis: This is only option.

Anyone can build my branch from source if they want. Git clone my branch, pip install ../jax-finufft. Works fine

unalmis: The time to brainstorm was two years ago when I mentioned this fundamental limitation of jax at developer meetings. We will not solve JAX's biggest problem with patch in DESC code.

The algorithm I created here has excellent performance on CPU and GPU, especially considering the challenge of an unknown number "hitting events" like bounce points is the main limitation of the modern jax computing paradigm.
@ddudt

unalmis: If you want to create a new algorithm feel free to, but this isn't really  the medium to discuss or bring that up.

desc/integrals/_bounce_utils.py
unalmis: There is already a short essay in the docstring that states informally the reason and formally the mathematics for why this is required. I will add the choice for the constant to that description and a parameter, but will not expose this to users since there is no need.

desc/integrals/_bounce_utils.py
unalmis: At this point it's a bad look for me to keep making desc issues since I don't work for desc, but you can make one. 

Preferably, instead of making the issue, just make the pr.

unalmis: Can change num transit to num field period too

desc/integrals/_interp_utils.py
unalmis: You can write a pallas kernel which supports dynamic shapes if u want.

desc/integrals/bounce_integral.py
unalmis: It's not magic it's the max possible amount as determined by the max fourier frequency. In practice it's way less than that because the high frequency coefficients are exponentially damped

unalmis:
> @unalmis Please do not mark comments as resolved unless they are addressed in the code or the discussion has concluded. Responding is not the same as resolving. Comments are helpful for everyone to read, but when they are resolved it makes them less visible for other reviewers to see and join the conversation.
> 
> I can show you how to enable GitHub notifications to see when there are new comments.

You had reopened two which is outside scope of this PR as `master` already has them. Such discussion should be done in github issues. I think the effort it takes for a reviewer to click see comment button is much less than the increased effort for me, the maintainer, to track which issues are actually resolvable and which are just unactionable due to jax or cosmetic personal preference that should be resolved elsewhere as they don't need my input. Github notifications don't work selectively on a PR. i get notifs for all of desc or none of it.

desc/integrals/bounce_integral.py
unalmis: added kwargs

unalmis:
@ddudt consider https://github.com/pullpo-io/conventional-comments for comprimise

desc/objectives/objective_funs.py
dpanici: FWIW I don't mind doing this, it helps to keep track of things and the issues themselves are public anyways.

tests/test_compute_everything.py
```diff
@@ -297,3 +303,75 @@ def need_special(name):
...
+def fft_grid_data(p):
+    """Compute fft grid quantities."""
+    if p != "desc.equilibrium.equilibrium.Equilibrium":
+        return {}
+
+    # TODO: can automate this later to add omngeneity, boozer transform, etc.
+    fft_names = ["effective ripple", "Gamma_c", "Gamma_c Velasco"]
+
+    eq = get("W7-X")
+    # ci and my laptop differ a bunch at rho = 0, so skip that
```
dpanici, line 317: Non-Blocking, just curiosity: I'm fine with skipping the axis if it is sensitive, but out of curiosity what OS are you using? 

I find I get minor diffs that still trip the test if I say re-saved quantities I computed on my mac, which then will have slight differences compared to the quantities computed on the CI's linux (I believe) runners. Usually I would just make sure I saved the quantities on some linux (non-wsl though I think I may have had issues there too) os like one of the university clusters.

dpanici (approved):
Looks good to me, thanks for taking some of the comments from that PR I had made. I would also like @ddudt to re-review since he had made a lot of comments his first pass around.

If possible, let's merge this one before the sparse pullback PR (#2170 , have not reviewed it yet) gets merged into here, so we can isolate the changes a bit and make the reviews of each more manageable

dpanici (approved):

desc/integrals/_bounce_utils.py
```diff
@@ -99,21 +102,23 @@ def bounce_points(pitch_inv, knots, B, num_well=-1):
     mask = flatten_mat(intersect >= 0)
     z1 = (dB_dz <= 0) & mask
     z2 = (dB_dz >= 0) & epigraph_and(mask, dB_dz)
+    del dB_dz
```
f0uriest, line 105: just curious, does this make a performance difference? I thought jax basically ignores `del` under jit?

desc/integrals/_interp_utils.py
```diff
@@ -8,13 +8,11 @@
 try:
     from jax_finufft import nufft2, options
 
-    can_use_nufft = True
```
f0uriest, line 11: don't we still need this to set the default value for eps correctly if finufft doesn't work?

desc/integrals/bounce_integral.py
```diff
@@ -384,14 +396,22 @@ def _objective_build(obj, names, eta):
...
         if Y_B is None:
-            Y_B = Y_B_rule(obj._grid, spline=True)
+            Y_B = Y_B_rule(obj._grid, obj._hyperparam["spline"])
             obj._hyperparam["Y_B"] = Y_B
         if obj._hyperparam["num_well"] is None:
             obj._hyperparam["num_well"] = num_well_rule(
-                obj._hyperparam["num_transit"], eq.NFP, Y_B
+                obj._hyperparam["num_transit"],
+                eq.NFP,
+                # Due to legacy reasons Y_B is resolution over full transit
```
f0uriest, line 405: wasn't there another PR that addressed this? I thought that got merged in here, so is this leftover?

desc/integrals/bounce_integral.py
```diff
@@ -795,22 +822,20 @@ def points(self, pitch_inv, num_well=None):
...
-        # Newton update has only been implemented for nuffts; contributions welcome.
-        z1, z2, _ = bounce_points(
-            pitch_inv, self._c["knots"], self._c["B(z)"], num_well
-        )
-        return z1, z2
+        if self._nufft_eps < 1e-14:
+            # FIXME: Newton update has only been implemented for nuffts; contributions
+            # welcome : copy logic in desc/equilibrium/coords._map_poloidal_coordinates.
+            return bounce_points(pitch_inv, self._c["knots"], self._c["B(z)"], num_well)
+        return regular_points(self, pitch_inv, num_well)
```
f0uriest, line 838: Just so I understand, there are basically 2x2 ways of computing bounce points. Either using a spline of B, or a chebyshev representation, and then also either using newton refinement or not. Except if we use chebyshev we always skip the newton update since we assume they're already accurate enough, so basically 3 ways. And if nuffts don't work we also skip the newton refinement bc its not implemented yet. And the default seems to be spline + newton? Is there a reason to prefer that over chebyshev? is it that much more accurate/faster?

desc/integrals/bounce_integral.py
f0uriest, line 896: isn't this now set in `Bounce2D.__init__`?

desc/integrals/_bounce_utils.py
unalmis: - I don't delete for performance, I delete to dereference it so that it doesn't get called again. When there's a lot of variables in namespace it's harder to keep track of which variables are still useble, so dereferencing is done for maintainability/safety reasons.
- i don't expect it or see it reduce memory under jit, although there's an open equinox issue that suggests in general it can. But given that these functions are not jitted and consume lots memory, I suppose there is utility to dereference as well.

desc/integrals/_interp_utils.py
unalmis: The `can_use_nufft` flag didn't work, so I removed it. To test whether nuffts work, the only way is to actually perform jax.grad on a nufft call on jax array, which is against what desc devs want, see #2156

desc/integrals/bounce_integral.py
unalmis: No it was done after #2170 and hence got merged into #2170 not this one. 

Sorry, but I don't plan ahead with optimal pr order, sparse happended out of blue, then dan suggested things to change, so I had to change on top of sparse pr, not this one

desc/integrals/bounce_integral.py
unalmis: After #2170 I have resolved all remaining  inconsistencies.

There are 2x2 methods yes: (spline true or false) Cartesian product with (nufft true or false).

Regarding bounce points,
1. `spline=True`. This finds initial guess for bounce points with splines, then does a Halley step to polish the guess. This single step reduces error from 1e-4 to 1e-10.
2. `spline=False`. This finds bounce points with Chebyshev series. No Halley step is done because the points are found to machine precision already. The reason we skip doing Halley is not because the points are accurate enough roots of fourier series- that's a very demanding since there is a stiffness condition where the singularity turns epsilon error into sqrt(epsilon) error. And the derivatives become non-integrable if the error is large enough. The reason we skip Halley is because we use the Chebyshev B to compute v_|| so the  bounce points error is always exactly zero. 

Whether to use nuffts is independent of this. 

On my CPU, fastest order is

1. nuffts + spline (default) 5 seconds for unconstrained Jacobian. (No dynamic shapes; with dynamic shapes about 2 sec)
2. Then turning off nuffts , exclusive OR, turning off spline. 12 seconds for Jacobian.
3. Turning off nuffts and turning off spline. 19 seconds for Jacobian.

I don't like having 2x2 options, but having a flag for `nuffts` is clearly necessary for the code to be usable. From a mathematical standpoint, having a flag for `spline` is also needed for me to claim proof of convergence and verufy integrability of optimization derivatives because `spline=False` is the only way to avoid a stiffness condition.

desc/integrals/bounce_integral.py
unalmis: Do we want to default to that?

desc/integrals/bounce_integral.py
unalmis: I choose the default as spline + nufft because I like the defaults to have best scaling laws because users often do resolution scans to get a feel for stuff. 

If I set chebyshev to default , then the eigenvalue solve for bounce points is Y_B^3 so it's got poor scaling that would make resolution scan annoying. Probably not an issue for gpu but anyway.

unalmis: https://github.com/PlasmaControl/DESC/blob/ac79068f7d174642d4af71383861b31e2567ecf1/desc/integrals/bounce_integral.py#L656-L657

f0uriest: OK, thanks. I think it would be helpful to include that note about performance somewhere in the docstrings, since it definitely wasn't obvious to me which one is better. I had thought using chebyshev everywhere was the fastest. We should also double check that those rankings hold up on GPU, since that's where we care most about performance.

As for what the default should be, I think we should pick that based on what's fastest on GPU at "reasonable" resolution (ie, the default resolution for the objective), with a note explaining when the different options might be preferable (based on hardware, resolution etc). 

Doesn't have to be in this PR but should at least be in #2170 since that cleans this up as well.

desc/integrals/bounce_integral.py
f0uriest: yeah I'd prefer if setting that once at init propagates it down to everywhere it's needed, but we can keep the option to override it locally if more fine control is needed.

unalmis: ok ill do on 2170 since this pr is laready approved

f0uriest (approved):

unalmis:
i don't have codecov override

desc/integrals/_bounce_utils.py
```diff
@@ -706,6 +727,15 @@ def theta_on_fieldlines(angle, iota, alpha, num_transit, NFP):
...
         Number of field periods.
+    X_min : int
+        See notes section. This parameter should never be changed.
+        It is included in the function signature for code optics only.
+        It is the number below which we short-circuit convergence to enforce
+        continuity by removing a discontinuity which is near machine precision
+        due to exponential convergence. This is not a hack; it has rigorous
+        mathematical justification regardless of the size of the removed
+        discontinuity, and does not bias the output beyond that of more
+        floating point operations in finite-precision.
```
ddudt, line 738: [nitpick](https://pullpo.io/cc?l=nitpick) 
If you don't like this in the function signature I said it could be a comment in the code. All I wanted was documentation to explain where the number 24 came from.

ddudt (approved):
[thought(non-blocking)](https://pullpo.io/cc?l=thought&d=non-blocking) 

I want to echo [this comment from Rory](https://github.com/PlasmaControl/DESC/pull/2170#pullrequestreview-4191730837) about having smaller PRs, each with a narrow scope. This was a big PR that covered a lot more than it originally advertised. More digestible PRs are easier and faster to review, and will make you a better software developer. 

Also I really like [this Chrome extension](https://github.com/pullpo-io/conventional-comments) that Kaya recommended and I would recommend it to others!

tests/test_compute_everything.py
unalmis: #2199

unalmis:
> i don't have codecov override

need codecov override

PR #2164, opened 2026-04-15
Prevent compilation in `obj.print_value`

YigitElma:
At the optimization, we call `objective.print_value` (even for verbose=0), which calls each sub-objective's `compute_unscaled` method. This triggers a compilation because during optimization, the only function that is compiled is either `ObjectiveFunction.compute_scalar` or `ObjectiveFunction.compute_scaled_error`. Even if we called `_Objective.compute_scaled_error`, this would trigger compilation since JAX doesn't store the compiled code of a wrapped function.

In this PR, `print_value` takes in the objective values as an array along with state vectors. Objective values either come from:
- the optimizer (if a vector-valued optimizer is used, like `lsq-exact`) 
- are computed via `ObjectiveFunction.compute_scaled_error`. 

Then, these vectors are sliced and passed to sub-objective `print_value` methods. Unscaled objective is obtained by reverse shift and scale operations. Since `shift` removes some information if there are `bounds`, the function falls back to calling `compute_unscaled` for that (this is an unavoidable compilation cost).

Same splitting is also used in `factorize_linear_constraints`, where we used to call sub-objective methods.

Caveats:
- If the user has an expensive and bounded objective along with a scalar optimizer that doesn't store the initial and final values (fse and fse0), this change actually computes the function twice (one with `compute_scaled_error` and one with `compute_unscaled`). Since the new `options` key can remove this cost, I assume this is acceptable. 

## Question
`test_mercier_print` and `test_magwell_print` were unnecessary since the parent `_Objective` covers all different cases; should I remove them? They don't add extra coverage. `test_plasma_vessel_distance_print`, `test_profile_objective_print`, and `test_boundary_error_print` cover all possible cases if I am not mistaken. Not important, but noting in the spirit of reducing redundant tests.


ddudt:
My suggestions:

- Add option to not compute initial & final values if the user doesn't need them (not printing them and doesn't care about result history)
- See if not JITing `compute_unscaled` is faster, since this function only gets called twice for the initial & final values

desc/objectives/objective_funs.py
```diff
@@ -635,26 +646,34 @@ def print_value(self, x, x0=None, constants=None):
...
-                else:
-                    out[obj._print_value_fmt] = [outi]
+            params0 = [None] * len(params)
+        assert len(params) == len(constants) == len(self.objectives)
+        assert len(params0) == len(constants) == len(self.objectives)
+
+        offset = 0
+        for par, par0, obj, const in zip(params, params0, self.objectives, constants):
+            dim = obj.dim_f
+            fi = f_full[offset : offset + dim]
```
YigitElma, line 756: I can do the same trick [here too](https://github.com/PlasmaControl/DESC/blob/7238beae389ea906fac5ff90bc3a607af051dcc7/desc/objectives/utils.py#L159).

The purpose there is slightly different, but I think we can still use `compute_scaled_error` (which is compiled couple lines above) by replacing `y2=0`. I am not sure but the current code seems like it always assumes constraints are not bounded.

YigitElma: Yes, whatever constraint was passed to `factorize_linear_constraints` is not bounded due to [this line](https://github.com/PlasmaControl/DESC/blob/7238beae389ea906fac5ff90bc3a607af051dcc7/desc/optimize/optimizer.py#L559)

desc/objectives/_free_boundary.py
```diff
@@ -255,13 +255,18 @@ def compute(self, eq_params, *field_params, constants=None):
         Bsq_err = (bsq_in - bsq_out) * g
         return jnp.concatenate([Bn_err, Bsq_err])
 
-    def print_value(self, args, args0=None, **kwargs):
+    def print_value(self, fse, f0se=None, args=None, args0=None, **kwargs):
```
f0uriest: `print_value` is part of the public API so I'd prefer to keep things backwards compatible. This could be done by making `fse` and `fse0` part of kwargs, or putting the new arguments at the end. And we should ensure that it still works correctly if the user only passes in `args` and/or `args0` without `fse` etc.

desc/objectives/objective_funs.py
```diff
@@ -1479,19 +1524,66 @@ def jvp_unscaled(self, v, x, constants=None):
         """
         return self._jvp(v, x, constants, "unscaled")
 
-    def print_value(self, args, args0=None, **kwargs):  # noqa: C901
-        """Print the value of the objective and return a dict of values."""
-        # compute_unscaled is jitted so better to use than than bare compute
+    def print_value(  # noqa: C901
+        self, fse, f0se=None, args=None, args0=None, **kwargs
```
f0uriest: see previous comment about keeping this backwards compatible

desc/objectives/objective_funs.py
```diff
@@ -593,7 +593,7 @@ def compute_scalar(self, x, constants=None):
         f = jnp.sum(self.compute_scaled_error(x, constants=constants) ** 2) / 2
         return f
 
-    def print_value(self, x, x0=None, constants=None):
+    def print_value(self, x, x0=None, fse=None, fse0=None, constants=None):
```
f0uriest: see previous comment about keeping this signature backwards compatible (new args should go at the end, possibly as keyword only)

desc/optimize/least_squares.py
```diff
@@ -467,6 +468,8 @@ def lsqtr(  # noqa: C901
         allx=allx,
         alltr=alltr,
     )
+    result["fse"] = f
```
f0uriest, line 471: `f` is already stored as `result.fun`
Could also consider storing the history of `f` like we do for `allx`?

desc/optimize/optimizer.py
```diff
@@ -174,6 +174,10 @@ def optimize(  # noqa: C901
             Supported](https://desc-docs.readthedocs.io/en/stable/optimizers.html)
             for more details on the options available to each specific optimizer.
 
+            Additionally, one can suppress the summary of objective values via
+            options["print_summary"]=False. This can prevent additional objective
```
f0uriest: isn't this already handled by the `verbose` flag?

desc/optimize/optimizer.py
```diff
@@ -364,14 +369,15 @@ def optimize(  # noqa: C901
...
-        # what to do about it.
-        with io.StringIO() as buf, contextlib.redirect_stdout(buf):
-            result = _print_output(things, things0, objective, constraints, result)
-            text_out = buf.getvalue()
-        if verbose > 0:
-            print(text_out)
+        # we always want to populate result dict with before/after values unless
+        # user specifically asks to skip, but may not always want to print, so
+        # we first capture the output and then decide what to do about it.
+        if print_summary:
```
f0uriest: this feels kind of confusing, if `print_summary` is true but `verbose = 0`?

desc/optimize/optimizer.py
```diff
@@ -485,12 +491,20 @@ def _print_output(things, things0, objective, constraints, result):
     print("{:=<{}}".format("", PRINT_WIDTH + w_divider))
 
     print(f"{'Start  -->   End':>{PRINT_WIDTH+21}}")
-    values = objective.print_value(objective.x(*state), objective.x(*state_0))
+    fse = result.get("fse", None)
```
f0uriest, line 488: nit: why `fse` rather than just `f`?

tests/test_objective_funs.py
```diff
@@ -2552,16 +2552,18 @@ def test_profile_objective_print(capsys):
     pre_width = len("Maximum ")
 
     def test(obj, values, print_init=False, normalize=False):
+        par = obj.xs(eq)
+        f_scaled = obj.compute_scaled_error(*par)
         if print_init:
             # print the initial value too. For this test, it is the
             # same as the final value
-            obj.print_value(obj.xs(eq), obj.xs(eq))
+            obj.print_value(f_scaled, f_scaled, args=par, args0=par)
```
f0uriest: we should keep the old version of this test and add the new one and make sure they match (ie user can pass either x or f and get the same printed values)

f0uriest (changes requested):
I appreciate the intention of this, but I think it needs some care to ensure we don't break the API. Ideally everything should be backwards compatible, which shouldn't be too hard.

Also we shouldn't need to change any old tests, we should just add new ones that ensure passing f gives the same result as passing x

desc/objectives/objective_funs.py
YigitElma: I think this one is already backwards compatible. I will change the other one

desc/optimize/least_squares.py
YigitElma: The key `f` is always in the `results` dictionary, but it is `compute_scalar` or `compute_scaled_error`. Choosing this name made it easier to pass to `print_value` later with minimal change.

I would rather not store `allf` since it might use redundant memory and make things jittable harder #1669

desc/optimize/optimizer.py
YigitElma: I considered changing the definition of `verbose`, but I couldn't come up with a clear one.

Currently, even `verbose`=0 internally calls `print_value`, but optionally prints that to the stdout. When you have really expensive objectives, you have to wait a long time for this redundant call (if you don't want to use that output). As @ddudt mentioned [here](https://github.com/PlasmaControl/DESC/pull/2164#issuecomment-4262089627)

desc/optimize/optimizer.py
YigitElma: The condition on line 379 takes that into account.

I think the confusing part is `print_summary=False` and `verbose=3`. That case, you will get all the verbose stuff like iteration information, build information etc, but the table at the end won't be printed. This is a verbosity flag specific to the table.

desc/optimize/optimizer.py
YigitElma: As I mentioned in the previous comment, scalar optimizers store only the `compute_scalar` but for `print_value`, we need the full `compute_scaled_error `(hence f_scaled_error=fse). This naming makes the distinction clear (at least for me)

ddudt (approved):

f0uriest (approved):

PR #2167, opened 2026-04-16
Add parameter to shift the max B theta to Omnigenity objective

### LINKED ISSUE #1143: Misaligned target and eq fields in omnigenity objective
For the attached 
[data.zip](https://github.com/user-attachments/files/16354911/data.zip)
OH equilibrium, the |B| in Boozer coordinates using the equilibrium object, and a field object have max and min(|B|) at different points.

eq_modB
field_modB

Since the definition of Boozer coordinates is the same, this implies that the field object's |B| is unphysical and incorrect.

To reproduce the result, run the script below with the data inside the attached zip file.
```
import pdb
import os
import numpy as np
from desc.plotting import *

fname_path1 = "eq_final.h5"
fname_path2 = "field_final.h5"
    
eq0 = Equilibrium.load(f"{fname_path1}")
field = Equilibrium.load(f"{fname_path2}")

N = int(200)
grid = LinearGrid(L = N)
rho = np.linspace(0, 1, N+1)

data_keys = ["iota", "D_Mercier"]
data1 = eq0.compute(data_keys, grid = grid)

rho0 = 1.0
iota = data1["iota"]

fig, ax, Boozer_data0 = plot_boozer_surface(eq0, rho=rho0, return_data=True)
plt.show()
plt.close()


fig, ax, Boozer_data1 = plot_boozer_surface(field, rho=rho0, iota=np.interp(rho0, rho, iota), return_data=True)
plt.show()
plt.close()

```

dpanici:
Resolves #1143

### DIFF
--- a/desc/objectives/_omnigenity.py
+++ b/desc/objectives/_omnigenity.py
@@ -503,8 +503,13 @@ class Omnigenity(_Objective):
     Errors are relative to a target field that is perfectly omnigenous,
     and are computed on a collocation grid in (ρ,η,α) coordinates.
 
-    This objective assumes that the collocation point (θ=0,ζ=0) lies on the contour of
-    maximum field strength ||B||=B_max.
+    By default this objective assumes that the collocation point (θ=0,ζ=0) of the
+    equilibrium lies on the contour of maximum field strength ||B||=B_max. The
+    ``B_max_theta_location`` argument relaxes this assumption: the field's (θ=0) point
+    (which corresponds to the field's B_max) is compared against the equilibrium's |B|
+    evaluated at θ = ``B_max_theta_location``.
+
+    A single well in |B| per field period is assumed for this objective.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


desc/objectives/_omnigenity.py
```diff
@@ -541,6 +546,11 @@ class Omnigenity(_Objective):
         computation time during optimization and only ``eq`` is allowed to change.
         If False, the field is allowed to change during the optimization and its
         associated data are re-computed at every iteration (Default).
+    B_max_theta_location : float, optional
```
ddudt, line 549: I don't like this name. Technically it isn't even the theta angle of B_max, since that will change as a function of the toroidal angle. I suggest `B_max_theta_offset` or something similar. Also you would need to add a similar option for the zeta offset.

desc/objectives/_omnigenity.py
```diff
@@ -824,11 +836,13 @@ def compute(self, params_1=None, params_2=None, constants=None):
...
         # additional computations that cannot be part of the regular compute API
 
+        B_max_theta_location = self._B_max_theta_location
+
         def _compute_B_eta_alpha(theta_B, zeta_B, B_mn):
             nodes = jnp.vstack(
                 (
                     jnp.zeros_like(theta_B),
-                    theta_B,
+                    theta_B + B_max_theta_location,
```
ddudt, line 845: This will only work for toroidal and helical omnigenity. There is also the similar issue for poloidal omnigenity (aka QI) where you want to shift the toroidal angle zeta_B

ddudt (changes requested):
My preferred solution is to create equivalents of `flip_theta` and `rotate_zeta` for pyQSC & pyQIC. That would solve #1143 for @dpanici's use case and improve the functionality of those codes. 

If we do use this solution, we need to also add a new parameter for shifting zeta_B in case B_max is not aligned with zeta_B=0 for OP equilibria.

dpanici:
Thanks for catching, I'll add a toroidal shift as well. Adding these here is easier than modifying both pyQSC and pyQIC, imo.

Issue #1872, opened 2025-08-20
Compute function and gradient simultaneously for reverse mode


unalmis:
Seems this would be useful so that the function and derivative are computed in one pass, in particular for objectives that are scalar outputs.

1. https://docs.jax.dev/en/latest/automatic-differentiation.html#evaluating-a-function-and-its-gradient-using-jax-value-and-grad
2. https://github.com/jax-ml/jax/pull/762
3. https://github.com/jax-ml/jax/pull/762#issuecomment-1002267121

dpanici:
https://github.com/PlasmaControl/DESC/tree/dp/stoch-val-and-grad

but it gives nan right now, I had to make _x_scale static for some reason that I do not yet comprehend

unalmis:
> FWIW, the typical optimization does have order of magnitude same fxn calls as jacobian calls. Does not remove the fact that the jac is usually (much) more expensive than f for us, but could be a savings still.

After #2170 , computing Jacobian is only 2x more time than computing the function without its derivative on my CPU. A GPU would probably crunch through the flops and prefer the ability to perform the computation in a single pass additional computation, so it might be faster to do at once on gpu.

PR #2170, opened 2026-04-17
Sparse pullback for big performance gain

### LINKED ISSUE #1489: Forward differentiation nan gradient, nonsingular computation
Due to recent numerical changes in JAX or maybe optimization routines, there is a now an `nan` leak when optimizing with forward mode differentiation for some quantities. The issue does **not** occur with reverse mode differentiation.

After debugging, I have identified the the source of the nan gradient generation. It is peculiar.
 - Nan only occurs in forward mode.
 - All computations are well-behaved. In particular, let $f_i$ be the set of operations to compute that when composed define the computation of interest $F$. All $f_i$ output finite values with numerical buffer between the boundary of the domain where the $f_i$ are defined. That is, there is no computation like $f_i = \arccos(1 - \epsilon)$ with $\epsilon \geq 0$ small, which would be problematic if epsilon was small because the gradient $\nabla f_i$ would then become `nan` .
 - The source of the `nan` is from a non-singular operation.
 - It is on iteration 0 of the optimization.

The nan is generated  by the `v_tau` term [here](https://github.com/PlasmaControl/DESC/blob/2cb0f3e5819b03ad354b8b35f9f4652d75a065c1/desc/compute/_fast_ion.py#L331) and [here](https://github.com/PlasmaControl/DESC/blob/2cb0f3e5819b03ad354b8b35f9f4652d75a065c1/desc/compute/_fast_ion.py#L183).
1. Removing `v_tau` from that line alone avoids the issue.
2. Redefining the integrand `_v_tau` to be 1, so that `v_tau` is computed as the length of the bounce integral $\int d\ell$ the `nan` persists.

Therefore, the nan gradient in forward mode arises from the computation of $d\ell /d \zeta = B / B^\zeta$. I have confirmed that this quantity is well-behaved.  Also, this can be computed in other routines at the exact same points with $\epsilon_{effective}$, and there the issue does not occur. 

FYI, I have found the JAX `nan` [debugging tools](https://jax.readthedocs.io/en/latest/debugging/flags.html) useful before, but for this problem they are not. Maybe related to https://github.com/jax-ml/jax/pull/25519.

Below is MWE. 

```python
eq0 = get("W7-X")
eq1 = eq0.copy()
k = 2  # which modes to unfix
print()
print("---------------------------------------")
print(f"Optimizing boundary modes M, N <= {k}")
print("---------------------------------------")
modes_R = np.vstack(
    (
        [0, 0, 0],
        eq1.surface.R_basis.modes[np.max(np.abs(eq1.surface.R_basis.modes), 1) > k, :],
    )
)
modes_Z = eq1.surface.Z_basis.modes[np.max(np.abs(eq1.surface.Z_basis.modes), 1) > k, :]
constraints = (
    ForceBalance(eq=eq1),
    FixBoundaryR(eq=eq1, modes=modes_R),
    FixBoundaryZ(eq=eq1, modes=modes_Z),
    FixPressure(eq=eq1),
    FixIota(eq=eq1),
    FixPsi(eq=eq1),
)
grid = LinearGrid(
    rho=np.linspace(0.2, 1, 3), M=eq1.M_grid, N=eq1.N_grid, NFP=eq1.NFP, sym=False
)
    (
        # Both Gamma_c Nemov and Gamma_c Velasco are affected. Effective ripple is not.
        GammaC(
           # pick anything
            eq1,
            grid=grid,
            X=16,
            Y=32,
            Y_B=128,
            num_transit=5,
            num_well=20 * 10,
            num_quad=8,
            num_pitch=10,
            deriv_mode="fwd"
        ),
    )
)
optimizer = Optimizer("proximal-lsq-exact")
(eq1,), _ = optimizer.optimize(
    eq1,
    objective,
    constraints,
    ftol=1e-4,
    xtol=1e-6,
    gtol=1e-6,
    maxiter=1,  # increase maxiter to 50 for a better result
    verbose=3,
    options={"initial_trust_ratio": 2e-3},
)
```

### LINKED ISSUE #2168: Sparsity preserving pullbacks
I will drop link to a useful technique I came up with that was valuable to me in another project to leverage sparsity. It would greatly benefit DESC as well.

#### example 
As an example, replace this [line](https://github.com/PlasmaControl/DESC/blob/c42a92b4db17f4ebb880153fa16df90d10efe7c0/desc/integrals/bounce_integral.py#L603) with the logic `sparse_cotangent(fun)(fun_data)` I wrote [here](https://github.com/jax-ml/jax/issues/36862#issuecomment-4265767967) (or some combination of the two). You will need to bypass the typical jax closure annoyances that arise due to how DESC wraps the jvp funs), but that should be standard (see equinox wrappers).
- With a fair comparison, the speed and memory of computing the Jacobian of $f \colon R^M \to R^N$ will improve by a factor of $N$.
- If you apply it to an objective that calls a function which is checkpointed, the speed could increase by a factor of $N^2$, but that assumes the checkpointed part is dominant computation.
- If the function you wrap with this is the dominant expense, then materializing the Jacobian of any objective is the same cost as a single vjp was before.

unalmis:
PR Age

If this will not be merged then all bounce integrals should be removed per #2225 

- [x] Resolves #2168 .
  - Yields factor of $N \times$ improvement in memory and compute speed to get Jacobian $R^{N \times M}$ of map $f \colon R^M \to R^N$. Basically if you had some expensive function, with proper application throughout you could get Jacobian for around the cost you'd get a single row before.
  - $N^2$ for check pointed funs
  - Even if you are only doing a single vjp, it is much faster and more memory effecient due to reasons discussed in links threads.
- [x] Removes
  - `bounce1d` optimization.
  - Unneeded flags (`is_reshaped`, `is_fourier`) that users said were confusing (backwards compatible) as well as the developer flags `Bref`, `Lref` that should not be there. 
- [x] Previously, ``pitch_batch_size`` was getting ignored. This fixes that by adding ``strip_dim0`` flag to ``batch_map``.
- [x] #2182 
- [x] #2185 
- [x] resolves #1489 on new jax versions 

## notes

- They are decorator functions, so it is non-invasive and can improves many objective such as bounce, balloon, qs, omni, surface integral, free surface etc, but more generally can apply to any atomic computation. Trivial to use now that I have done all the math and resolved implementation sharp bits.
- Note the CI benchmarks won't show that improvement since those objectives were onto $R^1$, but benchmarking with tensor-board indeed shows the factor of $N$ improvement in speed and memory computing the Jacobian. And this is not yet applied to the full pipeline. See #2171 .
- If the partial summation in #1508 is done so that the the transforms are factorized then we can extend the pullback wrappers to include the spectral to real space transform on each surface to by decorating the larger function with a sparse_pullback as well. Or use Fourier-Chebysev as i had suggested in #1243 to make it simpler. This would have thr advantage of further reducing memory. 
- This should renew interest in #1206 and #1872 and #465.


docs/notebooks/tutorials/EffectiveRipple.ipynb
unalmis, line 1: the optimization speed here has like a +- 6 minute variance on my computer depending on how hot it is, so if you want to measure compute or jacobian speed up you should just run python effective_ripple_profile with different settings for various batch sizes etc. most of the time is doing svd now

desc/derivatives.py
```diff
@@ -10,7 +10,209 @@
...
+
+
+@_sparse_pullback.def_fwd
+def _sparse_pullback_fwd(perturbed, y, *, fn):
+    out, vjp_fn = eqx.filter_vjp(fn, y)
+    # JAX cannot infer jvp from vjp for higher order derivative, so without
+    # further loss of generality, we stop tracing higher order derivative
+    # flow for debugging and performance benefits. We can stop this once JAX
+    # supports defining a custom_jvp and custom_vjp simultaneously.
+    return out, jax.lax.stop_gradient(vjp_fn(jnp.ones_like(out))[0])
```
unalmis, line 42: I just did this to ease debugging, but I'm sure it works without. not really a point though.

unalmis:
just text me when I can click merge.

desc/integrals/bounce_integral.py
unalmis, line 1: this just moves the two methods _defaults and _objective_build into the Options class.

desc/compute/_fast_ion.py
unalmis, line 1: changes in these compute files are so the closure conversion works better. also removed unneeded kwargs

desc/compute/_fast_ion.py
```diff
@@ -109,46 +108,28 @@ def _Gamma_c(params, transforms, profiles, data, **kwargs):
...
-            Y_B,
-            alpha,
-            num_transit,
-            quad,
-            nufft_eps=nufft_eps,
-            is_fourier=True,
-            spline=spline,
-            vander=vander,
+        pitch_inv, weight = Bounce2D.get_pitch_inv_quad(
+            data["min_tz |B|"], data["max_tz |B|"], opts.pitch_quad
```
unalmis: to reduce size of vjp

desc/integrals/bounce_integral.py
unalmis: and adds the spare call in def batch, and removes params is_fourier and is_reshaped to simplify for user since they can be figured out internally. all backwards compatible.

desc/compute/_fast_ion.py
unalmis: don't have to check since we have tests in test_compute_everything.

desc/compute/_old.py
unalmis: can delete this after someone does #2152

f0uriest (commented):
The sparse pullback stuff seems cool and gives a big performance boost, but again it seems like the vast majority of changes in this PR are unrelated to that. I would greatly prefer multiple small, self contained PRs rather than these giant monoliths that are impossible to review and actually understand.

desc/integrals/bounce_integral.py
```diff
@@ -894,11 +751,11 @@ def integrate(
         *,
         num_well=None,
         nufft_eps=1e-6,
-        is_fourier=False,
-        low_ram=False,
+        loop=False,
```
unalmis: rory wanted nufft eps = None then setdefault to self.nuffteps

unalmis:
when is this getting merged

desc/compute/_old.py
```diff
@@ -304,56 +253,44 @@ def _Gamma_c_Velasco_1D(params, transforms, profiles, data, **kwargs):
         J.L. Velasco et al. 2021 Nucl. Fusion 61 116059.
         https://doi.org/10.1088/1741-4326/ac2994.
 
-    This expression has a secular term that drives the result to zero as the number
```
f0uriest, line 307: is this comment no longer relevant?

tests/baseline/test_plot_gammac.png
f0uriest, line 1: this seems to have changed quite a lot, I thought this PR was just changing AD stuff?

CHANGELOG.md
```diff
@@ -17,17 +17,28 @@ New Features
...
 - [Significantly improves convergence of inverse stream maps](https://github.com/PlasmaControl/DESC/pull/1919).
-- Check-pointing to bounce integrals to improve speed and reduce memory of reverse mode differentiation.
 - Resolves a JAX memory regression in bounce integrals by avoiding materialization of a large tensor in memory. Previously, we had closed the issue by adding nuffts as a workaround. This update actually solves the issue for the case when a user specifies to not use nuffts as well.
 
+Breaking Changes
+
+- The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``num_field_periods``. This should make using a consistent resolution across different equilibria easier.
+- The parameter ``Y_B`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` is now the resolution over a single field period rather than a full toroidal transit. This should make using a consistent resolution across different equilibria easier.
+- The method name ``get_pitch_inv_quad`` has been shortened to ``pitch_quad`` to
```
f0uriest: if this isn't public than no need to mention it, if it is public then it shouldn't be changed without a deprection period and backwards compatibility

CHANGELOG.md
```diff
@@ -17,17 +17,28 @@ New Features
...
 - Reduces import time of `desc` modules.
     - Now, `desc.compute._build_data_index` uses depth-first search algorithm to construct the dependency tree.
     - Some of the default value computations at import time are removed (i.e. `desc.integrals.bounce_integral.default_quad`)
 - [Significantly improves convergence of inverse stream maps](https://github.com/PlasmaControl/DESC/pull/1919).
-- Check-pointing to bounce integrals to improve speed and reduce memory of reverse mode differentiation.
 - Resolves a JAX memory regression in bounce integrals by avoiding materialization of a large tensor in memory. Previously, we had closed the issue by adding nuffts as a workaround. This update actually solves the issue for the case when a user specifies to not use nuffts as well.
 
+Breaking Changes
+
+- The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``num_field_periods``. This should make using a consistent resolution across different equilibria easier.
```
f0uriest: I worry that `num_field_periods` will get confused with `NFP`, since it seems like passing `num_field_periods=eq.NFP` would generally be way too low. We could probably keep the old name and just make it clear in the docstring that transit means number of field period transits, not full toroidal transits

desc/objectives/_fast_ion.py
```diff
@@ -97,47 +83,51 @@ def __init__(
...
         num_pitch=65,
         pitch_batch_size=None,
         surf_batch_size=1,
         nufft_eps=1e-7,
         spline=True,
-        use_bounce1d=False,
         Nemov=True,
-        **kwargs,
     ):
+        errorif(
```
f0uriest, line 108: we should still support forward mode even if its slower

desc/objectives/_fast_ion.py
```diff
@@ -97,47 +83,51 @@ def __init__(
         normalize_target=True,
         loss_function=None,
         deriv_mode="auto",
-        jac_chunk_size=None,
```
f0uriest, line 105: why is this no longer an option?

desc/objectives/_neoclassical.py
```diff
@@ -67,46 +60,50 @@ def __init__(
...
         num_quad=32,
         num_pitch=51,
         pitch_batch_size=None,
         surf_batch_size=1,
         nufft_eps=1e-6,
         spline=True,
-        use_bounce1d=False,
-        **kwargs,
     ):
+        errorif(
```
f0uriest, line 86: see previous

desc/objectives/_neoclassical.py
```diff
@@ -67,46 +60,50 @@ def __init__(
         normalize_target=True,
         loss_function=None,
         deriv_mode="auto",
-        jac_chunk_size=None,
```
f0uriest, line 75: see previous

desc/batching.py
```diff
@@ -197,6 +197,21 @@ def vmap_chunked(
     - https://github.com/jax-ml/jax/issues/26689
     - https://github.com/jax-ml/jax/issues/27591
     - https://github.com/jax-ml/jax/issues/31919
+    - Due to an actively worked on issue in JAX,
+      https://docs.jax.dev/en/latest/jep/
```
f0uriest, line 202: according to the linked issue (https://github.com/jax-ml/jax/issues/1249) this was fixed 6 years ago

desc/derivatives.py
```diff
@@ -10,7 +10,197 @@
...
+
+@eqx.filter_custom_vjp
+def _sparse_pullback(y, *, fn):
+    return fn(y)
+
+
+@_sparse_pullback.def_fwd
+def _sparse_pullback_fwd(perturbed, y, *, fn):
+    out, vjp_fn = eqx.filter_vjp(fn, y)
+    p = eqx.filter(vjp_fn(jnp.ones_like(out))[0], perturbed)
```
f0uriest: note for self and other devs:
we're assuming that fn is diagonal along the first argument, so vjp(ones) is computing the diagonal of the jacobian. Then below in the backward pass we multiply that with the actual tangents that we want to pull back to get the full correct vjp.

desc/derivatives.py
```diff
@@ -10,7 +10,197 @@
...
+
+    return tree_map(apply, p)
+
+
+def sparse_pullback_map(fn, y):
+    """Wrapper for sparsity exploiting pullback.
+
+    Wraps the given map with logic to ensure cotangents flow through the diagonal
+    of its pullback. The derivatives will be exact for maps whose Jacobians are
+    block diagonal.
```
f0uriest: don't we need strictly diagonal? ie any dense jacobian is block diagonal with 1 block

desc/derivatives.py
```diff
@@ -10,7 +10,197 @@
...
+    fn = eqx.filter_closure_convert(fn, y)
+
+    @wraps(fn)
+    def wrapper(y):
+        return _sparse_pullback(y, fn=fn)
+
+    return wrapper
+
+
+def sparse_pullback(
```
f0uriest: looks like several paths through this function aren't tested, I think we should add small tests specifically for this to cover all bases

f0uriest (commented):
Still about 1/2 to go but leaving these here for now.

One big point is that if we're messing with custom AD stuff I think it would be good practice to add tests comparing AD of the relevant objectives to finite differences (can use very low res, don't care about physics convergence), both to check that the implementation is correct and also to guard against us accidentally applying the sparse pullback in places where its not strictly correct.

unalmis:
> Still about 1/2 to go but leaving these here for now.
> 
> One big point is that if we're messing with custom AD stuff I think it would be good practice to add tests comparing AD of the relevant objectives to finite differences (can use very low res, don't care about physics convergence), both to check that the implementation is correct and also to guard against us accidentally applying the sparse pullback in places where its not strictly correct.

- There are hundreds of tests for it
- Locally I tested with the jacobian equivalent of `test_compute_everything`. It worked to machine precision.
- I tested finite difference a bunch for these objectives, and it works well.
- I prefer to keep the tests I add as independent unit tests that test correctness of code.
- There are enough monolithic smoke tests. Part of the reason theses pr's are big is because I have to plumb through changes for the 100 different tests I have for these objectives. Adding more smoke tests makes it harder to maintain, annoys reviewers for pr leght, and my hunch is someone would eventually decide to delete for #914 

> > (can use very low res, don't care about physics convergence),

That is not true/possible. See the supplementary information in publications. Briefly, For nontrivial computational problems where not everything is C^infinty, an algorithm to solve a problem needs to have amazing convergence properties, and be robust to topology changes, for the duscretization error to be correlated enough nearby a given point in the optimization space for the finite difference derivative to have any chance if being accurate. (Again explained better in the pdf).

You can see that finite difference derivatives only make sense at high resolution computations of the algorithm. Auto diff makes sense at any resolution because it estimates the derivative from only information at a single point in the optimization landscape. (Of course if discretization error is high then over an optimization it could still stall as varying discretization error can affect the decent direction,but that's unrelated for this discussion). In general you'll need high res to get finite diff to match auto diff.

tests/baseline/test_plot_gammac.png
unalmis: It changed because the default resolution parameters changed. Dan wanted `Y_B` per field period. Also, we added the halley step on bounce points. docstrings mentions affect of `y_b`.

desc/objectives/_fast_ion.py
unalmis: See #1489 . Forward mode is supported now.

desc/objectives/_fast_ion.py
unalmis: Because we do not want to support jac_chunk_size ever for these objectives. It is mathematically proven to be inferior to sparse pullback.

desc/objectives/_neoclassical.py
unalmis: dido

desc/batching.py
unalmis: it's still an open issue for 2026 according to their design docs.

desc/derivatives.py
unalmis: yea so if you follow the link that is in the docstring i explain this in detail too

desc/derivatives.py
unalmis: No, all the bounce integral stuff is block diagonal not strictly diagonal. Strictly diagonal implies input and output are same shape. Block diagonal means the pushforwad can be diagonal and the pullback might not, or vice versa. For bounce integrals, the pushforward is not block diagonal. The pullback is block diagonal.

unalmis: yes i know its the same matrix but the term block diagonal describes the structure of the matrix - it is block diagonal if it is uniqulely determined by a pullback of ones or a pushfoward of ones. strictly diagonal replaces or with and.

CHANGELOG.md
unalmis: i have decided it is not public.

CHANGELOG.md
unalmis: transit is more confusing because it implies closure of the trajectory or at least a revolution. and then there is poloidal vs toroidal transit.

desc/derivatives.py
unalmis: i'm assuming it is block diagonal. could be "diagonal" along first n axes as you have used it.

desc/derivatives.py
unalmis: > ie any dense jacobian is block diagonal with 1 block

people use diagonal to mean lack of mixing. If you refer to a desnse jacobian as diagonal with 1 block, then means you are must also refer to your entire set of output basis vectors as 1 block. then obviously it is impossible to mix between blocks since you only have one.

desc/compute/_old.py
unalmis: Any weighted average, where the weights are bounded) of a sequence whose limit is zero is zero. It is an open and shut proof; no room for dispute. So I remove it because its not a resolution thing. In `ku/gamma` documentation is added to explain when these metrics are useful.

desc/integrals/_bounce_utils.py
```diff
@@ -838,25 +922,33 @@ def fast_chebyshev(theta, f, Y, num_t, modes_t, modes_z, *, vander=None):
...
-        cheb_pts(Y, theta.domain)[:, None] if vander is None else None,
+        z_eff,
         f,
         theta.domain,
         axis=-2,
         modes=modes_z,
         vander=vander,
     )[..., None, None, :, :]
-    f = irfft_mmt_pos(theta.evaluate(Y), f, num_t, modes=modes_t)
+    f = irfft_mmt_pos(theta.evaluate(Y), f, n=jnp.nan, modes=modes_t)
```
f0uriest, line 950: what is `n=jnp.nan` doing here? as far as i can tell its just a placeholder sentinel that gets ignored because modes is given explicitly? If so I'd prefer some other value so as not to trip the jax.debug_nans thing with false positives. (we already have hardcoded nans in other places but the fewer the better, and we should eventually replace those as well, the times ive had to debug nans in desc has been a huge pain)

desc/integrals/_bounce_utils.py
```diff
@@ -959,34 +1039,37 @@ def fast_cubic_spline(
...
+    t = idct_mmt(x, theta.cheb[..., None, :], vander=vander_t)
+    assert t.shape == (*lines, num_field_periods, Y)
 
-    if nufft_eps < 1e-14 or f.shape[-1] < 14:
-        # second condition for GPU
-        f = f[..., None, None, None, :, :]
-        f = irfft_mmt_pos(t, f, num_t, modes=modes_t)
+    if no_nufft(nufft_eps) or f.shape[-1] <= 16:
+        f = f[..., None, None, :, :]
+        f = irfft_mmt_pos(t, f, n=jnp.nan, modes=modes_t)
```
f0uriest, line 1052: same as prev about hardcoded nan

desc/integrals/bounce_integral.py
```diff
@@ -1014,77 +874,74 @@ def integrate(
...
+        data["|B|"] = B if is_chebyshev else mmt_for_bounce(z, t, self._c["|B|"])
         data["zeta"] = zeta
-
         return data
 
-    def interp_to_argmin(
-        self, f, points, *, nufft_eps=1e-6, is_fourier=False, **kwargs
-    ):
-        """Interpolate ``f`` to the deepest point pⱼ in magnetic well j.
+    def interp_to_argmin(self, f, points, *, nufft_eps=-1.0, **kwargs):
```
f0uriest, line 975: the docstring still says that if nufft_eps<1e-14 than they won't be used, but we also have if <0 use default?

f0uriest: Usually `None` is the sentinel for "use a default value", I'm assuming you want a float here to avoid recompilation? But if this is basically always wrapped in some outer jit at the objective level that shouldn't make any difference here.

desc/integrals/bounce_integral.py
```diff
@@ -1627,7 +1460,7 @@ def reshape(grid, f):
         """
         return grid.meshgrid_reshape(f, "raz")
 
-    def points(self, pitch_inv, num_well=None):
+    def points(self, pitch_inv, num_well=-1):
```
f0uriest, line 1516: see previous about `None` for default

desc/integrals/bounce_integral.py
f0uriest, line 1: `Bounce{1,2}D` are part of the public API (documented on RTD), and there are a bunch of API breaking changes here (renamed methods, renamed args, removed args etc). 

Given that we already know of several downstream users using these things for their own code we should make sure to gracefully catch these issues and throw deprecation warnings. I'd vote for not renaming methods at all (`get_pitch_inv_quad` -> `pitch_quad` doesn't really seem like a major readability/clarity improvement, but it will break user code). For deleted args we should still accept them in `**kwargs` and throw a warning to make clear that they don't have any effect and explain any substitutions that should be made.

desc/objectives/_fast_ion.py
f0uriest: In that case we should still accept it as part of `kwargs` and throw a deprecation warning to remind users to update their code if they were using this option.

desc/objectives/_fast_ion.py
f0uriest: The other issue here is that using vjp prevents using standard jax.hessian which is forward over reverse. If reverse is really always significantly faster than that may be ok, but it's worth getting other folks input.

If we do end up only supporting reverse mode it would be good to add to the `Notes` section of the docstring to remind users (and maybe eventually having a reverse only version of `objective.hess`)

desc/objectives/_neoclassical.py
```diff
@@ -67,46 +60,50 @@ def __init__(
...
         name="Effective ripple",
         grid=None,
         X=32,
         Y=32,
         Y_B=None,
-        alpha=jnp.array([0.0]),
-        num_transit=20,
+        alpha=None,
+        num_field_periods=20,
```
f0uriest: Not sure what happened to my previous comment but I worry that users will confuse this with `NFP`, plus its an argument name change that may break peoples code without proper handling. My vote would be to just keep the old name and update the docstring to explain it (`num_transits` still seems like a fair descriptor).

If you really want a new name I'd vote for `field_period_transits` to distinguish from NFP and make it clear that its to do with the field line/orbit rather than the underlying equilibrium. And it will need a `parse_argname_change` thing for backwards compatibility

desc/objectives/_neoclassical.py
```diff
@@ -67,46 +60,50 @@ def __init__(
         normalize_target=True,
         loss_function=None,
         deriv_mode="auto",
```
f0uriest: i think for the other reverse mode only objectives we changed the default to "rev" to make that clear? Also just in case there's some pathological case where our auto logic would select forward

desc/objectives/_fast_ion.py
f0uriest: It looks like in theory forward mode could still be possible by passing `sparse=False` when computing the bounce integrals? (assuming #1489 is fixed)

CHANGELOG.md
f0uriest: Ah ok i thought this was farther up.

I don't think it usually implies closure of the trajectory, eg in flux tubes it generally means how many turns around the device, but I think in reality what that means is the same as what we have here, "field period transits" rather than full toroidal transits.

desc/integrals/_bounce_utils.py
unalmis: this does not trip debug nan. It is much better programming practice to use nan to ensure the modes are used directly.

unalmis: debug nan is tripped when an nan is placed into an array. it is not tripped for setting a global variable to nan before compile time.

desc/integrals/_bounce_utils.py
unalmis: See previous comment; that is not how debug nan detects nan. (Your earlier thing is due to your incorrect understanding of JAX. It doesn't work that way)

desc/integrals/bounce_integral.py
unalmis: Any invalid input is a proper sentinel for a default value. Negative absolute error is an invalid input, therefore it is a valid sentinel for default value. This is common in python. Yes float avoid recomp. You should not assume its wrapped in outer jit. Moreover, you should always use proper programming techniques that work at every level rather than having a single failure point. and JAX has bugs.

desc/integrals/bounce_integral.py
unalmis: It is documented in the docstring, that `None` and and `-1`  have different behavior. Hence, this suggestion is formally not valid here.

desc/integrals/bounce_integral.py
unalmis: > and there are a bunch of API breaking changes here (renamed methods, renamed args, removed args etc).

No, the only API breaking changes are the ones discussed in the changelog.

> For deleted args we should still accept them in **kwargs and throw a warning to make clear that they don't have any effect and explain any substitutions that should be made.

This is done if there is a substitution to be made that cannot be automated.

desc/objectives/_neoclassical.py
unalmis: your previous comment remains where it was

unalmis: poloidal transit means 0 to 2pi. toroidal transit means 0 to 2pi. in either case, transit does not mean 0 to 2pi/NFP. if you have personal, cosmetic, preference for field_period_transits, as discussed in earlier PR comment, such cosmetic changes should be made on new PR that are childs of #2220 to avoid merge conflicts and days of retesting.

desc/objectives/_fast_ion.py
unalmis: There is nothing for users to update. Therefore a warning should not be thrown.

desc/objectives/_fast_ion.py
unalmis: Forward over reverse reverse over reverse etc. is all supported.

In any case, it would be wrong to make this PR contingent on such separate work, and it would be wrong to spend time doing such separate work before this is merged.

unalmis: > standard jax.hessian which is forward over reverse. If reverse is really always significantly faster than that may be ok

jax.hessian does not work on master for this in the first place; so the comment is irrelevant.

In any case you should be considering memory. forward without spare jacobian would barely fit in memroy. hessian has no chance unless you use sparsity. so I believe your suggested use case of someone being able to use objective with hessian is imaginary. Even if hess on master worked for these objectives, with the approach there it would impossible to fit in memory. I am not obligated to support 3rd order Hessian optimization for these objectives anyway because that would not improve optimization due to smoothness.

Therefore, it should not be in discussion in this PR.

desc/objectives/_fast_ion.py
unalmis: if you have personal cosmetic preference agin you can do that later

desc/objectives/_fast_ion.py
ddudt: [thought(blocking)](https://pullpo.io/cc?l=thought&d=blocking) 
Forward mode is actually very common. Individual objectives often prefer reverse mode, but typically we optimize with a list of many different objectives. If you set `ObjectiveFunction(objectives, deriv_mode="batched")` then it will use `deriv_mode="fwd"` for all of the sub-objectives. From the docs:

> deriv_mode : {"auto", "batched", "blocked"}
        Method for computing Jacobian matrices. ``batched`` uses forward mode, applied
        to the entire objective at once, and is generally the fastest for vector
        valued objectives. ...

So unfortunately I think we do need to make this PR contingent on a solution to using forward mode, because otherwise these changes will break user code. Also I think we have been very careful to always support both forward and reverse mode for all objectives as a design principal, so at the very least we should discuss the consequences in more detail before we abandon it.

desc/objectives/_fast_ion.py
ddudt: [suggestion](https://pullpo.io/cc?l=suggestion) 
Even if `jac_chunk_size` isn't used in the code anymore, I think we should still throw a deprecation warning. If users had been passing this argument before, we want to warn them that they can no longer do what they had been doing before.

ddudt (commented):
I haven't reviewed all the code yet, but I think this introduces some breaking changes. I'm not sure what the best solution is: ideally we would keep support for forward mode, but we at least need to add more error handling.

desc/objectives/_fast_ion.py
YigitElma: I think to prevent weird errors, we should add an additional check to `ObjectiveFunction.build()` method. Currently, we assume that all the objectives support forward and reverse mode differentiation, and we allow the users to choose `ObjectiveFunction(..., deriv_mode="batched")` which uses forward mode for everything, and silently overrides sub-objective deriv_modes to be "fwd". The check you have here won't be triggered in that case.

If #1489 still persists, I am fine removing forward mode for these objectives (which gives nans anyway), but the check I mentioned should be added since the assumption is now broken. You can add it after this one, for example, https://github.com/PlasmaControl/DESC/blob/1c7178d66a8cf7d839b58ff5d6aa366bb913c016/desc/objectives/objective_funs.py#L506

unalmis: Yigit can add his warning if he wants.

Rory and I had discussed that if my PRs which update my work will not be added to DESC, then bounce integrals should be removed as we don't want a poor/inefficient version made available on the most public platform when the published work under my name makes claims about the efficient version. That would lead to users not using my work; having a bad impression of it and reduce its impact. If Daniel wants to block it, please remove bounce integrals from DESC and add deprecation warnings for users to use https://github.com/unalmis/DESC.

unalmis:
> https://github.com/PlasmaControl/DESC/pull/2170#discussion_r3415951594

https://github.com/PlasmaControl/DESC/pull/2170#discussion_r3416922786

desc/objectives/_fast_ion.py
unalmis: They can do what they were doing before. The computation now is more efficient than jac_chunk_size=1 was before. What would the warning be?

@ddudt feel free to commit a deprecation warning as follows.

"Note that the computation is now faster than it would be with jac_chunk_size=infinite and uses less memory than jac_chunk_size=1. so jac chunk size does not do anything anymore. feel free to ignore this message as there is nothing actionable. Also feel free to make an issue to remove this message if the deprecation warning is annoying".

desc/integrals/bounce_integral.py
unalmis: and users jit this.

unalmis:
@ddudt @daniel-dudt it works in forward mode now. Master does not work in forward mode. So your blocking comment is not logical.

unalmis:
@dpanici when will this be merged?

unalmis:
@ddudt when will this be merged

unalmis:
It's needed for #2257

unalmis:
@f0uriest when will this be merged?

desc/integrals/_bounce_utils.py
```diff
@@ -69,6 +68,10 @@ def bounce_points(pitch_inv, knots, B, num_well=-1, return_mask=False):
 
         If there were fewer wells detected along a field line than the size of the
         last axis of the returned arrays, then that axis is padded with zero.
+    sentinel : float
+        Sentinel value which should be less ζ coordinate of all bounce points,
```
dpanici:
```suggestion
        Sentinel value which should be less than ζ coordinate of all bounce points,
```

desc/objectives/_neoclassical.py
```diff
@@ -71,47 +66,54 @@ def __init__(
...
+        deriv_mode="rev",
         name="Effective ripple",
         grid=None,
         X=32,
         Y=32,
         Y_B=None,
-        alpha=jnp.array([0.0]),
-        num_transit=20,
+        alpha=None,
+        num_field_periods=20,
```
dpanici: It is not important to me that we change the name here, but we can at least do

```python
if "num_transit" in kwargs:
    warnif(True ,DeprecationWarning,
    "argument num_transit has been deprecated in favor of num_field_periods,  converting to num_field_periods = num_transit*eq.NFP)
    num_field_periods=kwargs.pop("num_transit") * eq.NFP
```

desc/integrals/_bounce_utils.py
```diff
@@ -723,10 +809,10 @@ def theta_on_fieldlines(angle, iota, alpha, num_transit, NFP, *, X_min=24):
     alpha : jnp.ndarray
```
dpanici, line 808: does this function or elsewhere in Bounce2D require that the alphas be the same across all rho surfaces? 

Just wondering, if not, we could allow them to be different per surface and that would aid #2257 . Not suggesting we do this in this PR or anything like that, just asking out of informational reasons for future reference.

desc/integrals/bounce_integral.py
```diff
@@ -286,263 +288,92 @@ def __init__(
...
         check=False,
         vander=None,
         **kwargs,
     ):
         """Returns an object to compute bounce integrals."""
         assert grid.can_fft2
-        is_reshaped = is_reshaped or is_fourier
-        vander = setdefault(vander, {})
+        if "num_transit" in kwargs and num_field_periods == 20:  # default value
+            num_field_periods = kwargs["num_transit"] * grid.NFP
```
dpanici: This is ok with me but I'd prefer to also throw a warning to let user know that num_transit is not being used, instead of silently using the default, something like (sorry if this causes some formatting issue, the suggestions are hard to ensure they respect the column limit)

```suggestion
        if "num_transit" in kwargs and num_field_periods == 20:  # default value
            warnings.warn("`num_transit` kwarg is deprecated in favor of `num_field_periods`. Using `num_field_periods = num_transit*NFP`.", DeprecationWarning)
            num_field_periods = kwargs["num_transit"] * grid.NFP
```

I can make this change if you want

desc/derivatives.py
unalmis: It's all done and it was always tested

desc/integrals/bounce_integral.py
```diff
@@ -945,12 +808,9 @@ def integrate(
         nufft_eps : float
             Precision requested for interpolation with non-uniform fast Fourier
             transform (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
```
dpanici, line 839: would be good to add "if <0, then the Bounce2D's self._nufft_eps will be used"

desc/integrals/bounce_integral.py
```diff
@@ -1019,77 +882,74 @@ def integrate(
...
-    def _nufft(self, x, z1, z2, data, low_ram, eps, pitch_inv):
-        shape = (*z1.shape, x.size)
+    def _nufft(self, x, z1, z2, data, loop, eps, pitch_inv):
+        shape = z1.shape + (x.size,)
+
+        keys = list(data.keys())
+        keys.append("B^zeta")
+        c = list(data.values())
+        c.append(self._c["B^zeta"])
+        if not (is_chebyshev := isinstance(self._B, PiecewiseChebyshevSeries)):
```
dpanici, line 921: Nice use of `:=`, did not realize one could do this before now

desc/integrals/bounce_integral.py
```diff
@@ -1897,9 +1751,338 @@ def plot(self, l, m, pitch_inv=None, **kwargs):
...
+            For Chebyshev series, this is the max degree floor division by 2.
+
+        Returns
+        -------
+        num_well : int
+            A guess for the max number of wells that exist for any pitch angle
+            or field line after following it for the specified length.
+            The guess will ideally be more conservative than
+            ``num_field_periods*mins_per_field_period`` to enhance performance,
+            yet sill remain loose enough that all wells are always detected.
```
dpanici:
```suggestion
            yet still remain loose enough that all wells are always detected.
```

desc/integrals/bounce_integral.py
```diff
@@ -1897,9 +1751,338 @@ def plot(self, l, m, pitch_inv=None, **kwargs):
...
+            kwargs["z1"], kwargs["z2"] = _bounce_points(
+                pitch_inv, self._knots, B, sentinel=self._sentinel
+            )
             kwargs["k"] = pitch_inv
-        fig, ax = plot_ppoly(
-            PPoly(B.T, self._zeta), **set_default_plot_kwargs(kwargs, l, m)
+        return plot_ppoly(PPoly(B.T, self._knots), **kwargs)
+
+
+class Options(NamedTuple):
```
dpanici: I like this. If this is only supposed to be used for Bounce2D options, would it be worth specifically naming it Bounce2DOptions or OptionsBounce2D instead of Options? In case in the future another Options-like object becomes used elsewhere

desc/integrals/_bounce_utils.py
dpanici: this comment I think is out of date in lieu of comment [here](https://github.com/PlasmaControl/DESC/pull/2257/changes#r3648222941), seems you've progressed towards this already on #2215

desc/objectives/_fast_ion.py
```diff
@@ -151,13 +143,6 @@ def __init__(
...
             "spline": spline,
         }
-        if use_bounce1d:
-            self._hyperparam.pop("X")
-            self._hyperparam.pop("Y")
-            self._hyperparam.pop("pitch_batch_size")
-            self._hyperparam.pop("nufft_eps")
-            self._hyperparam.pop("spline")
-
         self._key = "Gamma_c" if Nemov else "Gamma_c Velasco"
```
dpanici, line 161:
```suggestion
    warnif("jac_chunk_size" in kwargs, DeprecationWarning, "ignoring jac_chunk_size as it is not used for this objective")
```

desc/objectives/_neoclassical.py
```diff
@@ -120,12 +122,6 @@ def __init__(
             "nufft_eps": nufft_eps,
             "spline": spline,
         }
-        if use_bounce1d:
-            self._hyperparam.pop("X")
-            self._hyperparam.pop("Y")
-            self._hyperparam.pop("pitch_batch_size")
-            self._hyperparam.pop("nufft_eps")
-            self._hyperparam.pop("spline")
```
dpanici, line 139:
```suggestion
    warnif("jac_chunk_size" in kwargs, DeprecationWarning, "ignoring jac_chunk_size as it is not used for this objective")
```

CHANGELOG.md
```diff
@@ -49,21 +48,30 @@ New Features
...
 - ``ObjectiveFunction.print_value`` can now use the previously computed ``compute_scaled_error`` values to print. For bounded objectives, we fall back to computing ``compute_unscaled``. Additionally, ``compute_scaled_error`` and array splitting are used in other parts of the code to prevent recompilation for one-time tasks, which makes initialization faster.
 
+Breaking Changes
+
+- The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``num_field_periods``. This should make using a consistent resolution across different equilibria easier. ``num_transit`` may still be used but note the equivalence ``num_field_periods = num_transit * grid.NFP``.
+- The parameter ``Y_B`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` is now the resolution over a single field period rather than a full toroidal transit. This should make using a consistent resolution across different equilibria easier.
+
```
dpanici:
```suggestion
- objectives using ``Bounce2D`` now do not support fwd mode differentiation for JAX versions <0.11.0. These objectives' derivatives are inefficient in fwd mode, so rev mode should be used for them, which is significantly more efficient with the introduction of sparse reverse-mode differentiation. 
```

dpanici (changes requested):
Thanks for all the work that went into this PR! I would like these cosmetic changes made. I can do them myself if you prefer. 

Main blocking comment I have is on the removal of jac_chunk_size: I get that these changes make the code much faster than before and make jac_chunk_size obsolete. It is good practice to then just throw a DeprecationWarning that it is no longer used or needed for this objective given these changes. I can make that commit as well (already made the suggestion in a comment).

The others are spelling or small warnings.

again, if you prefer I can make these changes myself as they are cosmetic. These are my main comments. 

The other thought is that the `Options` class is a great idea and maybe could be used elsewhere outside of Bounce in the future, but I will leave that for a future problem.

CHANGELOG.md
ddudt, line 1: Remember to move all of these changes up in the file now that there is a new version.

desc/compute/utils.py
```diff
@@ -88,7 +88,7 @@ def compute(  # noqa: C901
                     "instead.",
                     DeprecationWarning,
                 )
-    bad_kwargs = kwargs.keys() - allowed_kwargs
+    bad_kwargs = kwargs.keys() - allowed_kwargs - {"num_transit"}
```
ddudt, line 91: Should we just add this to `allowed_kwargs`? Or maybe make a new set called `deprecated_kwargs` to keep track of stuff like this?

desc/objectives/objective_funs.py
```diff
@@ -267,6 +269,9 @@ def collect_docs(
 
     # Copy to allow for objective-specific updates to docs
     docs_obj = docs.copy()
+    if not jac_chunk_size:
+        del docs_obj["jac_chunk_size"]
```
ddudt: [question](https://pullpo.io/cc?l=question) 
Why do these objectives not need `jac_chunk_size`? If I understand correctly, `jac_chunk_size` is only used for `deriv_mode="fwd"` and normally we should be using reverse mode for these objectives. But these objectives should still work in forward mode (at least with the newest version of JAX), in which case this kwarg should still be documented. 

Basically I think we can revert this change and always include `jac_chunk_size` in the docs.

desc/objectives/_fast_ion.py
ddudt: Now that we are supporting forward mode I think we need to keep the `jac_chunk_size` argument.

desc/objectives/_fast_ion.py
```diff
@@ -101,48 +87,54 @@ def __init__(
...
+        alpha=None,
+        num_field_periods=20,
         num_well=None,
         num_quad=32,
         num_pitch=65,
         pitch_batch_size=None,
         surf_batch_size=1,
         nufft_eps=1e-7,
         spline=True,
-        use_bounce1d=False,
```
ddudt, line 120: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
Clarifying question: so now we are not giving the option to use `Bounce1D` because `Bounce2D` is better overall and needed for these speed improvements? That's fine, I just want to make sure I understand.

desc/objectives/_neoclassical.py
ddudt: As stated previously, I think we can keep `jac_chunk_size` now that we are supporting forward mode.

CHANGELOG.md
ddudt: Looks like Dario already made this change in #2291

desc/integrals/_bounce_utils.py
```diff
@@ -106,102 +114,180 @@ def bounce_points(pitch_inv, knots, B, num_well=-1, return_mask=False):
...
     pitch_inv : jnp.ndarray
         Shape broadcasts with (num ρ, num α, num pitch).
-    z1, z2 : tuple[jnp.ndarray]
-        Shape (num ρ, num α, num pitch, num well).
+    z: jnp.ndarray
+        Shape (2, num ρ ?, num α, num pitch, num well).
+        Bounce points.
     mask : jnp.ndarray
-        Shape (num ρ, num α, num pitch, num well).
+        Shape (num ρ ?, num α, num pitch, num well).
```
ddudt, line 147: [question](https://pullpo.io/cc?l=question) 
What does this notation `ρ ?` mean?

desc/integrals/bounce_integral.py
```diff
@@ -74,11 +75,11 @@
 )
 
 
-class Bounce(eqx.Module, ABC):
+class _Bounce(eqx.Module, ABC):
     """Abstract class for bounce integrals."""
 
     @staticmethod
-    def get_pitch_inv_quad(min_B, max_B, num_pitch, simp=True):
+    def pitch_quad(min_B, max_B, num_pitch, **kwargs):
```
ddudt: Reiterating the essence of [this comment](https://github.com/PlasmaControl/DESC/pull/2170/changes#r3352272386): 

This is a change to the public API. Why do we need to change the name of this function? And why was the argument `simp` moved into `*kwargs`? It looks like it is still used and therefore should still be documented.

desc/integrals/bounce_integral.py
```diff
@@ -286,263 +288,92 @@ def __init__(
...
-    def batch(
-        fun,
-        fun_data,
-        desc_data,
-        angle,
-        grid,
-        num_pitch,
-        surf_batch_size=1,
-        simp=True,
-        expand_out=False,
```
ddudt, line 544: Deprecation warnings for `num_pitch`, `simp`, and `expand_out`?

desc/integrals/_bounce_utils.py
```diff
@@ -106,102 +114,180 @@ def bounce_points(pitch_inv, knots, B, num_well=-1, return_mask=False):
...
+        .reshape((-1,) + swapped_shape)  # then shape is (-1, ρ ?, b, α, λ, w)
+        .swapaxes(1, -4)  # recover shape (-1, b, ρ ?, α, λ, w)
+    )
 
 
 @partial(jax.custom_jvp, nondiff_argnums=(2,))
-def regular_points(o, pitch_inv, num_well):
-    """Bounce points then Newton, with regularized jvp.
+def bounce_points(o, pitch_inv, num_well):
+    """Bounce points then iterative solve, with regularized ift jvp.
```
ddudt, line 289: Deprecate name change?

desc/integrals/bounce_integral.py
```diff
@@ -1108,9 +970,6 @@ def interp_to_argmin(
         nufft_eps : float
             Precision requested for interpolation with non-uniform fast Fourier
             transform (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
```
ddudt, line 1001: Similar to [this comment](https://github.com/PlasmaControl/DESC/pull/2170/changes#r3731853289), need to update the docstrings for `nufft_eps` everywhere.

desc/integrals/bounce_integral.py
```diff
@@ -1108,9 +970,6 @@ def interp_to_argmin(
...
-        is_fourier : bool
```
ddudt, line 1111: Deprecate `is_fourier`?

desc/integrals/bounce_integral.py
```diff
@@ -1320,13 +1171,12 @@ def plot_angle_spectrum(
         angle,
         l,
         *,
-        truncate=0,
```
ddudt, line 1323: Deprecate `truncate`?

desc/integrals/bounce_integral.py
```diff
@@ -1542,16 +1390,7 @@ def __init__(
...
     @staticmethod
-    def batch(
-        fun,
-        fun_data,
-        desc_data,
-        grid,
-        num_pitch,
-        surf_batch_size=1,
-        simp=True,
-        expand_out=False,
```
ddudt, line 1553: See [previous comment](https://github.com/PlasmaControl/DESC/pull/2170/changes#r3786996170).

desc/integrals/bounce_integral.py
ddudt: [question](https://pullpo.io/cc?l=question) 
What is the behavior for `num_well=None`? I don't see it in the docstring if that is supposed to be a valid input value.

desc/integrals/bounce_integral.py
ddudt: I agree that we should give this class a more specific name.

desc/integrals/bounce_integral.py
```diff
@@ -244,25 +259,11 @@ class Bounce2D(Bounce):
...
-        If set to ``True``, then it is assumed that ``data["iota"]`` has shape
-        ``(grid.num_rho,)`` or is a scalar.
-    is_fourier : bool
-        If true, then it is assumed that ``data`` holds Fourier transforms
-        as returned by ``Bounce2D.fourier`` and ``data["iota"]`` has shape
-        ``(grid.num_rho,)`` or is a scalar. Default is false.
-    Bref : float
-        Optional. Reference magnetic field strength for normalization.
-    Lref : float
-        Optional. Reference length scale for normalization.
```
ddudt, line 262: Deprecate `is_reshaped`, `is_fourier`, `Bref`, `Lref`?

desc/integrals/bounce_integral.py
```diff
@@ -1444,44 +1297,32 @@ class Bounce1D(Bounce):
...
-    is_reshaped : bool
-        Whether the arrays in ``data`` are already reshaped to the expected form of
-        shape (..., num ζ) or (..., num α, num ζ) or
-        (num ρ, num α, num ζ). This option can be used to iteratively
-        compute bounce integrals one flux surface or one field line at a time,
-        respectively, reducing memory usage.
-        To do so, set to ``True`` and provide only those chunks of the reshaped data.
```
ddudt, line 1482: Deprecate `Bref`, `Lref`, `is_reshaped`?

ddudt (commented):
Looks good overall! Mostly requesting minor changes like additional deprecation warnings, which I will add to #2291 but am documenting here so I remember them. I also have a few clarifying questions for my own understanding.

desc/integrals/bounce_integral.py
unalmis: get_pitch_inv_quad still works just as fine. As described previously, it is shorter. What is the reason we define the class as `Equilibrium` instead of `PossibleEquilibriumWithFiniteForceBalanceError` ?

The public API is not changed at all. Anyone using their code on old versions works just as well on new version. Simp is provided through kwargs intentionally for maintenance purpose.

desc/integrals/_bounce_utils.py
unalmis: This is private function. it means that axis may not be there. Same as (..., num alpha, number pitch , num well ) but more explicit.

desc/objectives/_fast_ion.py
unalmis: Correct

desc/compute/utils.py
unalmis: Im fine with anything

unalmis:
> Looks good overall! Mostly requesting minor changes like additional deprecation warnings, which I will add to #2291 but am documenting here so I remember them. I also have a few clarifying questions for my own understanding.

thanks

CHANGELOG.md
unalmis: thanks

desc/objectives/objective_funs.py
unalmis: desc uses blunt hammer for jac chunking of outermost compute. causes redundant computation. better to chunk intermediate Jacobin to leverage sparsity. done via surf batch size

desc/integrals/_bounce_utils.py
unalmis: Its a private function, so users can't call it.

desc/integrals/bounce_integral.py
unalmis: It is not necessary since it automatically detects it and performs correct action. Deprecate would only pester users with warning to not worry since everything works fine.

desc/integrals/bounce_integral.py
unalmis: its just taken in from kwargs

desc/integrals/bounce_integral.py
unalmis: is reshaped is done automatically so we should not deprecate it. normalization should be done by user.

desc/objectives/objective_funs.py
unalmis: jac chunk size is not used at all. you can use reverse or (forward mode on new Jax versions). this objective will still be computed efficiently.

dpanici (dismissed):

CHANGELOG.md
```diff
@@ -79,7 +87,6 @@ Performance Improvements
     - Now, `desc.compute._build_data_index` uses depth-first search algorithm to construct the dependency tree.
     - Some of the default value computations at import time are removed (i.e. `desc.integrals.bounce_integral.default_quad`)
 - [Significantly improves convergence of inverse stream maps](https://github.com/PlasmaControl/DESC/pull/1919).
-- Check-pointing to bounce integrals to improve speed and reduce memory of reverse mode differentiation.
```
ddudt, line 89: Are these changes to the old log intentional? Usually we don't want to be re-writing this history.

dpanici: [At the time the change was made](https://github.com/PlasmaControl/DESC/pull/2170/changes/346938d2865f1c8c1f142d961824377a243a2705), I think it made sense to remove this since it was before the next release.

Since then, the code this referenced was put into interpax_fft, and also now [I think that checkpointing was made obsolete by the sparse pullbacks](https://github.com/unalmis/interpax_fft/releases/tag/v0.0.8).

I think the important thing is, at the time this change was made, it made sense to remove this line from the changelog to avoid having something like 
- added checkpointing
- removed checkpointing 
in the release notes of the same DESC version (which would not make sense to have).

Now, I think we don't need this removal so we should add it back for the sake of history. @unalmis I think I understood this correctly?

CHANGELOG.md
```diff
@@ -79,7 +87,6 @@ Performance Improvements
     - Now, `desc.compute._build_data_index` uses depth-first search algorithm to construct the dependency tree.
     - Some of the default value computations at import time are removed (i.e. `desc.integrals.bounce_integral.default_quad`)
 - [Significantly improves convergence of inverse stream maps](https://github.com/PlasmaControl/DESC/pull/1919).
```
dpanici, line 97:
```suggestion
- [Significantly improves convergence of inverse stream maps](https://github.com/PlasmaControl/DESC/pull/1919).
- Check-pointing to bounce integrals to improve speed and reduce memory of reverse mode differentiation.
```

CHANGELOG.md
unalmis: Yes that's correct

dpanici (dismissed):

CHANGELOG.md
ddudt: OK thanks. I was referring to all the edits to the Changelog, not just this one line. When I look at the diff I still see two other lines being changed in the sections for the previous release. I will note them in other comments.

CHANGELOG.md
```diff
@@ -64,7 +72,6 @@ New Features
 - Sub-objectives of an `ObjectiveFunction` can now have different `use_jit` values than the `ObjectiveFunction`. These objectives have to be built before building the `ObjectiveFunction`.
 - Adds ``num_neighbors`` parameter to ``CoilSetMinDistance`` that limits the pairwise distance computation to the nearest neighbors per coil, reducing memory useage for large coilsets.
 - Method to plot frequency spectrum of inverse stream map in field line coordinates ``Bounce2D.plot_angle_spectrum``.
-- Method to compute bounce integrals in batches is now added to the public API ``Bounce2D.batch``.
```
ddudt, line 68: See previous comment. I think we don't want to remove this line?

CHANGELOG.md
```diff
@@ -74,6 +81,7 @@ New Features
 Bug Fixes
 
 - Fixes SyntaxError thrown when loading hdf5 data from file-like objects.
+- Fixes ``pitch_batch_size`` argument getting ignored in compute functions.
```
ddudt: See previous comment. Should this line be moved into the latest section at the top of this file (not v0.17.2)?

CHANGELOG.md
```diff
@@ -22,6 +29,7 @@ Bug Fixes
 - Fixed bug that occured when passing in ``_surf_batch_size`` kwarg to ``Omnigenity`` and ``QuasisymmetryBoozer`` objectives
```
dpanici, line 30:
```suggestion
- Fixed bug that occured when passing in ``_surf_batch_size`` kwarg to ``Omnigenity`` and ``QuasisymmetryBoozer`` objectives
- Fixes ``pitch_batch_size`` argument getting ignored in compute functions.
```

CHANGELOG.md
dpanici:
```suggestion
```

CHANGELOG.md
```diff
@@ -64,7 +72,6 @@ New Features
 - Sub-objectives of an `ObjectiveFunction` can now have different `use_jit` values than the `ObjectiveFunction`. These objectives have to be built before building the `ObjectiveFunction`.
 - Adds ``num_neighbors`` parameter to ``CoilSetMinDistance`` that limits the pairwise distance computation to the nearest neighbors per coil, reducing memory useage for large coilsets.
 - Method to plot frequency spectrum of inverse stream map in field line coordinates ``Bounce2D.plot_angle_spectrum``.
```
dpanici, line 76:
```suggestion
- Method to plot frequency spectrum of inverse stream map in field line coordinates ``Bounce2D.plot_angle_spectrum``.
- Method to compute bounce integrals in batches is now added to the public API ``Bounce2D.batch``.
```

dpanici, line 76:
```suggestion
- Method to plot frequency spectrum of inverse stream map in field line coordinates ``Bounce2D.plot_angle_spectrum``.
- Method to compute bounce integrals in batches is now added to the public API ``Bounce2D.batch``.
```

CHANGELOG.md
dpanici: moving up to top level of changelog

CHANGELOG.md
dpanici: Checking v0.17.2 it was a public method still then so adding this back

dpanici (approved):

desc/objectives/_fast_ion.py
```diff
@@ -101,48 +86,69 @@ def __init__(
...
         try:
             import jax_finufft  # noqa: F401
-        except:  # noqa: E722
+        except Exception:
             warnif(
                 nufft_eps >= 1e-14,
                 msg="\njax-finufft is not installed properly.\n"
                 "Setting parameter nufft_eps to zero.\n"
-                "Performance will deteriorate significantly.\n",
+                "Performance may be somewhat slower.\n",
```
YigitElma, line 121: If the performance difference is not that big anymore, I would be happy to just remove `jax-finufft` dependency in the future

requirements.txt
```diff
@@ -1,10 +1,11 @@
 jax >= 0.6.2, != 0.7.1, < 0.10.0
+adv-jax-math >= 1.2, <= 1.2
```
YigitElma, line 2: Do we really need to use the most recent version for these to work properly or you just wanted to keep the min version the most recent?

requirements.txt
```diff
@@ -1,10 +1,11 @@
...
 colorama <= 0.4.6
 diffrax >= 0.6.0, <= 0.7.2
 equinox >=0.11.10, <=0.13.8
 h5py >= 3.0.0, <= 3.16.0
 interpax >= 0.3.3, < 0.4
-interpax_fft >= 0.0.6, <= 0.1.0
+interpax_fft >= 0.1.0, <= 0.1.0
```
YigitElma, line 8: Same as above. Will using an older version break code or just a preference?

YigitElma (approved):
A couple of thoughts for the future regarding maintainability. As discussed in the `equinox` issue, code in `adv-jax-math` requires regular maintenance for sharding changes of `jax` which we don't need here at all. If issues come up down the road, it may make sense to bring widely used functions like `batch_map` back into DESC. Also, it’s generally better to avoid frequent minimum version bumps unless a breaking change strictly requires it. For `interpax_fft` specifically, a version bump won't trigger a warning for users who already have the old version installed in their local environments.

requirements.txt
unalmis: the most recent version has the most relaxed dependency constraints

requirements.txt
unalmis: new versions patch Jax bugs with more relaxed Jax version constraints

unalmis:
> A couple of thoughts for the future regarding maintainability. As discussed in the `equinox` issue, code in `adv-jax-math` requires regular maintenance for sharding changes of `jax` which we don't need here at all. If issues come up down the road, it may make sense to bring widely used functions like `batch_map` back into DESC. Also, it’s generally better to avoid frequent minimum version bumps unless a breaking change strictly requires it. For `interpax_fft` specifically, a version bump won't trigger a warning for users who already have the old version installed in their local environments.

the adv-jax-math package is designed to have sharing completely separate from non-sharing so such issues are impossible. The version bumps relax/widen version support here.

PR #2172, opened 2026-04-19
use ast instead of pytest --collect-only for check_unmarked tests, which is much faster

dpanici:
Admittedly Claude made this for me but I added comments and what it is doing makes sense to me, and I tested it myself for unmarked stuff inside a test class and as a bare test. I didn't like that the old way we had it took like 4 seconds to check the tests every time I tried to make a commit, this one takes only like 0.2s.


f0uriest (approved):

YigitElma (approved):
Looks good. I tested for our usual test conventions, like single tests and tests in a class, and both work. The only limitation is [this case](https://docs.pytest.org/en/stable/example/markers.html#marking-whole-classes-or-modules), but we never use it, and I prefer that we never use it.

Self reminder to update `REQUIRED_MARKS` in #1495

ddudt (approved):

YigitElma:
@dpanici Can we merge this?

YigitElma:
I got permission to merge this from Dario.

PR #2188, opened 2026-04-28
Stateful optimization

### LINKED ISSUE #1034: Optimizers/objectives with auxiliary output

f0uriest:
First pass at #1034. For now this just adds the lowest level API - allowing desc optimizers to maintain and update state between iterations. 

Still to do:
- [ ] Plumb this up to objectives
- [ ] add state to scipy wrappers
- [ ] update proximal to use this?
- [ ] figure out how objectives declare that they need state, defaults, etc.

Some open questions:
- The state that we want to keep track of for the forward pass may be different for the tangent/backward pass. Consider an objective that depends on the solution of a linear system, $A(x) f = b(x)$ where $f$ is the output of the objective (and the state we want to maintain, ie, we solve Af=b iteratively) and $x$ is the input. The derivative would be $A(x) df = db/dx(x) - dA/dx(x) f$ - note that the state we want to keep track of here is $df$, not $f$. This is all to say that keeping track of state may accelerate forward calculations, but may not give much benefit for derivatives depending on the operation. As far as I can tell, it's not possible to have jax return aux information pertaining to the derivative without a ton of custom jvp/vjp logic for each objective.
- I suppose in the above example, always starting with df=0 would be reasonable? since if we're assuming that f doesn't change much then df should be small. So maybe we just always want to take 0 for the tangent state? This may depend on what the state is etc, so may require some custom jvp logic for stateful objectives.

Resolves #1034

### DIFF
--- a/desc/derivatives.py
+++ b/desc/derivatives.py
@@ -26,7 +26,7 @@ class _Derivative(ABC):
     """
 
     @abstractmethod
-    def __init__(self, fun, argnum=0, mode=None, **kwargs):
+    def __init__(self, fun, argnum=0, mode=None, has_aux=False, **kwargs):
         pass
 
     @abstractmethod
@@ -118,6 +118,13 @@ class AutoDiffDerivative(_Derivative):
         ``'hess'`` (Hessian of a scalar function),
         or ``'jvp'`` (Jacobian vector product)
         Default = ``'fwd'``
+    chunk_size : int
+        Will calculate the Jacobian
+        ``chunk_size`` columns at a time, instead of all at once.
+    has_aux : bool
+        Indicates whether fun returns a pair where the first element is considered
+        the output of the mathematical function to be differentiated and the second
+        element is auxiliary data. Default False.
 
     Raises
     ------
@@ -125,11 +132,14 @@ class AutoDiffDerivative(_Derivative):
 
     """
 
-    def __init__(self, fun, argnum=0, mode="fwd", chunk_size=None, **kwargs):
+    def __init__(
+        self, fun, argnum=0, mode="fwd", chunk_size=None, has_aux=False, **kwargs
+    ):
 
         self._fun = fun
         self._argnum = argnum
         self._chunk_size = chunk_size
+        self._has_aux = has_aux
         self._set_mode(mode)
 
     def compute(self, *args, **kwargs):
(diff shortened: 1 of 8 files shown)


unalmis:
> Some open questions:
> - The state that we want to keep track of for the forward pass may be different for the tangent/backward pass. Consider an objective that depends on the solution of a linear system, $A(x) f = b(x)$ where $f$ is the output of the objective (and the state we want to maintain, ie, we solve Af=b iteratively) and $x$ is the input. The derivative would be $A(x) df = db/dx(x) - dA/dx(x) f$ - note that the state we want to keep track of here is $df$, not $f$. This is all to say that keeping track of state may accelerate forward calculations, but may not give much benefit for derivatives depending on the operation. As far as I can tell, it's not possible to have jax return aux information pertaining to the derivative without a ton of custom jvp/vjp logic for each objective.

primal system is A f = b.
 - store the solution f.

The correct tangent system is on wikipedia.

 - store solution tangent df

update state,
- initialize next primal solve at f + df
- initialize next tangent solve at df + (dx old dot dx)/(dx old dot dx old) * (df - df old).

>  note that the state we want to keep track of here is $df$, not $f$.

keep both; see above.

unalmis:
> I suppose in the above example, always starting with df=0 would be reasonable? since if we're assuming that f doesn't change much then df should be small. So maybe we just always want to take 0 for the tangent state? This may depend on what the state is etc, so may require some custom jvp logic for stateful objectives.

The goal is to use known information to estimate the initial guess for the next primal and tangent solves. My suggestion above does that exactly to first order for the primal solve, and uses past history to do it for nearly first order for the tangent solve. You can choose any other estimation heurstic from optimization literature to estimate df_next too, e.g. momentum, mirror etc. In the worst case, the simplest initial guess for the next tangent solve is the previous tangent solve solution, e.g. df_new = df. choosing df_new = 0 doesn't make sense to me.

PR #2194, opened 2026-04-30
correct CGL nodes in `OmnigenousField.change_resolution`

### LINKED ISSUE #2189: OmnigenousField.change_resolution fails
When calling `field.change_resolution(L_B=3, M_B=3, L_x=3)` twice on the same `OmnigenousField` instance, the first call succeeds but the second call raises a `TypeError`:
TypeError: dot_general requires contracting dimensions to have the same shape, got (3,) and (4,).

The OmnigenousField is created by 
```python
field = OmnigenousField(
    L_B=1,
    M_B=3,
    L_x=1,
    M_x=1,
    N_x=1,
    NFP=eq.NFP,
    helicity=(0, eq.NFP),
    B_lm=np.array(
        [
            [0.8, 1.0, 1.2],
            [0, 0, 0],
        ]
    ).flatten(),
)
```

ddudt:
Resolves #2189. 

The Chebyshev-Gauss-Lobatto nodes used for changing the resolution were incorrect, but besides the linked issue this shouldn't have caused any problems besides less accurate interpolation.

### DIFF
--- a/desc/magnetic_fields/_core.py
+++ b/desc/magnetic_fields/_core.py
@@ -2992,15 +2992,15 @@ class OmnigenousField(Optimizable, IOAble):
         self._N_x = setdefault(N_x, self.N_x)
 
         # change well parameters and basis
-        rho = (  # Chebyshev-Gauss-Lobatto nodes
-            1 - np.cos(np.arange(old_L_B // 2, old_L_B + 1, 1) * np.pi / old_L_B)
-        ) / 2
+        # Chebyshev-Gauss-Lobatto nodes for radial interpolation
+        rho = (1 - np.cos(np.arange(old_L_B + 1) * np.pi / old_L_B)) / 2
         nodes = np.array([rho, np.zeros_like(rho), np.zeros_like(rho)]).T
 
         transform_fwd = self.B_basis.evaluate(nodes)
         transform_rev = jnp.linalg.pinv(transform_fwd)
         B_old = transform_fwd @ self.B_lm.reshape((old_L_B + 1, -1))
 
+        # linearly spaced nodes for eta interpolation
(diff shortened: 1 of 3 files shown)


YigitElma (approved):

desc/magnetic_fields/_core.py
```diff
@@ -2992,15 +2992,15 @@ def change_resolution(
         self._N_x = setdefault(N_x, self.N_x)
 
         # change well parameters and basis
-        rho = (  # Chebyshev-Gauss-Lobatto nodes
-            1 - np.cos(np.arange(old_L_B // 2, old_L_B + 1, 1) * np.pi / old_L_B)
-        ) / 2
+        # Chebyshev-Gauss-Lobatto nodes for radial interpolation
+        rho = (1 - np.cos(np.arange(old_L_B + 1) * np.pi / old_L_B)) / 2
```
rahulgaur104, line 2996: This changes both the number and the values of the rho points when change_resolution is called.

rahulgaur104 (approved):

desc/magnetic_fields/_core.py
ddudt: Yes the incorrect number of rho points for some resolutions was the cause of the issue #2189. 

I also changed the values of the rho points to the pattern where the CGL nodes are mapped from [-1, 1] to [0, 1] (consistent with what we do [here](https://github.com/PlasmaControl/DESC/blob/master/desc/basis.py#L1850)). The old values were only using the positive half of the CGL nodes, so they were sparser near the magnetic axis. This change isn't necessary, but I think it should be more accurate.

f0uriest (approved):

PR #2199, opened 2026-05-01
better root find

unalmis:
PR Age

Resolves https://github.com/PlasmaControl/DESC/pull/2157#discussion_r3106446266 .

- formula advertised in numerical recipes not backwards stable. causes issues. resolves that
- better continuity invariant preservation

Otherwise you get fake roots like this where condition number becomes singular, which poisons proceeding algorithm. Resolves that.

Screenshot From 2026-04-29 23-32-15


f0uriest (dismissed):
Looks fine but for ease of review please don't merge into #2168, wait until that is approved and merge both into master 1 by 1

dpanici (dismissed):
once sparsity is in, approved to merge

YigitElma (approved):

dpanici (approved):

PR #2212, opened 2026-05-14
Fix nested profiles and `_static_attrs`

### LINKED ISSUE #2175: Nested `ScaledProfile` and `PowerProfile`
If you scale a `ScaledProfile` or take the power of a `PowerProfile`, we currently nest the underlying profiles, and each time this happens, we add the `scale/power` to the optimizable params. Instead, we can do something like,
```python
class _Profile():
...
    def __mul__(self, x):
        """Multiply this profile by another or a constant."""
        if np.isscalar(x):
            if isinstance(self, ScaledProfile):
                self._scale *= x
                return self
            return ScaledProfile(x, self)
        elif isinstance(x, _Profile):
            return ProductProfile(self, x)
        else:
            raise NotImplementedError()

    def __pow__(self, x):
        """Raise this profile to a power."""
        if np.isscalar(x):
            if isinstance(self, PowerProfile):
                 self._power *= x
                 return self
            return PowerProfile(x, self)
        else:
            raise NotImplementedError()
```

The problems the current behavior might cause that I can think of,
- In an optimization where you are optimizing the profile, having 2 free parameters instead of 1 might cause degeneracy. 
- We make extra copies of the underlying profile, which is redundant
- If you want to write a linear constraint for coefficients of a profile, including the `scale/power` coefficients get a bit annoying.

### LINKED ISSUE #2184: Jax Arrays in `_static_attrs`
I thought the trick we have [here](https://github.com/PlasmaControl/DESC/blob/4b1c43b348e2e6ca8ba750948ffccdb3714f4f7f/desc/io/optimizable_io.py#L89) could also handle `jax.array`s by converting them to a tuple and then back to a `numpy` array. However, currently if you have a `jax` array in static attributes, the jacobian computation hangs indefinitely.

Minimal example,

```python
from desc.objectives.objective_funs import _Objective

class DummyObj(_Objective):
    _static_attrs = _Objective._static_attrs + ["_arr"]
    def __init__(self, eq, arr):
        self._arr = arr
        super().__init__(things=[eq])
        
    def build(self, use_jit=True, verbose=3):
        self._dim_f = 1
        
        return params["Psi"] * jnp.sum(self._arr)
    
eq = get("precise_QA")
# if you use numpy array instead everything works
obj = DummyObj(eq, arr=jnp.arange(3))
obj.build()

# first flattening is fine since there is no cache
f = obj.compute_scaled_error(*obj.xs(eq))
print(f)

# hangs indefinitely
J = obj.jac_scaled_error(*obj.xs(eq))
print(J[0]["Psi"])
```

Thanks @byoungj for realizing this in https://github.com/stellfoundry/ti_optimization/issues/4.

I don't think we should make it work, but we should at least mention it somewhere.

YigitElma:
Resolves #2184 
Resolves #2175

### DIFF
--- a/desc/io/optimizable_io.py
+++ b/desc/io/optimizable_io.py
@@ -10,8 +10,8 @@ import h5py
 import numpy as np
 from termcolor import colored
 
-from desc.backend import register_pytree_node
-from desc.utils import equals
+from desc.backend import jax, register_pytree_node
+from desc.utils import equals, warnif
 
 from .hdf5_io import hdf5Reader, hdf5Writer
 from .pickle_io import PickleReader, PickleWriter
@@ -126,6 +126,16 @@ class _AutoRegisterPytree(type):
 
             for key, val in obj.__dict__.items():
                 if key in static_attrs:
+                    if isinstance(val, jax.Array):
+                        warnif(
+                            True,
+                            UserWarning,
+                            "Detected jax array in _static_attrs. Converting it to "
+                            "numpy array to prevent stalling code.",
+                        )
+                        val = np.asarray(val)
+                        # update the obj attribute in place to prevent future warnings
+                        obj.__dict__.update({key: val})
                     aux_data += [(key, _make_hashable(val))]
                 else:
                     children[key] = val
--- a/desc/io/optimizable_io.py
+++ b/desc/io/optimizable_io.py
@@ -130,8 +130,9 @@ class _AutoRegisterPytree(type):
                         warnif(
                             True,
                             UserWarning,
-                            "Detected jax array in _static_attrs. Converting it to "
-                            "numpy array to prevent stalling code.",
+                            "Detected jax array in _static_attrs. Converting "
+                            f"{key} attribute of the {obj.__class__.__name__} "
+                            "to numpy array to prevent stalling code. ",
                         )
                         val = np.asarray(val)
                         # update the obj attribute in place to prevent future warnings
--- a/tests/test_profiles.py
+++ b/tests/test_profiles.py
@@ -18,7 +18,9 @@ from desc.profiles import (
     FourierZernikeProfile,
     HermiteSplineProfile,
     MTanhProfile,
+    PowerProfile,
     PowerSeriesProfile,
+    ScaledProfile,
     SplineProfile,
     TwoPowerProfile,
 )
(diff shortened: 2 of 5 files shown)


tests/test_profiles.py
```diff
@@ -564,3 +566,25 @@ def test_hermite_spline_solve(self):
...
+        pp = PowerSeriesProfile(
+            modes=np.array([0, 2, 4]), params=np.array([1, -2, 1]), sym=False
+        )
+        pp2 = pp.copy() * 2
+        assert isinstance(pp2, ScaledProfile)
+        assert pp2._scale == 2
+
+        pp2 *= 3
+        assert isinstance(pp2, ScaledProfile)
+        assert pp2._scale == 6
```
ddudt, line 584: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
We should also add the check:
```
assert isinstance(pp2._profile, PowerSeriesProfile)
```

tests/test_profiles.py
```diff
@@ -564,3 +566,25 @@ def test_hermite_spline_solve(self):
...
+
+        pp3 = pp.copy() ** 2
+        assert isinstance(pp3, PowerProfile)
+        assert pp3._power == 2
+
+        pp3 = pp3**3
+        assert isinstance(pp3, PowerProfile)
+        assert pp3._power == 6
```
ddudt, line 594: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
We should also add the check:
```
assert isinstance(pp3._profile, PowerSeriesProfile)
```

ddudt (changes requested):

ddudt (dismissed):

desc/io/optimizable_io.py
```diff
@@ -126,6 +126,16 @@ def _generic_tree_flatten(obj):
 
             for key, val in obj.__dict__.items():
                 if key in static_attrs:
+                    if isinstance(val, jax.Array):
+                        warnif(
+                            True,
+                            UserWarning,
+                            "Detected jax array in _static_attrs. Converting it to "
```
f0uriest: Would be useful to also print the object and name of the static attr so users don't have to go digging

f0uriest (approved):

ddudt (approved):

PR #2213, opened 2026-05-15
non-singular BIE solver, much faster and effecient interpolation for singular integral, and new magnetic field API

### LINKED ISSUE #890: Better FFT interpolation for singular integration

### LINKED ISSUE #1807: New magnetic field API

### LINKED ISSUE #1819: document tangential assumption of current potential classes

unalmis:
PR Age (see #1360)

- [x] All the solvers are thoroughly tested. See the shared paper to explain this PR. Significantly improved adjoint solve from earlier. It seems like this is still the only differentiable 3d laplace solver on the internet that works on multiply connected geometry.
- [x] Tested and confirmed agreement with `BIE3D` package.
- [x] This PR includes my FFT and DFT improvements significantly improve free surface optimization (both old and new) and anything that involves singular integrals. Even the flaky GitHub ci benchmarks [detect this](https://github.com/PlasmaControl/DESC/pull/2213#issuecomment-4460728207).
- [x] Non-invasive, backward compatible, new magnetic field api implemented. Without the new api, the code would be unreadable.
- There are other applications besides free surface. Free surface works too.see optimization in shared paper.
- The linear solve to compute the objective function only takes one gmres cycle. Gradient based optimization is very fast. Using a fast multipole technique would greatly improve performance. There are jax libraries that do this, but that can be a later pull request to merge after this.

## iterative plots
stop tolerance denotes typical stop tol in most applications. gmres/bicgstab is expected to always invert the linear system in a few iterations. after [stateful opt](https://github.com/PlasmaControl/DESC/pull/2188#issuecomment-4467398600), it will be less.

image

image


unalmis:
@YigitElma @dpanici when will this be merged

PR #2216, opened 2026-05-18
`"ni_rr"` compute function

ddudt:
Adds a compute function for $\frac{\partial^2 n_i}{\partial\rho^2}$ that I forgot to add in #2090.


YigitElma (approved):

rahulgaur104 (approved):
Looks good.

Issue #2219, opened 2026-05-19
`weight=0` for `_broadcast_input='Node'` fails


YigitElma:
MWE:
```python
eq = get("precise_QA")
coils = initialize_modular_coils(eq, num_coils=3, r_over_a=2)
# runs fine
obj = ObjectiveFunction(CoilLength(coils, weight=[0, 1, 2]))
obj.build()

# fails
obj = ObjectiveFunction(CoilTorsion(coils, weight=[0, 1, 2]))
obj.build()
```

Currently, the masking of the coil objective by `weight=0` trick doesn't work if the objective returns quantities on each node of the grid. There are multiple things to fix:
- [This line](https://github.com/PlasmaControl/DESC/blob/d41809ac64b94267373747868fc87b05131af494/desc/objectives/_coils.py#L219) fails since fancy indexing of python lists with numpy arrays is not possible
- `obj._coilset_tree["coilset_mask"]` doesn't have proper length to index `grid` which has length `num_coils`

@singh-jaydeep, can you think of any simple fix?

singh-jaydeep:
Jeez, that's my bad. We can keep a separate coil_mask (always of length num_coils) and grid_mask, so the coil_mask can be used here. I'll add this today, as well as some tests to cover this case.

YigitElma:
Resolved in #2221

PR #2221, opened 2026-05-19
Bug Fix: `weight=0` for `_broadcast_input='Node'` fails

singh-jaydeep:
PR #1921 added support for coil optimization with Pytree inputs for `weight`, `target`, and `bounds`. It relies on broadcasting and flattening the input Pytrees to have dimensions consistent with the objective's output shape. When `weight=0` for a given coil or coilset, the corresponding coils (or grid nodes) are dropped during the broadcasting process. However, this wasn't being done correctly for objectives like `CoilCurvature` that output a single value per grid node. 

Fix was suggested in original issue, #2219. We keep track of which coils are masked (`weight=0`) in the attribute `_coilset_tree["coilset_mask"]`. This always has length `num_coils`. The mask `_coilset_tree["objective_mask"]` matches the output shape of the objective, and in the case of per-node objectives, tracks which individual grid nodes are masked. Keeping both masks makes it easier to index things correctly when building the objective.

Also added some tests to catch bugs of this form. 

Addresses Issue #2219.


desc/objectives/_coils.py
```diff
@@ -311,13 +319,15 @@ def weight(self, weight):
...
-        """Expand an array in accordance with the attribute _broadcast_input.
+    def _coilset_broadcast(self, x, target="Coil"):
+        """Broadcast an array to dimensions consistent with "target".
 
         Parameters
         ----------
         x : float or list[float]
             Must be broadcastable to the structure of self._things[0].
+        target: str, optional
+            Optional string taking values "Coil" or "Node". Defaults to "Coil".
```
YigitElma: Need an assert statement since doesn't default to `self._broadcast_input`
```python
assert target in ["Node", "Coil"]
```

tests/test_objective_funs.py
```diff
@@ -4614,31 +4614,64 @@ def test_coil_objective_indices(DummyMixedCoilSet):
...
+    objE.build()
+
 
 @pytest.mark.unit
 def test_coil_objective_setter(DummyMixedCoilSet):
     """Tests setters for _CoilObjectives."""
-    coilset = load(load_from=str(DummyMixedCoilSet["output_path"]), file_format="hdf5")
-    obj = CoilLength(coilset)
+    coilsetA = load(load_from=str(DummyMixedCoilSet["output_path"]), file_format="hdf5")
+    coilsetB = MixedCoilSet((coilsetA[0], coilsetA[1]), check_intersection=False)
```
YigitElma, line 4654: Can you add a comment about the content of these coilsets i.e. MixedCoilSet including a coilset with 1 coil, coilset with 3 coil and so on?

tests/test_objective_funs.py
```diff
@@ -4614,31 +4614,64 @@ def test_coil_objective_indices(DummyMixedCoilSet):
     compB = objB.compute_scaled_error(None)
     np.testing.assert_allclose(compA, compB, atol=1e-13)
 
+    objC = CoilCurvature(coilsetA, weight=weight, normalize=False)
```
YigitElma, line 4629: Can you add a comment about the contents of the coilsets?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Bug Fixes
 - Fixes a bug in `OmnigenousField.change_resolution` when changing `L_B`.
 - Scaling a `ScaledProfile` or taking power of a `PowerProfile` now only updates the `scale`/`power` attributes instead of nesting the `ScaledProfile`/`PowerProfile`s.
 - `jax.Array`s in `_static_attrs` will be automatically converted to `np.ndarray` to prevent stalling code. In general, jax arrays should be omitted in `_static_attrs`.
+- Fixes a bug in `_CoilObjective` for objectives which are computed per-grid node when at least one entry of `weight` is zero.
```
YigitElma: I think there is also a breaking change here. If someone had a custom coil objective, they need to update the `[self._coilset_tree["coil_mask"]` to `[self._coilset_tree["objective_mask"]`. Can you also document that?

desc/objectives/_coils.py
```diff
@@ -59,7 +59,7 @@ class _CoilObjective(_Objective):
     "Coil" if the objective returns a single scalar per coil, and "Node"
     if it returns a scalar at every grid point. To be compatible with
     masking, compute function should apply the mask
-    self._coilset_tree["coilset_mask"] before returning data.
+    self._coilset_tree["objective_mask"] before returning data.
```
YigitElma, line 62: This is a breaking change, and needs to be documented in the changelog. But I like the new naming, which is easier to distinguish.

desc/objectives/_coils.py
```diff
@@ -133,7 +133,9 @@ def _build_coilset_tree():
                 params_tree["coils"] contains a nested list of 0s representing
                 individual coils and the coilsets to which they belong. Similarly,
                 params_tree["nodes"] lists the grid nodes associated with each coil.
-                params_tree["coilset_mask"] contains the indices in [0,self._dim_f-1]
+                params_tree["coil_mask"] contains the indices in [0,self._num_coils]
```
YigitElma: Should we keep the dict key name the same as before `coilset_mask`? The meaning is different but at least prevents breaking some code.

tests/test_objective_funs.py
```diff
@@ -4614,31 +4614,64 @@ def test_coil_objective_indices(DummyMixedCoilSet):
...
+    objD = CoilCurvature(coilsetB, normalize=False)
+    objC.build()
+    objD.build()
+    compC = objC.compute_scaled_error(None)
+    compD = objD.compute_scaled_error(None)
```
YigitElma, line 4634: Can you add tests regarding `dim_f` too?

YigitElma (commented):
I will test it myself too, but it looks good. I am personally fine with the breaking change; this was a pretty new addition, and going forward, I like the new name better.

CHANGELOG.md
YigitElma: Also, during merging master to previous PR, the changelog order got slightly messed up. We added #1921 in version 0.17.0, but it is under 0.16.0 in the file. Because of this, it was previously not in any changelog (we manually copy/paste the new changelog), now I updated the release notes, but it would still be nice to fix the changelog for future reference. Can you move these 3 lines to the end of v0.17.0 `New Features`?
```
- Adds support for optimization targeting individual coils in a coilset.
  - Coil objectives accept pytree inputs for `target`, `bounds`, and `weight`.
  - Able to set weights to zero, excluding certain coils from the objective.
```

desc/objectives/_coils.py
singh-jaydeep: Added!

tests/test_objective_funs.py
singh-jaydeep: Added

tests/test_objective_funs.py
singh-jaydeep: Added

CHANGELOG.md
singh-jaydeep: Updated. If the wording isn't clear, can change.

desc/objectives/_coils.py
singh-jaydeep: Sounds good, documented.

desc/objectives/_coils.py
singh-jaydeep: I switched back to coilset_mask.

tests/test_objective_funs.py
singh-jaydeep: Added

CHANGELOG.md
singh-jaydeep: Moved!

desc/objectives/_coils.py
```diff
@@ -166,12 +168,16 @@ def expand(t, idx=0):
...
                 "nodes": tree[1],
-                "coilset_mask": slice(None),
+                "coilset_mask": np.arange(self._num_coils),
+                "objective_mask": slice(None),
             }
             if np.any([w == 0 for w in tree_leaves(self._weight)]):
-                mask = self._coilset_broadcast(self._weight)
-                mask = np.nonzero(mask)[0]
-                self._coilset_tree["coilset_mask"] = mask
+                coilset_mask = self._coilset_broadcast(self._weight)
```
YigitElma: Nitpick: can you pass `target="Coil"` explicitly?

CHANGELOG.md
YigitElma: We made a new release recently. Can you move this above and add a title
```
0.17.2
------
```
below it

YigitElma (commented):
I will review it properly again later. Just wanted to mention the changelog issue

dpanici (dismissed):

desc/objectives/_coils.py
```diff
@@ -133,7 +133,9 @@ def _build_coilset_tree():
                 params_tree["coils"] contains a nested list of 0s representing
```
YigitElma: Actually, this docstring is a bit outdated? This function doesn't return `params_tree`, it sets `self._coilset_tree`. Tbh, not important since it is easy to understand but just pointing out.

YigitElma (dismissed):
Looks good to me. I also tested it with my own objective and it works.

desc/objectives/_coils.py
singh-jaydeep: Good catch. I changed it, leaving in the explanation of what each dictionary entry is. If it is too verbose let me know, I can remove it.

YigitElma (dismissed):

dpanici (dismissed):

desc/objectives/_coils.py
```diff
@@ -311,30 +319,34 @@ def weight(self, weight):
...
+        target: str, optional
+            Optional string taking values "Coil" or "Node". Defaults to "Coil".
 
         Returns
         -------
         arr: float or list[float]
             Float inputs are returned unchanged, and list inputs are
             expanded to size self._dim_f.
         """
+        assert target in ["Node", "Coil"]
```
ddudt: [nitpick(if-minor)](https://pullpo.io/cc?l=nitpick&d=if-minor) 
I'd prefer these to be lowercase, or check for a match without being case sensitive.

CHANGELOG.md
```diff
@@ -8,9 +8,15 @@ Performance Improvements
...
 - Fixes bug that was always setting NFP=1 in ``to_FourierRZ`` methods.
+- Fixes a bug in `_CoilObjective` for objectives which are computed per-grid node when at least one entry of `weight` is zero.
 - Fixes ``VMECIO.save`` metadata for current-density variables and corrects the
   asymmetric ``currvmns`` magnetic-axis extrapolation.
 
+Breaking Changes
+
+- Name change in `_CoilObjective` replacing `coilset_mask` with `objective_mask`. Custom
+  subclasses with `_broadcast_input="Node"` that previously used `coilset_mask` should
+  switch to `objective_mask`.
```
ddudt: [nitpick(if-minor)](https://pullpo.io/cc?l=nitpick&d=if-minor) 
You can have this all on a single line, we don't enforce the max line size for the Changelog.

ddudt (changes requested):

desc/objectives/_coils.py
singh-jaydeep: I switched to lowercase.

CHANGELOG.md
singh-jaydeep: Sounds good, fixed

desc/objectives/_coils.py
YigitElma: I am going to be annoying, but this is another breaking change. Instead can we do something like:
```python
assert target.lower() in ["node", "coil"]
```
The other change is breaking for almost every custom coil objective, I think

singh-jaydeep: makes sense, I'll do this instead

singh-jaydeep:
Main thing was addressing the backwards compatibility problem with changing from "Coil"/"Node" to "coil"/"node". Sticking with lowercase, but added .lower() where checks happen, so uppercase is fine. The docstrings state it is case insensitive, but lists defaults with lowercase.

Sorry for the additional changes, but ran the PR through claude and it suggested ways to make the code more robust. I implemented some of them.
- Because we have to overwrite things like self.weight, self.target, and self.bounds for broadcasting purposes, building objectives was not idempotent. Also building after setting new weights was problematic. So we now save the unbroadcast versions in self.weight_input, self.target_input, self.bounds_input, and self.grid_input. These do not change except via a setter, so you can build correctly multiple times.
- in _coilset_broadcast, the check involving len(tree_leaves) was different depending on whether the user gave the input array as a list or an np/jnp array. Made this uniform. The one thing a user cannot input (we could support it, but it would require an additional flattener) is something like weight = [1, 2, np.array([2,3])], i.e. lists which contain np/jnp arrays.
- extended some of the tests to include rebuilding objectives

tests/test_objective_funs.py
```diff
@@ -4596,13 +4602,14 @@ def test_coil_objective_input(DummyMixedCoilSet):
     assert np.size(obj._bounds[0]) == 1 and np.size(obj._bounds[1]) == 1
     assert np.size(obj._weight) == 1
 
+    weight = [0.0, 2.0, 3.0, 4.0]
     obj = CoilCurvature(coilset, bounds=bounds, weight=weight)
     obj.build()
```
ddudt, line 4608: Are these lines necessary? It doesn't look like they are used to test anything. Unless you are checking that `obj.build()` runs without errors?

ddudt (dismissed):
See question in comment, otherwise looks good.

YigitElma (dismissed):

tests/test_objective_funs.py
YigitElma: I think this is just checking that the `node` case build works.

desc/objectives/_coils.py
```diff
@@ -166,12 +166,16 @@ def expand(t, idx=0):
             self._coilset_tree = {
                 "coils": tree[0],
                 "nodes": tree[1],
-                "coilset_mask": slice(None),
+                "coilset_mask": np.arange(self._num_coils),
+                "objective_mask": slice(None),
             }
             if np.any([w == 0 for w in tree_leaves(self._weight)]):
```
YigitElma: maybe use `np.isclose(w, 0.0, atol=1e-12)` to support 0.0 for the `weight`

YigitElma (approved):
We can merge once the CI completes. This is discussed during issue triage today.

desc/objectives/_coils.py
YigitElma: Btw, I think this was not really necessary. But still not a problem to have
```
>>> a = 0.0
>>> b = 0
>>> a == b
True
>>> import numpy as np
>>> a = np.array([0.0])
>>> b = np.array([0])
>>> a == b
array([ True])
>>> a = np.array([0.0, 1.0, 2.0, 3])
>>> b = np.array([0, 1, 2, 3])
>>> a == b
array([ True,  True,  True,  True])
```

desc/objectives/_coils.py
```diff
@@ -169,13 +171,17 @@ def expand(t, idx=0):
...
+                np.isclose([w for w in tree_leaves(self._weight)], 0.0, atol=tol)
+            ):
                 coilset_mask = self._coilset_broadcast(self._weight, target="coil")
                 objective_mask = self._coilset_broadcast(
                     self._weight, self._broadcast_input
                 )
-                self._coilset_tree["coilset_mask"] = np.nonzero(coilset_mask)[0]
-                self._coilset_tree["objective_mask"] = np.nonzero(objective_mask)[0]
+                self._coilset_tree["coilset_mask"] = np.nonzero(coilset_mask > tol)[0]
+                self._coilset_tree["objective_mask"] = np.nonzero(objective_mask > tol)[
```
ddudt, line 182: Should this be `np.nonzero(np.abs(coilset_mask) > tol)`?

singh-jaydeep: I don't think the np.abs is necessary -- both coilset_mask and objective_mask consist of broadcasted values from self._weight, and those are verified to be non-negative.

ddudt (approved):

PR #2226, opened 2026-05-27
Removing objectives

### LINKED ISSUE #2225: Move bounce integrals, turbulent transport, fast ion objectives to external package
Rory and I had discussed that if my open PRs which update my work will not be added to DESC, then bounce integrals should be removed as we don't want a poor/inefficient version made available on the most public platform when the published work https://arxiv.org/abs/2412.01724v4 under my name makes claims about the efficient version. That would lead to users not using my work; having a bad impression of it and reduce its impact.

These include: objectives for neoclassical transport, fast ion confinement, turbulent transport, max J etc.,.
(e.g. epsilon effective, GammaC, GammaAlpha, GammaDelta, available energy, turbITG, max J, J parallel, direct omni) and associated algorithms such as sparsity preserving pullbacks, bounce integrals, etc. They can be moved to an external package so that other stellarator optimization libraries can use them.

The efficient and more accurate versions of all these objectives/algorithms is available on the fork 
https://github.com/unalmis/DESC.

unalmis:


unalmis:
Closing in favor of #2242

PR #2228, opened 2026-06-01
Fix the error in `field_line_integrate` with `bs_chunk_size`

### LINKED ISSUE #2214: `field_line_integrate` bug when specifying a `bs_chunk_size`
The function `field_line_integrate` is supposed to be able to take an integer for `bs_chunk_size`, but it throws an error: 
```
R, Z = field_line_integrate(
    R0,
    Z0,
    phi,
    field,
    chunk_size=None,
    bs_chunk_size=512,  # default None works but not an int
)
```
Error message:
```
ValueError: Non-hashable static arguments are not supported. An error occurred while trying to hash an object of type <class 'jax._src.interpreters.partial_eval.DynamicJaxprTracer'>, JitTracer(~int64[]). The error was:
TypeError: unhashable type: 'DynamicJaxprTracer'
```

YigitElma:
Resolves #2214 

`_odefun` is already jitted by the `filter_jit` of `diffeqsolve`, and additional jit shouldn't give performance improvement. However, it causes a weird behavior for static values. For example, `filter_jit` treats all the arguments that are not `jax.array`s or without `flatten` method as static, so `bs_chunk_size` becomes static (without us having to declare it as static as usual). For some reason, having the internal jit causes `bs_chunk_size` to become `JaxTracer` for tracing purposes and throws the error.

### DIFF
--- a/desc/magnetic_fields/_core.py
+++ b/desc/magnetic_fields/_core.py
@@ -19,7 +19,7 @@ from interpax import approx_df, interp1d, interp2d, interp3d
 from netCDF4 import Dataset, chartostring, stringtochar
 from scipy.constants import mu_0
 
-from desc.backend import jit, jnp, sign
+from desc.backend import jnp, sign
 from desc.basis import (
     ChebyshevDoubleFourierBasis,
     ChebyshevPolynomial,
@@ -2816,8 +2816,8 @@ def _intfun_wrapper(
     )
 
 
-@jit
 def _odefun(s, rpz, args):
+    # Note: this function is already jitted by diffeqsolve
     field, params, scale, source_grid, bs_chunk_size = args
     r = rpz[0]
     br, bp, bz = (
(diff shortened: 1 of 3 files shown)


desc/magnetic_fields/_core.py
```diff
@@ -2816,8 +2816,8 @@ def _intfun_wrapper(
     )
 
 
-@jit
```
rahulgaur104, line 2819: So the problem was this jit flag?

rahulgaur104 (approved):

desc/magnetic_fields/_core.py
YigitElma: Yeah, with this jit, we have to declare `bs_chunk_size` as static (relatively harder change). But `filter_jit` of `diffeqsolve` already does that automatically, so this is kind of redundant.

ddudt (approved):

desc/magnetic_fields/_core.py
f0uriest: Same issue: https://github.com/jax-ml/jax/issues/35090

f0uriest (approved):

PR #2234, opened 2026-06-03
Workaround for dependabot issue

YigitElma:
Hopefully, this will solve the issue


ddudt (approved):

rahulgaur104 (approved):
There's got to be an easier way than trial and error.

YigitElma:
> There's got to be an easier way than trial and error.

In this thread https://github.com/dependabot/dependabot-core/issues/14798, some people said they used this as a work around. I am not sure if this is the same as the old behavior but it is at least better than what is going on.

YigitElma:
Looks like it worked! #2235 #2236 #2237

PR #2239, opened 2026-06-09
`jit` compile tangent computation in Proximal

YigitElma:
- Removes the `batched_vectorize` call from Proximal, instead conducting the vectorized operation manually
- New function is fully jittable, hence prevents the repeated SVD calls
- Updates the clunky pad/clip/split operations inside Proximal
- Instead of computing JVP of `ForceBalance` for each tangent, only computes for a maximum of `dimc_per_thing[eq_idx]` tangents. For example, lets say you have 300 boundary coefficients and 100 coil coefficients inside a free boundary solve, master computes 400 JVPs, this PR computes 300. 

Note: Removing the `batched_vectorize` will help to make #1495 work with parallel force balance constraints too.

The document I shared in https://github.com/PlasmaControl/DESC/pull/1623#issuecomment-2877551214 might help to review the changes in this PR. One change from that document is that in this PR, I removed the `prox._feasible_tangents` and instead multiplied the related part by `eq_feasible_tangents`, these are equivalent because D@Z is exactly `eq_feasible_tangents`.


desc/optimize/_constraint_wrappers.py
```diff
@@ -1334,14 +1303,89 @@ def wrapper(*args, **kwargs):
...
+    )
+    tangents = dxdcv - dfdc @ feasible_tangents.T
+    return tangents
+
+
 @jit_if_possible
+def _get_tangent(
+    constraint, v, xf, constants, eq_feasible_tangents, dxdc, op="scaled_error"
+):
+    # Note: This function is vectorized over v. So, v is expected to be 1D array
```
YigitElma: This is not vectorized anymore. Update comment

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Changelog
 
 Performance Improvements
 
+- More efficient `ProximalProjection` jacobians especially if the `ForceBalance` constraint uses a small `jac_chunk_size`.
```
YigitElma: Performance might be only improved for chunked case.

desc/optimize/_constraint_wrappers.py
```diff
@@ -1334,14 +1303,89 @@ def wrapper(*args, **kwargs):
...
     # of same size as full equilibrium state vector. This function returns a 1D array.
 
-    # here we are forming (dF/dx)^-1 @ dF/dc
-    # where Fxh is dF/dx and Fc is dF/dc
+    # here we are forming (dF/dx)⁻¹ @ dF/dc
+    # where Fxh is dF/dxᵀ and Fc is dF/dc.
+    # Note: Fxh and its SVD do not depend on dc (the vectorized argument). Since the
+    # whole tangent computation is jitted as one program, we rely on the compiler to
+    # hoist this loop-invariant SVD out of the batched scan/vmap rather than
+    # recomputing it for every tangent.
```
YigitElma: Previously, I was trying to be smart by removing the transpose from the Fxh and handle that at the return operation (transpose of u and v etc) but that cause slowdown on CPU. Maybe SVD lowering of JAX on CPU is dependent on tall/wide?!

desc/optimize/_constraint_wrappers.py
```diff
@@ -1334,27 +1280,79 @@ def wrapper(*args, **kwargs):
...
+    # J @ [(tangent vectors in c direction) - (tangent vectors in x direction)@dfdc]
+    # Note: We will never form full Jacobian J, we will just compute the above
+    # expression by JVPs.
+
+    # v contains prox._args DoFs from eq and other objects (like coils, surfaces
+    # etc). Only the eq block changes when the equilibrium is re-solved.
+    vs = jnp.split(v, np.cumsum(dimc_per_thing)[:-1], axis=-1)
+    # JVPs are taken for the dxdc @ v tangents, but at most dimc of them are useful,
+    # so take whichever combination needs fewer of them. Applying v after the JVPs
+    # only pays off with a lot of coil, surface etc DoFs, ie. single stage.
```
dpanici, line 1313: I don't think I understand the difference bwtn these two paths

dpanici (dismissed):

desc/optimize/_constraint_wrappers.py
YigitElma: - `dim_opt` = number of free optimization variables (Rb+Zb+profiles+coils etc after linear constraints)
- `prox.dim_x` = total number of optimization variables (Rb+Zb+profiles+coils etc)
- `dim_c` = total number of eq optimization variables (Rb+Zb+profiles)
- `dim_xeq` = size of full eq state vector (Rlmn+Zlmn+Llmn+Rb+Zb etc)

`_proximal_eq_tangents` calls the JVP of the constraint with `eq_feasible_tangents` and tangents for the optimization variables. `vs` is a `dim_opt` by `prox.dim_x` sized array, when we take `vs[eq_idx]`, it becomes `dim_opt` by `dim_c`. So, the indexing operation doesn't change the number of tangents but makes the tangent vectors shorter.

We know that JVP wrt non-equilibrium optimization variables are 0, so if `dim_opt` > `dim_c`, we are going to have some trivially 0 JVPs. We can either take JVP wrt to `vs[eq_idx] @ dxdc.T` which is `dim_opt` by `dim_xeq` or wrt `dxdc.T` which is `dim_c` by `dim_xeq`. These branches compute the lower one with a static check.

YigitElma: I originilly had a comment like this
```
    # dxdc maps the eq DoFs (Rb_lmn, Zb_lmn, profiles) to the full eq state vector.
    # Everything in _proximal_eq_tangents is linear in the directions it gets, so we
    # can either take the JVPs for the columns of dxdc and apply v afterwards with a
    # matmul (first branch), or apply v first and take the JVPs only for the
    # directions in v (second branch). The JVPs dominate the cost, so we use the one
    # with fewer directions. The first branch never needs more than
    # dimc_per_thing[eq_idx] of them, since that is all the eq DoFs there are, so it
    # only pays off when v has more directions than that. Usually v holds the feasible
    # tangents of the outer LinearConstraintProjection and we optimize only some of
    # the boundary modes, so the second branch is the usual one and the first only
    # kicks in with a lot of coil, surface etc DoFs, important for single stage.
```
But it was very long, I can keep it if this is not clear.

ddudt (dismissed):
[question](https://pullpo.io/cc?l=question) 
Do these changes mean that it is faster to use a relatively small `jac_chunk_size` with `ForceBalance` as a nonlinear constraint in single-stage optimizations? Or is it still best to use the default (no chunking) for that objective if it can fit in memory?

YigitElma:
> [question](https://pullpo.io/cc?l=question) Do these changes mean that it is faster to use a relatively small `jac_chunk_size` with `ForceBalance` as a nonlinear constraint in single-stage optimizations? Or is it still best to use the default (no chunking) for that objective if it can fit in memory?

If it can fit in the memory, no chunking should still be faster. These changes basically remove the overhead of unjitted `batched_vectorize` call, i.e. multiple SVD's. Now, the `ProximalProjection` constraint chunk size act like any other objective's chunk size.

dpanici (dismissed):

desc/optimize/_constraint_wrappers.py
```diff
@@ -752,27 +745,15 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
...
-        # Since here we are setting the feasible direction for eq parameters only,
-        # we need to add 0 rows for eq fixed parameters and non-eq parameters which we
-        # handle by below operation
-        self._feasible_tangents[self._eq_idx] = self._feasible_tangents[self._eq_idx][
-            :, self._eq_unfixed_idx
-        ] @ (self._eq_Z * self._eq_D[self._eq_unfixed_idx, None])
-        self._feasible_tangents = jnp.concatenate(
-            [np.atleast_2d(foo) for foo in self._feasible_tangents], axis=-1
-        )
+        # Note: the full state vector version of the feasible tangents used to be
```
f0uriest: I'd rather not include comments that reference what's no longer here.

f0uriest (dismissed):

desc/optimize/_constraint_wrappers.py
YigitElma: I see, I just added this as a general explanation, but I can remove

desc/optimize/_constraint_wrappers.py
```diff
@@ -752,27 +745,15 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
...
+        # built here, but its eq block is just _eq_solve_objective._feasible_tangents
+        # and the other blocks are identity, so _proximal_get_tangents uses that.
```
```suggestion
```

dpanici (approved):

ddudt (approved):

Issue #2171, opened 2026-04-18
Single contangent pullback through compute pipeline


unalmis:
Should pullback a single contangent through the entire desc compute infrastructure to extend factor N memory and performance improvement, where N is number of rows in Jacobian, to full pipeline. 

All you have to do is put the sparse pullback decorator over that single function. Few lines changed . For meshgrids this is easy.

This would require the inputs and outputs the to the registered compute functions to be of shape (rho, theta, zeta, ...) . That's actually the natural shape that `np.meshgrid(rho,theta,zeta, ij)` returns. Then wrap the full compute pipeline in a single sparse pullback decorator except for Zernike transform.  Edit: see below for simpler 

Before exiting to return data to users use tree_map(meshgrid_reshape, data) so that retuned data to users is same structure as before.

unalmis:
It would also work if you do the tree map in the reverse direction , then apply sparse pullback, then tree map back. Less invasive

unalmis:
Also better because having a single call to sparse pullback in one code spot makes it easier to plumb through a flag on whether to use it or not based on if jacrev or jacfwd was selected. For special functions where jacfwd is useless we want to nest sparse pullback anyway so don't let the goal in this issue slow down the merge of #2170

PR #2242, opened 2026-06-15
update comments

unalmis:
Adds some comments to objectives

### DIFF
--- a/desc/compute/_fast_ion.py
+++ b/desc/compute/_fast_ion.py
@@ -88,6 +88,11 @@ def _drift2(data, B, pitch):
 def _Gamma_c(params, transforms, profiles, data, **kwargs):
     """Fast ion confinement proxy as defined by Nemov et al.
 
+    Notes
+    -----
+    A much more performant version is available at https://github.com/unalmis/DESC.
+    The reference 2 below refers to that implementation.
+
     [1] Poloidal motion of trapped particle orbits in real-space coordinates.
         V. V. Nemov, S. V. Kasilov, W. Kernbichler, G. O. Leitold.
         Phys. Plasmas 1 May 2008; 15 (5): 052501.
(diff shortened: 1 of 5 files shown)


YigitElma (approved):

rahulgaur104 (approved):
Full optimization performance benchmarks were never performed. For example, when optimizing an equilibrium, the new bounce integral from ku/bounce (merged) took almost exactly the same time as the old bounce integration routines. So compute maybe faster but optimization definitely was not. Maybe sparse pullback will change that, maybe not.

I don't actually care at this point, so approving this PR.

unalmis:
> Full optimization performance benchmarks were never performed. For example, when optimizing an equilibrium, the new bounce integral from ku/bounce (merged) took almost exactly the same time as the old bounce integration routines. So compute maybe faster but optimization definitely was not. Maybe sparse pullback will change that, maybe not.
> I don't actually care at this point, so approving this PR.

@rahulgaur104 's comment purpose is not clear to me, nor why he feels the need to advertise he does not care.

Focusing on the technical/work relevant content of that comment, his statements in reference to my work are verifiably false. 

To reduce spread of misinformation for others who are reading, note that
- `ku/bounce` (link: #1919 ) is a branch that primarily focused on correcting bugs in DESC and improving convergence; not performance.
- Still, the performance benefit from that PR is significant if one benchmarks it properly - e.g. compare how much faster it is when a low resolution solver is used (because the convergence improvements make the new low resolution solver as accurate as the old high resolution solver).
- All of this is shown in detail in #1919 .

Next, ku/sparse_pullback (link: #2170 ) and the following PRs reduces memory and improves performance by a factor of the number of flux surfaces the optimization is done with.
- This was benchmarked and proven (in math and in practice). One should not diminish the value of my work there, which is actually the largest performance boost in terms of memory+compute in desc to date. One performance benchmark demonstrating this (among many others that were shared with visualization) is available here (https://github.com/unalmis/DESC/blob/master/publications/unalmis2025/effective_ripple_profile.py).
- My later PRs (#2220 and #2171) continue this improvement by extending the sparse pullback through more of the compute pipeline.

I suggest making effort to become better informed before making negative claims on others' work.

PR #2244, opened 2026-06-17
Use qr_multiply to allow better reuse of factorisation

jpbrodrick89:
@YigitElma re your jax [discussion](https://github.com/jax-ml/jax/discussions/23149#discussioncomment-17321452) this is not a perfectly optimal solution to your problem but should be a nice universal 2-4x improvement (potentially more if your Jacobians are very tall). There are two flavours of improvement and they stack:

1. Replace `qr(mode="economic") + Q.T @ r` with `qr_multiply` which avoids materialising Q, this provides ~1.5x speedup in each QR solve based on rough benchmarks.
2. Reuse QR factorisation of undamped Jacobian, this reduces the computational cost of each alpha-solve by a factor (m+n)/2n (where height m > n) at the cost of one additional extra initial solve

$$
\begin{pmatrix}
J \\
\sqrt{a} I
\end{pmatrix}∆x =
\begin{pmatrix}
Q_0 \begin{pmatrix}
R_0 \\ 
0 \end{pmatrix}\\
\sqrt{a} I
\end{pmatrix}
$$

$$
\implies
\begin{pmatrix}
Q_0^{(\text{thin})T} & 0 \\
0 & I 
\end{pmatrix} 
\begin{pmatrix}
J \\
\sqrt{a} I
\end{pmatrix}=
\begin{pmatrix}
R_0 \\
\sqrt{a} I
\end{pmatrix} = 
Q_1 \begin{pmatrix}
R_1 \\
0
\end{pmatrix}
$$

So all one needs to do is a QR of the 2n x n matrix formed by vertically stacking $R_0$ and $\sqrt{alpha}I$. Alternatively, you could view this as Cholesky update and either scan over [jax.lax.linalg.cholesky_update](https://docs.jax.dev/en/latest/_autosummary/jax.lax.linalg.cholesky_update.html) (stable but likely as slow as your homebrewed version) or form the Gram matrix $R_0^T R_0 + \alpha I$ which is probably the fastest option but as always will struggle with ill-conditioning. What you ideally want is a wrapper for something like [Lapack](https://www.netlib.org/lapack//explore-html/d4/dad/group__tpqrt_gadb01a522e74a81897506c27a3508e35c.html#gadb01a522e74a81897506c27a3508e35c)'s `dtpqrt` which is specifically designed for factorising a stacked triangular plus trapezoidal matrix but I'm not sure if there's a GPU equivalent.

Note I implemented the `qr_multiply` in https://github.com/jax-ml/jax/pull/35104 so do shout if there are any issues. :warning: `qr_multiply` is not currently differentiable but it will be when https://github.com/jax-ml/jax/pull/36357 is eventually approved. If you are interested in running on GPU also note https://github.com/jax-ml/jax/pull/36575 which can provide a significant performance improvement when using `qr_multiply` (the default cusolver `ormqr` loses against `orgqr` + `gemm` for smaller Jacobians, this pure jax optimisation is faster pretty much across the board).

I did not adapt the wide cases (despite them also benefitting by a similar amount) because the API for the jax [ormqr](https://docs.jax.dev/en/latest/_autosummary/jax.lax.linalg.ormqr.html) does not currently match the [scipy one](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lapack.dormqr.html#scipy.linalg.lapack.dormqr) (because jax doesn't typically wrap `scipy.linalg.lapack`). If you decide you want to write an adapter it shouldn't be too painful.


jpbrodrick89:
Sorry this will require a minimum python version bump to 3.11 for jax>=0.10.0 compatibility (unless we write an alternative qr_multiply), as EOL is Oct 2026 hopefully that's acceptable, can push the changes if you're comfortable with that.

YigitElma:
Hi @jpbrodrick89, thank you for taking the time to explain and implement it! I will try to locally benchmark this for a couple of different problem sizes on the GPU and CPU. I think this can solve the issue we faced in https://github.com/PlasmaControl/DESC/pull/1755. Differentiability is not currently a problem; we are not differentiating over optimizations yet.

Although this new function is more efficient, I don't think we would like to bump our min `jax` version to `0.10.0` yet. We will have to first solve a couple of issues with JAX 0.10+ versions in #2235. After that, I think we can have a wrapper in `desc.backend` for this, such that if the user has a new version of jax, it can use this method; otherwise, older manual QR. We were planning to bump the Python version soon, so there shouldn't be a problem with that.

Having said that, I think the improvement in `trust_region_step_exact_qr` besides `qr_update` can still give a nice performance improvement.

desc/optimize/least_squares.py
```diff
@@ -302,15 +302,13 @@ def lsqtr(  # noqa: C901
             # try full newton step
             tall = J_a.shape[0] >= J_a.shape[1]
             if tall:
-                Q, R = qr(J_a, mode="economic")
-                p_newton = solve_triangular_regularized(R, -Q.T @ f_a)
+                z, R = qr_multiply(J_a, f_a, mode="right")
```
YigitElma: We can think of passing `z, R` to `trust_region_step_exact_qr`, which can save us 1 more QR. Since these functions are not really public, I think we can update the function signature. Just wide/tall distinction can be annoying.

desc/backend.py
```diff
@@ -86,7 +86,14 @@ def _diag_to_full(d, e):
...
     from jax.scipy.fft import dct, dctn, idct, idctn
-    from jax.scipy.linalg import block_diag, cho_factor, cho_solve, qr, solve_triangular
+    from jax.scipy.linalg import (
+        block_diag,
+        cho_factor,
+        cho_solve,
+        qr,
+        qr_multiply,
+        solve_triangular,
+    )
```
YigitElma: Maybe we can keep our range the same by having
```python
if Version(jax.__version__) >= Version("0.10.0"):
    from jax.scipy.linalg import qr_update
else:
    @functool.partial(jit, static_argnames="mode")
    def qr_update(A, b, mode="right"):
        Q, R = qr(A, mode="economic")
        if mode == "right":
            return Q.T@b, R
        elif mode == "left":
            return b.T@Q
        else:
            raise NotImplementedError()
```
The code above can be wrong, but the idea is to have something like this.

jpbrodrick89:
Great feedback thanks @YigitElma , should be fully backwards compatible now but the full Q materialisation free speedup will only be unlocked when you upgrade to Jax 0.10.0. Let me know if anything else is needed.

YigitElma:
@jpbrodrick89 Thanks! This looks good to me. #2245 will solve the issue with CI.

We generally require 2 approvals from active developers. I will discuss this with others during our dev-meeting tomorrow.

YigitElma:
Hi @jpbrodrick89, there was a minor bug with using the variable name `z` again in the Augmented Lagrangian least-squares optimizer. We were also using `z` for the augmented state vector. I renamed the variables `z -> Qt_f` like convention for both optimizers.

I also reordered the changelog because we had a recent release.

YigitElma:
I can run the benchmark on GPU later, but the CPU memory benchmark on CI is as follows:
image
I tried to label which peak corresponds to what.

jpbrodrick89:
I just realised that I could use my pure jax implementation of ormqr in https://github.com/jax-ml/jax/pull/36575 to provide a more efficient "fallback" (that even beats cusolver) as it doesn't depend on a rebuild of jaxlib so can probably run on any version of jax. Its not too complex and fairly self contained would you like to me to give it a whirl here or leave it as a follow up?

YigitElma:
> I just realised that I could use my pure jax implementation of ormqr in [jax-ml/jax#36575](https://github.com/jax-ml/jax/pull/36575) to provide a more efficient "fallback" (that even beats cusolver) as it doesn't depend on a rebuild of jaxlib so can probably run on any version of jax. Its not too complex and fairly self contained would you like to me to give it a whirl here or leave it as a follow up?

I think the current PR is a good improvement over the existing one. For the sake of keeping this one simple, I would vote to have your pure JAX implementation in a new PR if it is okay for you.

I am not very familiar with registering a primitive to JAX, but if I understand your PR correctly, instead of calling the cusolver ffi (which JAX does currently), you register your own fori_loop-based Householder implementation, am I correct?

DESC allows very different-sized optimizations. So, I am not sure what the net gain would be for different cases. If you would like to open a PR, I would be happy to test it.

jpbrodrick89:
Agreed, makes sense, this is enough for one PR. yes I use a blas 3 blocked python for loop which better utilises tensor cores rather than cusolver's apparently sequential. It works well on cpu but lapack is still better. I'll give it a whirl to see if theres a clean approach after this PR lands. Thanks!

YigitElma (approved):
Thanks!

ddudt (approved):

YigitElma:
Here are some GPU benchmarks. I solved `precise_QA` with different LMN resolution. I set the `jac_chunk_size` to 500 to focus on the QR region. Sharp peaks are jacobian computation and everthing in the middle is QR and trust region related. Jax version is 0.7.2 Cuda12 with my laptop RTX4080 12GB Vram.

<details>
<summary>Code</summary>

```python
if sys.argv[2] in ["GPU", "gpu"]:
    # Set the environment variable to use the GPU
    os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"
    from desc import set_device

    set_device("gpu")

from desc.backend import print_backend_info
from desc.examples import get

print_backend_info()

N = int(sys.argv[1])

eq = get("precise_QA")
eq.change_resolution(L=N, M=N, L_grid=2 * N, M_grid=2 * N)
eq.resolution_summary()
eq.set_initial_guess()
obj = ObjectiveFunction(ForceBalance(eq), jac_chunk_size=500, deriv_mode="batched")
obj.build()
print(f"Objective function deriv mode: {obj._deriv_mode}")
print(f"Objective function chunk size: {obj._jac_chunk_size}")

eq.solve(
    objective=obj,
    constraints=None,
    optimizer="lsq-exact",
    ftol=0,
    xtol=0,
    gtol=0,
    maxiter=int(sys.argv[3]),
    verbose=3,
    copy=False,
)
```

</details>

##  LM=16 N=8  (J is 19074 x 3336)
image

##  LMN=8   (J is 5346 x 856)
This one is way too fast to conclude on the GPU.
image

##  LM=20 N=8  (J is 29106 x 5188)
image

YigitElma:
Btw, I cannot get reliable benchmarks for newer `jax` versions (I tested with 0.10.1). But I think we can check this in #2235 after solving #2246.

YigitElma:
@jpbrodrick89 do you have any final changes in mind? If not, I think we can merge this.

jpbrodrick89:
Hi @YigitElma, no this looks nice and clean and well tested thanks for doing a production run so quickly, looks like its doing exactly what we hoped with the big difference on the first step.

I can try the pure jax ormqr as a follow-up of you're keen to try it.

By the way, are any of your team at EPS in a couple weeks? It would be great to connect!

YigitElma:
> I can try the pure jax ormqr as a follow-up of you're keen to try it.

Thanks! If you create a PR, I would be happy to test it. 

> By the way, are any of your team at EPS in a couple weeks? It would be great to connect!

Not that I am aware of. @dpanici @f0uriest @ddudt @rahulgaur104, are any of you joining?
If that doesn't work, we would be happy to have you in our dev-meeting sometime. I can share the details with you over email if you are interested.

PR #2247, opened 2026-06-19
Pure JAX implementation of ormqr for faster QR solves

jpbrodrick89:
Ported from https://github.com/jax-ml/jax/pull/36575, this was found to beat cusolver by a factor of 2 for your matrix sizes, so compounded with ormqr outperforming orgqr (raw vs economic) you should see at least a 3x speedup in each QR `ormqr` (not `qr_multiply` as a whole as the `geqrf` call is not sped up) call and probably a lower memory usage too. Crucially, **it does not require a jaxlib upgrade or rebuild.**

This employs a blocked householder multiplication tuned to an A100, your mileage may vary on other devices.

I tested this on my Mac CPU and observed speedups of 1.6-1.9x on each `qr_multiply` call.

Let me know if you think the added (hopefully temporary) maintenance burden is worth the performance gain to you?

I wasn't sure where in your codebase to put this so feel free to move this wherever you see fit.

```
(a) Big Jacobian qr_multiply(J, f) — tall, ratio 6

┌───────┬──────┬───────┬───────┬────────┬──────────┬────────────┐
│   m   │  n   │ orig  │  new  │ native │ orig/new │ native/new │
├───────┼──────┼───────┼───────┼────────┼──────────┼────────────┤
│ 1800  │ 300  │ 18.5  │ 10.1  │ 11.2   │ 1.83×    │ 1.11×      │
├───────┼──────┼───────┼───────┼────────┼──────────┼────────────┤
│ 5136  │ 856  │ 182   │ 116   │ 120    │ 1.57×    │ 1.03×      │
├───────┼──────┼───────┼───────┼────────┼──────────┼────────────┤
│ 20016 │ 3336 │ 6854  │ 4200  │ 4158   │ 1.63×    │ 0.99×      │
├───────┼──────┼───────┼───────┼────────┼──────────┼────────────┤
│ 31128 │ 5188 │ 24277 │ 12647 │ 12206  │ 1.92×    │ 0.97×      │
└───────┴──────┴───────┴───────┴────────┴──────────┴────────────┘

(b) Reduced loop qr_multiply([R; √α·I], zp) — (2n, n), per LM iteration

┌───────┬──────┬──────┬──────┬────────┬──────────┬────────────┐
│   m   │  n   │ orig │ new  │ native │ orig/new │ native/new │
├───────┼──────┼──────┼──────┼────────┼──────────┼────────────┤
│ 600   │ 300  │ 3.67 │ 2.16 │ 1.86   │ 1.70×    │ 0.86×      │
├───────┼──────┼──────┼──────┼────────┼──────────┼────────────┤
│ 1712  │ 856  │ 47.5 │ 25.8 │ 25.3   │ 1.84×    │ 0.98×      │
├───────┼──────┼──────┼──────┼────────┼──────────┼────────────┤
│ 5000  │ 2500 │ 1073 │ 611  │ 664    │ 1.76×    │ 1.09×      │
├───────┼──────┼──────┼──────┼────────┼──────────┼────────────┤
│ 10376 │ 5188 │ 6589 │ 3663 │ 3653   │ 1.80×    │ 1.00×      │
└───────┴──────┴──────┴──────┴────────┴──────────┴────────────┘
```


jpbrodrick89:
So I tried to get a handle on what was going on with tests, but I'm not very familiar with the codebase so might be a poor take.

1. The regression test seems to due to an ordering constraint that is hard for the augmented Lagrangian method to obey. I get very close to the 1e-4 limit on CI but don't reach it in 500 iterations. On my Mac I reach it in 320 iterations, but even the current status quo shows oscillation (i.e. once the constraint is met it can be broken again in subsequent iterations). I think the test isn't guaranteed to pass, and that it is flaky but I might have misunderstood.
2. Memory issues, CI reports ~30% peak memory increase, trying to run on my Mac I only measure a 10% increase. This is slightly counterintuitive to me, I tried a few fixed and none moved the dial. I think more detailed profiling is necessary to really understand what is driving the increase in memory. There's a chance this PR is a memory vs compute tradeoff that you might have to make an informed rational decision on.
3. Yes, I think the slight compile time increase is real but hopefully its worth the reduced runtime.

YigitElma:
> 1. I think the test isn't guaranteed to pass, and that it is flaky but I might have misunderstood.

That can be true. That test was prone to slight numerical changes that new jax versions introduce. It tends to be more unstable on CI, I am not sure why. But I think the problem here may be solved by changing the method that we solve the trust region problem to `svd`. The test solves an optimization problem that has bounds, and inside the bounds, the value/derivative of the objective and constraint becomes 0. This may cause the Jacobian to be singular, and in that case, the non-rank-revealing QR doesn't work. I can change the test.

But otherwise, how much numerical difference do you think your method has compared to before?

For other benchmarks, we can rerun the CI since it can be noisy due to GitHub runners. For example, this PR shouldn't have any effect on `test_objective_compile_dshape_current` or `test_build_transform_fft_lowres`. 

Memory/runtime/compile time is a hard one. Since these can also be problem dependent, we try to decide based on the most common problem types.

YigitElma:
Also, it might be worth to try jit compiling the new implementation, there seems to be multiple slicing operations that will trigger individual dispatches and compiling. We've previously seen that these individual costs can be overcome by jitting the outer function. For example, see #2153 

> Adds jit decorator to ObjectiveFunction.unpack_params() (this makes the difference in benchmarks, specifically [this line](https://github.com/PlasmaControl/DESC/blob/7238beae389ea906fac5ff90bc3a607af051dcc7/desc/objectives/utils.py#L156)) and to ObjectiveFunction.x(). The substantial speed-up (~1.5s) comes from the compilation overhead, but even after that, each call is 3-4 times faster (but these take couple milliseconds)

YigitElma:
> This may cause the Jacobian to be singular, and in that case, the non-rank-revealing QR doesn't work. I can change the test.

Yes, if I print the shape and the rank of the matrices we solve, the first step gives `(1133, 681) 473`, so it is singular. The rest of the iterations seem to be fine. Also, the test passed for me locally.

About the memory, this is a crude benchmark on laptop GPU again comparing PR (with `qr_multiply` jitted) and MASTER. Seems like `jit` solves the problem, let's see what CI benchmarks say.

image

Note: I removed jit/no jit comparison for clarity.

jpbrodrick89:
Ah, this is a good spot! I didn't know you were running your loop eagerly (I'm used to optimistix where all our loops are equinox while loops which jits the step function, any reason why don't jit your whole step function?). Looks a lot better with no jit, no memory issues. Would be interested to see how it performs on your production GPU as the laptop GPU may not be optimised enough to benefit from it (it relies on reasonable GPU kernel dispatch (we are launching a lot of kernels whereas cusolver ormqr is theoretically just one), lightning fast GEMM's (which NVIDIA GPU's have) and fairly quick TRSV's). If you just get 5% speedup like on your laptop GPU this is probably not worth it but I'mhoping it will be better than that.

YigitElma:
> any reason why don't jit your whole step function?

There are some objectives and routines that are very hard to make jit-friendly in DESC. Also, we want to keep our optimizers compatible with non-JAX functions. There is still some effort to make it possible, but it is not a priority right now.

I see around 10% improvement in speed and a 30-40% reduction in the max memory of the QR-related region. I think this is good even on my small GPU. We usually use A100's on Princeton clusters and NERSC Perlmutter, so I think your own benchmarks will be reflected here.

In general, computing the Jacobian takes significantly more memory than trust region operations, but this might be useful to minimize the total memory footprint when we compute the Jacobian column by column (then QR becomes the dominant memory bottleneck). Given that, I think we need to discuss/benchmark this further to make sure that it is worth the additional complexity.

I will trigger CI tests with older JAX versions to make sure that the dependency range is still correct.

YigitElma:
We can put the code in `desc.backend`

YigitElma (dismissed):
I just put the code to `desc.backend` and updated the test to use `svd` method since the linear system is singular. Still it would be good to have more approvals.

ddudt (dismissed):

YigitElma (approved):
I added a unit test, and put the code to `desc.backend`. It would be nice to 2 additional approvals @ddudt and @dpanici. 

@jpbrodrick89 if the changes I made looks good to you, we will try to merge this before making a new release (hopefully in 1-2 weeks).

ddudt (approved):

YigitElma:
Thanks for the contribution again! I will merge it.

PR #2215, opened 2026-05-17
available energy for turbulent transport, fast ion, and outward superbanana optimization

unalmis:
PR Age 

- exponentially accurate available energy . Validated against analytical results.
- energetic ion confinement proxies #2220 . Resolves #1748 .
- compute bug fixes #2218 .

Can use the implementation on https://github.com/unalmis/DESC for significantly improved performance.

ncsx_available_energy_wells_linear

ncsx_available_energy_wells_log

Issue #1745, opened 2025-05-29
caution when using non-analytic profiles


ddudt:
This is a general warning that if you use profiles that are not analytic throughout the entire domain, then numerical issues can arise. Related to Issue #1711. 

Most notably, the `PowerProfile` and `TwoPowerProfile` can result in profiles that are not analytic, such as $p = \sqrt{1 - \rho^2}$. In this example, the pressure gradient $\partial_\rho p$ is undefined at $\rho = 1$ and would cause issues with evaluating the `ForceBalance` objective on the plasma boundary. The derivative with respect to the exponent, like $\partial_\alpha p$ where $p = (1 - \rho^2)^\alpha$, is also undefined at $\rho = 1$ and $\alpha = \frac{1}{2}$ and would cause issues with evaluating the Jacobian of the `BoundaryError` objective. 

These are mathematical limitations and cannot be resolved with coding solutions. A realistic equilibrium should have analytic profiles, and the best way to avoid these numerical issues is to ensure that you use analytic profiles. This issue should only be resolved if we remove all profile classes that can possibly give non-analytic functions.

dpanici:
Note also that even if the pressure profile itself is analytic and fine, if it is composed of non-analytic profiles then there will still be issues (see #1711). That is, if pressure is defined as the product of two sqrt profiles, like $ p = nT = \sqrt{1 - \rho^2} \sqrt{1 - \rho^2} $, even though mathematically the pressure is analytic and fine ($1-\rho^2$), AD may still have issues at $\rho=1$ as it must take derivatives through the underlying, non-analytic sqrt profiles.

PR #2257, opened 2026-07-06
TrappedResonance

ejpaul:
Development of TrappedResonance objective function for EP optimization.


desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+            pitch_inv_weight, (grid.num_rho, n)
+        )
+
+    out = batch_map(fun, fun_data, surf_batch_size)
+
+    return out
+
+
+@register_compute_fun(
+    name="<L|r,a>",
```
dpanici: I think we already have this elsewhere
https://github.com/PlasmaControl/DESC/blob/508366a5afab22cc9ad1e185aec7ca005a1cb8c1/desc/compute/_geometry.py#L1193

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+    L_ra = simpson(
+        y=grid.meshgrid_reshape(1 / data["B^zeta"], "arz"),
+        x=grid.compress(grid.nodes[:, 2], surface_label="zeta"),
+        axis=-1,
+    )
+    data["<L|r,a>"] = grid.expand(jnp.abs(_alpha_mean(L_ra)))
+    return data
+
+
+def _s_drift_integrand(data, B, pitch):
```
dpanici: we I believe have this elsewhere

https://github.com/PlasmaControl/DESC/blob/508366a5afab22cc9ad1e185aec7ca005a1cb8c1/desc/compute/_fast_ion.py#L214

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+
+    Used in ``_trapped_EP_resonance``.
+    """
+    return safediv(
+        2 * data["cvdrift0"] * (2 - pitch * B),
+        jnp.sqrt(jnp.maximum(jnp.abs(1 - pitch * B), 1e-30)),
+    )
+
+
+def _alpha_drift_integrand(data, B, pitch):
```
dpanici, line 260: see
https://github.com/PlasmaControl/DESC/blob/508366a5afab22cc9ad1e185aec7ca005a1cb8c1/desc/compute/_fast_ion.py#L220

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+    eta_drift_avg = _jnpmean_nz(eta_drift, axis=1, fill=fill_value)
+    Omega = safediv(eta_drift_avg, omega_bounce_avg, fill=fill_value)
+    valid = (
+        _is_valid_value(eta_drift_avg, fill_value)
+        & _is_valid_value(omega_bounce_avg, fill_value)
+        & all_alpha_valid
+    )
+    Omega = jnp.where(valid, Omega, fill_value)
+
+    # Omega'(s) via finite differences
```
dpanici: Could we write this analytically? Is it worth it even?

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+    p_arr = kwargs.get("p_arr", None)
+    q_arr = kwargs.get("q_arr", None)
+    quad = kwargs.get("quad", None)
+    surf_batch_size = kwargs.get("surf_batch_size", 1)
+    num_eta = kwargs.get("num_eta", None)
+    weight_method = kwargs.get("weight_method", "linear")
+    Delta_Omega = kwargs.get("Delta_Omega", None)
+    wd_blur = kwargs.get("wd_blur", 1.25)
+    fill_value = kwargs.get("fill_value", 11)
+    eq = kwargs.get("eq", None)
```
dpanici: assuming this is an equilibrium object, it is best not to pass through to the underlying compute as we want to keep these pure and only w.r.t. the params dicts to make sure it is properly differentiable with JAX

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+
+    eta_vals = jnp.linspace(0, 2 * jnp.pi, num_eta, endpoint=False)
+    ft_denom = N * nfp - iotas * M
+    alpha_per_rho = eta_vals[None, :] * ft_denom[:, None] / nfp
+
+    eta_desc_grid = _build_eta_grid(eq, rhos, alpha_per_rho, zeta, iotas, params)
+    eta_grid = eta_desc_grid.source_grid
+
+    alpha_psa = jnp.linspace(0, 2 * jnp.pi, num_eta, endpoint=False)
+    psa_desc_grid = eq._get_rtz_grid(
```
dpanici: If this is the only thing `eq` is being used for, I would prefer if at the top of here we do `from desc.equilibrium.coords import get_rtz_grid` and use it from there to avoid needing to pass in the eq object as kwarg here

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+    base_data = compute_fun(
+        eq,
+        all_needed_keys,
+        params,
+        get_transforms(all_needed_keys, eq, base_grid, jitable=True),
+        internal_profiles,
+        data=data,
+    )
+
+    # Seed only per-surface (coordinates="r") quantities onto the 3D grids.
```
dpanici: this is a bit complex, I will try to think of a simpler way. Usually we would just precompute these things in the objective, but phase space things are trickier

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+        data_eta,
+        eta_grid,
+        num_pitch,
+        surf_batch_size,
+        pitch_invs=pitch_invs_use,
+        pitch_inv_weight=pitch_inv_weight_use,
+    )
+
+    # --- 1b. Barely-trapped filter ---
+    # Zero out wells whose poloidal bounce width delta_chi > 2.5*pi.
```
dpanici: to be consistent with the code below
```suggestion
    # Zero out wells whose poloidal bounce width delta_chi > 2*pi.
```

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+        Dictionary with keys: ``f_res``, ``Omega``, ``omega_bounce_avg``,
+        ``eta_drift_avg``, ``omega_bounce``, ``eta_drift``,
+        ``Omega_prime_s``, ``res_weight``, ``f_q_c``, ``f_q_s``,
+        ``f_q_abs``, ``Delta_s``,  ``s_res``,
+        ``f_res``.
+    """
+    m_alpha = 6.6446573450e-27
+    e_charge = 1.602e-19
+    Z = 2
+    KE = KE_frac * 5.6076e-13
```
dpanici: would be good to note what this is exactly... this is 3.5 meV converted to J right?

and then KE_frac is the fraction of the full alpha energy being considered?

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+        coordinates="rtz",
+        source_grid=raz_grid,
+        sort=False,
+        jitable=True,
+        _unique_rho_idx=unique_rho_idx,
+        _inverse_rho_idx=inverse_rho_idx,
+    )
+
+
+def _compute1D(
```
dpanici, line 195: If we can, would prefer to use Bounce2D as that is more efficient than 1D

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+        If ``True``, Delta_Omega calculation is clipped by
+        ``0.01 * max(Omega_eta) < Delta_Omega < 0.10 * max(Omega_eta)``.
+        This must be when using the ``bump`` weighting method and
+        ``Delta_Omega = None`` case. Otherwise this quantity is ignored.
+        Defaults to ``False``.
+
+    Returns
+    -------
+    result : dict
+        Dictionary with keys: ``f_res``, ``Omega``, ``omega_bounce_avg``,
```
dpanici: some definitions of each of these quantities would be useful for documentation purposes

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+    # Require particle to be trapped at all alpha/eta values for a given
+    # (rho, pitch, well).
+    all_alpha_valid = (_is_valid_value(omega_bounce, fill_value)).all(
+        axis=1
+    )  # (rho, pitch, well)
+
+    # Alpha-averaged frequencies → normalized precession Omega
+    omega_bounce_avg = _jnpmean_nz(omega_bounce, axis=1, fill=fill_value)
+    eta_drift_avg = _jnpmean_nz(eta_drift, axis=1, fill=fill_value)
+    Omega = safediv(eta_drift_avg, omega_bounce_avg, fill=fill_value)
```
dpanici: here, eta_drift_avg is actually the precession frequency omega_eta or omega_zeta right?

dario-panici (changes requested):
Overall looks good! Main change I'd request is to use the newer Bounce2D functionality for this, and an overall cleanup. The function is pretty complex and there are lots of grids being made within the compute method itself which I think maybe can be made simpler. I will try my hand this week at the Bounce2D version and see if I can think of ways to simplify it.

ejpaul:
> Overall looks good! Main change I'd request is to use the newer Bounce2D functionality for this, and an overall cleanup. The function is pretty complex and there are lots of grids being made within the compute method itself which I think maybe can be made simpler. I will try my hand this week at the Bounce2D version and see if I can think of ways to simplify it.

One issue with Bounce2D is that it uses a different Bcrit grid for each surface. We require a global Bcrit grid, since bounce-averages need to be compared from surface to surface. 

We require several grids, e.g. the eta grid must be distinct from the alpha grid. Let me know if you can see any redundancy. Happy to discuss further if it's helpful.

dpanici (commented):
forgot to leave my comments. I'll keep digging into this

desc/compute/_trapped_resonance.py
```diff
@@ -0,0 +1,917 @@
...
+    cropping_DOmega = kwargs.get("cropping_DOmega", False)
+
+    # --- 0. Build 3D grids and evaluate field data on them ---
+    base_grid = transforms["grid"]
+    iotas = base_grid.compress(data["iota"])
+    rhos = base_grid.compress(base_grid.nodes[:, 0])
+
+    eta_vals = jnp.linspace(0, 2 * jnp.pi, num_eta, endpoint=False)
+    ft_denom = N * nfp - iotas * M
+    alpha_per_rho = eta_vals[None, :] * ft_denom[:, None] / nfp
```
dpanici: I see the issue with using `Bounce2D` here, as here for a given rho, the alphas where quantities are needed vary. Bounce2D I think only accepts a single num_alpha size array for alpha and uses that for every provided rho value, so it cannot be used here without either going in and changing it to allow different alpha per rho, or using a for loop and computing things one surface at a time and re-building the Bounce2D function each time

desc/compute/_trapped_resonance.py
ejpaul: I don't think these are exactly the same. Trapped resonance only uses the periodic pieces.

desc/compute/_trapped_resonance.py
ejpaul: I'm not sure what you are asking for? A lot of this code is just checking for valid values.

desc/compute/_trapped_resonance.py
ejpaul: Not sure if we can (due to Bcrit grid needing to matching across rho grid).

desc/compute/_trapped_resonance.py
ejpaul: Agreed. Let me know how you'd like to proceed.

dpanici:
Another thought on the metric: does $\Delta s \sim |f_r|/\Omega'$ need to be raised to the 4th power? At the end of the day DESC will square the final objective value as part of a least-squares objective anyways. And as long as the radicand does not cross zero (in our case the numerator is bounded from below by zero so that should be fine), the square root should be differentiable right? Maybe I am missing something though

ejpaul:
> Another thought on the metric: does Δ s ∼ | f r | / Ω ′ need to be raised to the 4th power? At the end of the day DESC will square the final objective value as part of a least-squares objective anyways. And as long as the radicand does not cross zero (in our case the numerator is bounded from below by zero so that should be fine), the square root should be differentiable right? Maybe I am missing something though

This is a great suggestion. Right now the sum (corresponding to the phase-space average) is performed before returning the objective, so the entire sum is squared. It seems like it might make sense to pass the point-wise values (removing a square), then perform the sum-of-squares through the least-squares objective. I think one square may still be needed within the objective, since |f_r| = sqrt(...). I will take a look at this.

desc/compute/_trapped_resonance.py
unalmis: - At reasonable resolutions for the existing bounce integral objectives, the old `Bounce1D` consumes `100s of GB` of memory on the same tasks on which `Bounce2D` (e.g. on branches starting from #2215  or most optimally the [fork](https://github.com/unalmis/DESC/tree/master)) consumes `<=2 GB`. So using the old `Bounce1D` would limit the practical use of this objective.
- For the above reason, `Bounce1D` optimization was removed in #2170 .
- The logic to use the same `B_crit` across flux surfaces was exposed to be easier to enable some months ago; for example the `min_tz B` and `max_tz B` [here](https://github.com/unalmis/DESC/blob/2114b521500bc624b7e419195fb8c3f347e3668a/desc/compute/_turbulence.py#L233) can be replaced with a function that is independent of the flux surface . However, to benefit from this one would have to either merge #2215 into this branch or more simply from the fork the https://github.com/unalmis/DESC/tree/master .

ejpaul: Bounce2D now incorporated.

ejpaul:
> > Another thought on the metric: does Δ s ∼ | f r | / Ω ′ need to be raised to the 4th power? At the end of the day DESC will square the final objective value as part of a least-squares objective anyways. And as long as the radicand does not cross zero (in our case the numerator is bounded from below by zero so that should be fine), the square root should be differentiable right? Maybe I am missing something though
> 
> This is a great suggestion. Right now the sum (corresponding to the phase-space average) is performed before returning the objective, so the entire sum is squared. It seems like it might make sense to pass the point-wise values (removing a square), then perform the sum-of-squares through the least-squares objective. I think one square may still be needed within the objective, since |f_r| = sqrt(...). I will take a look at this.

I took an additional look at this strategy. If the sum is performed by the least-squares objective itself, it seems like each iteration is about 2.5 x more expensive. It also seems to converge to a pretty similar result. I don't see a clear benefit to this approach.

desc/compute/_trapped_resonance.py
unalmis: @ejpaul Do you want me to merge the bounce2d improvements into this branch and update your objective to benefit from them?

ejpaul: Thanks, @unalmis, that would be great.

dpanici: Agreed! Kaya's recent work really will significantly improve this, and if he has already made progress towards allowing alpha to differ from surface to surface or in passing a global pitch in #2215, I'd recommend rebasing this PR off of that one, or even making a new PR off of that rebase. 

I think the two things that were not able to be done with Bounce2D before were that
- it used per-flux-surface min_tz |B| and max_tz |B|, but this like you pointed out can be handled by just replacing those keys in desc_data with the global min and max before entering the Bounce2D.batch call (I think)
- Bounce2D assumed the same alpha values per flux surface. This is something that I think is an assumption  which could be relaxed (but please correct me if I am wrong), and the main place I see that needs changing is just in the theta_on_fieldlines computation, that the theta piecewise chebyshev series need to not all be for the same alpha values per rho (This should be doable? since the theta - alpha map at each flux surface is independent of the others?) 

If I am wrong on either of these though please correct me @unalmis and @ejpaul these were the main two things that your objective needed to use Bounce1D for right?

If either of these functionalities are in the other branches Kaya has, it is definitely worth rebasing this branch onto one of them, I defer to him for which is best. If there is some additional work to be done in the Bounce2D utilities to get the above two capabilities, we can do that at the level of this branch after re-basing it on the more upstream branch, or perhaps on a branch one level remove from this.

If @unalmis is willing to help on the revamp,  that would be very welcome! If not, I can also take a pass at it

ejpaul: I have done my best to merge in Bounce2D, and the objective agrees with the Bounce1D implementation. @unalmis I would appreciate if you could review this again.

desc/objectives/_trapped_resonance.py
```diff
@@ -0,0 +1,561 @@
...
+
+    Penalizes rational crossings of Omega_eta (the ratio between precessional
+    motion and bounce frequency) to minimize trapped energetic particle radial
+    motion due to resonances with magnetic field perturbations from omnigenity.
+
+    Parameters
+    ----------
+    eq : Equilibrium
+        Equilibrium that will be optimized to satisfy the Objective.
+    num_rho : int, optional
```
dpanici: We could also allow a user to pass in a custom array here. Reason for this is that some equilibria have pressure profiles that are not well-defined at rho=1 (https://github.com/PlasmaControl/DESC/issues/1745), in this current setup such an equilibrium gets NaN values for f_tr at the edge and at the node next to the edge due to `p_r` being nan at the edge, and thus `"cvdrift (periodic)"` which relies on `p_r` also is NaN, which trickles down to the trapped metric computation.

Something like `rho :  int or array`, mention in the docstring that it needs to be linearly spaced, and if is a number will be what the current default is, and we can add a check like this in the objective build like

```python
rho = np.linspace(0, 1, self._rho + 1)[1:] if isinstance(self._rho, int) else rho
errorif(not np.allclose(np.diff(rho)), ValueError, msg="rho array must be linearly spaced!")
errorif(np.any(np.isclose(rho,0.0)), ValueError, msg="rho array must not include the axis!")
```

desc/compute/_trapped_resonance.py
dpanici: I meant instead of doing finite differences to obtain Omega'(s), if we could compute Omega'(s) analytically (as in, compute the radial derivative of it explicitly instead of finite differences). I guess it would involve more bounce integrals though for the derivatives of the bounce-integrated quantities it depends on. It would at least eliminate the approximation for Omega'(s) here, not sure how important/significant that effect would be but just was wondering.

desc/objectives/_trapped_resonance.py
ejpaul: This is now addressed.

ejpaul:
@dario-panici I've now added analytic derivatives for Omega'.

PR #2264, opened 2026-07-10
Add `eq_fixed` option to `BoundaryError`

IssraAli:
An option to remove the equilibrium as an optimizable DOF when computing BoundaryError is included.


desc/objectives/_free_boundary.py
```diff
@@ -680,7 +763,8 @@ def compute(self, eq_params, *field_params, constants=None):
         Parameters
         ----------
         eq_params : dict
-            Dictionary of equilibrium degrees of freedom, eg Equilibrium.params_dict
+            Dictionary of equilibrium degrees of freedom, eg Equilibrium.params_dict,
```
YigitElma: I am not sure if the current assignment is correct. I would prefer renaming these to `params1` and `params2`. If `eq_fixed` then `field_params=params1`, else `eq_params=param1`. Finally if field_fixed then `field_params=None`.

YigitElma: See `PlasmaCoilSetDistanceBound` compute method for example

YigitElma (changes requested):
You can also add a test like `test_toroidal_flux` to check all paths run fine.

desc/objectives/_free_boundary.py
YigitElma: To clarify, the generic ObjectiveFunction methods don't know about the signature of individual compute functions. So, it will never send eq_params if eq is fixed. ObjectiveFunction just sends the params of the optimizable in the `self.things` in order, then we have to make the assignment.

YigitElma:
I think the reason tests are failing is that constants["field"] is a SumMagneticField and it expects a list for the params. So, the parameter assignment needs a bit more care. For this objective, field can be something like [MixedCoilSet(), ToroidalMagneticField()].

desc/objectives/_free_boundary.py
```diff
@@ -628,6 +640,9 @@ def build(self, use_jit=True, verbose=1):
 
         neq = 3 if self._sheet_current else 2  # number of equations we're using
 
+        if self._eq_fixed:
```
YigitElma: You already have this above in line 545, this is not necessary.

desc/objectives/_free_boundary.py
```diff
@@ -674,15 +757,19 @@ def build(self, use_jit=True, verbose=1):
...
+    def compute(self, *params, constants=None):
         """Compute boundary force error.
 
         Parameters
         ----------
-        eq_params : dict
-            Dictionary of equilibrium degrees of freedom, eg Equilibrium.params_dict
-        field_params : dict
-            Dictionary of field parameters, if field is not fixed.
+        params_1 : dict
```
YigitElma: Can you update the docstring? You don't need to write a long thing, but explain that this is the list of params, and will be assigned based on `self._eq_fixed` and `self._field_fixed`.

tests/test_objective_funs.py
```diff
@@ -610,6 +610,83 @@ def test_boundary_error_biest(self):
...
+
+        def test(eq_fixed=False, field_fixed=False):
+            obj = BoundaryError(
+                eq,
+                field,
+                eq_fixed=eq_fixed,
+                field_fixed=field_fixed,
+            )
+            obj.build()
+            if eq_fixed:
```
YigitElma: Actually, you should be able to call `obj.compute_scaled_error(*obj.xs())` if I am not mistaken. That will create the required params by using `obj.things`. Then, you can remove the conditionals.

tests/test_objective_funs.py
```diff
@@ -610,6 +610,83 @@ def test_boundary_error_biest(self):
...
```
YigitElma: Same as above

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ New Features
 - Adds an optional attribute `ion_density` to the `Equilibrium` class, to allow the ion density profile to be set independently of the electron density and effective atomic number. Also adds compute functions for ``"ni_rr"`` and ``"Zeff_rr"``.
 - Modernizes dependencies to use [``nvidia-ml-py``](https://pypi.org/project/nvidia-ml-py/) in place of [``nvgpu``](https://github.com/rossumai/nvgpu).
   If you are updating an existing software environment uninstall ``pynvml`` first and then reinstall the dependencies to correctly get ``nvidia-ml-py``.
+- Option for ``eq_fixed`` added to ``BoundaryError`` to remove equilibrium as a degree of freedom from boundary error objective.
```
YigitElma: Can you add this to the top of the file, and create a `New Features` title. Currently, it is on the old release section.

YigitElma (commented):
About the formatting and linting check, if you run the following, they will happen automatically everytime you commit something.
```
pip install pre-commit
pre-commit install
```
These and some other useful instructions can be found on the [Contributing Guide](https://github.com/PlasmaControl/DESC/blob/master/CONTRIBUTING.rst).

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+New Features
+
+- Added option for ``eq_fixed`` added to ``BoundaryError`` to remove equilibrium as a degree of freedom from boundary error objective.
```
YigitElma:
```suggestion
- Adds ``eq_fixed`` argument to ``BoundaryError`` to remove the equilibrium from the optimization. This can be used instead of adding a ``FixParameter(eq)`` constraint.
```
Or you can just fix the typo

YigitElma (commented):
Note to other devs: The failing test is related to jax-finufft version 1.2.0 and not caused by this PR.  We should solve that problem separetaly or wait CI to run that test on a different python version (which doesn't install jax-finufft 1.2.0)

ddudt (dismissed):
I fixed the failing test, simplified the new tests, and edited some of the docstrings. This approval doesn't really count.

desc/objectives/_free_boundary.py
```diff
@@ -651,6 +665,74 @@ def build(self, use_jit=True, verbose=1):
...
+            )
+            # need extra factor of B/2 bc we're evaluating on plasma surface
+            Bplasma = Bplasma + eval_data["B"] / 2
+
+            self._constants["source_data"] = source_data
+            self._constants["eval_data"] = eval_data
+            self._constants["Bplasma"] = Bplasma
+
+            # sheet current stuff
+            if self._sheet_current:
```
YigitElma, line 689: I think this should go before `virtual_casing_biot_savart` call above, because `K_vc` is required for proper virtual casing.

desc/objectives/_free_boundary.py
```diff
@@ -651,6 +665,74 @@ def build(self, use_jit=True, verbose=1):
...
+                if self._use_same_grid
+                else compute_fun(
+                    "desc.equilibrium.equilibrium.Equilibrium",
+                    self._eq_data_keys,
+                    params=self._eq.params_dict,
+                    transforms=eval_transforms,
+                    profiles=eval_profiles,
+                )
+            )
+
```
YigitElma, line 687: Move `if self._sheet_current:` block completely here. I think this was the only difference with the `compute` function, but an important one.

desc/objectives/_free_boundary.py
```diff
@@ -695,64 +778,84 @@ def compute(self, eq_params, *field_params, constants=None):
...
+            Bplasma = virtual_casing_biot_savart(
+                eval_data,
+                source_data,
+                constants["interpolator"],
+                chunk_size=self._B_plasma_chunk_size,
+            )
+            # need extra factor of B/2 bc we're evaluating on plasma surface
+            Bplasma = Bplasma + eval_data["B"] / 2
+
+            if self._sheet_current:
```
YigitElma, line 816: Same as above. The order of the computation matters for proper virtual casing.

YigitElma (changes requested):
Other than the order of the operations, it looks good to me.

desc/objectives/_free_boundary.py
```diff
@@ -760,11 +762,10 @@ def compute(self, *params, constants=None):
         Parameters
         ----------
         params : dict
-            1 or more dictionaries of params assigned in order of self.things,
-            respecting which degrees of freedom are fixed (via `eq_fixed`
-            and `field_fixed`). If eq_fixed, `params` are field_params, if
-            neither are fixed, `params[0]` takes equilibrium params, `params[1:]`
-            are field params)./
+            One or two dictionaries of parameters. If field_fixed=True, then it is the
```
IssraAli: Since the field can be a list, there can be more than two params here.

ddudt: [suggestion](https://pullpo.io/cc?l=suggestion) 
Oh you are right, my bad. If you change the first sentence back to "One or more dictionaries of parameters." then I think the rest of the docstring still works.

YigitElma (dismissed):

ddudt (dismissed):
I'm approving, but you should wait for at least one other approval since I authored some commits

lkadz (dismissed):

YigitElma (dismissed):
There was just a merge conflict in CHANGELOG

desc/objectives/_free_boundary.py
```diff
@@ -651,6 +665,74 @@ def build(self, use_jit=True, verbose=1):
...
+                    params=self._eq.params_dict,
+                    transforms=eval_transforms,
+                    profiles=eval_profiles,
+                )
+            )
+
+            # sheet current stuff
+            if self._sheet_current:
+                p = (
+                    "desc.magnetic_fields._current_potential."
```
dpanici: you can just use eq.surface, don't need to use the string. But this is fine, just he way it is formatted looks strange

ddudt (dismissed):

tests/test_compute_everything.py
```diff
@@ -76,11 +76,13 @@ def _compare_against_rpz(p, data, data_rpz, coordinate_conversion_func):
         if data_index[p][name]["dim"] != 3:
             continue
         res = coordinate_conversion_func(data, name) - data_rpz[name]
+        rtol = 1e-4 if "Gamma_" in name else 1e-8  # jax-finufft version differences
```
YigitElma:
```python
if Version(jax_finufft.__version__) > Version("1.2.0"):
    rtol = 1e-8
else:
    rtol = 1e-4 if "Gamma_" in name else 1e-8 
```

YigitElma (approved):

ddudt (approved):

dpanici (approved):

PR #2266, opened 2026-07-13
Add Benchmarks regarding Coil Optimization

YigitElma:
- Adds benchmarks for compute and jacobian of `QuadraticFlux` with a mixed coilset that is complex enough.
  - I tried to hit NFP, sym, and different coilset for loops of `compute_magnetic_field`
  - These tests will start running properly once the PR is merged, because those actions run under `pull_request_target`, that means they always use `master` code
- Fixes bug in `plot_coils` and `CoilSet.from_symmetry` that ignored the given `check_intersection` 
- Adds `check_intersection` to coil set initializers
- Sets the default of `check_intersection` to False in many places. Only exception is `FourierCurrentPotentialField.to_CoilSet` as it seems a bit more important.
- Passes built `LinearConstraintProjection` object in `perturb` benchmarks to make it faster. We internally use it in this way anyway. Also, we benchmark `LinearConstraintProjection.build()` separately. Also takes it out of `setup` to prevent triggering `build` multiple times.


desc/coils.py
```diff
@@ -2081,12 +2083,14 @@ def from_symmetry(cls, coils, NFP=1, sym=False, check_intersection=True):
                 fcoil.flip([0, 0, 1])
                 fcoil.current = -1 * coil.current
                 flipped_coils.append(fcoil)
-            coils = coils + flipped_coils
```
lkadz, line 2084: Maybe just doing this would prevent from coding the entire 'for' loop:
`coils = CoilSet(list(coils) + flipped_coils, check_intersection=False)`

YigitElma: I don't think that is a problem. `CoilSet` creation has some other redundant operations, and this skips those too.

desc/coils.py
```diff
@@ -2081,12 +2083,14 @@ def from_symmetry(cls, coils, NFP=1, sym=False, check_intersection=True):
...
         # next rotate the coilset for each field period
+        coilset = []
         for k in range(0, NFP):
-            rotated_coils = coils.copy()
-            rotated_coils.rotate(axis=[0, 0, 1], angle=2 * jnp.pi * k / NFP)
-            coilset += rotated_coils
+            for coil in base_coils:
+                rcoil = coil.copy()
+                rcoil.rotate(axis=[0, 0, 1], angle=2 * jnp.pi * k / NFP)
+                coilset.append(rcoil)
```
ddudt, line 2097: Why was this indented? It looks like now this only gets called `if sym`, but you can have a coil set with NFP>1 and sym=False. I think these needs to be un-indented like the original code.

desc/coils.py
```diff
@@ -3356,7 +3360,7 @@ def _linking_number(x1, x2, x1_s, x2_s, dx1, dx2):
     return ratio.sum()
 
 
-def initialize_modular_coils(eq, num_coils, r_over_a=2.0):
+def initialize_modular_coils(eq, num_coils, r_over_a=2.0, check_intersection=True):
```
ddudt: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
I would be in favor of changing the default to `check_intersection=False` everywhere. I think that is the more common use case and is significantly faster for large coil sets.

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -553,3 +555,71 @@ def run(x, prox):
...
+
+    def run(x, objective):
+        objective.jac_scaled_error(x).block_until_ready()
+
+    benchmark.pedantic(run, args=(x, objective), rounds=10, iterations=1)
+
+
+@pytest.mark.slow
+@pytest.mark.benchmark
+def test_objective_quadratic_flux_compute(benchmark):
```
ddudt: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
`test_objective_quadratic_flux_jac` and `test_objective_quadratic_flux_compute` are almost identical and only differ by one line.  You could put all the common code in a separate setup function that they both call to reduce the amount of redundant code. That setup function could also be called by `tests/benchmarks/memory_funcs.py/test_objective_quadratic_flux_jac` and the new tests in `tests/benchmarks/benchmark_gpu_small.py`. That would simplify the code by over 100 lines and be easier to maintain.

ddudt (changes requested):

desc/coils.py
YigitElma: The only change here is that instead of adding the coils by using the dunder `__add__` method, I append them in a list, then call the CoilSet only once. I don't think the other behavior changes. I need a for loop because `base_coils` is a list not a single `CoilSet`

desc/coils.py
YigitElma: I also prefer this, but I didn't want to change the default.

YigitElma:
Default to no check for intersections, change the docstring for new default, add to the changelog.

tests/benchmarks/benchmark_cpu_small.py
```diff
@@ -553,3 +554,51 @@ def run(x, prox):
...
+    run, x = _test_quadratic_flux(30, "jac")
+    benchmark.pedantic(run, args=(x,), rounds=10, iterations=1)
+
+
+@pytest.mark.slow
+@pytest.mark.benchmark
+def test_objective_quadratic_flux_compute(benchmark):
+    """Benchmark computing QuadraticFlux."""
+    run, x = _test_quadratic_flux(100, "compute")
+    benchmark.pedantic(run, args=(x,), rounds=50, iterations=1)
```
lkadz, line 604: The GPU file's compute benchmark uses `rounds=10` while CPU's uses `rounds=50 `. Is it intentional?

YigitElma: Good catch, thanks!

IssraAli (dismissed):

ddudt (dismissed):

lkadz (dismissed):

ddudt (dismissed):

YigitElma:
Sorry for the last minute changes, I realized that memory benchmark script had a bug in the import and the benchmarks in general were taking too long on CI (the jacobian one was 4 seconds on my laptop CPU but 40 seconds on CI). I fixed the bug and made the tests smaller.

ddudt (approved):

lkadz (approved):

PR #2267, opened 2026-07-15
Add surf_batch_size to Boozer transform to control memory usage

dpanici:
- Add `surf_batch_size` as a kwarg to the Boozer transform compute variables to allow control of memory usage when using with large number of surfaces 
- Added as a plumb-thru kwarg to
  - `QuasisymmetryBoozer` and `Omnigenity` objectives
  - `make_boozmn_output` utility function

Especially helpful when making high res boozmn.nc using DESC


desc/compute/_omnigenity.py
```diff
@@ -906,7 +955,7 @@ def _interp(x, B):
         return interp1d(x, eta_input, B, method="monotonic-0")
 
     # |B|_omnigeneous is an even function so B(-eta) = B(+eta) = B(|eta|)
-    B = vmap(_interp)(jnp.abs(eta), B_input.T)  # shape (nr, nt*nz)
+    B = vmap_chunked(_interp)(jnp.abs(eta), B_input.T)  # shape (nr, nt*nz)
```
YigitElma: Missing chunk_size and in_axis

YigitElma (changes requested):

desc/objectives/_omnigenity.py
```diff
@@ -590,6 +604,7 @@ def __init__(
         self.eta_weight = eta_weight
         self._eq_fixed = eq_fixed
         self._field_fixed = field_fixed
+        self._surf_batch_size = surf_batch_size
```
YigitElma, line 608: Do you need to add this to static attributes?

desc/objectives/_omnigenity.py
```diff
@@ -58,13 +63,15 @@ def __init__(
...
         jac_chunk_size=None,
+        surf_batch_size=None,
     ):
         if target is None and bounds is None:
             target = 0
         self._grid = grid
         self.helicity = helicity
         self.M_booz = M_booz
         self.N_booz = N_booz
+        self.surf_batch_size = surf_batch_size
```
YigitElma: May need to include in `static_attrs`

desc/objectives/_omnigenity.py
dpanici: I think you are right

YigitElma (approved):

desc/compute/_omnigenity.py
```diff
@@ -36,14 +37,17 @@
     grid_requirement={"is_meshgrid": True, "sym": False},
     M_booz="int: Maximum poloidal mode number for Boozer harmonics. Default 2*eq.M",
     N_booz="int: Maximum toroidal mode number for Boozer harmonics. Default 2*eq.N",
+    surf_batch_size="int: Number of flux surfaces to compute simultaneously. Defaults"
+    " to ``grid.num_rho`` e.g. compute all flux surfaces simultaneously. Decrease "
+    "to reduce memory required for computation.",
```
ddudt, line 42: [question](https://pullpo.io/cc?l=question) 
Really it defaults to `chunk_size=None` and that defaults to no chunking, right?

ddudt (commented):
Looks fine, but has it been tested? Like running a test with multiple surfaces, and using this to only compute a few surfaces at a time.

desc/compute/_omnigenity.py
YigitElma: Yes, there is no explicit default, but `kwargs.get("surf_batch_size")` returns `None` and it computes everything without chunking.

ddudt (approved):

PR #1907, opened 2025-09-12
Make continuation method fall back to shaping perturbation first if pressure first fails

dpanici:
- Adds `shaping_first` arg to `solve_continuation_automatic`
 - this would allow one to make shaping first then pressure perturbations which is most robust
- by default, still try pressure first, but then if pressure fails, do shaping then pressure first. This should ensure auto continuation is as robust as possible, but tries the more efficient pressure then shaping first before falling back on shaping then pressure.

Resolves #1903 (tested it on the stellarator there and it works well at least)
Resolves #1972 
Resolves #1914

PR #2268, opened 2026-07-17
Fix reactor_QA script to account for ScaledProfiles being returned now by rescale

dpanici:
In our reactor QA script, we use `FixCurrent` assuming the [0,1] indices are the constant and rho^1 modes of the profile. However after the changes in #1871, rescaling an eq now returns `ScaledProfile` object which means now that [0,1] indices are the scale and the constant mode of the power series.

this resulted in letting the linear term be free, and a nonzero derivative of current at the axis which is unphysical (this also leads to unphysical iota profiles due to the way the axis limit of iota works compared to how iota is computed away from rho=0). This fixes that unintentional bug and update the reactor_QA output accordingly.

The reactor_QA example .h5 file had this issue since #1907 updated it (my bad...). 

image


dpanici:
Test was failing bc the iota profile is not well-represented by an even power series for reactorQA (see below, eq1 is the real eq, eq2 is after loading from the asciiIO, where the iota was obtained with get_profile as a power series with default L=eq.L=8 and even symmetry), it was surprising it passed before. I loosened the tols even more now since it is more an exercist of the geometry and the logic for kinettics

image

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ Bug Fixes
 - Fixes bug that was always setting NFP=1 in ``to_FourierRZ`` methods.
 - Fixes ``VMECIO.save`` metadata for current-density variables and corrects the
   asymmetric ``currvmns`` magnetic-axis extrapolation.
+- Fixes bug in ``reactor_QA.py`` script where the current profile was allowed to have a nonzero rho^1 component, which resulted in annonphysical profile near-axis.
```
YigitElma:
```suggestion
- Fixes bug in ``reactor_QA.py`` script where the current profile was allowed to have a nonzero rho^1 component, which resulted in an unphysical profile near-axis.
```

desc/examples/reactor_QA.py
```diff
@@ -55,6 +55,9 @@
 
 # scale to ARIES-CS reactor size
 eq = rescale(eq, L=("a", a), B=("<B>", B), scale_pressure=False, copy=True, verbose=1)
```
YigitElma, line 57: I think the conversion we introduce here might not be apparent to the users. The current docstring says,
```
    Returns
    -------
    eq : Equilibrium or iterable of Equilibrium
        Same as input, but rescaled to the desired size and magnetic field strength.
        The profiles will now be ``ScaledProfile` classes, with the ``scale`` being
        the rescaling parameter computed in this function.
```
I think we can make it more obvious that the first element of the profile is `scale` which is also an optimizable parameter. Tbh, I had not realized that until #2175. Adding a warning might be an overkill and can cause many tests to fail, just better wording in the docstring can be good.

YigitElma (commented):
I don't know if it is really necessary, but I think the conversion is subtle enough to have more detailed explanation in docstring

ddudt (commented):
While reviewing this PR, I noticed that there were a few references to `Equilibrium.ion_density` missing after it was introduced in #2090. I added that fix to this PR and clarified some of the documentation. (Sorry for the scope creep.)

FYI I always do the following after calling `rescale` to avoid this issue (assuming the current profile was originally a power series):
```
eq = rescale(eq, L=("R0", R0), B=("B0", B0), scale_pressure=False, verbose=1)
eq.current = PowerSeriesProfile(params=eq.c_l[1:] * eq.c_l[0])
```

desc/objectives/_bootstrap.py
```diff
@@ -160,24 +160,27 @@ def build(self, use_jit=True, verbose=1):
             RuntimeError,
             "Bootstrap current calculation requires an ion temperature profile.",
         )
+        errorif(
+            eq.ion_density is None and eq.atomic_number is None,
+            RuntimeError,
+            "Bootstrap current calculation requires an ion density profile.",
```
dpanici, line 166: Does it? I thought if there is no ion density profilethen Zeff + quasineutrality is enough?

ddudt: It needs to compute the ion density. It can either do that from an `ion_density` profile or the `electron_density` and `atomic_number` profiles (assuming n_i = n_e / Z_eff). Before it was only checking for the `electron_density` profile, but we also need this check for either the `ion_density` or `atomic_number`.

desc/compat.py
```diff
@@ -179,9 +179,10 @@ def rescale(
     Returns
     -------
     eq : Equilibrium or iterable of Equilibrium
-        Same as input, but rescaled to the desired size and magnetic field strength.
-        The profiles will now be ``ScaledProfile` classes, with the ``scale`` being
-        the rescaling parameter computed in this function.
+        Equilibrium input rescaled to the desired size and magnetic field strength.
+        The current profile (but not iota) will become a ``ScaledProfile``, which may
+        change the size of its parameters array. If scale_pressure is True, then the
```
YigitElma:
```suggestion
        change the number of optimizable parameters of the profile by adding `scale` to the first index. If scale_pressure is True, then the
```
Something like this can prevent unfamiliar users from trying to figure out what the change is.

YigitElma (dismissed):
Looks good to me. I would prefer an even more elaborate explanation but non-blocking.

YigitElma (dismissed):

ddudt (dismissed):

desc/objectives/_power_balance.py
```diff
@@ -92,9 +92,10 @@ def build(self, use_jit=True, verbose=1):
         """
         eq = self.things[0]
         errorif(
-            eq.electron_density is None,
+            (eq.ion_density is None)
+            and (eq.electron_density is None or eq.atomic_number is None),
             ValueError,
-            "Equilibrium must have an electron density profile.",
+            "Equilibrium must have an ion density profile.",
```
dpanici, line 98:
```suggestion
            "Equilibrium must have an ion density profile. Please assign either an `ion_density` profile or `electron_density` and `atomic_number` profiles.",
```

YigitElma (approved):

ddudt (approved):
Since I made some of the changes this approval doesn't really count.

desc/objectives/_bootstrap.py
```diff
@@ -160,24 +160,27 @@ def build(self, use_jit=True, verbose=1):
...
+        )
```
lkadz, line 167:
```suggestion
        errorif(
            eq.atomic_number is None,
            RuntimeError,
            "Bootstrap current calculation requires an atomic number (Zeff) profile.",
        )
        errorif(
            eq.ion_density is None and eq.electron_density is None,
            RuntimeError,
            "Bootstrap current calculation requires an ion density or electron density profile.",
        )
```

lkadz: Previously it only guards against both being missing. It does not require `atomic_number` when `ion_density` is supplied directly. But the actual Redl formula requires `atomic_number` in `desc/compute/_bootstrap.py#L148`:

`Zeff = jnp.maximum(1 + 1.0e-14, profile_data["Zeff"])`

And per `desc/compute/_profiles.py#L472`:

```
def _Zeff(...):
    if profiles["atomic_number"] is not None:
        data["Zeff"] = profiles["atomic_number"].compute(...)
    else:
        data["Zeff"] = jnp.nan * data["0"]
```

docs/notebooks/tutorials/bootstrap_current.ipynb
```diff
@@ -147,7 +147,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": null,
```
lkadz, line 150: is it normal?

lkadz (commented):
Before approving, I just had these minor changes / questions

desc/objectives/_bootstrap.py
dpanici: https://github.com/PlasmaControl/DESC/blob/7841b520d649981d46adaf043c896370dea7909e/desc/equilibrium/equilibrium.py#L374

this is set to 1 if not given, so it would be hard to have kinetic profiles without the atomic number existing. @ddudt also said we want ion density to exist going forward so it is a conscious decision on the user to set

docs/notebooks/tutorials/bootstrap_current.ipynb
dpanici: yea don't worry about the jupyter metadata

desc/objectives/_bootstrap.py
ddudt: The Redl formula requires all kinetic profiles: Te, Ti, ne, ni, and Zeff. The only exception is that if ni is not given, it is assumed to be ne/Zeff. `atomic_number = 1` as a default in `Equilibrium.__init__` but you could initialize an eq with a pressure profile then switch it to kinetic profiles later and forget to set the atomic number.

PR #2277, opened 2026-07-29
Fix for NaN in reverse mode gradient of FourierPlanarCoil and Hessian Scaling Issue for small problems

### LINKED ISSUE #2276: `FourierPlanarCoil` still gives NaN gradient when axis is vertical
I thought we had fixed this before #1456 but I guess not?

```python
from desc.coils import FourierPlanarCoil
from desc.objectives import ObjectiveFunction, CoilLength

coil = FourierPlanarCoil(normal=[0,0,1])
opt = Optimizer("lsq-exact")

obj = ObjectiveFunction(CoilLength(coil))

# Works fine, no Nan
opt.optimize(coil,objective=obj,verbose=3,maxiter=2);


coil = FourierPlanarCoil(normal=[1e-18,1e-18,1])
opt = Optimizer("lsq-exact")

obj = ObjectiveFunction(CoilLength(coil))
# Nan Optimality when normal is close but not quite vertical
opt.optimize(coil,objective=obj,verbose=3,maxiter=2);
```

- No NaN in forward mode
- No NaN if the normal is [2e-8, 2e-8, 1], but NaN when the first two components are anything less than 2e-8
  - points to handling of the allclose [here ](https://github.com/PlasmaControl/DESC/blob/c2dfad10a39acf394e7bec544dbf73684ca1bc36/desc/compute/_curve.py#L229)which depends on the dotproduct I guess? though these are not aligned with negative Z axis...

Our current tests dont catch this, if I add 
```python
      test([0, 1e-8, 1])  # nearly parallel
      test([0, 1e-8, -1])  # nearly antiparallel
```
[to our test for this](https://github.com/PlasmaControl/DESC/blob/c2dfad10a39acf394e7bec544dbf73684ca1bc36/tests/test_objective_funs.py#L4079), these fail.

dpanici:
Resolves #2276

Fix was mainly in switching from axis-angle to quaternion for general case, and being sure that the antiparllel edge-case is handled by making a rotation of 180 degrees around a perpendicular vector composed of the input vector (ensuring that the derivative does not arbitrarily go to exactly zero wrt normal at that edge case, which could stall optijmizations even if the deriv was not NaN there)

Fixes
- NaN gradient occuring for horizontal coils
- zero gradient occuring for horizontal coils which arrests the optimization there incorrectly
- Hessian scaling issue for tiny problems (small # DOFs) which popped up in the tests

```python
from desc.coils import FourierPlanarCoil
from desc.objectives import ObjectiveFunction, CoilLength
import numpy as np

def test(normal):
    coil = FourierPlanarCoil(normal=normal)
    opt = Optimizer("lsq-exact")

    obj = ObjectiveFunction(CoilLength(coil))
    obj.build(verbose=0)
    g = obj.grad(obj.x(coil))
    return np.any(np.isnan(g))
for comp in np.concatenate([np.array([0.0]), np.logspace(-16,-6,11)]):
    was_nan = test([comp,comp,1.0])
    print(f"Did Normal of [{comp:1.2e}, {comp:1.2e},1.0] result in nan gradient?  {was_nan}")

```

On master:
```python
Did Normal of [0.00e+00, 0.00e+00,1.0] result in nan gradient?  False
Did Normal of [1.00e-16, 1.00e-16,1.0] result in nan gradient?  True
Did Normal of [1.00e-15, 1.00e-15,1.0] result in nan gradient?  True
Did Normal of [1.00e-14, 1.00e-14,1.0] result in nan gradient?  True
Did Normal of [1.00e-13, 1.00e-13,1.0] result in nan gradient?  True
Did Normal of [1.00e-12, 1.00e-12,1.0] result in nan gradient?  True
Did Normal of [1.00e-11, 1.00e-11,1.0] result in nan gradient?  True
Did Normal of [1.00e-10, 1.00e-10,1.0] result in nan gradient?  True
Did Normal of [1.00e-09, 1.00e-09,1.0] result in nan gradient?  True
Did Normal of [1.00e-08, 1.00e-08,1.0] result in nan gradient?  True
Did Normal of [1.00e-07, 1.00e-07,1.0] result in nan gradient?  False
Did Normal of [1.00e-06, 1.00e-06,1.0] result in nan gradient?  False
```

On this PR:

```python
Did Normal of [0.00e+00, 0.00e+00,1.0] result in nan gradient?  False
Did Normal of [1.00e-16, 1.00e-16,1.0] result in nan gradient?  False
Did Normal of [1.00e-15, 1.00e-15,1.0] result in nan gradient?  False
Did Normal of [1.00e-14, 1.00e-14,1.0] result in nan gradient?  False
Did Normal of [1.00e-13, 1.00e-13,1.0] result in nan gradient?  False
Did Normal of [1.00e-12, 1.00e-12,1.0] result in nan gradient?  False
Did Normal of [1.00e-11, 1.00e-11,1.0] result in nan gradient?  False
Did Normal of [1.00e-10, 1.00e-10,1.0] result in nan gradient?  False
Did Normal of [1.00e-09, 1.00e-09,1.0] result in nan gradient?  False
Did Normal of [1.00e-08, 1.00e-08,1.0] result in nan gradient?  False
Did Normal of [1.00e-07, 1.00e-07,1.0] result in nan gradient?  False
Did Normal of [1.00e-06, 1.00e-06,1.0] result in nan gradient?  False
```


dpanici:
```python
    axis = jnp.asarray(axis)
    norm = safenorm(axis)
    if angle is None:
        angle = norm
    eps = 1e3 * jnp.finfo(axis.dtype).eps # 1e3 instead of 1e2
    no_rotation = norm < eps
```
If the eps is made an order of magnitude larger, the 1e-13 results in no NaN.

So the issue comes when the axis magnitude from the cross of the normal and the Z-axis is just large enough that the norm becomes greater than our threshold for zero, but is small enough still that we don't trigger some other conditional on the magnitude of some intermediate quantity.

So I guess there is some inconsistency in thresholding somewhere in this.

I think it is that `safearccos` will return an infinite for the angle when their `dotprod` evaluates to 1 s.t. `abs(dotprod)==1` is True
https://github.com/PlasmaControl/DESC/blob/c2dfad10a39acf394e7bec544dbf73684ca1bc36/desc/utils.py#L945

and we then rely on that only happening when the Zaxis and normal are nearly parallel, so their cross product's norm is smaller than 1e2*eps in rotation_matrix and thus we never take cos(inf) but instead just return the identity matrix 
https://github.com/PlasmaControl/DESC/blob/c2dfad10a39acf394e7bec544dbf73684ca1bc36/desc/utils.py#L1008

or, if the axis is parallel to -Z, the correct reflection
https://github.com/PlasmaControl/DESC/blob/c2dfad10a39acf394e7bec544dbf73684ca1bc36/desc/compute/_curve.py#L228-L232

I know the answer is in the following data, just need to wrap my head around it:

```python
from desc.coils import FourierPlanarCoil
from desc.objectives import ObjectiveFunction, CoilLength
import numpy as np
from desc.utils import dot
from desc.utils import safearccos,safenormalize
from desc.backend import jnp
zaxis = jnp.array([0,0,1])


def test(normal):
    coil = FourierPlanarCoil(normal=normal,basis="xyz")
    print(f"norm of normal, pre-normalization: [{np.linalg.norm(coil.normal):1.16e}]")
    print(f"coil normal after normalization: [{coil.normal[0]:1.16e},{coil.normal[1]:1.16e},{coil.normal[2]:1.16e}]")

    opt = Optimizer("lsq-exact")

    obj = ObjectiveFunction(CoilLength(coil))
    obj.build(verbose=0)
    g = obj.grad(obj.x(coil))
    return np.any(np.isnan(g)), coil
for comp in np.concatenate([np.array([0.0]), np.logspace(-16,-6,11)]):
    was_nan, coil = test([comp,comp,1.0])
    print(f"Did Normal of [{comp:1.2e}, {comp:1.2e},1.0] result in nan gradient?  {was_nan}")
    dotprod = dot(zaxis,coil.normal)
    print(f"dot of normal and z axis == 1: {jnp.abs(dotprod) == 1}")
    print(f"arccos of dotprod: {safearccos(dotprod)}")


```

```python
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [0.0000000000000000e+00,0.0000000000000000e+00,1.0000000000000000e+00]
Did Normal of [0.00e+00, 0.00e+00,1.0] result in nan gradient?  False
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [9.9999999999999998e-17,9.9999999999999998e-17,1.0000000000000000e+00]
Did Normal of [1.00e-16, 1.00e-16,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [1.0000000000000001e-15,1.0000000000000001e-15,1.0000000000000000e+00]
Did Normal of [1.00e-15, 1.00e-15,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [1.0000000000000000e-14,1.0000000000000000e-14,1.0000000000000000e+00]
Did Normal of [1.00e-14, 1.00e-14,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [1.0000000000000000e-13,1.0000000000000000e-13,1.0000000000000000e+00]
Did Normal of [1.00e-13, 1.00e-13,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [9.9999999999999998e-13,9.9999999999999998e-13,1.0000000000000000e+00]
Did Normal of [1.00e-12, 1.00e-12,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [9.9999999999999994e-12,9.9999999999999994e-12,1.0000000000000000e+00]
Did Normal of [1.00e-11, 1.00e-11,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [1.0000000000000000e-10,1.0000000000000000e-10,1.0000000000000000e+00]
Did Normal of [1.00e-10, 1.00e-10,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [1.0000000000000001e-09,1.0000000000000001e-09,1.0000000000000000e+00]
Did Normal of [1.00e-09, 1.00e-09,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [1.0000000000000000e-08,1.0000000000000000e-08,1.0000000000000000e+00]
Did Normal of [1.00e-08, 1.00e-08,1.0] result in nan gradient?  True
dot of normal and z axis == 1: True
arccos of dotprod: inf
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [9.9999999999999003e-08,9.9999999999999003e-08,9.9999999999999001e-01]
Did Normal of [1.00e-07, 1.00e-07,1.0] result in nan gradient?  False
dot of normal and z axis == 1: False
arccos of dotprod: 1.4136482746161737e-07
norm of normal, pre-normalization: [1.0000000000000000e+00]
coil normal after normalization: [9.9999999999900003e-07,9.9999999999900003e-07,9.9999999999900013e-01]
Did Normal of [1.00e-06, 1.00e-06,1.0] result in nan gradient?  False
dot of normal and z axis == 1: False
arccos of dotprod: 1.4141194121979817e-06
```


OK, so issue is that when a normal is almost aligned with the z-axis (with a z-component of 1 but some additional other tiny but nonzero component), but the misalignment is so small that the norm of the vector is 1 to machine precision, then when we go to normalize the vector when it is set in the FourierPlanarCoil, the normalization does not change the z-axis of the normal.
So, that sets the scene for the bug: when the z-axis component of a FourierPlanarCoil's normal is exactly 1.0, but it has nonzero off-z-axis components to it (tiny enough that the norm of the vector is 1.0 to machine precision, thus the normalization of the normal does not reduce the z-axis of it to below 1.0)

Then, we can have this inconsistency occur where we in one part of the compute logic, (the safearccos), the checked quantity (dotprod of normal and Zaxis) is 1.0 to machine precision and thus we trigger the safe route there and return inf
but then inside of rotation_matrix, the norm of the axis (cross product of zaxis and normal) is NOT small enough to trigger the safe route there, and we end up returning things that result in nans (I THINK, I am not sure)

dpanici:
Updated test and solution


```python
from desc.coils import FourierPlanarCoil
from desc.objectives import ObjectiveFunction, CoilLength
import numpy as np
from desc.utils import dot, cross
from desc.utils import safearccos,safenormalize, safenorm
from desc.backend import jnp
zaxis = jnp.array([0,0,1])


def test(normal):
    coil = FourierPlanarCoil(normal=normal,basis="xyz")
    print(f"norm of normal, pre-normalization: [{np.linalg.norm(coil.normal):1.16e}]")
    print(f"coil normal after normalization: [{coil.normal[0]:1.16e},{coil.normal[1]:1.16e},{coil.normal[2]:1.16e}]")

    opt = Optimizer("lsq-exact")

    obj = ObjectiveFunction(CoilLength(coil))
    obj.build(verbose=0)
    g = obj.grad(obj.x(coil))
    return np.any(np.isnan(g)), coil
for comp in np.concatenate([np.array([0.0]), np.logspace(-16,-6,11)]):
    normal = np.array([comp,comp,1.0])
    was_nan, coil = test(normal)
    axis = cross(zaxis, normal)
    norm = safenorm(axis)

    eps = 1e8 * jnp.finfo(axis.dtype).eps
    no_rotation = norm < eps
    print(f"Did Normal of [{comp:1.2e}, {comp:1.2e},1.0] result in nan gradient?  {was_nan}")
    dotprod = dot(zaxis,coil.normal)
    
    print(f"arccos of dotprod: {safearccos(dotprod)}")
    print(f"zaxis x normal = rotation axis: {axis}")
    print(f"norm of rotation axis: {norm}")
    print(f"dot of normal and z axis == 1: {jnp.abs(dotprod) == 1}")
    print(f"is norm of rotation axis < eps?: {no_rotation}")
    print("#"*15)

```

We need a consistency between the two "safeness" logic checks: is the dot of the normal and zaxiz ==1 (or -1), and is the norm of the rotation axis < eps

Because the axis could contain two components which are < sqrt(eps) ~ 1e-8, when we take the norm of axis we square it and they become comparable to zero. 

Gist of it is, we should use sqrt(eps) instead of eps when we try to determine if an axis' norm is zero or not. Making that change results in no NaN.


Outstanding issue: the gradient wrt the normal vector before was ZERO for normal=[eps,eps,1] with eps<1e-15 and then NaN for 1e-15>eps>1e-8. With the change in this PR, the gradient is now just ZERO for normal=[eps,eps,1] with eps<1e-8. This can cause issues in optimization if your coil is initialized as horizontal

```python
normal = [0,0,1.0]

coil = FourierPlanarCoil(center = [10,1e-1,-3], r_n=0.5,normal=normal,basis="xyz", current=1e6)

from desc.objectives import SurfaceQuadraticFlux, FixCoilCurrent, FixParameters
surf = FourierRZToroidalSurface() # should be R=10 r=1 surface

# opt problem  fix the geometry of coil, the current, and its center location of the coil
# to minimize  Bn, should rotate from horizontal to vertical

cons = (FixCoilCurrent(coil), FixParameters(coil, {"r_n":True,"center":True}), FixParameters(surf))
obj = ObjectiveFunction(SurfaceQuadraticFlux(surf,coil,eval_grid=LinearGrid(N=10,M=10), field_grid=LinearGrid(N=10)))
opt = Optimizer("lsq-exact")
opt.optimize((surf,coil),objective=obj, constraints=cons,verbose=3,ftol=0,gtol=0);
```

This stalls with 0 gradient initially

```python
normal = [1e-7,0,1.0]

coil = FourierPlanarCoil(center = [10,1e-1,-3], r_n=0.5,normal=normal,basis="xyz", current=1e6)

from desc.objectives import SurfaceQuadraticFlux, FixCoilCurrent, FixParameters
surf = FourierRZToroidalSurface() # should be R=10 r=1 surface

# opt problem  fix the geometry of coil, the current, and its center location of the coil
# to minimize  Bn, should rotate from horizontal to vertical

cons = (FixCoilCurrent(coil), FixParameters(coil, {"r_n":True,"center":True}), FixParameters(surf))
obj = ObjectiveFunction(SurfaceQuadraticFlux(surf,coil,eval_grid=LinearGrid(N=10,M=10), field_grid=LinearGrid(N=10)))
opt = Optimizer("lsq-exact")
opt.optimize((surf,coil),objective=obj, constraints=cons,verbose=3,ftol=0,gtol=0);
```

This does not stall out and correctly moves the coil normal so that the dipole moment is parallel to the surface (minimizing Bn)

dpanici:
OK this works: switched to using quaternions for general rotation from one vector onto another, and for the antiparallel case made sure to use the vector that we want the deriv wrt (normal vector) to be nonzero so that case is covered too. This works for both parallel to z and antiparallel to z without nan OR zero gradient (for cases where e.g. coil is horizontal but we would expect the gradient to be nonzero

dpanici:
The test failing is not due to this PR really, as the main issue is that the rotation matrix derivative AD wrt normal vector z component can be really tiny/near zero in certain cases. This leads then to the Hessian scale (~1/diag of hess) being massive. Our epsilon check when Hess size is small (like in the test, only 8x8) allows things like 1/1e-14 to get thru which then completely ruins the scale of the problem.

This PR addresses that as well by just slightly increasing the epsilon under which scale_inv is converted to 1 (multiplies by a factor of 10). This was enough to overcome the numerical noise, it seems

Using this code:

```python
scales=[]
scales_inv=[]

hess_mats=[]
for normal_comp in np.logspace(-16,-1,16):
    c = FourierPlanarCoil(normal=[normal_comp,1.0,0.0])

    angle = np.arccos(np.dot(np.array([0,0,1]), c.normal))
    print(f"{angle=:1.2e}")
    
    obj = ObjectiveFunction(CoilLength(c))
    obj.build(verbose=0)
    h=obj.hess(obj.x(c))
    hess_mats.append(h)
    scale,scale_inv = compute_hess_scale(h)
    print(scale)
    scales.append(scale.copy())
    scales_inv.append(scale_inv.copy())
    
plt.figure()
plt.scatter(np.logspace(-16,-1,16), np.max(scales,axis=1))
plt.xlabel("off-normal component of normal vector")
plt.ylabel("Max xscale from hessian")
plt.xscale("log")
plt.yscale("log")


```


With the default epsilon we had before in `compute_hess_scale`, the scale was just noise (1/epsilon basically). It was proportional to the normal vector component which made no sense

image



Increasing epsilon by a factor of 10 in this PR at least makes it make sense, this now is proportional to one of the rotmat things which does not really matter anyways as that is always fixed in optimization. There is no more random numerical noise here too



image

tests/test_curves.py
```diff
@@ -552,9 +552,9 @@ def test_rotation(self):
         datax = cx.compute("x", grid=20, basis="xyz")
         datay = cy.compute("x", grid=20, basis="xyz")
         dataz = cz.compute("x", grid=20, basis="xyz")
-        np.testing.assert_allclose(datax["x"][:, 0], 0, atol=2e-16)  # only in Y-Z plane
-        np.testing.assert_allclose(datay["x"][:, 1], 0, atol=2e-16)  # only in X-Z plane
-        np.testing.assert_allclose(dataz["x"][:, 2], 0, atol=2e-16)  # only in X-Y plane
+        np.testing.assert_allclose(datax["x"][:, 0], 0, atol=5e-16)  # only in Y-Z plane
```
dpanici, line 555: relaxed this bc 2e-16 was a too tight tolerance. The new method for rotations has a few more arithmetic operations than the old which just pushed it past the 2e-16 threshold (it was failing at like 2.4e-16 which is well within machine precision).

desc/utils.py
```diff
@@ -1001,13 +1010,86 @@ def rotation_matrix(axis, angle=None):
...
+    Computes a rotation matrix that rotates vector u onto vector v.
+
+    Numerically stable when vectors are parallel or anti-parallel, and avoids
+    both NaN gradient when parallel/antiparallel, and also avoids zero
+    gradient (wrt v)
+
+    Parameters
+    ----------
+    u : array-like, shape(3,)
+        first vector, to be rotated onto second vector v
```
YigitElma, line 1031: These need to be in cartesian coordinates, right?

desc/utils.py
```diff
@@ -1001,13 +1010,86 @@ def rotation_matrix(axis, angle=None):
...
+    Returns
+    -------
+    rotmat : ndarray, shape(3,3)
+        Matrix to rotate points in cartesian (X,Y,Z) coordinates.
+
+    """
+    u = safenormalize(u)
+    v = safenormalize(v)
+
+    # 2. Compute the dot product
```
YigitElma, line 1044:
```suggestion
    # Compute the dot product
```

desc/optimize/utils.py
```diff
@@ -518,7 +518,7 @@ def compute_hess_scale(H, prev_scale_inv=None):
     """Compute scaling factors based on diagonal of Hessian matrix."""
     scale_inv = jnp.abs(jnp.diag(H))
     scale_inv = jnp.where(
-        scale_inv < jnp.finfo(H.dtype).eps * max(H.shape), 1, scale_inv
+        scale_inv < jnp.finfo(H.dtype).eps * max(H.shape) * 1e1, 1, scale_inv
```
YigitElma, line 521: What if we have something like,
```python
eps = jnp.maximum(jnp.finfo(H.dtype).eps * max(H.shape), 1e-12)
scale_inv = jnp.where(scale_inv < eps, 1, scale_inv)
```
1e-12 or something else can be chosen as the minimum threshold.

desc/geometry/core.py
```diff
@@ -190,7 +190,7 @@ def translate(self, displacement=[0, 0, 0]):
 
     def rotate(self, axis=[0, 0, 1], angle=0):
         """Rotate the curve by a fixed angle about axis in X,Y,Z coordinates."""
-        R = rotation_matrix(axis=axis, angle=angle)
+        R = rotation_matrix(axis=np.asarray(axis).astype(float), angle=angle)
```
YigitElma, line 193: Are these really needed? I would say that if `axis` stayed as an integer array, it would actually have better accuracy?!

desc/utils.py
```diff
@@ -1001,13 +1010,86 @@ def rotation_matrix(axis, angle=None):
...
+            ],
+            [
+                2 * (qx * qz - qw * qy),
+                2 * (qy * qz + qw * qx),
+                1 - 2 * (qx**2 + qy**2),
+            ],
+        ]
+    )
+    # where to return antiparallel R if needed, else normal R
+    return jnp.where(jnp.allclose(dot, -1.0), R_antiparallel, R)
```
YigitElma: Default rtol=1e-5 and atol=1e-8 here. Do the tests hit somewhere in between?

YigitElma (commented):
Couple nitpicks. It looks good but I am not sure if the edge cases still exist.

desc/geometry/core.py
dpanici: the issue comes later if we then try to grab its precision inside of rotation_matrix, it throws an error bc int is not a floating point dtype

f0uriest: shouldn't this use jnp

desc/optimize/utils.py
f0uriest: what were the cases this helped? another possible principled fix would be to scale based on the max of the diagonal rather than just the size.

dpanici: The case this had issue with was cases where the hessian was ~0 but not exactly zero (like 1e-14 level). So doing max of diagonal would work too I guess, do you mean something like

`scale_inv < jnp.max(scale_inv) , 1, scale_inv` or something?

dpanici: @f0uriest not sure if I understand your point here?

PR #2281, opened 2026-08-04
Remove large arrays before big memory operations

YigitElma:
This is a continuation of #1688. 

- When we compute the new Jacobian, the old one is still in memory. This historically caused the second Jacobian computation to have a larger peak. This PR deletes the old Jacobian and matrix decomposition-related stuff before the new computation.
- Replaces the remaining non-in-place J *= d with a jitted `scale_columns` that donates its buffer, so the scaled matrix is written into J's memory; the old reference is deleted immediately. Some of these were noted in [#1669 (comment)](https://github.com/PlasmaControl/DESC/issues/1669#issuecomment-2851084771).
- Operations like `g = jnp.dot(J.T, f)` create a copy of `J`. Since `f` is 1D, we can just compute `g = jnp.dot(f.T, J)` or equivalently `g = jnp.dot(f, J)` without explicitly forming `J.T`.
- The final J_h * 1 / d made two out-of-jit copies (J_h * 1, then / d). It now uses scale_columns too.

Some of these do not affect the peak memory, i.e. `jnp.dot(J.T, f)` change, but these are one-liners and can be useful in the future if they become the memory bottleneck. I also thought about updating `select_step` to take in `J.T@J` by using the decomposition stuff, but that is a bigger change and doesn't reduce peak memory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Performance Improvements
+
+- Improves memory management to reduce the base memory used during optimization while using `lsq-exact`, `lsq-auglag` and `fmin-auglag` optimizers.
+
+
+v0.17.2
+-------
+
 New Features
 
 - Adds ``eq_fixed`` argument to ``BoundaryError`` to remove the equilibrium from the optimization. This can be used instead of adding a ``FixParameter(eq)`` constraint.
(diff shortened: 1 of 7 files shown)


YigitElma:
For historical comparison, here are the same profiles from https://github.com/PlasmaControl/DESC/pull/1688#issuecomment-2803102758

- Current master vs this PR (`jac_chunk_size=500`)
memory-compare-gpu

- Current master vs this PR (`jac_chunk_size=1`)
memory-compare-gpu

Note: I was sampling memory at around 7kHz; this should resolve almost all peaks.

<details>
<summary>SVD/Cholesky</summary>

- SVD (`jac_chunk_size=500`)
memory-compare-gpu


- Cholesky (`jac_chunk_size=500`)
memory-compare-gpu


</details>

ddudt (approved):

dpanici (approved):

Issue #1952, opened 2025-10-02
Revert change to `plot_surfaces` `map_coordinates` call once toroidal angle is generalized


dpanici:
> Change this back once #568 is in 

 _Originally posted by @dpanici in [#1951](https://github.com/PlasmaControl/DESC/pull/1951/files#r2400152057)_

YigitElma:
This is already in the PRs todo list. Closing for cleaning purposes. The number of open issues makes it hard to see which ones are actually important. I would prefer to list these kind of todos only in the PR description.

PR #2282, opened 2026-08-04
Generalized toroidal angle (omega): re-derived against current master

### LINKED ISSUE #465: Generalize toroidal angle beyond phi cylindrical

rahulgaur104:
Parallel PR to Rory's generalized toroidal angle PR.
Here are the important point from PR #568 that have been addressed here
- [x] constant offset surface
- [x] Proximal projection
- [x] Default to fixing zeta==phi? Default to an empty W_lmn, modes_W.
- [x] to_FourierRZCurve methods should now be possible?
- [x] Fix calculation of A, a_major/a_minor, other stuff that needs to be computed at const phi, not const zeta. (this should already be done in other prs)
- [x] changing continuation/perturbation to work with the new coordinates (changing regularization)
- [x] make the new API compatible with this PR
 different resolutions for omega?
- [x] Revert change to plot_surfaces map_coordinates call once toroidal angle is generalized   https://github.com/PlasmaControl/DESC/issues/1952
- [x] Resolves https://github.com/PlasmaControl/DESC/issues/465
- [x] Add more tests to increase coverage. Maybe a tutorial?

I will add some results related to racetrack configurations soon. Results will include 

- [x]  ForceBalace solve with and without a generalized toroidal angle for a racetrack axis shape stellarator
- [x]  Comparison with the GVEC equilibrium for the same racetrack boundary shape 
- [x]  Optimization of a racetrack stellarator

### Future work
 ☐ near axis constraints with omega (currently the axis guess with the modified zeta works well)
 ☐ Add stuff for spectral condensation / optimal fitting (currently depends strongly on your initialization of omega).
Note: This is highly dependent on the geometry and should be addressed in a future PR. Finding the optimal omega for an arbitrary shape will definitely increase the size of this PR significantly (@f0uriest knows more). Without this feature, omega has to be supplied by the user.

### Test and figures: Normalized force error for a racetrack stellarator-mirror hybrid with and w/o $\omega$

Four cross-sections at equally spaced **lab** angles `phi` (not `zeta`) —
`plot_section` inverts `phi = zeta + omega`, so both rows are cut at the same
physical planes. Both rows share one colour scale.

| | `omega != 0` | `omega = 0` | ratio |
|---|---|---|---|
| max \|F\| normalized (ForceBalance objective) | **7.217e-03** | **1.375e-01** | 19.1x |
| avg \|F\| normalized (ForceBalance objective) | 9.344e-05 | 3.088e-03 | 33.0x |
| max \|F\|_normalized (volume grid) | 5.008e-02 | 1.005e+00 | 20.1x |
| mean \|F\|_normalized (volume grid) | 9.467e-04 | 3.037e-02 | 32.1x |
| boundary fit residual vs traced data | 4.298e-03 m | 1.325e-02 m | 3.1x |
| `omega` modes | 1268 | 0 | |
| `dim_x` | 10466 | 8850 | +18.3 % |
| nested | yes | yes | |

boundary_B
force_sections

You can see how much the generalized angle helps calculate the racetrack equilibrium.

### Comparison with GVEC
iota comparison with GVEC. The match is not great because the GVEC and DESC boundary are not identical and GVEC was run with a low resolution so it may not be fully converged
iota_vs_gvec

Summary by Claude.
```
Generalizes DESC so the computational toroidal coordinate `zeta` no longer has to
equal the cylindrical laboratory angle `phi`, by introducing a periodic toroidal
stream function `omega`:

phi(rho, theta, zeta) = zeta + omega(rho, theta, zeta)

stored spectrally as `W_lmn` (`W_basis`), with `Wb_lmn` on the boundary and `Wa_n`
on the axis. **`omega = 0` remains the default everywhere** and reproduces the
current behavior exactly.

Resolves #465. This is an alternative to #568, re-derived against current master
rather than continuing that branch (see "Relationship to #568" below).

## Approach

DESC already routed `phi`, the covariant/contravariant basis vectors, the
Jacobian, metric tensors, curvature, normals, the magnetic field and force
balance through `omega` — the registrations existed as zero placeholders. The
core of this PR connects those 30 registrations to real `W` transforms, which
generalizes the whole compute graph at once. No metric formula was hand-patched.

Highlights:

- `FourierRZToroidalSurface`: `W_lmn`/`W_basis` state with independent `Mz`/`Nz`,
  and `from_values` generalized to fit `omega = phi - zeta` when a `zeta` is
  supplied (accepts `(R, phi, Z)` or `(X, Y, Z)`).
- `FourierRZCurve`: `W_n` so the magnetic axis can carry a generalized angle;
  `x`, `x_s`, `x_ss`, `x_sss` re-derived for `phi = s + W(s)`.
- `Equilibrium`: `W_lmn` with `Lz`/`Mz`/`Nz`, save/load migration, initial guess,
  boundary/axis extraction, `is_nested`.
- Constraints: `BoundaryWSelfConsistency`, `AxisWSelfConsistency`, `FixBoundaryW`,
  `FixAxisW`, `FixOmegaGauge`, `FixOmegaInterior`, `FixZetaSFL`, wired into the
  getters, proximal projection and perturbations.
- New theory page: `docs/theory_generalized_toroidal_angle.rst`.

## No `.h5` regeneration

Adding `_W_lmn` to `_io_attrs_` makes DESC warn on every pre-existing `.h5` file,
which is an error under `filterwarnings = error`. Rather than regenerating the
bundled equilibria, this PR adds `_io_attrs_optional_` to `desc/io/hdf5_io.py`: a
per-class allowlist of attributes that may be absent from an older file without
warning, for attributes whose `_set_up` assigns a default reproducing the old
behavior. **No `.h5` files are modified in this PR.**

## Relationship to #568

Same goal and the same naming conventions (`W_lmn`, `W_basis`, `Wb_lmn`, `Wa_n`),
but re-derived against current master; that branch is 144 commits behind and the
two implementations conflict in 127 places, so this is a fresh branch rather than
a continuation. Deliberate differences:

- New constructor arguments are **keyword-only and appended**, so every existing
  positional call still works (#568 inserted them mid-signature).
- No `.h5` regeneration (see above).
- `change_resolution` `Lz`/`Mz`/`Nz` default to `None` and are preserved, rather
  than defaulting to `0` and silently wiping `omega` state.
- `omega = phi - zeta` is computed branch-cut-safely, plus the chart-validity check.
- `ZernikeRZToroidalSection` gets an explicit read-only `omega = 0` stub.
- `FixBoundaryW` is not normalized by `scales["a"]` (`omega` is an angle).
- `get_coeffs`/`set_coeffs` signatures are unchanged.
- `omega` constraints are added only when `omega` degrees of freedom exist.

Also handles master's `OMEGA_IS_0` markers: the closed-form `alpha -> theta_PEST`
inversion and the PEST fast path assume `phi == zeta`; with `omega != 0`, `omega`
depends on `theta` (the unknown being solved for) so they no longer decouple.
They are guarded and fall through to the general Newton solve, which handles
`alpha` and `theta_PEST` correctly since both are defined via `phi`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)
```


YigitElma:
I will try to review this soon, but I have a question about this plot. 
image

Am I being visually tricked or these 2 equilibria have slightly different LCFS? Can you maybe use `plot_comparison` to be sure?

rahulgaur104:
> I will try to review this soon, but I have a question about this plot. image
> 
> Am I being visually tricked or these 2 equilibria have slightly different LCFS? Can you maybe use `plot_comparison` to be sure?

Thanks, Yigit! Yeah, no hurry. I want to add more tests and get the coverage up before calling it ready.
Yes, they have slightly different boundaries. Both the methods used from_values to read raw point data but the generalized angle branch is supplied the data in Boozer coordinates and master takes the data in the cylindrical coordinate. Also, the resolution values supplied to from_values were not exactly the same.

I'll ensure that the boundaries are identical but I don't think that will significantly change the result.

rahulgaur104:
@YigitElma  Here's the new boundary comparison and ForceBalance calculation

boundary_compare_12
force_sections_6


Both the boundaries are solved at the same L=20, M=10, N=20 resolution.

desc/compute/_geometry.py
```diff
@@ -18,6 +18,25 @@
 from ..utils import cross, dot, safenorm
 from .data_index import register_compute_fun
 
+# Shared note for the constant-zeta cross-section family.  These are defined on
```
f0uriest, line 21: We should discuss the best way to handle this stuff. A few thoughts:
- Right now we have a few different ways of computing eg, `A(z)`, we should measure how much they vary on equilibria with and without omega, and possibly enforce that only one method gets used when omega !=0 so at least things are self consistent (and eg using a slightly different grid doesn't give you vastly different answers)
- In some cases we may want to maintain current behavior by enforcing that things be computed only on phi=const planes, i think there's a note about this somewhere in this file

desc/equilibrium/coords.py
```diff
@@ -122,7 +122,17 @@ def map_coordinates(  # noqa: C901
...
     # TODO (#1382): make this work for permutations of in/out basis
-    if outbasis == ("rho", "theta", "zeta"):
+    # The closed form inversions below assume the toroidal angle appearing in
+    # the field line label and the PEST mapping is the computational
+    # coordinate zeta, ie that omega = 0 so phi = zeta. With a generalized
+    # toroidal angle, omega depends on theta, which is the very unknown being
+    # solved for, so those inversions no longer decouple; in that case we fall
+    # through to the general Newton solve, which handles theta_PEST and alpha
+    # (both defined via phi) correctly.
+    omega_is_0 = OMEGA_IS_0 and (
```
f0uriest, line 132: I think this PR can remove the `OMEGA_IS_0` constant since it was meant to indicate that omega is everywhere identically zero which is no longer true

f0uriest: we may also want to add a shortcut for mapping to pest coordinates when omega is nonzero since its a common use case and there's probably some tricks to make it efficient

desc/equilibrium/equilibrium.py
```diff
@@ -156,6 +156,12 @@ class Equilibrium(IOAble, Optimizable):
         If True, and the default initial guess does not produce nested surfaces,
         run a small optimization problem to attempt to refine initial guess to improve
         coordinate mapping.
+    Lz, Mz, Nz : int (optional)
```
f0uriest: I'd prefer Lw, Mw, Nw etc for the omega resolution since we use W/w elsewhere for it

desc/equilibrium/equilibrium.py
```diff
@@ -207,7 +215,15 @@ class Equilibrium(IOAble, Optimizable):
         "_R_basis",
         "_Z_basis",
         "_L_basis",
+        "_W_basis",
+        "_Lz",
+        "_Mz",
+        "_Nz",
     ]
+    # equilibria saved before the generalized toroidal angle existed have no
```
f0uriest, line 237: I'd prefer not adding this optional part. `_set_up` in many cases today is designed to make things backwards compatible but we still throw a warning to let users know. If anything its advertising for new capabilities

desc/equilibrium/equilibrium.py
```diff
@@ -310,8 +329,31 @@ def __init__(
...
+        self._Mz = int(setdefault(Mz, getattr(self.surface, "Mz", 0)))
+        self._Nz = int(setdefault(Nz, getattr(self.surface, "Nz", 0)))
+        self._Lz = int(
+            setdefault(
+                Lz,
+                self.Mz if (self.spectral_indexing == "ansi") else 2 * self.Mz,
+            )
+        )
+
+        if hasattr(self._surface, "Mz"):
```
f0uriest: shouldn't every surface have an `Mz` attribute?

desc/equilibrium/equilibrium.py
```diff
@@ -338,6 +380,19 @@ def __init__(
             sym=self._Z_sym,
             spectral_indexing=self.spectral_indexing,
         )
+        # omega has the same (sin) stellarator symmetry parity as Z and lambda.
+        # Zero omega resolution means no generalized toroidal angle, so the basis
+        # must be empty: a non-symmetric basis at Lz=Mz=Nz=0 would still carry
+        # the (0,0,0) mode and give every asymmetric equilibrium a spurious
+        # degree of freedom. The symmetric case is already empty, sin(0) being 0.
+        self._W_basis = FourierZernikeBasis(
```
f0uriest, line 404: if Z is nonsymmetric then this will still have the 0,0,0 mode. I thought @ddudt added a symmetry option for "full without constant"?

desc/equilibrium/equilibrium.py
```diff
@@ -658,14 +761,26 @@ def change_resolution(
             if hasattr(p, "change_resolution"):
                 p.change_resolution(max(p.basis.L, self.L))
 
-        self.surface.change_resolution(
-            self.L, self.M, self.N, NFP=self.NFP, sym=self.sym
-        )
-        self.axis.change_resolution(self.N, NFP=self.NFP, sym=self.sym)
+        if hasattr(self.surface, "Mz"):
```
f0uriest: see previous comment

desc/equilibrium/equilibrium.py
```diff
@@ -1549,7 +1716,17 @@ def surface(self, new):
             self.sym == new.sym
         ), "Surface and Equilibrium must have the same symmetry"
         assert self.NFP == new.NFP, "Surface and Equilibrium must have the same NFP"
-        new.change_resolution(self.L, self.M, self.N)
+        if hasattr(new, "Mz") and ((new.Mz > self.Mz) or (new.Nz > self.Nz)):
```
f0uriest: this differs from how we treat the other coordinates and existing surfaces, is there a reason?

desc/equilibrium/equilibrium.py
```diff
@@ -2154,6 +2391,12 @@ def resolution_summary(self):
         """Print a summary of the spectral and real space resolution."""
         print("Spectral indexing: {}".format(self.spectral_indexing))
         print("Spectral resolution (L,M,N)=({},{},{})".format(self.L, self.M, self.N))
+        if self.W_basis.num_modes:
```
f0uriest, line 2444: I think we should always print this for completeness

desc/geometry/curve.py
```diff
@@ -39,6 +39,11 @@
 class FourierRZCurve(Curve):
     """Curve parameterized by Fourier series for R,Z in terms of toroidal angle phi.
 
+    When a nonzero ``W_n`` is given, the curve parameter ``s`` is a generalized
```
f0uriest, line 42: should update the previous line as well

desc/geometry/curve.py
```diff
@@ -53,19 +58,35 @@ class FourierRZCurve(Curve):
...
+    _static_attrs = Curve._static_attrs + [
+        "_sym",
+        "_NFP",
+        "_R_basis",
+        "_Z_basis",
+        "_W_basis",
+    ]
+    # curves saved before the generalized toroidal angle existed have no omega
+    # state; _set_up defaults it to zero (phi = s)
+    _io_attrs_optional_ = ["_W_n", "_W_basis"]
```
f0uriest, line 89: see previous comment

desc/geometry/curve.py
```diff
@@ -76,43 +97,86 @@ def __init__(
...
             raise ValueError("At least 1 coefficient for R must be supplied")
         if Z_n.size == 0:
             Z_n = np.array([0.0])
             modes_Z = np.array([0])
 
-        modes_R, modes_Z = np.asarray(modes_R), np.asarray(modes_Z)
+        modes_R, modes_Z, modes_W = (
+            np.asarray(modes_R),
+            np.asarray(modes_Z),
+            np.asarray(modes_W, dtype=int),
```
f0uriest, line 128: shouldnt they all be ints?

desc/geometry/curve.py
```diff
@@ -76,43 +97,86 @@ def __init__(
...
         N = max(NR, NZ)
         self._NFP = check_posint(NFP, "NFP", False)
         self._R_basis = FourierSeries(N, int(NFP), sym="cos" if sym else False)
         self._Z_basis = FourierSeries(N, int(NFP), sym="sin" if sym else False)
+        # No omega modes means no generalized toroidal angle, so the basis must be
+        # empty. A non-symmetric FourierSeries at N=0 still carries the n=0 mode,
+        # which would give every asymmetric curve -- and the coils that subclass
+        # them -- a spurious omega degree of freedom.
+        self._W_basis = FourierSeries(
+            NW, int(NFP), sym="sin" if (sym or not modes_W.size) else False
```
f0uriest, line 158: See previous note about symmetry=full without constant option. 

We could also just allow the constant mode but fix it to zero in the gauge constraints, that might be cleaner (similar to what we do for lambda)

desc/io/hdf5_io.py
```diff
@@ -120,10 +120,23 @@ def read_obj(self, obj, where=None):
         where : None or file instance
             specifies where to read obj from
 
+        Notes
+        -----
+        Attributes listed in the class attribute ``_io_attrs_optional_`` are
```
f0uriest, line 125: see previous note, I'm not crazy about this

desc/objectives/getters.py
```diff
@@ -353,15 +375,29 @@ def add_if_multiple(constraints, cls):
     if {"Z_lmn", "Zb_lmn"} <= params:
         constraints = add_if_multiple(constraints, BoundaryZSelfConsistency)
 
+    # omega constraints are only needed when the equilibrium actually has
+    # omega degrees of freedom; the default omega=0 (empty basis) case adds
+    # nothing to the optimization
+    has_omega = getattr(thing, "W_basis", None) is not None and thing.W_basis.num_modes
```
f0uriest: see previous comment, it might be cleaner to just give everything a W_basis and always fix the extra coefficients here

desc/objectives/linear_objectives.py
```diff
@@ -661,6 +661,119 @@ def compute(self, params, constants=None):
...
+        self._A = np.zeros((self._dim_f, eq.W_basis.num_modes))
+        Js = []
+        surf = eq.surface.rho if self._surface_label is None else self._surface_label
+        for i, (l, m, n) in enumerate(eq.W_basis.modes):
+            if eq.bdry_mode == "lcfs":
+                j = np.argwhere((modes[:, 1:] == [m, n]).all(axis=1))
+                Js.append(j.flatten())
+            else:
+                raise NotImplementedError(
+                    "bdry_mode is not lcfs, yell at Dario to finish poincare stuff"
```
f0uriest: this warning is out of date should now be "yell at @YigitElma to finish poincare stuff" 😆

desc/objectives/linear_objectives.py
```diff
@@ -1062,6 +1339,116 @@ def __init__(
         )
 
 
+class FixOmegaGauge(FixParameters):
+    """Fixes gauge freedom of the generalized toroidal angle.
+
+    For a fixed physical field, the computational toroidal angle may be shifted
+    per flux surface, zeta -> zeta + c(rho), with omega -> omega - c(rho),
```
f0uriest: I think any function of rho,theta is also a gauge freedom ie c(rho, theta) not just c(rho), so we may need to expand this

desc/objectives/linear_objectives.py
```diff
@@ -1062,6 +1339,116 @@ def __init__(
...
+            thing=eq,
+            params={"W_lmn": indices},
+            target=0,
+            normalize=normalize,
+            normalize_target=normalize_target,
+            name=name,
+        )
+
+
+class FixOmegaInterior(FixParameters):
```
f0uriest, line 1392: We should check if this is still necessary. As noted above I think the existing gauge constraint isnt complete which might be why things were getting weird without fixing omega entirely. With the proper gauge I think it should be similar to omega (a null direction of force balance in infinite resolution but helpful at finite resolution)

desc/objectives/linear_objectives.py
```diff
@@ -1062,6 +1339,116 @@ def __init__(
...
+    """Fixes the interior (higher radial mode) omega degrees of freedom.
+
+    For each poloidal/toroidal mode pair (m, n) of the omega basis, all radial
+    modes except the one with the lowest radial degree are fixed at their
```
f0uriest, line 1396: We may want to check if this is the best way to enforce this. We have 2 constraints on omega, this and the self consistency, so how you split that doesn't affect the math but does affect the numerics, and a bad choice may make the linear constraint projection more ill-conditioned. Self consistency fixes a sum, while this fixes individual dofs, leading the single unfixed one to compensate for all the others. Instead fixing an orthogonal sum might be better.

desc/plotting.py
```diff
@@ -1563,6 +1563,142 @@ def plot_fsa(  # noqa: C901
     return fig, ax
 
 
+def _phi_to_zeta_bisect(eq, nodes, niter=48):
```
f0uriest: bisection converges only logarithmically. Bracketed newton (basically what we do for the trust region subproblem) would likely be faster while avoiding incorrect branches

desc/plotting.py
```diff
@@ -1563,6 +1563,142 @@ def plot_fsa(  # noqa: C901
...
+    for _ in range(niter):
+        mid = 0.5 * (lo + hi)
+        too_big = phi_at(mid) > phi_target
+        hi = np.where(too_big, mid, hi)
+        lo = np.where(too_big, lo, mid)
+    nodes[:, 2] = 0.5 * (lo + hi)
+    return nodes
+
+
+def _pest_phi_to_zeta(eq, nodes, niter=12, tol=1e-10):
```
f0uriest, line 1613: if this is fast and robust we may want to make this a special case of map_coordinates for the omega != 0 case, but I worry that map coordinates is overhead dominated so multiple calls may be slower. Might be easier just to add box constraints to the newton solve we use to prevent branch jumps?

desc/plotting.py
```diff
@@ -1563,6 +1563,142 @@ def plot_fsa(  # noqa: C901
...
+        if np.nanmax(np.abs(zeta - zeta_prev)) < tol:
+            break
+
+    return np.stack([rho, theta, zeta], axis=1)
+
+
+def _find_failed_phi_inversion(eq, grid, phi_target, atol=1e-6):
+    """Flag points whose (rho, theta, zeta) do not reproduce the requested phi.
+
+    A Newton inversion of phi = zeta + omega started from ``zeta = phi`` can
```
f0uriest, line 1671: maybe im misunderstanding here, but shouldn't the worst case be that we get phi+2pi*k instead of phi? and that would be the same physical point, so what does this mean that it can converge to something outside the plasma?

desc/plotting.py
```diff
@@ -1970,38 +2101,44 @@ def plot_surfaces(eq, rho=8, theta=8, phi=None, ax=None, return_data=False, **kw
...
-                inbasis=["rho", "theta_PEST", "zeta"],
-                outbasis=["rho", "theta", "zeta"],
-                period=(np.inf, 2 * np.pi, 2 * np.pi),
-                guess=t_grid.nodes,
-                maxiter=30,
-            ),
-            sort=False,
-        )
+        t_grid_phi_target = t_grid.nodes[:, 2].copy()
+        # Resolves TODO(#568) for this caller.  The requested toroidal values
```
f0uriest: not sure this comment is useful

tests/test_generalized_toroidal.py
```diff
@@ -0,0 +1,1067 @@
...
+            np.testing.assert_allclose(
+                d[f"phi_{s}"],
+                (1 if s == "z" else 0) + d[f"omega_{s}"],
+                atol=1e-14,
+                err_msg=s,
+            )
+        assert np.max(np.abs(d["omega"])) > 0  # actually nonzero
+
+    @pytest.mark.unit
+    def test_omega_derivatives_vs_finite_differences(self):
```
f0uriest, line 338: i think this is basically already covered in test_transform, test_basis etc.

But testing higher order stuff vs finite differences (to check that the analytic derivatives in `_compute` are correct) would be good, ideally something similar to the stuff in `test_compute_funs` like `test_magnetic_field_derivatives`, `test_metric_derivatives` etc but with a nonzero stream function

tests/test_generalized_toroidal.py
```diff
@@ -0,0 +1,1067 @@
...
+            ("omega_r", (r + eps, t, z), (r - eps, t, z)),
+            ("omega_t", (r, t + eps, z), (r, t - eps, z)),
+            ("omega_z", (r, t, z + eps), (r, t, z - eps)),
+        ]:
+            fd = (w(*plus) - w(*minus)) / (2 * eps)
+            np.testing.assert_allclose(d[key], fd, rtol=1e-5, atol=1e-8, err_msg=key)
+
+    @pytest.mark.unit
+    def test_compute_everything_nonzero_omega(self):
+        """All standard rtz quantities stay finite with nonzero omega."""
```
f0uriest, line 396: I think we should be checking more than finite-ness.
Ideally we'd take the same simple but nontrivial equilibrium and parameterize it both with and without the stream function, then compute everything at shared points in R,phi,Z and compare, to make sure all of the formulas in the compute functions are correct.

f0uriest (commented):
Coarse pass for now, will look in more detail and try it out later

desc/io/hdf5_io.py
rahulgaur104: If we don't use io_attrs_optional, we have to add W_lmn to each equilibrium desc/examples or tests/inputs even when it's zero. Do you think that's a better idea?

desc/equilibrium/equilibrium.py
rahulgaur104: The options I see are `sin, cos, even, cos(t), "no n=0", False, None`. There's no "full basis minus the single constant mode" option anywhere. "no n=0" is not it either because it strips every n=0 mode (including m≠0, n=0 ones), not just the single (l=0,m=0,n=0) constant. 
Am I missing something?

desc/objectives/linear_objectives.py
rahulgaur104: I checked my driver script again and I don't use this constraint during force balance or optimization. I only fix the Omega on the boundary and freeing the omega inside the volume reduce the max |F|_normalized roughly by a factor of two

desc/objectives/linear_objectives.py
rahulgaur104: Comment above. Default is to free the Omega in the volume but fix on the boundary. For the case I ran, it gave better results and shouldn't cause the inconsistency between different constraints.

PR #2284, opened 2026-08-06
Add warning for batched mode when a sub objective has rev mode

### LINKED ISSUE #2243: Fix deriv mode = batched
If you set `ObjectiveFunction(objectives, deriv_mode="batched")` then it will use `deriv_mode="fwd"` for all of the sub-objectives.  This is not efficient for objectives that are more performant with reverse mode; and therefore, deriv mode = batched should support using the deriv mode of the subobjectives.

dpanici:
Does not resolve #2243 but at least warns users of the situation occuring

I am not sure if we can do batched as not fwd mode: maybe we can do it as like batched all at once but in reverse mode if all sub-objectives are rev mode.
Either way this will just throw a warning in the case that batched is set as ObjectiveFunction deriv_mode but any sub objective has rev mode, since that indicates a probable inefficiency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,18 @@
 Changelog
 =========
 
+New Features
+
+- Added warning for when ``deriv_mode="batched"`` is used in an ``ObjectiveFunction`` where one or more sub-objectives is using ``rev`` mode differentiation. Also adds more info about the derivative mode and Jacobian chunk sizes when building the objective with ``verbose>1``.
+
 Performance Improvements
 
 - Improves memory management to reduce the base memory used during optimization while using `lsq-exact`, `lsq-auglag` and `fmin-auglag` optimizers.
 
-
 Bug Fixes
 
 - Fixes bug in ``auglag`` optimizers which prevented them from accepting solver hyperparameters.
+
 
 v0.17.3
 -------
@@ -18,7 +22,6 @@ New Features
 - Adds ``eq_fixed`` argument to ``BoundaryError`` to remove the equilibrium from the optimization. This can be used instead of adding a ``FixParameter(eq)`` constraint.
 - Adds `check_intersection` argument to `initialize_modular_coils`, `initialize_helical_coils` and `initialize_saddle_coils`
 - Default value of `check_intersection` for coil related functions now defaults to False (no check). Previously, the default was True, and this was causing redundant checks.
-- Added warning for when ``deriv_mode="batched"`` is used in an ``ObjectiveFunction`` where one or more sub-objectives is using ``rev`` mode differentiation. Also adds more info about the derivative mode and Jacobian chunk sizes when building the objective with ``verbose>1``.
 
 Performance Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ New Features
 - Adds ``eq_fixed`` argument to ``BoundaryError`` to remove the equilibrium from the optimization. This can be used instead of adding a ``FixParameter(eq)`` constraint.
 - Adds `check_intersection` argument to `initialize_modular_coils`, `initialize_helical_coils` and `initialize_saddle_coils`
 - Default value of `check_intersection` for coil related functions now defaults to False (no check). Previously, the default was True, and this was causing redundant checks.
+- Added warning for when ``deriv_mode="batched"`` is used in an ``ObjectiveFunction`` where one or more sub-objectives is using ``rev`` mode differentiation. Also adds more info about the derivative mode and Jacobian chunk sizes when building the objective with ``verbose>1``.
 
 Performance Improvements
 
--- a/desc/objectives/objective_funs.py
+++ b/desc/objectives/objective_funs.py
@@ -511,6 +511,17 @@ class ObjectiveFunction(IOAble):
             else:
                 self._deriv_mode = "blocked"
 
+        warnif(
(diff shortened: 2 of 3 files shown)


desc/objectives/objective_funs.py
```diff
@@ -511,6 +511,17 @@ def build(self, use_jit=None, verbose=1):
...
 
+        warnif(
+            any((obj._deriv_mode == "rev") for obj in self.objectives)
+            and self._deriv_mode == "batched",
+            UserWarning,
+            "Using batched derivative mode, which performs fwd mode "
+            "differentiation of the entire objective, but one or more of the"
+            ' sub-objectives have deriv_mode="rev". It is recommended to use '
+            'deriv_mode="blocked", as these objective performances may suffer '
+            "in fwd mode, or may not even support reverse mode at all.",
```
YigitElma: The warning message is not very clear. I would prefer printing the deriv modes of sub-objectives here, since in general users don't choose the deriv modes but it is assigned based on the dim_x/dim_f of the sub-objective. For example in the failing test, `AspectRatio` defaults to reverse mode but for a user it is unknown why this happened.

Something like this can work maybe,
```python
rev_objs = [
    o.__class__.__name__ for o in self.objectives if o._deriv_mode == "rev"
]
warnif(
    len(rev_objs) > 0 and self._deriv_mode == "batched",
    UserWarning,
    "'batched' deriv_mode differentiates the whole ObjectiveFunction in "
    "forward mode, but these sub-objectives are set to use reverse mode "
    "(either automatically, from their input/output sizes, or by user): "
    f"{rev_objs}. In forward mode these may under-perform, or may not "
    "work at all. Consider 'blocked' deriv_mode. See the sub-objective "
    "docstrings for details.",
)
```

YigitElma (changes requested):
The message was not very clear, and there was a typo `or may not even support reverse mode at all` should be forward mode. 

Also, while we are at it, would we like to print something like this at the end?
```python
if verbose > 1:
    print(f"{self.name} deriv_mode : {self._deriv_mode}")
    if self._deriv_mode == "batched":
        print(f"{self.name} jac_chunk_size: {self._jac_chunk_size}")
    else:
        for o in self.objectives:
             print(f"{o.name} jac_chunk_size: {o._jac_chunk_size}")
    timer.disp("Objective build")
```

Should also fix the failing test

desc/objectives/objective_funs.py
dpanici: Thanks, sorry yea it should be fwd mode and yea I think it is a good idea to clearly show what all the sub objectivee are doing if blocked is being used, as well as the auto-picked jac chuink size

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ New Features
 - Adds ``eq_fixed`` argument to ``BoundaryError`` to remove the equilibrium from the optimization. This can be used instead of adding a ``FixParameter(eq)`` constraint.
 - Adds `check_intersection` argument to `initialize_modular_coils`, `initialize_helical_coils` and `initialize_saddle_coils`
 - Default value of `check_intersection` for coil related functions now defaults to False (no check). Previously, the default was True, and this was causing redundant checks.
+- Added warning for when ``deriv_mode="batched"`` is used in an ``ObjectiveFunction`` where one or more sub-objectives is using ``rev`` mode differentiation. Also adds more info about the derivative mode and Jacobian chunk sizes when building the objective with ``verbose>1``.
```
YigitElma: We made a new release. Some other PRs will probably add the v0.17.3 title before this, you can fix it then.

YigitElma (commented):
Other then the changelog issue it is fine.

YigitElma (approved):

YigitElma (dismissed):

desc/objectives/objective_funs.py
```diff
@@ -511,6 +511,21 @@ def build(self, use_jit=None, verbose=1):
             else:
                 self._deriv_mode = "blocked"
 
+        rev_objs = [
+            o.__class__.__name__ for o in self.objectives if o._deriv_mode == "rev"
```
ddudt: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
I would prefer `o.name`, since it would be more clear if the user has multiple of the same objectives and gave them unique names. Also because I think we typically use `obj.name` instead of the class name when printing objectives.

YigitElma:
```suggestion
            o.name for o in self.objectives if o._deriv_mode == "rev"
```

YigitElma (dismissed):

desc/objectives/objective_funs.py
dpanici:
```suggestion
            o.__name__ for o in self.objectives if o._deriv_mode == "rev"
```

desc/objectives/objective_funs.py
```diff
@@ -511,6 +511,21 @@ def build(self, use_jit=None, verbose=1):
             else:
                 self._deriv_mode = "blocked"
 
+        rev_objs = [
+            o.__name__ for o in self.objectives if o._deriv_mode == "rev"
+        ]
```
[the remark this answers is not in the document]
ddudt:
```suggestion
        rev_objs = [o.name for o in self.objectives if o._deriv_mode == "rev"]
```

YigitElma (approved):

ddudt (approved):

PR #2285, opened 2026-08-07
`SurfaceCurve` class and related optimization features

singh-jaydeep:
Adds a new `SurfaceCurve` class, representing curves which are constrained to lie on a (toroidal) surface. This PR is meant to integrate the features added in the open PRs #819, #844, and replaces #2081. 

Edit: The actual umbilic curvature objective from #819 was moved to #2321.

Features
- `FourierRZSurfaceCurve` class represents curves on a `FourierRZToroidalSurface` which can link poloidally or toroidally, and may circulate many times before closing. Parameterized using Fourier series for (theta(s), zeta(s)), s in [0,2pi). `FourierRZSurfaceCoil` is the coil version. These curves require either a surface or equilibrium as input. In the latter case, the surface is just the rho=1 flux surface.
- `SurfaceCurveConsistency` is a linear objective which accepts a surface/equilibrium and a `FourierRZSurfaceCurve`, computing the difference between the surface/equilibrium.surface's params and those of the curve. Should be used in optimization when the underlying surface appears across different objectives. This also supports coilsets.

Notes
- ~~For surface curve coils to work with NFP and sym, they need to implement rigid transformations like rotmat and shift. I set them to the identity and zero vector, respectively, but they do not make a lot of sense for surface curves. And since they are optimizable, pretty much all optimizations involving surface curves will need a FixParameters(curve, {"rotmat":True, "shift": True}). Not sure if there is a way around this.~~
- By default, all surface curves have the associated surface R_lmn, Z_lmn parameters as optimizable. So if you are optimizing a surface curve and don't want the surface to vary, users need to add FixParameteres(curve, {"R_lmn": True, "Z_lmn": True}).


singh-jaydeep:
Moving the notebook here: [surface_curve_optimizations.ipynb](https://github.com/user-attachments/files/31534500/surface_curve_optimizations.ipynb)

Here's an html version of the same thing: [surface_curve_optimizations.html](https://github.com/user-attachments/files/31534494/surface_curve_optimizations.html).

YigitElma:
Maybe split the objective related changes to a separate PR

PR #2286, opened 2026-08-10
perf: use jax.linearize when chunking jacobians in forward-mode

jpbrodrick89:
This should provide a runtime and compile time improvement at the possible cost of a very slight peak memory increase. I think if this pushes you over the limit you are probably in the region where you should be switching to iterative rather than direct solvers.

`jax.linearize` caches the residuals from the primal pass just like `jax.vjp` does which means sequentially evaluating (chunks of) jvp's it will be a lot faster. For example, if Proximal was fully jittable and went through this route this would have allowed the reuse of the SVD decomposition automatically. As it stands Proximal does not directly benefit but the improvement @YigitElma is making in #2239 should chain nicely with this (it would be interesting to see the benchmarks for these two PR's combined).

@unalmis this should probably be upstreamed in adv-jax-math, lmk if you'd like me to make a PR there.

PR #2287, opened 2026-08-10
Fix tests for older dependencies

YigitElma:
See https://github.com/PlasmaControl/DESC/pull/2286#issuecomment-5242963900

- I temporarily changed all unit and regression tests to use 3.10, so that all use jax-finufft 1.2.0. Everything passes now.
- Fixes some of the sensitive tests
- Adds a warning for possible rank deficiency during optimization
- Update `master_compute_data.pkl` after the `interpax` 0.3.15 release


ddudt (dismissed):

desc/optimize/utils.py
```diff
@@ -175,6 +175,20 @@ def gershgorin_bounds(H):
     return lb, ub
 
 
+@jit
+def _chol_failed(L):
+    """Whether a Cholesky factor is unusable because the matrix was not positive.
+
+    ``potrf`` only errors out, and so returns nan, when a pivot comes out
+    non-positive. A pivot at the noise level instead completes but gives a factor that
+    is useless to solve with, and which of the two a marginal matrix gets depends on
```
YigitElma, line 184: For example, `test_multiobject_optimization_al` was creating such matrices, failing on CI for #2293, but not locally

ddudt (dismissed):

desc/optimize/_desc_wrappers.py
```diff
@@ -45,6 +46,36 @@
...
+    """Warn if bounds on sub-objectives can make the Jacobian rank-deficient.
+
+    A sub-objective with bounds contributes zero residuals, and hence zero Jacobian
+    rows, whenever it is inside its bounds. The full rank of the (m, n) Jacobian is
+    min(m, n), so those rows can only cost rank if the ones that always remain are
+    fewer than that. Only ``"svd"`` handles the deficient case reliably, so warn if
+    the user hasn't picked a method themselves.
+    """
+    if "tr_method" in options:  # user picked a method, don't second guess it
+        return
```
f0uriest: do we want to change the default based on the number of bounds? and even if not, we may still want to warn even if the user selects qr themselves

f0uriest: alternatively on newer versions of jax i think we can use rank-revealing qr with a stabilized triangular solve? cheaper than svd, but may not be as accurate when its really rank deficient.

desc/optimize/_desc_wrappers.py
```diff
@@ -45,6 +46,36 @@
...
+    warnif(
+        m - dim_f_bounded < min(m, n),
+        UserWarning,
+        f"Objectives {[obj.name for obj in bounded]} use bounds instead of target, so "
+        + f"they can zero out {dim_f_bounded} of the {m} rows of the ({m}, {n}) "
+        + f"Jacobian and drop its rank below {min(m, n)}. The default 'qr' trust "
+        + "region method may then fail to solve the subproblem, in that case pass "
+        + "options={'tr_method': 'svd'}.",
+    )
+
```
YigitElma, line 82: If they passed use it. If they passed qr and it can become singular then warn. If they didn't pass, then use the safe option svd, and don't warn

desc/optimize/_desc_wrappers.py
YigitElma: We can add the rank revealing QR alternative in a later PR. I think our min version of JAX should already support that option.

desc/optimize/_desc_wrappers.py
```diff
@@ -45,6 +46,41 @@
...
+    if m - dim_f_bounded < min(m, n):
+        supplied = hasattr(options, "tr_method")
+        tr_method = options.get("tr_method", "qr")
+        warnif(
+            supplied and tr_method != "svd",
+            UserWarning,
+            f"Objectives {[obj.name for obj in bounded]} use bounds instead of target, "
+            + f"so they can zero out {dim_f_bounded} of the {m} rows of the ({m}, {n}) "
+            + f"Jacobian and drop its rank below {min(m, n)}. The trust region method "
+            + f"{tr_method} may then fail to solve the subproblem, in that case pass "
```
dpanici: Is the case obvious? Like does it nan or does it just fail to improbe objective? Maybe point out how it could pop up (if itis easy to tell)

desc/optimize/aug_lagrangian_ls.py
```diff
@@ -173,7 +173,9 @@ def lsq_auglag(  # noqa: C901
           Cholesky factorizations (generally 2-3), while ``"svd"`` uses one singular
           value decomposition. ``"cho"`` is generally the fastest for large systems,
           especially on GPU, but may be less accurate for badly scaled systems.
-          ``"svd"`` is the most accurate but significantly slower. Default ``"qr"``.
+          ``"svd"`` is the most accurate but significantly slower. If the Jacobian
+          can become rank-deficient due to the bounds, ``"svd"`` is recommended and
+          the default, otherwise the deafult is ``"qr"``.
```
dpanici:
```suggestion
          the default, otherwise the default is ``"qr"``.
```

desc/optimize/_desc_wrappers.py
```diff
@@ -45,6 +46,41 @@
...
+            + "options={'tr_method': 'svd'}.",
```
YigitElma:
```suggestion
            + f"{tr_method} may then fail to solve the subproblem, it is safer to use "
            + "options={'tr_method': 'svd'}.",
```

dpanici (dismissed):

desc/optimize/least_squares.py
```diff
@@ -139,7 +139,10 @@ def lsqtr(  # noqa: C901
           Cholesky factorizations (generally 2-3), while ``"svd"`` uses one singular
           value decomposition. ``"cho"`` is generally the fastest for large systems,
           especially on GPU, but may be less accurate for badly scaled systems.
-          ``"svd"`` is the most accurate but significantly slower. Default ``"qr"``.
+          ``"svd"`` is the most accurate but significantly slower. If the Jacobian
+          can become rank-deficient due to the bounds, ``"svd"`` is recommended and
+          the default, otherwise the deafult is ``"qr"``.
+          since the linear system has a chance to be rank-deficient. Default ``"qr"``.
```
YigitElma:
```suggestion
          the default, otherwise the default is ``"qr"``.
```

desc/optimize/_desc_wrappers.py
```diff
@@ -45,6 +46,41 @@
...
+    bounded = [obj for obj in sub.objectives if obj.bounds is not None]
+    dim_f_bounded = sum(obj.dim_f for obj in bounded)
+    # constraint rows never vanish, bounds there become slack variables instead
+    m = objective.dim_f + (0 if constraint is None else constraint.dim_f)
+    n = x0.size
+    if m - dim_f_bounded < min(m, n):
+        supplied = hasattr(options, "tr_method")
+        tr_method = options.get("tr_method", "qr")
+        warnif(
+            supplied and tr_method != "svd",
```
ddudt: [suggestion](https://pullpo.io/cc?l=suggestion) 
I think this is equivalent and simpler: 
```suggestion
        tr_method = options.get("tr_method", "svd")
        warnif(
            tr_method != "svd",
```

ddudt (commented):
See minor suggestion, otherwise it looks good.

desc/optimize/_desc_wrappers.py
YigitElma: I am fine with making the change, but to be honest, this looks more complicated to me to understand when we want to throw this warning :) I use `supplied` again later.

desc/optimize/_desc_wrappers.py
```diff
@@ -45,6 +46,41 @@
...
+    sub = objective
+    while hasattr(sub, "_objective"):  # unwrap Proximal/LinearConstraintProjection
+        sub = sub._objective
+    bounded = [obj for obj in sub.objectives if obj.bounds is not None]
+    dim_f_bounded = sum(obj.dim_f for obj in bounded)
+    # constraint rows never vanish, bounds there become slack variables instead
+    m = objective.dim_f + (0 if constraint is None else constraint.dim_f)
+    n = x0.size
+    if m - dim_f_bounded < min(m, n):
+        supplied = hasattr(options, "tr_method")
```
YigitElma: move this out

YigitElma: This was fine (`supplied` was defined in the proper place), but changing it anyway to Daniel's suggestion with additional comments.

desc/optimize/_desc_wrappers.py
```diff
@@ -45,6 +46,42 @@
...
+            UserWarning,
+            f"Objectives {[obj.name for obj in bounded]} use bounds instead of target, "
+            + f"so they can zero out {dim_f_bounded} of the {m} rows of the ({m}, {n}) "
+            + f"Jacobian and drop its rank below {min(m, n)}. The trust region method "
+            + f"{tr_method} may then fail to solve the subproblem, it is safer to use "
+            + "options={'tr_method': 'svd'}.",
+        )
+        # if not set by user, use SVD to avoid rank-deficiency issues
+        if not hasattr(options, "tr_method"):
+            options["tr_method"] = "svd"
```
dpanici, line 81: Does it make sense to let user know that SVD is being used now? the warning implies it is not being used. something like

"User did not explicitly supply tr_method, so switching to the safer SVD" so they know it is fine

dpanici: nvm I see it only warns when tr method is not svd

dpanici (approved):

ddudt (approved):

Issue #2051, opened 2026-01-05
Actually Deprecate things that have been deprecated in DESC


dpanici:

YigitElma:
I will delete the old issue and document what should be removed here

ddudt:
Remember to add #1769 to the list once it is merged. 

Should we be consistently using `FutureWarning` everywhere instead of `DeprecationWarning`? If so, we could also update accordingly as part of this issue.

YigitElma:
> Should we be consistently using `FutureWarning` everywhere instead of `DeprecationWarning`

From this [documentation](https://docs.python.org/3/library/warnings.html#warning-categories), I think we should use `FutureWarning`. `DeprecationWarning` is not shown to users unless they run a Python script (like `python main.py`). I faced this when I was testing code in a Jupyter notebook, and I didn't get any `DeprecationWarning`s. `pytest` captures all the warnings, so that is fine, but end user might not see them.

ddudt:
Remember to add #2170 (and #2291) to the list once it is merged.

PR #2291, opened 2026-08-13
Dp/requested changes sparse

dpanici:
@unalmis I put my requested changes in this PR pointing to yours, tests locally pass for test_integrals, test_objective_funs and test_compute_everything (paths where I thought warnings might pop up potentially), feel free to merge if you accept these change requests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,7 @@ Changelog
 Performance Improvements
 
 - Improves memory management to reduce the base memory used during optimization while using `lsq-exact`, `lsq-auglag` and `fmin-auglag` optimizers.
-- Sparse reverse-mode differentiation was introduced to DESC
-  to yield significant performance improvements [#2170](https://github.com/PlasmaControl/DESC/pull/2170).
-  Plumbing to use this method was added to DESC that
-  will be progressively taken advantage of in the future.
+- Sparse reverse-mode differentiation was introduced to yield significant performance improvements [#2170](https://github.com/PlasmaControl/DESC/pull/2170). Plumbing to use this method was added that will be progressively taken advantage of in the future.
 
 Breaking Changes and Deprecations
 
@@ -15,10 +12,10 @@ Breaking Changes and Deprecations
 - The parameter ``Y_B`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` is now the resolution over a single field period rather than a full toroidal transit. This should make using a consistent resolution across different equilibria easier.
 - Objectives using ``Bounce2D`` now do not support fwd mode differentiation for JAX versions <0.11.0.
 
-
 Bug Fixes
 
 - Fixes bug in ``auglag`` optimizers which prevented them from accepting solver hyperparameters.
+
 
 v0.17.3
 -------
@@ -48,10 +45,10 @@ Bug Fixes
     - Updates ``"reactor_QA"`` in ``desc.examples`` to fix this. Note that if using ``"reactor_QA"`` example from ``v0.16.0`` until this fix, the current profile in that example has this issue.
 - Fixes bug in `CoilSet.from_symmetry` that ignored the passed in `check_intersection` value. This caused redundant checks in various other functions such as `plot_coils`.
 
-
 Breaking Changes
 
 - Name change in `_CoilObjective` replacing `coilset_mask` with `objective_mask`. Custom subclasses with `_broadcast_input="node"` that previously used `coilset_mask` should switch to `objective_mask`.
+
 
 v0.17.2
 -------
@@ -85,8 +82,6 @@ Performance Improvements
 - [Significantly improves convergence of inverse stream maps](https://github.com/PlasmaControl/DESC/pull/1919).
 - Resolves a JAX memory regression in bounce integrals by avoiding materialization of a large tensor in memory. Previously, we had closed the issue by adding nuffts as a workaround. This update actually solves the issue for the case when a user specifies to not use nuffts as well.
 - ``ObjectiveFunction.print_value`` can now use the previously computed ``compute_scaled_error`` values to print. For bounded objectives, we fall back to computing ``compute_unscaled``. Additionally, ``compute_scaled_error`` and array splitting are used in other parts of the code to prevent recompilation for one-time tasks, which makes initialization faster.
-
-
 
 Deprecations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Breaking Changes and Deprecations
 
 - The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``field_period_transits``. This should make using a consistent resolution across different equilibria easier. The now-deprecated ``num_transit`` may still be used but note the equivalence ``field_period_transits = num_transit * grid.NFP``.
 - The parameter ``Y_B`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` is now the resolution over a single field period rather than a full toroidal transit. This should make using a consistent resolution across different equilibria easier.
-- Objectives using ``Bounce2D`` now do not support fwd mode differentiation for JAX versions <0.11.0. These objectives' derivatives are inefficient in fwd mode, so rev mode should be used for them, which is significantly more efficient with the introduction of sparse reverse-mode differentiation.
+- Objectives using ``Bounce2D`` now do not support fwd mode differentiation for JAX versions <0.11.0.
 
 
 Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,19 @@ Changelog
 Performance Improvements
 
 - Improves memory management to reduce the base memory used during optimization while using `lsq-exact`, `lsq-auglag` and `fmin-auglag` optimizers.
+- Sparse reverse-mode differentiation was introduced to DESC
+  to yield significant performance improvements [#2170](https://github.com/PlasmaControl/DESC/pull/2170).
+  Plumbing to use this method was added to DESC that
+  will be progressively taken advantage of in the future.
+
+Breaking Changes and Deprecations
+
+- The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``num_field_periods``. This should make using a consistent resolution across different equilibria easier. The now-deprecated ``num_transit`` may still be used but note the equivalence ``num_field_periods = num_transit * grid.NFP``.
+- The parameter ``Y_B`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` is now the resolution over a single field period rather than a full toroidal transit. This should make using a consistent resolution across different equilibria easier.
+- Objectives using ``Bounce2D`` now do not support fwd mode differentiation for JAX versions <0.11.0. These objectives' derivatives are inefficient in fwd mode, so rev mode should be used for them, which is significantly more efficient with the introduction of sparse reverse-mode differentiation.
 
 
-v0.17.2
+v0.17.3
 -------
 
 New Features
@@ -64,10 +74,6 @@ Bug Fixes
 
 Performance Improvements
 
-- Sparse reverse-mode differentiation was introduced to DESC
-  to yield significant performance improvements [#2170](https://github.com/PlasmaControl/DESC/pull/2170).
-  Plumbing to use this method was added to DESC that
-  will be progressively taken advantage of in the future.
 - Reduces import time of `desc` modules.
     - Now, `desc.compute._build_data_index` uses depth-first search algorithm to construct the dependency tree.
     - Some of the default value computations at import time are removed (i.e. `desc.integrals.bounce_integral.default_quad`)
@@ -75,10 +81,7 @@ Performance Improvements
 - Resolves a JAX memory regression in bounce integrals by avoiding materialization of a large tensor in memory. Previously, we had closed the issue by adding nuffts as a workaround. This update actually solves the issue for the case when a user specifies to not use nuffts as well.
 - ``ObjectiveFunction.print_value`` can now use the previously computed ``compute_scaled_error`` values to print. For bounded objectives, we fall back to computing ``compute_unscaled``. Additionally, ``compute_scaled_error`` and array splitting are used in other parts of the code to prevent recompilation for one-time tasks, which makes initialization faster.
 
-Breaking Changes
 
-- The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``num_field_periods``. This should make using a consistent resolution across different equilibria easier. ``num_transit`` may still be used but note the equivalence ``num_field_periods = num_transit * grid.NFP``.
-- The parameter ``Y_B`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` is now the resolution over a single field period rather than a full toroidal transit. This should make using a consistent resolution across different equilibria easier.
 
 Deprecations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Performance Improvements
 
 Breaking Changes and Deprecations
 
-- The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``num_field_periods``. This should make using a consistent resolution across different equilibria easier. The now-deprecated ``num_transit`` may still be used but note the equivalence ``num_field_periods = num_transit * grid.NFP``.
+- The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``field_period_transits``. This should make using a consistent resolution across different equilibria easier. The now-deprecated ``num_transit`` may still be used but note the equivalence ``field_period_transits = num_transit * grid.NFP``.
 - The parameter ``Y_B`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` is now the resolution over a single field period rather than a full toroidal transit. This should make using a consistent resolution across different equilibria easier.
 - Objectives using ``Bounce2D`` now do not support fwd mode differentiation for JAX versions <0.11.0. These objectives' derivatives are inefficient in fwd mode, so rev mode should be used for them, which is significantly more efficient with the introduction of sparse reverse-mode differentiation.
 
--- a/desc/integrals/_bounce_utils.py
+++ b/desc/integrals/_bounce_utils.py
@@ -5,7 +5,6 @@ these utilities are private, and although it is unlikely,
 their API may change without warning.
 """
 
-import warnings
 from functools import partial
 
 import numpy as np
@@ -283,18 +282,6 @@ def _acrobatics(z, t, c, NFP, eps, mask):
         .reshape((-1,) + swapped_shape)  # then shape is (-1, ρ ?, b, α, λ, w)
         .swapaxes(1, -4)  # recover shape (-1, b, ρ ?, α, λ, w)
     )
-
-
-@partial(jax.custom_jvp, nondiff_argnums=(2,))
-def regular_points(o, pitch_inv, num_well):
-    """Alias of the function bounce_points."""
-    warnings.warn(
-        FutureWarning(
-            "Function regular_points has been renamed to bounce_points, and "
-            + "bounce_points will be removed in a future release."
-        )
-    )
-    return bounce_points(o, pitch_inv, num_well)
 
 
 @partial(jax.custom_jvp, nondiff_argnums=(2,))
--- a/desc/integrals/_bounce_utils.py
+++ b/desc/integrals/_bounce_utils.py
@@ -147,9 +147,8 @@ def _halley(o, pitch_inv, z, mask, nufft_eps, diagnostic=0):
         Shape (num ρ ?, num α, num pitch, num well).
         Subset of points to refine.
     nufft_eps : float
-        Precision requested for interpolation with non-uniform fast Fourier
-        transform (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
-
+        Precision requested for interpolation with non-uniform fast Fourier transform
+        (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
         Should satisfy ε < εᵢₙ² where εᵢₙ is the error of the input points.
     diagnostic : int
         Positive integer denoting iteration step to print.
@@ -987,8 +986,8 @@ def fast_cubic_spline(
     modes_z : jnp.ndarray
         FFT Fourier modes in toroidal direction.
     nufft_eps : float
-        Precision requested for interpolation with non-uniform fast Fourier
-        transform (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
+        Precision requested for interpolation with non-uniform fast Fourier transform
+        (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
     vander_t : jnp.ndarray
         Precomputed transform matrix.
     vander_z : jnp.ndarray
--- a/desc/integrals/_bounce_utils.py
+++ b/desc/integrals/_bounce_utils.py
@@ -779,7 +779,7 @@ def argmin(z1, z2, f, mins, B_mins):
     return jnp.take_along_axis(f[..., None, None, :], where, axis=-1).squeeze(-1)
 
 
-def get_alphas(alpha, iota, num_field_periods, NFP):
+def get_alphas(alpha, iota, field_period_transits, NFP):
     """Get set of field line poloidal coordinates {Aᵢ | Aᵢ = (αᵢ₀, αᵢ₁, ..., αᵢ₍ₘ₋₁₎)}.
 
     Parameters
@@ -790,7 +790,7 @@ def get_alphas(alpha, iota, num_field_periods, NFP):
     iota : jnp.ndarray
         Shape (num ρ, ).
         Rotational transform normalized by 2π.
-    num_field_periods : int
+    field_period_transits : int
         Number of field periods to follow field line.
     NFP: int
         Number of field periods per toroidal transit.
@@ -804,10 +804,10 @@ def get_alphas(alpha, iota, num_field_periods, NFP):
     """
     alpha = alpha[:, None, None]
     iota = iota[:, None]
-    return alpha + iota * (2 * jnp.pi / NFP) * jnp.arange(num_field_periods)
+    return alpha + iota * (2 * jnp.pi / NFP) * jnp.arange(field_period_transits)
 
 
-def theta_on_fieldlines(angle, iota, alpha, num_field_periods, NFP, *, X_min=24):
+def theta_on_fieldlines(angle, iota, alpha, field_period_transits, NFP, *, X_min=24):
     """Parameterize θ on field lines α.
 
     Parameters
@@ -821,7 +821,7 @@ def theta_on_fieldlines(angle, iota, alpha, num_field_periods, NFP, *, X_min=24)
     alpha : jnp.ndarray
         Shape (num α, ).
         Starting field line poloidal labels {αᵢ₀}.
-    num_field_periods : int
+    field_period_transits : int
         Number of field periods to follow field line.
     NFP : int
         Number of field periods per toroidal transit.
@@ -879,7 +879,7 @@ def theta_on_fieldlines(angle, iota, alpha, num_field_periods, NFP, *, X_min=24)
     domain = (0, 2 * jnp.pi / NFP)
 
     # peeling off field lines
-    alpha = get_alphas(alpha, iota, num_field_periods, NFP)
+    alpha = get_alphas(alpha, iota, field_period_transits, NFP)
     if angle.ndim == 2:
         alpha = alpha.squeeze(1)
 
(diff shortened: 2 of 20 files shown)


ddudt:
[suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
In this PR can we also change the name of `num_field_periods` to `field_period_transits` as explained in [this comment](https://github.com/PlasmaControl/DESC/pull/2170/changes#r3352300622)?

desc/integrals/bounce_integral.py
```diff
@@ -79,7 +79,7 @@ class _Bounce(eqx.Module, ABC):
     """Abstract class for bounce integrals."""
 
     @staticmethod
-    def pitch_quad(min_B, max_B, num_pitch, **kwargs):
+    def get_pitch_inv_quad(min_B, max_B, num_pitch, simp=True):
```
ddudt: @unalmis Was there a compelling reason to change the name to `pitch_quad`? If so, I can revert this and replace it with an alias to throw a deprecation warning about the new function name, like I did for `regular_points` -> `bounce_points`. But the original name `get_pitch_inv_quad` sounds more descriptive so I prefer keeping it.

unalmis: pitch_quad is shorter. get_pitch_inv_quad still works so there is no need to remove pitch_quad. What do you mean like you did for regular to bounce points? Those were private functions so you don't need deprecation

desc/integrals/_bounce_utils.py
```diff
@@ -285,6 +285,18 @@ def _acrobatics(z, t, c, NFP, eps, mask):
     )
 
 
+@partial(jax.custom_jvp, nondiff_argnums=(2,))
+def regular_points(o, pitch_inv, num_well):
```
unalmis: see other comment; should remove this, its not needed.

desc/integrals/bounce_integral.py
```diff
@@ -96,6 +96,13 @@ def pitch_quad(min_B, max_B, num_pitch, **kwargs):
             points xₖ and weights wₖ for the approximation of the integral
             ∫₋₁¹ f(x) dx ≈ ∑ₖ wₖ f(xₖ). Then this method simply rescales
             the quadrature for integration between ``min_B`` and ``max_B``.
+        simp : bool, optional
+            If ``True``, then the pitch angles are chosen so that the quadrature
+            over the velocity coordinate of 1/λ is done with Simpson’s 1/3 in the
+            interior completed by an open midpoint scheme near the boundary such
+            that an accuracy of fourth order is preserved.
+            If ``False``, then an open midpoint scheme is returned, which is only
+            recommended for plotting purposes.
```
unalmis: Order of accuracy is dependent on the integrand and therefore the doctoring added in this PR is incorrect.

desc/integrals/bounce_integral.py
unalmis: simp is a private kwarg that should not be documented.

desc/integrals/bounce_integral.py
```diff
@@ -110,7 +117,6 @@ def pitch_quad(min_B, max_B, num_pitch, **kwargs):
                 msg="Floating point error impedes detection of bounce points "
                 f"near global extrema. Choose {num_pitch} < 1e5.",
             )
-            simp = kwargs.get("simp", True)
```
unalmis, line 113: this was intentional since simp is private kwarg.

desc/integrals/bounce_integral.py
```diff
@@ -122,7 +128,7 @@ def pitch_quad(min_B, max_B, num_pitch, **kwargs):
         w = w * grad_bijection_from_disc(min_B, max_B)
         return x, w
 
-    get_pitch_inv_quad = pitch_quad
```
unalmis, line 125: This was intentional

desc/integrals/bounce_integral.py
```diff
@@ -122,7 +128,7 @@ def pitch_quad(min_B, max_B, num_pitch, **kwargs):
...
+    get_pitch_inv_quad = get_pitch_inv_quad
```
unalmis: The original code was intentional; see other comment.

desc/integrals/bounce_integral.py
```diff
@@ -301,11 +307,38 @@ def __init__(
     ):
         """Returns an object to compute bounce integrals."""
         assert grid.can_fft2
-        if "num_transit" in kwargs and num_field_periods == 20:  # default value
-            num_field_periods = kwargs["num_transit"] * grid.NFP
+        warnif(
```
unalmis: Remove this warning. It doesn't matter whether reshaped is given or not. It is handled automatically. They are welcome to keep providing it if they want. It's just going to annoy users if you warn without need.

desc/integrals/bounce_integral.py
```diff
@@ -301,11 +307,38 @@ def __init__(
...
+            "is_reshaped" in kwargs,
+            FutureWarning,
+            "Argument is_reshaped has been deprecated and is no longer used.",
+        )
+        warnif(
+            "is_fourier" in kwargs,
+            FutureWarning,
+            "Argument is_fourier has been deprecated and is no longer used.",
```
unalmis: Same as above.

desc/integrals/bounce_integral.py
```diff
@@ -969,7 +1030,9 @@ def interp_to_argmin(self, f, points, *, nufft_eps=-1.0, **kwargs):
             between ``z1`` and ``z2`` resides in the epigraph of B.
         nufft_eps : float
             Precision requested for interpolation with non-uniform fast Fourier
-            transform (NUFFT). If less than ``1e-14`` then NUFFT will not be used.
+            transform (NUFFT). If positive but less than ``1e-14`` then NUFFT will not
+            be used. If negative then the nufft_eps assigned to the Bounce2D object at
+            instantiation will be used.
```
unalmis: It is bad practice to publicly advertise such behavior in the API. The original writing is preferred. The API does not promise that negative values will use the default. That is simply an internal choice we use to default to the sentinel. This is discussed in previous PR.

desc/integrals/bounce_integral.py
```diff
@@ -978,6 +1041,12 @@ def interp_to_argmin(self, f, points, *, nufft_eps=-1.0, **kwargs):
             ``f`` interpolated to the deepest point between ``points``.
 
         """
+        warnif(
+            "is_fourier" in kwargs,
+            FutureWarning,
+            "Argument is_fourier has been deprecated and is no longer used.",
+        )
+
```
unalmis: See previous comment. See also earlier discussion in previous PR. Users are welcome to keep providing this kwarg and they should not be pestered.

desc/integrals/bounce_integral.py
```diff
@@ -1202,6 +1271,11 @@ def plot_angle_spectrum(
             Matplotlib (fig, ax) tuple.
 
         """
+        warnif(
+            "truncate" in kwargs,
+            FutureWarning,
+            "Argument truncate has been deprecated and is no longer used.",
+        )
```
unalmis: This is wrong. It is used. Revert this change.

desc/integrals/bounce_integral.py
```diff
@@ -1348,9 +1422,24 @@ def __init__(
     ):
         """Returns an object to compute bounce integrals."""
         assert grid.is_meshgrid
+        warnif(
+            "is_reshaped" in kwargs,
+            FutureWarning,
+            "Argument is_reshaped has been deprecated and is no longer used.",
+        )
```
unalmis: Remove. see previous comment.

desc/integrals/bounce_integral.py
```diff
@@ -1871,10 +1978,10 @@ class Options(NamedTuple):
     alpha: jnp.ndarray
     loop: bool
     nufft_eps: float
-    num_field_periods: int
+    field_period_transits: int
     num_well: int
     pitch_batch_size: int
-    pitch_quad: tuple[jnp.ndarray]
+    get_pitch_inv_quad: tuple[jnp.ndarray]
```
unalmis: Looks like an AI/other tool refactored this incorrectly. Needs to be reverted. It's a precomputed attribute not a verb method to execute. Also not a backward issue since this PR is first time this logic is introduced.

desc/integrals/bounce_integral.py
```diff
@@ -1941,11 +2048,11 @@ def guess(
             alpha=jnp.zeros(1) if alpha is None else alpha,
             loop=loop,
             nufft_eps=nufft_eps,
-            num_field_periods=num_field_periods,
+            field_period_transits=field_period_transits,
             num_well=num_well,
             pitch_batch_size=pitch_batch_size,
-            pitch_quad=jax.lax.stop_gradient(simpson2(num_pitch)),
-            quad=Options._quad(eta, num_quad) if quad is None else quad,
+            get_pitch_inv_quad=jax.lax.stop_gradient(simpson2(num_pitch)),
```
unalmis: see previous comment.

desc/objectives/_fast_ion.py
```diff
@@ -67,7 +67,6 @@ class GammaC(_Objective):
             bounds_default="``target=0``.",
             normalize_detail=" Note: Has no effect for this objective.",
             normalize_target_detail=" Note: Has no effect for this objective.",
-            jac_chunk_size=False,
```
unalmis, line 70: If the 3d f-zerni to real transform is very large and there are tens of thousands of optimizable params, then I suppose there is a use case that jac_chunk_size=finite number even with surf_batch_size=infinity.

Still given that most expensive part of Jacobian chunking is controlled by surf_batch_size and not jac_chunk_size, regardless of whether forward or reverse mode is used, it can be confusing to users to see loss of performance when jac_chunk_size is set to finite number in reverse mode. Thus I suggest to set that variable to None default and then set `jac_chunk_size = jac_chunk_size if (deriv_mode == "fwd") else None) inside.

desc/objectives/_fast_ion.py
```diff
@@ -88,13 +87,14 @@ def __init__(
         normalize_target=True,
         loss_function=None,
         deriv_mode="rev",
+        jac_chunk_size=None,
```
unalmis: We already have **kwargs, since these objectives do not accept jac_chunk_size.

desc/objectives/_fast_ion.py
```diff
@@ -154,8 +168,8 @@ def __init__(
             normalize_target=normalize_target,
             loss_function=loss_function,
             deriv_mode=deriv_mode,
+            jac_chunk_size=jac_chunk_size,
```
unalmis: This is not correct. My original logic of forcing it to None is correct.

desc/objectives/_neoclassical.py
```diff
@@ -67,13 +66,14 @@ def __init__(
         normalize_target=True,
         loss_function=None,
         deriv_mode="rev",
+        jac_chunk_size=None,
```
unalmis: remove; see previous comment.

desc/objectives/_neoclassical.py
```diff
@@ -132,8 +146,8 @@ def __init__(
             normalize_target=normalize_target,
             loss_function=loss_function,
             deriv_mode=deriv_mode,
+            jac_chunk_size=jac_chunk_size,
```
unalmis: force to None; see previous comment.

desc/objectives/objective_funs.py
```diff
@@ -232,7 +232,6 @@ def collect_docs(
     normalize_target_detail=None,
     loss_detail=None,
     coil=False,
-    jac_chunk_size=True,
```
unalmis: Revert. this is intentional

desc/objectives/objective_funs.py
```diff
@@ -256,8 +255,6 @@ def collect_docs(
     coil : bool, optional
         Whether the objective is a coil objective. If ``True``, updates docs
         of ``target``, ``weight``, ``bounds``, and ``loss_function``.
-    jac_chunk_size : bool
```
unalmis: revert

desc/objectives/objective_funs.py
```diff
@@ -269,9 +266,6 @@ def collect_docs(
 
     # Copy to allow for objective-specific updates to docs
     docs_obj = docs.copy()
-    if not jac_chunk_size:
-        del docs_obj["jac_chunk_size"]
```
unalmis: revert

desc/plotting.py
```diff
@@ -4758,7 +4758,7 @@ def plot_gammac(
     # Extract pitch angle range
     minB = data0["min_tz |B|"][0]
     maxB = data0["max_tz |B|"][0]
-    inv_pitch, _ = Bounce2D.pitch_quad(minB, maxB, num_pitch)
+    inv_pitch, _ = Bounce2D.get_pitch_inv_quad(minB, maxB, num_pitch)
```
unalmis: revert

desc/utils.py
```diff
@@ -510,8 +510,8 @@ def parse_argname_change(arg, kwargs, oldname, newname):
     if oldname in kwargs:
         warnings.warn(
             FutureWarning(
-                f"argument {oldname} has been renamed to {newname}, "
-                + f"{oldname} will be removed in a future release"
+                f"Argument {oldname} has been renamed to {newname}, and "
+                + f"{oldname} will be removed in a future release."
```
unalmis: People are going to complain to me that this cosmetic change here should be in a separate PR. I suggest reverting to avoid that.

docs/notebooks/tutorials/EffectiveRipple.ipynb
unalmis, line 1: revert get_pitch_inv_quad to pitch_quad

tests/test_integrals.py
```diff
@@ -1091,7 +1091,7 @@ def test_bounce1d_checks(self):
             Bounce1D.required_names + ["min_tz |B|", "max_tz |B|", "g_zz"], grid=grid
         )
         bounce = Bounce1D(grid, data, check=True)
-        pitch_inv, _ = bounce.pitch_quad(
+        pitch_inv, _ = bounce.get_pitch_inv_quad(
```
unalmis: revert

tests/test_integrals.py
```diff
@@ -1289,7 +1289,9 @@ def drift_analytical(data):
 
         # Exclude singularity not captured by analytic approximation for pitch near
         # the maximum |B|. (This is captured by the numerical integration).
-        pitch_inv = Bounce1D.pitch_quad(np.min(B), np.max(B), 100, simp=False)[0][:-1]
+        pitch_inv = Bounce1D.get_pitch_inv_quad(np.min(B), np.max(B), 100, simp=False)[
```
unalmis: revert

tests/test_integrals.py
```diff
@@ -1509,12 +1511,12 @@ def test_bounce2d_checks(self):
...
             angle,
             alpha=alpha,
-            num_field_periods=38,
+            field_period_transits=38,
             check=True,
             spline=False,
             quad=chebgauss1(16),  # this is our own custom chebgauss1
         )
-        pitch_inv, _ = bounce.pitch_quad(
+        pitch_inv, _ = bounce.get_pitch_inv_quad(
```
unalmis: revert

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ Changelog
...
+  to yield significant performance improvements [#2170](https://github.com/PlasmaControl/DESC/pull/2170).
+  Plumbing to use this method was added to DESC that
+  will be progressively taken advantage of in the future.
+
+Breaking Changes and Deprecations
+
+- The parameter ``num_transit`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` and related functions has been changed to ``field_period_transits``. This should make using a consistent resolution across different equilibria easier. The now-deprecated ``num_transit`` may still be used but note the equivalence ``field_period_transits = num_transit * grid.NFP``.
+- The parameter ``Y_B`` in ``EffectiveRipple``, ``Gamma_c``, ``Bounce2D`` is now the resolution over a single field period rather than a full toroidal transit. This should make using a consistent resolution across different equilibria easier.
+- Objectives using ``Bounce2D`` now do not support fwd mode differentiation for JAX versions <0.11.0. These objectives' derivatives are inefficient in fwd mode, so rev mode should be used for them, which is significantly more efficient with the introduction of sparse reverse-mode differentiation.
```
unalmis: Do not include last statement about inefficient in fwd mode. They are welcome to use efficiently use forward mode in Jax versions 11.0. For that reason, I would also prefer retaining the default deriv_mode="auto" in these objectives rather than deriv_mode="rev". Rory suggested me change it to rev, but I do not think he knew fwd mode can work at that time.

desc/integrals/bounce_integral.py
ddudt: OK I will keep it as `get_pitch_inv_quad` if you don't mind. 

I mean like [this change](https://github.com/PlasmaControl/DESC/blob/dp/requested-changes-sparse/desc/integrals/_bounce_utils.py#L289) to throw a warning if you call the old function name. I know we don't really intend for users to call this directly, but it's still good practice and only a few lines of extra code that we can remove after a couple of releases have passed.

unalmis (changes requested):
Looks like there is some confusion that is causing some unnecessary deprecation warnings. Have marked those that should be reverted or excluded.

~~Also the two line change I added in downstream PR to not include a dummy argument in the docstring should be kept to avoid polluting an already complicated objective. As just a practical example of why that is also useful for developers, observe this PR reverts that change has introduced an inefficiency of allowing jac_chunk_size to be modified by the user. That will result in poor performance at no improvement.~~ I recall  I cached the Jacobin so doesn't matter.

desc/integrals/bounce_integral.py
unalmis: Okay on the regular points thing but fyi the whole file is an underscore and there is a docstring at top of file that warns every method here is private and may change without notice.

desc/integrals/_bounce_utils.py
unalmis: https://github.com/PlasmaControl/DESC/pull/2291#discussion_r3790251826

desc/integrals/bounce_integral.py
unalmis: Just to clarify, in original PR get_pitch_inv_quad still works. Just pitch_quad is added as an option.

desc/integrals/_bounce_utils.py
```diff
@@ -69,7 +70,7 @@ def _bounce_points(
         If there were fewer wells detected along a field line than the size of the
         last axis of the returned arrays, then that axis is padded with zero.
     sentinel : float
-        Sentinel value which should be less ζ coordinate of all bounce points,
+        Sentinel value which should be less than ζ coordinate of all bounce points,
```
unalmis, line 72: thanks

desc/objectives/_fast_ion.py
```diff
@@ -123,6 +123,20 @@ def __init__(
             nufft_eps = 0.0
         nufft_eps = float(nufft_eps)
 
+        warnif(
+            "use_bounce1d" in kwargs,
+            FutureWarning,
+            "Argument use_bounce1d has been deprecated and is no longer used.",
+        )
```
unalmis: No purpose to this deprecation warning. Both algorithms are expected to give same result in infinite precision / resolution. At finite resolution, the new one is strictly better for accuracy and performance.

desc/objectives/_neoclassical.py
```diff
@@ -102,6 +102,20 @@ def __init__(
             nufft_eps = 0.0
         nufft_eps = float(nufft_eps)
 
+        warnif(
+            "use_bounce1d" in kwargs,
+            FutureWarning,
+            "Argument use_bounce1d has been deprecated and is no longer used.",
+        )
```
unalmis: See previous comment

CHANGELOG.md
dpanici: I will remove the statement about inefficiency. I made a TODO in the objectives (and a corresponding issue) to change deriv mode to auto once we support jax 0.11.0 in our requirements

desc/integrals/bounce_integral.py
ddudt: We want to document all arguments, since the documentation is not just for users but also helpful for us developers to remember what things do.

desc/objectives/_fast_ion.py
ddudt: We will make this change later. See #2294

desc/integrals/bounce_integral.py
ddudt: Sorry, I had found and replaced all instances of `pitch_quad`. I didn't realize that the attribute `BounceOptions.pitch_quad` was different from the method `_Bounce.pitch_quad`. I think that is more reason to keep them as separate names.

desc/integrals/bounce_integral.py
ddudt: Since this is a change to the public API, we need to throw a `FutureWarning` (even though it is not a breaking change). These will get removed later, see #2051.

desc/integrals/bounce_integral.py
ddudt: OK I had missed that this was backwards compatible. I would strongly prefer to keep the old name `git_pitch_inv_quad` so there is no change to the public API (and no confusion with `BounceOptions.pitch_quad`). If we are going to rename the function then we need to throw a `FutureWarning` instead of this silent alias.

ddudt:
> Looks like there is some confusion that is causing some unnecessary deprecation warnings. Have marked those that should be reverted or excluded.

Yes I had misunderstood some things, thanks for the clarification. I now get that you were not introducing any breaking changes to the public API, but those changes should still be documented. 

Let's try to make some compromises to get this finished. I reverted my edits to the private functions and I'm fine with those changing. But we still want users to be warned about changes to the public API and those warnings can be removed later. 

Once you merge this PR we will merge #2170.

desc/integrals/bounce_integral.py
unalmis: The intention is to always use simp unless user supplies their own quadrature in future. DESC has plenty of methods where some features are buried into kwargs to avoid polluting the signature, especially when those kwargs are unchaged  in almost all cases.

desc/integrals/bounce_integral.py
unalmis: >  I would strongly prefer to keep the old name

We are keeping the old name. That is what the original PR already does.

The public API is unchanged. Users can use both methods. There is no need for a warning; both always work. Both are fully documented in public api. One is simpler to use and shorter.

desc/integrals/bounce_integral.py
unalmis: Removing kwarg only parameter is_fourier from the explicit signature into the implicit kwargs signature is NOT a change to the public API.



> These will get removed later, see #2051.

1. Why? Users are welcome to keep providing that kwarg indefinitely; the public API should remain unchanged, and that is what the downstream PR does. Your changes force the user to change their code. It is unnecessary.

2. There is no need to introduce them in the first place. Everything works automatically, and there is no need to pester users that their code is or will be broken when my approach allows it to work as is indefinitely.

desc/integrals/bounce_integral.py
unalmis: You are trying to remove the `_Bounce.pitch_quad` option that all the upstream PRs use because you prefer `_Bounce.get_pitch_inv_quad`. But the compromise for your preference is already there! [You can use `get_pitch_inv_quad` as you please](https://github.com/PlasmaControl/DESC/pull/2291#discussion_r3807734608). You can also use `_Bounce.pitch_quad`. Again, many prefer the latter.

desc/integrals/bounce_integral.py
ddudt: Let me clarify: Yes I agree that the interface itself is not changing. The issue is that user code with these old arguments is no longer doing what they originally thought it was (since the old arguments are no longer used). This is only meant to be a helpful warning to inform the user.

unalmis: >  The issue is that user code with these old arguments is no longer doing what they originally thought it was (since the old arguments are no longer used).

This is not true. The code does what it was always doing; precisely as the user specified it. Please let me know if I should explain this more.

> This is only meant to be a helpful warning to inform the user.

It is a harmful warning because the user is doing exactly what they should regardless. If you are driving a car with your seatbelt on and the car keeps dinging: warning: you should put your seatbelt on, that is harmful warning.

desc/integrals/bounce_integral.py
ddudt: `_Bounce.get_pitch_inv_quad` is already the name of this function in the master code. I'm trying to keep that existing name because adding a second name as an alias is confusing and unnecessary. Am I missing something? If you are worried about other downstream PRs that already reference `_Bounce.pitch_quad` I will happily make the changes for you.

desc/integrals/bounce_integral.py
```diff
@@ -1174,6 +1233,7 @@ def plot_angle_spectrum(
         norm=LogNorm(1e-7),
         h_ax_numticks=None,
         v_ax_numticks=None,
+        truncate=0,
```
unalmis, line 1206: We have plenty of methods in DESC that take in options through kwargs to avoid polluting the function signature. This change is not needed.

desc/integrals/bounce_integral.py
ddudt: Yes please explain this more, maybe I am misunderstanding something. 

First of all, are we only talking about `is_reshaped` or also `is_fourier`, `Bref`, and `Lref`? Those are all accepted arguments to this function in the master code. So if the new code in #2170 removes them but has the exact same functionality, are they already obsolete and useless on the master branch? I don't think that is the case, in which case the change should be documented. 

For `is_reshaped` in particular, it looks like you now try to automatically detect its value instead of relying on the user to pass this argument? That's a nice change, but I still don't see the harm in informing the user that they don't have to manually do that anymore.

desc/integrals/bounce_integral.py
unalmis: I have shared why I prefer pitch_quad in previous messages. I have shared that both methods can be used.

>  adding a second name as an alias is confusing and unnecessary.

It is simply one line of code for backwards compatibility.

>  If you are worried about other downstream PRs that already reference _Bounce.pitch_quad I will happily make the changes for you.

You cannot make the change for every end user. These PRs have been open for 6 months.

desc/integrals/bounce_integral.py
ddudt: I mostly agree that "this change is not needed" -- you made this change in #2170 and here I am reverting it back. This PR is meant to help you clean up that other PR so we can merge it. 

In this case the original code I am reverting to is preferable because it provides better documentation about what `trucate` does. The docstring for `kwargs` in this function says "keyword arguments to pass to matplotlib", which isn't true for `truncate`.

desc/integrals/bounce_integral.py
unalmis: The first two is_reshaped, is_fourier.

> So if the new code in https://github.com/PlasmaControl/DESC/pull/2170 removes them but has the exact same functionality

yes

> are they already obsolete and useless on the master branch? I don't think that is the case, in which case the change should be documented.

no to the first question. Also no to the second statement. These arguments specify code logic that is done automatically now. Since these arguments can be inferred without ambiguity from the other options that are explicitly specified by the user, they are completely unnecessary. Users are welcome to keep supplying them if they please. It is hostile to warn them to change their code when it is not necessary.

> That's a nice change, but I still don't see the harm in informing the user that they don't have to manually do that anymore.

The harm is that it is not necessary. Maybe an analogy will help. Cars don't remind people to put on their seatbelt when it is already on. Cars don't remind people to put on their seatbelt when the machine is off. If it did, it would be extremely distracting to the driver even if they can technically ignore the warning. It is the same when programming.

unalmis:
These are not resolved and break code.

1. https://github.com/PlasmaControl/DESC/pull/2291#discussion_r3807889567
2. https://github.com/PlasmaControl/DESC/pull/2291#discussion_r3807981258

> But we still want users to be warned about changes to the public API and those warnings can be removed later.

Please see second point on why the edits made here are counter to above statement. There is still incorrect understanding that those comments correct.

desc/integrals/bounce_integral.py
unalmis: ok this change isn't worth arguing over for me.

desc/integrals/bounce_integral.py
ddudt: OK that's what I thought that it is automatic now, thanks for clarifying. 

The warning will go away once the user stops passing the old arguments, so in your analogy wouldn't that be like the warning turning off after the driver puts their seatbelt on?

unalmis: no its like warning staying on while the seatbelt is on until the company decides the driver has been annoyed for long enough.

unalmis: I will review again after this is removed.

ddudt: I think this is a common and encouraged practice. One similar example I can find is from the [seaborn visualization library](https://github.com/mwaskom/seaborn/blob/master/seaborn/distributions.py#L1989). If users give obsolete arguments they automatically handle it but still warn the users to update their code.

unalmis: It's not relevant. Their's explicitly says it will error if that arguments is given because in new versions because a newer argument has replaced it.  They are breaking code and deprecating. 

In contrast, I am not changing the API. I am not breaking code. I will always support users passing in these arguments if they wish. Please stop trying to break the API.

desc/integrals/bounce_integral.py
YigitElma: I agree with Daniel here. If the only reason for this change is the shorter name, I prefer keeping the old name. First, having the same name for a key in the `Options` class and the method in the `_Bounce` classes is confusing. Also, even if the old name still works, having aliases creates redundant resistance to understanding the code.

That being said, I don't think this is worth arguing, so I am fine either way. Just wanted to share my opinion before merging anything.

dpanici: adding comment in-line as documentation and removing from docstring

dpanici:
@unalmis I think I made changes that are a compromise with what me and Daniel wanted. Changed all the get_inv_pitch_quad usages back to pitch_quad. Removed most of the unneeded warnings but kept a couple that it was not clear to me if the functionality is now automatic or not (expand_out for example)

unalmis (approved):

PR #2293, opened 2026-08-14
Misc

YigitElma:
I put bunch of random small changes to this PR, but happy to create separate ones. Also, we don't need to include all.

- Improve the LCP build method when dealing with large arrays. It also removes some of the big arrays completely, like versions of `_Z` to cut down the base memory. Same spirit as #2281, but this one also improves the optimization initialization time by roughly half on GPU.
- Improves compilation time of `qr_multiply` by using a `jax.lax.scan` instead of the unrolled for loop.
- Updates the initial value for `alpha`. The previous value of 0 was basically recalculating the same `p_newton`
- Improves the speed of `rpz/xyz` vector rotation on GPU, which is heavily used in free-boundary `_kernel_biot_savart`. On my laptop, the same code didn't have an impact on CPU speed, but CI benchmarks were 50% slower, so I made this a backend-dependent change.

Needs #2287 for some tests

Just a rough `eq.solve` memory benchmark:
memory-01_eq_solve-gpu

Note: Chunk size 464 is the case where `batched_vectorize` doesn't have a remainder part.

### DIFF
--- a/desc/optimize/least_squares.py
+++ b/desc/optimize/least_squares.py
@@ -24,6 +24,7 @@ from .utils import (
     compute_jac_scale,
     print_header_nonlinear,
     print_iteration_nonlinear,
+    scale_columns,
     solve_triangular_regularized,
 )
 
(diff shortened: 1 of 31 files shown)


desc/optimize/least_squares.py
singh-jaydeep, line 337: Should the value of alpha which gets passed here change? It looks like 0 still gets passed.

YigitElma: Good catch, I will change

desc/utils.py
YigitElma, line 1: This small change affects the free boundary a lot on my GPU (it is also faster on A100 but the effect is a couple of percent there; compile time is lower though), since this function is called inside the kernel many times.
Image

YigitElma: I will probably make this change based on backend.

PR #2297, opened 2026-08-19
Remove static_attrs variables from constants dicts in _omnigenity objectives

### LINKED ISSUE #2295: `_static_attr` that is stored in `_constants` throw tracer conversion error
Another reason to remove `constants`. If an attribute is set in `_static_attr` but passed inside `constants` and used from there as `constants['static']`, it throws a `TracerConversion` error during evaluation. For example,
```python
eq = get("precise_QA")
obj = ObjectiveFunction(QuasisymmetryBoozer(eq=eq, surf_batch_size=1))
obj.build()
obj.compute_scaled_error(obj.x(eq))
```
```
TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[].
The error occurred while tracing the function compute_scaled_error at 
/CODES/DESC/desc/objectives/objective_funs.py:1444 for jit. This concrete value was not 
available in Python because it depends on the value of the argument 
self[0]['_constants']['surf_batch_size'].
```

I think fixing `QuasisymmetryBoozer` is enough for this issue, but I still wanted to document this problem in general.

dpanici:
Resolves #2295

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -268,7 +268,7 @@ def test_qh_optimization():
 
     eq1 = run_qh_step(0, eq)
 
-    obj = QuasisymmetryBoozer(helicity=(1, eq1.NFP), eq=eq1)
+    obj = QuasisymmetryBoozer(helicity=(1, eq1.NFP), eq=eq1, surf_batch_size=1)
     obj.build()
     B_asym = obj.compute(*obj.xs(eq1))
 
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -270,7 +270,7 @@ def test_qh_optimization():
 
     obj = QuasisymmetryBoozer(helicity=(1, eq1.NFP), eq=eq1, surf_batch_size=1)
     obj.build()
-    B_asym = obj.compute(*obj.xs(eq1))
+    B_asym = obj.compute_unscaled(*obj.xs(eq1))
 
     np.testing.assert_array_less(np.abs(B_asym).max(), 1e-1)
     np.testing.assert_array_less(eq1.compute("a_major/a_minor")["a_major/a_minor"], 5)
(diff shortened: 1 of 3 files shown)


tests/test_examples.py
```diff
@@ -268,7 +268,7 @@ def test_qh_optimization():
 
     eq1 = run_qh_step(0, eq)
 
-    obj = QuasisymmetryBoozer(helicity=(1, eq1.NFP), eq=eq1)
+    obj = QuasisymmetryBoozer(helicity=(1, eq1.NFP), eq=eq1, surf_batch_size=1)
     obj.build()
     B_asym = obj.compute(*obj.xs(eq1))
```
YigitElma: If the intent is to test if the static attribute error is fixed here, then I would suggest calling `obj.compute_unscaled` since plain `compute` is not jitted and wouldn't recreate the issue.

YigitElma (commented):
Minor comment, otherwise looks good.

YigitElma (approved):

tests/test_examples.py
dpanici: Thanks for making the change

tests/test_examples.py
```diff
@@ -268,9 +268,9 @@ def test_qh_optimization():
 
     eq1 = run_qh_step(0, eq)
 
-    obj = QuasisymmetryBoozer(helicity=(1, eq1.NFP), eq=eq1)
+    obj = QuasisymmetryBoozer(helicity=(1, eq1.NFP), eq=eq1, surf_batch_size=1)
     obj.build()
-    B_asym = obj.compute(*obj.xs(eq1))
+    B_asym = obj.compute_unscaled(*obj.xs(eq1))
```
ddudt, line 273: [question](https://pullpo.io/cc?l=question) 
Is there a reason this had to change?

ddudt (approved):

tests/test_examples.py
YigitElma: The original problem only occurs under jit and `compute` method is not jitted, so we call the jitted equivalent of compute here.

PR #2298, opened 2026-08-19
Getting proximal to work with augmented Lagrangian optimizers

### LINKED ISSUE #873: Get `ProximalProjection` to work with augmented lagrangian
Needed when we want to exactly enforce equilibrium constraint which may be useful for stability optimization

### LINKED ISSUE #1720: Don't check whole state vector inside `ProximalProjection._update_equilibrium`
https://github.com/PlasmaControl/DESC/blob/1437e9b8d9e7bedcf923aeb89b04c447f303a436/desc/optimize/_constraint_wrappers.py#L866

@YigitElma can you maybe explain this more? Not sure I follow, I guess as it is currently coded, you could have say a single stage optimization where the coils change but the equilibrium DOFs did not change at all, and in that case one would not need to re-solve the equilibrium?

singh-jaydeep:
One way of addressing #873. Right now, the only nonlinear constraints which proximal methods accept are force balance related ones. `ProximalProjection` already handles the process of translating between the full set of state variables (including R_lmn, Z_lmn, L_lmn) and the reduced set of optimization variables, and ensures that an objective wrapped in proximal gets the proper tangents. For augmented lagrangian methods, the nonlinear constraints need those same tangents.

This adds a class, `ProximalState`, which is responsible for managing all properties of the equilibrium subproblem. Any proximal wrapped objective, e.g. the objective or a nonlinear constraint, can access the up-to-date equilibrium this way. The state also stores the equilibrium tangents for the present equilibrium, so multiple calls to grad start to just read the cache. It seems to work for the few cases I have tried.

Since this is an issue others have thought about more, there may be better ways of getting this to work. Lmk and I can refactor. 

Also some of the diff is due to changes from #2239, which will disappear once that is merged. 

Resolves #873 
Resolves #1720

### DIFF
--- a/desc/optimize/_constraint_wrappers.py
+++ b/desc/optimize/_constraint_wrappers.py
@@ -1267,12 +1267,11 @@ class ProximalState:
         self.eq = eq
         self.constraint = constraint
 
-        perturb_options = setdefault(perturb_options, {})
-        solve_options = setdefault(solve_options, {})
+        perturb_options = dict(setdefault(perturb_options, {}))
+        solve_options = dict(setdefault(solve_options, {}))
         self._solve_during_proximal_build = solve_options.pop(
             "solve_during_proximal_build", True
         )  # If user does not want the solve during build, mainly for debug purposes
-        perturb_options = {} if perturb_options is None else perturb_options
         perturb_options.setdefault("verbose", 0)
         perturb_options.setdefault("include_f", False)
         solve_options.setdefault("verbose", 0)
--- a/desc/optimize/_constraint_wrappers.py
+++ b/desc/optimize/_constraint_wrappers.py
@@ -572,6 +572,9 @@ class ProximalProjection(ObjectiveFunction):
     perturb_options, solve_options : dict
         dictionary of arguments passed to Equilibrium.perturb and Equilibrium.solve
         during the projection step.
+    state: ProximalState, optional
+        State manager for the equilibrium constraints. Default is None, in which
+        case the state is created.
     name : str
         Name of the objective function.
     """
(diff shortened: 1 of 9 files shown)


desc/optimize/aug_lagrangian_ls.py
```diff
@@ -557,9 +558,8 @@ def lagjac(z, y, mu, *args):
                 # if we update lagrangian params, need to recompute L and J
                 L = lagfun(f, c, y, mu)
                 Lcost = 0.5 * jnp.dot(L, L)
-                del J
-                J = lagjac(z, y, mu, *args)
-                njev += 1
+                J = J.at[f.size :].multiply((jnp.sqrt(mu) / jnp.sqrt(mu_old))[:, None])
```
singh-jaydeep: This change is not strictly necessary for the PR, but saves a Jacobian calculation.

desc/optimize/optimizer.py
```diff
@@ -599,16 +632,37 @@ def _maybe_wrap_nonlinear_constraints(
...
+                f"Method {wrapper}-{method} cannot accept nonlinear constraints. "
+                + "Consider an Augmented Lagrangian solver such as proximal-lsq-auglag."
+            )
+
         perturb_options = options.pop("perturb_options", {})
         solve_options = options.pop("solve_options", {})
-        objective = ProximalProjection(
-            objective,
-            constraint=_combine_constraints(nonlinear_constraints),
+        state = ProximalState(
```
singh-jaydeep, line 652: The ProximalState object is created here, outside of any particular ProximalProjection wrapper. This allows both the objective and nonlinear_constraints to see this state.

desc/optimize/optimizer.py
```diff
@@ -718,7 +772,7 @@ def get_combined_constraint_objectives(  # noqa: C901
         objective.build(verbose=verbose)
         if nonlinear_constraint is not None:
             nonlinear_constraint = LinearConstraintProjection(
-                nonlinear_constraint, linear_constraint
+                nonlinear_constraint, linear_constraint, **linear_constraint_options
```
singh-jaydeep, line 776: This change is unrelated to the PR, but linear_constraint_options was not being passed here.

desc/optimize/_constraint_wrappers.py
```diff
@@ -1220,78 +1068,136 @@ def _jvp(self, v, x, constants=None, op="scaled_error"):
...
-                jnp.split(xg, np.cumsum(self._dimx_per_thing)),
+                jnp.split(tangents, np.cumsum(self._state.dimx_per_thing), axis=-1),
+                jnp.split(x, np.cumsum(self._state.dimx_per_thing)),
                 op,
             )
 
-    def _get_tangent(self, v, xf, constants, op):
-        # Note: This function is vectorized over v. So, v is expected to be 1D array
-        # of size self.dim_x.
+    def _vjp(self, v, x, constants=None, op="scaled"):
```
singh-jaydeep, line 1091: Added a vjp for use with auglag

desc/optimize/optimizer.py
YigitElma: I need to check this but I think we intentionally removed this at some point. I need to look at the changelog, the intent was to remove the 2 separate xscales and just using one.

singh-jaydeep: Oh gotcha, I'll check too.

YigitElma: Oh ok, so we didn't remove it yet, but we were planning to. I think until we actually remove it, this change is fine, since currently there is an inconsistency between the 2 `LinearConstraintProjection`s (see 3 lines above)

desc/optimize/_constraint_wrappers.py
```diff
@@ -1304,23 +1210,282 @@ def constants(self):
...
 
-def jit_if_possible(func):
+        assert isinstance(constraint, ObjectiveFunction), (
+            "constraint should be instance of ObjectiveFunction." ""
+        )
+        for con in constraint.objectives:
+            errorif(
+                not con._equilibrium,
+                ValueError,
+                "ProximalProjection method cannot handle general "
```
YigitElma: We might need to update the error message here to `ProximalState`

desc/optimize/_constraint_wrappers.py
```diff
@@ -1304,23 +1210,282 @@ def constants(self):
...
+                + f"nonlinear constraint {con}.",
+            )
+            # can't have bounds on constraint bc if constraint is satisfied then
+            # Fx == 0, and that messes with Gx @ Fx^-1 Fc etc.
+            errorif(
+                con.bounds is not None,
+                ValueError,
+                "ProximalProjection can only handle equality constraints, "
```
```suggestion
                "ProximalState can only handle equality constraints, "
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -1304,23 +1210,282 @@ def constants(self):
...
+            self.constraint,
+            ObjectiveFunction(self.eq_linear_constraints),
+            name="Eq Update LinearConstraintProjection",
+        )
+        self.eq_solve_objective.build(use_jit=use_jit, verbose=verbose)
+
+        errorif(
+            self.constraint.things != [self.eq],
+            ValueError,
+            "ProximalProjection can only handle constraints on the equilibrium.",
```
```suggestion
            "ProximalState can only handle constraints on the equilibrium.",
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -1304,23 +1210,282 @@ def constants(self):
...
+        things = list(things)
+
+        layout_changed = (not hasattr(self, "things")) or (self.things != things)
+        self.eq_idx = things.index(self.eq)
+        self.things = things
+
+        # the full state vector includes all the parameters from all the things
+        # however, sub-objectives only need the part for their thing. We will
+        # use this to split the state vector into its components
+        self.dimx_per_thing = [t.dim_x for t in self.things]
```
YigitElma: I think `dimx_per_thing` and `dimc_per_thing` should stay as properties of `ProximalProjection`. They include redundant information about other optimizable thing (probably the reason you had to do the bookkeeping above).

I would prefer if this class only ever inputs eq related information, so whoever is calling this class should always pass `v_splits[eq_idx]` kind of vectors here if possible (I think it should be)

desc/optimize/_constraint_wrappers.py
```diff
@@ -1220,78 +1068,136 @@ def _jvp(self, v, x, constants=None, op="scaled_error"):
...
+        return self._jvp_given_tangents(tangents, xg, constants, op)
+
+    def _full_tangents(self, x, xf, constants, op):
+        """Gets the full matrix of tangents from self._state."""
+        state = self._state
+
+        # If self._state has not yet computed and stored the tangents,
+        # this function gets called.
+        def build_tangents():
+            return _proximal_get_tangents(
```
YigitElma: I think you can safely move this function to `ProximalState`. Almost all the arguments are state attributes anyway. You just need to handle `tuple(state.dimc_per_thing)` or pass it too.

desc/optimize/_constraint_wrappers.py
```diff
@@ -1220,78 +1068,136 @@ def _jvp(self, v, x, constants=None, op="scaled_error"):
...
+            constants[1],
+            state.eq_solve_objective._feasible_tangents,
+            state.dxdc,
+            state.dimc_per_thing,
+            state.eq_idx,
+            op,
+        )
+        return self._jvp_given_tangents(tangents, xg, constants, op)
+
+    def _full_tangents(self, x, xf, constants, op):
```
YigitElma: If you move the below definition to state itself, I don't think you need this as a separate method, it can be just called inline
```python
tangents = self._state.get_tangents(x, dimc_per_thing, op)
if self._objective._deriv_mode == "batched":
     # objective's method already know about its jac_chunk_size
     return getattr(self._objective, "jvp_" + op)(tangents, xg, constants[0])
else:
     return _proximal_jvp_blocked_pure(
          self._objective,
          jnp.split(tangents, np.cumsum(self._dimx_per_thing), axis=-1),
          jnp.split(xg, np.cumsum(self._dimx_per_thing)),
          op,
     )
```

desc/optimize/_constraint_wrappers.py
```diff
@@ -1220,78 +1068,136 @@ def _jvp(self, v, x, constants=None, op="scaled_error"):
...
+            Vectors to left-multiply the Jacobian by.
+        x : ndarray
+            Optimization variables.
+        constants : list
+            Constant parameters passed to sub-objectives. (Deprecated)
+        """
+        return self._vjp(v, x, constants, "unscaled")
+
+    @property
+    def dim_x(self):
```
YigitElma: For sake of easier review, can you move this definition to where it was above?

desc/optimize/_constraint_wrappers.py
```diff
@@ -1389,3 +1561,49 @@ def _proximal_jvp_blocked_pure(objective, vgs, xgs, op):
...
+    # dG/dc = J @ (tangent vectors in c direction)
+    # dG/dx = J @ (tangent vectors in x direction)
+    # So, dG/dc - dG/dx * (dF/dx)⁻¹ * dF/dc can be written as
+    # J @ [(tangent vectors in c direction) - (tangent vectors in x direction)@dfdc]
+    # Note: We will never form full Jacobian J, we will just compute the above
+    # expression by JVPs.
+
+    # v contains prox._args DoFs from eq and other objects (like coils, surfaces
+    # etc). Only the eq block changes when the equilibrium is re-solved.
+    vs = jnp.split(v, np.cumsum(dimc_per_thing)[:-1], axis=-1)
```
YigitElma: What I mean in above comments is something like this. This function only changes portion of the `vs`. So, by small changes to this function you can get rid of all the `things` updates. For example, if this function becomes,
```python
@jit_if_possible(static_argnames=("op"))
def _proximal_get_tangents(
    constraint,
    xf,
    veq,
    constants,
    eq_feasible_tangents,
    dxdc,
    eq_idx,
    op="scaled_error",
):
    if veq.ndim == 2 and veq.shape[0] > dxdc.shape[1]: # dxdc thing should be checked but either 1 or 0
        eq_tangents = veq @ _proximal_eq_tangents(
            constraint, xf, constants, eq_feasible_tangents, dxdc.T, op
        )
    else:
        dxdcv = veq @ dxdc.T
        # atleast_2d and reshape are to also handle a single (1D) direction
        eq_tangents = _proximal_eq_tangents(
            constraint, xf, constants, eq_feasible_tangents, jnp.atleast_2d(dxdcv), op
        )
        eq_tangents = eq_tangents.reshape(dxdcv.shape)
    return eq_tangents
```
Then do the split and concatenation outside of this function.

desc/optimize/aug_lagrangian_ls.py
YigitElma: I think this is nice to have. However, it will temporarily create a copy of `J` without jit. I think you can use `desc.optimize.utils.scale_columns` here with some proper `d`. IIRC, there should be a way of `d[None, :]` or something like that to scale rows. Feel free to update the name of the function to be general scaling of a matrix.

YigitElma (commented):
I like the `ProximalState` approach. With minor changes to #2239, I believe we can clean this up a lot.

As a summary, this is the chain of wrapping happening here,

Image

YigitElma:
Also a quick question, with these changes is there a way to still use the old way? It would be nice if this only happened for 'proximal-auglag' or 'proximal-lsq-auglag'.

desc/optimize/_constraint_wrappers.py
singh-jaydeep: These properties are now back in ProximalProjection. The only place where ProximalState still has to engage with the larger set of optimization variables is get_tangents.

desc/optimize/_constraint_wrappers.py
singh-jaydeep: I ended up passing in eq_idx as well. Let me know if this is what you were thinking. 

The state.get_tangents method is now called directly by the ProximalProjection wrapper. Just as an aside: while refactoring this I realized the tangent cache wasn't set up correctly, and resulting in lots of cache misses. This is due to the difference in how jvp and vjp were written, with some wrappers (like LCP) routing grad through jvp and others (ProximalProjection, after my changes) going through vjp. Now the jvp and vjp methods go through the same cache. A consequence is the added flag cache_tangents on ProximalState. If you don't want to use the cache, as is likely the case in proximal-lsq-exact (you don't need to share tangents across multiple wrappers), you use the default False. For proximal-auglag methods we set True.

desc/optimize/_constraint_wrappers.py
singh-jaydeep: Deleted this method and the jvp_given_tangents method.

desc/optimize/_constraint_wrappers.py
singh-jaydeep: Made this suggested change to _proximal_get_tangents. It now asks for the equilibrium portion of the tangents.

singh-jaydeep:
@YigitElma I implemented your suggested changes; let me know if anything is unclear. I'll think about whether there is a simple method to let users choose the old way.

desc/optimize/_constraint_wrappers.py
```diff
@@ -697,105 +617,48 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
...
-        self._eq_linear_constraints = maybe_add_self_consistency(
-            self._eq, self._eq_linear_constraints
-        )
-
-        # we don't always build here because in ~all cases the user doesn't interact
-        # with this directly, so if the user wants to manually rebuild they should
-        # do it before this wrapper is created for them.
+        # # we don't always build here because in ~all cases the user doesn't interact
+        # # with this directly, so if the user wants to manually rebuild they should
+        # # do it before this wrapper is created for them.
```
```suggestion
        # we don't always build here because in ~all cases the user doesn't interact
        # with this directly, so if the user wants to manually rebuild they should
        # do it before this wrapper is created for them.
```
Redundant change

desc/optimize/_constraint_wrappers.py
```diff
@@ -697,105 +617,48 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
...
-        self._eq_solve_objective.build(use_jit=use_jit, verbose=verbose)
 
-        errorif(
-            self._constraint.things != [self._eq],
-            ValueError,
-            "ProximalProjection can only handle constraints on the equilibrium.",
-        )
-
-        self._objectives = [self._objective, self._constraint]
+        self._state.build(use_jit=use_jit, verbose=verbose)
```
YigitElma, line 626: I guess it is safer to put this under `if not self._state.built:`. The second proximal projection in the augmented lagrangian don't need to build it again

desc/optimize/_constraint_wrappers.py
```diff
@@ -697,105 +617,48 @@ def build(self, use_jit=None, verbose=1):  # noqa: C901
...
+            If None, uses "things" of self._objectives.
+
+        """
+        super()._set_things(things)
+
+        # Sync "things" between the wrapper and objective.
+        # Does not include self._constraint
+        self._objective._set_things(self.things)
+
+        self._eq_idx = self.things.index(self._state.eq)
```
YigitElma, line 656: A bit of a nitpick, but I prefer keeping these in the build method, and removing this new definition of the `_set_things` all together. I personally think having many small helper methods actually makes the code harder to read (need to jump from function to function). But others prefer that I am fine.

desc/optimize/_constraint_wrappers.py
```diff
@@ -824,17 +687,23 @@ def unpack_state(self, x, per_objective=True):
                 + f"{self.dim_x} got {x.size}."
             )
 
-        xs = jnp.split(x, np.cumsum(self._dimc_per_thing))
+        xs = jnp.split(x, np.cumsum(self._dimc_per_thing)[:-1])
```
YigitElma, line 690: Was this a bug before? Can you explain why we need this?

desc/optimize/_constraint_wrappers.py
```diff
@@ -1099,6 +947,10 @@ def hess(self, x, constants=None):
         J = self.jac_scaled_error(x, constants)
         return J.T @ J
 
+    def _jac(self, x, constants=None, op="scaled"):
+        # passing v=None corresponds to jvp in all directions
```
YigitElma: Are you sure this works? We use `v` in some matrix operations too, I am not sure if this will work there

YigitElma (commented):
Looks much cleaner than befoer. Thanks for all the changes!

Image

Does this memory improvement have a specific reason? I cannot think of any. Can there be a bug?

desc/optimize/_constraint_wrappers.py
singh-jaydeep: Yeah you're right, we should not need this. At one point I was concerned about the last entry in cumsum, but it doesn't seem to be a problem.

desc/optimize/_constraint_wrappers.py
singh-jaydeep: The main reason for this being separate was due to a sync issue in optimizer.optimize(). When the different wrappers are built, they may have different "things". The combine_args call later in that function calls set_things with the definitive set of things, allowing us to sync the layout vectors (like dimx_per_thing) at that time. Another option could be to just collect the list of things earlier in optimize(), and pass that to the wrapper. I can try that.

desc/optimize/_constraint_wrappers.py
singh-jaydeep: The build() function for ProximalState has such a check. Would it be clearer to have the check here instead?

singh-jaydeep:
My best guess for the memory change is that the tangent cache is just returning the proximal tangents on the second and third call in the test. This would be a test specific thing, and wouldn't correspond to a savings in practice. I will try explicitly clearing the cache in the test between calls and see if that gives identical results to master.

YigitElma:
> I'll think about whether there is a simple method to let users choose the old way.

If the optimizer is `opt=Optimizer("proximal-lsq-auglag")`, then we can use this method. I think the passing only `"lsq-auglag"` might already give the old behavior.

desc/optimize/_constraint_wrappers.py
YigitElma: Oh, okay then never mind!

desc/optimize/_constraint_wrappers.py
singh-jaydeep: There are two ways I can see getting rid of `_set_things`. Neither one is great, but open to suggestions.
1. pass things to `ProximalProjection` when it is being built the first time. 
2. in `combine_args`, the function which calls `set_things` on each objective with the canonical things list, have a check if any of the objectives are proximal wrappers. If so, either have additional logic there or call `ProximalProjection.build()` again. The build method would have a check that if built, don't redo any expensive operations and just set these indexing vectors.

PR #2300, opened 2026-08-23
Handle zero Gershgorin bound in modified Cholesky

### LINKED ISSUE #2296: Special handling for Gershgorin lower bound of 0
If the lower bound is 0, then the logarithm in this line https://github.com/PlasmaControl/DESC/blob/7582fb387c0e5a6fbee4068bff621a5f92759643/desc/optimize/utils.py#L218 gives -inf, which result in Nan `alpha` values.

krystophny:
Fixes #2296

Use a scale-aware positive fallback when the Gershgorin lower bound is exactly zero, including matrices at extreme subnormal scale. The positive-definite path remains unchanged.

Validation:
- optimizer test module: 93 passed
- focused zero, singular, extreme-scale, and positive-definite oracles passed
- pre-commit hooks on changed files passed

### DIFF
--- a/desc/optimize/utils.py
+++ b/desc/optimize/utils.py
@@ -187,8 +187,9 @@ def _cholmod(A, maxiter=4):
     Attempts to find smallest alpha to the nearest order of magnitude,
     in maxiter steps (so 2**maxiter values of alpha will be scanned over).
     Scans over values of -log(abs(lb))< log(alpha) < log(abs(lb)). If the
-    Gershgorin lower bound is zero, a matrix-scaled machine epsilon supplies
-    the positive reference value for the logarithmic search.
+    Gershgorin lower bound is zero, a matrix-scaled machine epsilon, floored
+    at the dtype's smallest normal value, supplies the positive reference
+    value for the logarithmic search.
 
     Cost is approximately maxiter times the cost of a single cholesky
     factorization.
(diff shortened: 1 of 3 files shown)


desc/optimize/utils.py
```diff
@@ -186,7 +186,10 @@ def _cholmod(A, maxiter=4):
 
     Attempts to find smallest alpha to the nearest order of magnitude,
     in maxiter steps (so 2**maxiter values of alpha will be scanned over).
-    Scans over values of -log(abs(lb))< log(alpha) < log(abs(lb))
+    Scans over values of -log(abs(lb))< log(alpha) < log(abs(lb)). If the
```
YigitElma:
```suggestion
    Scans over values of -log(abs(lb)) < log(alpha) < log(abs(lb)). If the
```

YigitElma (commented):
Thanks for the contribution! The fix looks good to me.

Can you update the changelog for the bug fix too? I think without going too much into detail, we can have something like this,
```
Bug Fixes
---------
- Fixes bug in modified Cholesky factorization used by the trust-region
  subproblems when the Gershgorin
  lower bound of the Hessian was exactly zero (e.g. a Hessian with an all-zero row),
  producing NaN steps in ``fmintr`` and ``fmin-auglag`` with the default
  ``tr_method="exact"``, and in ``lsq-exact``/``lsq-auglag`` with ``tr_method="cho"``.
```

krystophny:
Added the requested concise changelog entry in d9b67ba. Thanks!

YigitElma (approved):
Thank you! This looks good to me.

dpanici (approved):

YigitElma:
@krystophny, if you don't want to make more changes, we can merge this PR into master. Thanks!

krystophny:
> @krystophny, if you don't want to make more changes, we can merge this PR into master. Thanks!

sure thx!

PR #2302, opened 2026-08-25
Ensure CoilCurvature and CoilTorsion are resolution independent

### LINKED ISSUE #2301: Certain coil objectives depend on resolution
For per-node objectives like CoilCurvature and CoilTorsion, the value of Objective.compute_scalar() depends on resolution. I'm not sure if there is another reason for the choice of quad_weights? There are some relevant comments in the original PR [#853](https://github.com/PlasmaControl/DESC/pull/853#issuecomment-2005709571).

Example: 
```
coil = FourierXYZCoil()
res_array = [10, 30, 50]
obj_array = []
for res in res_array:
    grid=LinearGrid(N=res)
    obj = CoilCurvature(coil,grid=grid,target=1)
    obj.build()
    obj_array.append(obj.compute_scalar(coil.params_dict))
obj_array
```
returns 
```
[Array(0.93996232, dtype=float64),
 Array(0.32359359, dtype=float64),
 Array(0.19543771, dtype=float64)]
```

singh-jaydeep:
Resolves #2301 by replacing `quad_weights` with their square roots for these objectives. Also strengthens the resolution independence test for coil objectives.


dpanici:
Can you also rerun the coil optimization notebook? Just so that they are up to date with master in case any minor changes have occurred

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Performance Improvements
 Bug Fixes
 
 - Fixes bug in ``auglag`` optimizers which prevented them from accepting solver hyperparameters.
+- Adjusts the `quad_weights` of `CoilCurvature` and `CoilTorsion` to ensure their outputs are roughly independent of grid resolution.
```
YigitElma: Is this just a fix for these objectives? I thought this changed almost all of the coil objectives.

tests/test_objective_funs.py
```diff
@@ -3807,11 +3807,15 @@ def test_compute_scalar_resolution_coils(self, objective):
         f = np.zeros_like(self.res_array, dtype=float)
         for i, res in enumerate(self.res_array):
             obj = ObjectiveFunction(
-                objective(coilset, grid=LinearGrid(N=int(5 + 3 * res))),
+                objective(coilset, grid=LinearGrid(N=int(5 + 3 * res)), target=1),
```
YigitElma, line 3812: Not that it matters, but any specific reason for target=1 instead of 0?

CHANGELOG.md
singh-jaydeep: The per-coil objectives like CoilLength override the quad_weights in _CoilObjective.build() in their own build methods (setting quad_weights to 1), so I had thought this change would not affect them.

tests/test_objective_funs.py
singh-jaydeep: Only reason was so that all coil objectives have nonzero compute_scalar values. If target=0, then things like CoilTorsion would return all zeros. I figured this was safer, but can change it (or document better).

CHANGELOG.md
YigitElma: Then maybe we can say, 
```suggestion
- Adjusts the `quad_weights` of coil objectives of type `_broadcast_input = "node"` to ensure their outputs are roughly independent of grid resolution.
```

YigitElma (commented):
Other than the additional constraint for mixed coilset optimization notebook seems to be very similar. Looks good to me. If you think the additional constraint is necessary I can approve.

YigitElma:
<p>ok ok, nevermind, I didn't see you removed that from objectives</p>

---

 <a href='https://app.reviewnb.com/PlasmaControl/DESC/pull/2302/discussion/#comment-5531282932'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

dpanici (changes requested):
I think notebook chnages are ~ok. Coil opt is sensitive to small things changing and we aren't going for state of the art results in the notebook.

I do want that comment in the notebook  next to the weight noting the old weight removed as it is unneeded, but other than that I'll approve once its done

dpanici (approved):

YigitElma (approved):

PR #2304, opened 2026-08-27
Option to use proper normalization for QS objectives

YigitElma:
During stage-1 optimization, if the volume or the average magnetic field is not constrained, the strength of the magnetic field can be reduced by inflating the equilibrium (while still having the same Psi). The actual definition of these objectives (for example, in Landreman&Paul 2022) already includes these normalizations. Our precomputed normalizations cannot prevent these kinds of "cheats" in some cases.

This PR adds the option to choose which equation to use by the new `scale_invariant` argument. The default keeps the old behavior.


dpanici (dismissed):
I'd prefer to just silently ignore normalize=True when the hatted metric is requested, as technically it IS normalized. Just set the normalization in the build to 1 in those cases

desc/objectives/_omnigenity.py
```diff
@@ -359,7 +397,13 @@ def compute(self, params, constants=None):
             profiles=constants["profiles"],
             helicity=constants["helicity"],
         )
-        return data["f_C"]
+        if self._mode == "fc":
+            return data["f_C"]
+        else:
+            Bave = surface_averages(
```
YigitElma: Change to local B instead of flux surface average

desc/objectives/_omnigenity.py
```diff
@@ -32,6 +33,13 @@ class QuasisymmetryBoozer(_Objective):
         Poloidal resolution of Boozer transformation. Default = 2 * eq.M.
     N_booz : int, optional
         Toroidal resolution of Boozer transformation. Default = 2 * eq.N.
+    mode : "fb" or "fb_hat", optional
```
YigitElma: Rename this to `scale_invariant` and make it bool

desc/objectives/_omnigenity.py
```diff
@@ -500,7 +560,13 @@ def compute(self, params, constants=None):
...
             profiles=constants["profiles"],
         )
-        return data["f_T"]
+        if self._mode == "ft":
+            return data["f_T"]
+        else:
+            Bave = surface_averages(
+                constants["transforms"]["grid"], q=data["|B|"], sqrt_g=data["sqrt(g)"]
+            )
+            return data["R0"] ** 2 * data["f_T"] / Bave**4
```
YigitElma: Use local R

YigitElma (commented):
Also add the equations in the docstring

tests/test_objective_funs.py
```diff
@@ -495,6 +495,79 @@ def test(eq):
...
+                assert obj._units == "(dimensionless)"
+                assert obj.normalization == 1
+            assert obj.compute_scaled_error(*obj.xs(eq)).size == obj.dim_f
+            f[mode] = obj.compute_unscaled(*obj.xs(eq))
+            assert np.all(np.isfinite(f[mode]))
+
+        # the harmonics on each surface are all scaled by the same factor, and the
+        # norm of the residuals on each surface should be between 0 and 1
+        ratio = (f["fb_hat"] / f["fb"]).reshape((rho.size, -1))
+        np.testing.assert_allclose(ratio / ratio[:, :1], 1, rtol=1e-12)
```
ddudt: [question](https://pullpo.io/cc?l=question) 
I don't understand what this is testing

YigitElma: That was testing if the normalization is constant per surface but yeah it is not necessary and a weird way to test. I changed it to just checking between 0 and 1

desc/objectives/_omnigenity.py
f0uriest, line 1: the new "scale_invariant" option seems like it might be confusing with "normalize", would it be cleaner to just make "scale_invariant" an option of "normalize"? ie
``normalize=True -> use fixed scaling``
``normalize="scale_invariant" -> use dynamic scaling``
or could make it `normalize="fixed` vs `normalize="dynamic"`

Another option would be to define a new compute qty for the normalized values, similar to what we did for ``F_normalized``?

YigitElma: I will try to create new compute quantities

desc/plotting.py
YigitElma, line 1: The normalization we use here is different than the new quantities. This function uses individual flux surface averages, then computes normalization, whereas if we used the new compute functions, it would be the flux surface average of the normalized output. If we want to have that, I will also need to update the `Basic QS Optimization` notebook.

YigitElma:
We discussed this during issue triage. Dario and Daniel agree that we can change the normalization used in the plotting function `plot_qs_error` to be consistent. What do you think @f0uriest ? I can update the function, notebook and couple benchmark images, in that case.

docs/notebooks/tutorials/qs_initial_guess.h5
YigitElma, line 1: I just loaded/saved this equilibrium because it lacked the `ion_density` profile.

desc/compute/_omnigenity.py
```diff
@@ -776,6 +796,84 @@ def _f_T(params, transforms, profiles, data, **kwargs):
...
 
 
+@register_compute_fun(
+    name="f_T_normalized",
+    label="\\frac{R^2 \\nabla \\psi \\times \\nabla B \\cdot \\nabla "
+    + "(\\mathbf{B} \\cdot \\nabla B)}{B^4}",
+    units="~",
+    units_long="None",
+    description="Triple product quasisymmetry metric, normalized by the major "
+    "radius and the local field strength",
```
ddudt:
```suggestion
    description="Triple product quasisymmetry metric, normalized by the cylindrical R "
    "coordinate and the local field strength",
```
to avoid confusion with `R0`

desc/compute/_omnigenity.py
```diff
@@ -776,6 +796,84 @@ def _f_T(params, transforms, profiles, data, **kwargs):
...
+    resolution_requirement="tz",
+    grid_requirement={"is_meshgrid": True, "sym": False},
+    matrix="ndarray: Transform matrix from the double-Fourier coefficients to the "
+    "double-angle coefficients, as returned by ``ptolemy_linear_transform``.",
+    idx="ndarray: Indices of the symmetry breaking modes, as returned by "
+    "``ptolemy_linear_transform``.",
+)
+def _f_B(params, transforms, profiles, data, **kwargs):
+    B_mn = data["|B|_mn_B"].reshape((transforms["grid"].num_rho, -1))
+    # B_mn has shape (num modes, num rho)
```
ddudt: Is this comment is about the following line? Because on the previous line it has shape (num_rho, num_modes). Maybe add comments on both lines to clarify

desc/compute/_omnigenity.py
```diff
@@ -776,6 +796,84 @@ def _f_T(params, transforms, profiles, data, **kwargs):
...
+    resolution_requirement="tz",
+    grid_requirement={"is_meshgrid": True, "sym": False},
+    matrix="ndarray: Transform matrix from the double-Fourier coefficients to the "
+    "double-angle coefficients, as returned by ``ptolemy_linear_transform``.",
+    idx="ndarray: Indices of the symmetry breaking modes, as returned by "
+    "``ptolemy_linear_transform``.",
+)
+def _f_B_normalized(params, transforms, profiles, data, **kwargs):
+    B_mn = data["|B|_mn_B"].reshape((transforms["grid"].num_rho, -1))
+    # B_mn has shape (num modes, num rho), normalize each surface
```
ddudt: see previous comment

desc/compute/_omnigenity.py
```diff
@@ -776,6 +796,84 @@ def _f_T(params, transforms, profiles, data, **kwargs):
...
+    grid_requirement={"is_meshgrid": True, "sym": False},
+    matrix="ndarray: Transform matrix from the double-Fourier coefficients to the "
+    "double-angle coefficients, as returned by ``ptolemy_linear_transform``.",
+    idx="ndarray: Indices of the symmetry breaking modes, as returned by "
+    "``ptolemy_linear_transform``.",
+)
+def _f_B(params, transforms, profiles, data, **kwargs):
+    B_mn = data["|B|_mn_B"].reshape((transforms["grid"].num_rho, -1))
+    # B_mn has shape (num modes, num rho)
+    B_mn = kwargs["matrix"] @ B_mn.T
```
ddudt: This might be more clear: 
```suggestion
    # B_mn has shape (num modes, num rho)
    B_mn = data["|B|_mn_B"].reshape((transforms["grid"].num_rho, -1)).T
    B_mn = kwargs["matrix"] @ B_mn
```

desc/objectives/_omnigenity.py
```diff
@@ -384,13 +430,27 @@ def helicity(self, helicity):
 class QuasisymmetryTripleProduct(_Objective):
     """Quasi-symmetry triple product error.
 
+    With B = ||𝐁||:
+
+    f_T = ∇ψ × ∇B ⋅ ∇(𝐁 ⋅ ∇B)  (T⁴/m²)
+
+    With ``scale_invariant`` this is made dimensionless with the major radius and
+    the local field strength:
```
ddudt:
```suggestion
    With ``scale_invariant`` this is made dimensionless with the local cylindrical
    radius and field strength:
```

ddudt (changes requested):
minor suggestions to improve clarity

desc/compute/_omnigenity.py
YigitElma: I think they have the same shape, even after the matrix multiplication (at least for the Boozer objective). I don't know if there is any case for the Boozer transform in which this doesn't hold

desc/compute/_omnigenity.py
YigitElma: Updated the comment

desc/compute/_omnigenity.py
ddudt: The matrix multiplication preserves the shape, but my point was that on the first line it is shape (num_rho, num_modes) and then in the next line we transpose it to (num_modes, num_rho). In my other comment I suggested changing the transposes so that `B_mn` is always the same shape. I just think it makes the code easier to follow if the same variable doesn't change shape.

YigitElma: Oh ok, I didn't see the transpose. I think my new comment should be correct.

ddudt: I still think it's confusing:
```
B_mn = data["|B|_mn_B"].reshape((transforms["grid"].num_rho, -1))  # shape = (num_rho, num_modes)
B_mn = kwargs["matrix"] @ B_mn.T  # shape = (num_modes, num_rho)
```
That's why I suggested changing it to:
```
# B_mn has shape (num modes, num rho)
B_mn = data["|B|_mn_B"].reshape((transforms["grid"].num_rho, -1)).T
B_mn = kwargs["matrix"] @ B_mn
```

ddudt (approved):

dpanici (approved):

